PR #362, opened 2020-02-06
Copy over pull secret to guardian namespace

Brian-McM:


bcreane:
@Brian-McM will this also go into `release-v1.2` branch (for Enterprise v2.7.0)?

Brian-McM:
@bcreane yes it does

PR #365, opened 2020-02-07
Create sa and rbac for managed cluster's compliance server, so it can…

rene-dekker:
Create sa and rbac for managed cluster's compliance server, so it can check rbac

### DIFF
--- a/pkg/render/compliance.go
+++ b/pkg/render/compliance.go
@@ -125,8 +125,8 @@ func (c *complianceComponent) Objects() []runtime.Object {
 		c.complianceGlobalReportPolicyAudit(),
 		c.complianceGlobalReportCISBenchmark(),
 
-		// Even though we do not have a compliance-server, we need a serviceaccount
-		// in the managed cluster for RBAC requests.
+		// We always need a sa and crb, whether a deployment of compliance-server is present or not.
+		// These two are used for rbac checks for managed clusters.
 		c.complianceServerServiceAccount(),
 		c.complianceServerClusterRoleBinding(),
 	)
--- a/pkg/render/compliance.go
+++ b/pkg/render/compliance.go
@@ -136,6 +136,13 @@ func (c *complianceComponent) Objects() []runtime.Object {
 			c.complianceServerService(),
 			c.complianceServerDeployment(),
 		)
+	} else {
+		// We still need a serviceaccount in the managed cluster for RBAC requests.
+		complianceObjs = append(complianceObjs,
+			c.complianceServerServiceAccount(),
+			c.complianceServerManagedClusterRole(),
+			c.complianceServerClusterRoleBinding(),
+		)
 	}
 
 	if c.openshift {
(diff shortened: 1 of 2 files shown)


pkg/render/compliance.go
```diff
@@ -136,6 +136,13 @@ func (c *complianceComponent) Objects() []runtime.Object {
 			c.complianceServerService(),
 			c.complianceServerDeployment(),
 		)
+	} else {
+		// We still need a serviceaccount in the managed cluster for RBAC requests.
+		complianceObjs = append(complianceObjs,
+			c.complianceServerServiceAccount(),
```
Brian-McM: Since the complianceServerServiceAccount() and complianceServerClusterRoleBinding() need to be in every cluster, could you just add them to the original list of objs please?

pkg/render/compliance.go
```diff
@@ -136,6 +136,13 @@ func (c *complianceComponent) Objects() []runtime.Object {
...
+			c.complianceServerManagedClusterRole(),
```
Brian-McM: Are you sure that this is only needed for a managed cluster? Don't we have subject access reviews in the management / standalone clusters as well?

pkg/render/compliance.go
rene-dekker: Alina argued that this might be more clear/maintainable as it is explicitly defined here, so I went along with that. I'm fine either way

pkg/render/compliance.go
rene-dekker: complianceServerClusterRole() Also contains these same permissions, as well as a few others that a managed cluster does not need.

pkg/render/compliance.go
Brian-McM: The intention of this block was to put in objects that were different in Manged / Management cases. I don't think it really improves readability, personally I think it promotes confusion as to why we put these in two places, especially if we need to remove / change the service account or role binding, somebody might think that they should only change it for one of the cases and leave it the same for the other.

pkg/render/compliance.go
Brian-McM: Gotcha

pkg/render/compliance.go
```diff
@@ -452,6 +459,20 @@ func (c *complianceComponent) complianceServerClusterRole() *rbacv1.ClusterRole
 	}
 }
 
+func (c *complianceComponent) complianceServerManagedClusterRole() *rbacv1.ClusterRole {
+	return &rbacv1.ClusterRole{
+		TypeMeta:   metav1.TypeMeta{Kind: "ClusterRole", APIVersion: "rbac.authorization.k8s.io/v1"},
+		ObjectMeta: metav1.ObjectMeta{Name: "tigera-compliance-server"},
+		Rules: []rbacv1.PolicyRule{
+			{
+				APIGroups: []string{"authorization.k8s.io"},
```
asincu: We need to allow get and list for `globalreporttypes`.

Brian-McM:
Thinking on this again, are you sure we don't want to just apply Role for the management cluster @rene-dekker? Are we sure we won't need the other rules as well?

pkg/render/compliance.go
asincu: I suggested this so we always have the service account, role and binding together to avoid confusion on how they are applied.

Brian-McM: What exactly do you think somebody will be confused about though?

pkg/render/compliance.go
```diff
@@ -124,18 +124,25 @@ func (c *complianceComponent) Objects() []runtime.Object {
 		c.complianceGlobalReportNetworkAccess(),
 		c.complianceGlobalReportPolicyAudit(),
 		c.complianceGlobalReportCISBenchmark(),
+
+		// Even though we do not have a compliance-server, we need a serviceaccount
```
tmjd: This comment is a little misleading I think. "Even though we do not have a compliance-server", this code is common to both managed and management. Maybe it should be "Even if we do not have a compliance-server".

pkg/render/compliance.go
rene-dekker: Well I agree that it might be weird to see the sa, cr and crb without a deployment. Someone not knowing what is going on might think that this is a mistake. I now clarified this with a comment, so I think we are safe to close this discussion.

PR #367, opened 2020-02-07
(CNX-11672) Increase eck operator memory to prevent OCP OOMKILL

Brian-McM:
The memory limit was just a little to low


matthewdupre:
Is this enough?  Should we go to 150 request and 200 limit?  It feels like we have reasonable grounds to expect it not to work with only 20MiB?

Maybe even 192 request 256 limit?

Brian-McM:
@matthewdupre these are the new limits in the manifest ECK published, and after testing it it seems stable at 150. I can of course increase it if you think it's necessary.

bcreane:
@Brian-McM , @matthewdupre - merge or back to the drawing board?

Brian-McM:
merge if @matthewdupre is ok with it

PR #371, opened 2020-02-08
add configurable node metrics port

rafaelvanoni:


pkg/controller/installation/core_controller.go
```diff
@@ -311,6 +311,11 @@ func fillDefaults(instance *operator.Installation) error {
 		}
 	}
 
+	if instance.Spec.NodeMetricsPort != nil && *instance.Spec.NodeMetricsPort == 0 {
```
caseydavenport: I'm not sure I understand the intention here. IIUC, if the user explicitly writes in `nodeMetricsPort: 0` then we'll assign port `9081` instead. Is that right?

pkg/render/node.go
```diff
@@ -398,6 +395,15 @@ func (c *nodeComponent) nodeDaemonset() *apps.DaemonSet {
 	annotations[typhaCAHashAnnotation] = AnnotationHash(c.typhaNodeTLS.CAConfigMap.Data)
 	annotations[nodeCertHashAnnotation] = AnnotationHash(c.typhaNodeTLS.NodeSecret.Data)
 
+	if c.cr.Spec.NodeMetricsPort != nil {
+		scrapingAnnotation := map[string]string{
+			"prometheus.io/scrape": "true",
+			"prometheus.io/port":   string(*c.cr.Spec.NodeMetricsPort),
+		}
+
+		annotations[prometheusHashAnnotation] = AnnotationHash(scrapingAnnotation)
```
caseydavenport: I don't think we need the annotation hash entry here for the scrape port. If the port changes, we're already going to be modifying an environment variable and the other annotation, meaning we'll trigger a rolling update without the need for the special hash annotation.

pkg/render/node.go
```diff
@@ -945,16 +954,26 @@ func (c *nodeComponent) nodeMetricsService() *v1.Service {
...
-					TargetPort: intstr.FromInt(int(nodeMetricsPort)),
-					Protocol:   v1.ProtocolTCP,
-				},
-			},
 		},
 	}
+
+	// If the caller hasn't specified a NodeMetricsPort, default it to 9081.
+	if c.cr.Spec.NodeMetricsPort != nil {
+		if *c.cr.Spec.NodeMetricsPort == 0 {
```
caseydavenport: I don't think this check belongs here in the render code. Mostly because the render code should be doing exactly what the user has specified in the CR (i.e., if the user specified `0`, then we should use port `0`).

pkg/apis/operator/v1/types.go
```diff
@@ -55,6 +55,10 @@ type InstallationSpec struct {
 	// +optional
 	// +kubebuilder:validation:Enum=Standalone,Management,Managed
 	ClusterManagementType ClusterManagementType `json:"clusterManagementType,omitempty"`
+
+	// NodeMetricsPort allows users to specify which port calico/node serves metrics on or "nil" to disable it.
```
ozdanborne: These comments appear in our docs as well. I'd suggest avoiding the word `nil` because it is go-specific whereas these comments apply to the yaml as well, and avoid referring to the 'user' as our other documentation does not.
```suggestion
	// NodeMetricsPort specifies which port calico/node serves metrics on. If omitted, then metrics are disabled.
```

pkg/render/node.go
```diff
@@ -932,10 +942,9 @@ func (c *nodeComponent) nodeLivenessReadinessProbes() (*v1.Probe, *v1.Probe) {
 	return lp, rp
 }
 
-// nodeMetricsService creates a Service which exposes the calico/node metrics
-// reporting endpoint.
+// nodeMetricsService creates a Service which exposes the calico/node metrics reporting endpoint.
 func (c *nodeComponent) nodeMetricsService() *v1.Service {
-	return &v1.Service{
+	service := v1.Service{
```
ozdanborne: This code seems to always create a service for the nodeMetrics, but conditionally attach ports to it. Is that right? I think we want to delete the service if we're not using it.

pkg/controller/installation/core_controller.go
rafaelvanoni: Yes... that's my understanding, we use "0" to mean "use the default value".

pkg/render/node.go
rafaelvanoni: Ok

pkg/render/node.go
rafaelvanoni: This was needed for the UTs, they call into `Render()` directly.

pkg/render/node.go
rafaelvanoni: Ok

caseydavenport: @ozdanborne sadly, simply removing the service and returning `nil` here won't delete the service - we'd need to write special-case logic to delete it.

I had the same thought, but I believe what @rafaelvanoni has written here right now is a good pragmatic solution. If we simply returned nil here, then going from a port to no port would result in the service remaining unchanged.

pkg/render/node.go
caseydavenport: I think we should adjust the UTs then, since they appear to not be a realistic test :)

pkg/controller/installation/core_controller.go
caseydavenport: I was imagining this behavior, really since it's a more intuitive / obvious UX.

- If a port is specified, then we use that port. 
- If no port is specified, then we disable metrics.

Is there really a use-case for "Use whatever port the code has arbitrarily chosen as the default"? @ozdanborne WDYT?

ozdanborne: +1 what casey said

pkg/render/node.go
ozdanborne: Ah neat. SGTM

caseydavenport: FWIW, this wouldn't be too hard. There's another PR that does this exact thing: https://github.com/tigera/operator/pull/382

pkg/controller/installation/core_controller.go
rafaelvanoni: Fixed

pkg/render/node.go
rafaelvanoni: Ok

pkg/render/node.go
rafaelvanoni: Ok

pkg/render/node.go
```diff
@@ -56,31 +52,41 @@ type nodeComponent struct {
...
 		c.nodeRole(),
 		c.nodeRoleBinding(),
 	}
+
+	var objsToDelete []runtime.Object
+
 	if c.cr.Spec.Variant == operator.TigeraSecureEnterprise {
 		// Include Service for exposing node metrics.
-		objs = append(objs, c.nodeMetricsService())
+		metricsServiceToCreate, metricsServiceToDelete := c.nodeMetricsService()
```
caseydavenport: I think this would be simpler to follow if we just keyed it off the existence of the `nodeMetricsPort`. In its current form, the logic for deciding if the metrics service should be created or deleted is split between here and the nodeMetricsService function. Refactoring like below will keep that logic here, and keep `nodeMetricsService`'s scope to be just "Give me a service based on my config". 

e.g., 

```
if nodeMetricsPort != nil {
  // Append returned service to objects to create
} else {
 // Append returned service to objects to delete.
}
```

rafaelvanoni: Ok

PR #374, opened 2020-02-10
Support configuring block size on IP pools

caseydavenport:


pkg/apis/operator/v1/types.go
```diff
@@ -211,6 +211,12 @@ type IPPool struct {
 	// Default: 'all()'
 	// +optional
 	NodeSelector string `json:"nodeSelector,omitempty"`
+
+	// BlockSize specifies the CIDR prefex length to use when allocating per-node IP blocks from
+	// the main IP pool CIDR.
+	// Default: 26
```
tmjd: We should specify the default for IPv4 and IPv6

pkg/controller/installation/core_controller.go
```diff
@@ -285,29 +285,40 @@ func fillDefaults(instance *operator.Installation) error {
...
-			// Default IPv4 address detection to "first found" if not specified.
 			if instance.Spec.CalicoNetwork.NodeAddressAutodetectionV4 == nil {
+				// Default IPv4 address detection to "first found" if not specified.
 				t := true
 				instance.Spec.CalicoNetwork.NodeAddressAutodetectionV4 = &operator.NodeAddressAutodetection{
 					FirstFound: &t,
 				}
 			}
+			if v4pool.BlockSize == nil {
+				var twentySix int32 = 26
```
tmjd: The variable name feels weird to me. (not requesting you change it, just a comment)

pkg/controller/installation/core_controller.go
```diff
@@ -329,7 +340,7 @@ func mergeProvider(cr *operator.Installation, provider operator.Provider) error
 	if cr.Spec.KubernetesProvider == operator.ProviderNone {
 		cr.Spec.KubernetesProvider = provider
 	}
-	log.WithValues("provider", cr.Spec.KubernetesProvider).Info("Determined provider")
+	log.WithValues("provider", cr.Spec.KubernetesProvider).V(1).Info("Determined provider")
```
tmjd: I would like to use constants in this situation because I'm not sure/confident on what 'level' a 1 is and I think constants would help us have a better shared understanding when deciding what level a message should be at. (I'm not suggesting you change in this PR but think we should perhaps go through and make that change separately.)

pkg/controller/installation/defaults_test.go
```diff
@@ -75,17 +80,29 @@ var _ = Describe("Defaulting logic tests", func() {
...
+							NATOutgoing:   "Enabled",
+							NodeSelector:  "has(thiskey)",
+							BlockSize:     &twentySeven,
+						},
+						{
+							CIDR:          "fd00::0/64",
+							Encapsulation: "None",
+							NATOutgoing:   "Enabled",
+							NodeSelector:  "has(thiskey)",
+							BlockSize:     &twentySeven,
```
tmjd: hmmm, this should be an error I think.

pkg/controller/installation/validation.go
```diff
@@ -95,7 +98,11 @@ func validateCustomResource(instance *operatorv1.Installation) error {
 			}
 
 			if v6pool.NodeSelector == "" {
-				return fmt.Errorf("ipPool.nodeSelector, should not be empty")
+				return fmt.Errorf("ipPool.nodeSelector should not be empty")
+			}
+
+			if v6pool.BlockSize != nil && (*v6pool.BlockSize > 128 || *v6pool.BlockSize < 116) {
+				return fmt.Errorf("ipPool.blockSize must be greater than 115 and less than or equal to 128")
```
tmjd: Should we validate that the blocksize is bigger than the cidr?

pkg/controller/installation/core_controller.go
tmjd: After some reading I think this should be 0 level. I think that is something we would always want to see in the logs.
https://github.com/go-logr/logr#how-do-i-choose-my-v-levels

pkg/render/node.go
```diff
@@ -773,45 +776,34 @@ func (c *nodeComponent) nodeEnvVars() []v1.EnvVar {
...
+				nodeEnv = append(nodeEnv, v1.EnvVar{Name: "CALICO_IPV4POOL_IPIP", Value: "Never"})
 			default:
-				nodeEnv = append(nodeEnv, v1.EnvVar{Name: "CALICO_IPV4POOL_IPIP",
-					Value: "Always"})
+				nodeEnv = append(nodeEnv, v1.EnvVar{Name: "CALICO_IPV4POOL_IPIP", Value: "Always"})
 			}
 
-			// Default for NAT Outgoing is enabled so it is only necessary to
-			// set when it is being disabled.
+			if v4pool.BlockSize != nil {
```
tmjd: This check shouldn't be needed since we default this.... but I think I prefer being defensive and keeping this check. (no change requested, just a comment)

pkg/controller/installation/core_controller.go
caseydavenport: The problem with this one is that it's logged every loop of the controller, which actually happens a lot, even if nothing is happening. It adds a lot of noise and log spam. 

Wonder if we can ratelimit, or log it somewhere else that's not in the main loop.

pkg/controller/installation/defaults_test.go
caseydavenport: Ah yeah, but we're not running the validation code in this test (just defaulting logic).

PR #375, opened 2020-02-10
For Openshift install put CNI config where multus expects

tmjd:


pkg/render/node.go
```diff
@@ -427,7 +427,7 @@ func (c *nodeComponent) cniDirectories() (string, string) {
 	var cniBinDir, cniNetDir string
 	switch c.provider {
 	case operator.ProviderOpenShift:
-		cniNetDir = "/etc/kubernetes/cni/net.d"
+		cniNetDir = "/var/run/multus/cni/net.d"
```
caseydavenport: I'm guessing we're going to need to set this as well: https://github.com/projectcalico/cni-plugin/blob/e1a1ea6834cad8235db6d0e87901cd144d60e265/k8s-install/scripts/install-cni.sh#L28

Or, adjust the `cniConfigMap` to have an explicit reference to the kubeconfig file rather than use `__KUBECONFIG_FILEPATH__`, which is auto-calculated based on expected absolute path on the host. https://github.com/tigera/operator/blob/5fa79b2b46e382c7f16c5f8a767f16030d097c6a/pkg/render/node.go#L298

tmjd: the install-cni script replaces that with `${HOST_CNI_NET_DIR}"/calico-kubeconfig` and HOST_CNI_NET_DIR is CNI_NET_DIR and we set CNI_NET_DIR to the `cniNetDir`.
Does that work for what you're thinking then?

PR #380, opened 2020-02-11
(SAAS-664) Make sure Elasticsearch / Kibana are cleaned up after LogStorage CR is deleted for all cluster types

Brian-McM:
If the cluster type was Managed, then the code path that cleans up Elasticsearch and Kibana before removing the finalizers on the LogStorage CR wasn't followed.

This commit also does the following:

- Update the Elasticsearch and Kibana watches to not use EnqueueRequestForOwner (so that we can tell when we're reconciling for elasticsearch or kibana)
- Add a watch for the ECK statefulset

### DIFF
--- a/pkg/controller/logstorage/logstorage_controller.go
+++ b/pkg/controller/logstorage/logstorage_controller.go
@@ -21,19 +21,25 @@ import (
 	"os"
 	"regexp"
 
+	apps "k8s.io/api/apps/v1"
+
+	"github.com/elastic/cloud-on-k8s/operators/pkg/utils/stringsutil"
+
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 
 	esalpha1 "github.com/elastic/cloud-on-k8s/operators/pkg/apis/elasticsearch/v1alpha1"
-	kibanaalpha1 "github.com/elastic/cloud-on-k8s/operators/pkg/apis/kibana/v1alpha1"
 	"github.com/go-logr/logr"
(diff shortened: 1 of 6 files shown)


pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -257,6 +267,24 @@ func (r *ReconcileLogStorage) Reconcile(request reconcile.Request) (reconcile.Re
...
+		}
+	} else if ls.DeletionTimestamp != nil {
+		var errMsg string
+
+		err := r.finalizeDeletion(ctx, ls)
+		if err != nil {
+			errMsg = err.Error()
+		}
+
+		r.status.SetDegraded("Finalizing deletion of LogStorage before continuing", errMsg)
```
caseydavenport: Maybe I'm missing something, but won't this set degraded even if err == nil? Should we put this within the `err != nil` block above?

caseydavenport: I'm also a little bit confused about when `r.status.setDegraded` is meant to be used vs `r.setDegraded`

Brian-McM: You're not missing something, although I might not be doing the right thing. I thought that we would want to set the state as degraded since until we're finished cleaning up the resources.

I guess since it's there's nothing actually wrong maybe I shouldn't add this? I wanted to convey to the users that this is in a state of deletion in the case that it takes awhile for the ECK operator to clean up it's resources, but maybe logs are the way to go?

Brian-McM: > I'm also a little bit confused about when r.status.setDegraded is meant to be used vs r.setDegraded

I should be using r.setDegraded each time

PR #381, opened 2020-02-11
add configurable controlPlaneNodeSelector

rafaelvanoni:


pkg/apis/operator/v1/types.go
```diff
@@ -55,6 +55,11 @@ type InstallationSpec struct {
 	// +optional
 	// +kubebuilder:validation:Enum=Standalone,Management,Managed
 	ClusterManagementType ClusterManagementType `json:"clusterManagementType,omitempty"`
+
+	// ControlPlaneNodeSelector is used to select control plane nodes on which to run specific Calico
+	// components.
```
tmjd: Can you add a comment that this applies to kube-controllers and apiserver?

PR #382, opened 2020-02-12
Remove compliance server. Add test suite for compliance.

rene-dekker:
Remove compliance server for managed clusters. Add test suite for compliance.

### DIFF
--- a/pkg/controller/compliance/compliance_controller_test.go
+++ b/pkg/controller/compliance/compliance_controller_test.go
@@ -44,7 +44,6 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/apimachinery/pkg/runtime/schema"
 
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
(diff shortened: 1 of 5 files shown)


caseydavenport:
It would also be really nice to standardize on a way to do this across controllers, I think the need will come up more. 

I wonder if this decision belongs in the render package long term? e.g., `component.GetResourceToDelete(args...)`? 

No need to change for this PR, but thought I'd mention it.

Brian-McM:
> The way it's done in this PR, we would render and create it, and then delete it right afterwards. Better to simply never create it in the first place (and clean it up if we happened to create it on an earlier iteration).

That's already in here, https://github.com/tigera/operator/blob/bb7b953301bfe4fd4e1196d5501990a48203f17c/pkg/render/compliance.go#L135. This was only an issue when switching from a standalone / management cluster to a managed cluster

pkg/controller/compliance/compliance_controller_test.go
```diff
@@ -0,0 +1,196 @@
...
+	var c client.Client
+	var ctx context.Context
+	var r ReconcileCompliance
+	var scheme *runtime.Scheme
+
+	BeforeSuite(func() {
+		// The schema contains all objects that should be known to the fake client when the test runs.
+		scheme = runtime.NewScheme()
+		Expect(apis.AddToScheme(scheme)).NotTo(HaveOccurred())
+		scheme.AddKnownTypes(schema.GroupVersion{Group: "apps", Version: "v1"}, &appsv1.Deployment{})
```
Brian-McM: You should look at this https://github.com/tigera/operator/blob/bb7b953301bfe4fd4e1196d5501990a48203f17c/pkg/controller/logstorage/logstorage_controller_test.go#L47

No need to add each known type individually, the libraries already have that available

pkg/controller/compliance/compliance_controller_test.go
```diff
@@ -0,0 +1,196 @@
...
+		scheme.AddKnownTypes(schema.GroupVersion{Group: "", Version: "v1"}, &rbacv1.ClusterRole{})
+		scheme.AddKnownTypes(schema.GroupVersion{Group: "", Version: "v1"}, &rbacv1.ClusterRoleBinding{})
+		scheme.AddKnownTypes(schema.GroupVersion{Group: "", Version: "v1"}, &rbacv1.Role{})
+		scheme.AddKnownTypes(schema.GroupVersion{Group: "", Version: "v1"}, &rbacv1.RoleBinding{})
+		scheme.AddKnownTypes(schema.GroupVersion{Group: "", Version: "v1"}, &appsv1.DaemonSet{})
+		Expect(operatorv1.SchemeBuilder.AddToScheme(scheme)).NotTo(HaveOccurred())
+
+		// Create a client that will have a crud interface of k8s objects.
+		c = fake.NewFakeClientWithScheme(scheme)
```
Brian-McM: It may not matter for the tests you have, but since you have this in BeforeSuite that means anything created, updated, or deleted with this client in one test with will appear on the other tests (I think). If you use BeforeEach then this is run before each test, so each test has a clean slate when it runs.

caseydavenport:
@Brian-McM Ah, lovely. I missed that we were already doing that :+1:

pkg/controller/compliance/compliance_controller_test.go
rene-dekker: That is indeed a cleaner solution, thank you. I will resolve the conversation now, but the commit containing this change will be pushed later.

PR #386, opened 2020-02-12
Change render.Component.Objects() to return two object slices

rafaelvanoni:
The goal of this change is to allow methods in the render package to return a list of objects to be deleted in addition to the list of objects rendered. This has become necessary in a couple cases where a user specified parameter removes the need for a component or service. For example, when the user sets (the soon-to-be-introduced) nodeMetricsPort to nil, we should return that service for deletion.

### DIFF
--- a/pkg/controller/utils/component.go
+++ b/pkg/controller/utils/component.go
@@ -73,18 +73,15 @@ func (c componentHandler) CreateOrUpdate(ctx context.Context, component render.C
 	statefulsets := []types.NamespacedName{}
 	objsToCreate, objsToDelete := component.Objects()
 
-	var err error
-	var key client.ObjectKey
-
 	for _, obj := range objsToCreate {
 		// Set CR instance as the owner and controller.
-		if err = controllerutil.SetControllerReference(c.cr, obj.(metav1.ObjectMetaAccessor).GetObjectMeta(), c.scheme); err != nil {
+		if err := controllerutil.SetControllerReference(c.cr, obj.(metav1.ObjectMetaAccessor).GetObjectMeta(), c.scheme); err != nil {
 			return err
 		}
 
 		logCtx := ContextLoggerForResource(c.log, obj)
 		var old runtime.Object = obj.DeepCopyObject()
-		key, err = client.ObjectKeyFromObject(obj)
+		key, err := client.ObjectKeyFromObject(obj)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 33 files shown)


pkg/render/render.go
```diff
@@ -39,8 +39,9 @@ var (
 )
 
 type Component interface {
-	// Objects returns all objects this component contains.
-	Objects() []runtime.Object
+	// Objects returns the lists of objects in this component that should be created and/or deleted during
+	// rendering.
+	Objects() ([]runtime.Object, []runtime.Object)
```
rene-dekker: Maybe we can name the returned params for clarity. Ex:
`Objects() (objsToCreate, objsToDelete []runtime.Object)`

pkg/controller/utils/component.go
```diff
@@ -147,6 +149,16 @@ func (c componentHandler) CreateOrUpdate(ctx context.Context, component render.C
...
 		status.AddStatefulSets(statefulsets)
 	}
+
+	for _, obj := range objsToDelete {
+		err := c.client.Delete(ctx, obj)
+		if err != nil {
+			logCtx := ContextLoggerForResource(c.log, obj)
+			logCtx.Error(err, "Error deleting object %v", obj)
+			return err
+		}
```
caseydavenport: We'll need to update the status manager with the deleted resources, similar to this bit of code: https://github.com/tigera/operator/pull/386/files#diff-73950099f4736837cc6709a49bcb595aR147-R151

But instead using the `status.Remove` equivalent (or whatever it's called).

pkg/controller/utils/component.go
```diff
@@ -71,16 +71,20 @@ func (c componentHandler) CreateOrUpdate(ctx context.Context, component render.C
 	daemonSets := []types.NamespacedName{}
 	deployments := []types.NamespacedName{}
 	statefulsets := []types.NamespacedName{}
-	for _, obj := range component.Objects() {
+	objsToCreate, objsToDelete := component.Objects()
+
+	var err error
```
caseydavenport: I don't think you need to move these outside here. You can probably actually just delete line 82, since the key is only used within that loop IIUC.

rafaelvanoni: It's used on line 163 too

caseydavenport: Right, but we could just do a `:=` there as well and then we don't need to declare these up here at all, right?

rafaelvanoni: Sure.. this works as well, right?

caseydavenport: It totally works, it's just a slightly confusing / non-standard structure.  Declaring variables up-front like this usually implies that the value assigned to the variable in some earlier context will be used in a later on, but that's not what we're doing here.

I'm just suggesting the change so that future readers don't think that there's some special link between the usage of key in the first loop and the usage in the second.

rafaelvanoni: Fair enough

PR #388, opened 2020-02-13
Remove lingering webhook created by eck when clustertype -> managed

rene-dekker:
SAAS-668
The ECK operator creates a webhook without owner refs, so the resource does not get deleted automatically.


pkg/controller/compliance/compliance_controller_test.go
```diff
@@ -1,22 +1,10 @@
-// Copyright (c) 2019 Tigera, Inc. All rights reserved.
-//
-// Licensed under the Apache License, Version 2.0 (the "License");
```
rene-dekker: This was somehow duplicated.

PR #390, opened 2020-02-14
Combine Management /  Managed cluster logic for LogStorage at the con…

Brian-McM:
…troller / render levels

This commit combines the separate logic for the Management / Managed cluster into one code path, instead of having the multiple render / controller paths. This was now possible with the addition of render components returning what they want deleted, and these changes greatly simplify the LogStorage logic over all

This commit fix ticket SAAS-664 as well, where we weren't cleaning up Elasticsearch correctly in the Managed Clusters


pkg/controller/status/status.go
```diff
@@ -301,75 +301,67 @@ func (m *statusManager) syncState() bool {
 	defer m.lock.Unlock()
 	progressing := []string{}
 	failing := []string{}
-	numDaemonSets := len(m.daemonsets)
```
Brian-McM: Some of this isn't really necessary, the num* fields weren't really used and the if len(.... was unnecessary and the cron job loop didn't have it, so I made it consistent.

I noticed this when I went to add the cron job count to the if statement at the bottom

pkg/controller/utils/component.go
```diff
@@ -159,13 +165,17 @@ func (c componentHandler) CreateOrUpdate(ctx context.Context, component render.C
...
 
 		key, err := client.ObjectKeyFromObject(obj)
-		switch obj.(type) {
-		case *apps.Deployment:
-			status.RemoveDeployments(key)
-		case *apps.DaemonSet:
-			status.RemoveDaemonsets(key)
-		case *apps.StatefulSet:
-			status.RemoveStatefulSets(key)
+		if status != nil {
```
Brian-McM: noticed this when adding the cronjob logic

pkg/render/common.go
```diff
@@ -131,6 +131,18 @@ func setServerAuth(x *x509.Certificate) error {
 	return nil
 }
 
+func CreateOperatorTLSSecret(
```
Brian-McM: I created wrapped exported functions to minimise the number of changes for this PR. I can, of course, just change the unexport function to an exported one but then there will be a lot more files to review.

pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -257,48 +275,276 @@ func (r *ReconcileLogStorage) Reconcile(request reconcile.Request) (reconcile.Re
...
 		return reconcile.Result{}, err
 	}
 
-	if network.Status.Variant != operatorv1.TigeraSecureEnterprise {
+	// These checks ensure that we're in the correct state to continue to the render function without causing a panic
+	if installationCR.Status.Variant != operatorv1.TigeraSecureEnterprise {
 		r.status.SetDegraded(fmt.Sprintf("Waiting for network to be %s", operatorv1.TigeraSecureEnterprise), "")
 		return reconcile.Result{}, nil
+	} else if ls == nil && installationCR.Spec.ClusterManagementType != operatorv1.ClusterManagementTypeManaged {
+		err := fmt.Errorf("LogStorage must exist for '%s' cluster type", installationCR.Spec.ClusterManagementType)
```
caseydavenport: Like we discussed I think this is confusing since in this branch the TigeraStatus object won't exist (due to the OnCRNotFound call above). 

Not a problem to call SetDegraded, per-se, but also doesn't add any value, so might be less confusing with a comment and/or removing the SetDegraded call.

pkg/render/common.go
caseydavenport: Appreciate this for the first round of review :+1: 

But longer term I don't see a reason to keep them separate functions - do you?

PR #392, opened 2020-02-19
Add toggle for FlexVol init container

sebrandon1:
Adds a configurable parameter to `CalicoNetworkSpec` to allow for the enabling/disabling (default: enabled) of the flexvol init container.


sebrandon1:
I have verified that this works via some internal testing.  Calico-Node comes up without the flexvol container with `flexVolInitContainerEnabled: false`.

PR #393, opened 2020-02-19
Migration patch fix

tmjd:
Update to v2.7.0 images
Fix patching daemonset to work when no nodeSelectors exist.

### DIFF
--- a/config/calico_versions.yml
+++ b/config/calico_versions.yml
@@ -1,11 +1,11 @@
 components:
   typha:
-    version: v3.11.1
+    version: v3.12.0
   calico/node:
-    version: v3.11.1-with-auto-backend
+    version: v3.12.0
   calico/cni:
-    version: v3.11.1
+    version: v3.12.0
   calico/kube-controllers:
-    version: v3.11.1
+    version: v3.12.0
   flexvol:
-    version: v3.11.1
+    version: v3.12.0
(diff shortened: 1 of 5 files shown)


pkg/components/versions.go
```diff
@@ -16,51 +16,51 @@ package components
...
 	// Overrides for Calico.
-	VersionTigeraNode            = "v2.7.0-0.dev-188-gc5255d3"
-	VersionTigeraTypha           = "v2.7.0-0.dev-74-g67bc0b1"
-	VersionTigeraKubeControllers = "v2.7.0-0.dev-109-gfa8b79c"
+	VersionTigeraNode            = "v2.7.0"
+	VersionTigeraTypha           = "v2.7.0"
+	VersionTigeraKubeControllers = "v2.7.0"
 
 	// API server images.
-	VersionAPIServer   = "v2.7.0-0.dev-32-gb094281f"
```
caseydavenport: Is this right? I think the existing v2.7.0-0.dev tags are more recent than the v2.7.0 release.

tmjd: Looking at the hash, `b094281f` is older than the tagged `v2.7.0`. So I think it is correct to update those.

caseydavenport: Oh, ok. yeah I think I misread the tag. :+1:

PR #394, opened 2020-02-19
accept empty string for installation.spec.kubernetesProvider

ozdanborne:


pkg/apis/operator/v1/types.go
```diff
@@ -41,7 +41,7 @@ type InstallationSpec struct {
 	// KubernetesProvider specifies a particular provider of the Kubernetes platform. This is often auto-detected.
 	// If specified, this enables provider-specific configuration and must match the auto-detected value (if any).
 	// +optional
-	// +kubebuilder:validation:Enum=EKS,GKE,AKS,OpenShift,DockerEnterprise
+	// +kubebuilder:validation:Enum=,EKS,GKE,AKS,OpenShift,DockerEnterprise
```
tmjd: We should add that empty string means default autodetection.

ozdanborne: Do you think "If specified..." implies this already? Most of the comments

tmjd: I feel like an empty value is something specified and could easily mean skip autodetection or something like that. Good point about the type though, I agree we shouldn't mention the type, just that an empty value means use the default autodetection.

PR #395, opened 2020-02-20
Adding doc for unsupported annotation.

tmjd:


README.md
```diff
@@ -106,3 +106,41 @@ To run all the unit tests, run:
 To run a specific test or set of tests, use the `GINKGO_FOCUS` argument.
 
 	make test GINKGO_FOCUS="component function tests"
+
+### Making temporary changes to components the operator manages
+
+The operator creates and manages resources and will reconcile them to be in the desired state. Due to the
```
caseydavenport: If we're going to document this here, we need to shout loud and clear that this is intended for development only and is not to be used in production clusters. Setting this annotation and using custom configuration can and will lead to unexpected and potentially catastrophic behavior during upgrade, etc.

README.md
```diff
@@ -106,3 +106,46 @@ To run all the unit tests, run:
...
+reconciliation it does, if a user makes direct changes to a resource the operator will revert those changes.
+To enable the user to make temporary changes, an annotation can be added to any resource directly managed by
+the operator which will cause the operator to no longer update the resource.
+Adding the following as an annotation to any resource will prevent the operator from making any future updates to the annotated resource:
+
+  *This annotation should only be temporarily added since it blocks the operator from it intended purpose.
```
caseydavenport: i think I'd like this to be stronger. It shouldn't say "should only be temporarily added", rather it should say "do not use this unless you are a developer working on the operator".

PR #397, opened 2020-02-20
[Cherry Pick] Enforce resource requirements for ElasticSearch pod.

Brian-McM:
* Add default request and limit values for for ES pod (to avoid ES pod eviction problem)
* Apply LogStorage resource requirements to ES pod (to allow user to override defaults)
* Apply explicit JVM heap size min and max arguments within ES pod

Origin commit https://github.com/tigera/operator/commit/a681835dbdf175fd571dbdddb24bbed303c4eacf


Brian-McM:
That's sort of correct. This uses the resource API fields in the pods resource limits (as the documentation implies), previously they were only being applied to the PVC resource limits. This also adds reasonable default if the resource limits aren't there

PR #403, opened 2020-02-21
Adding image path config

tmjd:

PR #412, opened 2020-02-25
Adding image path config

tmjd:
This is a "pick" of #403 
The pick needed a bit of rework since the image handling changed quite a bit between release-v1.2 and master.


tmjd:
I tested this by installing on a cluster created with `make cluster-create` and then putting all the images into my own repo and configured the Installation resource with a registry and imagePath and things just worked. I also verified after installation that the images were being pulled from my configured registry/imagePath by doing describe pod on several of the pods.

PR #416, opened 2020-02-25
set image tags to release-v2.7

bcreane:
Since hash releases are unreliable at the moment, let's have a fall-back for installing latest Tigera images. Note leaving Calico components (e.g. CNI) pinned to v3.12.0 for time being.


bcreane:
I verified this operator pulls correct version (`release-v2.7`) of Tigera images and the cluster is healthy.

caseydavenport:
Looks like this is fairly old and has some conflicts. Closing for now.

PR #417, opened 2020-02-26
custom FlexVolume path

rafaelvanoni:


sebrandon1:
This is fine.  As long as we are still able to disable it, that's fine with me.  👍

rafaelvanoni:
> This is fine. As long as we are still able to disable it, that's fine with me. +1

You can disable it by setting `FlexVolumePath` to "None".

pkg/controller/installation/core_controller.go
```diff
@@ -324,6 +320,24 @@ func fillDefaults(instance *operator.Installation) error {
...
+				instance.Spec.CalicoNetwork.FlexVolumePath = "/etc/kubernetes/kubelet-plugins/volume/exec/"
+			} else if instance.Spec.KubernetesProvider == operator.ProviderGKE {
+				instance.Spec.CalicoNetwork.FlexVolumePath = "/home/kubernetes/flexvolume/"
+			} else if instance.Spec.KubernetesProvider == operator.ProviderAKS {
+				instance.Spec.CalicoNetwork.FlexVolumePath = "/etc/kubernetes/volumeplugins/"
+			} else {
+				instance.Spec.CalicoNetwork.FlexVolumePath = "/usr/libexec/kubernetes/kubelet-plugins/volume/exec/"
+			}
+		} else if instance.Spec.CalicoNetwork.FlexVolumePath == "None" {
+			instance.Spec.CalicoNetwork.FlexVolumePath = ""
```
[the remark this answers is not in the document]
rafaelvanoni: I'm leaving the actual values out as this would likely be missed when/if the actual defaults change.

pkg/apis/operator/v1/types.go
```diff
@@ -121,10 +121,11 @@ type CalicoNetworkSpec struct {
 	// +optional
 	NodeAddressAutodetectionV6 *NodeAddressAutodetection `json:"nodeAddressAutodetectionV6,omitempty"`
 
-	// FlexVolInitContainerEnabled allows the toggling on/off of the FlexVol init container. If not specified,
-	// FlexVol will be enabled by default.
+	// FlexVolumePath optionally specifies a custom path for FlexVolume. If not specified, FlexVolume will be
+	// enabled by default. If set to 'None', FlexVolume will be disabled.
+	// Default: "/usr/libexec/kubernetes/kubelet-plugins/volume/exec/"
```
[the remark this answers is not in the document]
rafaelvanoni: I'm leaving the actual values out as this would likely be missed when/if the actual defaults change.

pkg/apis/operator/v1/types.go
```diff
@@ -66,6 +66,12 @@ type InstallationSpec struct {
 	// NodeMetricsPort specifies which port calico/node serves metrics on. If omitted, then metrics are disabled.
 	// +optional
 	NodeMetricsPort *int32 `json:"nodeMetricsPort,omitempty"`
+
+	// FlexVolumePath optionally specifies a custom path for FlexVolume. If not specified, FlexVolume will be
+	// enabled by default. If set to 'None', FlexVolume will be disabled. The default is based on the k8s
+	// provider.
```
caseydavenport: I think we should have validation on this field which asserts it is either "None", or a valid absolute path. WDYT?

e.g. "path/to/file" is not a valid option since it's a relative path. Neither is "none".

pkg/apis/operator/v1/types.go
```diff
@@ -66,6 +66,12 @@ type InstallationSpec struct {
...
+	// +optional
```
caseydavenport: I think we can add a kubebuilder tag here that will further validate, so users get errors from kubectl.

Something like this probably (you'll need to figure out the right regex to represent the valid options for this).

```
// +kubebuilder:validation:Pattern="None|^\/"
```

More info here: https://book.kubebuilder.io/reference/generating-crd.html

pkg/controller/installation/core_controller.go
```diff
@@ -325,6 +321,26 @@ func fillDefaults(instance *operator.Installation) error {
...
+			} else if instance.Spec.KubernetesProvider == operator.ProviderGKE {
+				instance.Spec.FlexVolumePath = "/home/kubernetes/flexvolume/"
+			} else if instance.Spec.KubernetesProvider == operator.ProviderAKS {
+				instance.Spec.FlexVolumePath = "/etc/kubernetes/volumeplugins/"
+			} else {
+				instance.Spec.FlexVolumePath = "/usr/libexec/kubernetes/kubelet-plugins/volume/exec/"
+			}
+		} else if instance.Spec.FlexVolumePath != "None" {
+			// Simple checks to make sure that the user provided a valid, absolute path to a directory.
+			if !path.IsAbs(instance.Spec.FlexVolumePath) {
```
caseydavenport: We have a separate function for validation - we should put that there rather than in `fillDefaults`

pkg/controller/installation/core_controller_test.go
```diff
@@ -90,10 +90,6 @@ var _ = Describe("Testing core-controller installation", func() {
 				Expect(i.Spec.CalicoNetwork.IPPools).To(HaveLen(0))
 				return
 			}
-			if calicoNet.FlexVolInitContainerEnabled == nil {
-				t := true
-				Expect(i.Spec.CalicoNetwork.FlexVolInitContainerEnabled).To(Equal(&t))
```
caseydavenport: Rather than just remove this test, can we replace the Expect() with the corresponding check in the new code? Same for the other tests where it's been removed.

pkg/render/node_test.go
```diff
@@ -57,6 +57,7 @@ var _ = Describe("Node rendering tests", func() {
 	})
 
 	It("should render all resources for a default configuration", func() {
+		defaultInstance.Spec.FlexVolumePath = "/usr/libexec/kubernetes/kubelet-plugins/volume/exec/"
```
caseydavenport: Might be nicer to just set this in the `BeforeEach` so that we don't need to do it in every It()?

rafaelvanoni: It's a different string depending on the platform

pkg/controller/installation/core_controller_test.go
rafaelvanoni: Not sure what you mean here, there already are render tests that check that a FlexVolume container was created given the correct string in `node_test.go`

caseydavenport: But this isn't a render test, this is a defaulting test. The check here is meant to check that the `caliconet.Spec.FlexVolumePath` gets defaulted to the correct value if it isn't set.

rafaelvanoni: I added those to `defaults_test.go`, do you want me to duplicate them here?

pkg/render/node_test.go
caseydavenport: Yeah, it defaults to a different string depending on the platform (though it can always be overridden by user config). I guess it's fine, but it confused me a bit.

pkg/controller/installation/defaults_test.go
```diff
@@ -38,8 +38,6 @@ var _ = Describe("Defaulting logic tests", func() {
 		Expect(*v4pool.BlockSize).To(Equal(int32(26)))
 		v6pool := render.GetIPv6Pool(instance.Spec.CalicoNetwork)
 		Expect(v6pool).To(BeNil())
-		t := true
```
caseydavenport: This is one of the other tests I was talking about - the old test validated that we properly defaulted the `FlexVolInitContainerEnabled` field. I think if we're removing that field and replacing it, we should add an equivalent test that makes sure the `FlexVolumePath` field is defaulted properly.

caseydavenport: Hm, I see there are other defaulting tests specifically for this field added below. That probably covers this, but threw me off since we've put all of the defaulting logic into this section in the past.

pkg/controller/installation/core_controller_test.go
caseydavenport: Nah that's ok. I was thrown off by them moving is all. So long as the tests exist I'm happy :+1:

PR #420, opened 2020-02-27
calico-node clusterrole: add secrets permissions

gianlucam76:
Such secrets permissions are needed to properly configure bgp password
in bgppeers


bcreane:
Likely want operator and manifest fixes cherry picked back v2.7 and v2.6, if required.

PR #421, opened 2020-02-27
Add config for gen-versions

tmjd:
- Include config files that can be used with gen-verisons
  so only one of the files needs to be specified instead of both.


tmjd:
This is targeted against release-v1.2, master already can build with updates to only EE or OS. After this is merged I expect to pick this to release-v1.1 and release-v1.0.

tmjd:
I also updated the generation template so it generates closer to what exists in the current version file.

PR #426, opened 2020-02-28
Update api comments so they render nicely

tmjd:
- Also switch instances of Tigera Secure to Calico Enterprise


tmjd:
These changes should render like this

image

PR #430, opened 2020-03-02
making changes to enable splunk as log destination

penkeysuresh:
* adding operator API spec for splunk storage
 * making changes to log collector controller
 * making changes to fluentd to render splunk env vars
 * adding test case to render splunk, updating others

The logic is pretty much similar to what is already done for `syslog` and `s3`. The test cases of other  destination updated for changes in `Fluentd`


pkg/controller/logcollector/logcollector_controller.go
```diff
@@ -226,6 +226,23 @@ func (r *ReconcileLogCollector) Reconcile(request reconcile.Request) (reconcile.
 		}
 	}
 
+	var splunkCredential *render.SplunkCredential
+	if instance.Spec.AdditionalStores != nil {
+		if instance.Spec.AdditionalStores.Splunk != nil {
+			splunkCredential, err = getSplunkCredential(r.client)
```
tmjd: We need a watch for this in the `add` function above to ensure if this changes the controller will re-Reconcile.

pkg/controller/logcollector/logcollector_controller.go
```diff
@@ -226,6 +226,23 @@ func (r *ReconcileLogCollector) Reconcile(request reconcile.Request) (reconcile.
...
+			if err != nil {
+				log.Error(err, "Error with Splunk credential secret")
+				r.status.SetDegraded("Error with Splunk credential secret", err.Error())
+				return reconcile.Result{}, err
+			}
+			if splunkCredential == nil {
```
tmjd: Is this required for all Splunk setups?
I don't know splunk config so this may be required, I just wanted to confirm.

pkg/controller/logcollector/logcollector_controller.go
penkeysuresh: Done, I added a watch along with the elastic secrets

pkg/controller/logcollector/logcollector_controller.go
penkeysuresh: @tmjd not sure.. can you elaborate the comment ?

tmjd: Is the splunkCredential required for all splunk endpoints? With this check, progress is blocked without the credential, is that expected?

pkg/render/fluentd_test.go
```diff
@@ -213,11 +214,148 @@ var _ = Describe("Tigera Secure Fluentd rendering tests", func() {
...
+			{name: "fluentd-node", ns: "tigera-fluentd", group: "apps", version: "v1", kind: "DaemonSet"},
+		}
+
+		i := 0
+		for _, expectedRes := range expectedResources {
+			ExpectResource(resources[i], expectedRes.name, expectedRes.ns, expectedRes.group, expectedRes.version, expectedRes.kind)
+			i++
+		}
+
+		ds := resources[3].(*apps.DaemonSet)
```
tmjd: Since this test is rendering with CA, I think it'd be good to check the volumes/volumeMounts.

pkg/controller/logcollector/logcollector_controller.go
tmjd: looking at getSplunkCredential is looks like the token is required but not the CA.

pkg/render/fluentd_test.go
penkeysuresh: added a generic check for volumes length in both test cases, explicate check with volume name in CA case.

pkg/controller/logcollector/logcollector_controller.go
penkeysuresh: Yes, once the user input the `SplunkSpec` in the `AdditionalStores` we require the token to be present, we don't expect the CA certificate. In `getSplunkCredential` we are checking for all the error cases are returning `nil` whenever there is an error case. 

Just so I get this correctly do we want to change the above to this ? This check is already done in the `getSplunkCredential`

```
if len(splunkCredential.Token) == 0 {
}
```

tmjd: I don't think any change is needed. Token is required but CA is not, and the getSplunkCredential takes care of checking that.
All good. :+1:

PR #433, opened 2020-03-04
CNX-11540: added summary spec

manojah99:


manojah99:
Did go.mod update as required

tmjd:
I believe a change to go.sum should be in this PR also. When I build your branch I see that my go.sum is modified and includes the new version that was added to go.mod.

PR #437, opened 2020-03-05
Bump versions to v2.7.1

mgleung:


config/ee_versions.yaml
```diff
@@ -1,36 +1,36 @@
...
   es-proxy:
     image: tigera/es-proxy
-    version: v2.7.0
+    version: v2.7.1
   eck-kibana:
     version: 7.3.2
   kibana:
     image: tigera/kibana
-    version: 7.3
+    version: v2.7.1
```
tmjd: Are you sure about this? I think it was v2.6.x at somepoint, went to 7.3 and now back to v2.7.1. What's up with that?

mgleung: I had a chat with Karthik about this one. I think we want to keep it synced with our release cadence now. I think we were using 7.3 for some of our development and it confusingly got leaked into our release process.

PR #447, opened 2020-03-12
Add override for git version

tmjd:


Makefile
```diff
@@ -201,6 +201,11 @@ ifeq ($(LOCAL_BUILD),true)
 	GIT_VERSION = $(shell git describe --tags --dirty --always)-dev-build
```
ozdanborne: I think you could achieve the same by using `?=` here and the line above
```suggestion
	GIT_VERSION ?= $(shell git describe --tags --dirty --always)-dev-build
```

tmjd: I could change the 1st one but if I change the one in LOCAL_BUILD would cause that one to never be set because of the 1st one.
Changing the 1st should be enough to do this though. Thanks.

Makefile
```diff
@@ -196,7 +196,7 @@ sub-tag-images-%:
 .PHONY: build
 
 # Get version from git.
-GIT_VERSION:=$(shell git describe --tags --dirty --always)
+GIT_VERSION?=$(shell git describe --tags --dirty --always)
 ifeq ($(LOCAL_BUILD),true)
 	GIT_VERSION = $(shell git describe --tags --dirty --always)-dev-build
 endif
```
ozdanborne: What about this?
```suggestion
ifeq ($(LOCAL_BUILD),true)
	GIT_VERSION ?= $(shell git describe --tags --dirty --always)-dev-build
else
	GIT_VERSION?=$(shell git describe --tags --dirty --always)
endif
```

tmjd: Done, thanks

PR #450, opened 2020-03-13
Update the ECK operator to v1.0.1

rene-dekker:
- This PR swaps out the ECK alpha operator with version v1.0.1. In the past the ES pods were part of a deployment, now they are part of a statefulset. 
- Some things that were previously part of tigera-operator.yaml have been moved to the code to ensure they are removed properly when logstorage is removed or for managed clusters.
- In 0.9 ECK an init-container was secretly added to the podspec of the ES pod. In this version I moved the init container to the code. Note that the init pod has a privileged user to set a sysctl command. We are still investigating if there is a safer approach to this. 

This should be merged in concert with the appropriate kube-controllers and docs changes.


pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -495,6 +503,18 @@ func (r *ReconcileLogStorage) elasticsearchSecrets(ctx context.Context) ([]*core
 	return secrets, nil
 }
 
+// Returns true if we want to apply a new trial license.
+func (r *ReconcileLogStorage) applyElasticTrialSecret(ctx context.Context) (bool, error) {
```
Brian-McM: Nit, but this function name sounds like it's actually applying the trial instead of checking if it exists.

pkg/render/logstorage.go
```diff
@@ -383,8 +397,27 @@ func (es elasticsearchComponent) podTemplate() corev1.PodTemplateSpec {
 		}
 	}
 
+	// Was already present, but hidden from view in ECK 0.9. We are investigating how to improve on this internally:
```
Brian-McM: Since this is an opensource repo I'm not sure we should mention what we're doing internally...

pkg/render/logstorage.go
```diff
@@ -653,7 +687,7 @@ func (es elasticsearchComponent) eckOperatorStatefulSet() *appsv1.StatefulSet {
 					Containers: []corev1.Container{{
 						Image: components.GetReference(components.ComponentElasticsearchOperator, es.installation.Spec.Registry, es.installation.Spec.ImagePath),
 						Name:  "manager",
-						Args:  []string{"manager", "--operator-roles", "all", "--enable-debug-logs=false"},
+						Args:  []string{"manager", "--operator-roles", "all", "--log-verbosity=2"},
```
Brian-McM: What is level 2 equivalent to? Info?

pkg/render/logstorage.go
```diff
@@ -674,17 +708,17 @@ func (es elasticsearchComponent) eckOperatorStatefulSet() *appsv1.StatefulSet {
 							},
 							Requests: corev1.ResourceList{
 								"cpu":    resource.MustParse("100m"),
-								"memory": resource.MustParse("20Mi"),
+								"memory": resource.MustParse("50Mi"),
 							},
 						},
 						Ports: []corev1.ContainerPort{{
-							ContainerPort: 9876,
+							ContainerPort: 9443,
```
Brian-McM: Just wanted to double check this change is correct

pkg/render/logstorage.go
```diff
@@ -716,26 +750,27 @@ func (es elasticsearchComponent) kibanaCR() *kbv1alpha1.Kibana {
...
 			Version: components.ComponentEckKibana.Version,
 			Image:   components.GetReference(components.ComponentKibana, es.installation.Spec.Registry, es.installation.Spec.ImagePath),
-			Config: &cmneckalpha1.Config{
+			Config: &cmnv1.Config{
 				Data: map[string]interface{}{
 					"server": map[string]interface{}{
 						"basePath":        fmt.Sprintf("/%s", KibanaBasePath),
 						"rewriteBasePath": true,
 					},
+					"elasticsearch.ssl.certificateAuthorities": []string{"/usr/share/kibana/config/elasticsearch-certs/tls.crt"},
```
Brian-McM: If the user provides the kibana secret is this where ECK ends up mounting it?

pkg/render/logstorage.go
rene-dekker: Thanks for reminding me, it should be 0.


> Verbosity level of logs. -2=Error, -1=Warn, 0=Info, 0 and above=Debug

pkg/render/logstorage.go
rene-dekker: Yeah, although this is not specifically documented as such to my knowledge, I found out that this is where the cert is mounted.

pkg/render/logstorage.go
rene-dekker: ECK seems to have changed since 0.9.0 in commit 54c3f4a35dab7853fda4fd952bd812b71bd1eea3 . If we want things to be the same we can revert this line and also change it in the webhook service that is defined in this file. I am fine either way.

pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -495,6 +503,18 @@ func (r *ReconcileLogStorage) elasticsearchSecrets(ctx context.Context) ([]*core
 	return secrets, nil
 }
 
+// Returns true if we want to apply a new trial license.
```
caseydavenport: This function comment isn't super useful, since this information is essentially contained in the function name :) 

I'd prefer if there was some more explanation here - i.e., in what circumstances do we expect to need a trial license? What happens if a user wants to switch from a trial license to a non-trial license? etc.

PR #451, opened 2020-03-16
SAAS-663 Add tigerastatus for managed cluster

Suraiya-Hameed:
- Remove the status if related CR is not found.
- Add a new status `management-cluster-connection` to managed cluster that will let the user know if its connected to management cluster

### DIFF
--- a/pkg/controller/status/status.go
+++ b/pkg/controller/status/status.go
@@ -106,10 +106,13 @@ func (m *statusManager) Run() {
 	go func() {
 		// Loop forever, periodically checking dependent objects for their state.
 		for {
+			if m.removeTigeraStatus() {
+				// Wait for CR to be available.
+				time.Sleep(5 * time.Second)
+				continue
+			}
+
 			if !m.syncState() {
-				if !m.enabled {
-					m.crCleanup()
-				}
 				// Waiting to be in sync.
 				time.Sleep(5 * time.Second)
 				continue
@@ -133,10 +136,6 @@ func (m *statusManager) Run() {
 				m.setDegraded(m.degradedReason(), m.degradedMessage())
 			} else {
 				m.clearDegraded()
-			}
-
-			if !m.enabled {
-				m.crCleanup()
 			}
 
 			time.Sleep(5 * time.Second)
@@ -416,13 +415,19 @@ func (m *statusManager) syncState() bool {
 	return m.explicitDegradedReason != ""
 }
 
-// crCleanup removes the status displayed for CR in the TigeraStatus
-func (m *statusManager) crCleanup() {
-	ts := &operator.TigeraStatus{ObjectMeta: metav1.ObjectMeta{Name: m.component}}
-	err := m.client.Delete(context.TODO(), ts)
-	if err != nil && !apierrs.IsNotFound(err) {
-		log.WithValues("error", err).Info("Failed to remove TigeraStatus", m.component)
+// removeTigeraStatus returns true and removes the status displayed in TigeraStatus if corresponding CR not found
+func (m *statusManager) removeTigeraStatus() bool {
+	m.lock.Lock()
+	defer m.lock.Unlock()
+	if !m.enabled {
+		ts := &operator.TigeraStatus{ObjectMeta: metav1.ObjectMeta{Name: m.component}}
+		err := m.client.Delete(context.TODO(), ts)
+		if err != nil && !apierrs.IsNotFound(err) {
+			log.WithValues("error", err).Info("Failed to remove TigeraStatus", m.component)
+		}
+		return true
 	}
+	return false
 }
 
 // podsFailing takes a selector and returns if any of the pods that match it are failing. Failing pods are defined
--- a/pkg/controller/status/status.go
+++ b/pkg/controller/status/status.go
@@ -153,6 +153,14 @@ func (m *statusManager) OnCRNotFound() {
 	m.clearProgressing()
 	m.lock.Lock()
 	defer m.lock.Unlock()
+	// Remove status displayed for CR, if CR was available earlier but not now.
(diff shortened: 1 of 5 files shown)


pkg/controller/status/status.go
```diff
@@ -153,6 +153,14 @@ func (m *statusManager) OnCRNotFound() {
 	m.clearProgressing()
 	m.lock.Lock()
 	defer m.lock.Unlock()
+	// Remove status displayed for CR, if CR was available earlier but not now.
+	if m.enabled {
```
Brian-McM: I don't think we want to remove this here, all interactions with the API are done in the Run() loop. This function should only set the flag that the tigera-status should be removed, and then some part of the go routine in the Run() loop should remove the tigera-status (look here https://github.com/tigera/operator/blob/1ff13fc9f3d7962c591282f9442b6dd0e881c3a4/pkg/controller/status/status.go#L105).

 I would also consider if it makes sense to clear everything from the status manager (i.e. deployments, daemonsets,...).

Suraiya-Hameed: I've moved the logic to Run() loop and tested :+1: 

> I would also consider if it makes sense to clear everything from the status manager (i.e. deployments, daemonsets,...).
This is already done when m.enabled flag is set to false in OnCRNotFound()

pkg/controller/status/status.go
```diff
@@ -301,6 +302,17 @@ func (m *statusManager) syncState() bool {
 	defer m.lock.Unlock()
 	progressing := []string{}
 	failing := []string{}
+
+	// Remove status displayed for CR, if CR not found.
+	if !m.enabled {
```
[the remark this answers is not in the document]
Suraiya-Hameed: Initially I had the cleanup code at the end of `Run()`, but the flow never goes beyond `syncState` as all the resources are cleaned up when `enabled` is set to false in `onCRNotFound`. Flow gets stuck in this loop https://github.com/tigera/operator/blob/master/pkg/controller/status/status.go#L108

How about calling crCleanup twice, once at the end of `syncState` when we have no more resources to watch and once at the end of `Run()`? Though I am not sure if the one at the end of `Run()` will ever get called, since we cleanup resource when CR not found.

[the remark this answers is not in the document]
Brian-McM: I'm still not a fan of syncState having that logic, this isn't "syncing the state" of the resources. You could run it in that Run()'s loop but regardless of the outcome of syncState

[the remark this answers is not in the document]
Suraiya-Hameed: True, I've moved it out.

pkg/apis/operator/v1/managementclusterconnection_types.go
```diff
@@ -40,6 +40,8 @@ type ManagementClusterConnection struct {
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 
 	Spec ManagementClusterConnectionSpec `json:"spec,omitempty"`
+
+	Status ManagementClusterConnectionStatus `json:"status,omitempty"`
```
Brian-McM: I can't really remember how this Status would be used, is the State value in this ever set?

pkg/controller/status/status.go
```diff
@@ -407,6 +416,15 @@ func (m *statusManager) syncState() bool {
 	return m.explicitDegradedReason != ""
 }
 
+// Remove status displayed for CR
```
Brian-McM: I think this message could be a little more explicit, mentioning specifically that the "tigera-status" for the CR is removed. You should also take a look at this https://blog.golang.org/godoc-documenting-go-code. The standard for go documentation is to start the comment with the name of the element being described.

pkg/controller/status/status.go
```diff
@@ -407,6 +416,15 @@ func (m *statusManager) syncState() bool {
 	return m.explicitDegradedReason != ""
 }
 
+// crCleanup removes the status displayed for CR in tigerastatus
```
Brian-McM: Nit "in the TigeraSTatuses" (I think we normally use the capitalise / camel case for resource names)

pkg/controller/status/status.go
```diff
@@ -131,6 +135,10 @@ func (m *statusManager) Run() {
 				m.clearDegraded()
 			}
 
+			if !m.enabled {
```
caseydavenport: Looks like we have this in two places within this loop - why is that? Can we resolve these to a single check? 

I think you can just put this at the top of the loop?

```go
if !m.enabled {
   cleanup()
   time.Sleep( 5 * time.Second)
   continue
} 

if !m.syncState() {
 . . .
}
...
```

pkg/controller/status/status.go
```diff
@@ -106,6 +107,9 @@ func (m *statusManager) Run() {
 		// Loop forever, periodically checking dependent objects for their state.
 		for {
 			if !m.syncState() {
+				if !m.enabled {
```
caseydavenport: Looks like everywhere else we access `m.enabled` we're using a mutex to prevent concurrent access. I think we should do that here as well.

Suraiya-Hameed: We are using mutex while writing into `m.enabled`, looks like we don't use it for read https://github.com/tigera/operator/blob/master/pkg/controller/status/status.go#L455

caseydavenport: @Suraiya-Hameed if you look at all the places that we call `set(` you will see that the mutex is already held by the caller.

caseydavenport:
It would be nice to have an FV test in the repo that makes sure this logic is sound. Could you add one of those as well? 

I'm thinking a test that deletes a CR (e.g., Installation) and asserts that the corresponding TigeraStatus is eventually removed.

PR #455, opened 2020-03-17
Fix nodeMetricsPort to adjust correct setting

caseydavenport:


pkg/render/node_test.go
```diff
@@ -793,22 +789,26 @@ var _ = Describe("Node rendering tests", func() {
 			}),
 	)
 
-	It("should not export FELIX_PROMETHEUSREPORTERPORT if NodeMetricsPort is nil", func() {
+	It("should not enabled prometheus metrics if NodeMetricsPort is nil", func() {
```
doublek: nit: typo s/enabled/enable.

PR #456, opened 2020-03-19
Operator: Customize rolling update parameters

rafaelvanoni:


pkg/apis/operator/v1/types.go
```diff
@@ -91,6 +92,11 @@ type InstallationSpec struct {
 	// kubernetesProvider.
 	// +optional
 	FlexVolumePath string `json:"flexVolumePath,omitempty"`
+
+	// MaxUnavailable determines the max number of daemonset pods that can be unavailable during an update.
+	// Default: 1
+	// +optional
+	MaxUnavailable *intstr.IntOrString `json:"maxUnavailable,omitempty"`
```
caseydavenport: Wondering if we should expose the full update strategy section wholesale: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.17/#daemonsetupdatestrategy-v1-apps

That would give us some more flexibility if we want to let users customize this further. 

I think we also need to consider what this means - there are multiple deployments / daemonsets / etc that this field could apply to across the operator, and it's not obvious from the name / comment which one this applies to.

I think the intention is that this only applies to the node daemonset?

tmjd: Do we really want to allow the user to switch the type to "OnDelete"?
And what is the implication for the namespace migration?

pkg/controller/migration/namespace_migration.go
```diff
@@ -469,18 +470,36 @@ func (m *CoreNamespaceMigration) getNodesToMigrate() []*v1.Node {
...
+		if csMaxUnavailable != nil {
+			n, err := intstr.GetValueFromIntOrPercent(csMaxUnavailable, int(csD), false)
+			if err != nil {
+				if (csD - int32(n)) >= 0 {
+					csD -= int32(n)
+				}
+			}
+		}
+
+		if ksD <= ksR && csD <= csR {
```
tmjd: I'm not sure if this is what we want. I think this could result in double the number of unavailable pods.
As I was just thinking through this, I think we should use the MaxUnavailable from the calico-system daemonset and make sure that ksR + csR is greater than or equal to that value. I think that would best simulate what a Daemonset RollingUpdate update would do.

pkg/apis/operator/v1/types.go
caseydavenport: @tmjd good question. I think probably not generically, but it could be useful for debugging, etc.

I suspect we should be validating that field to make sure it is set to "RollingUpdate", since that's the only mode that we support / test right now. I do think it still makes sense to maintain the structure, in case we want to support other modes in the future, or if there are additional fields added to the rolling update strategy that we want to enable. WDYT?

caseydavenport: For namespace migration, we probably only care about the maxUnavailable field.

pkg/apis/operator/v1/types.go
```diff
@@ -91,6 +92,8 @@ type InstallationSpec struct {
 	// kubernetesProvider.
 	// +optional
 	FlexVolumePath string `json:"flexVolumePath,omitempty"`
+
+	NodeUpdateStrategy appsv1.DaemonSetUpdateStrategy `json:"nodeUpdateStrategy,omitempty"`
```
caseydavenport: This field is missing API doc comment. 

The comments get generated into documentation for users, so each field needs one.

pkg/controller/migration/namespace_migration.go
caseydavenport: Yeah, agreed - as it's currently ready I think you could end up with 2*maxUnavailable pods not ready.

We should compare the total number of unready pods vs the total number of desired pods.

pkg/controller/migration/namespace_migration.go
```diff
@@ -469,35 +469,46 @@ func (m *CoreNamespaceMigration) getNodesToMigrate() []*v1.Node {
...
+		csD, csR, csMaxUnavailable, err := m.getNumPodsDesiredAndReady(common.CalicoNamespace, nodeDaemonSetName)
 		if err != nil {
 			return false, err
 		}
 
-		// TODO: When we support configuring adjusting the rolling update unavailable,
-		// we should use that configuration here.
-		if ksD == ksR && csD == csR {
+		// MaxUnavailable represents the number of pods that are allowed to be down during an update. We
+		// decrement the number of desired pods by that value.
```
tmjd: nit: this comment doesn't match what was actually done though logically it is the same.

pkg/apis/operator/v1/types.go
```diff
@@ -92,6 +93,11 @@ type InstallationSpec struct {
 	// kubernetesProvider.
 	// +optional
 	FlexVolumePath string `json:"flexVolumePath,omitempty"`
+
+	// NodeUpdateStrategy can be used to customize the desired update strategy, such as the MaxUnavailable
+	// field (which defaults to a value of one).
```
caseydavenport: I don't think we need to / should call out the default for a sub-field on the top-level struct.

PR #457, opened 2020-03-20
Move admission controller support behind a tech-preview feature flag

doublek:


pkg/controller/apiserver/apiserver_controller_test.go
```diff
@@ -0,0 +1,47 @@
...
+		Expect(supportEnabled).To(Equal(expectedResult))
+	}
+
+	It("should parse tech-preview annotation", func() {
+		checkAnnotation("enabled", true)
+		checkAnnotation("Enabled", true)
+		checkAnnotation("somethingelse", false)
+		checkAnnotation("", false)
+	})
+})
```
tmjd: WDYT about adding a test for no annotation on the APIServer?

doublek: Will do.

PR #458, opened 2020-03-20
Add HostPort configuration

tmjd:


pkg/apis/operator/v1/types.go
```diff
@@ -145,6 +166,12 @@ type CalicoNetworkSpec struct {
 	// IPv6 addresses will not be auto-detected.
 	// +optional
 	NodeAddressAutodetectionV6 *NodeAddressAutodetection `json:"nodeAddressAutodetectionV6,omitempty"`
+
+	// HostPort specifies if the plugin for HostPort will be configured as part of the CNI configuration.
+	// Default: Enabled
+	// +optional
+	// +kubebuilder:validation:Enum=Enabled,Disabled
+	HostPort *HostPortType `json:"hostPort,omitempty"`
```
caseydavenport: I think we should call this `HostPorts` with an `s`.

"Enable or disable host ports" is how I'd say it in english.

pkg/render/node.go
```diff
@@ -320,6 +320,16 @@ func (c *nodeComponent) nodeCNIConfigMap() *v1.ConfigMap {
 		assign_ipv6 = "false"
 	}
 
+	var portmap string = ""
+	if c.cr.Spec.CalicoNetwork.HostPort != nil && *c.cr.Spec.CalicoNetwork.HostPort == operator.HostPortEnabled {
+		portmap = `,
+	{
```
caseydavenport: hm. the formatting here is pretty rough. Wonder if it would be nicer just to put this all on a single line, since it's obvious and harder to accidentally mess up with a spacing issue.

e.g., 

```
portmap = `,{"type": "portmap", "snat": true, "capabilities": {"portMappings": true}}`
```

WDYT?

pkg/render/node_test.go
```diff
@@ -66,7 +68,37 @@ var _ = Describe("Node rendering tests", func() {
...
+          "assign_ipv6" : "false"
+      },
+      "policy": {
+          "type": "k8s"
+      },
+      "kubernetes": {
+          "kubeconfig": "__KUBECONFIG_FILEPATH__"
+      }
+    },
+	{
```
caseydavenport: Indentation here looks like it's a couple of spaces off, I think.

pkg/apis/operator/v1/types.go
```diff
@@ -145,6 +166,12 @@ type CalicoNetworkSpec struct {
 	// IPv6 addresses will not be auto-detected.
 	// +optional
 	NodeAddressAutodetectionV6 *NodeAddressAutodetection `json:"nodeAddressAutodetectionV6,omitempty"`
+
+	// HostPorts specifies if the plugin for HostPorts will be configured as part of the CNI configuration.
```
caseydavenport: I think this should be more use-facing - e.g. "HostPorts configures whether or not Calico will support Kubernetes HostPorts".

The plugin is an implementation detail, not API.

pkg/render/node.go
```diff
@@ -416,6 +418,10 @@ func (c *nodeComponent) nodeDaemonset() *apps.DaemonSet {
 	annotations[typhaCAHashAnnotation] = AnnotationHash(c.typhaNodeTLS.CAConfigMap.Data)
 	annotations[nodeCertHashAnnotation] = AnnotationHash(c.typhaNodeTLS.NodeSecret.Data)
 
+	if cniCfgMap != nil {
```
caseydavenport: good spot

pkg/render/node_test.go
```diff
@@ -66,7 +68,32 @@ var _ = Describe("Node rendering tests", func() {
...
+          "assign_ipv4" : "true",
+          "assign_ipv6" : "false"
+      },
+      "policy": {
+          "type": "k8s"
+      },
+      "kubernetes": {
+          "kubeconfig": "__KUBECONFIG_FILEPATH__"
+      }
+    },{"type": "portmap", "snat": true, "capabilities": {"portMappings": true}}
```
caseydavenport: lol yeah this is pretty horrible looking too, isn't it... oh well. Maybe we should add a newline before this?

PR #461, opened 2020-03-24
Decrease default shards to 1

asincu:
Decrease the number of shards to 1 instead of 5.

### DIFF
--- a/pkg/controller/logstorage/logstorage_controller.go
+++ b/pkg/controller/logstorage/logstorage_controller.go
@@ -595,23 +595,22 @@ func (r *ReconcileLogStorage) getKibanaService(ctx context.Context) (*corev1.Ser
 	return &svc, nil
 }
 
-func calculateFlowShards(nodes *operatorv1.Nodes, defaultShards int) int {
-	if nodes == nil {
+func calculateFlowShards(nodesSpecifications *operatorv1.Nodes, defaultShards int) int {
+	if nodesSpecifications == nil || nodesSpecifications.ResourceRequirements == nil || nodesSpecifications.ResourceRequirements.Requests == nil {
 		return defaultShards
 	}
 
-	var nodesInt, coresInt, shardPerNode int
-	nodesInt = int(nodes.Count)
-	coresInt = int(nodes.ResourceRequirements.Requests.Cpu().AsInt64())
-	shardPerNode = coresInt / 4
+	var nodes = nodesSpecifications.Count
+	var cores, _ = nodesSpecifications.ResourceRequirements.Requests.Cpu().AsInt64()
+	var shardPerNode = int(cores) / 4
 
-	if nodesInt <= 0 || coresInt <= 0 || shardPerNode <= 0{
+	if nodes <= 0 || cores <= 0 || shardPerNode <= 0{
 		return defaultShards
 	}
 
 	if shardPerNode < defaultShards {
-		return nodesInt * defaultShards
+		return int(nodes) * defaultShards
 	}
 
-	return nodesInt * shardPerNode
+	return int(nodes) * shardPerNode
 }
--- a/pkg/controller/logstorage/logstorage_controller.go
+++ b/pkg/controller/logstorage/logstorage_controller.go
@@ -604,12 +604,8 @@ func calculateFlowShards(nodesSpecifications *operatorv1.Nodes, defaultShards in
 	var cores, _ = nodesSpecifications.ResourceRequirements.Requests.Cpu().AsInt64()
 	var shardPerNode = int(cores) / 4
 
-	if nodes <= 0 || cores <= 0 || shardPerNode <= 0{
+	if nodes <= 0 || shardPerNode <= 0 {
 		return defaultShards
-	}
-
-	if shardPerNode < defaultShards {
-		return int(nodes) * defaultShards
 	}
 
 	return int(nodes) * shardPerNode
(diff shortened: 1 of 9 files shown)


pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -57,7 +57,7 @@ const (
 	defaultResolveConfPath             = "/etc/resolv.conf"
 	defaultLocalDNS                    = "svc.cluster.local"
 	tigeraElasticsearchUserSecretLabel = "tigera-elasticsearch-user"
-	defaultElasticsearchShards         = 5
+	defaultElasticsearchShards         = 1
```
rene-dekker: I think we need a separate PR to make this happen for Helm based installations. Maybe it can be dont from calico-private with an env setting, but probably we want to lower the default in the docker-fluentd repo as well.

tmjd:
There does not look to currently be a way for the user to configure shards for ES. Is there something I'm missing?
If it is correct that there is not a way to configure shards, is that known in this decision to switch the default to 1.
Also this will change any current installs from 5 to 1 as the value in the ConfigMap will be updated when a Reconcile is ran.

pkg/controller/logstorage/logstorage_controller.go
matthewdupre: Do we not have independent control of this for each type of index?  I think we do sometimes want to have more shards (probably based on ES resources) for the flow logs index.

asincu:
@tmjd  That is the desired approach. For flow logs, it should be calculated based on the number of nodes and cpu's to have granularity. I think we want any new index that is created to have 1 shard while old indices to maintain their previous shards.

pkg/render/fluentd.go
```diff
@@ -335,8 +335,6 @@ func (c *fluentdComponent) container() corev1.Container {
 func (c *fluentdComponent) envvars() []corev1.EnvVar {
 	envs := []corev1.EnvVar{
 		{Name: "FLUENT_UID", Value: "0"},
-		{Name: "ELASTIC_FLOWS_INDEX_SHARDS", Value: "5"},
```
matthewdupre: Do these variables get set somewhere else that I haven't seen?

asincu: @matthewdupre Yup. Here: https://github.com/tigera/operator/blob/master/pkg/render/fluentd.go#L445-L446

pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -592,3 +594,23 @@ func (r *ReconcileLogStorage) getKibanaService(ctx context.Context) (*corev1.Ser
...
+
+	var nodes = nodesSpecifications.Count
+	var cores, _ = nodesSpecifications.ResourceRequirements.Requests.Cpu().AsInt64()
+	var shardPerNode = int(cores) / 4
+
+	if nodes <= 0 || cores <= 0 || shardPerNode <= 0{
+		return defaultShards
+	}
+
+	if shardPerNode < defaultShards {
```
tmjd: I don't think with the new default of 1 that this case will ever be used, since shardPerNode is calculated with integer math.

pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -592,3 +594,23 @@ func (r *ReconcileLogStorage) getKibanaService(ctx context.Context) (*corev1.Ser
...
+func calculateFlowShards(nodesSpecifications *operatorv1.Nodes, defaultShards int) int {
+	if nodesSpecifications == nil || nodesSpecifications.ResourceRequirements == nil || nodesSpecifications.ResourceRequirements.Requests == nil {
+		return defaultShards
+	}
+
+	var nodes = nodesSpecifications.Count
+	var cores, _ = nodesSpecifications.ResourceRequirements.Requests.Cpu().AsInt64()
+	var shardPerNode = int(cores) / 4
+
+	if nodes <= 0 || cores <= 0 || shardPerNode <= 0{
```
tmjd: I think `cores <= 0 || shardPerNode <= 0` is redundant, I think `shardPerNode <= 0` would be sufficient.

pkg/render/elasticsearchclusterconfig.go
```diff
@@ -94,6 +100,7 @@ func (c ElasticsearchClusterConfig) ConfigMap() *corev1.ConfigMap {
 			"clusterName": c.clusterName,
 			"replicas":    strconv.Itoa(c.replicas),
 			"shards":      strconv.Itoa(c.shards),
+			"flowShards":  strconv.Itoa(c.flowShards),
```
tmjd: What you've linked in another comment https://github.com/tigera/operator/blob/master/pkg/render/fluentd.go#L445-L446 also indicates that there is no separation of FLOW and other shards in fluentd. I think that means that this flowShards will do nothing.

tmjd:
> I think we want any new index that is created to have 1 shard while old indices to maintain their previous shards.

Is that because the number of shards is defined when an index is created? But we create new indexes each day right? So the the current day would be finished with the current value but the next day would have the new default. Maybe this is what we expect but I just want to make sure that is what we're expecting.

asincu:
@tmjd Yes, that is what we are expecting. I did some tests yesterday and installed this branch on top a cluster that already has ES working.  New indices will have shards set to 1 and old indices will still ingest data.

Screenshot from 2020-03-24 16-09-37
Screenshot from 2020-03-24 16-15-19

PR #462, opened 2020-03-24
Update versions

tmjd:


bcreane:
Hey @tmjd - how was this update generated?

tmjd:
I used some of the latest builds for the enterprise and the v3.13.1 Calico components then ran `make gen-versions` and committed those updates.

pkg/components/enterprise.go
```diff
@@ -15,135 +15,158 @@
...
 	ComponentTigeraNode = component{
-		Version: "v2.7.0",
-		Digest:  "sha256:205c5e15543ef4c0f166a77326a061f5d99a1e20f6c9e4a0a8227b620333d280",
+		Version: "v2.7.0-0.dev-300-ge5619d5",
+		Digest:  "sha256:2b7434aedd81e84b9dccc14608f0af099c4f1e6926375b8ca0c8b646934ad064",
 		Image:   "tigera/cnx-node",
 	}
-
+	
+
```
ozdanborne: Odd that these newlines were added. Perhaps you opened it in your editor and it applied formatting? We should really run the output of the template through `gofmt`. But fine as is

PR #465, opened 2020-03-25
Clarify that cluster management type is for enterprise only

caseydavenport:


pkg/controller/installation/validation.go
```diff
@@ -143,6 +143,13 @@ func validateCustomResource(instance *operatorv1.Installation) error {
 			instance.Spec.FlexVolumePath)
 	}
 
+	if instance.Spec.Variant == operatorv1.TigeraSecureEnterprise {
```
tmjd: I think you mean to check != or need to change to operatorv1.Calico

caseydavenport: doh yeah, fixed.

PR #473, opened 2020-03-26
[WIP] FIX! Flow shards is not being set for managed clusters

asincu:


pkg/render/elasticsearchclusterconfig.go
```diff
@@ -56,10 +56,19 @@ func NewElasticsearchClusterConfigFromConfigMap(configMap *corev1.ConfigMap) (*E
...
+		if flowShards, err = strconv.Atoi(configMap.Data["flowShards"]); err != nil {
+			return nil, errors.Wrap(err, "'flowShards' must be an integer")
+		}
+	}
+
 	return &ElasticsearchClusterConfig{
 		clusterName: configMap.Data["clusterName"],
 		replicas:    replicas,
 		shards:      shards,
+		flowShards: flowShards,
```
matthewdupre: This bug wouldn't have happened if we'd called the main NewESCC() function here instead of replicating.  I think it's worth changing.

PR #475, opened 2020-03-26
update prometheus and calico-system namespace creation

ozdanborne:
stop creating the prometheus namespace, instead relying on it already
being created. and move creation of imagepullsecrets to the manager
controller which is more tied to prometheus than the installation
controller.

This presents a situation where the prometheus namespace doesn't exist. This chance of this happening is reduced because it is included in our manifests (currently). Even so, I tested the error handling of this code:
- when using imagepullsecrets, you would get a nice error pointing out that the secrets couldn't be copied to the tigera-prometheus namespace because it doesn't exist.
- However, when not using imagepullsecrets, the code would silently succeed that the manager would just be not-functional because prometheus is down. As such, I added a check to the manager controller to validate that the prometheus namespace exists. An example tigerastatus of this happening is here:

```
Status:
  Conditions:
    Last Transition Time:  2020-04-22T20:45:28Z
    Status:                False
    Type:                  Available
    Last Transition Time:  2020-04-22T20:45:28Z
    Status:                False
    Type:                  Progressing
    Last Transition Time:  2020-04-22T20:45:28Z
    Message:               Dependency on tigera-prometheus not satisfied
    Reason:                prometheus namespace does not exist
    Status:                True
    Type:                  Degraded
Events:                    <none>
```

### DIFF
--- a/pkg/controller/manager/manager_controller.go
+++ b/pkg/controller/manager/manager_controller.go
@@ -114,6 +114,10 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 		return fmt.Errorf("manager-controller failed to watch Network resource: %v", err)
 	}
 
+	if err = utils.AddNamespaceWatch(c, common.TigeraPrometheusNamespace); err != nil {
+		return fmt.Errorf("manager-controller failed to watch the '%s' namespace: %v", common.TigeraPrometheusNamespace, err)
+	}
+
 	return nil
 }
 
--- a/pkg/controller/manager/manager_controller.go
+++ b/pkg/controller/manager/manager_controller.go
@@ -20,6 +20,7 @@ import (
 	"time"
 
 	operatorv1 "github.com/tigera/operator/pkg/apis/operator/v1"
+	"github.com/tigera/operator/pkg/common"
 	"github.com/tigera/operator/pkg/controller/compliance"
 	"github.com/tigera/operator/pkg/controller/installation"
 	"github.com/tigera/operator/pkg/controller/status"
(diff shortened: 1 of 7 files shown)


pkg/render/namespaces.go
```diff
@@ -17,51 +17,8 @@ package render
...
-
-	operator "github.com/tigera/operator/pkg/apis/operator/v1"
-	"github.com/tigera/operator/pkg/common"
-)
-
-const (
-	TigeraPrometheusNamespace = "tigera-prometheus"
 )
 
-func Namespaces(cr *operator.Installation, openshift bool, pullSecrets []*corev1.Secret) Component {
```
ozdanborne: @tmjd  I opened this PR to show / ask why node is creating namespace resources using this Component interface instead of just running these four lines of code like the other controllers:

```
		objs = append(objs, createNamespace(TigeraPrometheusNamespace, c.openshift))
	if len(c.pullSecrets) > 0 {
		objs = append(objs, copyImagePullSecrets(c.pullSecrets, TigeraPrometheusNamespace)...)
	}
```

Do the other controllers need to be updated, is the Installation controller outdated, or is there some other special case that the Installation resource needs to account for?

tmjd: The core was written a little differently than the other controllers is probably the main reason.

pkg/render/render.go
```diff
@@ -191,7 +191,6 @@ type calicoRenderer struct {
 func (r calicoRenderer) Render() []Component {
 	var components []Component
 	components = appendNotNil(components, PriorityClassDefinitions(r.installation))
-	components = appendNotNil(components, Namespaces(r.installation, r.provider == operator.ProviderOpenShift, r.pullSecrets))
```
tmjd: This won't work because the namespace needs to exist for the ConfigMaps, Secrets, and Typha components, so moving it to the "Node" component will not work.

pkg/controller/manager/manager_controller.go
```diff
@@ -215,6 +216,17 @@ func (r *ReconcileManager) Reconcile(request reconcile.Request) (reconcile.Resul
 		return reconcile.Result{}, nil
 	}
 
+	// check that prometheus is running
+	ns := &corev1.Namespace{}
```
tmjd, line 220: I think we should add a watch for this too https://github.com/tigera/operator/pull/475/files#diff-8962c60cf9c10aa1ac7a2ffb1422ae3cL68
So if it is missing and then gets added the Reconcile will be triggered.

PR #477, opened 2020-03-27
Fix cloud-on-k8s pin

tmjd:
v1.0.1 is not a valid git tag and caused problems generating the
reference API docs. This tag was unused I guess since there is a
replace line for the package.


stevegaossou:
Weird, it was [changed in this PR](https://github.com/tigera/operator/pull/450/files), but it looks like the build was passing.

PR #478, opened 2020-03-27
Update operator CRDs based on release-v1.4 branch

tmjd:


doublek:
I don't have context around these particular fields and they don't seem to be related to the metrics port in particular.

stevegaossou:
Okay I was just curious why they got removed in Casey's PR for `Fix nodeMetricsPort to adjust correct setting`.

tmjd:
These changes are not bundled with the operator, they will have no impact on the release. It is just committing the changes that come from running `make gen-files`. So there should be no impact for release-v1.4.

PR #479, opened 2020-03-27
Fix GIT_VERSION in makefile

tmjd:


Makefile
```diff
@@ -197,9 +197,9 @@ sub-tag-images-%:
 
 # Get version from git.
 ifeq ($(LOCAL_BUILD),true)
-	GIT_VERSION?=$(shell git describe --tags --dirty --always)-dev-build
+  GIT_VERSION?=$(shell git describe --tags --dirty --always)-dev-build
 else
-	GIT_VERSION?=$(shell git describe --tags --dirty --always)
+  GIT_VERSION?=$(shell git describe --tags --dirty --always)
```
stevegaossou: Just curious about this fix. 

I thought Makefiles required tags for indentation.

tmjd: Tabs mean that a line is part of a target, we didn't want this to be part of a target so changing the tab to spaces makes them always execute when the if branch says they should.

PR #480, opened 2020-03-28
Fix up portmap cni config spacing

tmjd:
Changing tabs to spaces to match the rest of the cni config


pkg/render/node.go
```diff
@@ -325,7 +325,7 @@ func (c *nodeComponent) nodeCNIConfigMap() *v1.ConfigMap {
 	var portmap string = ""
```
rene-dekker: type string is redundant

PR #484, opened 2020-03-30
[cherry-pick] Operator: Customize rolling update parameters

rafaelvanoni:


pkg/controller/installation/validation.go
```diff
@@ -149,6 +150,18 @@ func validateCustomResource(instance *operatorv1.Installation) error {
 			instance.Spec.FlexVolumePath)
 	}
 
+	if instance.Spec.NodeUpdateStrategy.Type != appsv1.RollingUpdateDaemonSetStrategyType {
+		return fmt.Errorf("Installation spec.NodeUpdateStrategy.type '%s' is not supported",
+			instance.Spec.NodeUpdateStrategy.RollingUpdate)
+	}
+
+	if instance.Spec.Variant == operatorv1.Calico {
```
tmjd: I don't think this is was a separate change.

rafaelvanoni: There was another reference to `Variant` in `validation_test.go` too

PR #486, opened 2020-04-02
GlobalAlertTemplates are not created for the managed cluster

asincu:
When creating a managed cluster, global alert templates should be present in the managed cluster. 

This requires that the creation of the global alert templates should be moved from tigera-manager to a component that gets installed in each cluster (ex: intrusion detection).


asincu:
This PR was merged.

PR #487, opened 2020-04-02
Allow for a configurable storage class name.

rene-dekker:
Introduce a new field to the LogStorage CR(D) to allow for users to specify the storage class that is used by the ECK operator to create PVC's based on storage classes other than the default ("tigera-elasticsearch).

### DIFF
--- a/pkg/controller/logstorage/logstorage_controller.go
+++ b/pkg/controller/logstorage/logstorage_controller.go
@@ -58,6 +58,7 @@ const (
 	defaultLocalDNS                    = "svc.cluster.local"
 	tigeraElasticsearchUserSecretLabel = "tigera-elasticsearch-user"
 	defaultElasticsearchShards         = 1
+	DefaultElasticsearchStorageClass   = "tigera-elasticsearch"
 )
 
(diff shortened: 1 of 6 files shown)


pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -348,9 +347,17 @@ func (r *ReconcileLogStorage) Reconcile(request reconcile.Request) (reconcile.Re
 
 		var flowShards = calculateFlowShards(ls.Spec.Nodes, defaultElasticsearchShards)
 		clusterConfig = render.NewElasticsearchClusterConfig(render.DefaultElasticsearchClusterName, ls.Replicas(), defaultElasticsearchShards, flowShards)
-		if err := r.client.Get(ctx, client.ObjectKey{Name: render.ElasticsearchStorageClass}, &storagev1.StorageClass{}); err != nil {
+
+		// Check if there is a StorageClass available to run Elasticsearch on.
+		scn := ls.Spec.StorageClassName
```
Brian-McM, line 352: This is a small thing, but since you're writing back to ls.Spec.StorageClassName, there's really no need for another variable. ls.Spec.StorageClassName will always have a non empty name.

PR #494, opened 2020-04-07
[release-v1.4] ES NodeSelector

rene-dekker:
Allows the user to add a NodeSelector to the podSpec of ES by adding it to the LogStorage CR. This allows some of our users to schedule the ES onto the pods with the right volumes and resources.

### DIFF
--- a/pkg/apis/operator/v1/logstorage_types.go
+++ b/pkg/apis/operator/v1/logstorage_types.go
@@ -60,6 +60,12 @@ type LogStorageSpec struct {
 	// Default: tigera-elasticsearch
 	// +optional
 	StorageClassName string `json:"storageClassName,omitempty"`
+
+	// NodeSelector gives you more control over the node that Elasticsearch will run on. The contents of NodeSelector will
+	// be added to the PodSpec of the Elasticsearch nodes. For the pod to be eligible to run on a node, the node must have
+	// each of the indicated key-value pairs as labels.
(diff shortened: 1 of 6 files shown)


pkg/apis/operator/v1/logstorage_types.go
```diff
@@ -60,6 +60,12 @@ type LogStorageSpec struct {
 	// Default: tigera-elasticsearch
 	// +optional
 	StorageClassName string `json:"storageClassName,omitempty"`
+
+	// NodeSelector gives you more control over the node that Elasticsearch will run on. The contents of NodeSelector will
+	// be added to the PodSpec of the Elasticsearch nodes. For the pod to be eligible to run on a node, the node must have
+	// each of the indicated key-value pairs as labels.
+	// +optional
+	NodeSelector map[string]string `json:"nodeSelector,omitempty"`
```
tmjd, line 68: This will only apply to the ES 'worker' pods. Is there something we could add to the name to indicate that? Or should this apply to the operator as well?

caseydavenport: Yeah good question... should the ECK operator be goverened by the `controlPlanNodeSelector` on the Installation resource, really?

rene-dekker: How about `DataNodeSelector` to remove the ambiguity?

PR #498, opened 2020-04-07
[release-v1.2] ES NodeSelector

rene-dekker:
Allows the user to add a NodeSelector to the podSpec of ES by adding it to the LogStorage CR. This allows some of our users to schedule the ES onto the nodes with the right volumes and resources.


pkg/render/elasticsearch_test.go
```diff
@@ -80,6 +84,9 @@ var _ = Describe("Elasticsearch rendering tests", func() {
 		resources := component.Objects()
 		Expect(len(resources)).To(Equal(len(expectedResources)))
 
+		// Verify that the node selectors are passed into the Elasticsearch pod spec.
+		Expect(resources[9].(*esalpha1.Elasticsearch).Spec.Nodes[0].PodTemplate.Spec.NodeSelector["label"]).To(Equal("value"))
```
tmjd, line 88: I would like to see this done as 2 separate tests.
* 1 that is more of a default render test
* 1 with a NodeSelector set, ideally a specific test just for setting and checking that NodeSelector renders as expected

PR #501, opened 2020-04-07
Enable dynamic webhook

gianlucam76:
`kind/enhancement` 
`enterprise`

### DIFF
--- a/pkg/controller/apiserver/apiserver_controller.go
+++ b/pkg/controller/apiserver/apiserver_controller.go
@@ -17,7 +17,6 @@ package apiserver
 import (
 	"context"
 	"fmt"
-	"strings"
 	"time"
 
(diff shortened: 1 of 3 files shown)


tmjd:
The PR that added the tech preview annotation was #457, I think we should revert more of that PR in this change. Specifically I'm referring to removing isAdmissionControllerSupportEnabled and the additional flag that had been added to the render.APIServer call. That probably also means the pkg/controller/apiserver test files can be removed.

gianlucam76:
Thanks @tmjd . Idea was to change the logic from default disabled to default enabled, hence this PR (though I agree the function is useless now). Anyhow, with said, I will revert #457 in master tomorrow and then cherry pick to 1.4 as well for calico enterprise 2.8.

gianlucam76:
replaced with https://github.com/tigera/operator/pull/502

PR #502, opened 2020-04-08
Revert "Move admission controller support behind a tech-preview feature flag"

gianlucam76:
Reverts tigera/operator#457

Dynamic admission webhook will be enabled by default.

kind/enhancement
enterprise


pkg/render/apiserver.go
```diff
@@ -578,7 +574,6 @@ func (c *apiServerComponent) apiServerContainer() corev1.Container {
 			fmt.Sprintf("--secure-port=%d", apiServerPort),
 			"--audit-policy-file=/etc/tigera/audit/policy.conf",
 			"--audit-log-path=/var/log/calico/audit/tsee-audit.log",
-			fmt.Sprintf("--enable-admission-controller-support=%t", c.enableAdmissionControllerSupport),
```
tmjd, line 581: Is this not necessary anymore? In other words is that flag no longer needed to enable the controller on the API server?

gianlucam76: it is not needed anymore. Thanks.

PR #508, opened 2020-04-10
Add MultiInterfaceMode to installation spec

Brian-McM:
## Description
This configures if multiple interfaces are allowed per pod and what configures them. We only allow multus for now.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/deploy/crds/operator_v1_installation_crd.yaml
+++ b/deploy/crds/operator_v1_installation_crd.yaml
@@ -88,6 +88,14 @@ spec:
                     for pods on the Calico network. Default: 1410'
                   format: int32
                   type: integer
+                multiInterfaceMode:
+                  description: 'MultiInterfaceMode configures what will configure
+                    multiple interface per pod. Only valid for Calico Enterprise installations.
+                    Default: None'
+                  enum:
+                  - None
+                  - Multus
+                  type: string
                 nodeAddressAutodetectionV4:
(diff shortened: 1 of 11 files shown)


deploy/crds/operator_v1_installation_crd.yaml
```diff
@@ -88,6 +88,14 @@ spec:
                     for pods on the Calico network. Default: 1410'
                   format: int32
                   type: integer
+                multiInterfaceMode:
+                  description: 'MultiInterfaceMode configures what will configure
+                    multiple interface per pod. Only valid for EE installations. Default:
```
tmjd, line 93: I don't think EE is appropriate to use here, should probably be `Calico Enterprise`

pkg/render/node.go
```diff
@@ -635,6 +635,12 @@ func (c *nodeComponent) cniEnvvars() []v1.EnvVar {
 			},
 		},
 	}
+
+	if c.cr.Spec.CalicoNetwork != nil {
```
tmjd, line 639: Should this be gated on Spec.Variant === Enterprise (like the others)?

Brian-McM: yes it should

caseydavenport:
@tmjd double check I put the right milestone on this one!

PR #509, opened 2020-04-15
CA cert for Voltron must be provided

asincu:
## Description

Voltron needs to receive a certificate CA and private key in order to start up the tunnel server. This will be used to create certificates for the managed clusters that will be used to connect to the tunnel server. 

Operator would create a certificate that used the value `voltron` as `SAN`. Voltron had this value hardcoded as the ServerName when checking certficates. If we need to provide a custom certificate, a DNS SAN entry would look like `*.mydomain.com`. Gurdian will not be able to connect to the server, as it will check for the value `voltron` to match `*.mydomain.com`.  A part of the fix is to remove this dependency in Voltron and default to the hostname provided to connect. 

Generating a default server with DNS entry set to `voltron` is a redundant step, as this certificate will never allow Guardians to connect to the Voltron sever. This is why we need to provide the secret as a mandatory step before registering any clusters. Generating the proper CA certificate requires that we know how our clients would like to make Voltron discoverable and it depends on the architecture of our clients. They can choose to create a public DNS entry for this service or create a reachable IP for this service. The root CA will need to be created with DNS entry like `*.mydomain.com` and/or IP entry like `x.x.x.x` matching voltron. 

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


pkg/render/manager.go
```diff
@@ -92,14 +92,8 @@ func Manager(
...
-	var tunnelSecrets []*corev1.Secret
+	var tunnelSecrets = []*corev1.Secret{tunnelSecret}
 	if management {
-		// If there is no secret create one and add it to the operator namespace.
-		if tunnelSecret == nil {
-			tunnelSecret = voltronTunnelSecret()
-			tunnelSecrets = append(tunnelSecrets, tunnelSecret)
-		}
-
 		tunnelSecrets = append(tunnelSecrets, CopySecrets(ManagerNamespace, tunnelSecret)...)
```
stevegaossou, line 97: I guess our approach here is that we won't have the operator call this out to the user. 

The new normal is that the user must always provide a tunnel cert (just like they must provide a tunnel service) manually. 

So we just need to document it in our user docs. 

I recall this is what was discussed already. I'm just writing it out just to confirm and job my own memory.

tmjd: In reference to your 1st sentence. The operator will report a missing certificate if one is not provided. See above at https://github.com/tigera/operator/pull/509/files#diff-8962c60cf9c10aa1ac7a2ffb1422ae3cR303

asincu:
Clients will be required to use self signed certificates and not provide their own CA certificates. In case they do, they need to add a subject alternate domain for `voltron`.

PR #510, opened 2020-04-15
Update operator to include new env variable

shirlz8:
Adding a new environment variable to store name of the node, used in bgp monitoring

### DIFF
--- a/pkg/render/fluentd_test.go
+++ b/pkg/render/fluentd_test.go
@@ -48,7 +48,7 @@ var _ = Describe("Tigera Secure Fluentd rendering tests", func() {
 		esConfigMap = render.NewElasticsearchClusterConfig("clusterTestName", 1, 1, 1)
 	})
 
-	It("should render all resources for a default configuration", func() {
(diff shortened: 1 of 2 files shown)


shirlz8:
@tmjd Added some tests for the env variables!

pkg/render/fluentd_test.go
```diff
@@ -70,6 +70,28 @@ var _ = Describe("Tigera Secure Fluentd rendering tests", func() {
...
+				ValueFrom: &corev1.EnvVarSource{
+					FieldRef: &corev1.ObjectFieldSelector{FieldPath: "spec.nodeName"},
+				},
+			},
+		}
+
+		for _, expected := range expectedEnvs {
+
+			Expect(envs).To(ContainElement(
+				expected))
```
stevegaossou, line 93: Very Minor: Formatting seems a bit off here. Was there a reason for the newline in line 91 and having the Expect statement split across two lines?

shirlz8: Updated!

PR #511, opened 2020-04-15
adding curator service account

penkeysuresh:
## Description

* adding a service account for curator to be able to restrict curator [PSP](https://kubernetes.io/docs/concepts/policy/pod-security-policy/) permissions

Pod Security Policy are based on service account used by the pod, so creating a new SA for curator allows us to create stricter PSP. This SA's sole purpose is to separate the curator from getting associated with `default` SA

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/logstorage.go
+++ b/pkg/render/logstorage.go
@@ -71,7 +71,8 @@ const (
 
 	LogStorageFinalizer = "tigera.io/eck-cleanup"
 
-	EsCuratorName = "elastic-curator"
+	EsCuratorName           = "elastic-curator"
+	EsCuratorServiceAccount = "tigera-elastic-curator"
 
 	// As soon as the total disk utilization exceeds the max-total-storage-percent,
 	// indices will be removed starting with the oldest. Picking a low value leads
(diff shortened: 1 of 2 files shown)


penkeysuresh:
From my initial testing, making the image run as root seemed sufficient. Didn't encounter any errors when it made not privileged. This needs to be tested more, since the errors might have gotten silenced etc.

penkeysuresh:
converting this to draft, operator needs more changes for curator component

tmjd:
What is the need for the Service Account. There are no role bindings for it so it doesn't seem useful to add a SA. Is it just so it doesn't get associated with a default one? Could you provide some reasoning for the SA. Thanks.
Also I think any RunAsNonRoot should be tested on an OpenShift cluster.

penkeysuresh:
as discussed on slack, we are creating new SA so that it curator doesn’t get associated with the default one, this new SA helps us create stricter pod security policies for curator as current PSP are based on service account used by the component.

I'll mention this in the code and also update the PR as discussed.

PR #512, opened 2020-04-16
Try out a GH actions workflow for commenting on PRs

caseydavenport:
## Description

Starting to delve into GH actions and see if they are useful. Trying a
basic comment action.

Uses this script for doing the commenting: https://github.com/harupy/comment-on-pr/blob/master/entrypoint.py

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/.github/workflows/comment-on-pr.yaml
+++ b/.github/workflows/comment-on-pr.yaml
@@ -1,0 +1,13 @@
+name: Remind about milestones and labels
+on: pull_request
+jobs:
+  comment:
+    name: Remind labels and milestones
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
+      - uses: harupy/comment-on-pr@master
+        env:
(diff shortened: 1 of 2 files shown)


.github/workflows/comment-on-pr.yaml
```diff
@@ -0,0 +1,13 @@
+name: Remind about milestones and labels
+on: pull_request
+jobs:
+  comment:
+    name: Remind labels and milestones
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
```
tmjd, line 8: I don't think this looks like it would be right. Should it be actions/pr? I've got no idea.

tmjd: Nevermind looks like this is needed to have access to the template.md in the repo.

caseydavenport:
We'll see if it works - I think it needs to be in master in order to run :D

PR #514, opened 2020-04-17
Use CNI private for Calico Enterprise installs

Brian-McM:
Multi NICs is only going to be available for Calico Enterprise installations which is why this change needs to happen

### DIFF
--- a/config/enterprise_versions.yml
+++ b/config/enterprise_versions.yml
@@ -68,3 +68,6 @@ components:
   guardian:
     image: tigera/guardian
     version: v2.7.0-0.dev-61-g73e2b02
+  tigera-cni:
+    image: tigera/cni
+    version: v3.0.0-0.dev-39-g3f96541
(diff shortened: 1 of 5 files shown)


config/enterprise_versions.yml
```diff
@@ -68,3 +68,6 @@ components:
   guardian:
     image: tigera/guardian
     version: v2.7.0-0.dev-61-g73e2b02
+  cni:
```
ozdanborne, line 71: The key of this yaml entry is important. 
- It needs to match what calico-private and tigera/process are using.
- It also should ideally be unique so that the key for the OS component and the Private variant of that same component don't use the same key. (this is a new rule because if they are identical they can't be put in the same versions.yml file which is annoying).


How about this intead?
```suggestion
  tigera-cni:
```

ozdanborne:
If we're no longer using OS calico/cni, we should update versions.yml of calico-private accordingly, since it's used to generate the image pull/retag/push instructions for airgapped clusters.

Remove this entry: https://github.com/tigera/calico-private/blob/master/_data/versions.yml#L94-L97

And add a new entry for "tigera-cni"

config/enterprise_versions.yml
Brian-McM: Sounds good!

ozdanborne:
You'll also need to update tigera/process to set "tigera-cni" when it's generating versions.yml to pass to gen-versions.

PR #533, opened 2020-04-22
Allow updates to Logstorage resource limits

Suraiya-Hameed:
## Description
Updating the Logstroage resource limit fails as storage requirements are immutable from ECK 1.0 https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-orchestration.html#k8s-orchestration-limitations

This PR uses hash of the volumeClaimTemplates as NodeSet name so a new StatefulSet is created if storage requirements change.

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/logstorage.go
+++ b/pkg/render/logstorage.go
@@ -546,12 +546,17 @@ func (es elasticsearchComponent) elasticsearchCluster() *esv1.Elasticsearch {
 // As storage requirements of NodeSets are immutable,
 // renaming a NodeSet automatically creates a new StatefulSet with new PersistentVolumeClaim.
 // https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-orchestration.html#k8s-orchestration-limitations
-func nodeSetName(pvcTemplate corev1.PersistentVolumeClaim) string{
+func nodeSetName(pvcTemplate corev1.PersistentVolumeClaim) string {
 	pvcTemplateHash := fnv.New64a()
-	templateBytes, _ := json.Marshal(pvcTemplate)
+	templateBytes, err := json.Marshal(pvcTemplate)
+	if err != nil {
+		log.V(5).Info("Failed to create unique name for ElasticSearch NodeSet.", "err", err)
+		return "es"
+	}
 	pvcTemplateHash.Write(templateBytes)
 	return hex.EncodeToString(pvcTemplateHash.Sum(nil))
 }
+
 func (es elasticsearchComponent) eckOperatorWebhookSecret() *corev1.Secret {
 	return &corev1.Secret{
 		ObjectMeta: metav1.ObjectMeta{
--- a/pkg/render/logstorage.go
+++ b/pkg/render/logstorage.go
@@ -15,7 +15,10 @@
 package render
 
 import (
+	"encoding/hex"
+	"encoding/json"
 	"fmt"
+	"hash/fnv"
 	"strings"
 
 	"github.com/elastic/cloud-on-k8s/pkg/utils/stringsutil"
@@ -499,6 +502,7 @@ func memoryQuantityToJVMHeapSize(q *resource.Quantity) string {
 // render the Elasticsearch CR that the ECK operator uses to create elasticsearch cluster
 func (es elasticsearchComponent) elasticsearchCluster() *esv1.Elasticsearch {
 	nodeConfig := es.logStorage.Spec.Nodes
+	pvcTemplate := es.pvcTemplate()
 
 	return &esv1.Elasticsearch{
 		TypeMeta: metav1.TypeMeta{Kind: "Elasticsearch", APIVersion: "elasticsearch.k8s.elastic.co/v1"},
@@ -522,7 +526,7 @@ func (es elasticsearchComponent) elasticsearchCluster() *esv1.Elasticsearch {
 			NodeSets: []esv1.NodeSet{
 				{
 					Count: int32(nodeConfig.Count),
-					Name:  "es",
+					Name: nodeSetName(pvcTemplate),
 					Config: &cmnv1.Config{
 						Data: map[string]interface{}{
 							"node.master": "true",
@@ -530,7 +534,7 @@ func (es elasticsearchComponent) elasticsearchCluster() *esv1.Elasticsearch {
 							"node.ingest": "true",
 						},
 					},
-					VolumeClaimTemplates: []corev1.PersistentVolumeClaim{es.pvcTemplate()},
+					VolumeClaimTemplates: []corev1.PersistentVolumeClaim{pvcTemplate},
 					PodTemplate:          es.podTemplate(),
 				},
 			},
@@ -538,6 +542,16 @@ func (es elasticsearchComponent) elasticsearchCluster() *esv1.Elasticsearch {
 	}
 }
 
+// nodeSetName returns thumbprint of PersistentVolumeClaim object as string.
+// As storage requirements of NodeSets are immutable,
+// renaming a NodeSet automatically creates a new StatefulSet with new PersistentVolumeClaim.
+// https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-orchestration.html#k8s-orchestration-limitations
+func nodeSetName(pvcTemplate corev1.PersistentVolumeClaim) string{
+	pvcTemplateHash := fnv.New64a()
+	templateBytes, _ := json.Marshal(pvcTemplate)
+	pvcTemplateHash.Write(templateBytes)
+	return hex.EncodeToString(pvcTemplateHash.Sum(nil))
+}
 func (es elasticsearchComponent) eckOperatorWebhookSecret() *corev1.Secret {
 	return &corev1.Secret{
 		ObjectMeta: metav1.ObjectMeta{
--- a/pkg/render/logstorage_test.go
+++ b/pkg/render/logstorage_test.go
@@ -16,7 +16,6 @@ package render_test
 
 import (
 	"fmt"
-
 	esv1 "github.com/elastic/cloud-on-k8s/pkg/apis/elasticsearch/v1"
 	kbv1 "github.com/elastic/cloud-on-k8s/pkg/apis/kibana/v1"
 	. "github.com/onsi/ginkgo"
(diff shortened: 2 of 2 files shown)


pkg/render/logstorage.go
```diff
@@ -522,22 +526,32 @@ func (es elasticsearchComponent) elasticsearchCluster() *esv1.Elasticsearch {
 			NodeSets: []esv1.NodeSet{
 				{
 					Count: int32(nodeConfig.Count),
-					Name:  "es",
+					Name: "es-" + nodeSetNameSuffix(pvcTemplate),
```
rene-dekker, line 529: You can omit the "es-" bit

pkg/render/logstorage_test.go
```diff
@@ -267,6 +267,99 @@ var _ = Describe("Elasticsearch rendering tests", func() {
 
 		Context("Deleting LogStorage", deleteLogStorageTests(operatorv1.ClusterManagementTypeStandalone))
 
+		Context("Updating LogStorage resource", func() {
```
rene-dekker, line 270: Does this test fail without your fix? I'm not sure if the fake k8s client we use actually captures the problem.

rene-dekker: Could you add an assertion on a spec change?

pkg/render/logstorage.go
Brian-McM: Is a prefix added to the node set automatically? If not we should add "tigera" or something like that.

pkg/render/logstorage.go
```diff
@@ -522,22 +526,32 @@ func (es elasticsearchComponent) elasticsearchCluster() *esv1.Elasticsearch {
...
 }
 
+// nodeSetNameSuffix returns thumbprint of PersistentVolumeClaim object as string.
+// As storage requirements of NodeSets are immutable,
+// renaming a NodeSet automatically creates a new StatefulSet with new PersistentVolumeClaim.
+func nodeSetNameSuffix(pvcTemplate corev1.PersistentVolumeClaim) string{
+	pvcTemplateHash := md5.New()
+	templateBytes, _ := json.Marshal(pvcTemplate)
+	pvcTemplateHash.Write(templateBytes)
+	// Slice to fit NodeSets Name's max length of 23 char
```
stevegaossou, line 552: @Suraiya-Hameed could you reference the actual Elastic doc that states this? 

Important restriction and good to provide a reference here.

pkg/render/logstorage.go
Suraiya-Hameed: no, what we give is used as it is for NodeSet name, with the current change pod name becomes something like `tigera-secure-es-es-1dacabc7d15654312c8c-0` from original `tigera-secure-es-es-0`.

Suraiya-Hameed: I'll remove `es-` prefix from the NodeSet name, it'll be less redundant, so the statefulset generated will be `tigera-secure-es-<thumbprint>`

rene-dekker: Is it hard to get a shorter hash?

stevegaossou: > You can omit the "es-" bit
> Is a prefix added to the node set automatically? If not we should add "tigera" or something like that.

Agree with Rene. You don't need any prefix in the NodeSet name, because ECK already adds another prefix (from the Elasticsearch resource name). 

I was curious about this myself, so I looked at the Elastic docs. 

[Doc states](https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-orchestration.html#k8s-statefulsets): 
> the StatefulSet name is built from the Elasticsearch resource name and the NodeSet name. Each Pod will be assigned the StatefulSet name suffixed by an ordinal. The corresponding Elasticsearch node has the same name as the Pod.

So for example, if the NodeSet name is `es` and our Elasticsearch resource name is `tigera-secure` (which is what it is currently), then from the Elastic docs it sounds like StatefulSet name should be `tigera-secure-es`. It's actually `tigera-secure-es-es` (I checked in a running cluster). 

The Elasticsearch pod created from the StatefulSet is called `tigera-secure-es-es-0`. 

So other than the double `es` it seems to fit what the documentation says.

pkg/render/logstorage.go
stevegaossou: I couldn't find a mention of this max length in the docs. Could you share the doc you found on this?

pkg/render/logstorage.go
```diff
@@ -522,22 +526,33 @@ func (es elasticsearchComponent) elasticsearchCluster() *esv1.Elasticsearch {
...
+// nodeSetName returns thumbprint of PersistentVolumeClaim object as string.
+// As storage requirements of NodeSets are immutable,
+// renaming a NodeSet automatically creates a new StatefulSet with new PersistentVolumeClaim.
+// https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-orchestration.html#k8s-orchestration-limitations
+func nodeSetName(pvcTemplate corev1.PersistentVolumeClaim) string{
+	pvcTemplateHash := md5.New()
+	templateBytes, _ := json.Marshal(pvcTemplate)
+	pvcTemplateHash.Write(templateBytes)
+	// Slice to fit NodeSets Name's max length of 23 char
+	return hex.EncodeToString(pvcTemplateHash.Sum(nil))[:23]
```
stevegaossou, line 554: Everything hinges on this returning a unique hash. What guarantees the first 20 chars are unique?

pkg/render/logstorage_test.go
```diff
@@ -267,6 +267,104 @@ var _ = Describe("Elasticsearch rendering tests", func() {
...
+			It("should create new NodeSet", func() {
+				logStorage2 := &operator.LogStorage{
```
stevegaossou, line 272: Minor: Why is this called `logStroage2`? The number 2 seems like noise (not important).

pkg/render/logstorage.go
Suraiya-Hameed: https://github.com/elastic/cloud-on-k8s/blob/0404e804124f51fb7095f2ff2eb905b399067b42/pkg/apis/elasticsearch/v1beta1/elasticsearch_types.go#L63

pkg/render/logstorage_test.go
stevegaossou: Everything about this fix hinges on the function `nodeSetName()` generating a unique (20-char) hash substring given corev1.PersistentVolumeClaim instances that contain different values. 

Would like to see a few more tests perhaps just for `nodeSetName()` itself (unit style tests) to illustrate this. Are there an edge cases where this isn't true?

pkg/render/logstorage.go
Suraiya-Hameed: >the StatefulSet name is built from the Elasticsearch resource name and the NodeSet name. Each Pod will be assigned the StatefulSet name suffixed by an ordinal. The corresponding Elasticsearch node has the same name as the Pod.
https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-orchestration.html#k8s-statefulsets
Thats how we are getting `tigera-secure-es-es-0`

pkg/render/logstorage.go
Suraiya-Hameed: I couldn't find documentation around it, I got the error saying max length is 23 when I used full hash output (32 char).

pkg/render/logstorage.go
Suraiya-Hameed: 128bit is the minimum length of hash output, that would be 32 char. I was looking at options to get 23 char unique value, but all the docs I referred suggested truncating as collision will be pretty low even with that.

Brian-McM: > Everything hinges on this returning a unique hash. What guarantees the first 20 chars are unique?
It would be _extremely_ unlikely that even the first 5 characters would be the same, using 23 characters is quite the overkill. One character differences in two strings going through a hash function usually generate wildly different hashes.

This will generate huge pod names, which might seem a little weird to our customers, and I think the likely hood of collision with even 5 suffixes is so small.

One other note is that only one suffix ever exists at a time, and we don't need to generate a hash that's never been used before, we only need to generate a hash that was different from the previous ones. The likely hood of generating two hash in a row with 5 character likes that have different generated from unique PVCs is extremely unlikely, to say the least.

Suraiya-Hameed: I've switched to non-cryptographic hash with smaller output length.

pkg/render/logstorage.go
```diff
@@ -522,22 +526,32 @@ func (es elasticsearchComponent) elasticsearchCluster() *esv1.Elasticsearch {
...
 	}
 }
 
+// nodeSetName returns thumbprint of PersistentVolumeClaim object as string.
+// As storage requirements of NodeSets are immutable,
+// renaming a NodeSet automatically creates a new StatefulSet with new PersistentVolumeClaim.
+// https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-orchestration.html#k8s-orchestration-limitations
+func nodeSetName(pvcTemplate corev1.PersistentVolumeClaim) string{
+	pvcTemplateHash := fnv.New64a()
+	templateBytes, _ := json.Marshal(pvcTemplate)
```
Brian-McM, line 551: This returns an error and it should be handled. I think the pattern in render is to panic, since we don't return errors. If there's an error the below line likely panicsince templateBytes will be nil.

Brian-McM: Maybe confirm with @caseydavenport or @tmjd that this is OK, sometimes our opinions with panicing differ. This needs to be handle somehow though.

caseydavenport: blech, yeah it's probably fine.

I don't really like the pattern that we've established to "panic" in render - it's not very graceful or resilient to potential bugs in the code.

Ideally we'd propagate an error back up to report to the user in the TigeraStatus, but that's a larger change than I want to make you do in this PR.

Something better but not as good would be to log out a warning and then return sensible output. e.g., if we can't calculate a hash, we just return `es` like we do today?

Suraiya-Hameed: >Something better but not as good would be to log out a warning and then return sensible output. e.g., if we can't calculate a hash, we just return es like we do today?

Sure sounds better than panicking.

PR #541, opened 2020-04-23
Secret manager-tls must be shared with compliance and kube-controllers

asincu:
## Description

Manager-tls secret must be shared with compliance-server and kube-controllers. This will be added to the set of trusted certificates when making calls to retrieve compliance reports from a managed cluster. 

This change is required as we are currently using InsecureSkipVerify: true when making this request.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

```release-note
Verify TLS between management cluster components and managed clusters. 
```

### DIFF
--- a/pkg/controller/installation/core_controller.go
+++ b/pkg/controller/installation/core_controller.go
@@ -453,6 +453,21 @@ func (r *ReconcileInstallation) Reconcile(request reconcile.Request) (reconcile.
 		return reconcile.Result{}, err
 	}
 
+	var managerTLSSecret *corev1.Secret
+	managerTLSSecret, err = utils.ValidateCertPair(r.client,
+		render.ManagerTLSSecretName,
+		render.ManagerSecretKeyName,
+		render.ManagerSecretCertName,
+	)
+
+	if instance.Spec.ClusterManagementType == operator.ClusterManagementTypeManagement {
+		if err != nil {
+			log.Error(err, "Invalid manager TLS Cert")
+			r.status.SetDegraded("Error validating manager TLS certificate", err.Error())
+			return reconcile.Result{}, err
+		}
+	}
+
 	typhaNodeTLS, err := r.GetTyphaFelixTLSConfig()
 	if err != nil {
 		log.Error(err, "Error with Typha/Felix secrets")
@@ -497,6 +512,7 @@ func (r *ReconcileInstallation) Reconcile(request reconcile.Request) (reconcile.
 		instance,
 		pullSecrets,
 		typhaNodeTLS,
+		managerTLSSecret,
 		birdTemplates,
 		instance.Spec.KubernetesProvider,
 		netConf,
--- a/pkg/controller/manager/manager_controller.go
+++ b/pkg/controller/manager/manager_controller.go
@@ -97,7 +97,7 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 			render.VoltronTunnelSecretName, render.ComplianceServerCertSecret,
 		} {
 			if err = utils.AddSecretsWatch(c, secretName, namespace); err != nil {
-				return fmt.Errorf("compliance-controller failed to watch the secret '%s' in '%s' namespace: %v", secretName, namespace, err)
+				return fmt.Errorf("manager-controller failed to watch the secret '%s' in '%s' namespace: %v", secretName, namespace, err)
 			}
 		}
 	}
@@ -205,6 +205,28 @@ func (r *ReconcileManager) Reconcile(request reconcile.Request) (reconcile.Resul
 		return reconcile.Result{}, err
 	}
 
+	// Check that if the manager certpair secret exists that it is valid (has key and cert fields)
+	// If it does not exist then this function returns a nil secret but no error and a self-signed
+	// certificate will be generated when rendering below.
+	tlsSecret, err := utils.ValidateCertPair(r.client,
+		render.ManagerTLSSecretName,
+		render.ManagerSecretKeyName,
+		render.ManagerSecretCertName,
+	)
+
+	// An error is returned in case the read cannot be performed of the secret does not match the expected format
+	// In case the secret is not found, the error and the secret will be nil. This check needs to be done for all
+	// cluster types. For management cluster, we also need to check if the secret was created before hand.
+	if err != nil {
+		r.status.SetDegraded("Error validating manager TLS certificate", err.Error())
+		return reconcile.Result{}, err
+	}
+	if installation.Spec.ClusterManagementType == operatorv1.ClusterManagementTypeManagement && tlsSecret == nil {
+		err = fmt.Errorf("manager TLS Secret not found")
+		r.status.SetDegraded("No TLS manager certificate", err.Error())
+		return reconcile.Result{}, err
+	}
+
 	// Check that compliance is running.
 	compliance, err := compliance.GetCompliance(ctx, r.client)
 	if err != nil {
@@ -231,19 +253,6 @@ func (r *ReconcileManager) Reconcile(request reconcile.Request) (reconcile.Resul
 		return reconcile.Result{}, err
 	}
 
-	// Check that if the manager certpair secret exists that it is valid (has key and cert fields)
-	// If it does not exist then this function returns a nil secret but no error and a self-signed
-	// certificate will be generated when rendering below.
-	tlsSecret, err := utils.ValidateCertPair(r.client,
-		render.ManagerTLSSecretName,
-		render.ManagerSecretKeyName,
-		render.ManagerSecretCertName,
-	)
-	if err != nil {
-		log.Error(err, "Invalid TLS Cert")
-		r.status.SetDegraded("Error validating TLS certificate", err.Error())
-		return reconcile.Result{}, err
-	}
 
 	pullSecrets, err := utils.GetNetworkingPullSecrets(installation, r.client)
 	if err != nil {
--- a/pkg/render/compliance.go
+++ b/pkg/render/compliance.go
@@ -49,10 +49,12 @@ const (
 	ComplianceServerKeyName    = "tls.key"
 
 	complianceServerTLSHashAnnotation = "hash.operator.tigera.io/tls-certificate"
+	ManagerTLSHashAnnotation          = "hash.operator.tigera.io/manager-certificate"
 )
 
 func Compliance(
 	esSecrets []*corev1.Secret,
+	managerSecret *corev1.Secret,
 	installation *operatorv1.Installation,
 	complianceServerCertSecret *corev1.Secret,
 	esClusterConfig *ElasticsearchClusterConfig,
@@ -79,6 +81,7 @@ func Compliance(
 
 	return &complianceComponent{
 		esSecrets:                   esSecrets,
+		managerSecret:               managerSecret,
 		installation:                installation,
 		esClusterConfig:             esClusterConfig,
 		pullSecrets:                 pullSecrets,
@@ -89,6 +92,7 @@ func Compliance(
 
 type complianceComponent struct {
 	esSecrets                   []*corev1.Secret
+	managerSecret               *corev1.Secret
 	installation                *operatorv1.Installation
 	esClusterConfig             *ElasticsearchClusterConfig
 	pullSecrets                 []*corev1.Secret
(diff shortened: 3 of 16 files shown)


pkg/render/manager.go
```diff
@@ -80,6 +82,7 @@ func Manager(
 			ManagerSecretCertName,
 			825*24*time.Hour, // 825days*24hours: Create cert with a max expiration that macOS 10.15 will accept
 			nil,
+			ManagerServiceExternalDNS, ManagerServiceInternalDNS,
```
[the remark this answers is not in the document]
stevegaossou: On closer inspection, I guess this was intentional, since the function is variadic with the variable amount of hostnames `hostnames ...string`. 

But in that case you could use a slice and then use "..." to make it clearer. 

```suggestion
	hostnames := []string{ManagerServiceExternalDNS, ManagerServiceInternalDNS}
	...
	tlsKeyPair, err = CreateOperatorTLSSecret(nil,
			ManagerTLSSecretName,
			ManagerSecretKeyName,
			ManagerSecretCertName,
			825*24*time.Hour, // 825days*24hours: Create cert with a max expiration that macOS 10.15 will accept
			nil,
			hostnames...,
		)
```

pkg/controller/installation/core_controller.go
```diff
@@ -453,6 +453,25 @@ func (r *ReconcileInstallation) Reconcile(request reconcile.Request) (reconcile.
...
+			825*24*time.Hour, // 825days*24hours: Create cert with a max expiration that macOS 10.15 will accept
+			nil,
+			render.ManagerServiceExternalDNS,
+			render.ManagerServiceInternalDNS,
+		)
+		if err != nil {
+			r.SetDegraded("Error generating self signed certificates for tigera-manager", err, reqLogger)
+			return reconcile.Result{}, err
+		}
+	}
```
stevegaossou, line 473: Looks like we reproduce the tigera-manager cert secret for Compliance and for Kube controller, using two different ways. 

Compliance we use `utils.ValidateCertPair()` and then pass to Compliance render, where it does an actual copy of the secret. 

Kube controllers, we do this `render.CreateOperatorTLSSecret()` directly. 

Was there a specific reason for the difference? I was going to guess that it's because Compliance component has its own render (but then it looks like Kube controller does as well).

pkg/render/manager.go
```diff
@@ -33,15 +33,17 @@ import (
...
-	ManagerSecretKeyName    = "key"
-	ManagerSecretCertName   = "cert"
-	ManagerOIDCConfig       = "tigera-manager-oidc-config"
-	ManagerOIDCWellknownURI = "/usr/share/nginx/html/.well-known"
-	ManagerOIDCJwksURI      = "/usr/share/nginx/html/discovery"
+	managerPort               = 9443
+	managerTargetPort         = 9443
+	ManagerNamespace          = "tigera-manager"
+	ManagerServiceInternalDNS = "tigera-manager.tigera-manager.svc"
+	ManagerServiceExternalDNS = "localhost"
```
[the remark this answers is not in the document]
asincu: I think we need localhost for the browser and external service and tigera-manager.tigera-manager.svc to be used internally in the k8s cluster.

pkg/controller/installation/core_controller.go
asincu: Compliance does not create the secret, it just copies it in its namespace. The core-controllers creates the secret and that make it available to be other components.

pkg/render/manager.go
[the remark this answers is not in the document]
asincu: Hm ... I take the thumbs up back. :) I do not think we need to create a slice. The purpose of a variadic function is to allow the caller to use any numbers of parameters.

pkg/render/manager.go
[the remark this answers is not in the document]
asincu: I think I will rename them as I see how the perspective can cause confusion.

pkg/render/manager.go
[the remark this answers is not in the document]
stevegaossou: Yes, the variadic part of this is the hostnames, which can be any number of them. 

I was saying that if you intended to have both ManagerServiceExternalDNS and ManagerServiceInternalDNS on the same line to indicate they are both hostnames, then a slice with the `...` allows you to communicate that more explicitly. 

You can ignore what I said if that wasn't the intention. Either way it's minor.

pkg/controller/installation/core_controller.go
stevegaossou: My question is then: 
- Why does core-controllers need to create the secret and make it available? 

Before this Manager needed the secret, which it must have gotten a copy from getting it passed from the tigera operator namespace. So why does core-controller need to create its own (instead of a copy). 

I understand that Compliance and Kube controller both now need access to this Manager TLS secret, but why does core controller need to create a new version? 

This is more for my understanding, since I'm missing a step here.

stevegaossou:
Should get a review from @tmjd or @caseydavenport.

pkg/controller/manager/manager_controller.go
```diff
@@ -205,6 +205,20 @@ func (r *ReconcileManager) Reconcile(request reconcile.Request) (reconcile.Resul
...
+		render.ManagerTLSSecretName,
+		render.ManagerSecretKeyName,
+		render.ManagerSecretCertName,
+	)
+	if err != nil {
+		log.Error(err, "Invalid TLS Cert")
+		r.status.SetDegraded("Error validating TLS certificate", err.Error())
+		return reconcile.Result{}, err
+	}
+
```
tmjd, line 221: Also when installation.clusterManagementType indicates management then we should validate that tlsSecret is not nil.

pkg/render/compliance.go
```diff
@@ -601,8 +613,28 @@ func (c *complianceComponent) complianceServerDeployment() *appsv1.Deployment {
...
 						},
 					},
+				}},
+				{
+					Name: "manager-cert",
+					VolumeSource: corev1.VolumeSource{
+						Secret: &corev1.SecretVolumeSource{
+							DefaultMode: &defaultMode,
+							SecretName:  ManagerTLSSecretName,
+							Optional:   &optional,
```
tmjd, line 623: I don't think we should make this optional. We know if the cert is being rendered or not so we can be confident if this mount is needed or not.

pkg/render/compliance.go
```diff
@@ -156,6 +159,9 @@ func (c *complianceComponent) Objects() ([]runtime.Object, []runtime.Object) {
 	}
 
 	complianceObjs = append(complianceObjs, secretsToRuntimeObjects(CopySecrets(ComplianceNamespace, c.esSecrets...)...)...)
+	if c.managerSecret != nil {
```
tmjd, line 162: See Line 545 (AnnotationHash) for the annotation that we need to add for this Secret.

pkg/render/compliance.go
```diff
@@ -581,6 +588,11 @@ func (c *complianceComponent) complianceServerDeployment() *appsv1.Deployment {
 						Name:      "cert",
 						MountPath: "/code/apiserver.local.config/certificates",
 						ReadOnly:  true,
+					},
+					{
```
tmjd, line 592: As with the VolumeSource, this should be added only if the secret is not nil.

pkg/render/compliance_test.go
```diff
@@ -90,9 +91,90 @@ var _ = Describe("compliance rendering tests", func() {
 		})
 	})
 
+	Context("Management cluster", func() {
+		It("should render all resources for a default configuration", func() {
+			component, err := render.Compliance(nil, &corev1.Secret{
```
tmjd, line 96: With the secret specified we should check that the mount is added to the deployment.

pkg/render/kube-controllers_test.go
```diff
@@ -100,10 +101,45 @@ var _ = Describe("kube-controllers rendering tests", func() {
...
+
+		// Should render the correct resources.
+		ExpectResource(resources[0], "calico-kube-controllers", "calico-system", "", "v1", "ServiceAccount")
+		ExpectResource(resources[1], "calico-kube-controllers", "", "rbac.authorization.k8s.io", "v1", "ClusterRole")
+		ExpectResource(resources[2], "calico-kube-controllers", "", "rbac.authorization.k8s.io", "v1", "ClusterRoleBinding")
+		ExpectResource(resources[3], "calico-kube-controllers", "calico-system", "apps", "v1", "Deployment")
+		ExpectResource(resources[4], render.ManagerTLSSecretName, "calico-system", "", "v1", "Secret")
+
+		// The Deployment should have the correct configuration.
+		ds := resources[3].(*apps.Deployment)
```
tmjd, line 134: We should verify the mount and volume are present when the secret is specified.

pkg/render/kube-controllers.go
```diff
@@ -259,6 +269,33 @@ func (c *kubeControllersComponent) controllersDeployment() *apps.Deployment {
...
+						},
+					},
+					Volumes: []v1.Volume{
+						{
+							Name: "manager-cert",
+							VolumeSource: v1.VolumeSource{
+								Secret: &v1.SecretVolumeSource{
+									DefaultMode: &defaultMode,
+									SecretName:  ManagerTLSSecretName,
+									Optional:   &optional,
```
tmjd, line 286: Same as in the compliance, we know if the secret exists or not so there should be no need for this to be an optional mount.

pkg/render/kube-controllers.go
```diff
@@ -30,23 +30,30 @@ import (
...
 func (c *kubeControllersComponent) Objects() ([]runtime.Object, []runtime.Object) {
-	return []runtime.Object{
+	kubeControllerObjects := []runtime.Object{
 		c.controllersServiceAccount(),
 		c.controllersRole(),
 		c.controllersRoleBinding(),
 		c.controllersDeployment(),
-	}, nil
+	}
+	if c.managerSecret != nil {
```
tmjd, line 52: It doesn't go here but in render/kube-controllers.go but with this same case check we need to add an AnnotationHash to the deployment to make sure the deployment will be updated/restarted when this secret is CRUD'ed.

pkg/render/compliance.go
```diff
@@ -601,8 +613,28 @@ func (c *complianceComponent) complianceServerDeployment() *appsv1.Deployment {
...
 					},
+				}},
+				{
+					Name: "manager-cert",
+					VolumeSource: corev1.VolumeSource{
+						Secret: &corev1.SecretVolumeSource{
+							DefaultMode: &defaultMode,
+							SecretName:  ManagerTLSSecretName,
+							Optional:   &optional,
+							Items: []corev1.KeyToPath{
```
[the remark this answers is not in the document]
asincu: I have kept only the certificate, as we do not need the key.

pkg/controller/installation/core_controller.go
```diff
@@ -453,6 +453,13 @@ func (r *ReconcileInstallation) Reconcile(request reconcile.Request) (reconcile.
 		return reconcile.Result{}, err
 	}
 
+	var managerTLSSecret *corev1.Secret
+	managerTLSSecret, err = utils.ValidateCertPair(r.client,
```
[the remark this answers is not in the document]
asincu: Isn't this done already here https://github.com/tigera/operator/blob/master/pkg/controller/installation/core_controller.go#L122 ?

pkg/render/kube-controllers.go
asincu: We might need to sync up on this comment. I think the comment might have been needed on a different file than `render/kube_controllers.go` ?

pkg/controller/manager/manager_controller.go
asincu: Added in https://github.com/tigera/operator/pull/541/commits/2ca889a930515473ddc193328e46b06c8760bc73

pkg/controller/installation/core_controller.go
```diff
@@ -453,6 +453,13 @@ func (r *ReconcileInstallation) Reconcile(request reconcile.Request) (reconcile.
...
+		render.ManagerTLSSecretName,
+		render.ManagerSecretKeyName,
+		render.ManagerSecretCertName,
+	)
+
```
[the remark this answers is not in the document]
asincu: Added in https://github.com/tigera/operator/pull/541/commits/2ca889a930515473ddc193328e46b06c8760bc73

pkg/controller/installation/core_controller.go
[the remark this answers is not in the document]
tmjd: Yep, you're right.

pkg/render/kube-controllers.go
tmjd: You've done what I wanted with the kubeControllerAnnotations

pkg/controller/installation/core_controller.go
```diff
@@ -453,6 +453,24 @@ func (r *ReconcileInstallation) Reconcile(request reconcile.Request) (reconcile.
...
+	if instance.Spec.ClusterManagementType == operator.ClusterManagementTypeManagement {
+		if managerTLSSecret == nil {
```
tmjd, line 464: This should be nil for a new cluster, so I don't think we should have this check.

tmjd:
Trying to trigger Semaphore v2 job

PR #543, opened 2020-04-24
Expose BGP metrics for Prometheus.

stevegaossou:
## Description
- Adding new port to calico-node-metrics service to expose BGP metrics for Prometheus.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/node.go
+++ b/pkg/render/node.go
@@ -43,6 +43,9 @@ var (
 	// The port used by calico/node to report Calico Enterprise internal metrics.
 	// This is separate from the calico/node prometheus metrics port, which is user configurable.
 	nodeReporterPort int32 = 9081
+	// The port used by calico/node to report Calico Enterprise BGP metrics.
+	// This is currently not intended to be user configurable.
+	nodeBGPReporterPort int32 = 9900
 )
 
 // Node creates the node daemonset and other resources for the daemonset to operate normally.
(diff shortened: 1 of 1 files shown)


pkg/render/node.go
```diff
@@ -992,6 +995,12 @@ func (c *nodeComponent) nodeMetricsService() *v1.Service {
 					TargetPort: intstr.FromInt(int(nodeReporterPort)),
 					Protocol:   v1.ProtocolTCP,
 				},
+				v1.ServicePort{
```
tmjd, line 998: Does the comment on this function apply to this port also? I think the answer is yes but just wanted to call that out.

pkg/render/node.go
```diff
@@ -43,6 +43,9 @@ var (
 	// The port used by calico/node to report Calico Enterprise internal metrics.
 	// This is separate from the calico/node prometheus metrics port, which is user configurable.
 	nodeReporterPort int32 = 9081
+	// The port used by calico/node to report Calico Enterprise BGP metrics.
+	// This is currently not intended to be user configurable.
+	nodeBGPReporterPort int32 = 9900
```
tmjd, line 48: For `nodeReporterPort` there is also an env var set on the node container, do we need the same for this?

stevegaossou: Good question. 

There is an ENV variable, but the plan for now was to use its defaults. Unless, we think it's better to have a node container ENV in Operator (just for documenting the variable itself), I was going to leave it out until we actually want to expose it to the user in the future.

pkg/render/node.go
stevegaossou: Thanks for catching that. Fixed now.

pkg/render/node.go
tmjd: I think leaving it out is a good plan. Ideally we can do that for most things and that way if we add datastore configuration for options we don't need to remove env options to use them.

stevegaossou:
Tested it using local version of operator and it works as expected.

PR #550, opened 2020-04-24
Add operator support for RBAC for Flowviz for MCM.

rene-dekker:
When a user is not allowed to list pods/endpoints in a given namespace, it should also not see them in the flow visualization in the manager UI. This feature was implemented in an earlier release and now made compatible with multi-cluster management.

When a given user makes a request in the UI from the management cluster to a managed cluster, an rbac call will be made in the managed cluster by guardian, while impersonating the tigera-manager. For that reason we need the sa & ns & rbac to be setup there.

### DIFF
--- a/pkg/controller/clusterconnection/clusterconnection_controller.go
+++ b/pkg/controller/clusterconnection/clusterconnection_controller.go
@@ -16,16 +16,19 @@ package clusterconnection
 
 import (
 	"context"
+	"errors"
 	"fmt"
+
 	"github.com/tigera/operator/pkg/controller/status"
 
 	"github.com/tigera/operator/pkg/controller/installation"
+	controllermanager "github.com/tigera/operator/pkg/controller/manager"
 	"github.com/tigera/operator/pkg/controller/utils"
 	"github.com/tigera/operator/pkg/render"
 
 	operatorv1 "github.com/tigera/operator/pkg/apis/operator/v1"
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -87,6 +90,12 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 		return fmt.Errorf("%s failed to watch Network resource: %v", controllerName, err)
 	}
 
+	// Watch for changes to primary resource Manager
+	err = c.Watch(&source.Kind{Type: &operatorv1.Manager{}}, &handler.EnqueueRequestForObject{})
+	if err != nil {
+		return fmt.Errorf("%v failed to watch primary resource: %v", controllerName, err)
+	}
+
 	return nil
 }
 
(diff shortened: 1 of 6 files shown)


pkg/components/enterprise.go
```diff
@@ -15,165 +15,141 @@
...
 var (
-	
 	ComponentAPIServer = component{
 		Version: "v2.7.0-0.dev-70-gcf61025a",
 		Digest:  "sha256:99001709c7e9079bd41f466ff75bfac87369c84b4096994a04fbceaf3c447c85",
 		Image:   "tigera/cnx-apiserver",
 	}
-	
-	
+
```
rene-dekker, line 23: Is there a reason we have two newlines here? Goimports keeps cleaning them up for me, which I think is  good.

pkg/controller/clusterconnection/clusterconnection_controller.go
```diff
@@ -145,6 +154,21 @@ func (r *ReconcileConnection) Reconcile(request reconcile.Request) (reconcile.Re
 			operatorv1.ClusterManagementTypeManaged))
 	}
 
+	// Verify that the manager is not running.
+	_, err = controllermanager.GetManager(ctx, r.Client)
```
asincu, line 158: I suggest checking for cluster type managed instead of checking that the manager is not running.

pkg/render/manager.go
```diff
@@ -256,6 +260,21 @@ func (c *managerComponent) managerVolumes() []v1.Volume {
 				},
 			},
 		},
+		{
+			// We only want to mount the cert, not the private key to es-proxy to establish a connection with voltron.
+			Name: ManagerTLSSecretCertName,
```
asincu, line 265: I suggest making this read only.

pkg/controller/clusterconnection/clusterconnection_controller.go
rene-dekker: You are right. Right now there is no way you want guardian if it is not a managed cluster.

pkg/render/guardian.go
```diff
@@ -277,3 +282,51 @@ func (c *GuardianComponent) container() []v1.Container {
 		},
 	}
 }
+
+// This is a service account that is being impersonated when calls are made from the management cluster to do
+// subject access reviews in a managed cluster.
+func (c *GuardianComponent) managerImpersonationServiceAccount() *v1.ServiceAccount {
```
tmjd, line 288: Instead of creating a second function here for the SA, Role, and Binding, WDYT about making common ones from the ones in `pkg/render/manager.go` and pass in a flag to set which version to use? I think only the ClusterRole would change so that is probably the only one that needs the flag.
I'm thinking I don't want to see one of these get changed and not the other.
It looks like there is no reason for the function to be attached to the component.

pkg/render/manager.go
tmjd, line 265: Is it correct for this to be present all the time? (In other words is this needed for standalone clusters?)

pkg/components/enterprise.go
tmjd: That's the way the template file is written. We can change that template.

pkg/render/manager.go
```diff
@@ -447,6 +466,9 @@ func (c *managerComponent) managerEsProxyContainer() corev1.Container {
 		Image:           components.GetReference(components.ComponentEsProxy, c.installation.Spec.Registry, c.installation.Spec.ImagePath),
 		LivenessProbe:   c.managerEsProxyProbe(),
 		SecurityContext: securityContext(),
+		VolumeMounts: []corev1.VolumeMount{
```
tmjd, line 469: Same comment as above. (Do we need this all the time?)

rene-dekker: It can be removed.

pkg/render/guardian.go
rene-dekker: I didn't really want to break a pattern of them being methods, but seeing that you suggest it, I'm on board with the idea.

PR #556, opened 2020-04-29
Remove annotations for secret

asincu:
## Description

Removing annotations for manager-secret tls as the compliance and kube-controllers are restarted continuously by operator.

Please see the large number of replica sets for compliance and kube-controllers

perpetum_mobile

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


pkg/render/compliance.go
```diff
@@ -675,10 +675,6 @@ func complianceAnnotations(c *complianceComponent) map[string]string {
 		ManagerTLSHashAnnotation:          AnnotationHash(c.managerSecret),
```
stevegaossou, line 675: Doesn't this line mean the annotation is still set?

asincu: Yeah, but it does not seem to matter. I removed the method.

asincu: Yes, for compliance we still have to set an annotation. The problematic one is the one for the secret.

PR #558, opened 2020-04-29
Updates for Multi-NICs

Brian-McM:
Update pins and propogate MULTI_INTERFACE_MODE to the api server and query server

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/apiserver.go
+++ b/pkg/render/apiserver.go
@@ -568,6 +568,14 @@ func (c *apiServerComponent) apiServerContainer() corev1.Container {
 
 	isPrivileged := true
 
+	env := []corev1.EnvVar{
+		{Name: "DATASTORE_TYPE", Value: "kubernetes"},
+	}
+
+	if c.installation.Spec.CalicoNetwork != nil && c.installation.Spec.CalicoNetwork.MultiInterfaceMode != nil {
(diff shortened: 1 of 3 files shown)


pkg/render/apiserver.go
```diff
@@ -568,6 +568,16 @@ func (c *apiServerComponent) apiServerContainer() corev1.Container {
 
 	isPrivileged := true
 
+	env := []corev1.EnvVar{
+		{Name: "DATASTORE_TYPE", Value: "kubernetes"},
+	}
+
+	if c.installation.Spec.Variant == operator.TigeraSecureEnterprise {
```
caseydavenport, line 575: The apiserver is an enterprise only component, so don't need this check.

caseydavenport: Also, we should prevent setting `MultiInterfaceMode` to non-nil for Calico OSS clusters in validation anyway

Brian-McM: Right, I think we do but I'll double check

Brian-McM: We don't actually, but I'll make a separate PR to handle that.

PR #559, opened 2020-04-29
Remove MultiInterfaceMode if not EE installation

Brian-McM:
## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/installation/core_controller.go
+++ b/pkg/controller/installation/core_controller.go
@@ -328,9 +328,14 @@ func fillDefaults(instance *operator.Installation) error {
 			instance.Spec.CalicoNetwork.HostPorts = &hp
 		}
 
-		if instance.Spec.CalicoNetwork.MultiInterfaceMode == nil {
-			mm := operator.MultiInterfaceModeNone
-			instance.Spec.CalicoNetwork.MultiInterfaceMode = &mm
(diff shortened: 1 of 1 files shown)


pkg/controller/installation/core_controller.go
```diff
@@ -328,9 +328,14 @@ func fillDefaults(instance *operator.Installation) error {
 			instance.Spec.CalicoNetwork.HostPorts = &hp
 		}
 
-		if instance.Spec.CalicoNetwork.MultiInterfaceMode == nil {
-			mm := operator.MultiInterfaceModeNone
-			instance.Spec.CalicoNetwork.MultiInterfaceMode = &mm
+		if instance.Spec.Variant == operator.TigeraSecureEnterprise {
```
caseydavenport, line 331: I don't think this should be defaulting - it should be validation.

If the user specified both of these, we can't action it - we don't know which of the two is there true intent, and the two are not compatible with each other.

caseydavenport: Should be a check here: https://github.com/tigera/operator/blob/542547fdb4df61f69e1d8ce8283ea03a91b54ee3/pkg/controller/installation/validation.go#L30:6

Brian-McM: Ya that's a good point, I'll change that

PR #585, opened 2020-05-22
only build individual commits if on master or release branch

ozdanborne:
## Description

**Background**

Our initial `make cd CONFIRM=true` is broken down into three groups of image tag / pushes:
1. Branch (`quay.io/tigera/operator:release-v1.1`)
1. Release / Tags (`quay.io/tigera/operator:v1.1.0`)
1. Commit (`quay.io/tigera/operator:v1.1.0-0.dev-373-g12770777`)

This PR prevents 2 from being built and pushed to dockerhub from a branch that isn't `master` or `release-v*` (e.g. feature branches). 

**Release Builds**

The changes here prevent a problem where if you push a branch that is at the same commit as a tag, it will trigger a rebuild and overwrite of that release image on the docker registry.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -345,10 +345,7 @@ ifndef BRANCH_NAME
 	$(error BRANCH_NAME is undefined - run using make <target> BRANCH_NAME=var or set an environment variable)
 endif
 	$(MAKE) tag-images-all push-all push-manifests push-non-manifests IMAGETAG=${BRANCH_NAME} EXCLUDEARCH="$(EXCLUDEARCH)"
-	# only build individual commits if on master or release-v* branch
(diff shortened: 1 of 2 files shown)


tmjd:
I think we should keep the `#3` builds. They are dev builds, so even if we would rebuild it and push the same if the hash were the same I don't think we would care. Is there something I'm missing on why that would be bad?

ozdanborne:
@tmjd Yeah that's understandable. In the final paragraph above, I try to explain that I'm preventing us from having two images with `v1.0.0-0.dev-374-<commit>` tags. But the <commit> will be unique so it's not a _huge_ deal...

PR #598, opened 2020-05-29
making fluentd, apiserver, reporter privileged only on openshift

penkeysuresh:
## Description

Making changes to conditionally enable `privileged` mode for `apiserver`, `fluentd` and `compliance-reporter` on OCP so that on other platforms these pods can run without that permission. OCP needs privileged mode to access host path volumes. 

 

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/apiserver.go
+++ b/pkg/render/apiserver.go
@@ -566,7 +566,11 @@ func (c *apiServerComponent) apiServerContainer() corev1.Container {
 		{Name: "tigera-apiserver-certs", MountPath: "/code/apiserver.local.config/certificates"},
 	}
 
-	isPrivileged := true
+	isPrivileged := false
+	//On OpenShift apiserver needs privileged access to write audit logs to host path volume
+	if c.openshift {
+		isPrivileged = true
+	}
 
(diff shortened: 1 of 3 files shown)


tmjd:
This looks ok to me but makes me think. What is it that is enforcing the requirement on OpenShift making privileged necessary? Is this required on OpenShift because of a certain controller or component that could be installed on other clusters?

penkeysuresh:
This is due to SE-Linux context that's getting used. Currently this get set based on SELinux security module loaded on the host operating system. 

On other platforms (example gcp kubeadm) when the container is run as a `non root` user, SE Linux role defaults to `docker-default` (restrictive). But when we change this to `root` it changes to `unconfined`. This allows the container to run virtually without any confinements and allows it to access/write to host path volumes etc. 

On OCP when the container is run as `non root` the role defaults to `container_t` (restrictive) and stays the same when user becomes `root`. Where as when we make this to be privileged it becomes `spc_t` (super privileged container) and allows write access to host path volumes. On OCP I tried setting the roles manually which might have write access without much success.

PR #602, opened 2020-06-01
Adding new voltron-tls cert for voltron within tigera-manager component.

stevegaossou:
## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/installation/core_controller.go
+++ b/pkg/controller/installation/core_controller.go
@@ -455,7 +455,7 @@ func (r *ReconcileInstallation) Reconcile(request reconcile.Request) (reconcile.
 
 	var managerTLSSecret *corev1.Secret
 	managerTLSSecret, err = utils.ValidateCertPair(r.client,
-		render.ManagerTLSSecretName,
+		render.VoltronTLSSecretName,
 		render.ManagerSecretKeyName,
 		render.ManagerSecretCertName,
 	)
--- a/pkg/render/compliance.go
+++ b/pkg/render/compliance.go
@@ -49,12 +49,11 @@ const (
 	ComplianceServerKeyName    = "tls.key"
 
 	complianceServerTLSHashAnnotation = "hash.operator.tigera.io/tls-certificate"
-	ManagerTLSHashAnnotation          = "hash.operator.tigera.io/manager-certificate"
 )
 
 func Compliance(
 	esSecrets []*corev1.Secret,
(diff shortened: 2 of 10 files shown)


pkg/controller/installation/core_controller.go
```diff
@@ -455,7 +455,7 @@ func (r *ReconcileInstallation) Reconcile(request reconcile.Request) (reconcile.
 
 	var managerTLSSecret *corev1.Secret
 	managerTLSSecret, err = utils.ValidateCertPair(r.client,
-		render.ManagerTLSSecretName,
+		render.VoltronTLSSecretName,
```
stevegaossou, line 458: Ignore this. I'm not sure why I added this change. This should stay `ManagerTLSSecretName`.

pkg/render/compliance.go
```diff
@@ -49,12 +49,11 @@ const (
 	ComplianceServerKeyName    = "tls.key"
 
 	complianceServerTLSHashAnnotation = "hash.operator.tigera.io/tls-certificate"
-	ManagerTLSHashAnnotation          = "hash.operator.tigera.io/manager-certificate"
```
stevegaossou, line 52: Looks like this isn't being used anymore, so I thought it would be better to remove it.

pkg/render/render.go
```diff
@@ -120,21 +124,41 @@ func Calico(
...
 		if err != nil {
 			return nil, fmt.Errorf("generating certificates for manager was not finalized due to %v", err)
 		}
 		tss = append(tss, managerTLSSecret)
 	}
 
+	// In a Calico Enterprise management cluster we need a secondary TLS certificate for internal communication
+	// between voltron and other components (e.g. kube controller and compliance server).
+	var voltronTLSSecret *corev1.Secret
+	if isManagement {
```
stevegaossou, line 137: I think this is the only place we need this generation because we only need Kube controller and compliance server to have this cert set when it's a management cluster. 

Looks like both of those components treat this as optional setting so it should support this only being available for management cluster type.

stevegaossou: I also don't check that it's empty before doing this, because this `voltron-tls` secret is not intended to be overridden by the user.

stevegaossou:
Closed in favour of @asincu's PR: 
https://github.com/tigera/operator/pull/606

PR #606, opened 2020-06-03
Generate a certificate to be used for internal traffic for multiclust…

asincu:
…er management

## Description

When multicluster management is enabled, Voltron will make use of 3 certificates:
- manager-tls (This certificate can be changed with certificated registered to a valid dns domain). This certificate will be used for TLS communication between a browser and tigera UI
- manager-internal-tls (This certificate will be used for internal traffic within K8s cluster). This will be used when compliance/kube-controller/es-proxy need to initiate a TLS connection with tigera-manager.svc or localhost via Voltron)
- tigera-management-cluster-connection (This is a ca certificate that is generated by operator and provided to Voltron). This certificate is used to generate certificates for managed clusters.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/compliance/compliance_controller.go
+++ b/pkg/controller/compliance/compliance_controller.go
@@ -17,8 +17,9 @@ package compliance
 import (
 	"context"
 	"fmt"
-	corev1 "k8s.io/api/core/v1"
 	"time"
+
+	corev1 "k8s.io/api/core/v1"
 
 	operatorv1 "github.com/tigera/operator/pkg/apis/operator/v1"
 	"github.com/tigera/operator/pkg/controller/installation"
@@ -214,16 +215,16 @@ func (r *ReconcileCompliance) Reconcile(request reconcile.Request) (reconcile.Re
 		return reconcile.Result{}, err
 	}
 
-	var managerCertSecret *corev1.Secret
+	var managerInternalTLSSecret *corev1.Secret
 	if network.Spec.ClusterManagementType == operatorv1.ClusterManagementTypeManagement {
-		managerCertSecret, err = utils.ValidateCertPair(r.client,
-			render.ManagerTLSSecretName,
-			render.ManagerSecretCertName,
-			render.ManagerSecretKeyName,
+		managerInternalTLSSecret, err = utils.ValidateCertPair(r.client,
+			render.ManagerInternalTLSSecretName,
+			render.ManagerInternalSecretCertName,
+			render.ManagerInternalSecretKeyName,
 		)
 		if err != nil {
-			log.Error(err, fmt.Sprintf("failed to retrieve / validate %s", render.ManagerTLSSecretName))
-			r.status.SetDegraded(fmt.Sprintf("failed to retrieve / validate  %s", render.ManagerTLSSecretName), err.Error())
+			log.Error(err, fmt.Sprintf("failed to retrieve / validate %s", render.ManagerInternalSecretCertName))
+			r.status.SetDegraded(fmt.Sprintf("failed to retrieve / validate  %s", render.ManagerInternalSecretKeyName), err.Error())
 			return reconcile.Result{}, err
 		}
 	}
@@ -245,7 +246,7 @@ func (r *ReconcileCompliance) Reconcile(request reconcile.Request) (reconcile.Re
 	reqLogger.V(3).Info("rendering components")
 	openshift := r.provider == operatorv1.ProviderOpenShift
 	// Render the desired objects from the CRD and create or update them.
-	component, err := render.Compliance(esSecrets, managerCertSecret, network, complianceServerCertSecret, esClusterConfig, pullSecrets, openshift)
+	component, err := render.Compliance(esSecrets, managerInternalTLSSecret, network, complianceServerCertSecret, esClusterConfig, pullSecrets, openshift)
 	if err != nil {
 		log.Error(err, "error rendering Compliance")
 		r.status.SetDegraded("Error rendering Compliance", err.Error())
--- a/pkg/controller/manager/manager_controller.go
+++ b/pkg/controller/manager/manager_controller.go
@@ -221,11 +221,6 @@ func (r *ReconcileManager) Reconcile(request reconcile.Request) (reconcile.Resul
 		r.status.SetDegraded("Error validating manager TLS certificate", err.Error())
 		return reconcile.Result{}, err
 	}
-	if installation.Spec.ClusterManagementType == operatorv1.ClusterManagementTypeManagement && tlsSecret == nil {
-		err = fmt.Errorf("manager TLS Secret not found")
-		r.status.SetDegraded("No TLS manager certificate", err.Error())
-		return reconcile.Result{}, err
-	}
 
 	// Check that compliance is running.
 	compliance, err := compliance.GetCompliance(ctx, r.client)
@@ -252,7 +247,6 @@ func (r *ReconcileManager) Reconcile(request reconcile.Request) (reconcile.Resul
 		}
 		return reconcile.Result{}, err
 	}
-
 
 	pullSecrets, err := utils.GetNetworkingPullSecrets(installation, r.client)
 	if err != nil {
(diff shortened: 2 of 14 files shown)


pkg/controller/compliance/compliance_controller.go
```diff
@@ -214,16 +214,16 @@ func (r *ReconcileCompliance) Reconcile(request reconcile.Request) (reconcile.Re
 		return reconcile.Result{}, err
 	}
 
-	var managerCertSecret *corev1.Secret
+	var managerInternalTLSSecret *corev1.Secret
```
rene-dekker, line 217: nit: could you reformat the imports?

pkg/render/manager.go
```diff
@@ -302,20 +303,25 @@ func (c *managerComponent) managerVolumes() []v1.Volume {
 		v = append(v,
 			v1.Volume{
 				// We only want to mount the cert, not the private key to es-proxy to establish a connection with voltron.
-				Name: ManagerTLSSecretCertName,
+				Name: ManagerInternalTLSSecretName,
 				VolumeSource: v1.VolumeSource{
 					Secret: &v1.SecretVolumeSource{
```
rene-dekker, line 308: Does this mean you are also mounting a private key to es-proxy now, unlike what the comment says?

pkg/render/manager_test.go
```diff
@@ -90,10 +91,31 @@ var _ = Describe("Tigera Secure Manager rendering tests", func() {
...
+		var voltron = deployment.Spec.Template.Spec.Containers[2]
+		Expect(len(voltron.VolumeMounts)).To(Equal(3))
+		Expect(voltron.VolumeMounts[0].Name).To(Equal(render.ManagerTLSSecretName))
+		Expect(voltron.VolumeMounts[0].MountPath).To(Equal("/certs/https"))
+		Expect(voltron.VolumeMounts[1].Name).To(Equal(render.KibanaPublicCertSecret))
+		Expect(voltron.VolumeMounts[1].MountPath).To(Equal("/certs/kibana"))
+		Expect(voltron.VolumeMounts[2].Name).To(Equal(render.ComplianceServerCertSecret))
+		Expect(voltron.VolumeMounts[2].MountPath).To(Equal("/certs/compliance"))
+
+		//Expect 3 volumes mapped to 3 secrets
```
rene-dekker, line 105: Looks like 4 volumes mapped to 3 secrets

rene-dekker, line 105: Looks like 4.

pkg/render/manager.go
asincu: Talked offline: We can declare the volume twice: a declaration with cert and key for Voltron and a declaration with only the cert for es proxy.

pkg/render/manager.go
```diff
@@ -195,6 +204,9 @@ func (c *managerComponent) managerDeployment() *appsv1.Deployment {
 	}
 	if c.management {
 		annotations[voltronTunnelHashAnnotation] = AnnotationHash(c.tunnelSecrets[0].Data)
+		if len(c.tunnelSecrets) == 2 {
```
ozdanborne, line 207: This seems kindof flimsy. Should we really be grouping the internalSecret in with `c.tunnelSecrets`? Also wondering why `c.tunnelSecrets` is a slice in the first place. Should we just have `c.tunnelSecret` and `c.internalSecret`?

pkg/render/manager.go
```diff
@@ -315,6 +317,26 @@ func (c *managerComponent) managerVolumes() []v1.Volume {
...
+
+		// Append volume for tunnel certificate
+		v = append(v, v1.Volume{
+			Name: VoltronTunnelSecretName,
+			VolumeSource: v1.VolumeSource{
+				Secret: &v1.SecretVolumeSource{
+					SecretName: VoltronTunnelSecretName,
+				},
+			},
+		},
```
ozdanborne, line 339: nit: just include this in the above `append`
```suggestion
			},
			v1.Volume{
				Name: VoltronTunnelSecretName,
				VolumeSource: v1.VolumeSource{
					Secret: &v1.SecretVolumeSource{
						SecretName: VoltronTunnelSecretName,
					},
				},
			},
```

pkg/render/manager.go
asincu: I think that would be better. I do not really like the fact that all of these are slices and we only populate the first element and use it to generate hashes.

pkg/render/manager.go
asincu: Sounds good :)

pkg/render/manager.go
ozdanborne: Idk I see you're mostly just following the pattern here. Seems odd that we love setting up slices of secrets when we usually expect the length to just be 1 or 0. I'm open to keeping it as is...

ozdanborne:
You mentioned three certs in the description:
- manager-tls
- manager-internal-tls
- tigera-management-cluster-connection

But it seems like your PR removed `manager-tls` and replaced it with `manager-internal-tls`?  Am I misreading that?

pkg/render/render.go
```diff
@@ -109,14 +109,15 @@ func Calico(
...
 
-	if managerTLSSecret == nil && cr.Spec.Variant == operator.TigeraSecureEnterprise && cr.Spec.ClusterManagementType == operator.ClusterManagementTypeManagement {
-		// Generate CA and TLS certificate for tigera-manager
-		log.Info("Creating secret for manager")
+	if managerInternalTLSSecret == nil && cr.Spec.Variant == operator.TigeraSecureEnterprise && cr.Spec.ClusterManagementType == operator.ClusterManagementTypeManagement {
+		// Generate CA and TLS certificate for tigera-manager for internal traffic within the K8s cluster
+		// The certificate will be issued for ManagerServiceDNS and localhost
+		log.Info("Creating secret for internal manager credentials")
 		var err error
-		managerTLSSecret, err = CreateOperatorTLSSecret(nil,
```
ozdanborne, line 116: don't we still need to create this somewhere?

asincu: @ozdanborne `manager-tls` will either be created by the operator in https://github.com/tigera/operator/pull/606/files#diff-29e68fd54befc00ec93a59c7359836d0R90 or a client is going to provided it.

asincu:
> You mentioned three certs in the description:
> 
> * manager-tls
> * manager-internal-tls
> * tigera-management-cluster-connection
> 
> But it seems like your PR removed `manager-tls` and replaced it with `manager-internal-tls`? Am I misreading that?

`manager-internal-tls` will be created by kube-controller and distributed all components that need it: compliance, kube-controllers and es -proxy. The certficate + private key will be loaded into Voltron.
Previously, the same cert was used for both internal communication and browser - Tigera UI interaction.

pkg/render/manager.go
```diff
@@ -77,10 +81,12 @@ func Manager(
 	oidcConfig *corev1.ConfigMap,
 	management bool,
 	tunnelSecret *corev1.Secret,
+	internalTrafficSecret *corev1.Secret,
 ) (Component, error) {
 	tlsSecrets := []*corev1.Secret{}
+	tlsAnnotations := make(map[string]string)
 
-	if tlsKeyPair == nil && !management {
```
stevegaossou, line 83: Hey @asincu do you remember why there was a check for whether or not it was a management cluster? 

Just want to understand what scenario it was accounting for and why it's no longer needed.

pkg/render/render.go
stevegaossou: If this section now generates `internal-manager-tls` instead, it doesn't really need the `localhost` domain in the certificate (since it is intended to be called only from components that have access to the service DNS name). 

It probably doesn't make it less secure, but it's just not really needed for `internal-manager-tls`. The `localhost` was useful for `manager-tls` when a user tries to access Manager UI through port-forwarding https://localhost:9443.

pkg/controller/manager/manager_controller.go
```diff
@@ -331,6 +326,21 @@ func (r *ReconcileManager) Reconcile(request reconcile.Request) (reconcile.Resul
 				return reconcile.Result{}, nil
 			}
 		}
```
stevegaossou, line 328: This is a somewhat sidetrack issue, but do we want the above section (lines 323 - 333)? 

Since we don't support users bringing their own secret for the tunnel certificate, we don't need logic that involves checking or trying to move the secret over into `tigera-manager` namespace. Better to remove it to close the loop.

pkg/render/manager.go
asincu: Yup, when the cert was shared, it was created by the kube-controller renderer.

pkg/controller/manager/manager_controller.go
asincu: Let's do it on a separate PR.

PR #611, opened 2020-06-08
Make metrics port configurable

### LINKED ISSUE #515: metrics port hardcoded to 8383 in tigera-operator conflicting with CNV
## Expected Behavior

Metrics port in tigera-operator should be configurable, and default value should be different than 8383 which is used by nmstate-handler pods from CNV (Container Native Virtualization in RedHat Openshift 4.x. Instead it is hardcoded to 8383 here: https://github.com/tigera/operator/blob/master/pkg/daemon/daemon.go#L35
 

## Current Behavior

Currently it is not possible to properly deploy CNV (Container Native Virtualization) on RedHat Openshift 4.x with Calico and tigera-operator. Both tigera-operator and one of nmstate-handler pods (part of CNC) run with host network and try to bind to port 8383. Since tigera-operator is deployed first, nmstate-handler keeps crashing and CNV deployment cannot finish properly.

## Possible Solution

1. make metrics port configurable, and
2. change default port 8383 to something else

1. deploy RedHat Openshift 4.3.12 with Calico as specified in RedHat documentation and the following page: https://docs.projectcalico.org/getting-started/openshift/installation
2. deploy CNV according to RedHat documentation: https://docs.openshift.com/container-platform/4.3/cnv/cnv_install/installing-container-native-virtualization.html
3. list pods in openshift-cnv namespace and look for crashing nmstate-handler running on the same node as tigera-operator
4. scale down tigera-operator to 0 replicas to stop tigera-operator pod
5. restart failed nmstate-handler pod - it should be able to run successfully
6. inspect status of hco/kubevirt-hyperconverged in openshift-cnv namespace - the last status condition should state "Reconcile completed successfully"

## Your Environment

* Operating System and version: Redhat OpenShift 4.3.12 with Calico deployed using tigera-operator v1.3.3

ozdanborne:
## Description

Fixes #515

Replaces #604 

Operator currently serves prometheus metrics on port 8383. This is not configurable. This is problematic because operator is run in hostNetworked mode, and will conflict with any other service running on port 8383 on the node. Several users have reported this happening due to a service in openshift that also uses 8383.

It seems like this code hasn't been touched since the original boiler plate was checked in during the initial commit. Given that, I've taken some liberties to adjust the default behavior from here on out.

This PR:
- Exposes host and port via the `METRICS_HOST` and `METRICS_PORT` environment variables. If both are empty strings, metrics will be shut off.
- Changes default port to 8484. This will be used if only `METRICS_HOST` is specified.

I did some validation testing:
- :heavy_check_mark: running without setting `METRICS_HOST` and `METRICS_PORT` (or setting them to a blank string) results in metrics not being run
- :heavy_check_mark: running with `METRICS_HOST=localhost` runs metrics on 8383
- :heavy_check_mark: running with `METRICS_PORT=8384` runs metrics on 8484

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [x] Appropriate labels:

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -31,7 +31,7 @@ import (
 )
 
 var (
-	defaultMetricsPort int32 = 8383
+	defaultMetricsPort int32 = 8484
 	namespace                = ""
 )
 
@@ -115,7 +115,7 @@ func metricsAddr() string {
 		// the controller-runtime accepts '0' to denote that metrics should be disabled.
 		return "0"
 	}
-	// if just a host is specified, listen on port 8383 of that host.
+	// if just a host is specified, listen on port 8484 of that host.
 	if metricsHost != "" && metricsPort == "" {
 		// the controller-runtime will choose a random port if none is specified.
 		// so use the defaultMetricsPort in that case.
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -16,7 +16,6 @@ package daemon
 
 import (
 	"context"
-	"fmt"
 	"os"
 
 	"github.com/operator-framework/operator-sdk/pkg/leader"
@@ -31,9 +30,7 @@ import (
 )
 
 var (
-	metricsHost       = "0.0.0.0"
-	metricsPort int32 = 8383
-	namespace         = ""
+	namespace = ""
 )
 
 var log = logf.Log.WithName("daemon")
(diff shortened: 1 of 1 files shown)


pkg/daemon/daemon.go
```diff
@@ -46,6 +43,13 @@ func Main() {
 		os.Exit(1)
 	}
 
+	metricsAddr := os.Getenv("METRICS_BIND_ADDRESS")
```
caseydavenport, line 46: You might want this to be two separate variables - METRICS_HOST and METRICS_PORT.

Main reason being that users can then use the Kubernetes downward API to set the HOST portion to the node's IP address based on where the pod lands, without needing to bind to all addresses.

PR #613, opened 2020-06-09
Add node awareness to Elasticsearch cluster

Brian-McM:
## Description
This commit allows you to set the "nodeSets" field in the LogStorage CR to define node awareness setting for the Elasticsearch nodes. This "node awareness" adds node affinities and cluster awareness setting (Elasticsearch configuration) so that Elasticsearch nodes distributed across k8s nodes with certain attributes, i.e. availability zone, and the Elasticsearch cluster will assign replicas to Elasticsearch nodes with different attributes than the primary shard.

### DIFF
--- a/pkg/apis/operator/v1/logstorage_types.go
+++ b/pkg/apis/operator/v1/logstorage_types.go
@@ -73,9 +73,33 @@ type Nodes struct {
 	// Count defines the number of nodes in the Elasticsearch cluster.
 	Count int64 `json:"count,omitempty"`
 
+	// NodeSets defines configuration specific to each Elasticsearch Node Set
+	// +optional
+	NodeSets []NodeSet `json:"nodeSets,omitempty"`
+
 	// ResourceRequirements defines the resource limits and requirements for the Elasticsearch cluster.
 	// +optional
 	ResourceRequirements *corev1.ResourceRequirements `json:"resourceRequirements,omitempty"`
+}
+
+// NodeSets defines configuration specific to each Elasticsearch Node Set
+type NodeSet struct {
+	// SelectionAttributes defines K8s node attributes a NodeSet should use when setting the Node Affinity selectors and
+	// Elasticsearch cluster awareness attributes for the Elasticsearch nodes. The list of SelectionAttributes are used
+	// to define Node Affinities and set the node awareness configuration in the running Elasticsearch instance.
+	SelectionAttributes []NodeSetSelectionAttribute `json:"selectionAttributes,omitempty"`
+}
+
+// NodeSetSelectionAttribute defines a K8s node "attribute" the Elasticsearch nodes should be aware of. The "Name" and "Value"
+// are used together to set the "awareness" attributes in Elasticsearch, while the "NodeLabel" and "Value" are used together
+// to define Node Affinity for the Pods created for the Elasticsearch nodes.
+type NodeSetSelectionAttribute struct {
+	// +required
+	Name string `json:"name"`
+	// +required
+	NodeLabel string `json:"nodeLabel"`
+	// +required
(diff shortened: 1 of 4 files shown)


pkg/apis/operator/v1/logstorage_types.go
```diff
@@ -73,11 +73,32 @@ type Nodes struct {
...
+	// Elasticsearch cluster awareness attributes for the Elasticsearch nodes. The list of SelectionAttributes are used
+	// to define Node Affinities and set the node awareness configuration in the running Elasticsearch instance.
+	SelectionAttributes []NodeSetSelectionAttribute `json:"selectionAttributes,omitempty"`
+}
+
+// NodeSetSelectionAttribute defines a K8s node "attribute" the Elasticsearch nodes should be aware of. The "Name" and "Value"
+// are used together to set the "awareness" attributes in Elasticsearch, while the "NodeLabel" and "Value" are used together
+// to define Node Affinity for the Pods created for the Elasticsearch nodes.
+type NodeSetSelectionAttribute struct {
+	Name      string `json:"name, omitempty"`
```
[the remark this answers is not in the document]
Brian-McM: Ya true, it's not valid to omit them

pkg/render/logstorage_test.go
```diff
@@ -439,6 +440,379 @@ var _ = Describe("Elasticsearch rendering tests", func() {
...
+						[]*corev1.Secret{
+							{ObjectMeta: metav1.ObjectMeta{Name: render.TigeraKibanaCertSecret, Namespace: render.OperatorNamespace()}},
+							{ObjectMeta: metav1.ObjectMeta{Name: render.TigeraKibanaCertSecret, Namespace: render.KibanaNamespace}},
+						}, true,
+						[]*corev1.Secret{
+							{ObjectMeta: metav1.ObjectMeta{Name: "tigera-pull-secret"}},
+						}, operator.ProviderNone, nil, nil, nil, "cluster.local", true)
+
+					createResources, _ := component.Objects()
+					nodeSets := createResources[15].(*esv1.Elasticsearch).Spec.NodeSets
```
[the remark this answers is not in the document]
Brian-McM: Will do, I just copied what was there for the other tests.

pkg/render/logstorage.go
```diff
@@ -525,22 +522,112 @@ func (es elasticsearchComponent) elasticsearchCluster() *esv1.Elasticsearch {
...
+	var nodeSets []esv1.NodeSet
+	if nodeConfig.NodeSets == nil {
+		nodeSet := es.nodeSetTemplate()
+		nodeSet.Name = nodeSetName(pvcTemplate)
+		nodeSet.Count = int32(nodeConfig.Count)
+		nodeSet.PodTemplate = es.podTemplate()
+
+		nodeSets = append(nodeSets, nodeSet)
+	} else {
+		baseNumNodes := nodeConfig.Count / int64(len(nodeConfig.NodeSets))
```
tmjd, line 546: What about the case where NodeSets is not nil and Count is zero/unset?
Won't we end up with no nodeSets? Is that valid?

Brian-McM: Good point! I'll make sure the "if" is hit in that case.

Brian-McM: I actually think this will panic, can't divide by 0. Rookie mistake.

tmjd: I'm glad you noticed that but I was meaning if `nodeConfig.Count` was 0.

Brian-McM: If that's 0 in either case there will be no nodes set. We should probably default or warn about this, I'll do this in a separate PR.

PR #620, opened 2020-06-12
Tunnel secret will be created by the API server

asincu:
## Description

The tunnel secret will be created by the API server, instead of the manager. This change is needed in order to share the CA with the API server, that will be solely responsible for generating managed cluster resources. At creation, a client cert and key will be provided as a one time event. The cert and private key are signed by the CA from secret `tigera-management-cluster-connection`

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/apiserver/apiserver_controller.go
+++ b/pkg/controller/apiserver/apiserver_controller.go
@@ -17,6 +17,7 @@ package apiserver
 import (
 	"context"
 	"fmt"
+	v1 "k8s.io/api/core/v1"
 	"time"
 
 	operatorv1 "github.com/tigera/operator/pkg/apis/operator/v1"
@@ -164,6 +165,21 @@ func (r *ReconcileAPIServer) Reconcile(request reconcile.Request) (reconcile.Res
 		return reconcile.Result{}, err
 	}
 
+	var isManagement = network.Spec.ClusterManagementType == operatorv1.ClusterManagementTypeManagement
+	var tunnelCASecret *v1.Secret
+	if isManagement {
+		tunnelCASecret, err = utils.ValidateCertPair(r.client,
+			render.VoltronTunnelSecretName,
+			render.VoltronTunnelSecretKeyName,
+			render.VoltronTunnelSecretCertName,
+		)
+		if err != nil {
+			log.Error(err, "Invalid TLS Cert")
+			r.status.SetDegraded("Error validating TLS certificate", err.Error())
+			return reconcile.Result{}, err
+		}
+	}
+
 	pullSecrets, err := utils.GetNetworkingPullSecrets(network, r.client)
 	if err != nil {
 		log.Error(err, "Error retrieving Pull secrets")
@@ -185,7 +201,8 @@ func (r *ReconcileAPIServer) Reconcile(request reconcile.Request) (reconcile.Res
 
 	// Render the desired objects from the CRD and create or update them.
 	reqLogger.V(3).Info("rendering components")
-	component, err := render.APIServer(network, aci, tlsSecret, pullSecrets, r.provider == operatorv1.ProviderOpenShift)
+	component, err := render.APIServer(network, aci, tlsSecret, pullSecrets, r.provider == operatorv1.ProviderOpenShift,
+		tunnelCASecret)
 	if err != nil {
 		log.Error(err, "Error rendering APIServer")
 		r.status.SetDegraded("Error rendering APIServer", err.Error())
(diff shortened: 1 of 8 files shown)


pkg/controller/manager/manager_controller.go
```diff
@@ -315,13 +315,12 @@ func (r *ReconcileManager) Reconcile(request reconcile.Request) (reconcile.Resul
...
 
-		// If clusterType is management and the customer brings its own cert, copy it over to the manager ns.
+		// We expect that the secret that holds the certificates for tunnel certificate generation
+		// is already created by the Api Server
 		tunnelSecret = &corev1.Secret{}
 		err := r.client.Get(ctx, client.ObjectKey{Name: render.VoltronTunnelSecretName, Namespace: render.OperatorNamespace()}, tunnelSecret)
 		if err != nil {
 			if errors.IsNotFound(err) {
-				tunnelSecret = nil
-			} else {
```
tmjd, line 324: I don't think there is any need for the `IsNotFound` check. I believe any error for the secret should result in SetDegraded here?

pkg/render/apiserver.go
```diff
@@ -41,8 +41,10 @@ const (
 
 var apiServiceHostname = apiServiceName + "." + APIServerNamespace + ".svc"
 
-func APIServer(installation *operator.Installation, aci *operatorv1beta1.AmazonCloudIntegration, tlsKeyPair *corev1.Secret, pullSecrets []*corev1.Secret, openshift bool) (Component, error) {
+func APIServer(installation *operator.Installation, aci *operatorv1beta1.AmazonCloudIntegration, tlsKeyPair *corev1.Secret, pullSecrets []*corev1.Secret, openshift bool, isManagement bool, tunnelCASecret *corev1.Secret) (Component, error) {
```
tmjd, line 44: Since the installation is being passed in I don't think we need the isManagement flag. (I'd be on board with dropping the openshift flag too)

pkg/render/apiserver_test.go
```diff
@@ -326,6 +328,135 @@ var _ = Describe("API server rendering tests", func() {
...
+
+		i := 0
+		for _, expectedRes := range expectedResources {
+			ExpectResource(resources[i], expectedRes.name, expectedRes.ns, expectedRes.group, expectedRes.version, expectedRes.kind)
+			i++
+		}
+
+		By("Validating the newly created tunnel secret")
+		// Use the x509 package to validate that the cert was signed with the privatekey
+		validateTunnelSecret(resources[13].(*corev1.Secret))
```
tmjd, line 395: Should use the GetResource (pkg/render/render_utils_test.go) function here instead of a direct index into the resources slice. It'll make sure we're checking the right resource even if the resources change. (of course we'll have to adjust the code above though)

pkg/render/manager.go
```diff
@@ -105,12 +107,6 @@ func Manager(
 	tlsAnnotations[tlsSecretHashAnnotation] = AnnotationHash(tlsKeyPair.Data)
 
 	if management {
-		// If there is no secret create one and add it to the operator namespace.
-		if tunnelSecret == nil {
-			tunnelSecret = voltronTunnelSecret()
-			tlsSecrets = append(tlsSecrets, tunnelSecret)
```
tmjd, line 111: Is this line not needed too?

pkg/render/voltron_secret.go
```diff
@@ -35,6 +38,23 @@ const (
...
+func voltronTunnelSecret() *corev1.Secret {
+	key, cert := ceateSelfSignedVoltronSecret()
+	return &corev1.Secret{
+		TypeMeta: metav1.TypeMeta{Kind: "Secret", APIVersion: "v1"},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      VoltronTunnelSecretName,
+			Namespace: OperatorNamespace(),
+		},
+		Data: map[string][]byte{
+			"cert": []byte(cert),
```
tmjd, line 52: You've got constants for these key names, you should use them :grinning:

pkg/render/manager.go
asincu: This copied the secret from the operator namespace. Since we are not generating it, we should only render the secret that is copied inside the tigera-manager namespace.

pkg/render/apiserver_test.go
asincu: :100: Thanks for the tip!

pkg/controller/apiserver/apiserver_controller.go
```diff
@@ -164,6 +165,21 @@ func (r *ReconcileAPIServer) Reconcile(request reconcile.Request) (reconcile.Res
 		return reconcile.Result{}, err
 	}
 
+	var isManagement = network.Spec.ClusterManagementType == operatorv1.ClusterManagementTypeManagement
+	var tunnelCASecret *v1.Secret
+	if isManagement {
+		tunnelCASecret, err = utils.ValidateCertPair(r.client,
```
tmjd, line 171: I think we need a watch added for this secret.

asincu: @tmjd Added: https://github.com/tigera/operator/pull/626

PR #621, opened 2020-06-12
Update to operator-sdk v0.18.1, and gen CRDs

caseydavenport:
## Description

Switches the CRDs to the v1 API

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

```release-note
Update CRDs to v1
```


deploy/crds/operator.tigera.io_amazoncloudintegrations_crd.yaml
```diff
@@ -0,0 +1,90 @@
+apiVersion: apiextensions.k8s.io/v1
```
caseydavenport, line 1: It's unfortunate that it uses a different name for the CRD. Let me see if I can trick git into treating this as a "moved" file.

pkg/apis/operator/v1/intrusiondetection_types.go
```diff
@@ -42,6 +42,7 @@ type IntrusionDetectionStatus struct {
 // of this resource is supported. It must be named "tigera-secure".
 // +k8s:openapi-gen=true
 // +kubebuilder:subresource:status
+// +kubebuilder:resource:scope=Cluster
```
tmjd, line 45: Is the `+genclient:nonNamespaced` still needed?

Makefile
```diff
@@ -417,10 +417,9 @@ endif
 ###############################################################################
 # Utilities
 ###############################################################################
-## Generating code after API changes
+## Generating code after API changes. Requires operator-sdk v0.18.1.
 gen-files:
-	operator-sdk generate k8s
-	operator-sdk generate openapi
```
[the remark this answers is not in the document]
caseydavenport: This command no longer exists in the latest operator-sdk. I think that this is the command that actually used to generate crds (there wasn't a generate crds subcommand before)

pkg/apis/operator/v1/intrusiondetection_types.go
caseydavenport: The `genclient` stuff is all for building k8s golang clients.

The `kubebuilder` tags are for CRD yaml files. 

So, I think they are both used for different things.

Makefile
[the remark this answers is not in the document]
tmjd: There is still a k8s generate that is needed to `Generates Kubernetes code for custom resource`, without that we won't see our deepcopy code generated.

PR #622, opened 2020-06-12
make amazon cloud CRD optional

ozdanborne:
## Description

The Operator will deadlock if the new v1beta1 amazoncloudintegration CRD is not created. We encountered this briefly when merging operator code in conjunction with docs code. We now always include the CRD on docs. 

However, down the road, it may be desirable to not force installation of this CRD when a user isn't running on AWS nor have any plans to do so.

This PR copies the mechanism already established by the core controller to not launch the Enterprise controllers if we're not running Enterprise. In this case, we skip launching the AmazonCloudIntegration controller if we're OS or not using EKS networking. 

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/amazoncloudintegration/amazoncloudintegration_controller.go
+++ b/pkg/controller/amazoncloudintegration/amazoncloudintegration_controller.go
@@ -46,7 +46,7 @@ var log = logf.Log.WithName("controller_amazoncloudintegration")
 // Add creates a new AmazonCloudIntegration Controller and adds it to the Manager. The Manager will set fields on the Controller
 // and Start it when the Manager is Started.
 func Add(mgr manager.Manager, opts options.AddOptions) error {
-	if !opts.EnableEnterpriseControllers {
+	if !opts.AmazonCRDExists {
 		// No need to start this controller.
 		return nil
 	}
--- a/pkg/controller/amazoncloudintegration/amazoncloudintegration_controller.go
+++ b/pkg/controller/amazoncloudintegration/amazoncloudintegration_controller.go
@@ -30,6 +30,7 @@ import (
 	operatorv1 "github.com/tigera/operator/pkg/apis/operator/v1"
 	operatorv1beta1 "github.com/tigera/operator/pkg/apis/operator/v1beta1"
 	"github.com/tigera/operator/pkg/controller/installation"
+	"github.com/tigera/operator/pkg/controller/options"
 	"github.com/tigera/operator/pkg/controller/status"
 	"github.com/tigera/operator/pkg/controller/utils"
 	"github.com/tigera/operator/pkg/render"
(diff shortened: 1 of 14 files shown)


pkg/controller/apiserver/apiserver_controller.go
```diff
@@ -171,13 +173,16 @@ func (r *ReconcileAPIServer) Reconcile(request reconcile.Request) (reconcile.Res
...
 
-	aci, err := utils.GetAmazonCloudIntegration(ctx, r.client)
-	if errors.IsNotFound(err) {
-		aci = nil
-	} else if err != nil {
-		log.Error(err, "Error reading AmazonCloudIntegration")
-		r.status.SetDegraded("Error reading AmazonCloudIntegration", err.Error())
-		return reconcile.Result{}, err
+	var aci *operatorv1beta1.AmazonCloudIntegration
+	if r.provider == operatorv1.ProviderEKS {
```
caseydavenport, line 177: IMO this is the wrong way around.

I think it should be an error for the users to create an AmazonCloudIntegration resource if the provider is not EKS. IIUC, the current code will just never query the "ACI" unless the provider is EKS.

I think we should query both, and perform cross validation with an error message if an invalid configuration is provided. That way, we can alert the user to incorrect configuration rather than silently ignore it.

caseydavenport: Hm, I suppose we do this validation here right now: https://github.com/tigera/operator/pull/622/files#diff-f99c8b3f1d4c6a1c8cc66cb40b40e023R162

pkg/controller/amazoncloudintegration/amazoncloudintegration_controller.go
```diff
@@ -45,7 +45,7 @@ var log = logf.Log.WithName("controller_amazoncloudintegration")
 // Add creates a new AmazonCloudIntegration Controller and adds it to the Manager. The Manager will set fields on the Controller
 // and Start it when the Manager is Started.
 func Add(mgr manager.Manager, provider operatorv1.Provider, tsee bool) error {
-	if !tsee {
+	if !tsee || provider != operatorv1.ProviderEKS {
```
caseydavenport, line 48: I don't think this is quite the assertion that I was imagining.

I think this is saying that we still _always_ need to install this CRD on AWS. What I was imagining is that we never require the CRD, but if the CRD exists then we run this controller.

ozdanborne: I initially looked into implementing it the way you described, but it was a slightly more involved change because you need to update the function call on every controller's Add method. Which I can totally do, but...

>I think this is saying that we still always need to install this CRD on AWS

You're exactly correct here, but I thought that was a pretty fair UX / requirement of the user. After all, the original use case that prompted this PR was "i don't want to apply amazon-specific crd's if i'm not on amazon", which both approaches still accomplish.

Also note that what is there in the PR is in line with the other tech debt that we have accumulated in the preview version of this feature, in that we've already overloaded the EKS provider to mean something it is not: indicating that the AWS CNI plugin is in use.

pkg/controller/installation/core_controller.go
```diff
@@ -132,7 +132,7 @@ func add(mgr manager.Manager, r *ReconcileInstallation) error {
 	}
 
 	// Only watch the AmazonCloudIntegration if the tsee API is available
-	if r.enterpriseCRDsExist {
+	if r.autoDetectedProvider == operator.ProviderEKS {
```
caseydavenport, line 135: Shouldn't this be both conditions, with an &&?

PR #625, opened 2020-06-16
Add clusterrole to components that need authentication reviews

rene-dekker:
Add clusterrole to components that need authentication reviews for cross cluster requests

The authentication reviews api is an api that exchanges an auth header for user information. Necessary for:
- kube-controllers to view and modify secrets in managed cluster
- compliance to fetch report data from managed cluster

Token reviews are no longer used.

### DIFF
--- a/pkg/render/compliance.go
+++ b/pkg/render/compliance.go
@@ -472,6 +472,8 @@ func (c *complianceComponent) complianceServerClusterRole() *rbacv1.ClusterRole 
 
 	if c.installation.Spec.ClusterManagementType == operatorv1.ClusterManagementTypeManagement {
 		// For cross-cluster requests, an authentication review will be done for authenticating the compliance-server.
+		// Requests on behalf of the compliance-server will be sent to Voltron, where an authentication review will take
+		// place with its bearer token.
 		clusterRole.Rules = append(clusterRole.Rules, rbacv1.PolicyRule{
 			APIGroups: []string{"projectcalico.org"},
 			Resources: []string{"authenticationreviews"},
--- a/pkg/render/compliance.go
+++ b/pkg/render/compliance.go
@@ -382,7 +382,6 @@ func (c *complianceComponent) complianceReporterPodTemplate() *corev1.PodTemplat
 		privileged = true
 	}
 
-
 	envVars := []corev1.EnvVar{
 		{Name: "LOG_LEVEL", Value: "warning"},
 		{Name: "TIGERA_COMPLIANCE_JOB_NAMESPACE", Value: ComplianceNamespace},
@@ -454,7 +453,7 @@ func (c *complianceComponent) complianceServerServiceAccount() *corev1.ServiceAc
 }
 
 func (c *complianceComponent) complianceServerClusterRole() *rbacv1.ClusterRole {
-	return &rbacv1.ClusterRole{
+	clusterRole := &rbacv1.ClusterRole{
 		TypeMeta:   metav1.TypeMeta{Kind: "ClusterRole", APIVersion: "rbac.authorization.k8s.io/v1"},
 		ObjectMeta: metav1.ObjectMeta{Name: "tigera-compliance-server"},
 		Rules: []rbacv1.PolicyRule{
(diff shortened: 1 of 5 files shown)


pkg/render/manager.go
```diff
@@ -578,12 +578,6 @@ func managerClusterRole(impersonationOnly bool) *rbacv1.ClusterRole {
 
 	if !impersonationOnly {
 		cr.Rules = append(cr.Rules,
-			rbacv1.PolicyRule{
-
-				APIGroups: []string{"authentication.k8s.io"},
-				Resources: []string{"tokenreviews"},
-				Verbs:     []string{"create"},
-			},
```
stevegaossou, line 586: Does Voltron not need at least one of `tokenreviews` or `authenticationreviews` for its AuthN checking?

pkg/render/kube-controllers.go
```diff
@@ -156,6 +156,15 @@ func (c *kubeControllersComponent) controllersRole() *rbacv1.ClusterRole {
...
 		role.Rules = append(role.Rules, extraRules...)
+
+		if c.cr.Spec.ClusterManagementType == operator.ClusterManagementTypeManagement {
+			// For cross-cluster requests an authentication review will be done for authenticating the kube-controllers.
+			role.Rules = append(role.Rules, rbacv1.PolicyRule{
+				APIGroups: []string{"projectcalico.org"},
+				Resources: []string{"authenticationreviews"},
+				Verbs:     []string{"create"},
+			})
+		}
```
stevegaossou, line 167: Would be good to add some test coverage for these config changes.

pkg/render/manager.go
rene-dekker: No, the Authorization headers that it receives are sent to /apis/projectcalico.org/authenticationreviews and for this reason, we can remove the permission from here, but we need to add it to the sa's that do cross cluster requests.

pkg/render/kube-controllers.go
rene-dekker: Added

pkg/render/compliance.go
```diff
@@ -463,18 +462,23 @@ func (c *complianceComponent) complianceServerClusterRole() *rbacv1.ClusterRole
...
-			},
 			{
 				APIGroups: []string{"authorization.k8s.io"},
 				Resources: []string{"subjectaccessreviews"},
 				Verbs:     []string{"create"},
 			},
 		},
 	}
+
+	if c.installation.Spec.ClusterManagementType == operatorv1.ClusterManagementTypeManagement {
```
doublek, line 473: Will this also work for standalone clusters?

pkg/render/manager.go
doublek, line 586: Doesn't es-proxy need a Policyrule for `authenticationreviews`?

pkg/render/compliance.go
rene-dekker: Yes, it should. Let me give an example:

Standalone
- Jane uses UI and goes to compliance page
- Request is made with her header
- Jane has a role that includes authenticationReviews/create
- Compliance server sends this header to api-server in the form of an authenticationReviews/create call

Now Management: Jane wants to access cluster X
- Same 4 steps
- Now compliance server needs data from other cluster
- A request on behalf of the compliance-server is made to voltron, to be tunneled to X
- Voltron will then use the header that belongs to the compliance-server and make the authenticationReviews/create call

rene-dekker: You might wonder why es-proxy does not need this permission then. (tigera-manager SA)
I did too.
It turns out that the rbac checks done in the other cluster are on behalf of Jane again. A special config is made for Jane in the rbac code, and so it is Jane who will be impersonated.

pkg/render/manager.go
rene-dekker: Please see my comment for the other one above. I tried it out and I think it is not necessary.

pkg/render/compliance.go
doublek: Makes sense. Thanks for explaining the reason.

pkg/render/compliance.go
```diff
@@ -463,18 +462,23 @@ func (c *complianceComponent) complianceServerClusterRole() *rbacv1.ClusterRole
...
+		// For cross-cluster requests, an authentication review will be done for authenticating the compliance-server.
```
doublek, line 474: Can you add a comment here as to why you don't need this for standalone clusters.

pkg/render/compliance_test.go
```diff
@@ -174,6 +182,14 @@ var _ = Describe("compliance rendering tests", func() {
 			Expect(dpComplianceServer.Spec.Template.Spec.Volumes[1].Secret.SecretName).To(Equal(render.ManagerInternalTLSSecretName))
 			Expect(dpComplianceServer.Spec.Template.Spec.Volumes[2].Name).To(Equal("elastic-ca-cert-volume"))
 			Expect(dpComplianceServer.Spec.Template.Spec.Volumes[2].Secret.SecretName).To(Equal(render.ElasticsearchPublicCertSecret))
+
+			clusterRole := resources[28].(*rbacv1.ClusterRole)
```
tmjd, line 186: We should use GetResource from `pkg/render/render_utils_test.go` instead of indexing in directly. It'll make the test more resilient to changing the resources.

pkg/render/kube-controllers_test.go
```diff
@@ -97,6 +100,10 @@ var _ = Describe("kube-controllers rendering tests", func() {
 		ds := resources[3].(*apps.Deployment)
 
 		Expect(ds.Spec.Template.Spec.Containers[0].Image).To(Equal("test-reg/tigera/kube-controllers:" + components.ComponentTigeraKubeControllers.Version))
+
+		clusterRole := resources[1].(*rbacv1.ClusterRole)
```
tmjd, line 104: Same comments in this file for getting specific resources.

pkg/render/manager.go
rene-dekker: There is a thread on this PR with Karthik that further clarifies the flow ^

pkg/render/compliance_test.go
rene-dekker: Thanks for making me aware of this.

PR #637, opened 2020-06-17
Add clusterrole rule for authentication reviews for the tigera-manager

rene-dekker:
Add clusterrole rule for authentication reviews for the tigera-manager

Related to the implementation of authentication reviews as a replacement for token reviews.

### DIFF
--- a/pkg/render/guardian.go
+++ b/pkg/render/guardian.go
@@ -80,7 +80,7 @@ func (c *GuardianComponent) Objects() ([]runtime.Object, []runtime.Object) {
 		// Add tigera-manager service account for impersonation
 		createNamespace(ManagerNamespace, c.openshift),
 		managerServiceAccount(),
-		managerClusterRole(true),
+		managerClusterRole(c.installation.Spec.ClusterManagementType),
 		managerClusterRoleBinding(),
 	)
 
--- a/pkg/render/manager_test.go
+++ b/pkg/render/manager_test.go
@@ -22,6 +22,7 @@ import (
 	appsv1 "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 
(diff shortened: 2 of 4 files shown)


pkg/render/guardian.go
```diff
@@ -80,7 +80,7 @@ func (c *GuardianComponent) Objects() ([]runtime.Object, []runtime.Object) {
 		// Add tigera-manager service account for impersonation
 		createNamespace(ManagerNamespace, c.openshift),
 		managerServiceAccount(),
-		managerClusterRole(true),
+		managerClusterRole(operatorv1.ClusterManagementTypeManaged),
```
asincu, line 83: Could we pass `installation.Spec.ClusterManagementType` ?

pkg/render/manager_test.go
```diff
@@ -126,6 +127,11 @@ var _ = Describe("Tigera Secure Manager rendering tests", func() {
 		Expect(d.Spec.Template.Spec.Containers[0].Name).To(Equal("tigera-manager"))
 		Expect(d.Spec.Template.Spec.Containers[0].Env[8].Name).To(Equal("CNX_POLICY_RECOMMENDATION_SUPPORT"))
 		Expect(d.Spec.Template.Spec.Containers[0].Env[8].Value).To(Equal("true"))
+
+		clusterRole := GetResource(resources, render.ManagerClusterRole, "", "rbac.authorization.k8s.io", "v1", "ClusterRole").(*rbacv1.ClusterRole)
+		Expect(len(clusterRole.Rules)).To(Equal(2))
```
tmjd, line 132: Please switch this to and the 2 expects below to just one that checks the whole rules has the proper elements.
`Expect(clusterRole.Rules).To(Contain(<slice of expected rules>))`
That will check the 'len' and if there are ever any additions change to the Rules then updating the one Expect will cover the len change and the rule change. Also if we do have an error here, the error will include what was expected and the actual, making it more obvious what the problem was.
(I'm noticing this because of some recent issues with some rule tests).

pkg/render/manager_test.go
```diff
@@ -256,13 +262,19 @@ var _ = Describe("Tigera Secure Manager rendering tests", func() {
 		Expect(deployment.Spec.Template.Spec.Volumes[5].Secret.SecretName).To(Equal(render.VoltronTunnelSecretName))
 		Expect(deployment.Spec.Template.Spec.Volumes[6].Name).To(Equal("elastic-ca-cert-volume"))
 		Expect(deployment.Spec.Template.Spec.Volumes[6].Secret.SecretName).To(Equal(render.ElasticsearchPublicCertSecret))
+
+		clusterRole := GetResource(resources, render.ManagerClusterRole, "", "rbac.authorization.k8s.io", "v1", "ClusterRole").(*rbacv1.ClusterRole)
+		Expect(len(clusterRole.Rules)).To(Equal(3))
```
tmjd, line 267: same comment here

PR #639, opened 2020-06-17
WIP: Upgrade the default resource settings for Elastic

rene-dekker:
The previous defaults are insufficient and lead to occasional e2e failures. 

Some of our documentation for installations already instruct to apply logstorage containing the new default settings, while other installation types fall back on these defaults.

### DIFF
--- a/pkg/render/logstorage.go
+++ b/pkg/render/logstorage.go
@@ -345,33 +345,25 @@ func (es elasticsearchComponent) pvcTemplate() corev1.PersistentVolumeClaim {
 
 // Generate the pod template required for the ElasticSearch nodes (controls the ElasticSearch container)
 func (es elasticsearchComponent) podTemplate() corev1.PodTemplateSpec {
-	// Setup default configuration for ES container
+	// Setup default configuration for ES container. For more information on managing resources, see:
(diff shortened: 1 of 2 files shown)


pkg/render/logstorage.go
```diff
@@ -345,33 +345,25 @@ func (es elasticsearchComponent) pvcTemplate() corev1.PersistentVolumeClaim {
...
 // Generate the pod template required for the ElasticSearch nodes (controls the ElasticSearch container)
 func (es elasticsearchComponent) podTemplate() corev1.PodTemplateSpec {
-	// Setup default configuration for ES container
+	// Setup default configuration for ES container. For more information on managing resources, see:
+	// https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-managing-compute-resources.html and
+	// https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-jvm-heap-size.html#k8s-jvm-heap-size
+
 	esContainer := corev1.Container{
 		Name: "elasticsearch",
-		// Important note: Following Elastic ECK docs, the recommended practice is to set
```
rene-dekker, line 351: I left the links in there for reference to the readers, but also removed some comments as they were not correct or in line with our current defaults.

PR #638, opened 2020-06-17
Adjust typha autoscale

tmjd:
## Description

I think we should go to something like this PR for typha scaling.
Here is a go playground if you want to see the calculation in action.
https://play.golang.org/p/cG1mEyp0StU

## For PR author

- [ ] Tests for change.
- [ ] If changing pkg/apis/, run `make gen-files`
- [ ] If changing versions, run `make gen-versions`

## For PR reviewers

A note for code reviewers - all pull requests must have the following:

- [ ] Milestone set according to targeted release.
- [ ] Appropriate labels:
  - `kind/bug` if this is a bugfix.
  - `kind/enhancement` if this is a a new feature.
  - `enterprise` if this PR applies to Calico Enterprise only.

PR #647, opened 2020-06-21
Dockeree tolerations and new Typha replicaset algorithm

doucol:
## Description
This PR ensures that an exists toleration to calico-node for key: "node.kubernetes.io/not-ready". This is required for at least DockerEE. We are also adjusting the replicaset #s for Typha to make sure we only have 1 typha for clusters where there is <= 4 nodes and then scale up from there. You can see the scaling in the code / comments below. Once we get above 1000 nodes, we calc using a max node per typha of 200.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


pkg/controller/installation/typha_autoscaler.go
```diff
@@ -39,16 +39,14 @@ const (
...
-//     2              2
-//     3              3
-//   250              4
-//   500              5
-//  1000              6
-//  1500              7
-//  2000              8
-//  2000+            10
+//   Nodes          Replicas
+// <=    4                 1
```
tmjd, line 43: Just because we have a small cluster doesn't mean we shouldn't treat it as production. We should have HA in production, so as soon as we can have more than 1 typha we should. Ideally running 3 (when we can) so we have redundancy even when updating.

tmjd: I think we should do something like https://github.com/tigera/operator/pull/638 instead of having un-explained scalings of typha for the nodes. I think we should have some calculation based on the number of nodes. @caseydavenport WDYT?

pkg/render/node.go
```diff
@@ -491,11 +491,13 @@ func (c *nodeComponent) nodeDaemonset(cniCfgMap *v1.ConfigMap) *apps.DaemonSet {
 
 // nodeTolerations creates the node's tolerations.
 func (c *nodeComponent) nodeTolerations() []v1.Toleration {
-	return []v1.Toleration{
+	t := []v1.Toleration{
 		{Operator: v1.TolerationOpExists, Effect: v1.TaintEffectNoSchedule},
 		{Operator: v1.TolerationOpExists, Effect: v1.TaintEffectNoExecute},
+		{Operator: v1.TolerationOpExists, Key: "node.kubernetes.io/not-ready"},
```
tmjd, line 497: I think if we only need this when running on DockerEE then we should only tolerate it when that is the Provider.

pkg/controller/installation/typha_autoscaler_test.go
```diff
@@ -117,10 +144,10 @@ func verifyTyphaReplicas(c client.Client, expectedReplicas int) {
 	Eventually(func() int32 {
 		err := test.GetResource(c, typha)
 		Expect(err).To(BeNil())
-		// Just return an invalid number that will never match an expected replica count.
+		// Replicas defaults to 1:  https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#replicas
 		if typha.Spec.Replicas == nil {
-			return -1
+			return 1
```
tmjd, line 149: In the case where replicas should be 1, do we still expect our autoscaling to set 1 as the value? If so I think we should keep this as it was.

pkg/controller/installation/typha_autoscaler_test.go
```diff
@@ -117,10 +144,10 @@ func verifyTyphaReplicas(c client.Client, expectedReplicas int) {
...
 		}
 		return *typha.Spec.Replicas
-	}, 1*time.Second).Should(BeEquivalentTo(expectedReplicas))
+	}, 1).Should(BeEquivalentTo(expectedReplicas))
```
tmjd, line 152: Since this is a duration, I think 1 is 1 nanosecond (https://golang.org/pkg/time/#Duration).
Though https://onsi.github.io/gomega/#eventually says that if the number is a float64 then it is interpreted as seconds.
I'm not really sure, are you sure that this is still being interpreted as 1 second. I think I'd prefer to keep it as it was for clarity sake.

pkg/render/node.go
```diff
@@ -491,11 +491,12 @@ func (c *nodeComponent) nodeDaemonset(cniCfgMap *v1.ConfigMap) *apps.DaemonSet {
 
 // nodeTolerations creates the node's tolerations.
 func (c *nodeComponent) nodeTolerations() []v1.Toleration {
-	return []v1.Toleration{
-		{Operator: v1.TolerationOpExists, Effect: v1.TaintEffectNoSchedule},
```
caseydavenport, line 495: Are we sure it's OK to remove these tolerations?

pkg/controller/installation/typha_autoscaler_test.go
doucol: i believe this was treated properly as 1 second, but agree that we should not have lost that clarity - this has been fixed

pkg/controller/installation/typha_autoscaler_test.go
doucol: Our logic for updating typha replicas will not update the deployment if the existing replicas is nil (defaults to 1) and the expected replicas is 1. What we can do is to change that logic so that it always updates the deployment if replicas is nil.

pkg/render/node.go
doucol: K8s will automatically add this and the 'unreachable' toleration (Effect: NoExecute) for Daemonsets, but since we are setting tolerations explicitly, I don't think K8s will automatically add them. Perhaps this isn't correct, but then I don't understand why it didn't get automatically added in this case. Since we are already adding tolerations explicitly, I think we should go ahead and add both. Also, I don't have any evidence of efficacy (docs or otherwise) for the tolerations we've added for NoSchedule/NoExecute without Keys. What do they do?

pkg/controller/installation/typha_autoscaler.go
doucol: I think this is just a matter of being practical. In the case of <= 4 nodes, we can't afford to have 4 typhas running - that doesn't make any sense to me. Now, we could state that we should always have a minimum of 2 and ideally at least 3. But, what if we have <= 2 nodes? It seems like we should always try to have a *functional* cluster as our top priority followed closely by "production ready".

I don't really understand your comment about the "un-explained scalings". The new algorithm attempts to blend the practical with the ideal and I think it does a pretty decent job of balancing the two. That said, we might decide that we should adjust the #s slightly on the low end. For example:  <=2 (1 typha), <= 5 (2 typha) - however, we'll likely then need to add a toleration to the typha deployment so that it can run on master nodes.

pkg/render/node.go
doucol: I don't have any evidence for efficacy (docs or otherwise). What do they do?

caseydavenport: These tolerate any taint which has the effect "NoSchedule" or "NoExecute", no matter what the key / value for the particular taint is. They act as a "catch-all".

pkg/controller/installation/typha_autoscaler.go
caseydavenport: > // <=    4                 1

Isn't this suggesting that for any cluster less than 4 nodes we'll just use a single Typha? I think @tmjd meant that we should do something like this:

- 1 node -> 1 typha
- 2 nodes -> 2 typhas
- 3 nodes -> 3 typhas
- 4 nodes - > 3 typhas
- 5 nodes -> 3 typhas
. . .
- 250 nodes -> 4 typhas

pkg/controller/installation/typha_autoscaler.go
```diff
@@ -78,25 +76,23 @@ func newTyphaAutoscaler(client client.Client, options ...typhaAutoscalerOption)
...
 	case nodes <= 500:
 		return 5
 	case nodes <= 1000:
 		return 6
-	case nodes <= 1500:
-		return 7
-	case nodes <= 2000:
-		return 8
+	default:
+		return (nodes / maxNodesPerTypha) + 1
```
caseydavenport, line 94: I think we should set a hard cap on the number of typhas that we'll run via this calclulation - say at 20.

pkg/render/node.go
doucol: > These tolerate any taint which has the effect "NoSchedule" or "NoExecute", no matter what the key / value for the particular taint is. They act as a "catch-all".

K8s does this for empty keys/effect with operator of exists:
From here: https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/
"Note: There are two special cases: An empty key with operator Exists matches all keys, values and effects which means this will tolerate everything. An empty effect matches all effects with key key."

But, I haven't read anything that states you can use an exists operator with a stated effect and an empty key to get wildcard behavior...but perhaps what you are describing is undocumented...or I'm not understanding properly! 

However, the evidence that I have suggests that these are not working as intended. In the DockerEE case, the nodes have a "not-ready" taint with effect of NoSchedule and our calico-node pods would not get deployed.

caseydavenport: Yeah, it's possible that the behavior has changed. That documentation isn't super clear about it, but does imply that you must leave both key and effect empty in order to get the "match all" behavior.

However, I do know that we have used these particular tolerations on our DaemonSets in order to get this behavior for a very long time. Here's a v3.0 manifest that has this toleration on it! https://github.com/projectcalico/calico/blob/release-legacy/v3.0/getting-started/kubernetes/installation/hosted/calico.yaml#L112-L114

Maybe newer versions of k8s have tightened up what they allow? I'd expect API validation to catch an invalid configuration, though definitely not impossible that k8s has a bug here. 

> However, the evidence that I have suggests that these are not working as intended. In the DockerEE case, the nodes have a "not-ready" taint with effect of NoSchedule and our calico-node pods would not get deployed. With a "not-ready" toleration added, calico-node gets deployed.

What version of k8s is that DockerEE cluster? I'm interested in seeing if I can chase this down.

pkg/controller/installation/typha_autoscaler.go
tmjd: Yes I was suggesting we need 1:1, 2:2, 3:3, 4:3...
By un-explained I meant our other scaling levels at 250 and 500 and so on.

pkg/controller/installation/typha_autoscaler.go
```diff
@@ -39,16 +39,14 @@ const (
...
-//   250              4
-//   500              5
-//  1000              6
-//  1500              7
-//  2000              8
-//  2000+            10
+//   Nodes          Replicas
+// <=    2                 1
+// <=    5                 2
+// <=  100                 3
```
tmjd, line 45: I chatted with @caseydavenport a bit about this and we need to stick to 3 when 3 are possible. It is what is documented as suggested. Without pulling in more people and discussing with a wider audience we should have 3 when 3 are available.

pkg/controller/installation/typha_autoscaler.go
tmjd: please add a cap on the number of typhas like Casey suggested.
Also I think this means we'll be at 6 typhas until 1200 since I believe the division is integer math meaning 1199/200 = 5, I think that is ok but I thought I should point that out.

pkg/render/common.go
```diff
@@ -293,3 +293,12 @@ func secretsToRuntimeObjects(secrets ...*v1.Secret) []runtime.Object {
 	}
 	return objs
 }
+
+func GetStandardTolerations() []v1.Toleration {
```
tmjd, line 297: I think maybe something like GetNoExNoSchedTolerations or GetCriticalTolerations would be more descriptive. WDYT?

caseydavenport:
@doucol @tmjd is this still in progress? Looks like it has a conflict now.

tmjd:
I think this should be closed. I think we don't need the Dockeree tolerations. @doucol feel free to reopen if you disagree.

PR #666, opened 2020-06-26
Add pod security policies to the installation

mgleung:
## Description

Add pod security policies to the installation. These pod security policies will be installed by default though no effect will occur unless the pod security policy controller is enabled in the cluster. Since OpenShift uses their own version of pod security policies (security context constraints), pod security policies will only be installed in non-OpenShift clusters.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/apiserver.go
+++ b/pkg/render/apiserver.go
@@ -21,6 +21,7 @@ import (
 	"github.com/tigera/operator/pkg/components"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	policyv1beta1 "k8s.io/api/policy/v1beta1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -138,6 +139,10 @@ func (c *apiServerComponent) Objects() ([]runtime.Object, []runtime.Object) {
 		c.webhookReaderClusterRole(),
 		c.webhookReaderClusterRoleBinding(),
 	)
+
+	if !c.openshift {
+		objs = append(objs, c.apiServerPodSecurityPolicy())
+	}
 
 	return objs, nil
 }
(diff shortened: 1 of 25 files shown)


pkg/controller/installation/typha_autoscaler.go
```diff
@@ -78,25 +78,7 @@ func newTyphaAutoscaler(client client.Client, options ...typhaAutoscalerOption)
...
-		return 5
-	case nodes <= 1000:
-		return 6
-	case nodes <= 1500:
-		return 7
-	case nodes <= 2000:
-		return 8
-	}
-	return 10
+	return 1
```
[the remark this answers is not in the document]
mgleung: sorry, that was leftover from a custom build I did for something else.

pkg/render/apiserver.go
```diff
@@ -1038,3 +1054,67 @@ func (c *apiServerComponent) tigeraNetworkAdminClusterRole() *rbacv1.ClusterRole
...
+				policyv1beta1.Projected,
+				policyv1beta1.Secret,
+				policyv1beta1.DownwardAPI,
+				policyv1beta1.PersistentVolumeClaim,
+				policyv1beta1.HostPath,
+			},
+			HostNetwork: false,
+			HostPorts: []policyv1beta1.HostPortRange{
+				policyv1beta1.HostPortRange{
+					Min: int32(0),
```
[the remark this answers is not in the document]
mgleung: The intention was to allow all ports. I believe port 0 is a fake port that could be specified when the system is supposed to assign a free port number. Might be OK to skip since I don't think any components specifically call on it, but since the intention of the policy was to not limit any ports (those should be restricted via network policy), I left all the host ports open.

pkg/render/logstorage.go
```diff
@@ -151,6 +152,10 @@ type elasticsearchComponent struct {
 func (es *elasticsearchComponent) Objects() ([]runtime.Object, []runtime.Object) {
 	var toCreate, toDelete []runtime.Object
 
+	if es.provider == operatorv1.ProviderOpenShift {
```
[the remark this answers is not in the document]
mgleung: I think I forgot to remove this. This isn't the right place for setting this up even if the logic matched up.

pkg/render/compliance_test.go
```diff
@@ -172,7 +188,7 @@ var _ = Describe("compliance rendering tests", func() {
 			ExpectGlobalReportType(resources[21], "policy-audit")
 			ExpectGlobalReportType(resources[22], "cis-benchmark")
 
-			var dpComplianceServer = resources[len(expectedResources)-1].(*v1.Deployment)
+			var dpComplianceServer = resources[len(expectedResources)-6].(*v1.Deployment)
```
tmjd, line 191: Could you change this to use GetResource from pkg/render/render_utils_test.go? Its use would have made no change necessary.

pkg/render/fluentd.go
```diff
@@ -646,3 +654,61 @@ func (c *fluentdComponent) eksLogForwarderVolumes() []corev1.Volume {
...
+		},
+		RoleRef: rbacv1.RoleRef{
+			APIGroup: "rbac.authorization.k8s.io",
+			Kind:     "ClusterRole",
+			Name:     eksLogForwarderName,
+		},
+		Subjects: []rbacv1.Subject{
+			{
+				Kind:      "ServiceAccount",
+				Name:      "default",
```
tmjd, line 690: I'm not sure how I feel about using the default ServiceAccount here. Should we be creating one so we can do this binding?
Has there been any conversation around just using the default here?

pkg/render/fluentd_test.go
```diff
@@ -51,7 +51,7 @@ var _ = Describe("Tigera Secure Fluentd rendering tests", func() {
 	It("should render with a default configuration", func() {
 		component := render.Fluentd(instance, nil, esConfigMap, s3Creds, splkCreds, filters, eksConfig, nil, installation)
 		resources, _ := component.Objects()
-		Expect(len(resources)).To(Equal(2))
+		Expect(len(resources)).To(Equal(5))
```
tmjd, line 54: I think you should move this to after then you can use len(expectedResources).

pkg/render/fluentd_test.go
```diff
@@ -71,7 +74,7 @@ var _ = Describe("Tigera Secure Fluentd rendering tests", func() {
 			i++
 		}
 
-		ds := resources[1].(*apps.DaemonSet)
+		ds := resources[4].(*apps.DaemonSet)
```
tmjd, line 77: Please use GetResource here too

pkg/render/fluentd_test.go
```diff
@@ -105,7 +108,7 @@ var _ = Describe("Tigera Secure Fluentd rendering tests", func() {
 		}
 		component := render.Fluentd(instance, nil, esConfigMap, s3Creds, splkCreds, filters, eksConfig, nil, installation)
 		resources, _ := component.Objects()
-		Expect(len(resources)).To(Equal(3))
+		Expect(len(resources)).To(Equal(6))
```
tmjd, line 111: Same here, move to below expectedResources and use len for the expected value

pkg/render/fluentd_test.go
```diff
@@ -171,7 +177,7 @@ var _ = Describe("Tigera Secure Fluentd rendering tests", func() {
 		}
 		component := render.Fluentd(instance, nil, esConfigMap, s3Creds, splkCreds, filters, eksConfig, nil, installation)
 		resources, _ := component.Objects()
-		Expect(len(resources)).To(Equal(2))
+		Expect(len(resources)).To(Equal(5))
```
tmjd, line 180: Same here

pkg/render/fluentd_test.go
```diff
@@ -191,7 +200,7 @@ var _ = Describe("Tigera Secure Fluentd rendering tests", func() {
 			i++
 		}
 
-		ds := resources[1].(*apps.DaemonSet)
+		ds := resources[4].(*apps.DaemonSet)
```
tmjd, line 203: GetResource please.

pkg/render/fluentd_test.go
```diff
@@ -241,12 +250,12 @@ var _ = Describe("Tigera Secure Fluentd rendering tests", func() {
...
 		instance.Spec.AdditionalStores = &operatorv1.AdditionalLogStoreSpec{
 			Splunk: &operatorv1.SplunkStoreSpec{
-				Endpoint:    "https://1.2.3.4:8088",
+				Endpoint: "https://1.2.3.4:8088",
 			},
 		}
 		component := render.Fluentd(instance, nil, esConfigMap, s3Creds, splkCreds, filters, eksConfig, nil, installation)
 		resources, _ := component.Objects()
-		Expect(len(resources)).To(Equal(4))
+		Expect(len(resources)).To(Equal(7))
```
tmjd, line 258: Ok I'm going to stop saying Same here. Please fix up the rest of the Expect(len(resources)) and the resources[X] places. Thanks.

pkg/render/intrusion_detection.go
```diff
@@ -545,3 +553,53 @@ func (c *intrusionDetectionComponent) globalAlertTemplates() []runtime.Object {
...
+		},
+		Subjects: []rbacv1.Subject{
+			{
+				Kind:      "ServiceAccount",
+				Name:      "intrusion-detection-controller",
+				Namespace: IntrusionDetectionNamespace,
+			},
+			{
+				Kind:      "ServiceAccount",
+				Name:      "default",
```
tmjd, line 600: Why do we need default here?

mgleung: The installer job doesn't have a a service account associated with it so we need to make this accessible to the job by applying it to the `default` service account.

pkg/render/node.go
```diff
@@ -624,6 +639,9 @@ func (c *nodeComponent) flexVolumeContainer() v1.Container {
 		Name:         "flexvol-driver",
 		Image:        components.GetReference(components.ComponentFlexVolume, c.cr.Spec.Registry, c.cr.Spec.ImagePath),
 		VolumeMounts: flexVolumeMounts,
+		SecurityContext: &v1.SecurityContext{
```
tmjd, line 642: I think we want to add this but why not add it to the CNI container also?

mgleung: I missed it oops. :P

MuhtasimTanmoy:
Now that [pod-security-policy](https://kubernetes.io/docs/concepts/security/pod-security-policy) is deprecated, is any work ongoing to support [pod-security-admission](https://kubernetes.io/docs/concepts/security/pod-security-admission)?

PR #667, opened 2020-06-26
Fix vulnerabilities

gianlucam76:


build/bin/remove_unneeded
```diff
@@ -0,0 +1,110 @@
+#!/bin/bash
+
+# Sanity check: the file is created by the Dockerfile.
+if [ ! -f /in-the-container ]; then
+  echo "Don't run this outside the container!"
+  exit 1
+fi
+
+# Array of basic tools needed by entrypoint and user_setup.
+basicNeed=("bin/sh" "/bin/bash" "/usr/bin/coreutils")
```
bcreane, line 10: I'd love to get get rid of some of these (`bash` in particular), but I think the scope is fine for a quicker `release-v1.2` fix.

build/bin/remove_unneeded
```diff
@@ -0,0 +1,110 @@
...
+libraries=()
+
+# Get all that is needed by basicNeed
```
bcreane, line 13: this comment is a bit minimal - you're filtering on shared libraries, identified by `\.so`.

build/bin/remove_unneeded
```diff
@@ -0,0 +1,110 @@
...
+for i in "${basicNeed[@]}"; do
+    ldd=$(ldd "${i}" \
+        | grep so \
```
bcreane, line 16: This pattern is broad - can we do something like `grep -P "\.so\.\d?"` so that we only match shared libraries?

build/bin/remove_unneeded
```diff
@@ -0,0 +1,110 @@
...
+        | sed -e '/^[^\t]/ d' \
```
bcreane, line 17: Is it possible to simplify all these `sed`, `sort`, `awk` etc commands? For example, In my limited testing, the following pulls out the shared library path:

```
ldd /usr/bin/bash | grep -P "\.so\.\d?" | awk '{print $1}'
```

build/bin/remove_unneeded
```diff
@@ -0,0 +1,110 @@
...
+        fi
+    done
+done
+
+# Recursively find all that is needed
+loop=1
+while [ "$loop" -eq 1 ]; do 
+    loop=0
+    for library in "${libraries[@]}"; do
+        ldd=$(ldd "$library" \
```
bcreane, line 40: consider refactoring this identical code into a bash function.

bcreane, line 40: please rename the `ldd` variable to something different from the `ldd` binary.

build/bin/remove_unneeded
```diff
@@ -0,0 +1,110 @@
...
+            fi
+        done
+        if [ "$remove" -eq 1 ]; then
+            rm -rf ${entry}
+        fi
+    done
+done
+
+# Remove un-needed tools. Keep only what is needed by user_setup and entrypoint
+leave=(mkdir chown chmod rm whoami ls sh bash)
```
bcreane, line 96: leave -> "keep"?

build/bin/remove_unneeded
```diff
@@ -0,0 +1,110 @@
+#!/bin/bash
```
bcreane, line 1: nit, consider renaming `remove_unneeded` to `remove_unneeded.sh` - the suffix is a useful hint.

bcreane: Also, if you don't already, please make sure to remove `remove_unneeded.sh` from the scratch image.

gianlucam76: It is not copied to scratch.

build/bin/remove_unneeded
gianlucam76: Thanks. I changed to grep "\.so\.[0-9]*"

build/bin/remove_unneeded
gianlucam76: I simplified a little bit but if it's OK I would prefer keeping the sed as it gives me full path instead of just library name.

build/bin/remove_unneeded
gianlucam76: bash unfortunately is not that easy. The script is bash not sh (easier to write). I will keep this in mind for master though.

PR #668, opened 2020-06-29
Update status once installation CR is written

lmm:
## Description

This PR writes an empty status block after we've written the installation CR to the datastore. We want this change because an operator scorecard test checks for the existence of the `status` in the CR after applying the CR manifest. The operator scorecard tests are run when we submit an operator metadata bundle to certify a new version of our operator.

We'll want to backport this to 1.6, 1.7 and 1.8 to cover Calico v3.14 and v3.15.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/installation/core_controller.go
+++ b/pkg/controller/installation/core_controller.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"net"
 	"os"
+	"reflect"
 	"strings"
 	"time"
 
@@ -426,6 +427,16 @@ func (r *ReconcileInstallation) Reconcile(request reconcile.Request) (reconcile.
 
 	ctx := context.Background()
 
+	// Get the installation object if it exists so that we can save the original
+	// status before we merge/fill that object with other values.
+	instance := &operator.Installation{}
+	if err := r.client.Get(ctx, utils.DefaultInstanceKey, instance); err != nil && apierrors.IsNotFound(err) {
+		reqLogger.Info("Installation config not found")
+		r.status.OnCRNotFound()
+		return reconcile.Result{}, nil
+	}
+	status := instance.Status
+
 	// Query for the installation object.
 	instance, err := GetInstallation(ctx, r.client, r.autoDetectedProvider)
 	if err != nil {
@@ -454,6 +465,17 @@ func (r *ReconcileInstallation) Reconcile(request reconcile.Request) (reconcile.
 	if err = r.client.Update(ctx, instance); err != nil {
 		r.SetDegraded("Failed to write defaults", err, reqLogger)
 		return reconcile.Result{}, err
+	}
+
+	// A status is needed at this point for operator scorecard tests.
+	// status.variant is written later but for some tests the reconciliation
+	// does not get to that point.
+	if reflect.DeepEqual(status, operator.InstallationStatus{}) {
+		instance.Status = operator.InstallationStatus{}
+		if err = r.client.Status().Update(ctx, instance); err != nil {
+			r.SetDegraded("Failed to write default status", err, reqLogger)
+			return reconcile.Result{}, err
+		}
 	}
 
 	// The operator supports running in a "Calico only" mode so that it doesn't need to run TSEE specific controllers.
--- a/test/mainline_test.go
+++ b/test/mainline_test.go
@@ -269,6 +269,7 @@ func setupManager() (client.Client, manager.Manager) {
 	err = controller.AddToManager(mgr, options.AddOptions{
 		DetectedProvider:    operator.ProviderNone,
 		EnterpriseCRDExists: true,
+		AmazonCRDExists:     true,
 	})
 	Expect(err).NotTo(HaveOccurred())
 	return mgr.GetClient(), mgr
--- a/test/mainline_test.go
+++ b/test/mainline_test.go
@@ -17,6 +17,7 @@ package test
 import (
 	"context"
 	"fmt"
+	"reflect"
 	"strings"
 	"time"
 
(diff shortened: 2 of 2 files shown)


pkg/controller/installation/core_controller.go
```diff
@@ -456,6 +456,14 @@ func (r *ReconcileInstallation) Reconcile(request reconcile.Request) (reconcile.
 		return reconcile.Result{}, err
 	}
 
+	// A status is needed at this point for operator scorecard tests.
+	// status.variant is written later but for some tests the reconciliation
+	// does not get to that point.
+	instance.Status = operator.InstallationStatus{}
```
caseydavenport, line 462: This will overwrite any previously written status. Is that what we want here? Or should we be maintaining any status that the controller had previously written?

pkg/controller/installation/core_controller.go
```diff
@@ -456,6 +456,14 @@ func (r *ReconcileInstallation) Reconcile(request reconcile.Request) (reconcile.
...
+	if err = r.client.Status().Update(ctx, instance); err != nil {
```
ozdanborne, line 463: Makes sense during the first run. But on subsequent runthroughs of the reconcilliation loop, won't this nuke existing statuses? Or is that not a problem because we mainly write status to the tigerastatus CR instead of the installation CR?

FWIW, I ran into this same gap while working on a different feature and this is what I came up with: https://github.com/ozdanborne/operator/commit/df37bb48d34c9543aa6cefd6f622eeb55e776df1

pkg/controller/installation/core_controller.go
lmm: No, that wasn't right. I've updated the commit so that we only do the pre-write of the installation status if there was no status at the start of the reconcile.

pkg/controller/installation/core_controller.go
lmm: What I had was wrong. But thanks for sharing that diff... I've written my fixed commit so that it might merge with your future changes a bit more easily.

pkg/controller/installation/core_controller.go
```diff
@@ -456,6 +467,17 @@ func (r *ReconcileInstallation) Reconcile(request reconcile.Request) (reconcile.
 		return reconcile.Result{}, err
 	}
 
+	// A status is needed at this point for operator scorecard tests.
+	// status.variant is written later but for some tests the reconciliation
+	// does not get to that point.
+	if reflect.DeepEqual(status, operator.InstallationStatus{}) {
```
tmjd, line 473: Why wait till here to do this status update? Why not do it right after the `if` block you added above?

test/mainline_test.go
```diff
@@ -125,6 +126,38 @@ var _ = Describe("Mainline component function tests", func() {
...
+					return err
+				}
+				if instance.Status.Variant != operator.Calico {
+					return fmt.Errorf("installation status not Calico yet")
+				}
+				return nil
+			}, 60*time.Second).Should(BeNil())
+
+			By("Checking that the installation status does not change")
+			Consistently(func() error {
```
tmjd, line 147: Why will this not change? shouldn't the Variant get written?

test/mainline_test.go
```diff
@@ -236,6 +269,7 @@ func setupManager() (client.Client, manager.Manager) {
 	err = controller.AddToManager(mgr, options.AddOptions{
 		DetectedProvider:    operator.ProviderNone,
 		EnterpriseCRDExists: true,
+		AmazonCRDExists:     true,
```
tmjd, line 272: Was this changed needed/intended?

test/mainline_test.go
lmm: The block just above checks that the variant is written before continuing.

pkg/controller/installation/core_controller.go
lmm: I figure we should wait until the merged/filled installation is written first before updating the status.

test/mainline_test.go
lmm: This fixes an issue with the tests where the tests in mainline_test.go exit early because we hit this line: https://github.com/tigera/operator/blob/master/pkg/controller/installation/core_controller.go#L490

We install all the crds right before the tests are run so we end up hitting that "reboot" and skip the remainder of tests in mainline_test.go

lmm:
Thanks for looking @tmjd 

> What status information does the operator scorecard expect to be present?

The test checks whether the Installation CR has a status block present. It's `checkstatustest` in this list: https://sdk.operatorframework.io/docs/scorecard/scorecard/#basic-operator

> A 'real' status that we could include here is a 'Valid' field indicating that the resource has been parsed and looks to be valid.

That's a reasonable idea, maybe we can revisit that later, wdyt?

> Maybe since we intend our main status reporting to be through the tigerastatus resource we should include a reference in the Installation.Status that points to the tigerastatus? (just throwing out ideas)

:+1: I think there's something to explore there for sure. But for the sake of satisfying the scorecard test maybe we can just add in the empty status block for now since it's an easier change to approve since we're not changing the API.

pkg/controller/installation/core_controller.go
```diff
@@ -426,6 +427,16 @@ func (r *ReconcileInstallation) Reconcile(request reconcile.Request) (reconcile.
 
 	ctx := context.Background()
 
+	// Get the installation object if it exists so that we can save the original
+	// status before we merge/fill that object with other values.
+	instance := &operator.Installation{}
+	if err := r.client.Get(ctx, utils.DefaultInstanceKey, instance); err != nil && apierrors.IsNotFound(err) {
```
tmjd, line 433: Do we really need this extra get? Couldn't we use the status on the object that comes from GetInstallation?

test/mainline_test.go
tmjd: Oh right, I was reading that previous one as checking that Variant was not Calico. My bad.

pkg/controller/installation/core_controller.go
lmm: The main reason we are doing this here is because GetInstallation actually does two things: it gets the Installation then it merges defaults and such into that Installation before returning it. I don't think changing the signature of GetInstallation makes sense (since other usages of GetInstallation don't need the original Instance).

The Installation returned by GetInstallation will have its status already set to the default.

PR #670, opened 2020-06-29
Move to scratch

gianlucam76:


build/Dockerfile.amd64
```diff
@@ -22,9 +29,8 @@ LABEL name="Calico Operator" \
       description="Calico Operator manages the lifecycle of a Calico or Tigera Secure installation on Kubernetes or OpenShift" \
       maintainer="Laurence Man <laurence@tigera.io>"
 
-# Add in top-level license file
-RUN mkdir /licenses
-COPY LICENSE /licenses
+COPY --from=ubi /etc/passwd /etc/passwd
```
doublek, line 32: Do you know who uses this file? I thought it was used only as a part of entrypoint script and perhaps now it's not required?

gianlucam76: It does not appear to be needed. I can remove if that's what we want.

gianlucam76: Not needed. I can remove it if that's what we want.

doublek: If it's not strictly required I am okay removing it. I would like @tmjd's opinion as well.

gianlucam76: Just FYI, removing following I was able to start from scratch and all worked (tested with crc).
 
-COPY --from=ubi /etc/passwd /etc/passwd
-COPY --from=ubi /licenses /licenses

doublek: I think `-COPY --from=ubi /licenses /licenses` might be required but probably not the copy for `/etc/passwd`.

tmjd: I don't know of a reason it is needed. If it works without I'm happy for it to be removed.
Awesome @gianlucam76, thanks for doing this.

build/Dockerfile.amd64
```diff
@@ -22,20 +23,13 @@ LABEL name="Calico Operator" \
       description="Calico Operator manages the lifecycle of a Calico or Tigera Secure installation on Kubernetes or OpenShift" \
       maintainer="Laurence Man <laurence@tigera.io>"
 
-# Add in top-level license file
-RUN mkdir /licenses
-COPY LICENSE /licenses
```
doublek, line 27: I think license is still required for certification requirements. Sorry if I wasn't clear in my previous comment. My comment was to only remove /etc/passwd which shouldn't be required anymore.

build/Dockerfile.amd64
```diff
@@ -22,20 +33,13 @@ LABEL name="Calico Operator" \
...
     USER_NAME=operator
 
 # install operator binary
 COPY build/_output/bin/operator-amd64 ${OPERATOR}
 
-COPY build/bin /usr/local/bin
-RUN  /usr/local/bin/user_setup
-
-ENTRYPOINT ["/usr/local/bin/entrypoint"]
+ENTRYPOINT ["/usr/local/bin/operator"]
```
tmjd, line 43: minor thing: Can we use `${OPERATOR}` here too?

PR #678, opened 2020-07-07
Compliance: mount /usr/bin to different directory

gianlucam76:
This change is needed by https://github.com/tigera/compliance/pull/283


pkg/render/compliance.go
```diff
@@ -859,7 +859,7 @@ func (c *complianceComponent) complianceBenchmarkerDaemonSet() *appsv1.DaemonSet
 		{Name: "var-lib-kubelet", MountPath: "/var/lib/kubelet", ReadOnly: true},
 		{Name: "etc-systemd", MountPath: "/etc/systemd", ReadOnly: true},
 		{Name: "etc-kubernetes", MountPath: "/etc/kubernetes", ReadOnly: true},
-		{Name: "usr-bin", MountPath: "/usr/bin", ReadOnly: true},
+		{Name: "usr-bin", MountPath: "/usr/local_bin", ReadOnly: true},
```
doublek, line 862: nit: Can you rename this to `/host/usr/bin` or something like that, assuming that it's possible to put this in such a path and not break a bunch of stuff. `local_bin` feels too close to `/usr/local/bin` for my tastes.

PR #682, opened 2020-07-08
Add ClusterServiceVersion bundling

lmm:
## Description

This PR adds plumbing to generate ClusterServiceVersion (CSV) bundles which we use to validate and publish the operator on the RH catalog.

Two new make targets are added:
- `gen-csv` creates a CSV for every operator version we want to publish;
- `gen-bundle` gathers all generated CSVs, adds them to the ones already committed, and creates a zip file

This PR also removes operator manifests from the deploy directory since they are not used for CSV generation. Instead, users should use the operator manifests from our [docs](https://docs.projectcalico.org/getting-started/kubernetes/).

### Testing

- The CSV bundling was tested by generating them and uploading them for validation testing. I also locally installed these bundles in my OCP cluster to verify the operator metadata in the OpenShift Console.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


Makefile
```diff
@@ -442,6 +442,22 @@ $(BINDIR)/gen-versions: $(shell find ./hack/gen-versions -type f)
...
 
+## Generate a ClusterServiceVersion package.
+# E.g. make gen-csv VERSION=1.6.2 PREV_VERSION=0.0.0
+#
+# VERSION: the operator version to generate a CSV for.
+# PREV_VERSION: the operator version that this CSV will replace. If there is
+#               no previous version, use 0.0.0
+.PHONY: gen-csv
+gen-csv: hack/bin/operator-sdk-$(OPERATOR_SDK_VERSION)
+	hack/gen-csv/csv.sh
```
tmjd, line 453: I think we should containerize this. Would `$(CONTAINERIAED) hack/gen-csv/csv.sh` work?
Same for gen-bundle below?

RELEASING.md
```diff
@@ -40,3 +40,33 @@ You should have no local changes and tests should be passing.
...
+
+4. Create the CSV bundle:
+
+   ```
+   make gen-bundle
+   ```
+
+   This step will add the CSV we generated earlier to `deploy/olm-catalog` and create a zip file in `build/_output/bundle/bundle.zip`.
+
+5. `git add deploy/olm-catalog`, commit the changes, and have the PR reviewed and merged into master.
```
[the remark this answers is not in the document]
lmm: Yes, that's it. We need to keep track of previously generated CSV's. The bundle we submit to RH will have all versions we have published (plus any new versions).

lmm:
> Should we be committing these bundles to this repo? Did you consider any other options?

Pre-generating the CSV's as needed was an option that was considered but the trade off is that we'd need some other way to keep track of the operator versions published on RH. The other thing is that the many operators in the operator-sdk ecosystem already check-in their CSVs, and since we already follow a lot of the operator-sdk conventions already, the proposed process in this PR is consistent.

> I'm not sure if bundling the CRDs in this repo is right, I'm wondering if instead of that we should be pulling from the docs. The ones here are more for testing purposes than anything and the 'real' ones should be coming from the docs. Or maybe part of the process needs to be pulling the latest Calico CRDs from the docs.

I had a brief chat with @caseydavenport about this and the gist of that was that maybe we want the manifests in `deploy/` to  be the "real" ones that the docs follow. I think that simplifies the process for CSV bundling but makes it slightly awkward if contributors submit changes to operator manifests in the calico repo. Maybe Casey had some more thoughts on this.

RELEASING.md
```diff
@@ -40,3 +40,33 @@ You should have no local changes and tests should be passing.
...
+
+## Creating a release on RH Catalog
+
+1. After the semaphore job in the releasing steps is complete, and images have been tagged and pushed, checkout the tag you released.
+
+1. Create the ClusterServiceVersion (CSV), using the tag version for VERSION and the version that this replaces in PREV_VERSION.
+   The versions are semver strings. For example:
+
+   ```
+   make gen-csv VERSION=1.3.1 PREV_VERSION=1.3.0
```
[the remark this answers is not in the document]
lmm: The PREV_VERSION will set the `spec.replaces` field in the CSV but in the `make gen-bundle` command we loop through and process all of the CSVs, and may end up overwriting the `spec.replaces` field in some CSVs. This addresses the scenario you've highlighted: adding a new version that is in-between 2 existing versions.

E.g. say we start with `v1.31 <= v1.6.2`. (The `<=` here will mean "replaces").

Then we have `v1.3.2` that we've released and want to publish. After generating the CSV and the bundle, the new bundle we commit to master will have a version chain that looks like:

```
v1.3.1 <= v1.3.2 <= v1.6.2
```

We will have updated the `spec.replaces` field in the `v1.6.2` CSV from `v1.3.1` to `v1.3.2`.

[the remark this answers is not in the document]
lmm: One other thing to call out is that the tooling here is proposing a single channel of versions. I don't think we need anything more than this right now. In the future we could have separate channels for alpha or beta versions (like this [operator](https://github.com/IBM/ibm-ingress-nginx-operator/blob/master/deploy/olm-catalog/ibm-ingress-nginx-operator/ibm-ingress-nginx-operator.package.yaml)) or maybe separate channels per release branch (this would allow a user to stay pinned to a minor version of Calico).

Makefile
lmm: Actually, using the gobuild container won't work because it doesn't have docker itself. The csv.sh script uses `docker pull` and `docker inspect`. Maybe it's still okay since we're using the operator-sdk binary in hack/ ?

RELEASING.md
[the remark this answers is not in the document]
tmjd: So I'm understanding correctly you're saying there is some 'magic' in gen-bundle that will change the v1.6.2 spec.replaces field with v1.3.2 when it originally had v1.3.1.
Have you had any chats about the channels? I think I'm onboard with only having the one.

Makefile
tmjd: I want to use the containerized because I'm concerned about using the operator-sdk specifically. That's because I think it also depends on the go version (at least from previous experience).

lmm: Yeah you're right. I've addressed that with b5caf0b

Once I get https://github.com/projectcalico/go-build/pull/115 approved+merged+released, that should fix the build

RELEASING.md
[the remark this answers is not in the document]
lmm: > So I'm understanding correctly you're saying there is some 'magic' in gen-bundle that will change the v1.6.2 spec.replaces field with v1.3.2 when it originally had v1.3.1.

Yes, once the script has copied over any CSVs to `deploy/olm-catalog`, it sorts the CSV versions in descending order and overwrites the spec.replaces field in the CSVs (all but the last CSV since being last it replaces nothing).

> Have you had any chats about the channels? I think I'm onboard with only having the one.

Not yet, just this one we're having now.

caseydavenport:
> Should we be committing these bundles to this repo? Did you consider any other options?

Yeah, this strongly weirds me out, but it seems to be the pattern that other RH operators are according to my discussion with @lmm . 

The main difference I think, based on my conversation with @lmm was that the other operators have a single release stream (e.g., the only maintain the master branch, not additional release branches).

Makefile
```diff
@@ -442,6 +442,50 @@ $(BINDIR)/gen-versions: $(shell find ./hack/gen-versions -type f)
...
+#             "quay.io/tigera/operator@sha256:5e1d551b5a711592472f4a3cc4645698d5f826da4253f0d47cfa5d5b641a2e1a",
+#             "..."
+#         ],
+#         ...
+#     }
+# ]
+.PHONY: get-digest
+get-digest: prepull-image
+	@echo Getting operator image digest...
+	$(eval OPERATOR_IMAGE_DIGEST=$(shell sh -c "docker image inspect quay.io/tigera/operator:v$(VERSION) | jq -r '.[0].RepoDigests[] | select(. | contains(\"quay.io/tigera/operator\"))' "))
```
tmjd, line 481: This is better but I still don't like depending on having jq on the host system.
It would be easy enough to just pass in the whole inspect output right?

tmjd:
> > I'm not sure if bundling the CRDs in this repo is right, I'm wondering if instead of that we should be pulling from the docs. The ones here are more for testing purposes than anything and the 'real' ones should be coming from the docs. Or maybe part of the process needs to be pulling the latest Calico CRDs from the docs.
> 
> I had a brief chat with @caseydavenport about this and the gist of that was that maybe we want the manifests in `deploy/` to be the "real" ones that the docs follow. I think that simplifies the process for CSV bundling but makes it slightly awkward if contributors submit changes to operator manifests in the calico repo. Maybe Casey had some more thoughts on this.

I think I'd be worried about moving the CRDs into the operator repo without some automation around copying them to the docs and a better understanding of what the process around that would be. Mainly because a lot of integration testing uses the docs so if they get out of sync that would not be good.

I feel like it would be simple enough to download the CRDs as part of the CSV generation step, I guess the caveat with that would be that we should wait until the Calico release is done and the CRDs are 'officially' available.

@caseydavenport WDYT?

Makefile
lmm: Fixed. I had to base64 encode it to get around escaping issues.

lmm:
> I feel like it would be simple enough to download the CRDs as part of the CSV generation step

Thanks @tmjd, I've added this with https://github.com/tigera/operator/pull/682/commits/9d8ca79b8f4d1a29dc2e1b62c8c416eebe950d52 and https://github.com/tigera/operator/pull/682/commits/2d0fea32e8666466dd1c43beb68f3142e580d8d6

hack/gen-csv/get-manifests.sh
```diff
@@ -0,0 +1,105 @@
...
+# This script downloads Calico and operator manifests that are used to generate
+# ClusterServiceVersions.
+set -ex
+
+# Get the base path for the Calico docs site. This will be used to download manifests.
+CALICO_BASE_URL=https://docs.projectcalico.org
+
+if [ -f config/calico_versions.yml ]; then
+    CALICO_VERSION=$(yq read config/calico_versions.yml components.typha.version)
+elif [ -f config/os_versions.yaml ]; then
```
tmjd, line 12: Same as my other comment about supporting multiple versions.

RELEASING.md
```diff
@@ -40,3 +40,33 @@ You should have no local changes and tests should be passing.
 1. Close the milestone for this release. https://github.com/tigera/operator/milestones
 
 1. Go to https://github.com/tigera/operator/milestones and create any new milestones that should exist (e.g., next patch release)
+
+## Creating a release on RH Catalog
+
+1. After the semaphore job in the releasing steps is complete, and images have been tagged and pushed, checkout the tag you released.
+
+1. Create the ClusterServiceVersion (CSV), using the tag version for VERSION and the version that this replaces in PREV_VERSION.
```
tmjd, line 48: Should we include any notes about needing updates if any new Calico CRDs were added since previous releases?

RELEASING.md
```diff
@@ -40,3 +40,33 @@ You should have no local changes and tests should be passing.
 1. Close the milestone for this release. https://github.com/tigera/operator/milestones
 
 1. Go to https://github.com/tigera/operator/milestones and create any new milestones that should exist (e.g., next patch release)
+
+## Creating a release on RH Catalog
```
tmjd, line 44: Should we include something that only releases targeting Calico (not Enterprise) should be added to the RH Catalog?

lmm: 556c0655aca9c858cdac91e71b563de44a1ff432

RELEASING.md
lmm: Yes, thanks: 556c0655aca9c858cdac91e71b563de44a1ff432

hack/gen-csv/get-manifests.sh
```diff
@@ -0,0 +1,105 @@
...
+    fi
+
+    # Download the Calico CRDs into CSV dir.
+    for resource in $CALICO_RESOURCES; do
+        curl ${CALICO_BASE_URL}/manifests/ocp/crds/calico/kdd/02-crd-${resource}.yaml --output ${CSV_DIR}/02-crd-${resource}.yaml
+    done
+}
+
+function downloadCalicoCRDs() {
+    if [ "${CALICO_VERSION}" == "v3.13" ] || [ "${CALICO_VERSION}" == "v3.14" ]; then
```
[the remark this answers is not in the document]
lmm: Addressed with 4784baea83d16d550189605be134ad75325eea83

hack/gen-csv/get-manifests.sh
lmm: Addressed with 4784baea83d16d550189605be134ad75325eea83

Makefile
```diff
@@ -445,6 +445,49 @@ $(BINDIR)/gen-versions: $(shell find ./hack/gen-versions -type f)
...
+gen-csv: hack/bin/operator-sdk-$(OPERATOR_SDK_VERSION) get-digest
+	$(eval EXTRA_DOCKER_ARGS += -e OPERATOR_IMAGE_INSPECT="$(OPERATOR_IMAGE_INSPECT)" -e VERSION=$(VERSION) -e PREV_VERSION=$(PREV_VERSION))
+	$(CONTAINERIZED) hack/gen-csv/csv.sh
+
+.PHONY: prepull-image
+prepull-image:
+	@echo Pulling operator image...
+	docker pull quay.io/tigera/operator:v$(VERSION)
+
+# Get the digest for the image. 'docker inspect' returns output like the example
```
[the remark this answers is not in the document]
lmm: Updated the comment.

RELEASING.md
```diff
@@ -41,7 +41,14 @@ You should have no local changes and tests should be passing.
 
 1. Go to https://github.com/tigera/operator/milestones and create any new milestones that should exist (e.g., next patch release)
 
-## Creating a release on RH Catalog
+## Updates for new Calico CRDs
+
+If the release includes new Calico CRDs, add the new CRDs to `hack/gen-csv/get-manifests.sh`.
```
tmjd, line 46: Also the template file if I understand correctly.

lmm: Thanks, updated. For some reason i thought I had automated that section of the template file.

PR #688, opened 2020-07-09
Tag and push release images to RedHat Connect for validation

lmm:
## Description

This PR adds a step to the release process which tags and pushes the newly released operator and operator-init images to RedHat Connect for validation.
It depends on a secret that has been added to Semaphore. That secret exports a pair of env variables for each image: one variable identifies the operator "project" on RH Connect, the other is a registry key used to login to scan.connect.redhat.com.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/hack/maybe-build-release.sh
+++ b/hack/maybe-build-release.sh
@@ -1,4 +1,5 @@
 #!/bin/bash
+set -e
 
 tag=$(git describe --exact-match --tags HEAD)
 if [[ -z "${tag}" ]]; then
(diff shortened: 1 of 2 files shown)


hack/maybe-build-release.sh
```diff
@@ -16,3 +17,21 @@ make release VERSION=${tag}
...
+
+echo "Tagging and pushing operator images to RedHat Connect for certification..."
+
+docker login -u unused scan.connect.redhat.com --password-stdin <<< ${OPERATOR_RH_REGISTRY_KEY}
+quayImage=quay.io/tigera/operator:$VERSION
+redhatImage=scan.connect.redhat.com/$OPERATOR_RH_PROJECTID/operator:$VERSION
+docker pull $quayImage
+docker tag $quayImage $redhatImage
+# Pushes to scan.connect.redhat.com fail if the image exists already. Return
+# a success exit code so we can retry image pushes if needed.
```
tmjd, line 29: I don't understand the comment. I don't see any retry being done.
If we're counting on the push failing if the image exists already, could we just check for it before doing the push?

lmm: I updated the script as you suggested to check if the image exists

PR #690, opened 2020-07-10
BPF Dataplane Mode

tomastigera:
## Description

The idea is that since we cannot rely on kube-proxy being present and thus kubernetes service working, we need k8s client with direct connection, hence we need to be able to override the host:port and these values need to be propagated to other components, namely Felix and CNI. There might be better, more automated, ways of getting these values, however, these are going to be very much environment dependent.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


pkg/apis/operator/v1/types.go
```diff
@@ -221,6 +233,13 @@ type CalicoNetworkSpec struct {
 	// +optional
 	// +kubebuilder:validation:Enum=None;Multus
 	MultiInterfaceMode *MultiInterfaceMode `json:"multiInterfaceMode,omitempty"`
+
+	// BPFDataplaneMode configures whether the BPF data plane is enabled anda in
+	// what form.
+	// Default: Disabled
+	// +optional
+	// +kubebuilder:validation:Enum=Disabled;Enabled;EnabledKeepKubeProxy
+	BPFDataplaneMode *BPFDataplaneMode `json:"bpfDataplaneMode,omitempty"`
```
caseydavenport, line 242: We should discuss where this option belongs. Not 100% obvious to me that this beloings in `CalicoNetworkSpec`.

pkg/controller/installation/core_controller.go
```diff
@@ -677,6 +697,14 @@ func (r *ReconcileInstallation) Reconcile(request reconcile.Request) (reconcile.
 		return reconcile.Result{}, err
 	}
 
+	// If enabling BPF disable kube-proxy as the last thing to keep services
+	// working as long as possible at least for host-networked pods.
+	if kubeProxyState == utils.KubeProxyRuns && netConf.BPFEnabled && !netConf.BPFKeepKubeProxy {
```
caseydavenport, line 702: I thought we weren't going to have logic in the operator to disable kube-proxy?

pkg/controller/utils/utils.go
```diff
@@ -244,3 +245,131 @@ func GetAmazonCloudIntegration(ctx context.Context, client client.Client) (*oper
...
+// GetK8sEndpoint returns an ip:port of a kubernetes endpoint
+func GetK8sEndpoint(c client.Client) (string, int, error) {
+	var (
+		ip   string
+		port int
+		err  error
+	)
+
+	eps := &corev1.Endpoints{}
+	err = c.Get(context.Background(), client.ObjectKey{Name: "kubernetes", Namespace: "default"}, eps)
```
caseydavenport, line 258: This looks like it's using the Kubernetes API to query the k8s endpoints? Why do we need to query an endpoint if we already have API access :D

pkg/controller/installation/core_controller.go
```diff
@@ -689,6 +717,15 @@ func GenerateRenderConfig(install *operator.Installation) render.NetworkConfig {
 	// If CalicoNetwork is specified, then use Calico networking.
 	if install.Spec.CalicoNetwork != nil {
 		config.CNI = render.CNICalico
+		if install.Spec.CalicoNetwork.BPFDataplaneMode != nil {
```
caseydavenport, line 720: I don't think we need this extra abstraction - the `Installation` object is passed through to render so we can just use it directly there.

I understand that there is already some abstraction here and it's inconsistent, but the vast majority of fields are passed straight through via the Installation object into the `render` code, and this is the exception case.

pkg/controller/installation/core_controller.go
```diff
@@ -495,6 +495,26 @@ func (r *ReconcileInstallation) Reconcile(request reconcile.Request) (reconcile.
...
 
+	if netConf.BPFEnabled {
+		var err error
+		netConf.K8sHost, netConf.K8sPort, err = utils.GetK8sEndpoint(r.client)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+	}
+
+	kubeProxyState, err := utils.GetKubeProxyState(r.client)
```
caseydavenport, line 506: Like we discussed in Slack, I do not think the operator should be in the business of modifying the kube-proxy (either installing it, modifying it, removing it, etc). 

Users should either install or remove the kube-proxy as needed independently of installing our operator.

pkg/apis/operator/v1/types.go
caseydavenport: For example, we might want to do something like this:
- Always have the operator set KUBERNETES_SERVICE_PORT and KUBERNETES_SERVICE_HOST
- Users enable / disable BPF itself via FelixConfiguration resources. 

Would that work? It wouldn't require any API changes here.

pkg/controller/installation/core_controller.go
tomastigera: ohhh this is not supposed to be here anymore as well as the other similar, I thought that I removed it :-/ thanks for spotting this!

pkg/controller/installation/core_controller.go
tomastigera: sure, I am all for that, was wondering why we rewrite one thing into another, but felt obliged to follow the pattern :-)

pkg/controller/utils/utils.go
tomastigera: man, I think I pushed a wrong branch! I will close this PR

PR #695, opened 2020-07-11
BPF enable

tomastigera:
## Description

The idea is that since we cannot rely on kube-proxy being present and thus kubernetes service working, we need k8s client with direct connection, hence we need to be able to override the host:port and these values need to be propagated to other components, namely Felix and CNI. There might be better, more automated, ways of getting these values, however, these are going to be very much environment dependent.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/deploy/crds/operator.tigera.io_installations_crd.yaml
+++ b/deploy/crds/operator.tigera.io_installations_crd.yaml
@@ -39,6 +39,14 @@ spec:
                 description: CalicoNetwork specifies configuration options for Calico
                   provided pod networking.
                 properties:
+                  bpfDataplaneMode:
+                    description: 'BPFDataplaneMode configures whether the BPF data
+                      plane is enabled anda in what form. Default: Disabled'
+                    enum:
+                    - Disabled
+                    - Enabled
+                    - EnabledDSR
+                    type: string
                   hostPorts:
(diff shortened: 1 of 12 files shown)


deploy/crds/operator.tigera.io_installations_crd.yaml
```diff
@@ -35,6 +35,14 @@ spec:
             description: Specification of the desired state for the Calico or Calico
               Enterprise installation.
             properties:
+              bpfDataplaneMode:
+                description: 'BPFDataplaneMode configures whether the BPF data plane
+                  is enabled anda in what form. Default: Disabled'
```
fasaxc, line 40: Typo: anda

deploy/crds/operator.tigera.io_installations_crd.yaml
```diff
@@ -35,6 +35,14 @@ spec:
...
+                enum:
+                - Disabled
+                - Enabled
+                - EnabledDSR
```
fasaxc, line 44: Not sure if DSR should be combined in here or left as a separate option (or even configured through felix config).  Probably need opinions form the operator team.

tomastigera:
Replaced by simpler https://github.com/tigera/operator/pull/700

PR #696, opened 2020-07-12
read existing config

ozdanborne:
## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/migration/convert/k8s.go
+++ b/pkg/controller/migration/convert/k8s.go
@@ -46,7 +46,7 @@ func (r *CheckedDaemonSet) uncheckedVars() []string {
 
 // getEnv gets the value of an environment variable and marks that it has been checked.
 func (r *CheckedDaemonSet) getEnv(ctx context.Context, client client.Client, container string, key string) (*string, error) {
-	c := getContainers(r.Spec.Template.Spec, container)
+	c := getContainer(r.Spec.Template.Spec, container)
 	if c == nil {
 		return nil, ErrIncompatibleCluster{fmt.Sprintf("couldn't find %s container in existing daemonset", container)}
 	}
--- a/pkg/controller/migration/convert/k8s.go
+++ b/pkg/controller/migration/convert/k8s.go
@@ -1,0 +1,101 @@
+package convert
+
+import (
+	"context"
+	"fmt"
+
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// CheckedDaemonSet keeps track of which fields have been 'checked' by handlers.
+// This is done so that at the end of the migration, any 'unchecked' fields can be reported
+// and errored.
+type CheckedDaemonSet struct {
+	appsv1.DaemonSet
+
+	checkedVars map[string]checkedFields
+}
+
+// uncheckedVars returns a list of all environment variables which
+// were not checked by handlers.
+func (r *CheckedDaemonSet) uncheckedVars() []string {
+	unchecked := []string{}
+
+	for _, t := range r.Spec.Template.Spec.Containers {
+		for _, v := range t.Env {
+			if _, ok := r.checkedVars[t.Name].envVars[v.Name]; !ok {
+				unchecked = append(unchecked, t.Name+"/"+v.Name)
+			}
+		}
+	}
+
+	for _, t := range r.Spec.Template.Spec.InitContainers {
+		for _, v := range t.Env {
+			if _, ok := r.checkedVars[t.Name].envVars[v.Name]; !ok {
+				unchecked = append(unchecked, t.Name+"/"+v.Name)
+			}
+		}
+	}
+
+	return unchecked
+}
+
+// getEnv gets the value of an environment variable and marks that it has been checked.
+func (r *CheckedDaemonSet) getEnv(ctx context.Context, client client.Client, container string, key string) (*string, error) {
+	c := getContainers(r.Spec.Template.Spec, container)
+	if c == nil {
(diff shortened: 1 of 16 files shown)


pkg/controller/migration/convert/network.go
```diff
@@ -0,0 +1,165 @@
...
+	if _, ok := c.pluginCNIConfig["portmap"]; ok {
+		hp := v1.HostPortsEnabled
+		install.Spec.CalicoNetwork.HostPorts = &hp
+	} else {
+		hp := v1.HostPortsDisabled
+		install.Spec.CalicoNetwork.HostPorts = &hp
+	}
+
+	// CNI bandwidth plugin
+	if _, ok := c.pluginCNIConfig["bandwidth"]; ok {
```
tmjd, line 83: I think this will block upgrades for now. I think we should just be ignoring this for now.

pkg/controller/migration/convert/network.go
```diff
@@ -0,0 +1,165 @@
...
+
+	if c.calicoCNIConfig.MTU == -1 {
+		// if MTU is -1, we assume it was us who replaced it when doing initial CNI
+		// config loading. We need to pull it from the correct source
+		mtu, err := c.node.getEnv(ctx, c.client, containerInstallCNI, "CNI_MTU")
+		if err != nil {
+			return err
+		}
+		if mtu == nil {
+			return fmt.Errorf("couldn't detect MTU")
```
tmjd, line 100: I don't think this would need to be an error case necessarily, if this isn't set previously are we ok with just letting the operator default it?

pkg/controller/migration/convert/network.go
```diff
@@ -0,0 +1,165 @@
...
+		}
+
+		// TODO: clean this up
```
tmjd, line 103: What clean up are you wanting to do here? Make sure the conversions don't error?

pkg/controller/migration/convert/network.go
```diff
@@ -0,0 +1,165 @@
...
+	} else {
+		// user must have hardcoded their CNI instead of using our cni templating engine.
+		// use the hardcoded value.
+		mtu := int32(c.calicoCNIConfig.MTU)
+		install.Spec.CalicoNetwork.MTU = &mtu
+	}
+
+	// CNI IPAM pools
+	if len(c.calicoCNIConfig.IPAM.IPv4Pools) != 0 {
+		return ErrIncompatibleCluster{"cni ipam ranges not suported"}
```
tmjd, line 116: I think this should be `cni config assigning specified IPPools is not supported`. WDYT?

pkg/controller/migration/convert/network.go
```diff
@@ -0,0 +1,165 @@
...
+		return operatorv1.NodeAddressAutodetection{CanReach: dest}, nil
+	}
+
+	// skip-interface
+	if strings.HasPrefix(method, AUTODETECTION_METHOD_SKIP_INTERFACE) {
+		ifStr := strings.TrimPrefix(method, AUTODETECTION_METHOD_SKIP_INTERFACE)
+		return operatorv1.NodeAddressAutodetection{SkipInterface: ifStr}, nil
+	}
+
+	return operatorv1.NodeAddressAutodetection{}, errors.New("unrecognized option for AUTODETECTION_METHOD_SKIP_INTERFACE: " + method)
```
tmjd, line 164: AUTODETECTION_METHOD_SKIP_INTERFACE isn't the right thing to report error on. I wonder if this function should return an error that the higher level function should add context to (because this function can be used for IPv6 or IPv4.

pkg/controller/migration/convert/utils.go
```diff
@@ -0,0 +1,17 @@
+package convert
+
+import corev1 "k8s.io/api/core/v1"
+
+func getContainers(spec corev1.PodSpec, name string) *corev1.Container {
```
tmjd, line 5: This function only gets 1 container so I feel like the name should not be plural.

pkg/controller/migration/convert/network.go
ozdanborne: Ah I see that in `install-cni.sh` if `__CNI_MTU__` is in the cni conf but `CNI_MTU` env isn't set, we just use 1500. I returned this error because I thought install-cni would throw an error. I will remove this.

pkg/controller/migration/convert/network.go
ozdanborne: It just seemed bad that it took me three lines of code to go from from string to int to int32 to *int32

pkg/controller/migration/convert/k8s.go
```diff
@@ -0,0 +1,101 @@
...
+			Namespace: "kube-system",
+		}, &cm)
+		if err != nil {
+			return "", err
+		}
+		v := cm.Data[e.ValueFrom.ConfigMapKeyRef.Key]
+		return v, nil
+	}
+
+	// TODO: if we just need to check that a variable _is_ a secretRef, fieldRef, and resourceFieldRef,
```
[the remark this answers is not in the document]
ozdanborne: Yeah, I think we'll need to tackle it when we add the check for say NodeName. Should be trivial, but don't want to add an unused function yet...

pkg/controller/migration/convert/network.go
ozdanborne: I went with `specification of IPPools via cni config is not supported.` WDTY?

pkg/controller/migration/convert/network.go
ozdanborne: Good point. I adjusted the message and wrapped that context from the caller :+1:

pkg/controller/migration/convert/network.go
tmjd: I'm just trying to think what would be the most understandable for a user. WDYT about including `ipv4_pool` in the message. Maybe we could just say "specifying ipv4_pools in cni config is not supported"?

PR #699, opened 2020-07-14
Add new CNI Spec and Type to Installation

tmjd:
## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


pkg/controller/installation/validation.go
```diff
@@ -145,6 +148,37 @@ func validateCustomResource(instance *operatorv1.Installation) error {
...
 
+	switch instance.Spec.CNI.Type {
+	case operatorv1.PluginCalico:
+	case operatorv1.PluginGKE:
+		switch instance.Spec.KubernetesProvider {
+		case operatorv1.ProviderEKS:
+			fallthrough
+		case operatorv1.ProviderAKS:
+			return fmt.Errorf("kubernetesProvider of %s is invalid with CNI.type %s",
+				instance.Spec.KubernetesProvider, instance.Spec.CNI.Type)
```
ozdanborne, line 159: Neat. I've never seen a `fallthrough` before.

If I understand it correctly, doesn't this do the same as the following (in an arguably less-clear way?)
```suggestion
		case operatorv1.ProviderEKS, operatorv1.ProviderAKS:
			return fmt.Errorf("kubernetesProvider of %s is invalid with CNI.type %s",
				instance.Spec.KubernetesProvider, instance.Spec.CNI.Type)
```

pkg/render/config.go
```diff
@@ -24,7 +24,41 @@ const (
...
+	}
+
+	// Set other provider-specific settings.
+	switch install.Spec.KubernetesProvider {
+	case operatorv1.ProviderDockerEE:
+		networkConfig.NodenameFileOptional = true
+	}
+
+	// If CalicoNetwork is specified, then use Calico networking.
+	if install.Spec.CalicoNetwork == nil {
```
ozdanborne, line 47: I'm confused, the comment says `If CalicoNetwork is specified`, but then the function checks if it's nil...

pkg/controller/installation/validation.go
tmjd: Right, your way is clearer. Sorry I think my C++ background is showing.

pkg/controller/installation/validation.go
```diff
@@ -145,6 +148,37 @@ func validateCustomResource(instance *operatorv1.Installation) error {
...
+		switch instance.Spec.KubernetesProvider {
+		case operatorv1.ProviderGKE:
+			fallthrough
+		case operatorv1.ProviderEKS:
+			return fmt.Errorf("kubernetesProvider of %s is invalid with CNI.type %s",
+				instance.Spec.KubernetesProvider, instance.Spec.CNI.Type)
+		}
+	default:
+		return fmt.Errorf("Invalid CNI.type %s, it should be one of %s",
+			instance.Spec.CNI.Type, strings.Join(operatorv1.CNIPluginTypesString, ","))
```
ozdanborne, line 179: if this is the only place `CNIPluginTypesString` is used, why not just put the actual strings here?

pkg/apis/operator/v1/types.go
```diff
@@ -312,6 +316,47 @@ type IPPool struct {
...
+type CNIPluginType string
+
+const (
+	PluginCalico    CNIPluginType = "Calico"
+	PluginGKE       CNIPluginType = "GKE"
+	PluginAmazonVPC CNIPluginType = "AmazonVPC"
+	PluginAzureVNET CNIPluginType = "AzureVNET"
+)
+
+var CNIPluginTypes []CNIPluginType = []CNIPluginType{
```
ozdanborne, line 329: Is this used anywhere?

pkg/render/config.go
tmjd: I'm not sure what I was thinking here. I'm removing that if and the associated code block.

pkg/controller/installation/validation.go
tmjd: Because then the strings to will be defined 2 places, once where the type is defined and then also here.
With the current way it will be more obvious that a change should be made when adding a new PluginType. If we just listed them here then when adding a new PluginType it wouldn't be obvious that this code should be updated too.

pkg/apis/operator/v1/types.go
tmjd: I'm not sure if it is. But perhaps I should use it to validate the PluginType.

pkg/render/config.go
tmjd: I found why that was needed. Need to skip the rest of the processing if no CalicoNetwork exists.

ozdanborne: Ok.  One other concern I have about this function:

Based on the single switch case, it looks like we intend on this function growing as we add more CNI plugins. And it seems a bit fragile that developers need to know whether to add that logic before or after this nil check / return depending on the situation. The comment you've added does help, but worth considering if we can make this less fragile. Not sure what I'd suggest to accomplish that, so I'm fine merging as is and we can re-evaluate as those new cases appear.

tmjd: I could switch the check to `if install.Spec.CalicoNetwork != nil` and then move all the remainder of the code into the code block.

PR #700, opened 2020-07-14
KUBERNETES_SERVICE_HOST/PORT override

tomastigera:
If there is no kube-proxy (or it is to be removed), e.g. when we enable
    BPF, we need to override the KUBERNETES_SERVICE_HOST/PORT for operator
    so that it can access k8s. The same setting needs to be forwarded to
    other components like felix, typha and cni.

    There are various ways how to override the variables, using env, envFrom using
     optional config map or just setting the variables straight in the env.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/common.go
+++ b/pkg/render/common.go
@@ -371,3 +371,19 @@ func basePodSecurityPolicy() *policyv1beta1.PodSecurityPolicy {
 		},
 	}
 }
+
+type K8sServiceEndpoint struct {
+	Host string
+	Port string
+}
+
+func (k8s K8sServiceEndpoint) EnvVars() []v1.EnvVar {
+	if k8s.Host == "" || k8s.Port == "" {
+		return nil
+	}
+
+	return []v1.EnvVar{
+		{Name: "KUBERNETES_SERVICE_HOST", Value: k8s.Host},
+		{Name: "KUBERNETES_SERVICE_PORT", Value: k8s.Port},
+	}
+}
--- a/pkg/render/common.go
+++ b/pkg/render/common.go
@@ -372,11 +372,15 @@ func basePodSecurityPolicy() *policyv1beta1.PodSecurityPolicy {
 	}
 }
 
+// K8sServiceEndpoint is the Host/Port of the K8s endpoint.
 type K8sServiceEndpoint struct {
 	Host string
 	Port string
 }
 
+// EnvVars returns a slice of v1.EnvVars KUBERNETES_SERVICE_HOST/PORT if the Host and Port
+// of the K8sServiceEndpoint were set. It returns a nil slice if either was empty as both
+// need to be set.
 func (k8s K8sServiceEndpoint) EnvVars() []v1.EnvVar {
 	if k8s.Host == "" || k8s.Port == "" {
 		return nil
--- a/pkg/render/node.go
+++ b/pkg/render/node.go
@@ -17,7 +17,6 @@ package render
 import (
 	"fmt"
 	"net"
-	"os"
 	"strconv"
 
 	operator "github.com/tigera/operator/pkg/apis/operator/v1"
@@ -53,49 +52,32 @@ var (
 	nodeBGPReporterPort int32 = 9900
 )
 
-func getK8sEndpoint() (string, string, error) {
-	host := os.Getenv("KUBERNETES_SERVICE_HOST")
-	port := os.Getenv("KUBERNETES_SERVICE_PORT")
-
-	if host == "" || port == "" {
-		return "", "", fmt.Errorf("k8s host and/or port override env vars empty")
-	}
-
-	return host, port, nil
-}
-
 // Node creates the node daemonset and other resources for the daemonset to operate normally.
 func Node(
+	k8sServiceEp K8sServiceEndpoint,
 	cr *operator.Installation,
 	bt map[string]string,
 	tnTLS *TyphaNodeTLS,
 	aci *operator.AmazonCloudIntegration,
 	migrate bool,
 ) Component {
-	node := &nodeComponent{
+	return &nodeComponent{
+		k8sServiceEp:    k8sServiceEp,
 		cr:              cr,
 		birdTemplates:   bt,
 		typhaNodeTLS:    tnTLS,
 		amazonCloudInt:  aci,
 		migrationNeeded: migrate,
 	}
-
-	if k8sHost, k8sPort, err := getK8sEndpoint(); err == nil {
-		node.k8sHost = k8sHost
-		node.k8sPort = k8sPort
-	}
-
-	return node
 }
 
 type nodeComponent struct {
+	k8sServiceEp    K8sServiceEndpoint
 	cr              *operator.Installation
 	birdTemplates   map[string]string
 	typhaNodeTLS    *TyphaNodeTLS
 	amazonCloudInt  *operator.AmazonCloudIntegration
 	migrationNeeded bool
-	k8sHost         string
-	k8sPort         string
 }
 
 func (c *nodeComponent) Objects() ([]runtime.Object, []runtime.Object) {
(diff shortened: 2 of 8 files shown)


tmjd:
Why do we use _OVERRIDE_ env vars instead of just using the standard KUBERNETES_SERVICE_HOST (and PORT)? Is that so the operator can tell if it is using overridden values and needs to pass them on?
Don't the same changes that were made to node.go need to be made to typha.go?

tomastigera:
> Why do we use _OVERRIDE_ env vars instead of just using the standard KUBERNETES_SERVICE_HOST (and PORT)? Is that so the operator can tell if it is using overridden values and needs to pass them on?

Hmmm one of the reasons was to be able to see if the override wasn't set if operators was enabling bpf. But since we dont do it anymore, we might just use the KUBERNETES_SERVICE_HOST/PORT, will  try it that way.

> Don't the same changes that were made to node.go need to be made to typha.go?

good point, will do

pkg/render/node.go
```diff
@@ -59,14 +71,21 @@ func Node(
...
+	node := &nodeComponent{
 		cr:              cr,
 		netConfig:       nc,
 		birdTemplates:   bt,
 		typhaNodeTLS:    tnTLS,
 		amazonCloudInt:  aci,
 		migrationNeeded: migrate,
 	}
+
+	if k8sHost, k8sPort, err := getK8sEndpoint(); err == nil {
```
caseydavenport, line 83: I'm not a fan of adding this dependency on the KUBERNETES_SERVICE_HOST / PORT environment variables to the render code (and really the operator in general), since it's totally valid for them to not be present in some scenarios. 

As an example, a pretty standard way of testing the operator code during development is something like this:

```
KUBECONFIG=./kubeconfig.yaml go run ./cmd/manager
```

Could we use the auto-configured kube-client in order to extract the configured API endpoint and pass it to render instead? Would that work?

caseydavenport: Maybe me you and @tmjd should hop on a call to get alignment on what we want to do here.

tomastigera: > As an example, a pretty standard way of testing the operator code during development is something like this:
> 
> ```
> KUBECONFIG=./kubeconfig.yaml go run ./cmd/manager
> ```

That scenario works. You need to set all 3. KUBECONFIG takes precedence and KUBERNETES_SERVICE_HOST/PORT are just forwarded. If it is in-cluster, KUBECONFIG (afaict) is not set and thus the operator's client uses the override as well.

caseydavenport: My point is that we previously did not have a dependency on the KUBERNETES_SERVICE_HOST and KUBERNETES_SERVICE_PORT environment variables.

If you ran the code without them set, then it still worked. With this change, if those env vars are not set, then the code will error. I'm not yet convinced that's an OK change. Does that make sense?

caseydavenport: > If it is in-cluster, KUBECONFIG (afaict) is not set and thus the operator's client uses the override as well.

It would totally be possible for someone to:
- Mount in a kubeconfig file
- Set the KUBECONFIG environment variable. 

With this change, that workflow is no longer possible (not that I'm aware of anyone using it that way necessarily, but trying to show how this is a breaking change).

tomastigera: > If you ran the code without them set, then it still worked. With this change, if those env vars are not set, then the code will error. I'm not yet convinced that's an OK change. Does that make sense?

Then the forward values are not set and felix may not work correctly if it cannot reach k8s. In a cluster with kube-proxy that is OK as felix will use whatever values it has set. In a cluster without kube-proxy, e.g. when enabling bpf, operator will not be able to connect to k8s without this being set properly. That is a requirement.

We can revert to using the OVERRIDE vars. Is that what you suggest?

tmjd:
I think this PR needs to be rebased on master to fix the tests.

tmjd:
I think the changes here look good. The one thing I would consider is if we should get those env vars in an init function so we're not grabbing them on each Reconcile. I think reading them each time is probably fine but they won't be changing while running so it would also work to read them one time at startup.

pkg/render/common.go
```diff
@@ -371,3 +371,23 @@ func basePodSecurityPolicy() *policyv1beta1.PodSecurityPolicy {
...
+type K8sServiceEndpoint struct {
+	Host string
+	Port string
+}
+
+// EnvVars returns a slice of v1.EnvVars KUBERNETES_SERVICE_HOST/PORT if the Host and Port
+// of the K8sServiceEndpoint were set. It returns a nil slice if either was empty as both
+// need to be set.
+func (k8s K8sServiceEndpoint) EnvVars() []v1.EnvVar {
+	if k8s.Host == "" || k8s.Port == "" {
```
caseydavenport, line 385: Wonder if we should be doing some sort of validation on these before we pass them through to the individual components.

Seems like we should add logic to check for scenarios like this:

- An invalid port number of host is provided
- One field is provided but not the other

WDYT? We'd probably want to add that validation to core_controller.go

pkg/render/node_test.go
```diff
@@ -45,6 +45,8 @@ var _ = Describe("Node rendering tests", func() {
 	one := intstr.FromInt(1)
 	defaultNumExpectedResources := 6
 
+	k8s := render.K8sServiceEndpoint{}
```
caseydavenport, line 48: just a nit, but can we call this something other than "k8s" - it's not a particularly descriptive name, and lots of stuff here is relevant to Kubernetes!

Other options that come to mind:

- kubernetesServiceEndpoint
- k8sAPIInfo
- apiServerLocation
- ? ? ?

pkg/render/common.go
tomastigera: > * An invalid port number of host is provided

I though about that, but can we do anything beyond "the value provided to port is a 16bit number" ? Or "is HOST parsable IP"? What if it is a DN?

> * One field is provided but not the other

It requires both to be passed on. Maybe can can do the test in the `init()` and log that one is missing and thus not passing anything on :+1:

pkg/render/node_test.go
tomastigera: sure k8sServiceEp ?

pkg/render/common.go
caseydavenport: > Maybe can can do the test in the init() and log that one is missing and thus not passing anything on

I don't think it belongs in the init. We have an estalished pattern for doing validation and reporting it to the user via a TigeraStatus message (see SetDegraded) - we should use that same mechansim. 

> I though about that, but can we do anything beyond "the value provided to port is a 16bit number" ? Or "is HOST parsable IP"? What if it is a DN?

Yep, this is still better than nothing. We should be able to parse URL provide to make sure it is a valid IP or domain name.

tomastigera: I added the validation, but note that if the provided values are wrong, operator will not work correctly, will not be able to connect. I think it will fail sooner  than these validations happen. It will appear in its logs as "cannot connect ..." I am still not sure that hardening it here adds much value except if the operator runs from outside the cluster or uses kubeconfig.

caseydavenport: > I added the validation, but note that if the provided values are wrong, operator will not work correctly, will not be able to connect. I think it will fail sooner than these validations happen

oof, yes you are correct.

tomastigera: i am inclined to revert the last commit with the validations and assume that if operator can use these values for its own connection, then it is all OK. If not, the error will be visible with operator's pod `describe pod`

caseydavenport: @tomastigera that sounds good to me. Sorry for leading you astray.

tomastigera: ok, no worries, we would not know if we did not do the excercise :-) I removed the validations and rebased the code on master.

PR #703, opened 2020-07-15
convert config that is configurable via operator api

ozdanborne:
## Description

This includes (see individual commits):
- mem / cpu limits / requests for node, kubeControllers, & typha
- flexvolpath
- node updatestrategy
- nodeselector (kubecontrollers)

Also included: made the converter test apart of the package so the handler tests (which test private methods) could use the the helper methods provided by the converter tests.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/migration/convert/core.go
+++ b/pkg/controller/migration/convert/core.go
@@ -46,6 +46,20 @@ func handleCore(c *components, install *Installation) error {
 	// node update-strategy
 	install.Spec.NodeUpdateStrategy = c.node.Spec.UpdateStrategy
 
+	// node volumes
+	for _, vol := range c.node.Spec.Template.Spec.Volumes {
+		switch vol.Name {
+		case "flexvol-driver-host":
+			// prefer user-defined flexvolpath over detected value
+			if install.Spec.FlexVolumePath == "" {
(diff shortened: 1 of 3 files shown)


pkg/controller/migration/convert/core.go
```diff
@@ -9,6 +13,50 @@ func handleCore(c *components, install *Installation) error {
...
+	for _, vol := range c.node.Spec.Template.Spec.Volumes {
+		switch vol.Name {
+		case "flexvol-driver-host":
+			if vol.HostPath == nil {
+				return ErrIncompatibleCluster{"volume 'flexvol-driver-host' must be a HostPath"}
+			}
+			install.Spec.FlexVolumePath = vol.HostPath.Path
+		}
+	}
+
```
tmjd, line 59: I think we should check if install.Spec.FlexVolumePath is not set at this point then we should set it to none. Otherwise it will get defaulted later.

tmjd: We might also want to verify that the flexvol init container doesn't exist in the case when we set FlexVolumePath to None.

PR #717, opened 2020-07-16
Fix being able to specify CNI plugin with no Provider

tmjd:
## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/installation/core_controller.go
+++ b/pkg/controller/installation/core_controller.go
@@ -300,11 +300,6 @@ func fillDefaults(instance *operator.Installation) error {
 			msg := "Installation spec.calicoNetwork must not be set for provider %s"
 			return fmt.Errorf(msg, instance.Spec.KubernetesProvider)
 		}
-	default:
-		if instance.Spec.CalicoNetwork == nil {
-			// For all other platforms, default to using Calico networking.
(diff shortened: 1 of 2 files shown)


pkg/controller/installation/core_controller.go
```diff
@@ -882,6 +882,10 @@ func isOpenshiftOnAws(install *operator.Installation, ctx context.Context, clien
 }
 
 func updateInstallationForOpenshiftNetwork(i *operator.Installation, o *configv1.Network) error {
+	// If CNI plugin is specified and not Calico then skip any CalicoNetwork initialization
+	if i.Spec.CNI != nil && i.Spec.CNI.Type != operator.PluginCalico {
```
caseydavenport, line 886: I don't think this is really correct - we sometimes want a CalicoNetworkSpec even if not using Calico CNI.

caseydavenport: If you recall, this was one of my concerns about this particular expression of the API - `CalicoNetwork` != `cni.type:Calico`

tmjd: With only the CNI plugin being able to be specified right now I don't think there are any cases where we'd want CalicoNetwork. Did we allow CalicoNetwork to be specified with the EKS, AKS, or GKE providers?

PR #719, opened 2020-07-16
Allow enabling / disabling BGP

caseydavenport:
## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


pkg/controller/installation/defaults_test.go
```diff
@@ -121,6 +129,36 @@ var _ = Describe("Defaulting logic tests", func() {
...
+		instance := &operator.Installation{
+			Spec: operator.InstallationSpec{
+				CNI: &operator.CNISpec{
+					Type: operator.PluginAmazonVPC,
+				},
+				CalicoNetwork: &operator.CalicoNetworkSpec{},
+			},
+		}
+		fillDefaults(instance)
+		Expect(*instance.Spec.CalicoNetwork.BGP).To(Equal(operator.BGPDisabled))
```
tmjd, line 158: I don't think fillDefaults should be creating an invalid configuration. This seems like a problem.

pkg/render/node.go
```diff
@@ -891,49 +853,81 @@ func (c *nodeComponent) nodeEnvVars() []v1.EnvVar {
...
+	if c.cr.Spec.CalicoNetwork.MTU != nil {
+		ipipMtu = strconv.Itoa(int(*c.cr.Spec.CalicoNetwork.MTU))
+		vxlanMtu = strconv.Itoa(int(*c.cr.Spec.CalicoNetwork.MTU))
+		wireguardMtu = strconv.Itoa(int(*c.cr.Spec.CalicoNetwork.MTU))
+	}
+	nodeEnv = append(nodeEnv, v1.EnvVar{Name: "FELIX_VXLANMTU", Value: vxlanMtu})
+	nodeEnv = append(nodeEnv, v1.EnvVar{Name: "FELIX_WIREGUARDMTU", Value: wireguardMtu})
+
+	// Configure whether or not BGP should be enabled.
+	if !bgpEnabled(c.cr) {
```
tmjd, line 889: I think we should have this wrapped in a check for `if c.cr.Spec.CNI.Type == operator.PluginCalico` and set CALICO_NETWORKING_BACKEND=none if not.

pkg/render/node.go
```diff
@@ -891,49 +853,81 @@ func (c *nodeComponent) nodeEnvVars() []v1.EnvVar {
...
+		// not actually using Calico VXLAN for networking (e.g., via an alternative CNI plugin).
+		nodeEnv = append(nodeEnv, v1.EnvVar{Name: "CALICO_NETWORKING_BACKEND", Value: "vxlan"})
+	} else {
+		// BGP is enabled.
+		nodeEnv = append(nodeEnv, v1.EnvVar{Name: "CALICO_NETWORKING_BACKEND", Value: "bird"})
+		nodeEnv = append(nodeEnv, v1.EnvVar{Name: "FELIX_IPINIPMTU", Value: ipipMtu})
+	}
+
+	// IPv4 auto-detection configuration.
+	v4Method := getAutodetectionMethod(c.cr.Spec.CalicoNetwork.NodeAddressAutodetectionV4)
```
tmjd: The getAutodetectionMethod calls need to be protected with `CalicoNetwork != nil`

pkg/controller/installation/defaults_test.go
tmjd: I see that the initial instance is invalid currently too, so I guess that's fine. I think we should include the Expect below and just have it expecting an error for now. Then when the validation is changed this test will fail and get updated.

pkg/controller/installation/validation.go
```diff
@@ -92,24 +92,24 @@ func validateCustomResource(instance *operatorv1.Installation) error {
...
 
-			valid := false
-			for _, t := range operatorv1.EncapsulationTypes {
-				if v4pool.Encapsulation == t {
-					valid = true
+			// Verify the specified encapsulation type is valid.
+			switch v4pool.Encapsulation {
+			case operatorv1.EncapsulationIPIP, operatorv1.EncapsulationIPIPCrossSubnet:
+				// IPIP currently requires BGP to be running in order to program routes.
+				if instance.Spec.CalicoNetwork.BGP == nil || *instance.Spec.CalicoNetwork.BGP == operatorv1.BGPDisabled {
```
tmjd, line 99: Since we always default BGP in CalicoNetwork, the nil check isn't really needed.

pkg/controller/installation/validation.go
```diff
@@ -92,24 +92,24 @@ func validateCustomResource(instance *operatorv1.Installation) error {
...
+					return fmt.Errorf("IPIP encapsulation requires that BGP is enabled")
 				}
-			}
-			if !valid {
+			case operatorv1.EncapsulationVXLAN, operatorv1.EncapsulationVXLANCrossSubnet:
+			case operatorv1.EncapsulationNone:
```
tmjd, line 103: Since it is currently invalid to have non-Calico CNI and CalicoNetwork, the case of EncapsulationNone also needs to have BGP enabled, I think.

tmjd: This might change soon with adding IPAM HostLocal as an option.

pkg/controller/installation/defaults_test.go
caseydavenport: Yeah, exactly. `fillDefaults` is fine, it's the initial configuration that we're passing in that's invalid. 

I'll fix the assertion.

pkg/controller/installation/validation.go
caseydavenport: yeah, but in case there is a bug this is a better failure mode than a panic

pkg/render/node.go
caseydavenport: > and set CALICO_NETWORKING_BACKEND=none if not.

This is intentionally VXLAN even for non-calico CNI plugins in this PR, since there is a known issue that node IPs are not populated on the node object if set to "none".

I guess this is really just a workaround for a bug in another part of the code, though.

tmjd: Does that mean there is a problem with other CNI plugins currently that have that value set to none? Or is this only an issue for the Encryption case where we want to use wireguard with a other CNI plugins?

PR #730, opened 2020-07-20
fix bug where detected config fails

ozdanborne:
## Description

Removes the call to `DeepCopyInto`. `DeepCopyInto` works well with pointer fields because it checks if the thing you are copying from is nil before copying. But for non nil, it will copy no matter what. This manifested itself into non-nil fields which the user has set in their Installation beiing overwritten with whatever the default empty value is. Specifically, the `Variant` was being overwritten to Calico.

This PR refactors how the installation is manipualted by the conversion code. Instead, a pointer to the user-configured install is passed into the converter.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/migration/convert/convert.go
+++ b/pkg/controller/migration/convert/convert.go
@@ -26,7 +26,7 @@ var ctx = context.Background()
 
 // Installation represents the configuration pulled from the existing install.
 type Installation struct {
-	operatorv1.Installation
+	*operatorv1.Installation
 
 	// The following fields are not yet exposed in the
 	// operator API, and serve as a temporary store during prototyping.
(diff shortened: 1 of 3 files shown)


pkg/controller/migration/convert/convert.go
```diff
@@ -103,35 +103,30 @@ func getComponents(ctx context.Context, client client.Client) (*components, erro
...
 	}
 
 	// check for unchecked env vars
 	if uncheckedVars := comps.node.uncheckedVars(); len(uncheckedVars) != 0 {
-		return nil, ErrIncompatibleCluster{fmt.Sprintf("unexpected env var: %s", uncheckedVars)}
+		return ErrIncompatibleCluster{fmt.Sprintf("unexpected env var: %s", uncheckedVars)}
 	}
 
-	return &config.Installation, nil
+	return nil
```
tmjd, line 131: I think we should have an explicit step here that merges an Installation that we generate from the handlers with the Installation that is passed into this function.
I think it will be problematic to try to ensure every place in the handlers that touch a field will validate that the field is not already set.

PR #734, opened 2020-07-21
cni-plugin moved to scratch

gianlucam76:
cni-plugin has replaced install-cni.sh with install go binary


doublek:
@gianlucam76 perhaps this should be closed?

PR #735, opened 2020-07-21
convert config that is configurable via operator api

ozdanborne:
## Description

This is a reopening of https://github.com/tigera/operator/pull/703 which I accidentally temporarily merged

This includes (see individual commits):
- mem / cpu limits / requests for node, kubeControllers, & typha
- flexvolpath
- node updatestrategy
- nodeselector (kubecontrollers)

Also included: made the converter test apart of the package so the handler tests (which test private methods) could use the the helper methods provided by the converter tests.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/migration/convert/core.go
+++ b/pkg/controller/migration/convert/core.go
@@ -46,6 +46,20 @@ func handleCore(c *components, install *Installation) error {
 	// node update-strategy
 	install.Spec.NodeUpdateStrategy = c.node.Spec.UpdateStrategy
 
+	// node volumes
+	for _, vol := range c.node.Spec.Template.Spec.Volumes {
+		switch vol.Name {
+		case "flexvol-driver-host":
+			// prefer user-defined flexvolpath over detected value
+			if install.Spec.FlexVolumePath == "" {
+				if vol.HostPath == nil {
+					return ErrIncompatibleCluster{"volume 'flexvol-driver-host' must be a HostPath"}
+				}
+				install.Spec.FlexVolumePath = vol.HostPath.Path
+			}
+		}
+	}
+
 	// TODO: handle these vars appropriately
 	c.node.ignoreEnv("calico-node", "WAIT_FOR_DATASTORE")
 	c.node.ignoreEnv("calico-node", "CLUSTER_TYPE")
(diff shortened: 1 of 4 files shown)


pkg/controller/migration/convert/core.go
```diff
@@ -9,6 +13,73 @@ func handleCore(c *components, install *Installation) error {
...
+	}
+
+	// kube-controllers nodeSelector
+	install.Spec.ControlPlaneNodeSelector = c.kubeControllers.Spec.Template.Spec.NodeSelector
+
+	// node update-strategy
+	install.Spec.NodeUpdateStrategy = c.node.Spec.UpdateStrategy
+
+	// node volumes
+	for _, vol := range c.node.Spec.Template.Spec.Volumes {
```
tmjd, line 50: There seem to be 2 separate processing of the Volumes for flexvol-driver-host here. Is that needed?

pkg/controller/migration/convert/core.go
```diff
@@ -9,6 +13,73 @@ func handleCore(c *components, install *Installation) error {
...
+			if vol.HostPath == nil {
+				return ErrIncompatibleCluster{"volume 'flexvol-driver-host' must be a HostPath"}
+			}
+			if fv := getContainer(c.node.Spec.Template.Spec, "flexvol-driver"); fv == nil {
+				return ErrIncompatibleCluster{"detected 'flexvol-driver-host' volume but no 'flexvol-driver' init container"}
+			}
+			install.Spec.FlexVolumePath = vol.HostPath.Path
+		}
+	} else {
+		// verify that no flexvol container is set
```
tmjd, line 77: I think here we should be setting `install.Spec.FlexVolumePath = None`. That's how we disable it with the operator installation.

tmjd: I think this branch should end with setting `install.Spec.FlexVolumePath = None`. It is not sufficient to leave it unset because when our defaulting logic runs we would set a default path.

pkg/controller/migration/convert/core.go
ozdanborne: oops, meant to remove the first one when i added the second.

PR #742, opened 2020-07-22
add v1beta1 to crd definition

ozdanborne:
## Description

Seems k8s doesn't like it when a CRD drops an api-version. We could have all users `kubectl delete crd amazoncloudintegration.tigera.io`, but that's a pain. Instead, since there's only a couple of users using this CRD, we can just do a half-baked migration where we include the old CRD definition. I think @doucol has tighter upgrade instructions somewhere for   users who are using the v1beta1.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


deploy/crds/operator.tigera.io_amazoncloudintegrations_crd.yaml
```diff
@@ -88,3 +88,80 @@ spec:
...
+          status:
+            description: AmazonCloudIntegrationStatus defines the observed state of
+              AmazonCloudIntegration
+            properties:
+              state:
+                description: State provides user-readable status.
+                type: string
+            type: object
+        type: object
+    served: true
```
tmjd, line 164: What if we changed this to false?

ozdanborne:
@tmjd thanks, great feedback.

I did some experimentation, and your suggestions all work, even when an AmazonCloudIntegration exists. I did have some trouble modifying the AmazonCloudIntegration after applying the new CRD, but I was able to delete and recreate it (which I believe was @doucol 's upgrade plan anyways).

PR #743, opened 2020-07-22
Changes for Calico networking on EKS

caseydavenport:
## Description

- Make defaults for EKS w/ Calico CNI more sensible.
- Add workaround for controller->pod connections for API server

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/installation/core_controller.go
+++ b/pkg/controller/installation/core_controller.go
@@ -340,20 +340,39 @@ func fillDefaults(instance *operator.Installation) error {
 		// Default IP pools, only if it is nil. If it is an empty slice then that
 		// means no default IPPools should be created.
 		if instance.Spec.CalicoNetwork.IPPools == nil {
-			instance.Spec.CalicoNetwork.IPPools = []operator.IPPool{
-				{CIDR: "192.168.0.0/16"},
+			switch instance.Spec.KubernetesProvider {
+			case operator.ProviderEKS:
+				// On EKS, default to a CIDR that doesn't overlap with the host range,
+				// and also use VXLAN encap by default.
+				instance.Spec.CalicoNetwork.IPPools = []operator.IPPool{
+					{
+						CIDR:          "172.16.0.0/16",
+						Encapsulation: operator.EncapsulationVXLAN,
+					},
+				}
+			default:
+				instance.Spec.CalicoNetwork.IPPools = []operator.IPPool{
+					{CIDR: "192.168.0.0/16"},
+				}
 			}
 		}
 
-		// Default BGP enablement based on CNI plugin. We default to BGP
-		// enabled for Calico CNI, but disabled for all others.
+		// Default BGP enablement based on CNI plugin and provider.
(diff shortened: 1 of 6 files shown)


pkg/controller/installation/core_controller.go
```diff
@@ -324,20 +324,39 @@ func fillDefaults(instance *operator.Installation) error {
 		// Default IP pools, only if it is nil. If it is an empty slice then that
 		// means no default IPPools should be created.
 		if instance.Spec.CalicoNetwork.IPPools == nil {
-			instance.Spec.CalicoNetwork.IPPools = []operator.IPPool{
-				{CIDR: "192.168.0.0/16"},
+			switch instance.Spec.KubernetesProvider {
+			case operator.ProviderEKS:
+				// On EKS, default to a CIDR that doens't overlap with the host range,
```
bcreane, line 329: typo - `doesn't`

pkg/render/node_test.go
```diff
@@ -402,6 +402,240 @@ var _ = Describe("Node rendering tests", func() {
...
+		i := 0
+		for _, expectedRes := range expectedResources {
+			ExpectResource(resources[i], expectedRes.name, expectedRes.ns, expectedRes.group, expectedRes.version, expectedRes.kind)
+			i++
+		}
+
+		cniCmResource := GetResource(resources, "cni-config", "calico-system", "", "v1", "ConfigMap")
+		Expect(cniCmResource).ToNot(BeNil())
+		cniCm := cniCmResource.(*v1.ConfigMap)
+		Expect(cniCm.Data["config"]).To(Equal(`{
```
tmjd, line 440: We should use `MatchJSON` instead of `Equal` to make our tests not care about whitespace in the config. (I've got another PR where I switched the others to this.)

pkg/render/apiserver.go
```diff
@@ -541,6 +541,15 @@ rules:
 func (c *apiServerComponent) apiServer() *appsv1.Deployment {
 	var replicas int32 = 1
 
+	hostNetwork := false
+	if c.installation.Spec.KubernetesProvider == operator.ProviderEKS &&
+		c.installation.Spec.CNI.Type == operator.PluginCalico {
+		// Workaround the fact that webhooks don't work for non-host-networked pods
+		// when in this networking mode on EKS, because the control plane nodes don't run
+		// Calico.
+		hostNetwork = true
```
tmjd, line 550: Should we have a field in the API that indicates this?

caseydavenport: I don't think so - this is really a workaround for a limitation in the combination of:
- How EKS control plane does not allow pod scheduling
- How Calico networking doesn't make pod IPs routable to instances outside of the cluster.

It's not something that can / should be user-configurable, it's just required to make the user's provided intent "work".

pkg/render/node_test.go
caseydavenport: oo that's neat

PR #744, opened 2020-07-22
Pass OIDC prefixes from Authentication CR to KubeControllers

Brian-McM:
## For PR author

- [ ] Tests for change.

## For PR reviewers

- [x] Appropriate labels:

### DIFF
--- a/pkg/controller/installation/core_controller.go
+++ b/pkg/controller/installation/core_controller.go
@@ -179,6 +179,11 @@ func add(mgr manager.Manager, r *ReconcileInstallation) error {
 		if err != nil {
 			return fmt.Errorf("tigera-installation-controller failed to watch primary resource: %v", err)
 		}
+
+		err = c.Watch(&source.Kind{Type: &operator.Authentication{}}, &handler.EnqueueRequestForObject{})
+		if err != nil {
+			return fmt.Errorf("tigera-installation-controller failed to watch primary resource: %v", err)
+		}
 	}
 
 	return nil
@@ -566,8 +571,16 @@ func (r *ReconcileInstallation) Reconcile(request reconcile.Request) (reconcile.
 		render.ManagerInternalSecretKeyName,
 	)
 
+	logStorageExists, err := utils.LogStorageExists(ctx, r.client)
+	if err != nil {
+		log.Error(err, "Error checking if LogStorage exists")
+		r.SetDegraded("Error checking if LogStorage exists", err, reqLogger)
+		return reconcile.Result{}, err
+	}
+
 	var managementCluster *operator.ManagementCluster
 	var managementClusterConnection *operator.ManagementClusterConnection
+	var authentication interface{}
 	if r.enterpriseCRDsExist {
 		managementCluster, err = utils.GetManagementCluster(ctx, r.client)
 		if managementCluster != nil {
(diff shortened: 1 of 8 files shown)


pkg/controller/installation/core_controller.go
```diff
@@ -640,6 +645,13 @@ func (r *ReconcileInstallation) Reconcile(request reconcile.Request) (reconcile.
 		}
 	}
 
+	authentication, err := utils.GetAuthentication(ctx, r.client)
```
matthewdupre, line 648: I would have expected to see some kind of if authentication CRD exists here (perhaps like the AWSCloud one above).  I guess GetAuthentication() returns the default one?

pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -227,6 +224,11 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 		return fmt.Errorf("log-storage-controller failed to watch primary resource: %v", err)
 	}
 
+	err = c.Watch(&source.Kind{Type: &operatorv1.Authentication{}}, &handler.EnqueueRequestForObject{})
```
matthewdupre, line 227: We're always including Authentication in Calico Enterprise installs right?

pkg/controller/utils/utils.go
```diff
@@ -273,3 +276,63 @@ func GetAmazonCloudIntegration(ctx context.Context, client client.Client) (*oper
 
 	return instance, nil
 }
+
+// getAuthentication
+func GetAuthentication(ctx context.Context, client client.Client) (interface{}, error) {
+	var authenticationConfig interface{}
+
+	authenticationCR := &operatorv1.Authentication{}
```
matthewdupre, line 284: Does this default everything correctly?  (It gets returned like this in the not found case.)

pkg/controller/logstorage/logstorage_controller.go
Brian-McM: Yes

pkg/controller/utils/utils.go
Brian-McM: This CR isn't returned, right now the only things that will be returned will be either nil or render.OIDCAuthentication. This just allows use to do a type switch to see if we should add OIDC configuration instead of doing a bunch of checks on the authentication CR.

This also prevents the need to change the function definitions each time we add an auth method. Nothing has really changed here from what was in the logstorage controller, this was just moved the the core controller could use the same logic.

pkg/controller/logstorage/logstorage_controller.go
Brian-McM: yes

pkg/controller/installation/core_controller.go
Brian-McM: GetAuthentication() returns a struct for the authentication method found. It doesn't return the CRD, it parses the authentication CRD, determines the type (OIDC for right now), figures out any missing bits, and returns a struct with everything needed to setup the auth method. For OIDC this includes looking up the oidc secret to retrieve the client id and secret as well as parsing out the values we need from the well known config.

pkg/controller/installation/core_controller.go
```diff
@@ -179,6 +179,11 @@ func add(mgr manager.Manager, r *ReconcileInstallation) error {
 		if err != nil {
 			return fmt.Errorf("tigera-installation-controller failed to watch primary resource: %v", err)
 		}
+
+		err = c.Watch(&source.Kind{Type: &operator.Authentication{}}, &handler.EnqueueRequestForObject{})
+		if err != nil {
+			return fmt.Errorf("log-storage-controller failed to watch primary resource: %v", err)
```
matthewdupre, line 185: s/log-storage/core/

PR #745, opened 2020-07-22
Migrate non-calico CNI plugins

tmjd:
## Description

I'm not confident on using the install-cni init container as the main flag for putting the config migration into non-Calico CNI mode. I had initially used CALICO_NETWORKING_BACKEND but then thought maybe that wasn't good enough. I think another good option would be to use the FELIX_INTERFACEPREFIX.
So I'd like feed back on which to use as the initial indication we're using a non-Calico CNI.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/migration/convert/cni.go
+++ b/pkg/controller/migration/convert/cni.go
@@ -12,6 +12,15 @@ import (
 // loadCNI loads CNI config into the components for all other handlers to use.
 // No verification is done here beyond checking for valid json.
 func loadCNI(c *components) error {
+
+	cntnr := getContainer(c.node.Spec.Template.Spec, containerInstallCNI)
+	if cntnr == nil {
+		// It is valid to not have an install-cni container in the cases
+		// of non Calico CNI so nothing more to do in that case.
+		log.Print("no install-cni container detected on node")
+		return nil
+	}
+
 	cniConfig, err := c.node.getEnv(ctx, c.client, containerInstallCNI, "CNI_NETWORK_CONFIG")
 	if err != nil {
 		return err
--- a/pkg/controller/migration/convert/core.go
+++ b/pkg/controller/migration/convert/core.go
@@ -1,6 +1,10 @@
 package convert
 
 import (
+	"fmt"
+	"strconv"
+	"strings"
+
 	operatorv1 "github.com/tigera/operator/pkg/apis/operator/v1"
 )
 
@@ -24,7 +28,7 @@ func handleCore(c *components, install *Installation) error {
 
 	// kube-controllers
 	kubeControllers := getContainer(c.kubeControllers.Spec.Template.Spec, "calico-kube-controllers")
-	if len(kubeControllers.Resources.Limits) > 0 || len(kubeControllers.Resources.Requests) > 0 {
+	if kubeControllers != nil && (len(kubeControllers.Resources.Limits) > 0 || len(kubeControllers.Resources.Requests) > 0) {
 		install.Spec.ComponentResources = append(install.Spec.ComponentResources, &operatorv1.ComponentResource{
 			ComponentName:        operatorv1.ComponentNameKubeControllers,
 			ResourceRequirements: kubeControllers.Resources.DeepCopy(),
@@ -67,6 +71,10 @@ func handleCore(c *components, install *Installation) error {
 		install.Spec.FlexVolumePath = "None"
 	}
 
+	if err = handleFelixNodeMetrics(c, install); err != nil {
+		return err
+	}
+
 	// TODO: handle these vars appropriately
 	c.node.ignoreEnv("calico-node", "WAIT_FOR_DATASTORE")
 	c.node.ignoreEnv("calico-node", "CLUSTER_TYPE")
@@ -82,11 +90,41 @@ func handleCore(c *components, install *Installation) error {
 	c.node.ignoreEnv("calico-node", "FELIX_HEALTHENABLED")
 	c.node.ignoreEnv("calico-node", "FELIX_USAGEREPORTINGENABLED")
 	c.node.ignoreEnv("calico-node", "FELIX_TYPHAK8SSERVICENAME")
+	c.node.ignoreEnv("calico-node", "FELIX_LOGSEVERITYSYS")
 	c.node.ignoreEnv("upgrade-ipam", "KUBERNETES_NODE_NAME")
 	c.node.ignoreEnv("upgrade-ipam", "CALICO_NETWORKING_BACKEND")
 	c.node.ignoreEnv("install-cni", "CNI_CONF_NAME")
 	c.node.ignoreEnv("install-cni", "KUBERNETES_NODE_NAME")
 	c.node.ignoreEnv("install-cni", "SLEEP")
+
+	return nil
+}
+
+func handleFelixNodeMetrics(c *components, install *Installation) error {
+	metricsEnabled, err := c.node.getEnv(ctx, c.client, containerCalicoNode, "FELIX_PROMETHEUSMETRICSENABLED")
+	if err != nil {
+		return err
+	}
+	if metricsEnabled != nil && strings.ToLower(*metricsEnabled) == "true" {
+		var _9091 int32 = 9091
+		install.Spec.NodeMetricsPort = &_9091
+		port, err := c.node.getEnv(ctx, c.client, containerCalicoNode, "FELIX_PROMETHEUSMETRICSPORT")
+		if err != nil {
+			return err
+		}
+		if port != nil {
+			p, err := strconv.ParseInt(*port, 10, 32)
+			if err != nil || p <= 0 || p > 65535 {
+				return ErrIncompatibleCluster{fmt.Sprintf(
+					"invalid port defined in FELIX_PROMETHEUSMETRICSPORT(%s), it should be 1-65535 ", *port)}
+			}
+			i := int32(p)
+			install.Spec.NodeMetricsPort = &i
+		}
+	} else {
+		// Ignore the metrics port if metrics is disabled.
+		c.node.ignoreEnv(containerCalicoNode, "FELIX_PROMETHEUSMETRICSPORT")
+	}
 
 	return nil
 }
--- a/pkg/controller/migration/convert/network.go
+++ b/pkg/controller/migration/convert/network.go
@@ -17,7 +17,7 @@ const (
 
 func handleNetwork(c *components, install *Installation) error {
 
-	// FELIX_DEFAULTENDPOINTTOHOSTACTION
+	// Verify FELIX_DEFAULTENDPOINTTOHOSTACTION is set to Accept because that is what the operator sets it to.
 	defaultWepAction, err := c.node.getEnv(ctx, c.client, containerCalicoNode, "FELIX_DEFAULTENDPOINTTOHOSTACTION")
 	if err != nil {
 		return err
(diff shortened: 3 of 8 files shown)


pkg/controller/migration/convert/cni.go
```diff
@@ -14,6 +14,12 @@ import (
 func loadCNI(c *components) error {
 	cniConfig, err := c.node.getEnv(ctx, c.client, containerInstallCNI, "CNI_NETWORK_CONFIG")
 	if err != nil {
+		if IsContainerNotFound(err) {
```
ozdanborne, line 17: ok, but couldn't you just call `getContainer` first instead of introducing this error type?

pkg/controller/migration/convert/core.go
```diff
@@ -9,6 +15,10 @@ func handleCore(c *components, install *Installation) error {
 		return ErrIncompatibleCluster{"only DATASTORE_TYPE=kubernetes is supported at this time"}
 	}
 
+	if err = handleFelixNodeMetrics(c, install); err != nil {
```
ozdanborne, line 18: I don't think we want handlers calling other handlers. Can you move this up to where `handleCore` is called from?

pkg/controller/migration/convert/network.go
```diff
@@ -16,34 +16,46 @@ const (
...
+		return ErrIncompatibleCluster{
+			fmt.Sprintf("unexpected FELIX_DEFAULTENDPOINTTOHOSTACTION: '%s'. Only 'accept' is supported.", *defaultWepAction),
+		}
+	}
+
+	if c.calicoCNIConfig != nil {
+		return handleCalicoCNI(c, install)
+	} else {
+		return handleNoCNIConfiguration(c, install)
+	}
```
ozdanborne, line 35: again, I think this belongs up a level where handleNetwork is called.

pkg/controller/migration/convert/network.go
```diff
@@ -119,6 +126,107 @@ func handleNetwork(c *components, install *Installation) error {
...
+				"detected GKE CNI plugin, FELIX_IPTABLESFILTERALLOWACTION was set to %s, expected it to be 'Return'.",
+				*filterAllow)}
+		}
+	default:
+		return ErrIncompatibleCluster{
+			fmt.Sprintf("unexpected FELIX_INTERFACEPREFIX value: '%s'. Only 'eni, avz, gke' are supported.", *prefix),
+		}
+	}
+
+	c.node.ignoreEnv(containerCalicoNode, "IP")
```
ozdanborne, line 224: is this `TODO` or permanent?

tmjd: It is until we support non-Calico CNI and CalicoNetwork.

pkg/controller/migration/convert/network.go
tmjd: I see these as sub-handlers of handleNetwork. But I guess they could be individual handlers. I'll make the switch.

pkg/controller/migration/convert/network.go
tmjd: Actually no, ignore probably isn't right.

pkg/controller/migration/convert/core.go
ozdanborne: Glad you added it to the caller, but looks like you forgot to remove it here

pkg/controller/migration/convert/network.go
```diff
@@ -65,7 +81,7 @@ func handleNetwork(c *components, install *Installation) error {
 	if am != nil {
 		tam, err := getAutoDetection(*am)
 		if err != nil {
-			return ErrIncompatibleCluster{"error parsing IP_AUTODETECTION_METHOD: " + err.Error()}
+			return ErrIncompatibleCluster{fmt.Sprintf("%s, error parsing IP_AUTODETECTION_METHOD: "+err.Error(), context)}
```
ozdanborne, line 84: nit, you're already using fmt.Sprintf - why not go the extra mile?
```suggestion
			return ErrIncompatibleCluster{fmt.Sprintf("%s, error parsing IP_AUTODETECTION_METHOD: %s", context, err)}
```

pkg/controller/migration/convert/network.go
```diff
@@ -119,6 +130,107 @@ func handleNetwork(c *components, install *Installation) error {
...
 }
 
+func handleNonCalicoCNI(c *components, install *Installation) error {
+	plugin, err := getCNIPlugin(c)
+	if err != nil {
+		return err
+	}
+	if plugin == operatorv1.PluginCalico {
+		return nil
+	}
```
ozdanborne, line 140: You've implemented this correctly based on how I've put all he handlers in a slice. But I'm wondering aloud if handlers should include the logic to decide if they need to be run. I wonder if we should remove that slice and just have one big function that runs these kind of checks and calls the handlers accordingly.

pkg/controller/migration/convert/network.go
```diff
@@ -16,34 +16,50 @@ const (
 )
 
 func handleNetwork(c *components, install *Installation) error {
-	// TODO: be more elegant about creation of this spec field if it already exists
-	if install.Spec.CalicoNetwork == nil {
-		install.Spec.CalicoNetwork = &operatorv1.CalicoNetworkSpec{}
+
+	// FELIX_DEFAULTENDPOINTTOHOSTACTION
+	defaultWepAction, err := c.node.getEnv(ctx, c.client, containerCalicoNode, "FELIX_DEFAULTENDPOINTTOHOSTACTION")
```
caseydavenport, line 21: We should track this one, I think.

Either:
- The default is already Accept, and so we can just remove our setting of the env var.
- The default should be changed to Accept so we can remove our env var.

Then, we can simply convert this env var into datastore config if we see it. 

No work to do here now, though.

pkg/controller/migration/convert/network.go
```diff
@@ -16,34 +16,50 @@ const (
...
+func handleCalicoCNI(c *components, install *Installation) error {
+	plugin, err := getCNIPlugin(c)
+	if err != nil {
+		return err
+	}
+	if plugin != operatorv1.PluginCalico {
+		return nil
+	}
+
+	context := fmt.Sprintf("detected %s CNI plugin", plugin)
```
caseydavenport, line 43: I'd recommend against calling this `context` - easily confused with the `context` package in go.

pkg/controller/migration/convert/network.go
```diff
@@ -16,34 +16,50 @@ const (
...
-	// FELIX_DEFAULTENDPOINTTOHOSTACTION
-	defaultWepAction, err := c.node.getEnv(ctx, c.client, containerCalicoNode, "FELIX_DEFAULTENDPOINTTOHOSTACTION")
-	if err != nil {
-		return err
-	}
-	if defaultWepAction != nil && strings.ToLower(*defaultWepAction) != "accept" {
-		return ErrIncompatibleCluster{
-			fmt.Sprintf("unexpected FELIX_DEFAULTENDPOINTTOHOSTACTION: '%s'. Only 'accept' is supported.", *defaultWepAction),
-		}
+	// TODO: be more elegant about creation of this spec field if it already exists
```
caseydavenport, line 60: Not sure I follow - what more do you think we could do to resolve this TODO? This seems fine to me.

pkg/controller/migration/convert/network.go
```diff
@@ -119,6 +130,107 @@ func handleNetwork(c *components, install *Installation) error {
...
+	}
+
+	if install.Spec.CNI == nil {
+		install.Spec.CNI = &operatorv1.CNISpec{}
+	}
+
+	switch plugin {
+	case operatorv1.PluginAmazonVPC:
+		install.Spec.CNI.Type = plugin
+		// FELIX_IPTABLESMANGLEALLOWACTION
```
caseydavenport, line 167: This comment isn't super useful - I'd recommend removing it or flushing it out to explain why we care about this field (probably the latter)

pkg/controller/migration/convert/network.go
```diff
@@ -119,6 +130,107 @@ func handleNetwork(c *components, install *Installation) error {
...
+		mangleAllow, err := c.node.getEnv(ctx, c.client, containerCalicoNode, "FELIX_IPTABLESMANGLEALLOWACTION")
+		if err != nil {
+			return err
+		}
+
+		// check that IPTABLESMANGLE is set Return
```
caseydavenport, line 173: Same here - that we're checking it is obvious, _why_ is less so.

pkg/controller/migration/convert/network.go
```diff
@@ -119,6 +130,107 @@ func handleNetwork(c *components, install *Installation) error {
...
+		} else if *mangleAllow != "Return" {
+			return ErrIncompatibleCluster{fmt.Sprintf("%s, FELIX_IPTABLESMANGLEALLOWACTION was %s, expected it to be 'Return'.", context, *mangleAllow)}
+		}
+
+		// FELIX_IPTABLESFILTERALLOWACTION
+		filterAllow, err := c.node.getEnv(ctx, c.client, containerCalicoNode, "FELIX_IPTABLESFILTERALLOWACTION")
+		if err != nil {
+			return err
+		}
+		// check that IPTABLESFILTER is set Return
```
caseydavenport, line 201: IPTABLESFILTERALLOWACTION?

pkg/controller/migration/convert/network.go
```diff
@@ -119,6 +130,107 @@ func handleNetwork(c *components, install *Installation) error {
...
+		if filterAllow == nil {
+			return ErrIncompatibleCluster{fmt.Sprintf("%s, FELIX_IPTABLESFILTERALLOWACTION was unset, expected it to be 'Return'.", context)}
+		} else if *filterAllow != "Return" {
+			return ErrIncompatibleCluster{fmt.Sprintf("%s, FELIX_IPTABLESFILTERALLOWACTION was set to %s, expected it to be 'Return'.", context, *filterAllow)}
+		}
+	default:
+		return ErrIncompatibleCluster{
+			fmt.Sprintf("unsupported plugin migrate: '%s'.", plugin),
```
caseydavenport, line 209: "unsupported plugin migrate" isn't really a complete thought. Maybe "unable to migrate plugin %s: unsupported"?

pkg/controller/migration/convert/network.go
```diff
@@ -119,6 +130,107 @@ func handleNetwork(c *components, install *Installation) error {
...
+	}
+
+	// TODO: Handle configuration with IPs and Pools specified.
+	// We need to relax the restriction on CalicoNetwork and non-Calico CNI to do this.
+
+	ip, err := c.node.getEnv(ctx, c.client, containerCalicoNode, "IP")
+	if err != nil {
+		return err
+	}
+	if ip != nil && *ip != "" {
```
caseydavenport, line 220: It's probably OK to relax this validation here. Having it set shouldn't cause any harm.

pkg/controller/migration/convert/network.go
```diff
@@ -155,3 +267,29 @@ func getAutoDetection(method string) (operatorv1.NodeAddressAutodetection, error
 
 	return operatorv1.NodeAddressAutodetection{}, errors.New("unrecognized method: " + method)
 }
+
+func getCNIPlugin(c *components) (operatorv1.CNIPluginType, error) {
+	// FELIX_INTERFACEPREFIX
```
caseydavenport, line 272: Similarly here - comment isn't saying much :D

pkg/controller/migration/convert/network.go
tmjd: I think this was a note that I just copied. But I agree I'm not sure what else would be done.

PR #747, opened 2020-07-23
cni command change for enterprise release-v3.2.

tmjd:
## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/node.go
+++ b/pkg/render/node.go
@@ -644,15 +644,20 @@ func (c *nodeComponent) cniContainer() v1.Container {
 		{MountPath: "/host/etc/cni/net.d", Name: "cni-net-dir"},
 	}
 
+	command := "/install-cni.sh"
+
 	image := components.GetReference(components.ComponentCalicoCNI, c.cr.Spec.Registry, c.cr.Spec.ImagePath)
 	if c.cr.Spec.Variant == operator.TigeraSecureEnterprise {
 		image = components.GetReference(components.ComponentTigeraCNI, c.cr.Spec.Registry, c.cr.Spec.ImagePath)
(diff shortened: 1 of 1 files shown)


pkg/render/node.go
```diff
@@ -644,15 +644,20 @@ func (c *nodeComponent) cniContainer() v1.Container {
...
 	}
 
+	command := "/install-cni.sh"
+
 	image := components.GetReference(components.ComponentCalicoCNI, c.cr.Spec.Registry, c.cr.Spec.ImagePath)
 	if c.cr.Spec.Variant == operator.TigeraSecureEnterprise {
 		image = components.GetReference(components.ComponentTigeraCNI, c.cr.Spec.Registry, c.cr.Spec.ImagePath)
+		// Enterprise release-v3.2 has cni changes that converted shell script to golang binary. The same change
+		// is not in the corresponding os version.
+		command = "/opt/cni/bin/install"
```
ozdanborne, line 654: I appreciate the fix you've done here. but why isn't this in OS?

Could we just PR a change to cni-plugin which runs `ln -s /install-cni.sh /opt/cni/bin/install` during the image build until it is added there too?

tmjd: It is in OS but not in a released version of OS.

tmjd: We could change the cni-plugin but it is simpler to just change what we're deploying instead of adding cruft to the cni-plugin.

tmjd: I picked the changes that are in for the v3.2 enterprise release. I think we should update the images to match what it uses when it is released. So I thought it would be best to just go with what it has.

ozdanborne: OH I didn't notice the base branch on this PR was our feature branch. LGTM

PR #760, opened 2020-07-24
Relax validation only allowing CalicoNetwork with Calico CNI plugin

tmjd:
## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/installation/validation.go
+++ b/pkg/controller/installation/validation.go
@@ -69,10 +69,6 @@ func validateCustomResource(instance *operatorv1.Installation) error {
 
 	// Verify Calico settings, if specified.
 	if instance.Spec.CalicoNetwork != nil {
-		// Currently, calicoNetwork is valid only when using the Calico CNI plugin.
-		if instance.Spec.CNI.Type != operatorv1.PluginCalico {
-			return fmt.Errorf("spec.CalicoNetwork requires spec.cni.type 'Calico'")
-		}
 
 		nPools := len(instance.Spec.CalicoNetwork.IPPools)
 		if nPools > 2 {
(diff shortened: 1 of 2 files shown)


pkg/controller/installation/validation_test.go
```diff
@@ -161,20 +161,20 @@ var _ = Describe("Installation validation tests", func() {
...
-			Entry("should disallow AzureVNET", operator.PluginAzureVNET),
+			Entry("GKE", operator.PluginGKE),
+			Entry("AmazonVPC", operator.PluginAmazonVPC),
+			Entry("AzureVNET", operator.PluginAzureVNET),
 		)
 	})
 	Describe("validate non-calico CNI plugin Type", func() {
 		BeforeEach(func() {
-			instance.Spec.CalicoNetwork = nil
+			//instance.Spec.CalicoNetwork = nil
```
caseydavenport, line 177: Is this intentional?

pkg/controller/installation/validation.go
```diff
@@ -155,8 +167,14 @@ func validateCustomResource(instance *operatorv1.Installation) error {
 					v6pool.NATOutgoing, strings.Join(operatorv1.NATOutgoingTypesString, ","))
 			}
 
-			if v6pool.NodeSelector == "" {
-				return fmt.Errorf("ipPool.nodeSelector should not be empty")
+			if instance.Spec.CNI.Type == operatorv1.PluginCalico {
+				if v6pool.NodeSelector == "" {
```
caseydavenport, line 171: This validation isn't Calico specific, right? The node-selector must _always_ be non empty?

tmjd: You're right.

tmjd: I'm going to change the IPv4 code also. In both cases I'll move the `NodeSelector == ""` check to after Plugin check block. So in the case of non-Calico we'll check the NodeSelector twice.

PR #774, opened 2020-07-24
Add cluster-admin permissions to the compliance controller for DockerEE

mgleung:
## Description

In order to start pods with privileged permissions or host path volumes we need to grant the creating pod cluster-admin privileges in DockerEE (https://docs.mirantis.com/docker-enterprise/v3.0/dockeree-products/ucp/authorization/access-control-model.html#secure-kubernetes-defaults). Our other pods are unaffected because the operator orchestrates their creation, but in the compliance-controller's case, it will start jobs that require host path volumes.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


doublek:
Let's hold off on merging this, until we run out of other options.

PR #777, opened 2020-07-27
Add Kibana access to predefined cluster roles.

rene-dekker:


pkg/render/apiserver.go
```diff
@@ -933,7 +933,7 @@ func (c *apiServerComponent) tigeraUserClusterRole() *rbacv1.ClusterRole {
 			APIGroups: []string{"lma.tigera.io"},
 			Resources: []string{"*"},
 			ResourceNames: []string{
-				"flows", "audit*", "events", "dns",
+				"flows", "audit*", "events", "dns", "kibana_login",
```
stevegaossou, line 936: Hey @rene-dekker, so I read the comment for this section and it says: 

```
	// If this is a managed cluster the rule to access the clusters indices in Elasticsearch need to be added to the management
	// cluster
```

It was added a while ago in this PR https://github.com/tigera/operator/pull/356/files and then you augmented it with the new ManagementCluster CR logic. 

The comment is confusing, it doesn't seem accurate with the "If this is a managed cluster". Also it talks specifically about ES indices, but with the new resource names added here for Kibana it is now more than just ES indices. 

Could you fix it?

pkg/render/apiserver.go
```diff
@@ -1052,7 +1052,7 @@ func (c *apiServerComponent) tigeraNetworkAdminClusterRole() *rbacv1.ClusterRole
 			APIGroups: []string{"lma.tigera.io"},
 			Resources: []string{"*"},
 			ResourceNames: []string{
-				"flows", "audit*", "events", "dns",
+				"flows", "audit*", "events", "dns", "elasticsearch_superuser",
```
stevegaossou, line 1055: Same thing here as above, about fixing the comment.

```
	// If this is a managed cluster the rule to access the clusters indices in Elasticsearch need to be added to the management
	// cluster

```

pkg/render/apiserver_test.go
```diff
@@ -218,6 +218,13 @@ var _ = Describe("API server rendering tests", func() {
 		Expect(d.Spec.Template.Spec.Volumes[1].ConfigMap.Items[0].Key).To(Equal("config"))
 		Expect(d.Spec.Template.Spec.Volumes[1].ConfigMap.Items[0].Path).To(Equal("policy.conf"))
 		Expect(len(d.Spec.Template.Spec.Volumes[1].ConfigMap.Items)).To(Equal(1))
+
+		clusterRole := GetResource(resources, "tigera-network-admin", "", "rbac.authorization.k8s.io", "v1", "ClusterRole").(*rbacv1.ClusterRole)
+		Expect(clusterRole.Rules).To(ConsistOf(networkAdminPolicyRules))
+
+		clusterRole = GetResource(resources, "tigera-ui-user", "", "rbac.authorization.k8s.io", "v1", "ClusterRole").(*rbacv1.ClusterRole)
+		Expect(clusterRole.Rules).To(ConsistOf(uiUserPolicyRules))
```
stevegaossou, line 226: Not sure if this description is still accurate (about what resources to expected in the render): 
https://github.com/tigera/operator/pull/777/files#diff-969ab7a6ef1d9769dd6b6d7977f3300fR87-R100

If it's accurate could you update it with the new ClusterRoles you added? If it's not accurate, I think we should just remove it.

PR #788, opened 2020-07-30
block migration for unsupported configurations

ozdanborne:
## Description

- tolerations
- volume mounts
- canal

I also converted typha into a pointer in the components struct since it is an optional component.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/migration/convert/aws-cni-policy-only_test.go
+++ b/pkg/controller/migration/convert/aws-cni-policy-only_test.go
@@ -184,20 +184,5 @@ func awsCNIPolicyOnlyConfig() []runtime.Object {
 				},
 			},
 		},
-		&appsv1.Deployment{
-			ObjectMeta: metav1.ObjectMeta{
-				Name:      "calico-kube-controllers",
-				Namespace: "kube-system",
-			},
-			Spec: appsv1.DeploymentSpec{
-				Template: corev1.PodTemplateSpec{
-					Spec: corev1.PodSpec{
-						Containers: []corev1.Container{{
-							Name: "calico-node",
-						}},
-					},
-				},
-			},
-		},
 	}
 }
--- a/pkg/controller/migration/convert/core.go
+++ b/pkg/controller/migration/convert/core.go
@@ -27,25 +27,31 @@ func handleCore(c *components, install *Installation) error {
 	}
 
 	// kube-controllers
-	kubeControllers := getContainer(c.kubeControllers.Spec.Template.Spec, "calico-kube-controllers")
-	if kubeControllers != nil && (len(kubeControllers.Resources.Limits) > 0 || len(kubeControllers.Resources.Requests) > 0) {
-		install.Spec.ComponentResources = append(install.Spec.ComponentResources, &operatorv1.ComponentResource{
-			ComponentName:        operatorv1.ComponentNameKubeControllers,
-			ResourceRequirements: kubeControllers.Resources.DeepCopy(),
-		})
+	if c.kubeControllers != nil {
+		kubeControllers := getContainer(c.kubeControllers.Spec.Template.Spec, "calico-kube-controllers")
+		if kubeControllers != nil && (len(kubeControllers.Resources.Limits) > 0 || len(kubeControllers.Resources.Requests) > 0) {
+			install.Spec.ComponentResources = append(install.Spec.ComponentResources, &operatorv1.ComponentResource{
+				ComponentName:        operatorv1.ComponentNameKubeControllers,
+				ResourceRequirements: kubeControllers.Resources.DeepCopy(),
+			})
+		}
 	}
 
-	// typha resource limits. typha is optional so check for nil first
-	typha := getContainer(c.typha.Spec.Template.Spec, "calico-typha")
-	if typha != nil && (len(typha.Resources.Limits) > 0 || len(typha.Resources.Requests) > 0) {
-		install.Spec.ComponentResources = append(install.Spec.ComponentResources, &operatorv1.ComponentResource{
-			ComponentName:        operatorv1.ComponentNameTypha,
-			ResourceRequirements: typha.Resources.DeepCopy(),
-		})
+	if c.typha != nil {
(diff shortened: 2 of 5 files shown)


pkg/controller/migration/convert/aws-cni-policy-only_test.go
```diff
@@ -192,6 +192,16 @@ func awsCNIPolicyOnlyConfig() []runtime.Object {
 			Spec: appsv1.DeploymentSpec{
 				Template: corev1.PodTemplateSpec{
 					Spec: corev1.PodSpec{
+						Tolerations: []corev1.Toleration{
```
tmjd, line 195: I messed up here, actually this Deployment needs to be removed, there is no calico-kube-controllers in the amazon-vpc-cni repo.
This file was meant to represent https://raw.githubusercontent.com/aws/amazon-vpc-cni-k8s/master/config/master/calico.yaml. I think I had copied this and forgot to remove it.

pkg/controller/migration/convert/core.go
```diff
@@ -145,3 +216,21 @@ func handleFelixNodeMetrics(c *components, install *Installation) error {
...
+func checkTolerations(existing []corev1.Toleration, expected ...corev1.Toleration) error {
+	if len(existing) != len(expected) {
+		return fmt.Errorf("missing expected tolerations. have: %+v. expecting: %+v", existing, expected)
+	}
+
+OUTER:
+	for _, t := range expected {
+		for _, k := range existing {
+			if k == t {
+				continue OUTER
```
tmjd, line 229: This is just to continue the outer loop so doing a break here should work.

ozdanborne: I don't think that's correct - If I break the inner loop, I'll immediately hit the `return fmt.Errorf`

pkg/controller/migration/convert/aws-cni-policy-only_test.go
ozdanborne: I added a commit which removes the kubeController deployment from this test, makes the kubeController in components a pointer, and doesn't throw an error in `getComponents` if it doesn't detect the kubecontrollers deployment :+1:

pkg/controller/migration/convert/core.go
```diff
@@ -71,6 +78,76 @@ func handleCore(c *components, install *Installation) error {
 		install.Spec.FlexVolumePath = "None"
 	}
 
+	// Volumes for lib-modules, xtables-lock, var-run-calico, var-lib-calico, or policysync have been changed
+	v := getVolume(c.node.Spec.Template.Spec, "lib-modules")
+	if v == nil || v.HostPath == nil || v.HostPath.Path != "/lib/modules" {
+		return ErrIncompatibleCluster{"must mount /lib/modules from host"}
```
tmjd, line 84: Should this message be more like `expected calico-node to have Volume 'lib-modules' with host path '/lib/modules'`? (same for the others below too)

pkg/controller/migration/convert/core.go
```diff
@@ -71,6 +78,76 @@ func handleCore(c *components, install *Installation) error {
...
+		},
+		corev1.Toleration{
+			Effect:   corev1.TaintEffectNoSchedule,
+			Operator: corev1.TolerationOpExists,
+		},
+		corev1.Toleration{
+			Effect:   corev1.TaintEffectNoExecute,
+			Operator: corev1.TolerationOpExists,
+		}); err != nil {
+		return fmt.Errorf("calico-node has incompatible tolerations: %v", err)
```
tmjd, line 123: should be ErrIncompatibleCluster

pkg/controller/migration/convert/core.go
```diff
@@ -71,6 +78,76 @@ func handleCore(c *components, install *Installation) error {
...
+		if err := checkTolerations(c.kubeControllers.Spec.Template.Spec.Tolerations,
+			corev1.Toleration{
+				Key:      "CriticalAddonsOnly",
+				Operator: corev1.TolerationOpExists,
+			},
+			corev1.Toleration{
+				Effect: corev1.TaintEffectNoSchedule,
+				Key:    "node-role.kubernetes.io/master",
+			}); err != nil {
+			return fmt.Errorf("kube-controllers has incompatible tolerations: %v", err)
```
tmjd, line 137: should be ErrIncompatibleCluster

pkg/controller/migration/convert/core.go
```diff
@@ -71,6 +78,76 @@ func handleCore(c *components, install *Installation) error {
...
+		}
+	}
+
+	// check typha tolerations
+	if c.typha != nil {
+		if err := checkTolerations(c.typha.Spec.Template.Spec.Tolerations, corev1.Toleration{
+			Key:      "CriticalAddonsOnly",
+			Operator: corev1.TolerationOpExists,
+		}); err != nil {
+			return fmt.Errorf("typha has incompatible tolerations: %v", err)
```
tmjd, line 147: ditto

pkg/controller/migration/convert/core.go
tmjd: Yep you're right.
Do you like this pattern? It is just a GOTO which I always feel should be avoided if possible.
One alternative would be to use a bool for if the toleration is found and use that to skip the `return fmt.Errorf`, then break would work.
But if you're not bothered by the 'goto' I can live with it.

pkg/controller/migration/convert/core.go
```diff
@@ -145,3 +222,21 @@ func handleFelixNodeMetrics(c *components, install *Installation) error {
 
 	return nil
 }
+
+func checkTolerations(existing []corev1.Toleration, expected ...corev1.Toleration) error {
+	if len(existing) != len(expected) {
+		return fmt.Errorf("missing expected tolerations. have: %+v. expecting: %+v", existing, expected)
```
tmjd, line 228: The errors in this function should be ErrIncompatibleCluster.

PR #791, opened 2020-07-31
Set routeSource in Felix configuration to WorkloadIPs for aws-vpc CNI

realgaurav:
## Description

Set `routeSource` in Felix configuration to `WorkloadIPs` for aws-vpc CNI.

Default value for `routeSource` in Felix configuration is `CalicoIPAM`. With value set to `CalicoIPAM`, dataplane in Felix doesn't get RouteUpdate for remote workloads manifesting into route issues for encryption functionality.

## For PR author

- [x] Tests for change.
- [ ] ~If changing pkg/apis/, run `make gen-files`~
- [ ] ~If changing versions, run `make gen-versions`~

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/node.go
+++ b/pkg/render/node.go
@@ -1022,6 +1022,10 @@ func (c *nodeComponent) nodeEnvVars() []v1.EnvVar {
 	}
 	nodeEnv = append(nodeEnv, v1.EnvVar{Name: "FELIX_IPTABLESBACKEND", Value: "auto"})
 
+	if c.cr.Spec.CNI.Type != operator.PluginCalico {
+		nodeEnv = append(nodeEnv, v1.EnvVar{Name: "FELIX_ROUTESOURCE", Value: "WorkloadIPs"})
+	}
+
 	if c.amazonCloudInt != nil {
 		nodeEnv = append(nodeEnv, GetTigeraSecurityGroupEnvVariables(c.amazonCloudInt)...)
 		nodeEnv = append(nodeEnv, v1.EnvVar{
--- a/pkg/render/node_test.go
+++ b/pkg/render/node_test.go
@@ -739,6 +739,7 @@ var _ = Describe("Node rendering tests", func() {
 			{Name: "FELIX_IPTABLESBACKEND", Value: "auto"},
 			{Name: "FELIX_INTERFACEPREFIX", Value: "eni"},
 			{Name: "FELIX_IPTABLESMANGLEALLOWACTION", Value: "Return"},
+			{Name: "FELIX_ROUTESOURCE", Value: "WorkloadIPs"},
 		}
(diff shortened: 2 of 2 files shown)


pkg/render/node.go
```diff
@@ -1011,6 +1011,7 @@ func (c *nodeComponent) nodeEnvVars() []v1.EnvVar {
 	case operator.PluginAmazonVPC:
 		nodeEnv = append(nodeEnv, v1.EnvVar{Name: "FELIX_INTERFACEPREFIX", Value: "eni"})
 		nodeEnv = append(nodeEnv, v1.EnvVar{Name: "FELIX_IPTABLESMANGLEALLOWACTION", Value: "Return"})
+		nodeEnv = append(nodeEnv, v1.EnvVar{Name: "FELIX_ROUTESOURCE", Value: "WorkloadIPs"})
```
caseydavenport, line 1014: I wonder if we should actually be setting this for all plugins whenever calico IPAM isn't being used.

The two options are:
- CalicoIPAM
- WorkloadIPs

Seems like if `spec.cni.ipam.type` is Calico then we should use CalicoIPAM, otherwise we should use WorkloadIPs. WDYT?

realgaurav: What you are suggesting does make sense to me. However I am not sure of the impact or possible regression.
Mainly because I haven't looked at all the code dealing with `routeSource` in Felix, I am a bit skeptical of making this change in general.

The code where I saw `routeSource` impacting encryption functionality is fairly new, so it's possible we have just not stumbled on this issue for other platforms. My suggestion would be to go ahead with this change and see if e2e runs show any relatable failures. If we do see similar issue, then generalize.

Please let me know what you think.

robbrockbank: Feels like we might be missing an option if pod IPs are restricted to a specific CIDR for each node.

realgaurav: @robbrockbank do you mean for in general or for aws-vpc platform? I was able to verify this change with aws-vpc CNI.

robbrockbank: For general... in particular i'm not sure what happens regarding aggregation of IPs with host local.  But that might be handled differently so perhaps not an issue.

robbrockbank: But in regards to Casey's immediate comment - i agree in that it feels like we should use WorkloadIPs for non-calico IPAM

caseydavenport: Yeah, it should be a safe change. The intention of WorkloadIPs was specifically to handle non-Calico IPAM scnearios, including host-local IPAM. Maybe we could also add a "NodeCIDRs` option or something, but that's really an optimization.

realgaurav: Thanks Rob & Casey. I will address this.

PR #808, opened 2020-08-03
Lots of IPAM migration/blocking

tmjd:
## Description

I think I've got some good stuff here. I'll take a look again tomorrow but I think a 1st review would be good.
TODO: I need to add a test to actually use the calicoDefaultConfig that I've added.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


pkg/controller/migration/convert/cni.go
```diff
@@ -1,14 +1,47 @@
...
+// IPAMConfig represents the IP related network configuration.
+// This nests Range because we initially only supported a single
+// range directly, and wish to preserve backwards compatability
+type HostLocalIPAMConfig struct {
+	*Range
+	Name       string
+	Type       string         `json:"type"`
+	Routes     []*types.Route `json:"routes"`
+	DataDir    string         `json:"dataDir"`
+	ResolvConf string         `json:"resolvConf"`
```
ozdanborne, line 31: Do we need _all_ of this stuff?

pkg/controller/migration/convert/cni.go
```diff
@@ -42,11 +75,30 @@ func loadCNI(c *components) error {
...
+				raw := CNIHostLocalIPAM{}
+				if err := json.Unmarshal(plugin.Bytes, &raw); err != nil {
+					return fmt.Errorf("failed to parse cni config for raw IPAM: %w", err)
+				}
+
+				// Use a Decoder so we can set DisallowUnknownFields so if there are
+				// any unknown fields we will detect that.
+				x := HostLocalIPAMConfig{}
+				dec := json.NewDecoder(bytes.NewReader(raw.IPAM))
+				dec.DisallowUnknownFields()
```
ozdanborne, line 91: Similar to my previous comment, I wonder if at a higher level we don't need to go validating that their host-local CNI is valid, and can just parse the bits we need...

pkg/controller/migration/convert/cni.go
```diff
@@ -1,14 +1,47 @@
...
 
+// Used to parse the IPAM section out of the full CNI configuration
+type CNIHostLocalIPAM struct {
+	IPAM json.RawMessage `json:"ipam"`
+}
+
+// IPAMConfig represents the IP related network configuration.
+// This nests Range because we initially only supported a single
+// range directly, and wish to preserve backwards compatability
+type HostLocalIPAMConfig struct {
```
ozdanborne, line 25: Did you consider just importing this as well? https://github.com/containernetworking/plugins/blob/master/plugins/ipam/host-local/backend/allocator/config.go#L43-L52

pkg/controller/migration/convert/network.go
```diff
@@ -110,10 +147,123 @@ func handleCalicoCNI(c *components, install *Installation) error {
...
 		return ErrIncompatibleCluster{"IpAddrsNoIpam not supported"}
 	}
+	if c.calicoCNIConfig.ContainerSettings.AllowIPForwarding {
+		return ErrIncompatibleCluster{"AllowIPForwarding not supported"}
+	}
+
+	return nil
+}
+
+func subhandleCalicoIPAM(c *components, install *Installation) error {
```
ozdanborne, line 157: minor suggestion - could probably _just_ pass in `c.calicoCNIConfig.IPAM` instead of the whole `*components` struct.

pkg/controller/migration/convert/cni.go
tmjd: Yep the problem is the Subnet we use (`usePodCidr`) is not actually valid and the Calico CNI replaces that field when actually running. So I needed to be able to replace its definition.

pkg/controller/migration/convert/cni.go
tmjd: Not really, it is just what I copied over from containernetworking/plugins

pkg/controller/migration/convert/cni.go
tmjd: My intention here wasn't to validate that it was valid but ensure there were no fields we don't know how to handle.

tmjd: I'm not concerned that what they have is valid or not but that we can't reproduce their config.

pkg/controller/migration/convert/network.go
tmjd: Hmm, I kind of moved some other stuff into the subhandlers, see what you think about that.
I think I like keeping the (sub)handlers all with the same interface, but I'm not sure if that's a good reason.
... Looks like I might have broke something.

pkg/controller/migration/convert/network.go
```diff
@@ -41,30 +41,67 @@ func handleCalicoCNI(c *components, install *Installation) error {
...
+		install.Spec.CNI.IPAM.Type = operatorv1.IPAMPluginCalico
+		install.Spec.CalicoNetwork.BGP = operatorv1.BGPOptionPtr(operatorv1.BGPEnabled)
+		switch {
+		case netBackend == nil:
+		case *netBackend == "":
+		case *netBackend == "bird":
+		case *netBackend == "vxlan":
+			install.Spec.CalicoNetwork.BGP = operatorv1.BGPOptionPtr(operatorv1.BGPDisabled)
+		default:
+			return ErrIncompatibleCluster{fmt.Sprintf("%s and IPAM plugin %s, CALICO_NETWORKING_BACKEND %s is not valid",
```
ozdanborne: Thinking out loud, does this belong in `validation.go`?

tmjd: I don't think we'd have a way to represent the config in the Installation API, so I don't think it would be possible to just migrate the config and have validation pick up that something was wrong.

pkg/controller/migration/convert/network.go
```diff
@@ -98,22 +108,158 @@ func handleCalicoCNI(c *components, install *Installation) error {
...
+		// this check is just double checking what the caller should have already
+		// validated.
+		return nil
+	}
+
+	// TODO: Is there anything we need to do with the Name field on the CNI plugin or IPAM
+	// plugin? We obviously don't have a way currently to persist them.
+
+	// CALICO_NETWORKING_BACKEND
+	netBackend, err := c.node.getEnv(ctx, c.client, containerCalicoNode, "CALICO_NETWORKING_BACKEND")
```
ozdanborne, line 137: I think this PR you're working on is the first scenario where it makes sense for us to have the processing in these subhandlers depend on processing done in the calling handler (handleNetwork). 

have handleNetwork process `CALICO_NETWORKING_BACKEND` by setting fields in the Installation resource. Then, have these handlers use conditionals off of that instead of having to run through that processing a second time.

Alternatively, we move those into functions which can be called from both places.

tmjd: This field doesn't directly translate into a single field so I don't think we can read it and put it into the Installation resource.
I can move it into one function that both subhandle functions call.

ozdanborne: The way you've moved it into a function looks a million times better. But I just want to continue the original thought -

it seems like it mostly translates into `install.Spec.CalicoNetwork.BGP`. We could have 

```go
	netBackend, err := c.node.getEnv(ctx, c.client, containerCalicoNode, "CALICO_NETWORKING_BACKEND")
	if err != nil {
		return "", err
	}

	switch {
	case netBackend == nil, netBackend == "", *netBackend == "bird":
		install.Spec.CalicoNetwork.BGP = operatorv1.BGPOptionPtr(operatorv1.BGPEnabled)
	case *netBackend == "vxlan", *netBackend == "none":
		install.Spec.CalicoNetwork.BGP = operatorv1.BGPOptionPtr(operatorv1.BGPDisabled)
	default:
		return "", fmt.Errorf("CALICO_NETWORKING_BACKEND %s is not valid", *netBackend)
	}
```

Then the subhandlers could check `install.Spec.CalicoNetwork.BGP`. And if, given the cni / ipam, it's incompatible, they could throw an error.

tmjd: I think that would leave us open to trying to migrate something we don't support. Specifically with HostLocal `vxlan` backend is not valid yet with your suggestion it would look the same as backend of `none`.
There is also a less likely chance with Calico CNI with `none` as the backend.

pkg/controller/migration/convert/cni.go
ozdanborne: We already handle a replacement of `__CNI_MTU__` in the CNI parsing code. Is there a reason we don't want to reuse that approach for this one?

pkg/controller/migration/convert/network.go
```diff
@@ -98,22 +108,174 @@ func handleCalicoCNI(c *components, install *Installation) error {
...
+func subhandleCalicoIPAM(c *components, install *Installation) error {
+	if c.calicoCNIConfig.IPAM.Type != "calico-ipam" {
+		// I don't know if this should be an error or not, the way it is written
+		// this check is just double checking what the caller should have already
+		// validated.
+		return nil
+	}
+
+	// TODO: Is there anything we need to do with the Name field on the CNI plugin or IPAM
+	// plugin? We obviously don't have a way currently to persist them.
```
ozdanborne, line 156: did you answer these questions yet?

pkg/controller/migration/convert/network.go
```diff
@@ -98,22 +108,174 @@ func handleCalicoCNI(c *components, install *Installation) error {
...
+// subhandleHostLocalIPAM checks all fields in the Host Local IPAM configuration,
+// if any fields have unexpected values an error message will be returned.
+// The function tries to collect all the errors and report one message.
+// If there are no errors and the config can be added to the 'install' passed
+// in then nil is returned.
+func subhandleHostLocalIPAM(c *components, install *Installation) error {
+	if c.calicoCNIConfig.IPAM.Type != "host-local" {
+		// I don't know if this should be an error or not, the way it is written
+		// this check is just double checking what the caller should have already
+		// validated.
```
ozdanborne, line 201: did you answer these questions yet?

pkg/controller/migration/convert/network.go
```diff
@@ -41,30 +41,40 @@ func handleCalicoCNI(c *components, install *Installation) error {
...
-		return ErrIncompatibleCluster{fmt.Sprintf("%s, required cni config was not found ", errCtx)}
+		return ErrIncompatibleCluster{fmt.Sprintf("%s, required Calico cni config was not found ", errCtx)}
 	}
 
 	if install.Spec.CNI == nil {
 		install.Spec.CNI = &operatorv1.CNISpec{}
 	}
 	install.Spec.CNI.Type = plugin
 
+	if install.Spec.CNI.IPAM == nil {
```
ozdanborne, line 53: all these nil checks here - is it safe to say that they will all need to be initialized during a migration since we are detecting the existing networking mode? If so, maybe we can do this earlier?

pkg/controller/migration/convert/cni.go
ozdanborne: Can we reproduce their config if they've set `ResolvConf`? I think you need to drop any field in here that we don't know how to migrate so that the later unmarshal using `DisallowUnknownFields` will raise an error

pkg/controller/migration/convert/network.go
```diff
@@ -98,22 +108,174 @@ func handleCalicoCNI(c *components, install *Installation) error {
...
+func subhandleHostLocalIPAM(c *components, install *Installation) error {
+	if c.calicoCNIConfig.IPAM.Type != "host-local" {
+		// I don't know if this should be an error or not, the way it is written
+		// this check is just double checking what the caller should have already
+		// validated.
+		return nil
+	}
+
+	if c.hostLocalIPAMConfig == nil {
+		return ErrIncompatibleCluster{"no configuration"}
```
ozdanborne, line 206: What's the scenario where this error would be hit?  I think it's when they're using host-local, but for some reason, `getComponents` didn't initialize `c.hostLocalIPAMConfig` with the parsed config. The only situation where that woudl occur is probably developer error? Maybe CNI parsing for host-local should just be done here in the host-local handler instead of during `getComponents` :shrug:

pkg/controller/migration/convert/network.go
ozdanborne: it might help cut down on all the nil checks that you have to do if these just accepted the bits they're processing :shrug:

pkg/controller/migration/convert/network.go
tmjd: I don't think I would want to do this earlier. The CalicoNetwork and IPAM fields may not be set in all cases (like for non-Calico CNI). I believe the CNI one will always be done but I think I'd prefer to keep it close to where we actually set it.

pkg/controller/migration/convert/network.go
tmjd: I think it would be a developer error. Because when processing the CNI if IPAM.Type is hostlocal then a hostLocalConfig should be parsed and if that failed it should be an error there, which should block us from ever getting to this point.

I did consider doing that parsing here but thought it was better to keep those 2 jobs separated. With the loadCNI loading the config into our structs and then just processing those structs here.

pkg/controller/migration/convert/network.go
```diff
@@ -98,22 +116,167 @@ func handleCalicoCNI(c *components, install *Installation) error {
...
+		return "", fmt.Errorf("CALICO_NETWORKING_BACKEND %s is not valid", *netBackend)
+	}
+}
+
+// subhandleCalicoIPAM checks all fields in the Calico IPAM configuration,
+// if any fields have unexpected values an error message will be returned.
+// The function tries to collect all the errors and report one message.
+// If there are no errors and the config can be added to the passed in 'install'
+// then nil is returned.
+func subhandleCalicoIPAM(node CheckedDaemonSet, client client.Client, cnicfg calicocni.NetConf, install *Installation) error {
```
ozdanborne, line 164: I think this looks much cleaner by not passing in the entire components struct. However:
- I don't think the variable marking functionality of the `CheckedDaemonset` will work if you're passing this by value in to this function. It needs a pointer so that subsequent calls to `getEnv` will update the correct map.
- However, I notice that the only place `node` and `client` are used are during the call to `getNetworkingBackend`. Perhaps you just want call `getNetworkingBacked` in the caller, and then pass in the networking backend only?

tmjd: CheckedDaemonset holds the ignored information in a map which is a reference so when it is copied the copy can still update the ignore information.
I've pulled out the getNetworkingBackend so only the backend value is passed into the subhandle functions.

pkg/controller/migration/convert/cni.go
```diff
@@ -1,14 +1,47 @@
...
+	"net"
+	"reflect"
 	"strings"
 
 	"github.com/containernetworking/cni/libcni"
+	"github.com/containernetworking/cni/pkg/types"
 )
 
+// Used to parse the IPAM section out of the full CNI configuration
+type CNIHostLocalIPAM struct {
```
caseydavenport, line 18: does this need to be a public struct?

pkg/controller/migration/convert/network.go
```diff
@@ -5,6 +5,9 @@ import (
 	"fmt"
 	"strings"
 
+	calicocni "github.com/projectcalico/cni-plugin/pkg/types"
```
caseydavenport, line 8: Aren't we trying to avoid importing projectcalico/cni and projectcalico/libcalico-go?

pkg/controller/migration/convert/network.go
```diff
@@ -98,22 +121,157 @@ func handleCalicoCNI(c *components, install *Installation) error {
...
+		invalidFields = append(invalidFields, "ipam.ipv6pools field is unsupported")
+	}
+
+	if len(invalidFields) > 0 {
+		return ErrIncompatibleCluster{"configuration could not be migrated: " + strings.Join(invalidFields, ",")}
+	}
+	return nil
+}
+
+const UnsupportedMsg = "is unsupported"
```
caseydavenport, line 200: Is this really worthwhile? Seems like it's used inconsistently throughout the file anyway (some spell it out, some use this constant).

IMO we should just remove the constant and spell out the words "is unsupported" - it's not that many extra characters!

pkg/render/node.go
```diff
@@ -423,7 +423,7 @@ func (c *nodeComponent) getCalicoIPAM() string {
 	)
 }
 
-func (c *nodeComponent) getHostLocalIPAM() string {
+func BuildHostLocalIPAM(cns *operator.CalicoNetworkSpec) string {
```
caseydavenport, line 426: Should this be private?

PR #813, opened 2020-08-05
add more validation during migration

ozdanborne:
## Description

Validate the following:

1. annotations
2. nodeSelectors
3. `CNI_CONF_NAME`

See individual commits for more clear organization

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/migration/convert/core.go
+++ b/pkg/controller/migration/convert/core.go
@@ -160,6 +160,14 @@ func handleCore(c *components, install *Installation) error {
 		if e != nil && (e.ValueFrom == nil || e.ValueFrom.FieldRef == nil || e.ValueFrom.FieldRef.FieldPath != "spec.nodeName") {
 			return ErrIncompatibleCluster{"KUBERNETES_NODE_NAME on 'install-cni' container must be unset or be a FieldRef to 'spec.nodeName'"}
 		}
+
+		n, err := c.node.getEnv(ctx, c.client, containerInstallCNI, "CNI_CONF_NAME")
+		if err != nil {
+			return err
+		}
+		if n != nil && *n != "10-calico.conflist" {
+			return ErrIncompatibleCluster{"CNI_CONF_NAME on 'install-cni' container must be '10-calico.conflist'"}
+		}
 	}
 
 	// TODO: handle these vars appropriately
@@ -176,7 +184,6 @@ func handleCore(c *components, install *Installation) error {
 	c.node.ignoreEnv("calico-node", "FELIX_LOGSEVERITYSYS")
 	c.node.ignoreEnv("upgrade-ipam", "KUBERNETES_NODE_NAME")
 	c.node.ignoreEnv("upgrade-ipam", "CALICO_NETWORKING_BACKEND")
-	c.node.ignoreEnv("install-cni", "CNI_CONF_NAME")
 	c.node.ignoreEnv("install-cni", "SLEEP")
 
 	return nil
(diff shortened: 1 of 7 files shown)


pkg/controller/migration/convert/core_test.go
```diff
@@ -246,4 +246,44 @@ var _ = Describe("core handler", func() {
...
+			})
+		})
+		Context("typha", func() {
+			ExpectAnnotations(func(annotations map[string]string) {
+				comps.typha.Annotations = annotations
+			})
+			ExpectAnnotations(func(annotations map[string]string) {
+				comps.typha.Spec.Template.Annotations = annotations
+			})
+		})
```
tmjd, line 344: I'd appreciate a test with the safe-to-evict annotations on typha

pkg/controller/migration/convert/core.go
```diff
@@ -156,6 +156,14 @@ func handleCore(c *components, install *Installation) error {
 		return err
 	}
 
+	epHostAction, err := c.node.getEnv(ctx, c.client, containerCalicoNode, "FELIX_DEFAULTENDPOINTHOSTACTION")
```
tmjd, line 158: This check is already in network.go. You should remove it from there.
I think having it here is probably the right place.

pkg/controller/migration/convert/core_test.go
```diff
@@ -90,15 +90,55 @@ var _ = Describe("core handler", func() {
...
+			It("should not set nodeSelector if none is set", func() {
+				Expect(handleCore(&comps, i)).ToNot(HaveOccurred())
+				Expect(i.Spec.ControlPlaneNodeSelector).To(BeEmpty())
+			})
+			It("should carry forward nodeSelector on kube-controllers", func() {
+				nodeSelector := map[string]string{"foo": "bar"}
+				comps.kubeControllers.Spec.Template.Spec.NodeSelector = nodeSelector
+				Expect(handleCore(&comps, i)).ToNot(HaveOccurred())
+				Expect(i.Spec.ControlPlaneNodeSelector).To(Equal(nodeSelector))
+			})
```
tmjd, line 105: An interesting test here would be if there is a `kubernetes.io/os: windows` nodeSelector.

pkg/controller/migration/convert/core.go
ozdanborne: I thought the same thing, but I think we both were confusing it with `FELIX_IPTABLESMANGLEALLOWACTION` because I checked and there are no checks for `FELIX_IPTABLESMANGLEALLOWACTION`.

pkg/controller/migration/convert/core_test.go
ozdanborne: There is one! Line 122 :smile: https://github.com/tigera/operator/pull/813/files#diff-ca6adc83aab5bc88a29eb05f6f51a32aR123

pkg/controller/migration/convert/core.go
tmjd: https://github.com/tigera/operator/blob/tech-preview-migrate-install/pkg/controller/migration/convert/network.go#L22 is what I'm talking about

pkg/controller/migration/convert/core_test.go
tmjd: I meant on kube-controllers, I think what you linked is only done for typha and node.

pkg/controller/migration/convert/core.go
ozdanborne: Whoops! This one had a typo in the env var name which is why I didn't find the other. Fixed!

PR #816, opened 2020-08-07
Handle IP Migration

tmjd:
## Description

This includes adding CRD bindings for IPPools which then the IP Migration uses to retrieve the IPPools from the datastore.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


pkg/controller/installation/core_controller.go
```diff
@@ -226,16 +227,21 @@ func GetInstallation(ctx context.Context, client client.Client, provider operato
 }
 
 // GetInstallation returns the default installation instance with defaults populated.
-func getInstallation(ctx context.Context, client client.Client, provider operator.Provider) (*operator.Installation, error) {
```
ozdanborne, line 229: Surely there must be a way to create a single clientset from these two individual ones?

pkg/controller/migration/convert/network.go
```diff
@@ -437,16 +440,213 @@ func getCNIPlugin(c *components) (operatorv1.CNIPluginType, error) {
 	}
 }
 
-func handleIPPool(c *components, install *Installation) error {
+func handleIPPools(c *components, install *Installation) error {
+	pools, err := c.crdClient.IPPools().List(metav1.ListOptions{})
+	if err != nil {
```
ozdanborne, line 445: do we need special handling in case no pools exist?

pkg/controller/migration/convert/network.go
```diff
@@ -437,16 +440,213 @@ func getCNIPlugin(c *components) (operatorv1.CNIPluginType, error) {
 	}
 }
 
-func handleIPPool(c *components, install *Installation) error {
+func handleIPPools(c *components, install *Installation) error {
```
ozdanborne, line 443: could you add a comment to this function explaining what you and casey agreed on for the design here? My understanding is it's this:
```
// handleIPPools sets the installation's initialPools field. We first check the CALICO_IPV4POOL_* env vars on calico-node. 
// However, the user may have migrated off of that pool since it is only initially used when no other pool exists.
// As such, we make a client request to the pools in the datastore, and find one ipv4 and one ipv6 pool if at least one of each exists.
// When executing this client request, we prefer the pool if it matches the CALICO_IPV4POOL_* env var. However, we will pick any other if we cannot find one that matches.
```

pkg/controller/migration/convert/network.go
```diff
@@ -437,16 +440,213 @@ func getCNIPlugin(c *components) (operatorv1.CNIPluginType, error) {
...
+				return ErrIncompatibleCluster{"CNI config indicates assign_ipv6 true but there were no valid V6 pools found"}
+			}
+		} else {
+			if v6pool != nil {
+				return ErrIncompatibleCluster{"CNI config indicates assig_ipv6 false but a V6 pool was found"}
+			}
+		}
+	}
+
+	// Covert any found CRD pools into Operator pools and add them.
```
ozdanborne, line 503:
```suggestion
	// Convert any found CRD pools into Operator pools and add them.
```

hack/boilerplate/boilerplate.go.txt
```diff
@@ -0,0 +1 @@
+// Copyright (c) YEAR Tigera, Inc. All rights reserved.
```
ozdanborne, line 1: should this be checked in?

pkg/controller/migration/convert/convert_test.go
```diff
@@ -16,10 +19,21 @@ import (
 
 var _ = Describe("Parser", func() {
 	var ctx = context.Background()
+	var fakeCrd crdv1client.CrdV1Interface
+	BeforeEach(func() {
+		fakeCrd = fakecrd.NewSimpleClientset().CrdV1()
```
ozdanborne, line 24: I think you can remove the duplicate call in the beforeeach if you call this during initialization
```suggestion
	var fakeCrd  = fakecrd.NewSimpleClientset().CrdV1()
	BeforeEach(func() {
```

pkg/controller/migration/convert/network.go
```diff
@@ -437,16 +440,213 @@ func getCNIPlugin(c *components) (operatorv1.CNIPluginType, error) {
...
+		return nil, err
+	}
+	if pool != nil {
+		return pool, nil
+	}
+	return nil, nil
+}
+
+// convertPool converts the src (CRD) pool into an Installation/Operator IPPool
+func convertPool(src crdv1.IPPool) (*operatorv1.IPPool, error) {
```
ozdanborne, line 619: I don't see a reason for this to return a pointer when it never returns `nil, nil`.
```suggestion
func convertPool(src crdv1.IPPool) (operatorv1.IPPool, error) {
```

hack/boilerplate/boilerplate.go.txt
tmjd: Yeah we need this file for `make gen-datastore-bindings`.

pkg/controller/migration/convert/network.go
```diff
@@ -437,16 +440,213 @@ func getCNIPlugin(c *components) (operatorv1.CNIPluginType, error) {
...
+	p := operatorv1.IPPool{CIDR: src.Spec.CIDR}
+
+	ip := src.Spec.IPIPMode
```
ozdanborne, line 622: If an empty string mode means never, i think you could calm this code down a lot by setting it to that up front:
```suggestion
	ip := src.Spec.IPIPMode
	if ip == "" {
		ip = crdv1.IPIPModeNever
	}
```

pkg/controller/migration/convert/convert_test.go
tmjd: Are you suggesting only have 1 fakecrd.NewSimpleClientset() in the BeforeEach (which I agree with) or are you suggesting that I move the call out of BeforeEach all together?

ozdanborne: Right, just the former. I found it odd having two instances of `fakeCrd = fakecrd.NewSimpleClientset().CrdV1()` inside the beforeeach.

pkg/controller/migration/convert/network.go
tmjd: I don't think so.
The only addition along those lines is maybe we need some checks that we have pools for certain setups. @caseydavenport WDYT?
I think calico-ipam is covered below with the checks around assign_ip but for host-local would pools be required that we should check for?

caseydavenport: no need to assert on pools for anything other than Calico IPAM

pkg/controller/migration/convert/network.go
caseydavenport: Do we really need logic that complicated? Why not just something like this:

- We query all the IP pools
- If pools with the default-vX-pool name exist, use those.
- Otherwise, pick the first IPv4 and first IPv6 pools that we find.

hack/boilerplate/boilerplate.go.txt
caseydavenport: Should replace YEAR with 2020, if you still need this file after what we discussed earlier.

pkg/apis/addtoscheme_crd.projectcalico.org_v1.go
```diff
@@ -0,0 +1,24 @@
...
+
+package apis
+
+import (
+	crdv1 "github.com/tigera/operator/pkg/apis/crd.projectcalico.org/v1"
+)
+
+func init() {
+	// Register the types with the Scheme so the components can map objects to GroupVersionKinds and back
+	AddToSchemes = append(AddToSchemes, crdv1.SchemeBuilder.AddToScheme)
```
caseydavenport, line 23: I think we can just do this where we add all the other schemes (no need to make a new file for this)

pkg/apis/crd.projectcalico.org/v1/ippool.go
```diff
@@ -0,0 +1,140 @@
...
+			Kind:       KindIPPoolList,
+			APIVersion: "crd.projectcalico.org/v1",
+		},
+	}
+}
+
+type Mode string
+
+const (
+	Undefined   Mode = ""
```
caseydavenport, line 124: go formatting is off here?

pkg/apis/crd.projectcalico.org/v1/ippool.go
```diff
@@ -0,0 +1,140 @@
...
+
+	// The block size to use for IP address assignments from this pool. Defaults to 26 for IPv4 and 112 for IPv6.
+	BlockSize int `json:"blockSize,omitempty"`
+
+	// Allows IPPool to allocate for a specific node by label selector.
+	NodeSelector string `json:"nodeSelector,omitempty" validate:"omitempty,selector"`
+
+	// Deprecated: this field is only used for APIv1 backwards compatibility.
+	// Setting this field is not allowed, this field is for internal use only.
+	IPIP *IPIPConfiguration `json:"ipip,omitempty" validate:"omitempty,mustBeNil"`
```
caseydavenport, line 67: We don't need this field - it's internal only, so not relevant to the operator uses of IP pools.

pkg/apis/crd.projectcalico.org/v1/ippool.go
```diff
@@ -0,0 +1,140 @@
...
+
+	// Deprecated: this field is only used for APIv1 backwards compatibility.
+	// Setting this field is not allowed, this field is for internal use only.
+	NATOutgoingV1 bool `json:"nat-outgoing,omitempty" validate:"omitempty,mustBeFalse"`
```
caseydavenport, line 71: same here, I think we can remove this one

pkg/apis/crd.projectcalico.org/v1/register.go
```diff
@@ -0,0 +1,72 @@
...
+
+var (
+	GroupVersion  = schema.GroupVersion{Group: GroupName, Version: "v1"}
+	schemeBuilder = runtime.NewSchemeBuilder(addKnownTypes)
+
+	// Install is a function which adds this version to a scheme
+	Install = schemeBuilder.AddToScheme
+
+	// SchemeGroupVersion generated code relies on this name
+	// Deprecated
```
caseydavenport, line 39: Have you checked if we still need these deprecated vars here?

hack/boilerplate/boilerplate.go.txt
tmjd: No longer needed.

pkg/apis/crd.projectcalico.org/v1/ippool.go
tmjd: What do you think is off here?
It looks like it aligned on the `=`.
I tried removing the spaces in `Undefined   Mode` but my go fmt put them back.

pkg/controller/migration/convert/network.go
tmjd: Currently the code does what you've suggested @caseydavenport  with the added step of 1st trying to pick the one specified in the `*POOL_CIDR` variable. So I should drop the POOL_CIDR search?

pkg/controller/migration/convert/convert_test.go
```diff
@@ -6,46 +6,63 @@ import (
...
 )
 
 var _ = Describe("Parser", func() {
 	var ctx = context.Background()
+	var pool *crdv1.IPPool
+	var scheme *runtime.Scheme
+	BeforeEach(func() {
+		scheme = kscheme.Scheme
+		err := apis.AddToScheme(scheme)
+		Expect(err).NotTo(HaveOccurred())
```
ozdanborne, line 28: You might be able to do this in a `BeforeSuite` in the suite initializer. But I know how test initialization with ginkgo can be a bit of a rabbit hole so happy for you to just leave as is.

pkg/controller/migration/convert/network.go
```diff
@@ -437,16 +440,211 @@ func getCNIPlugin(c *components) (operatorv1.CNIPluginType, error) {
...
+// pools in the datastore.
+// We read the pools from the datastore and select the appropriate ones.
+// See selectInitialPool for details on which pool will be selected.
+// Since the operator only supports one v4 and one v6 only one of each will be picked
+// if they exist.
+func handleIPPools(c *components, install *Installation) error {
+	//pools, err := c.crdClientset.CrdV1().IPPools().List(metav1.ListOptions{})
+	//if err != nil {
+	//	return err
+	//}
```
ozdanborne, line 453: probably ok to nuke this now, right?
```suggestion
```

pkg/apis/crd.projectcalico.org/v1/ippool.go
caseydavenport: just looked weird to me, but I can believe that my eyes don't speak `gofmt`

pkg/apis/crd.projectcalico.org/v1/ippool.go
```diff
@@ -0,0 +1,132 @@
...
+
+type Mode string
+
+const (
+	Undefined   Mode = ""
+	Always           = "always"
+	CrossSubnet      = "cross-subnet"
+)
+
+type IPIPConfiguration struct {
```
caseydavenport, line 121: This struct is unused.

pkg/apis/crd.projectcalico.org/v1/ippool.go
caseydavenport: Are these constants actually used? The values actually look wrong to me.

caseydavenport: I think they're only used in the IPIPConfiguration, which itself is unused.

pkg/controller/migration/convert/network.go
caseydavenport: I don't mind much either way - I think if I was writing from scratch I wouldn't bother trying to compare the pools that exist with the ones in the POOL_CIDR variables, but not a problem per-se, just slightly simpler / easier to understand.

pkg/apis/crd.projectcalico.org/v1/ippool.go
```diff
@@ -0,0 +1,132 @@
+// Copyright (c) 2017 Tigera, Inc. All rights reserved.
```
ozdanborne, line 1: just noticed a few 2017 copyrights. worth updating?

PR #819, opened 2020-08-08
Add ClusterServiceVersions

lmm:
## Description

This PR adds the original v1.3.1 CSV and the most recent v1.5.2, v1.5.3 and v1.7.3 CSV's generated with `make gen-csv` and `make gen-bundle`.

However, the v1.7.3 csv has been modified outside of the make targets mentioned above. 4a98598 changes the crds to use apiextensions.k8s.io/v1beta1 so that the bundle will pass upstream validation tests.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/deploy/olm-catalog/tigera-operator/1.3.1/operator_v1_installation_crd.yaml
+++ b/deploy/olm-catalog/tigera-operator/1.3.1/operator_v1_installation_crd.yaml
@@ -1,0 +1,176 @@
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: installations.operator.tigera.io
+spec:
+  group: operator.tigera.io
+  names:
+    kind: Installation
+    listKind: InstallationList
+    plural: installations
+    singular: installation
+  scope: Cluster
+  subresources:
+    status: {}
+  validation:
+    openAPIV3Schema:
+      properties:
+        apiVersion:
+          description: 'APIVersion defines the versioned schema of this representation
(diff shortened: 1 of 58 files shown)


tmjd:
Does this need any changes now that we know the certification doesn't support v1 (only v1beta1 currently)?

lmm:
> Does this need any changes now that we know the certification doesn't support v1 (only v1beta1 currently)?

Sorry, I should've made this PR a draft once I got the updated info. Yeah, we may need to make some changes; still TBD though.

tmjd:
I noticed that there are no operator manifest, cluster role, or bindings as part of this. Can you remind me where those come from with OLM?

I'm wondering if we need to look at our tooling to generate the v1beta manifests. Do we have any idea how long we'll need to be using v1beta1 for the CRDs?

lmm:
>I noticed that there are no operator manifest, cluster role, or bindings as part of this. Can you remind me where those come from with OLM?

The operator deployment and rbac manifests are provided as inputs to the CSV generation process: https://github.com/tigera/operator/blob/master/hack/gen-csv/get-manifests.sh#L29-L31

 The operator deployment and rbac are then embedded into the resulting csv.

> I'm wondering if we need to look at our tooling to generate the v1beta manifests. Do we have any idea how long we'll need to be using v1beta1 for the CRDs?

When chatting with @caseydavenport , we were looking at this as just a one-off. I think that is still doable if we are okay with delaying v3.16 certification.

lmm:
Thanks @tmjd !

PR #842, opened 2020-08-14
Enable auto-detection based on CIDR

caseydavenport:
## Description

Adds operator support for this PR: https://github.com/projectcalico/node/pull/518

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


pkg/controller/installation/validation.go
```diff
@@ -233,6 +233,9 @@ func validateNodeAddressDetection(ad *operatorv1.NodeAddressAutodetection) error
 	if ad.FirstFound != nil && *ad.FirstFound {
 		numEnabled++
 	}
+	if len(ad.CIDRS) != 0 {
+		numEnabled++
```
tmjd, line 237: We should probably check that the CIRDs are valid shouldn't we?

pkg/apis/operator/v1/types.go
```diff
@@ -257,6 +257,10 @@ type NodeAddressAutodetection struct {
 	// specified IP or domain.
 	// +optional
 	CanReach string `json:"canReach,omitempty"`
+
+	// CIDRS enables IP auto-detection based on which addresses on the nodes are within
+	// one of the provided CIDRs.
+	CIDRS []string `json:"cidrs,omitempty"`
```
tmjd, line 263: Is there anything that can be added here so that there is CIDR validation done on these strings by the apiserver?

caseydavenport: I don't think so. There's no "validate as cidr" option on CRDs to my knowledge, and regexes for any possible IPv4 / v6 address are _disgusting_. we'll need to do it in code.

PR #846, opened 2020-08-18
block migration on invalid ipv6

ozdanborne:
## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/migration/convert/network_test.go
+++ b/pkg/controller/migration/convert/network_test.go
@@ -14,6 +14,7 @@ import (
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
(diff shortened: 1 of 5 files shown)


pkg/controller/migration/convert/network_test.go
```diff
@@ -76,6 +77,7 @@ var _ = Describe("Convert network tests", func() {
 			}, operatorv1.PluginGKE),
 		)
 		It("should convert AWS CNI install", func() {
+			Skip("eks not supported until typha prometheus metrics support is added")
```
ozdanborne, line 80: @tmjd I wonder if we want to just ignore user's prometheus metrics settings for typha and disable them until we have a way to migrate them. Otherwise, we're knocking out all clusters installed with our aws-cni-policy-only  manifests

tmjd: I'd say that ignoring is probably the route to go for right now but we should make it a priority to add that capability.

Issue #847, opened 2020-08-18
when installing operator over calico v3.15.x using k8s v1.18.x calico-node doesn't migrate


cwedgwood:
rbac lacks patch verb

tmjd:
The Calico v3.15 docs have been updated to include the patch verb for daemonsets and deployments.
https://docs.projectcalico.org/manifests/tigera-operator.yaml

PR #848, opened 2020-08-18
Add permission to patch daemonsets

cwedgwood:
Fixes: #847


tmjd:
Thanks for submitting this. The file you updated is expected to only be used on OpenShift with Operator Lifecycle Manager which is expected to never be used for a migration because Calico as a manifest install isn't supported on OpenShift and other reasons.
I've submitted https://github.com/projectcalico/calico/pull/3903 which is where I think the change should go and then once it is in we can pick it to v3.15.

I think we should go ahead and close this one unless there is a specific reason you think this file should be updated.

cwedgwood:
thanks, i wondered where it should go given the paths didn't "make sense"

much appreciated

PR #849, opened 2020-08-19
Migration: Do not add Resource if it already exists

tmjd:
## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/migration/convert/core.go
+++ b/pkg/controller/migration/convert/core.go
@@ -2,6 +2,7 @@ package convert
 
 import (
 	"fmt"
+	"reflect"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


pkg/controller/migration/convert/core.go
```diff
@@ -207,6 +205,36 @@ func handleCore(c *components, install *Installation) error {
...
+			ComponentName:        compName,
+			ResourceRequirements: rescReq.DeepCopy(),
+		})
+		return nil
+	}
+	if reflect.DeepEqual(existingRR, rescReq) {
+		return nil
+	}
+
+	return ErrIncompatibleCluster{fmt.Sprintf("ResourcesRequirements for component %s did not match between Installation and migration source", compName)}
```
ozdanborne, line 235: Maybe in the future we should just prefer whatever is set in the `Installation` but this seems ok for now while the "user-specified install" use case is still not supported

PR #873, opened 2020-09-09
Refactor CNI into package, remove dependency on projectcalico/cni-plugin

ozdanborne:
## Description

Changes:
- Moved CNI into own package
- Moved the various network fields into a new `NetworkComponents` struct instead of having them directly in the main Components struct.
- Updated main cni Parsing function to accept a string instead of the entire components struct, allowing for more isolated testing without having to initialize an entire components struct.
- Improved templating in the cni tests so the same main template can be used across all tests.
- Copied used types from projectcalico/cni-plugin, removing dependency on it

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/migration/cni/calicotypes.go
+++ b/pkg/controller/migration/cni/calicotypes.go
@@ -14,16 +14,13 @@ type CalicoConf struct {
 		IPv4Pools  []string `json:"ipv4_pools,omitempty"`
 		IPv6Pools  []string `json:"ipv6_pools,omitempty"`
 	} `json:"ipam,omitempty"`
-	// Args                 Args   `json:"args"`
-	MTU                  int    `json:"mtu"`
-	Nodename             string `json:"nodename"`
-	NodenameFileOptional bool   `json:"nodename_file_optional"`
-	DatastoreType        string `json:"datastore_type"`
-	EtcdEndpoints        string `json:"etcd_endpoints"`
-	EtcdDiscoverySrv     string `json:"etcd_discovery_srv"`
-	LogLevel             string `json:"log_level"`
-	// Policy               Policy `json:"policy"`
-	// Kubernetes           Kubernetes        `json:"kubernetes"`
+	MTU                  int               `json:"mtu"`
+	Nodename             string            `json:"nodename"`
+	NodenameFileOptional bool              `json:"nodename_file_optional"`
+	DatastoreType        string            `json:"datastore_type"`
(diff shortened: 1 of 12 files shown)


pkg/controller/migration/cni/calicotypes.go
```diff
@@ -0,0 +1,46 @@
...
+	// Args                 Args   `json:"args"`
+	MTU                  int    `json:"mtu"`
+	Nodename             string `json:"nodename"`
+	NodenameFileOptional bool   `json:"nodename_file_optional"`
+	DatastoreType        string `json:"datastore_type"`
+	EtcdEndpoints        string `json:"etcd_endpoints"`
+	EtcdDiscoverySrv     string `json:"etcd_discovery_srv"`
+	LogLevel             string `json:"log_level"`
+	// Policy               Policy `json:"policy"`
+	// Kubernetes           Kubernetes        `json:"kubernetes"`
```
ozdanborne, line 26: There's a few fields on this struct which are not currently used by our migration code. As such, it seemed unnecessary to copy in the type definitions. I've commented them out in this file, but I think I'll remove them entirely.

tmjd: I think that is ok even when the fields are in the CNI config because it is an extra step to make sure unknown fields cause an error with unmarshalling.

tmjd (changes requested):
Looks good, just a minor request.
I wonder if as part of the work you're doing if we should consider more log messages instead of removing them. I would think we should err on the side of more logs since this code won't always be running and it may be difficult to figure out what happened.

pkg/controller/migration/convert/convert.go
```diff
@@ -103,11 +94,33 @@ func getComponents(ctx context.Context, client client.Client) (*components, erro
...
 	return comps, err
 }
 
+// loadCNI pulls the CNI network config from it's env var source within components
+// and then returns the parsed data.
+func loadCNI(comps *components) (nc cni.NetworkComponents, err error) {
+	// do some upfront processing of CNI by loading it into comps
+	c := getContainer(comps.node.Spec.Template.Spec, containerInstallCNI)
+	if c == nil {
+		return
```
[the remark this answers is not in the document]
tmjd: In addition to the logs, I was also suggesting that we should keep the comment that no install cni container is allowed because it is not expected for non-Calico CNI. I don't think that is necessary though.

tmjd (approved):
LGTM

PR #875, opened 2020-09-10
Add volume mounts for CNI logging and change the default CNI log level to INFO

mgleung:
## Description

Adds the volume mounts for the CNI plugin logs so that the logs are accessible on each node in the cluster. This also changes the default CNI log level to INFO so information will be written by default.

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/apis/operator/v1/types.go
+++ b/pkg/apis/operator/v1/types.go
@@ -374,7 +374,7 @@ func (cp CNIPluginType) String() string {
 }
 
 type CNISpec struct {
-	// Specifies the CNI plugin that will be used in the Calico or Calico Enterprise installation.
+	// CNIPluginType specifies the CNI plugin that will be used in the Calico or Calico Enterprise installation.
 	// For KubernetesProvider GKE, this field defaults to GKE. For KubernetesProvider AKE, this field
 	// defaults to AzureVNET. For KubernetesProvider EKS, this field defaults to AmazonVPC. For all other
 	// KubernetesProviders this field defaults to Calico.
@@ -385,6 +385,12 @@ type CNISpec struct {
 	// Default: Calico
 	// +kubebuilder:validation:Enum=Calico;GKE;AmazonVPC;AzureVNET
 	Type CNIPluginType `json:"type"`
+
+	//LogLevel specifies the log level the CNI plugin is configured to run at.
+	// Default: INFO
+	// +optional
+	// +kubebuilder:validation:Enum=ERROR;WARNING;INFO;DEBUG
+	LogLevel string `json:"log_level"`
 }
 
 // InstallationStatus defines the observed state of the Calico or Calico Enterprise installation.
--- a/pkg/apis/operator/v1/types.go
+++ b/pkg/apis/operator/v1/types.go
@@ -374,7 +374,7 @@ func (cp CNIPluginType) String() string {
 }
 
 type CNISpec struct {
-	// CNIPluginType specifies the CNI plugin that will be used in the Calico or Calico Enterprise installation.
+	// Specifies the CNI plugin that will be used in the Calico or Calico Enterprise installation.
 	// For KubernetesProvider GKE, this field defaults to GKE. For KubernetesProvider AKE, this field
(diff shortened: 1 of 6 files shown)


pkg/apis/operator/v1/types.go
```diff
@@ -385,6 +385,12 @@ type CNISpec struct {
 	// Default: Calico
 	// +kubebuilder:validation:Enum=Calico;GKE;AmazonVPC;AzureVNET
 	Type CNIPluginType `json:"type"`
+
+	//LogLevel specifies the log level the CNI plugin is configured to run at.
+	// Default: INFO
+	// +optional
+	// +kubebuilder:validation:Enum=ERROR;WARNING;INFO;DEBUG
+	LogLevel string `json:"log_level"`
```
caseydavenport, line 393: Need to think about this some more. I think for now I would recommend not including the API changes in this PR, so that they are separate from the more important change of mounting in the files. 

As an aside, this should probably be its own type, with values of format `Error,Warning,Info,Debug` (note capitalization).

pkg/render/node_test.go
```diff
@@ -652,8 +663,11 @@ var _ = Describe("Node rendering tests", func() {
 		amazonVPCInstalllation := &operator.Installation{
 			Spec: operator.InstallationSpec{
 				KubernetesProvider: operator.ProviderEKS,
-				CNI:                &operator.CNISpec{Type: operator.PluginAmazonVPC},
-				FlexVolumePath:     "/usr/libexec/kubernetes/kubelet-plugins/volume/exec/",
+				CNI: &operator.CNISpec{
+					Type:     operator.PluginAmazonVPC,
+					LogLevel: "INFO",
```
caseydavenport, line 668: I'd expect this to be an invalid config (we don't support the LogLevel field with non-Calico CNI plugins).

That said, I think we should remove LogLevel from this PR for now while we figure out how we want to expose CNI config more generally

pkg/apis/operator/v1/types.go
mgleung: The capitalization is weird because our docs list the options in all uppercase whereas we seem to use `info` in other places. Not sure what the values really should be.

mgleung: I can remove the API for now, I don't think anyone will be changing the log level (for now) so it should be fine. I was just hoping to add the configuration just in case, but it's not entirely thought through at this point.

pkg/render/node.go
```diff
@@ -653,6 +656,7 @@ func (c *nodeComponent) cniContainer() v1.Container {
 	cniVolumeMounts := []v1.VolumeMount{
 		{MountPath: "/host/opt/cni/bin", Name: "cni-bin-dir"},
 		{MountPath: "/host/etc/cni/net.d", Name: "cni-net-dir"},
+		{MountPath: "/host/var/log/calico/cni", Name: "cni-log-dir"},
```
tmjd, line 659: Why does the container need this MountPath? Is there some action that the install does that needs to access the log dir?

tmjd (commented):
One comment (I'm assuming you'll be dropping the log level config change).

pkg/apis/operator/v1/types.go
caseydavenport: Well, this is the operator API, so it doesn't matter really what the CNI plugin says, we need to be consistent within the operator API itself. 

I think it does toLower() on it anyway :D

pkg/render/node.go
mgleung: Yeah, this isn't needed. I mistook the installer container with where it was actually running.

tmjd (commented):
Is the title still correct? Specifically the `change the default CNI log level to INFO`. I would have guessed previously we were logging at INFO level.

mgleung:
The log level needed to be set specifically to `Info` for it to actually log at that level. I didn't add it last time so I had to set it explicitly in this PR.

tmjd (approved):
LGTM

PR #876, opened 2020-09-10
verify and set resoure requirements for individual fields

Suraiya-Hameed:
## Description

**Issue:** Non specified LogStorage ResourceRequirements fields set to 0 if resourceRequirements block is specified

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/logstorage.go
+++ b/pkg/render/logstorage.go
@@ -403,13 +403,7 @@ func (es elasticsearchComponent) pvcTemplate() corev1.PersistentVolumeClaim {
 		Spec: corev1.PersistentVolumeClaimSpec{
 			AccessModes: []corev1.PersistentVolumeAccessMode{corev1.ReadWriteOnce},
 			Resources: corev1.ResourceRequirements{
-				Limits: corev1.ResourceList{
-					"cpu":    resource.MustParse("2"),
-					"memory": resource.MustParse("3Gi"),
-				},
 				Requests: corev1.ResourceList{
-					"cpu":     resource.MustParse("1"),
-					"memory":  resource.MustParse("2Gi"),
 					"storage": resource.MustParse("10Gi"),
 				},
 			},
@@ -420,14 +414,7 @@ func (es elasticsearchComponent) pvcTemplate() corev1.PersistentVolumeClaim {
 	// If the user has provided resource requirements, then use the user overrides instead
 	if es.logStorage.Spec.Nodes != nil && es.logStorage.Spec.Nodes.ResourceRequirements != nil {
 		userOverrides := *es.logStorage.Spec.Nodes.ResourceRequirements
-
-		// If the user provided overrides does not contain a storage quantity, then we still need to
-		// set a default
-		if _, ok := userOverrides.Requests["storage"]; !ok {
-			userOverrides.Requests["storage"] = resource.MustParse("10Gi")
-		}
-
-		pvcTemplate.Spec.Resources = userOverrides
+		overridePvcRequirements(&pvcTemplate.Spec.Resources, userOverrides)
 	}
 
 	return pvcTemplate
@@ -460,16 +447,7 @@ func (es elasticsearchComponent) podTemplate() corev1.PodTemplateSpec {
 	// If the user has provided resource requirements, then use the user overrides instead
 	if es.logStorage.Spec.Nodes != nil && es.logStorage.Spec.Nodes.ResourceRequirements != nil {
 		userOverrides := *es.logStorage.Spec.Nodes.ResourceRequirements
-		esContainer.Resources = corev1.ResourceRequirements{
-			Limits: corev1.ResourceList{
-				"cpu":    *userOverrides.Limits.Cpu(),
-				"memory": *userOverrides.Limits.Memory(),
-			},
-			Requests: corev1.ResourceList{
-				"cpu":    *userOverrides.Requests.Cpu(),
-				"memory": *userOverrides.Requests.Memory(),
-			},
-		}
+		overrideResourceRequirements(&esContainer.Resources, userOverrides)
 
 		// Now extract the memory request value to compute the recommended heap size for ES container
 		recommendedHeapSize := memoryQuantityToJVMHeapSize(esContainer.Resources.Requests.Memory())
@@ -665,7 +643,7 @@ func (es elasticsearchComponent) nodeSets() []esv1.NodeSet {
 
 	var nodeSets []esv1.NodeSet
 	if nodeConfig.NodeSets == nil || len(nodeConfig.NodeSets) < 1 {
-		nodeSet := es.nodeSetTemplate()
+		nodeSet := es.nodeSetTemplate(pvcTemplate)
 		nodeSet.Name = nodeSetName(pvcTemplate)
 		nodeSet.Count = int32(nodeConfig.Count)
 		nodeSet.PodTemplate = es.podTemplate()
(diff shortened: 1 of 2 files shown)


pkg/render/logstorage.go
```diff
@@ -403,13 +403,7 @@ func (es elasticsearchComponent) pvcTemplate() corev1.PersistentVolumeClaim {
 		Spec: corev1.PersistentVolumeClaimSpec{
 			AccessModes: []corev1.PersistentVolumeAccessMode{corev1.ReadWriteOnce},
 			Resources: corev1.ResourceRequirements{
-				Limits: corev1.ResourceList{
-					"cpu":    resource.MustParse("2"),
-					"memory": resource.MustParse("3Gi"),
-				},
```
Suraiya-Hameed, line 409: Removing cpu and memory resources from PVC spec as they are not needed, only storage is valid for PVC.
https://kubernetes.io/docs/concepts/storage/persistent-volumes/#introduction

Brian-McM: Good to clean out the cruft.

marvin-tigera:
OK, Adding "merge-when-ready" label and will merge the PR when it's ready.

pkg/render/logstorage.go
```diff
@@ -1419,3 +1395,60 @@ func (es elasticsearchComponent) kibanaPodSecurityPolicy() *policyv1beta1.PodSec
...
+
+// If user has set ResourceRequirements, replace individual field's default value with user's value.
+// If user provided both Limits and Requests, use them.
+// As ResourceRequirements.Limits must be >= ResourceRequirements.Requests,
+// if user provided just Limits, and Limits is <= default Requests, set Requests value as user's Limits value,
+// if user provided just Requests, and Requests is >= default Limits, set Limits value as user's Requests value.
+func overrideResourceRequirements(defaultReq *corev1.ResourceRequirements, userOverrides corev1.ResourceRequirements) {
+	if userOverrides.Limits.Cpu().Value() != 0 {
+		defaultReq.Limits["cpu"] = *userOverrides.Limits.Cpu()
+		if userOverrides.Requests.Cpu().Value() == 0 && defaultReq.Requests.Cpu().Value() > userOverrides.Limits.Cpu().Value() {
```
Brian-McM, line 1407: I don't think we should overwrite the value if the user has specifically set the limit. If the user sets a bad value we should either set it as that's there intent, or fail so they know that something went wrong. 

If kuberentes does not allow a limit to be smaller the the requests than it will fail so we don't need to fail. If kuberenets does allow it then we should allow it as it may be the users intent.

pkg/render/logstorage.go
```diff
@@ -1419,3 +1395,60 @@ func (es elasticsearchComponent) kibanaPodSecurityPolicy() *policyv1beta1.PodSec
...
 	return psp
 }
+
+// If user has set ResourceRequirements, replace individual field's default value with user's value.
+// If user provided both Limits and Requests, use them.
+// As ResourceRequirements.Limits must be >= ResourceRequirements.Requests,
+// if user provided just Limits, and Limits is <= default Requests, set Requests value as user's Limits value,
+// if user provided just Requests, and Requests is >= default Limits, set Limits value as user's Requests value.
+func overrideResourceRequirements(defaultReq *corev1.ResourceRequirements, userOverrides corev1.ResourceRequirements) {
+	if userOverrides.Limits.Cpu().Value() != 0 {
```
Brian-McM, line 1405: I think we should check if the values was actually set, not that it's 0. If a user wants to set it to 0, they should be allowed. It may be that ends up being a special value in the future, but regardless we should not override user intent.

Brian-McM (changes requested):

marvin-tigera:

pkg/render/logstorage.go
Brian-McM: We chatted an I had a misunderstanding about this, the limits are only updated if there's requests but no limits set.

marvin-tigera:
OK, Adding "merge-when-ready" label and will merge the PR when it's ready.

pkg/render/logstorage.go
```diff
@@ -1419,3 +1395,60 @@ func (es elasticsearchComponent) kibanaPodSecurityPolicy() *policyv1beta1.PodSec
 	psp.GetObjectMeta().SetName("tigera-kibana")
 	return psp
 }
+
+// If user has set ResourceRequirements, replace individual field's default value with user's value.
+// If user provided both Limits and Requests, use them.
+// As ResourceRequirements.Limits must be >= ResourceRequirements.Requests,
+// if user provided just Limits, and Limits is <= default Requests, set Requests value as user's Limits value,
+// if user provided just Requests, and Requests is >= default Limits, set Limits value as user's Requests value.
+func overrideResourceRequirements(defaultReq *corev1.ResourceRequirements, userOverrides corev1.ResourceRequirements) {
```
Brian-McM, line 1404: Instead of using some pointer magic, could just return a resource requiment object? There's no real benefit over returning an object, and these defaults are no long defaults after this.

If I was to factor out this logic into a function, I would have a declaration like this:
`func calculateResourceRequirements(defaultReq *corev1.ResourceRequirements, userProvided *corev1.ResourceRequirements) corev1.ResourceRequirements`.

Could you also clean up the function comment, it seems like maybe this should be bullet pointed, with a starting comment like "calculateResourceRequirements combines the default resources with the user specified resources as with the following rules:".

pkg/render/logstorage.go
```diff
@@ -1419,3 +1395,60 @@ func (es elasticsearchComponent) kibanaPodSecurityPolicy() *policyv1beta1.PodSec
...
+		}
+	}
+}
+
+// Storage is the only valid resource requirement for PVC, extract that if user has set it.
+// If user provided both Limits and Requests, use them
+// As ResourceRequirements.Limits must be >= ResourceRequirements.Requests,
+// If user has provided just Limits, and Limits is <= default Requests, set Requests value as user's Limits value.
+// We don not set default Limits for storage, so don't have to handle case where user has set only Requests.
+func overridePvcRequirements(defaultReq *corev1.ResourceRequirements, userOverrides corev1.ResourceRequirements) {
```
Brian-McM, line 1436: Similar to the override function above, I don't think we should be modifying the object through a parameter pointer, we should return something the resource requirements.

pkg/render/logstorage.go
```diff
@@ -1419,3 +1395,60 @@ func (es elasticsearchComponent) kibanaPodSecurityPolicy() *policyv1beta1.PodSec
...
+// If user has set ResourceRequirements, replace individual field's default value with user's value.
+// If user provided both Limits and Requests, use them.
+// As ResourceRequirements.Limits must be >= ResourceRequirements.Requests,
+// if user provided just Limits, and Limits is <= default Requests, set Requests value as user's Limits value,
+// if user provided just Requests, and Requests is >= default Limits, set Limits value as user's Requests value.
+func overrideResourceRequirements(defaultReq *corev1.ResourceRequirements, userOverrides corev1.ResourceRequirements) {
+	if _, ok := userOverrides.Limits["cpu"]; ok {
+		defaultReq.Limits["cpu"] = *userOverrides.Limits.Cpu()
+		if _, ok := userOverrides.Requests["cpu"]; !ok && defaultReq.Requests.Cpu().Value() > userOverrides.Limits.Cpu().Value() {
+			defaultReq.Requests["cpu"] = *userOverrides.Limits.Cpu()
```
Brian-McM, line 1408: Along the lines of my other "anti override" comments, since you are modifying this default object you lose what was originally there. I see two problems with this:
1. In reviewing this I need to make sure that you are not accidentally comparing the new value of the defaults when you meant to compare the old, none existent values
2. If somebody needs check the original defaults for some other reason as an extension of the abilities here there is a very big chance of a bug being implemented.
3. If there is code outside of this function after this function that for some reason needs the original default values then a bug could be easily introduced.

I think you should create a separate, empty resource and populate the values with either the defaults or the user provided resources.

Brian-McM (changes requested):
I'm a little concerned about the pointer passing and overriding in the "override" functions. I added some comments expressing my concerns

marvin-tigera:

marvin-tigera:
OK, Adding "merge-when-ready" label and will merge the PR when it's ready.

Brian-McM (approved):
Looks good!

PR #879, opened 2020-09-14
Fix wrong permission on a volume mount

mgleung:
## Description

Fixes a wrong permission on a CNI log volume mount added in https://github.com/tigera/operator/pull/875

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/node.go
+++ b/pkg/render/node.go
@@ -772,7 +772,7 @@ func (c *nodeComponent) nodeVolumeMounts() []v1.VolumeMount {
 		}
 		nodeVolumeMounts = append(nodeVolumeMounts, extraNodeMounts...)
 	} else if c.cr.Spec.CNI.Type == operator.PluginCalico {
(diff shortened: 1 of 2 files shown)


tmjd:
Can you give some detail on why it can't/shouldn't be ReadOnly? It made sense to me that it was ReadOnly so I'm surprised to see that removed.

mgleung:
Sure thing, The CNI plugin also needs to write out those logs and since the CNI plugin is run from within the `calico-node` container, the `calico-node` container volume mount needs to allow the directory to be writable. It is definitely needs read permissions for the diagnostics, but since we collect and write out all the information from the same pod, it'll need both read and write permissions.

mgleung:
I take it back, the CNI plugin is just installed from the init container and run on the node so this should be fine being read-only. We don't need this.

PR #882, opened 2020-09-16
Move init image functionality into the operator itself

tmjd:
## Description

On OpenShift on AWS we ensure the SecurityGroups for the masters and workers allow BGP, IPIP, and Typha traffic. We have been using an `operator-init` image that the operator has been launching as a job. This PR removes that init image and instead the adds to the operator a flag that does what the init image did to configure the Security Groups.
We'll still launch a job for doing the AWS SG setup but it will run the operator image with the new flag `--aws-sg-setup`. This is necessary because the DNS configuration of the operator deployment does not function before networking is up but the Job does have functional DNS before the cluster DNS is working.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


tmjd:
This does not work yet because the same DNS config was not used as on the operator as the init image. My initial tests with this were running it on a cluster that was already functional and because of that did not notice the DNS problem.

cmd/manager/main.go
```diff
@@ -15,12 +15,14 @@
...
 
 import (
+	"context"
 	"flag"
 	"fmt"
 	"net/url"
 	"os"
 	"runtime"
 
+	"github.com/tigera/operator/pkg/awsSGSetup"
```
ozdanborne, line 25: this casing scares me. should this be `aws-sg-setup`?

cmd/manager/main.go
```diff
@@ -115,8 +115,39 @@ func main() {
...
+		if err != nil {
+			log.Error(err, "")
+			os.Exit(1)
+		}
+		os.Exit(0)
+	}
+
 	// Run the Daemon
 	daemon.Main()
 }
```
ozdanborne, line 153:
```suggestion
	} else {
		// Run the Daemon
		daemon.Main()
	}
}
```

pkg/awsSGSetup/aws_security_group_setup.go
```diff
@@ -0,0 +1,320 @@
...
+
+// setupAWSSecurityGroups updates the master and worker security groups used in an
+// OpenShift AWS setup. It sets a time that should be checked before attempting
+// to do another call to setup.
+func SetupAWSSecurityGroups(ctx context.Context, client client.Client) error {
+	// Grab ConfigMap kube-system aws-creds
+	//		get aws_access_key_id and aws_secret_access_key
+	awsKeyId, awsSecret, err := getAWSCreds(ctx, client)
+	if err != nil {
+		return fmt.Errorf("failed to update AWS SecurityGroups: %s", err.Error())
```
ozdanborne, line 51: nitpick, you can use `%v` instead of calling `.Error()`

```suggestion
		return fmt.Errorf("failed to update AWS SecurityGroups: %v", err)
```
Any reason you've done it the way you have? It's just habit for me to use `%v`, I wonder if it's habit for you to do that^

Note that in go 1.14, you can now use `%w` to wrap the error. Which might be the correct thing to do now

pkg/awsSGSetup/aws_security_group_setup.go
```diff
@@ -0,0 +1,320 @@
...
+	secretByte, ok := creds.Data["aws_secret_access_key"]
+	if !ok {
+		return "", "", fmt.Errorf("aws-creds ConfigMap does not have key aws_secret_access_key")
+	}
+	return string(idByte), string(secretByte), nil
+}
+
+//readMeta queries the metadata at the specified subpath on the instance.
+func readMeta(subpath string) (string, error) {
+	u, err := url.Parse("http://169.254.169.254/latest/meta-data")
```
ozdanborne, line 156: Did you end up investigating the `https://docs.aws.amazon.com/sdk-for-go/api/aws/ec2metadata/` and/or the `https://docs.aws.amazon.com/sdk-for-go/api/aws/endpoints/` apis?

pkg/awsSGSetup/aws_security_group_setup.go
```diff
@@ -0,0 +1,320 @@
...
+				GroupId: aws.String(s.srcSGId),
+			}},
+			IpProtocol: aws.String(s.protocol),
+			FromPort:   s.port,
+			ToPort:     s.port,
+		}})
+		_, err := cli.AuthorizeSecurityGroupIngress(in)
+		if err != nil {
+			log.Error(err, "Failed to add Ingress rule", "toSG", toSG, "ingressSrc", s.String())
+			return err
```
ozdanborne, line 314: If this function returns an error, that error is ultimately printed to stdout. I think you can just return this error instead of logging it. something like...
```suggestion
			return fmt.Errorf("Failed to add Ingress rule to SG '%s' for traffic from '%s': %v", "err, toSG, s)
```

pkg/awsSGSetup/aws_security_group_setup.go
```diff
@@ -0,0 +1,320 @@
...
+					break
+				}
+			}
+			if skip {
+				break
+			}
+		}
+		if skip {
+			continue
+		}
```
ozdanborne, line 302: This code is pretty wacky. I wonder if you could clean it up if you added a function that checks if two ingress rules are equal and/or a function that checks if a rule already exists in a slice of rules.

pkg/awsSGSetup/aws_security_group_setup.go
```diff
@@ -0,0 +1,320 @@
...
+)
+
+var log = logf.Log.WithName("AWS_SG_Setup")
+var TRACE = 7
+var DEBUG = 5
+
+// setupAWSSecurityGroups updates the master and worker security groups used in an
+// OpenShift AWS setup. It sets a time that should be checked before attempting
+// to do another call to setup.
+func SetupAWSSecurityGroups(ctx context.Context, client client.Client) error {
```
ozdanborne, line 46: no tests? :scream_cat:

cmd/manager/main.go
tmjd: Why do you think this is better? we often return early out of a function and this feels like the same kind of thing.

pkg/awsSGSetup/aws_security_group_setup.go
tmjd: I'm not sure. I think it is habit but I've started switching to using `%v`. I did some testing after I saw you making some changes like this recently and didn't find a difference so I've started doing it the way you suggested, I just forgot here.

pkg/awsSGSetup/aws_security_group_setup.go
tmjd: I'm confused. I had switched to use ec2metadata but obviously lost that switch. Dang it. :angry:

pkg/awsSGSetup/aws_security_group_setup.go
tmjd: Actually I had made that change to use `%v` but that was another thing I lost at some point. :angry:

pkg/awsSGSetup/aws_security_group_setup.go
tmjd: Yeah this is pretty dependent on the APIs k8s client and more so the AWS go sdk. It would either require mocking those which would probably be a lot of work or isolating the API clients from the other code and only testing the "other" code which would be fairly minimal.

pkg/awssgsetup/aws_security_group_setup.go
```diff
@@ -0,0 +1,286 @@
...
+	secretByte, ok := creds.Data["aws_secret_access_key"]
+	if !ok {
+		return "", "", fmt.Errorf("aws-creds ConfigMap does not have key aws_secret_access_key")
+	}
+	return string(idByte), string(secretByte), nil
+}
+
+// getVPCid gets the VPC id by querying the instance metadata.
+// curl http://169.254.169.254/latest/meta-data/network/interfaces/macs/ | head -n 1
+// curl http://169.254.169.254/latest/meta-data/network/interfaces/macs/$mac/vpc-id
```
ozdanborne, line 162:
```suggestion
```

ozdanborne (approved):
Awesome, using that aws package makes this code a million times better :+1:

PR #884, opened 2020-09-18
Cleanup tests

ozdanborne:
## Description

- Move prometheus tests into `core_test.go` since the prometheus processing lives in `core.go`
- Fixup a convert.go test that must have been written before we had helper methods for initializing valid components because it checked for an 'errincompatiblecluster' when it wasn't necessary.
- move 'components' out of `convert.go` and into its own `components.go`

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


pkg/controller/migration/convert/components_test.go
```diff
@@ -0,0 +1,169 @@
+package convert
```
ozdanborne, line 1: I wonder if this file should be called `testutils_test.go` ...

pkg/controller/migration/convert/convert_test.go
```diff
@@ -41,15 +41,9 @@ var _ = Describe("Parser", func() {
...
-			ObjectMeta: v1.ObjectMeta{
-				Name:      "calico-node",
-				Namespace: "kube-system",
-			},
-		}, emptyKubeControllerSpec(), pool, emptyFelixConfig())
+		c := fake.NewFakeClientWithScheme(scheme, emptyNodeSpec(), emptyKubeControllerSpec(), pool, emptyFelixConfig())
 		err := Convert(ctx, c, &operatorv1.Installation{})
-		// though it will detect an install, it will be in the form of an incompatible-cluster error
-		Expect(err).To(BeAssignableToTypeOf(ErrIncompatibleCluster{}))
+		Expect(err).ToNot(HaveOccurred())
```
tmjd, line 46: This is inverting the test. Why did that change?

tmjd (commented):
I think there may be one issue here with a test.

pkg/controller/migration/convert/convert_test.go
ozdanborne: I intentionally flipped it. It's hard to explain, but I'll try to.

Part of the reason this is convoluted is because `Convert` used to return an error and an Installation resource. Nowadays, we pass in  a pointer to the Installation resource we've instantiated.

But if we go back in time to when it used to return both...

The test is basically checking that `Convert` doesn't return `nil, nil` when a calico-node daemonset exists. However, this test was written before we had helper methods for initializing valid components. So it passed a calico-node daemonset into the fake client which it knew would cause an ErrIncompatibleCluster because it's missing a ton of required spec fields, and then asserted that `Convert` is detecting that daemonset and not just returning `nil, nil`.

tmjd: Ok so since this test was changed to use the `emptyNodeSpec()` the node is now valid enough that an error is not expected.

pkg/controller/migration/convert/components_test.go
tmjd: Yeah I think a different name would be good.

tmjd (approved):
LGTM, if you want to change the name of components_test.go I'm onboard with that too.

PR #915, opened 2020-09-22
Degrade Calico status on typha autoscale issues

Brian-McM:
This commit ties the Calico tigera status to the typha autoscalar, degrading the status if there's an issue.

An error check that has been added is to make sure that there are enough linux nodes available to schedule the required typha pods on, and now if there's not, an error is logged and the Calico tigera status is set to degraded

If the auto scalar degraded the status, the core controller will see it if it reconciles and trigger the autoscalar run immediately. If this run fails the controller will set the degraded status again and try to re reconcile later. Note the the ticker in typha is reset if the run is triggered by the controller.

## For PR author

- [x] Tests for change.

## For PR reviewers

- [x] Appropriate labels:

### DIFF
--- a/pkg/controller/installation/core_controller.go
+++ b/pkg/controller/installation/core_controller.go
@@ -87,20 +87,24 @@ func newReconciler(mgr manager.Manager, opts options.AddOptions) (*ReconcileInst
 	if err != nil {
 		return nil, fmt.Errorf("Failed to initialize Namespace migration: %v", err)
 	}
+
+	statusManager := status.New(mgr.GetClient(), "calico")
+
 	r := &ReconcileInstallation{
 		config:               mgr.GetConfig(),
 		client:               mgr.GetClient(),
 		scheme:               mgr.GetScheme(),
 		watches:              make(map[runtime.Object]struct{}),
 		autoDetectedProvider: opts.DetectedProvider,
-		status:               status.New(mgr.GetClient(), "calico"),
-		typhaAutoscaler:      newTyphaAutoscaler(mgr.GetClient()),
+		status:               statusManager,
+		typhaAutoscaler:      newTyphaAutoscaler(mgr.GetClient(), statusManager),
 		namespaceMigration:   nm,
 		amazonCRDExists:      opts.AmazonCRDExists,
 		enterpriseCRDsExist:  opts.EnterpriseCRDExists,
 	}
 	r.status.Run()
-	r.typhaAutoscaler.run()
+	r.typhaAutoscaler.start()
+
 	return r, nil
 }
 
@@ -548,6 +552,15 @@ func (r *ReconcileInstallation) Reconcile(request reconcile.Request) (reconcile.
 		if err = r.client.Status().Update(ctx, instance); err != nil {
 			r.SetDegraded("Failed to write default status", err, reqLogger)
 			return reconcile.Result{}, err
+		}
+	}
+
+	// If the autoscalar is degraded then trigger a run and recheck the degraded status. If it is still degraded after the
+	// the run the reset the degraded status and requeue the request.
+	if r.typhaAutoscaler.isDegraded() {
+		if err := r.typhaAutoscaler.triggerRun(); err != nil {
+			r.SetDegraded("Failed to scale typha", err, reqLogger)
+			return reconcile.Result{RequeueAfter: 30 * time.Second}, nil
 		}
 	}
 
--- a/pkg/controller/installation/typha_autoscaler.go
+++ b/pkg/controller/installation/typha_autoscaler.go
@@ -18,6 +18,8 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/tigera/operator/pkg/controller/status"
+
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 
 	"github.com/tigera/operator/pkg/common"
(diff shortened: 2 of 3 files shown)


pkg/controller/installation/typha_autoscaler_test.go
```diff
@@ -95,12 +103,44 @@ var _ = Describe("Test typha autoscaler ", func() {
...
+		_ = createNode(c, "node1", map[string]string{"kubernetes.io/os": "linux"})
+		_ = createNode(c, "node2", map[string]string{"kubernetes.io/os": "linux"})
+		_ = createNode(c, "node3", map[string]string{"kubernetes.io/os": "window"})
+		_ = createNode(c, "node4", map[string]string{"kubernetes.io/os": "window"})
+
+		// Create the autoscaler and run it
+		ta := newTyphaAutoscaler(c, statusManager, typhaAutoscalerPeriod(10*time.Millisecond))
+		ta.start()
+
+		time.Sleep(1 * time.Second)
```
tmjd, line 131: Instead of using a sleep here it should work just to do `ta.isDegraded()`, that would ensure that autoscale was ran and SetDegraded would be set if degraded.

pkg/controller/installation/typha_autoscaler.go
```diff
@@ -65,10 +70,14 @@ func typhaAutoscalerPeriod(syncPeriod time.Duration) typhaAutoscalerOption {
...
-func newTyphaAutoscaler(client client.Client, options ...typhaAutoscalerOption) *typhaAutoscaler {
-	ta := new(typhaAutoscaler)
-	ta.client = client
-	ta.syncPeriod = defaultTyphaAutoscalerSyncPeriod
+func newTyphaAutoscaler(client client.Client, statusManager status.StatusManager, options ...typhaAutoscalerOption) *typhaAutoscaler {
+	ta := &typhaAutoscaler{
+		client:         client,
+		statusManager:  statusManager,
+		syncPeriod:     defaultTyphaAutoscalerSyncPeriod,
+		triggerRunChan: make(chan chan error, 100),
```
tmjd, line 78: Should we really be making these channels with 100 buffer? It should be fine to be 1 for both.

tmjd (commented):
My thoughts

pkg/controller/installation/core_controller.go
```diff
@@ -551,6 +555,15 @@ func (r *ReconcileInstallation) Reconcile(request reconcile.Request) (reconcile.
 		}
 	}
 
+	// If the autoscalar is degraded then trigger a run and recheck the degraded status. If it is still degraded after the
+	// the run the reset the degraded status and requeue the request.
+	if r.typhaAutoscaler.isDegraded() {
+		if err := r.typhaAutoscaler.triggerRun(); err != nil {
+			r.SetDegraded("Failed to scale typha", err, reqLogger)
+			return reconcile.Result{RequeueAfter: 30 * time.Second}, err
```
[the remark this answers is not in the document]
Brian-McM: Ah right! You would think I'd remember this as I've been burned by it so many times lol.

pkg/controller/installation/typha_autoscaler.go
```diff
@@ -99,29 +108,91 @@ func (t *typhaAutoscaler) getExpectedReplicas(nodes int) int {
...
-					typhaLog.Error(err, "Could not scale Typha deployment")
+					// Return the error so the "caller" can decided what to do with the error
+					errCh <- err
+				} else {
+					degraded = false
 				}
+
+				close(errCh)
+
+				ticker.Stop()
```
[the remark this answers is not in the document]
Brian-McM: My reasoning was to make sure that it doesn't try to immediately autoscale again if it was triggered manually. If there was 5 seconds left on the ticker and this was triggered manually the autoscale would be kicked off again almost immediately.

It's not _that_ big a deal if we don't reset the ticker, but on the other hand it's really easy to do so I figured why not.

pkg/controller/installation/typha_autoscaler_test.go
Brian-McM: Technically there's still a race there, although it's a bit smaller. The initial autoscale run is done in the go routine, so the degraded check could theoretically beat the run.

I think putting this to a milli second then doing the degraded check would be good, just to ensure this. WDYT?

pkg/controller/installation/typha_autoscaler.go
Brian-McM: Meh, probably not. I guess we don't need to, the go routine is essentially locked while triggerRun and isDegraded are being called anyway.

Brian-McM: I set it to 0, which means it blocks until the object is read off of the channel. I think that's probably what we want.

pkg/controller/installation/typha_autoscaler.go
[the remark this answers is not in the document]
tmjd: Yeah I guess getting the list of nodes might be significant in a large cluster. So I'm good with keeping it.

pkg/controller/installation/typha_autoscaler_test.go
tmjd: The ta.isDegraded() will block until the loop and select happens once and writes the degraded status back. Before the loop even runs we do an autoscale and set degraded so I don't think there is a race here.

tmjd (approved):
LGTM

marvin-tigera:

marvin-tigera:

caseydavenport (commented):
Whoops, never submitted this. It's only one nitty comment though :D

PR #923, opened 2020-09-30
migration fixes for Azure

ozdanborne:
## Description

- ignore cluster-autoscaler daemonset annotation
- degrade install-cni container error to warning
- fix bug where azv prefix is incorrectly set to avz

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


tmjd (approved):
I think this is reasonable.
I'm a little concerned with ignoring the install CNI container but I think it should be ok in the case of other CNI plugins.

PR #930, opened 2020-10-06
Default CNI.Type to AmazonVPC if aws-node daemonset exists

tmjd:
## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


ozdanborne (approved):
LGTM, but I am curious how this all works when using calico CNI in EKS, which I think was recently made possible?

tmjd:
We chatted offline but just to have it here also.
Currently in EKS the `cni.type` is defaulted to AmazonVPC so this change would not have any effect. With this change though maybe we should consider removing the defaulting based on EKS provider and only depend on the aws-node daemonset existing for defaulting to AmazonVPC.

PR #931, opened 2020-10-07
Use -felix-ready for calico-node readiness on OpenShift

lmm:
## Description

We previously had to use `-bird-ready` for the calico-node readiness probe command because Felix did not support configurable health ports at the time. This PR fixes the readiness probe (since calico-node supports[ configurable Felix health ports](https://github.com/projectcalico/node/pull/241) already).

This also fixes OCP with VXLAN and no BGP clusters not able to come up healthy since that would use `-bird-ready` in the readiness command.

I'm not sure of the milestone but this would need to go into the operator slated for Calico v3.17.0

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/node_test.go
+++ b/pkg/render/node_test.go
@@ -2387,21 +2387,11 @@ func verifyProbes(ds *apps.DaemonSet, isOpenshift, isEnterprise bool) {
 	Expect(found).To(BeTrue())
 
 	switch {
-	case !isOpenshift && !isEnterprise && !bgp:
+	case !bgp:
 		expectedReadiness.Exec.Command = []string{"/bin/calico-node", "-felix-ready"}
-	case !isOpenshift && !isEnterprise && bgp:
+	case bgp && !isEnterprise:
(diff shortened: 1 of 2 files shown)


pkg/render/node_test.go
```diff
@@ -2383,16 +2369,42 @@ func verifyProbes(ds *apps.DaemonSet, isOpenshift, isEnterprise bool) {
...
+				bgp = true
+			}
+			found = true
+			break
+		}
+	}
+	Expect(found).To(BeTrue())
+
+	switch {
+	case !isOpenshift && !isEnterprise && !bgp:
```
tmjd, line 2390: I think isOpenshift can be omitted from these switch cases since it now no longer impacts the Readiness command.

tmjd (commented):
One cleanup I think needed on the tests.
Looking at when the change went in that this PR references, I think we could have had these changes the whole time, we probably were just copying what was in the manifests.

tmjd (approved):
LGTM
Please squash before merging.
I updated the labels, please see if you agree with them.

PR #935, opened 2020-10-08
check for addon manager

ozdanborne:
## Description

Addon manager will only reconcile resources with the `addonmanager.kubernetes.io/mode` label on them. As such, we can block upgrades if we detect that label on any components.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/migration/convert/addon_manager.go
+++ b/pkg/controller/migration/convert/addon_manager.go
@@ -1,0 +1,32 @@
+package convert
+
+import operatorv1 "github.com/tigera/operator/api/v1"
+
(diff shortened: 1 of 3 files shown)


tmjd (approved):
I'm wondering if there is a mode that we could work with but I think we should go with this for now  and if we find a mode that we could work with in the future we can address it then.

ozdanborne:
There's two modes: Reconcile, and EnsureExists. I think in the future we can open up support to EnsureExists if we don't delete the resources, but rather just scale them down to zero instances.

Until then though, yeah lets go with this.

PR #938, opened 2020-10-09
Dont wipe status at start

tmjd:
## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/status/status.go
+++ b/pkg/controller/status/status.go
@@ -78,7 +78,7 @@ type statusManager struct {
 	statefulsets map[string]types.NamespacedName
 	cronjobs     map[string]types.NamespacedName
 	lock         sync.Mutex
-	enabled      bool
+	enabled      *bool
 
 	// Track degraded state as set by external controllers.
 	degraded               bool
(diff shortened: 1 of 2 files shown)


Suraiya-Hameed:
LGTM

pkg/controller/status/status.go
```diff
@@ -78,7 +78,8 @@ type statusManager struct {
 	statefulsets map[string]types.NamespacedName
 	cronjobs     map[string]types.NamespacedName
 	lock         sync.Mutex
-	enabled      bool
+	crQueried    bool
```
ozdanborne, line 81: I've looked through this a few times but I'm still not seeing the importance of knowing if the CR was queried or not. What is this flag supposed to indicate?

tmjd: It lets us know that we have at least Queried for the CR (because OnCRFound or OnCRNotFound has been called at least once). This prevents us from removing the tigerastatus when we haven't tried querying the CR yet.

tmjd: Previously we would just remove the tigerastatus if enabled was false and it was false at startup. So to make sure we only remove it after we've queried it I added this flag.

pkg/controller/installation/core_controller.go
```diff
@@ -538,7 +538,8 @@ func mergeProvider(cr *operator.Installation, provider operator.Provider) error
 // Result.Requeue is true, otherwise upon completion it will remove the work from the queue.
 func (r *ReconcileInstallation) Reconcile(request reconcile.Request) (reconcile.Result, error) {
 	reqLogger := log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
-	reqLogger.V(1).Info("Reconciling Installation.operator.tigera.io")
+	reqLogger.Info("Reconciling Installation.operator.tigera.io")
+	defer reqLogger.Info("Finished reconciling network installation")
```
caseydavenport, line 542: Is there a way to tell now when we have reached the end of this function vs. exiting early for some reason?

tmjd: Not currently.
Should I add a flag that we set at the end before returning to be able to tell that difference?
Or just drop this change. The thing is I don't think we always have a message when the Reconcile finishes (early).

caseydavenport: I don't think it's really worth adding a flag.

Either leave it as it is in this PR if you think it's worthwhile, or revert and add logging to the places where it is missing would be my suggestion.

pkg/controller/status/status.go
```diff
@@ -475,8 +494,8 @@ func (m *statusManager) containerErrorMessage(p corev1.Pod, c corev1.ContainerSt
 }
 
 func (m *statusManager) set(conditions ...operator.TigeraStatusCondition) {
-	if !m.enabled {
-		// Never set any conditions unless the status manager is enabled.
+	if m.enabled == nil || !*m.enabled {
+		// Never set any conditions unless the CR is present.
```
caseydavenport, line 498: I think this should still say "is enabled". 

At the moment, we know we set this bit based on CR, but from the context of the status manager this variable doesn't signify if the CR is present, rather it signifies:

- enabled
- disabled
- waiting-for-information.

pkg/controller/status/status.go
```diff
@@ -160,7 +171,8 @@ func (m *statusManager) OnCRNotFound() {
 	m.clearProgressing()
 	m.lock.Lock()
 	defer m.lock.Unlock()
-	m.enabled = false
+	_f := false
```
caseydavenport, line 174: what's the `_` for? Is that a convention I'm not aware of?

pkg/controller/status/status.go
```diff
@@ -415,11 +427,18 @@ func (m *statusManager) syncState() bool {
 	return m.explicitDegradedReason != ""
 }
 
+// isCRQueried returns true if corresponding CR has been queried
+func (m *statusManager) isCRQueried() bool {
```
caseydavenport, line 431: Similar to my comment below, I think this should be something like "isInitialized()", since the status manager code is meant to be more-or-less agnostic to the logic that is driving it.

caseydavenport (approved):
I added some nits around naming, but this LGTM.

pkg/controller/status/status.go
tmjd: I thought I've seen that before for variables that are only created for the purpose of being able to get a pointer to them. If not I'll drop the `_`.

PR #941, opened 2020-10-14
Fine-grained control of log types for Syslog forwarding and alert forwarding 

stevegaossou:
## Description

- Adding new LogTypes list field for LogCollector SyslogStoreSpec to allow fine-grained control of log types forward to syslog 
- Propagate LogCollector SyslogStoreSpec to ID controller so that we can set env when IDS event log type is set

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


tmjd (changes requested):

pkg/render/fluentd.go
```diff
@@ -427,6 +424,43 @@ func (c *fluentdComponent) envvars() []corev1.EnvVar {
 					},
 				)
 			}
+
+			// When LogTypes is not defined explicitly, the default is to enable
+			// all the log types.
+			if syslog.LogTypes == nil {
```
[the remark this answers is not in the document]
stevegaossou: Yeah, that's my bad. 

I added this not knowing the convention for adding defaults and meant to have a comment here to call that out so that we could talk about how that works.

I'll go ahead and do that.

pkg/controller/intrusiondetection/intrusiondetection_controller.go
```diff
@@ -87,6 +88,12 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 		return fmt.Errorf("intrusiondetection-controller failed to watch installer job: %v", err)
 	}
 
+	// Watch for changes to to primary resource LogCollector, to determine if syslog forwarding is
+	// turned on or off.
+	// Since this syslog forwarding is optional, we don't need to return an error if LogCollector
+	// is not found.
```
stevegaossou, line 94: @tmjd not sure about this part. 

Syslog forwarding will only be set if this is a management or standalone cluster, so I suppose I could try to get a `managementClusterConnection` resource, if I find one then it must be managed cluster and I don't need to return an error (proceed as if nothing bad happened). 

Of course, the other angle to this is even for a management or standalone cluster, syslog forwarding is optional. So if it cannot watch a LogCollector then perhaps it should proceed as if nothing bad happened regardless? 

That's why I have it like this right now.

pkg/render/fluentd.go
```diff
@@ -427,6 +424,43 @@ func (c *fluentdComponent) envvars() []corev1.EnvVar {
...
+					corev1.EnvVar{Name: "SYSLOG_DNS_LOG", Value: "true"},
+					corev1.EnvVar{Name: "SYSLOG_FLOW_LOG", Value: "true"},
+					corev1.EnvVar{Name: "SYSLOG_IDS_EVENT_LOG", Value: "true"},
+				)
+			} else {
+				for _, t := range syslog.LogTypes {
+					switch t {
+					case operatorv1.SyslogLogAuditEE:
+						envs = append(envs,
+							corev1.EnvVar{Name: "SYSLOG_AUDIT_EE_LOG", Value: "true"},
```
asincu, line 443: Could we extract these values as constants ? Seems like they are being used in multiple places.

pkg/render/fluentd.go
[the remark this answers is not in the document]
asincu: I wonder if we can the make the default value of `syslog.LogTypes` to be []syslog.LogTypes{SyslogLogAuditEE,
	SyslogLogAuditKube,
	SyslogLogDNS,
	SyslogLogFlows,
	SyslogLogIDSEvents,}.  What happens if you do not want to export any log to syslog ? How do you disable the feature ?

pkg/controller/intrusiondetection/intrusiondetection_controller.go
```diff
@@ -87,6 +88,12 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
+	_ = c.Watch(&source.Kind{Type: &operatorv1.LogCollector{}}, &handler.EnqueueRequestForObject{})
```
asincu, line 95: Do we need to handle the error returned ?

pkg/render/fluentd.go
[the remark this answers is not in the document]
stevegaossou: > I wonder if we can the make the default value of syslog.LogTypes to be []syslog.LogTypes{SyslogLogAuditEE,
SyslogLogAuditKube,
SyslogLogDNS,
SyslogLogFlows,
SyslogLogIDSEvents,}

That's an interesting suggestion. You can certainly set default values through the kubebuilder markers. I wonder what the behaviour is though if you have a LogTypes field defined but have it set to empty. 

> What happens if you do not want to export any log to syslog ? How do you disable the feature ?

If you don't want to export any logs of any types to syslog then you should not have the SyslogStoreSpec section at all in your LogCollector CR. 

So if you added it and then decide you don't want to export any logs to syslog, you need to remove the SyslogStoreSpec section entirely. That's the clearest way. 

I think it probably also works if you have the SyslogStoreSpec defined and define an empty LogTypes list. That should work too. But I would treat this second way as unsupported (not documented).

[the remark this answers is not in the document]
asincu: I think this it is a bit confusing. I would rather not make this expansion and let the user declare what it needs to forward to syslog.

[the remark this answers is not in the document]
stevegaossou: Spoke with @asincu some more. 

It sounds like the design of this field wasn't quite clean enough and it missed some clarity. 

Here's a summary of where we're at: 
- LogTypes is not optional, since if a user has SyslogStoreSpec defined it means they want to export at least one type of log 
- We were trying to compensate for this (in case the user left the field empty) by having it auto-declare the list with all valid log types
- But actually we should make LogTypes required, and make it default to all the valid log types 

We can default it simply be using the `+kubebuilder:default` marker. 

@tmjd thoughts?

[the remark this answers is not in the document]
stevegaossou: So the definition will now look like this: 
```yaml
syslog:
  description: If specified, enables exporting of flow, audit, and
    DNS logs to syslog.
  properties:
    endpoint:
      description: 'Location of the syslog server. example: tcp://1.2.3.4:601'
      type: string
    logTypes:
      default:
      - AuditEE
      - AuditKube
      - DNS
      - Flows
      - IDSEvents
      description: LogTypes contains a list of types of logs to
        export to syslog. By default, if this field is omitted,
        it will be set to include all possible values.
      items:
        description: SyslogLogType represents the allowable log
          types for syslog. Allowable values are AuditEE, AuditKube,
          DNS, Flows and IDSEvents. * AuditEE corresponds to audit
          logs for Enterprise API custom resources. * AuditKube
          corresponds to audit logs for Kubernetes API resources.
          * DNS corresponds to DNS logs generated by Calico node.
          * Flows corresponds to flow logs generated by Calico node.
          * IDSEvents corresponds to event logs for the intrusion
          detection system (anomaly detection, suspicious IPs, suspicious
          domains and global alerts).
        enum:
        - AuditEE
        - AuditKube
        - DNS
        - Flows
        - IDSEvents
        type: string
      type: array
    packetSize:
      description: 'PacketSize defines the maximum size of packets
        to send to syslog. In general this is only needed if you
        notice long logs being truncated. Default: 1024'
      format: int32
      type: integer
  required:
  - endpoint
  - logTypes
```

pkg/controller/intrusiondetection/intrusiondetection_controller.go
stevegaossou: See my [comment above](https://github.com/tigera/operator/pull/941/files#r504936180). 

> Of course, the other angle to this is even for a management or standalone cluster, syslog forwarding is optional. So if it cannot watch a LogCollector then perhaps it should proceed as if nothing bad happened regardless?

My main concern was that LogCollector was not critical like the other resources ID controller watches for. So I was thinking returning an error here was too severe. 

But I suppose no matter what kind of cluster this is in, if the user has Intrusion Detection running then they must also have a LogCollector defined. 

... I'll add an error here.

pkg/render/fluentd.go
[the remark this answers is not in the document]
tmjd: I believe setting the Field to required means the user must specify the field when creating the CR. Is that incorrect?
With my understanding, if the user is specifying SyslogSpec then they also have to provide the field. Having it optional allows the user to not specify the field and allows us to default it.

I would say the field is not required because we do not need the user specify information in that field.

If you're confident that the LogCollector.SyslogSpec can be created with no logTypes field and the list will be populated with the list of defaults I'd be ok with this but I do not think that is how it works.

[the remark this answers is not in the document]
stevegaossou: I'll test it out against a cluster and confirm what the actual behaviour is.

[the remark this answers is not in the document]
stevegaossou: Okay verified that with the field marked as `required` validation prevents it from being empty and it's not automatically populated by the defaults. 

```
# logcollectors.operator.tigera.io "tigera-secure" was not valid:
# * : Invalid value: "The edited file failed validation": ValidationError(LogCollector.spec.additionalStores.syslog): missing required field "logTypes" in io.tigera.operator.v1.LogCollector.spec.additionalStores.syslog
``` 

However, if I set the field back to `optional` **and** I have the default listing defined in the schema it gets auto populated. 

So what I inputed into the LogCollector CR: 
```
spec:
  additionalStores:
    syslog:
      endpoint: tcp://tinysyslog.default.svc:5140
      packetSize: 1024
```

What I got once I saved it: 
```
spec:
  additionalStores:
    syslog:
      endpoint: tcp://tinysyslog.default.svc:5140
      logTypes:
      - AuditEE
      - AuditKube
      - DNS
      - Flows
      - IDSEvents
      packetSize: 1024
```

@tmjd I think that matches up with what are original expectation was. And without needing to write an explicit logic to do that ourselves. 

@asincu are you onboard with this behaviour now? I can see how it was confusing when I was explaining it. 

Here's what we have: 
- If user defines SyslogStoreSpec section and doesn't set logTypes, then it gets automatically populated with the default list (as shown above in this comment)
- If the user decides later on that they don't want to export any logs of any type to syslog, they need to remove the entire SyslogStoreSpec section
- If the user wants to filter down to a smaller subset of log types to export to syslog, then they edit logTypes and remove the types that they don't want to export

👍 👎 ?

pkg/render/fluentd.go
stevegaossou: It's only defined in two places now (here and then in the test). 

I'm going to leave it alone because it looks like the convention in most of these is to use the string value, so I want to be consistent with what exists.

pkg/render/fluentd.go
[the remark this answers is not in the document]
tmjd: That sounds reasonable to me.
@caseydavenport is there any reason with this behavior ^ that we need to include the defaulting logic in our code?

I do have a little concern about having the defaulting happening in two different places (I'm mainly thinking it might be confusing when debugging?). For logTypes it happens in the CRD machinery and everything else happening in the controller code. I don't think we could ever move all the defaulting into the CRD either but maybe we can put defaulting like this (where we are just populating a field if the user doesn't specify it) in the CRD and keep "auto-detection" in the controllers.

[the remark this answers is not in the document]
tmjd: I am a bit on the fence if we need to do this defaulting ourselves.
But both Casey and I have concerns with using the CRD defaulting so I think we should have the logCollector controller handle the defaulting. In addition to the other concerns mentioned, I do not think there will be many fields where we will be able to use the CRD defaulting so I don't want just a few fields defaulted there.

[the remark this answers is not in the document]
asincu: To be quite honest, I found the defaulting a bit confusing. I see this more as applying a mathematical function. You provide an empty set or a set with zero elements, you get nothing in return. You provide a collection with only DNS log types, you get only DNS logs. You provide a collection that contains all the log types, you get all the logs in return. I think what we are saying is that if you provide an empty set, you get everything. If you provide a set zero elements you get nothing. 

If the defaulting is being done by the controller, does that mean that the operator needs to make an extra call to update the resource after it was created ?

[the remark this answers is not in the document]
stevegaossou: > But both Casey and I have concerns with using the CRD defaulting so I think we should have the logCollector controller handle the defaulting. In addition to the other concerns mentioned, I do not think there will be many fields where we will be able to use the CRD defaulting so I don't want just a few fields defaulted there.

@caseydavenport @tmjd okay I can buy into the reason that fragmenting our defaulting across two different approaches just adds overhead and lacks consistency. 

I'll change it. 

> If the defaulting is being done by the controller, does that mean that the operator needs to make an extra call to update the resource after it was created ?
@asincu from what Erik said it looks like our general pattern does involve an extra update. First we have to grab the CR and examine it, then if a condition is met we update it (here if the logTypes field is empty or not defined but the Syslog section is defined, then populate logTypes with all values). 

In the case of the autobuilder CRD mechanism, I think it might avoid any extra update (perhaps it piggy backs on the original update the user triggered, by running the validation check and then modifying the CR before writing it.

> To be quite honest, I found the defaulting a bit confusing. I see this more as applying a mathematical function. You provide an empty set or a set with zero elements, you get nothing in return. 
@asincu I know it feels a bit awkward from where you standing (i.e. a user would never be able to set logTypes field to empty list). It doesn't let the user do whatever they want with the list, there's a extra mechanism (with default behaviour) happening on top. 

However, if you look at it from a user experience / product perspective it's basically saying the user doesn't have to set logTypes manually (for user convenience/ease of use) and it will auto-populate to having all the types explicitly in the CR (for backwards compatibility if we change the default behaviour in the future).

In the end, if the user wants to not have any logs sent, they _can_ do that, but it requires they remove the Syslog section entirely.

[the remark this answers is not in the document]
stevegaossou: Hey @tmjd, the PR is ready for another pass now. I'm hoping to get this merged in the next day or so (before the release branches are cut). 

Here's the [latest changes](https://github.com/tigera/operator/pull/941/commits/4259fbabdb4e4141149098a4f082b1680591b18c).

pkg/render/intrusion_detection.go
```diff
@@ -647,6 +647,16 @@ func (c *intrusionDetectionComponent) globalAlertTemplates() []runtime.Object {
...
+
+	psp.Spec.Volumes = append(psp.Spec.Volumes, policyv1beta1.HostPath)
+	psp.Spec.AllowedHostPaths = []policyv1beta1.AllowedHostPath{
+		{
+			PathPrefix: "/var/log/calico",
+			ReadOnly:   false,
+		},
+	}
+
+	psp.Spec.RunAsUser.Rule = policyv1beta1.RunAsUserStrategyRunAsAny
```
stevegaossou, line 659: Had to add this because of our PSP validation now, otherwise ID controller does not have permissions to write event logs for Fluentd to tail.

pkg/controller/logcollector/logcollector_controller.go
```diff
@@ -245,6 +247,67 @@ func (r *ReconcileLogCollector) Reconcile(request reconcile.Request) (reconcile.
 		}
 	}
 
+	if instance.Spec.AdditionalStores != nil {
+		if instance.Spec.AdditionalStores.Syslog != nil {
+			syslog := instance.Spec.AdditionalStores.Syslog
+
+			// Try to grab the ManagementClusterConnection CR (if present) because we need it for some
+			// validation with respect to Syslog.logTypes.
+			managementClusterConnection, mccErr := utils.GetManagementClusterConnection(ctx, r.client)
```
tmjd, line 256: Should check if mccErr is IsNotFound and if anything else then we should degrade on that err.

tmjd: Might even want to explicitly set managmenetClusterConnection = nil in the case of IsNotFound.

pkg/controller/logcollector/logcollector_controller.go
```diff
@@ -245,6 +247,67 @@ func (r *ReconcileLogCollector) Reconcile(request reconcile.Request) (reconcile.
...
+				// when upgrading to a future release (regardless of whether or not we change
+				// the defaulting for logTypes).
+				if err = r.client.Update(ctx, instance); err != nil {
+					r.status.SetDegraded(
+						"Failed to write default log types for Syslog configuration",
+						err.Error(),
+					)
+					return reconcile.Result{}, err
+				}
+			} else {
```
tmjd, line 290: I think the code in this `else` clause should always be executed.

pkg/render/intrusion_detection.go
tmjd: Hmm, does this need something different for Openshift? It looks like the other places we set PSP we only do that if not Openshift.

pkg/controller/logcollector/logcollector_controller.go
```diff
@@ -245,6 +247,67 @@ func (r *ReconcileLogCollector) Reconcile(request reconcile.Request) (reconcile.
...
+				// Set all available log types (except for v1.SyslogLogIDSEvents)
+				instance.Spec.AdditionalStores.Syslog.LogTypes = []v1.SyslogLogType{
+					v1.SyslogLogAuditEE,
+					v1.SyslogLogAuditKube,
+					v1.SyslogLogDNS,
+					v1.SyslogLogFlows,
+				}
+
+				// If this is not a managed cluster (i.e. no ManagementClusterConnection CR), then
+				// include v1.SyslogLogIDSEvents as well.
```
tmjd, line 270: @caseydavenport I can't remember if you were involved in discussing this behavior of changing the defaulting behavior based on ManagementClusterConnection CR existence.
I think I suggested this behavior but am second guessing it because there is a race condition of LogCollector CR being created and the ManagementClusterConnection CR.

Will it be a common occurrence to create the LogCollector CR and then later create the ManagementClusterConnection?

pkg/render/fluentd.go
```diff
@@ -427,6 +424,35 @@ func (c *fluentdComponent) envvars() []corev1.EnvVar {
 					},
 				)
 			}
+
+			// When LogTypes is not defined explicitly, the default is to enable
```
tmjd, line 428: I think this comment should be removed with the defaulting happening in the controller.

pkg/controller/logcollector/logcollector_controller.go
caseydavenport: Hm, I don't think I was involved.

I'm not a fan of this approach. Really, I think it's just not an API design that will fly.

 Like you say its relying on the order of events the user takes rather than the state that they provide, which breaks one of the more fundamental goals we're trying to provide in our API - for it to be fully declarative.

I think we just need to pick a default and stick with it, and then the user or our docs need to change it when it needs to be changed. The API is not intended to be zero-config for the user. Trying to remove the need for the user to ever configure the product while great in some limited capacity, when taken to the extreme is a "bad thing".

caseydavenport: What we _can_ do is have validation on this field (as well as on the ManagementClusterConnection CR) that makes sure this field doesn't include SyslogLogIDSEvents if this is a management cluster. 

I don't know the background... is it _really_ not allowed for this to be set on a managed cluster?

config/crd/bases/operator.tigera.io_logcollectors.yaml
```diff
@@ -109,6 +109,29 @@ spec:
...
+                            DNS, Flows and IDSEvents. * AuditEE corresponds to audit
+                            logs for Enterprise API custom resources. * AuditKube
+                            corresponds to audit logs for Kubernetes API resources.
+                            * DNS corresponds to DNS logs generated by Calico node.
+                            * Flows corresponds to flow logs generated by Calico node.
+                            * IDSEvents corresponds to event logs for the intrusion
+                            detection system (anomaly detection, suspicious IPs, suspicious
+                            domains and global alerts).
+                          enum:
+                          - AuditEE
```
caseydavenport, line 128: Beware of abbreviations in our API please!

Maybe this?

AuditEE -> Audit
AuditKube -> AuditKubernetes

stevegaossou: > AuditEE -> Audit
AuditKube -> AuditKubernetes

Hmm, yeah I guess the `AuditKube` isn't a good one. Will change to `AuditKubernetes`. 

As for `AuditEE`, don't we tend to attach Enterprise to things that apply to CaliEnt? I guess `AuditEnterprise` isn't necessary better? I wish we had a consistent name for things that apply to CaliEnt that could be attached to things related to CaliEnt.

pkg/render/intrusion_detection.go
stevegaossou: > Hmm, does this need something different for Openshift? 

Yes the piece specific to Openshift is adding the privileged flag in SecurityContext section. That's [done here](https://github.com/tigera/operator/pull/941/files#diff-41a87360d53438a77f76d9f2a90345ded1d71cb682d4db44041ee2f16a78a504R394-R400).

 >  looks like the other places we set PSP we only do that if not Openshift.
Hmm, not sure about that. Maybe @mgleung could confirm this.

pkg/controller/logcollector/logcollector_controller.go
stevegaossou: > Hm, I don't think I was involved.

Yes, this wasn't discussed during the design phase. I unfortunately caught this late in the game (last week) and the discussion was mainly in [a Slack conversation](https://tigera.slack.com/archives/CJMB52RPG/p1602786223088700). 

Gist of it is that IDSEvent forwarding happens differently from the other log types because we're pulling the values directly out of the LogStorage (ES) which in an MCM setup is centralized. The other log types are always per cluster, so it behaves the same way regardless of the cluster (standalone/management/managed). 

Having this checking is more for the user's benefit as an extra warning so they know that for a managed cluster setting the IDSEvent option does nothing because it's not supported in a managed cluster. If you look at the logic below this comment thread, we don't add the IDSEvent during defaulting, unless the cluster is not a managed cluster. 



> I think we just need to pick a default and stick with it, and then the user or our docs need to change it when it needs to be changed. The API is not intended to be zero-config for the user. Trying to remove the need for the user to ever configure the product while great in some limited capacity, when taken to the extreme is a "bad thing".

Just to be clear the defaulting behaviour of logTypes isn't directly tied to this extra check. 

> Like you say its relying on the order of events the user takes rather than the state that they provide, which breaks one of the more fundamental goals we're trying to provide in our API - for it to be fully declarative.

My question is then, what is the declarative way to check this? The most certain way to determine whether this cluster happens to be managed is looking for the ManagementClusterConnection CR.

pkg/controller/logcollector/logcollector_controller.go
stevegaossou: Hmm, by having the else we exclude the case where `logTypes` is not set. Why would we need this checking if the logTypes is undefined or empty?  

You cannot loop over `syslog.LogTypes` if it's empty or undefined.

config/crd/bases/operator.tigera.io_logcollectors.yaml
caseydavenport: > don't we tend to attach Enterprise to things that apply to CaliEnt? 

I don't know that we do in user-facing situations. My thinking was really predicated on a couple of things:
- "AuditEnterprise" isn't a well-known pattern or standard. It's not immediately obvious to all users what this means.
- This API object is already scoped within Calico Enterprise. We don't need to reiterate to the user that this configures Calico Enterprise, because if they are looking at this object they are already trying to configure Calico Enterprise.

Maybe, could you explain the difference between AuditKube and Audit?

pkg/render/intrusion_detection.go
mgleung: We don't set up PSPs in OpenShift because in OpenShift we should be using the equivalent Security Context Constraints (SCCs). We do not currently have any preset SCCs set up but instead at the moment point users towards the documentation around the required permissions in order to get those set up properly.

mgleung: We require privileged permissions in OpenShift in order to use host mount volumes because of SELinux restrictions

pkg/controller/logcollector/logcollector_controller.go
caseydavenport: > Just to be clear the defaulting behaviour of logTypes isn't directly tied to this extra check.

Really? I thought that this code was determining what entrires to put in the `logTypes` field when it is not defined, which is what I mean by defaulting :) Maybe I'm misunderstanding something, though? 

I think the correct way to do this is not to try to default the `logTypes` field based on whether or not a ManagementClusterConnection exists. 

Instead, always default to the same thing, and have _validation_ check that the provided configuration is consistent with itself. This validation can / should occur in both the logcollector and logstorage controllers.  

The problem with attempting to default based on the presence of another resource is that it takes something declarative - "here is my desired state" and turns it into something imperative - "here are the steps you need to take to get to my desired state". 

For the user, we don't want different behavior if they execute these commands:

```
kubectl create -f management-cluster-connection.yaml
kubeclt create -f log-storage.yaml
```

compared to if they run these commands:

```
kubeclt create -f log-storage.yaml
kubectl create -f management-cluster-connection.yaml
```

config/crd/bases/operator.tigera.io_logcollectors.yaml
tmjd: From what I remember, I think basically AuditKube is for Kube API server handled resources and AuditEnterprise is for Tigera API server handled resources.

pkg/render/intrusion_detection.go
stevegaossou: > We don't set up PSPs in OpenShift because in OpenShift we should be using the equivalent Security Context Constraints (SCCs).

Okay that's my bad. I will wrap this a check to ensure it's not Openshift. 

> We do not currently have any preset SCCs set up but instead at the moment point users towards the documentation around the required permissions in order to get those set up properly.

IIUC you're saying operator shouldn't auto set SSC with privileged for Openshift, correct? 

Because I can see existing examples where we do auto-configure SSC with privileged for Openshift: 
- [fluentd](https://github.com/tigera/operator/blob/master/pkg/render/fluentd.go#L359)
- [compliance](https://github.com/tigera/operator/blob/master/pkg/render/compliance.go#L493)
- [apiserver](https://github.com/tigera/operator/blob/master/pkg/render/apiserver.go#L662)

config/crd/bases/operator.tigera.io_logcollectors.yaml
caseydavenport: Isn't that an implementation detail though? Do we expect all users to know which APIs are handled by which components in the system? 

Maybe a subtle distinction, but I think we need to classify these things by how a user would see them rather than an engineer. e.g., "APIs provided by Calico Enterprise" and "APIs provided by Kubernetes".

Alternatively, we go more flexible and we let the user specify the API groups / versions / kinds that they want audited.

pkg/controller/logcollector/logcollector_controller.go
tmjd: If the code in the else comes after the above if block then the logTypes should always be set.

One way to think of it is the code in the if block is defaulting and the code in the else is validating.
We should be able to default and then validate the config. If you want to keep the loop protected with a check that logTypes is not nil I would not object to that.

pkg/controller/logcollector/logcollector_controller.go
tmjd: An option we may want to consider:
Since `logTypes` is in a Syslog sub-struct under LogCollector, and not configured by default, maybe we should change it to a required field and not default it at all and only validate the field being set appropriately.
Though I guess that would be a problem on upgrades.  If we are defaulting the field to one of the sets (the one with or without IDSEvents) then we will be 'incorrectly' defaulting the logTypes for Managed or Management clusters.

config/crd/bases/operator.tigera.io_logcollectors.yaml
stevegaossou: > From what I remember, I think basically AuditKube is for Kube API server handled resources and AuditEnterprise is for Tigera API server handled resources.

Yes, that's correct.

> Isn't that an implementation detail though? Do we expect all users to know which APIs are handled by which components in the system?

It was something that came out of discussing with Product what options we wanted to provide the user to control what gets forwarded to Syslog. It was the maximum level of granularity, but I can also see how one can argue that audit logs are audit logs, regardless of whether it's K8s resources or our own custom resources. 

I can double check with Phil on whether there was a specific customer case that drove needing to control the split between two types of audit logs.

pkg/controller/logcollector/logcollector_controller.go
stevegaossou: I am inclined to keep this as simple as possible. 

Making `logTypes` required and just forcing the user to set it makes a lot of sense to me (that's what I had prior to setting default values). 

> Though I guess that would be a problem on upgrades.

What would it be a problem on upgrades? I think it was only a problem when we were considering having default logic, but not explicitly declaring what the default values where in the CR. 

If we don't do any defaulting at all, the user has to explicitly set the field and then whatever they set carries over during upgrade (just like it would if we had set default values explicitly for them). 

Thoughts?

config/crd/bases/operator.tigera.io_logcollectors.yaml
stevegaossou: Just head back from Product. There's no real need to separate the two kinds of audit logs. 

I'll change it so that they are both controlled by a single user-facing option. 

I'll go with `Audit` as the log type option.

pkg/controller/logcollector/logcollector_controller.go
tmjd: > What would it be a problem on upgrades?
Currently if they have log forwarding configured, the user would see a change in behavior. Assuming there is no defaulting then the empty set is either invalid or will cause no logs to be forwarded which I would imagine will not be the behavior before the upgrade.

pkg/render/intrusion_detection.go
stevegaossou: @tmjd I think we can close this thread. 

> Hmm, does this need something different for Openshift? It looks like the other places we set PSP we only do that if not Openshift.

You were correct on this. And it is different. 

This PSP for ID controller is [disabled if it's openshift](https://github.com/tigera/operator/blob/master/pkg/render/intrusion_detection.go#L88-L93).

For openshift we have the [SSC set to privileged here](https://github.com/tigera/operator/pull/941/files#diff-41a87360d53438a77f76d9f2a90345ded1d71cb682d4db44041ee2f16a78a504R394-R400).

pkg/controller/logcollector/logcollector_controller.go
stevegaossou: Okay, I see you mean now Erik. 

I rather like the validation of having `logTypes` be required, because it puts the onus on the user to get their Syslog settings correct on their own. 

What about this setup: 
- Have the new `logTypes` field marked as required
    - This forces any new users that enable Syslog in v3.4 and beyond to set `logTypes` with whatever options they want
- For upgrades from previous versions, we keep a defaulting mechanism
    - The defaulting mechanism would work like what we have here already (set default options when it detects a LogCollector CR where the Syslog.LogTypes is empty/undefined and saves the change) 
   - We preserve the old behavior from before logTypes field **and** avoid this problem of checking whether it is a managed cluster or not, **by making the default behaviour only set the options to Audit, DNS and Flows (which is what it used to do)**
    - This means that the user has to explicitly turn on IDSEvents if they want it, which works well since it's new for v3.4 release 
    - If a user sets IDSEvent and this cluster happens to be in a managed cluster, we will also have a check for this that will communicate this isn't supported using the degraded status (whenever operator happens to detect a ManagementClusterConnection CR present)
        - This should be fine now because there won't be a race condition between defaulting (write to LogCollector) and checking for ManagementClusterConnection CR
    - Moving forward all new users who enable Syslog, will be forced to explicitly specify which log types they want because the field will be required and thus having explicit values will preserve expected behavior through into future releases 

Thoughts?

tmjd: Good summary. Seems reasonable to me.

stevegaossou: Okay, here is [the latest changes](https://github.com/tigera/operator/pull/941/commits/d931446f47ca10adbb9af8f7ff701bf7d4217c86), which follow the summary in [my previous comment](https://github.com/tigera/operator/pull/941/#discussion_r508866008) above.

I did some testing with a scenario where I had a LogCollector that did not have Syslog.LogTypes defined and then I upgraded to the latest operator and of the CRD. I can confirm that it correctly added the defaults (Audit, DNS, Flows). In any other case I am forced to set at least one value for LogTypes.

tmjd (commented):
Looking good. Just a test request and one detail in the Intrusion Detection PSP.

pkg/controller/logcollector/logcollector_controller.go
```diff
@@ -245,6 +247,69 @@ func (r *ReconcileLogCollector) Reconcile(request reconcile.Request) (reconcile.
...
+				// option was not available prior to the logTypes field being introduced). This ensures
+				// existing users continue to get the same expected behavior for Syslog forwarding.
+				instance.Spec.AdditionalStores.Syslog.LogTypes = []v1.SyslogLogType{
+					v1.SyslogLogAudit,
+					v1.SyslogLogDNS,
+					v1.SyslogLogFlows,
+				}
+
+				// Write the default list back to the datastore.
+				if err = r.client.Update(ctx, instance); err != nil {
```
caseydavenport, line 302: Shouldn't this update occur outside of the "if syslog.logtypes is not specified" branch?

in most of our controllers we update the resource with the determined defaults always.

stevegaossou: IIUC, you're asking why have a specific update statement just for the Syslog.Logtypes, and not have a generic update statement at the end of the reconcile? Is my understanding correct? 

I could do that, if in the future there's other defaults that need to get added to LogCollector it doesn't make sense to have multiple update statements. 

Although, one advantage to having a specific update here is that set degraded message can be specific and thus potentially more helpful to the user.

pkg/render/intrusion_detection.go
```diff
@@ -574,6 +647,16 @@ func (c *intrusionDetectionComponent) globalAlertTemplates() []runtime.Object {
 func (c *intrusionDetectionComponent) intrusionDetectionPodSecurityPolicy() *policyv1beta1.PodSecurityPolicy {
 	psp := basePodSecurityPolicy()
 	psp.GetObjectMeta().SetName("intrusion-detection")
+
+	psp.Spec.Volumes = append(psp.Spec.Volumes, policyv1beta1.HostPath)
```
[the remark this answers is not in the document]
stevegaossou: Sure, can add that in. 

One thing related to this change to keep in mind, ID controller container now runs as root user.

pkg/controller/logcollector/logcollector_controller.go
caseydavenport: > IIUC, you're asking why have a specific update statement just for the Syslog.Logtypes, and not have a generic update statement at the end of the reconcile? Is my understanding correct?

Yep, in general the pattern we have established for controllers in the operator is that it:

- Queries the CR
- Populates any defaults
- Performs validation
- Updates the CR in the API
- Renders resources

Really well exemplified in the installation controller: https://github.com/tigera/operator/blob/master/pkg/controller/installation/core_controller.go#L581-L600

We probably should have been doing this in this controller from the get-go.

Basically, the reason for always writing it back after defaults is so that we can ensure consistent behavior for users on upgrade, even if we change what the code itself is doing. If a users has LogTypes=[X,Y,Z] in version A, and we change the defaults in version B, the user can upgrade from A->B without fear of behavior changing unexpectedly, since their desired state is spelled out explicitly in the config. 

I haven't done a comprehensive review of this controller in its current form, so maybe I am way off-base, but if there are any other values that we're defaulting in the "render" stage we should be moving those up to the "populate defaults" stage so we can write them back (not in this PR obviously, but just in general).

Anyway, a long-winded response to basically say "Even though we're only doing it for one field right now, we should be doing it for all of them so lets structure the code with that in mind"

pkg/render/intrusion_detection_test.go
```diff
@@ -27,10 +28,21 @@ var _ = Describe("Intrusion Detection rendering tests", func() {
 	It("should render all resources for a default configuration", func() {
 		esConfigMap := render.NewElasticsearchClusterConfig("clusterTestName", 1, 1, 1)
 
+		// Initialize a default LogCollector instance to use.
+		lc := &operatorv1.LogCollector{}
+		lc.Spec.AdditionalStores = &operatorv1.AdditionalLogStoreSpec{
```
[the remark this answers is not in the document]
stevegaossou: Done.

stevegaossou:
@tmjd @caseydavenport I think I've addressed all the feedback now. 

Could one of you guys take a final look? 🙏

pkg/controller/logcollector/logcollector_controller.go
stevegaossou: Cool, thanks for the explainer with the example. I'll be sure to remember the general pattern you outlined with the example moving forward. 

Makes sense to me. I've moved the update out so it's generic.

pkg/render/intrusion_detection.go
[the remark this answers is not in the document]
stevegaossou: Done.

pkg/controller/logcollector/logcollector_controller.go
```diff
@@ -301,9 +367,23 @@ func (r *ReconcileLogCollector) Reconcile(request reconcile.Request) (reconcile.
 		return reconcile.Result{RequeueAfter: 30 * time.Second}, nil
 	}
 
-	// Everything is available - update the CRD status.
+	// Update the LogCollector instance with any changes that have occurred.
+	if isModified {
```
caseydavenport, line 371: This update should actually come before we render and apply resources, immediately after any defaulting is done.

caseydavenport: If rendering fails, the user should be able to query the CR to see exactly what state was being used to render.

pkg/controller/logcollector/logcollector_controller.go
tmjd: It looks like you've put the "Updates the CR in the API" after the "Renders resources". I think the other controllers do the "Updates the CR in the API" as soon as the CR has been updated, so before even retrieving and validating that the config like Secrets and other CRs are valid.

pkg/controller/logcollector/logcollector_controller.go
stevegaossou: Right. I'll move it above the Fleuntd related rendering.

caseydavenport (commented):
LGTM so long as @tmjd is good with it

tmjd (approved):
LGTM

PR #944, opened 2020-10-15
Use ILM to manage ES Index

Suraiya-Hameed:
## Description

Create ILM policies for all ES indices based on retention period and storage size set in LogStorage resource

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


pkg/controller/logstorage/logstorage_controller_test.go
```diff
@@ -535,3 +540,15 @@ func setUpLogStorageComponents(cli client.Client, ctx context.Context, storageCl
 		}),
 	).ShouldNot(HaveOccurred())
 }
+
+func (*mockESClient) NewElasticsearchClient(client.Client, context.Context) error {
+	return nil
+}
+
+func (*mockESClient) SetElasticsearchIlmPolicies(context.Context, *operatorv1.LogStorage, int64) error {
```
rene-dekker, line 548: I think ILM should be three capital letters. In golang abbreviations are all uppercase when the three letters stand for three separate words. (Like URL for example)

pkg/controller/utils/elasticsearch.go
```diff
@@ -62,3 +131,261 @@ func GetElasticsearchClusterConfig(ctx context.Context, cli client.Client) (*ren
...
+		return err
+	}
+
+	es.client = esClient
+	return nil
+}
+
+// SetElasticsearchIlmPolicies creates ILM policies for each timeseries based index using the retention period and storage size in LogStorage
+func (es *EsClient) SetElasticsearchIlmPolicies(ctx context.Context, ls *operatorv1.LogStorage, totalEsStorage int64) error {
+	for _, v := range IndexDiskMapping {
```
rene-dekker, line 166: I was thinking if you would change this method to first call a method to create a list of policies and then a method to apply the policies, your code would become more readable. This struct IndexDiskMapping could disappear and be a variable inside of the method that created the policies

rene-dekker (changes requested):

pkg/controller/utils/elasticsearch.go
```diff
@@ -62,3 +131,261 @@ func GetElasticsearchClusterConfig(ctx context.Context, cli client.Client) (*ren
 
 	return render.NewElasticsearchClusterConfigFromConfigMap(configMap)
 }
+
+type ElasticClient interface {
+	NewElasticsearchClient(client.Client, context.Context) error
+	SetElasticsearchIlmPolicies(context.Context, *operatorv1.LogStorage, int64) error
```
rene-dekker, line 137: Since the interface is already an ElasticClient, I think putting Elasticsearch in the methods is redundant.

pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -689,3 +697,25 @@ func calculateFlowShards(nodesSpecifications *operatorv1.Nodes, defaultShards in
...
+		err := r.esClient.NewElasticsearchClient(r.client, ctx)
+		if err != nil {
+			log.Error(err, "Failed to create ES client")
+			return err
+		}
+	}
+	return r.esClient.SetElasticsearchIlmPolicies(ctx, ls, getTotalEsDisk(ls))
+}
+
+func getTotalEsDisk(ls *operatorv1.LogStorage) int64 {
```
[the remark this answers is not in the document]
Suraiya-Hameed: Using pointer to the logstorage wont work in this case, esClient is part of ReconcileLogStorage struct (as we dont want to create new connection everytime logstorage reconciles) whereas logstorage[*operatorv1.LogStorage] gets created every-time during reconcile.

pkg/controller/utils/elasticsearch.go
```diff
@@ -62,3 +102,323 @@ func GetElasticsearchClusterConfig(ctx context.Context, cli client.Client) (*ren
...
+			// There is no option to set retention for benchmark and events log in LogStorage, set default values used by curator
+			retention = 91
+		}
+		pd := BuildIlmPolicy(totalEsStorage, pctOfTotalDisk, pct, retention)
+		listPolicy[indexName] = pd
+	}
+	return listPolicy
+}
+
+func BuildIlmPolicy(totalEsStorage int64, totalDiskPercentage float64, percentOfDiskForLogType float64, retention int) PolicyDetail {
```
rene-dekker, line 254: I think ILM should be capitalized in all these funcs.

pkg/controller/utils/elasticsearch.go
```diff
@@ -62,3 +102,323 @@ func GetElasticsearchClusterConfig(ctx context.Context, cli client.Client) (*ren
...
+	return &es, nil
+}
+
+// SetILMPolicies creates ILM policies for each timeseries based index using the retention period and storage size in LogStorage
+func (es *EsClient) SetILMPolicies(ls *operatorv1.LogStorage) error {
+	policyList := es.ListILMPolicies(ls)
+	return es.CreateOrUpdatePolicies(policyList)
+}
+
+func (es *EsClient) ListILMPolicies(ls *operatorv1.LogStorage) map[string]PolicyDetail {
```
rene-dekker, line 132: I think it can be rewritten to
```
func (es *EsClient) ListILMPolicies(ls *operatorv1.LogStorage) map[string]PolicyDetail {
	totalEsStorage := getTotalEsDisk(ls)
	pctOfTotalDisk := 0.7
	pctOfDiskAuditSnapshot := 0.1 / float64(6)
	
	return map[string]PolicyDetail{
		"tigera_secure_ee_flows" : BuildIlmPolicy(totalEsStorage, pctOfTotalDisk, 0.9, int(*ls.Spec.Retention.Flows)),
		"tigera_secure_ee_dns" : BuildIlmPolicy(totalEsStorage, pctOfTotalDisk, 0.05, 8),
		"tigera_secure_ee_bgp" : BuildIlmPolicy(totalEsStorage, pctOfTotalDisk, 0.05, 8),

		"tigera_secure_ee_audit_ee" : BuildIlmPolicy(totalEsStorage, pctOfTotalDisk, pctOfDiskAuditSnapshot, int(*ls.Spec.Retention.AuditReports)),
		"tigera_secure_ee_audit_kube" : BuildIlmPolicy(totalEsStorage, pctOfTotalDisk, pctOfDiskAuditSnapshot, int(*ls.Spec.Retention.AuditReports)),
		"tigera_secure_ee_snapshots" : BuildIlmPolicy(totalEsStorage, pctOfTotalDisk, pctOfDiskAuditSnapshot, int(*ls.Spec.Retention.Snapshots)),
		"tigera_secure_ee_benchmark_results" : BuildIlmPolicy(totalEsStorage, pctOfTotalDisk, pctOfDiskAuditSnapshot, 91),
		"tigera_secure_ee_events" : BuildIlmPolicy(totalEsStorage, pctOfTotalDisk, pctOfDiskAuditSnapshot, 91),
		
	}
}
```
And then policiesForIndicesWithMajorSpace and policiesForIndicesWithMinorSpace can be removed.

pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -540,6 +540,13 @@ func (r *ReconcileLogStorage) Reconcile(request reconcile.Request) (reconcile.Re
 			r.status.SetDegraded("Waiting for curator secrets to become available", "")
 			return reconcile.Result{}, nil
 		}
+
+		// ES should be in ready phase when execution reaches here, apply ILM polices
+		if err = r.esClient.SetILMPolicies(r.client, ctx, ls); err != nil {
+			log.Info("Waiting for ES ILM policies and templates to get created")
```
rene-dekker, line 546: I think log.error could be a better choice, that way you can log the err as well.

pkg/controller/utils/elasticsearch.go
```diff
@@ -62,3 +103,271 @@ func GetElasticsearchClusterConfig(ctx context.Context, cli client.Client) (*ren
...
+	if es.client == nil {
+		user, password, root, err := getClientCredentials(client, ctx)
+		if err != nil {
+			return err
+		}
+		h := &http.Client{
+			Transport: &http.Transport{TLSClientConfig: &tls.Config{RootCAs: root}},
+		}
+
+		es.lock.Lock()
```
rene-dekker, line 135: I think the lock should happen as soon as you enter the if statement.

pkg/controller/utils/elasticsearch.go
```diff
@@ -62,3 +103,271 @@ func GetElasticsearchClusterConfig(ctx context.Context, cli client.Client) (*ren
...
+					"actions": map[string]interface{}{
+						"readonly": map[string]interface{}{},
+						"set_priority": map[string]interface{}{
+							"priority": 50,
+						},
+					},
+				},
+				"delete": map[string]interface{}{
+					"min_age": pd.deleteAge,
+					"actions": map[string]interface{}{
```
rene-dekker, line 233: It seems like the whole actions section can be omitted, or does es not allow that?

pkg/controller/utils/elasticsearch.go
```diff
@@ -62,3 +103,271 @@ func GetElasticsearchClusterConfig(ctx context.Context, cli client.Client) (*ren
...
+		"tigera_secure_ee_benchmark_results":  buildILMPolicy(totalEsStorage, minorPctOfTotalDisk, pctOfDisk, 91),
+		"tigera_secure_ee_events":             buildILMPolicy(totalEsStorage, minorPctOfTotalDisk, pctOfDisk, 91),
+	}
+}
+
+func (es *EsClient) createOrUpdatePolicies(ctx context.Context, listPolicy map[string]policyDetail) error {
+	for indexName, pd := range listPolicy {
+		policyName := indexName + "_policy"
+
+		res, err := es.client.XPackIlmGetLifecycle().Policy(policyName).Do(ctx)
```
rene-dekker, line 180: The docs of es library:
```
// XPackIlmPutLifecycle adds or modifies an ilm policy.
```
Does that mean we can just skip the Get Verb and go straight to the Put? That might save some code.

pkg/controller/utils/elasticsearch.go
```diff
@@ -62,3 +103,271 @@ func GetElasticsearchClusterConfig(ctx context.Context, cli client.Client) (*ren
...
+		if err != nil {
+			return err
+		}
+		h := &http.Client{
+			Transport: &http.Transport{TLSClientConfig: &tls.Config{RootCAs: root}},
+		}
+
+		es.lock.Lock()
+		defer es.lock.Unlock()
+		esClient, err := newElastic(user, password, render.ElasticsearchHTTPSEndpoint, h)
```
rene-dekker, line 137: I think there is no need to have a separate newElastic func. The contents could just live here, together with the code that creates the http client.

pkg/controller/utils/elasticsearch.go
```diff
@@ -17,13 +17,54 @@ package utils
 
 import (
 	"context"
-
+	"crypto/tls"
```
rene-dekker, line 20: Please organize the imports.

pkg/controller/utils/elasticsearch.go
Suraiya-Hameed: correct, action is explicitly needed for ES.

pkg/controller/utils/elasticsearch.go
```diff
@@ -62,3 +103,271 @@ func GetElasticsearchClusterConfig(ctx context.Context, cli client.Client) (*ren
 
 	return render.NewElasticsearchClusterConfigFromConfigMap(configMap)
 }
+
+type ElasticClient interface {
+	SetILMPolicies(client.Client, context.Context, *operatorv1.LogStorage) error
+}
+
+type EsClient struct {
```
rene-dekker, line 111: I don't think this EsClient should need to be exported.

pkg/controller/utils/elasticsearch.go
```diff
@@ -62,3 +103,271 @@ func GetElasticsearchClusterConfig(ctx context.Context, cli client.Client) (*ren
...
+}
+
+func newElastic(user, password, url string, h *http.Client) (*elastic.Client, error) {
+
+	options := []elastic.ClientOptionFunc{
+		elastic.SetURL(url),
+		elastic.SetHttpClient(h),
+		elastic.SetErrorLog(logrus.StandardLogger()),
+		elastic.SetSniff(false),
+		//elastic.SetTraceLog(logrus.StandardLogger()),
```
rene-dekker, line 260: remove?

pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -31,7 +31,7 @@ import (
 	"github.com/tigera/operator/pkg/controller/status"
 	"github.com/tigera/operator/pkg/controller/utils"
 	"github.com/tigera/operator/pkg/render"
-
+
```
rene-dekker, line 34: Save actions can also help you remove trailing spaces

pkg/controller/utils/elasticsearch.go
Suraiya-Hameed: That's needed for test to create a mock client. If you look at the elasticsearch_tests.go, it create a mock http client that intercepts call to elasticsearch and validates what is being sent.

pkg/controller/utils/elasticsearch.go
Suraiya-Hameed: XPackIlmPutLifecycle updates the policy and thereby updating the policy version
- I don't think we want the policy version number to be updated each time LogStorage reconciles, that might cause issues in log run where version hits whatever max value is allowed for that type.
- Also, ES notices that the policy version index is using is different than current policy version, so it has to do extra work https://www.elastic.co/guide/en/elasticsearch/reference/current/update-lifecycle-policy.html

So its better to split them up, we are not going to always call PUT, but yes it does involve making a GET call and comparing the values with LogStorage.

rene-dekker: Cool

rene-dekker (approved):
lgtm

pkg/controller/utils/elasticsearch.go
```diff
@@ -17,13 +17,55 @@ package utils
...
 )
 
+const (
+	ElasticsearchRetentionFactor = 4
+	DefaultMaxIndexSizeGi        = 30
+	ElasticConnRetries           = 10
+	ElasticConnRetryInterval     = "500ms"
+)
+
+type Policy struct {
```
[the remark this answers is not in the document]
Suraiya-Hameed: ES repo is just using map[string]interface{} for policy and it is not doing much of validation or parsing with it, mostly just sending it across to ES.

pkg/controller/utils/elasticsearch.go
```diff
@@ -62,3 +104,266 @@ func GetElasticsearchClusterConfig(ctx context.Context, cli client.Client) (*ren
...
+			return err
+		}
+		h := &http.Client{
+			Transport: &http.Transport{TLSClientConfig: &tls.Config{RootCAs: root}},
+		}
+
+		options := []elastic.ClientOptionFunc{
+			elastic.SetURL(render.ElasticsearchHTTPSEndpoint),
+			elastic.SetHttpClient(h),
+			elastic.SetErrorLog(logrus.StandardLogger()),
```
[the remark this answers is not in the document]
Suraiya-Hameed: Interface that `elastic.SetErrorLog()` expects is different from the one available in `sigs.k8s.io/controller-runtime/pkg/log` package, so had to use logrus package.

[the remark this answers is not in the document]
tmjd: Hmm it only needs to meet the interface
```
type Logger interface {        
    Printf(format string, v ...interface{})
} 
```
I think we should wrap the logger we use with something like
```
type logrWrappedESLogger struct {
  l logr.Logger
}
func (l logrWrappedESLogger) Printf(format string, v ...interface{}) {
  l.l.Error(nil, fmt.Printf(format, v...))
}
```
I've used the Error function in the wrapped call because we're passing it to SetErrorLog.
WDYT?
That was we ensure all our logs are in the same format and have the same metadata.

pkg/controller/utils/elasticsearch.go
```diff
@@ -62,3 +104,265 @@ func GetElasticsearchClusterConfig(ctx context.Context, cli client.Client) (*ren
...
+	if err := es.createElasticClient(client, ctx); err != nil {
+		return err
+	}
+	policyList := es.listILMPolicies(ls)
+	return es.createOrUpdatePolicies(ctx, policyList)
+}
+
+func (es *esClient) createElasticClient(client client.Client, ctx context.Context) error {
+	if es.client == nil {
+		es.lock.Lock()
```
tmjd, line 132: This lock should be before accessing the `es.client`, as it is it would be possible that two calls could happen to createElasticClient and if they both do the if check before the lock then they would both create the client.

pkg/controller/utils/elasticsearch.go
```diff
@@ -62,3 +112,266 @@ func GetElasticsearchClusterConfig(ctx context.Context, cli client.Client) (*ren
...
+	return age
+}
+
+func getClientCredentials(client client.Client, ctx context.Context) (string, string, *x509.CertPool, error) {
+	esSecret := &corev1.Secret{}
+	if err := client.Get(ctx, types.NamespacedName{Name: render.ElasticsearchOperatorUserSecret, Namespace: render.OperatorNamespace()}, esSecret); err != nil {
+		if !errors.IsNotFound(err) {
+			return "", "", nil, err
+		}
+		log.Info("Elasticsearch public cert secret not found yet")
```
tmjd, line 325: Why do we continue here? Don't we have to have this secret?
Also isn't this log message incorrect? It should be `tigera-operator/<ElasticsearchOperatorUserSecret>`

tmjd (approved):

PR #946, opened 2020-10-16
Merge converted and configured Installation CR

tmjd:
## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/installation/merge.go
+++ b/pkg/controller/installation/merge.go
@@ -1,0 +1,190 @@
+// Copyright (c) 2020 Tigera, Inc. All rights reserved.
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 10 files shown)


pkg/controller/installation/merge.go
```diff
@@ -0,0 +1,189 @@
...
+	az := reflect.DeepEqual(a, reflect.Zero(reflect.TypeOf(a)).Interface())
+	bz := reflect.DeepEqual(b, reflect.Zero(reflect.TypeOf(b)).Interface())
+	if az && bz {
+		return Same
+	}
+	if reflect.DeepEqual(a, b) {
+		return Same
+	}
+	if az {
+		return BOnlySet
```
ozdanborne, line 119: I might not be reading this code correctly, but it looks like when  az and bz are both set to different values, we incorrectly return `BOnlySet`

tmjd: az and bz will be true if the associated a or b is the nil or zero value, otherwise it will be false.
If they both are the zero value then they're the same and returned after the 1st check.
If only one is set then that should cause the appropriate BOnlySet or AOnlySet to be returned.
If both are set then nither `if az` or `if bz` will hit.

I'm adding a comment to the `az` and `bz` setting to make it clearer what they do.

ozdanborne (approved):

PR #947, opened 2020-10-16
Add Dex to the Operator for adding IDPs to Calico Enterprise

rene-dekker:
Add dex to the operator for adding IDPs.


pkg/apis/operator/v1/authentication_types.go
```diff
@@ -38,34 +43,37 @@ type AuthenticationStatus struct {
 type AuthenticationOIDC struct {
 	// IssuerURL is the URL to the OIDC provider.
 	// +required
-	IssuerURL string `json:"issuerURL"`
+	IssuerURL string `json:"issuerURL,omitempty"`
```
Brian-McM, line 46: why did you add the omitempty if it's a required field?

pkg/controller/utils/utils.go
```diff
@@ -274,62 +271,32 @@ func GetAmazonCloudIntegration(ctx context.Context, client client.Client) (*oper
...
 
-// getAuthentication
-func GetAuthentication(ctx context.Context, client client.Client) (interface{}, error) {
-	var authenticationConfig interface{}
-
-	authenticationCR := &operatorv1.Authentication{}
-	if err := client.Get(ctx, DefaultTSEEInstanceKey, authenticationCR); err != nil {
-		if !kerrors.IsNotFound(err) {
-			return nil, err
+// GetAuthentication finds the authentication CR in your cluster. If you do not want to return an error, but (nil,nil)
```
Brian-McM, line 274: Why have you change this, and removed the type checking through the project? I did it this way so that this would be extendable to more than just OIDC. There could be custom configuration that needs to be calculated that's not in the CR, which is why I created a new config object to hold that for OIDC i.e. like the stuff you look up from the jwksurl. There could be more stuff like this for other authentication types, so the idea is that you just return the configuration struct for the specific type, pass that single configuration into the render method (instead of requiring the function definition to hold all types), and then just switch on the type for the parts you need to configure.

Brian-McM (changes requested):
The main question I have for this first pass is why have you removed the type switching for the authentication? There's more data than what's in the CR that needs to be passed to render, and it stops us from having to add a parameter for every authentication method we're going to use.

pkg/controller/utils/utils.go
rene-dekker: All the kibana settings other than the clientSecret are now static.

pkg/controller/manager/manager_controller.go
```diff
@@ -367,18 +362,45 @@ func (r *ReconcileManager) Reconcile(request reconcile.Request) (reconcile.Resul
...
+		r.status.SetDegraded("Error while fetching Authentication", err.Error())
+		return reconcile.Result{}, err
+	}
+
+	var dexCfg render.DexConfig
+	if authentication != nil {
+		dexTLSSecret := &corev1.Secret{}
+		if err := r.client.Get(ctx, types.NamespacedName{Name: render.DexTLSSecretName, Namespace: render.OperatorNamespace()}, dexTLSSecret); err != nil {
+			if errors.IsNotFound(err) {
+				return reconcile.Result{RequeueAfter: 10 * time.Second}, nil
```
Brian-McM, line 383: Don't think you need the RequeueAfter

pkg/oidc/oidc.go
```diff
@@ -17,7 +18,14 @@ type WellKnownConfig struct {
 }
 
 func LookupWellKnownConfig(issuerURL string) (*WellKnownConfig, error) {
```
Brian-McM, line 20: Is this needed anymore?

pkg/render/compliance.go
```diff
@@ -653,9 +661,10 @@ func (c *complianceComponent) complianceServerDeployment() *appsv1.Deployment {
 			ImagePullSecrets: getImagePullSecretReferenceList(c.pullSecrets),
 			Containers: []corev1.Container{
 				ElasticsearchContainerDecorate(corev1.Container{
-					Name:  ComplianceServerName,
-					Image: components.GetReference(components.ComponentComplianceServer, c.installation.Spec.Registry, c.installation.Spec.ImagePath),
-					Env:   envVars,
+					Name:            ComplianceServerName,
+					Image:           "gcr.io/tigera-dev/cnx/tigera/compliance-server:rene", //todo: reset components.GetReference(components.ComponentComplianceServer, c.installation.Spec.Registry, c.installation.Spec.ImagePath)
```
Brian-McM, line 665: Were these forgotten?

pkg/render/dex_utils.go
```diff
@@ -0,0 +1,288 @@
+// Copyright (c) 2019 Tigera, Inc. All rights reserved.
```
Brian-McM, line 1: Could we please have this named something other than dex_utils please. You could just name the dex_config.

pkg/render/dex_utils.go
```diff
@@ -0,0 +1,288 @@
...
+
+	corev1 "k8s.io/api/core/v1"
+
+	oprv1 "github.com/tigera/operator/api/v1"
+)
+
+type ConnectorType string
+
+const (
+	ConnectorTypeOIDC      = "oidc"
```
Brian-McM, line 30: Are these variables specific to Dex? If yes, could you please prefix them with Dex or something like that.

Also, if these aren't used outside the package please leave them unexported.

pkg/render/dex_utils.go
```diff
@@ -0,0 +1,288 @@
...
+	// ClientSecret returns the secret for Dex' auth endpoint
+	ClientSecret() []byte
+	TLSSecret() *corev1.Secret
+	IdpSecret() *corev1.Secret
+	DexSecret() *corev1.Secret
+	GoogleServiceAccountSecret() []byte
+	OpenshiftRootCA() []byte
+
+	// AppendDexEnv adds env that is used to configure pods with dex options.
+	AppendDexEnv(env []corev1.EnvVar, prefix string) []corev1.EnvVar
```
Brian-McM, line 55: Could you get rid of the "Append" functions and just return the env list please? This appending logic doesn't really help anything.

Brian-McM (changes requested):

api/v1/authentication_types.go
```diff
@@ -60,28 +65,31 @@ type AuthenticationOIDC struct {
...
+	// Deprecated. Please use Authentication.Spec.GroupsPrefix instead.
 	// +optional
 	GroupsPrefix string `json:"groupsPrefix,omitempty"`
 }
 
+// AuthenticationOIDC is the configuration needed to setup Openshift.
+type AuthenticationOpenshift struct {
+	// IssuerURL is the URL to the Opensift OAuth provider. Ex.: https://api.my-ocp-domain.com:6443
+	// +required
+	IssuerURL string `json:"issuerURL"`
```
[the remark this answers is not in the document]
rene-dekker: No, it is not. An early version of the design did attempt to do that, but we decide to stick with the jargon that is specific to the connector type.

Brian-McM (commented):
Forgot to submit these comment earlier, sorry :(

pkg/render/logstorage.go
```diff
@@ -430,6 +420,12 @@ func (es elasticsearchComponent) podTemplate() corev1.PodTemplateSpec {
 	// https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-managing-compute-resources.html and
 	// https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-jvm-heap-size.html#k8s-jvm-heap-size
 
+	var volumeMounts []corev1.VolumeMount
+	if es.dexCfg != nil {
+		volumeMounts = []corev1.VolumeMount{
```
Brian-McM, line 425: Could you please add a RequiredVolumeMounts() function to the dexCfg that returns this instead.

pkg/render/logstorage.go
```diff
@@ -499,15 +495,32 @@ func (es elasticsearchComponent) podTemplate() corev1.PodTemplateSpec {
...
+		ObjectMeta: metav1.ObjectMeta{
+			Annotations: annotations,
+		},
 		Spec: corev1.PodSpec{
 			InitContainers:     initContainers,
 			Containers:         []corev1.Container{esContainer},
 			ImagePullSecrets:   getImagePullSecretReferenceList(es.pullSecrets),
 			NodeSelector:       es.logStorage.Spec.DataNodeSelector,
 			ServiceAccountName: "tigera-elasticsearch",
+			Volumes: []corev1.Volume{
```
Brian-McM, line 511: Could you  please add a RequiredVolumes() method to the dexConfig to get the volumes so they don't need to be created here.

pkg/render/logstorage.go
```diff
@@ -734,24 +745,22 @@ func (es elasticsearchComponent) nodeSetTemplate(pvcTemplate corev1.PersistentVo
...
-			"op.authorization_endpoint":   oidcAuth.AuthorizationEndpoint,
-			"op.token_endpoint":           oidcAuth.TokenEndpoint,
-			"op.jwkset_path":              oidcAuth.JWKSetURI,
-			"op.userinfo_endpoint":        oidcAuth.UserInfoEndpoint,
-			"rp.post_logout_redirect_uri": fmt.Sprintf("https://%s:9443/tigera-kibana/logged_out", oidcAuth.SiteURL),
-			"claims.principal":            oidcAuth.UsernameClaim,
-			"claims.groups":               oidcAuth.GroupsClaim,
+			"rp.redirect_uri":             fmt.Sprintf("%s/tigera-kibana/api/security/oidc/callback", es.dexCfg.ManagerURI()),
+			"rp.requested_scopes":         es.dexCfg.RequestedScopes(),
+			"rp.post_logout_redirect_uri": fmt.Sprintf("%s/tigera-kibana/logged_out", es.dexCfg.ManagerURI()),
```
Brian-McM, line 755: I think that most of these ones that are using fmt.Sprintf or are hard coded should be returned by the dexCfg. I think it might make sense to have functions like TokenEndpoint(), UserInfoEndpoint() and Authorization() endpoints so that anything else that requires this information wouldn't have to know to hard code it or how to fmt.Sprintf it, it just gets it from the config. This also helps if something changes, then you wouldn't need to update a bunch of different places.

pkg/render/dex_config.go
```diff
@@ -0,0 +1,359 @@
...
+		managerURI:     baseUrl,
+	}
+}
+
+type dexBaseCfg struct {
+	authentication *oprv1.Authentication
+	tlsSecret      *corev1.Secret
+	idpSecret      *corev1.Secret
+	dexSecret      *corev1.Secret
+	managerURI     string
```
Brian-McM, line 169: I think you should rename this to base URI and add functions like IssuerEndpoint, TokenEndpoint, and whatnot. We're encapsulating Dex here so the manager shouldn't be mentioned, and the config should be able to provide the exact endpoints that a user of this requires.

Brian-McM (changes requested):

pkg/render/dex_config.go
```diff
@@ -0,0 +1,359 @@
...
+	// ManagerURI returns the address where the Manager UI can be found. Ex: https://example.org
+	RequestedScopes() []string
+	// UsernameClaim returns the part of the JWT that represents a unique username.
+	UsernameClaim() string
+	// GroupsClaim returns the part of the JWT that represents the list of user groups.
+	GroupsClaim() string
+	baseConfig
+}
+
+type baseConfig interface {
```
Brian-McM, line 83: Could you add a more specific name to this please, what exactly this is a base of. This is a package of many different types, not specific to dex so anything else in this package could use this.

I also think that if you're going to make some sort of base config it would be much more useful to have a base that was something like this:

```
type K8sRenderer interface {
   RequiredAnnotations() map[string]string
   RequiredSecrets(namespace string) []*corev1.Secret
   RequiredEnv(prefix string) []corev1.EnvVar
   RequiredVolumes() []corev1.Volume
   RequiredVolumeMounts() []corev1.VolumeMount
}
```

Maybe think of a better name then K8sRenderer but the concept is that anything can implement this. I like that you're using interfacing and compositions of interfaces and structs, but I think the concept of the interfaces and the structure needs to be considered first, over just trying to minimise duplicating code.

For instance, is DexConfig really a composition DexKeyValidatorConfig? Does anything would set up the dex config also be a KeyValidator, or have you combining them because they have similar functions? I think both the DexConfig and the DexKeyValidator should implement the K8sRenderer interface, and if you _really_ wanted to add a common interface for the ManagerURI() then create a separate dex specific interface, and compose multiple interfaces in the DexConfig and DexKeyValidatorConfig.

pkg/render/dex.go
```diff
@@ -0,0 +1,350 @@
...
+			Namespace: DexNamespace,
+		},
+		Data: map[string]string{
+			"config.yaml": string(bytes),
+		},
+	}
+}
+
+// This func prepares the configuration and objects that will be rendered related to the connector and its secrets.
+func getConnector(dexConfig DexConfig) map[string]interface{} {
```
Brian-McM, line 323: Everything in this could just be in the DexConfig. If this is the only thing you need the ConnectorType for then you could just have a GetConnector function on the DexConfig that creates and returns this map, it's all stuff specific to what's in the dex config.

pkg/render/compliance.go
```diff
@@ -713,31 +722,34 @@ func (c *complianceComponent) complianceServerPodSecurityPolicy() *policyv1beta1
...
 	return mounts
 }
 
-func complianceVolumes(defaultMode int32, managerSecret *corev1.Secret) []corev1.Volume {
+func (c *complianceComponent) complianceVolumes() []corev1.Volume {
 	var volumes = []corev1.Volume{{
 		Name: "cert",
 		VolumeSource: corev1.VolumeSource{
 			Secret: &corev1.SecretVolumeSource{
-				DefaultMode: &defaultMode,
```
Brian-McM, line 739: Did you mean to remove this?

pkg/render/compliance.go
```diff
@@ -751,14 +763,13 @@ func complianceVolumes(defaultMode int32, managerSecret *corev1.Secret) []corev1
...
 		}}}
 
-	if managerSecret != nil {
+	if c.managerInternalTLSSecret != nil {
 		volumes = append(volumes,
 			corev1.Volume{
 				Name: ManagerInternalTLSSecretName,
 				VolumeSource: corev1.VolumeSource{
 					Secret: &corev1.SecretVolumeSource{
-						DefaultMode: &defaultMode,
```
Brian-McM, line 760: And this?

pkg/render/dex.go
```diff
@@ -0,0 +1,320 @@
...
+
+const (
+	// Manifest object variables
+	DexNamespace     = "tigera-dex"
+	DexObjectName    = "tigera-dex"
+	DexPort          = 5556
+	DexTLSSecretName = "tigera-dex-tls"
+
+	OIDCSecretName               = "tigera-oidc-credentials"
+	OpenshiftSecretName          = "tigera-openshift-credentials"
```
Brian-McM, line 41: I think some of these should probably be moved to the dex_config.go file, shouldn't they?

pkg/render/dex_config.go
```diff
@@ -0,0 +1,408 @@
...
+	issuer         string
+	connectorType  string
+}
+
+func (d *dexBaseCfg) ManagerURI() string {
+	return d.managerURI
+}
+
+func (d *dexBaseCfg) UsernameClaim() string {
+	claim := "email"
```
Brian-McM, line 170: might be worth a constant name "defaultUsernameClaim".

pkg/render/dex_config.go
```diff
@@ -0,0 +1,408 @@
...
+	if d.connectorType == connectorTypeOIDC && d.authentication.Spec.OIDC.UsernameClaim != "" {
+		claim = d.authentication.Spec.OIDC.UsernameClaim
+	}
+	return claim
+}
+
+func (d *dexBaseCfg) GroupsClaim() string {
+	claim := "groups"
```
Brian-McM, line 178: might be worth a constant name "defaultGroupClaim".

pkg/render/logstorage.go
```diff
@@ -599,7 +604,7 @@ func (es elasticsearchComponent) elasticsearchCluster(secureSettings bool) *esv1
 		},
 		Spec: esv1.ElasticsearchSpec{
 			Version: components.ComponentEckElasticsearch.Version,
-			Image:   components.GetReference(components.ComponentElasticsearch, es.installation.Spec.Registry, es.installation.Spec.ImagePath),
+			// todo: add again Image:   components.GetReference(components.ComponentElasticsearch, es.installation.Spec.Registry, es.installation.Spec.ImagePath),
```
Brian-McM, line 607: Don't forget this please

pkg/render/render.go
```diff
@@ -250,3 +250,14 @@ func appendNotNil(components []Component, c Component) []Component {
 	}
 	return components
 }
+
+type K8sAttributes interface {
```
caseydavenport, line 254: Is this interface used? A quick search doesn't have it show up.

Looks like it has dex specific stuff in it but isn't in a dex specific package?

Brian-McM (commented):
Couple of small nits but looking great!

pkg/render/manager.go
```diff
@@ -609,6 +591,11 @@ func managerClusterRole(managementCluster, managedCluster, openshift bool) *rbac
 				Resources: []string{"serviceaccounts", "namespaces"},
 				Verbs:     []string{"list"},
 			},
+			{
```
caseydavenport, line 594: Could you please add a comment explaining exactly how these permissions are used?

I know we failed to do that for some of the other ones, but it's really important to document what permissions we're providing and why they are required.

pkg/render/render.go
rene-dekker: Hey Casey, the interface is used in dex_config.go and implemented by 3 other types. The idea is to keep this generic and maybe make the constructors of render objects smaller across the project.

pkg/render/compliance.go
rene-dekker: yes, this is the default value that k8s otherwise puts here.

pkg/render/compliance.go
rene-dekker: It is the k8s default and therefore redundant.

pkg/render/logstorage.go
caseydavenport, line 607: This should be commented.

api/v1/authentication_types.go
```diff
@@ -52,36 +57,39 @@ type AuthenticationStatus struct {
...
 
-	// If specified, GroupsPrefix is prepended to each group obtained from the claims specified by GroupsClaim. Note that
-	// Kibana does not support a groups prefix, so this prefix is removed from Kubernetes Groups when translating log access
-	// ClusterRoleBindings into Elastic.
+	// Deprecated. Please use Authentication.Spec.GroupsPrefix instead.
 	// +optional
 	GroupsPrefix string `json:"groupsPrefix,omitempty"`
 }
 
+// AuthenticationOIDC is the configuration needed to setup Openshift.
```
caseydavenport, line 86: Comment doesn't match the struct

api/v1/authentication_types.go
```diff
@@ -52,36 +57,39 @@ type AuthenticationStatus struct {
 
 // AuthenticationOIDC is the configuration needed to setup OIDC.
 type AuthenticationOIDC struct {
-	// IssuerURL is the URL to the OIDC provider.
+	// UpstreamIssuerURL is the URL to the OIDC provider.
```
caseydavenport, line 60: Comment doesn't match the field

api/v1/authentication_types.go
```diff
@@ -52,36 +57,39 @@ type AuthenticationStatus struct {
...
-	// Kibana does not support a groups prefix, so this prefix is removed from Kubernetes Groups when translating log access
-	// ClusterRoleBindings into Elastic.
+	// Deprecated. Please use Authentication.Spec.GroupsPrefix instead.
 	// +optional
 	GroupsPrefix string `json:"groupsPrefix,omitempty"`
 }
 
+// AuthenticationOIDC is the configuration needed to setup Openshift.
+type AuthenticationOpenshift struct {
+	// UpstreamIssuerURL is the URL to the Opensift OAuth provider. Ex.: https://api.my-ocp-domain.com:6443
```
caseydavenport, line 88: Comment doesn't match the field

pkg/render/render.go
Brian-McM: The same items need to be rendered in a few different places, so we came up with this pattern so we could minimise the amount of duplicate around things like volume mounts, copying secrets, setting annotations, and figured this would be a great pattern to use in the future for other components that need these items rendered in multiple places.

api/v1/authentication_types.go
```diff
@@ -22,26 +22,31 @@ import (
...
-	AuthMethodOIDC AuthMethod = "OIDC"
-)
-
 // AuthenticationSpec defines the desired state of Authentication
 type AuthenticationSpec struct {
-	// Method configures the method of authentication used by Kibana.
-	// Default: Basic
-	// +kubebuilder:validation:Enum=Basic;OIDC
-	// +required
-	Method AuthMethod `json:"method,omitempty"`
```
caseydavenport, line 35: I can't remember - did we decide it was OK to remove this field entierly? Any existing users which are setting this field will start getting errors where they previously did not.

api/v1/manager_types.go
```diff
@@ -22,38 +22,12 @@ import (
 
 // ManagerSpec defines configuration for the Calico Enterprise manager GUI.
 type ManagerSpec struct {
-	// Auth defines the authentication strategy for the Calico Enterprise manager GUI.
-	// +optional
-	Auth *Auth `json:"auth,omitempty"`
```
caseydavenport, line 27: Same here - this might break existing users.

We'll need to spell this out really loudly in the release notes.

pkg/controller/apiserver/apiserver_controller.go
```diff
@@ -268,7 +268,7 @@ func (r *ReconcileAPIServer) Reconcile(request reconcile.Request) (reconcile.Res
 	}
 
 	// Everything is available - update the CRD status.
-	instance.Status.State = operatorv1.APIServerStatusReady
+	instance.Status.State = operatorv1.TigeraStatusReady
```
caseydavenport, line 271: I think this is fine, but for context we intentionally had different definitions for the different CR status fields in case we needed to represent different statuses for different CRs.

I suppose we can always split again if the need arises.

Brian-McM: This was my suggestion, I haven't seen a case where we've actually done that and @rene-dekker was following the pattern and adding another "Ready" field. I figured that if something wants a different status it could define it's own and use it, but if it doesn't need it there's a generic one.

pkg/controller/authentication/authentication_controller.go
```diff
@@ -0,0 +1,269 @@
...
+		log.Error(err, fmt.Sprintf("Waiting for network to be %s", oprv1.TigeraSecureEnterprise))
+		r.status.SetDegraded(fmt.Sprintf("Waiting for network to be %s", oprv1.TigeraSecureEnterprise), "")
+		return reconcile.Result{}, nil
+	}
+
+	// Make sure authentication and managementclusterconnection are not present at the same time.
+	_, err = utils.GetManagementClusterConnection(ctx, r.client)
+	if err == nil {
+		log.Error(err, "Having Authentication CR and ManagementCluster CR at the same time is not supported")
+		r.status.SetDegraded("Having Authentication CR and ManagementCluster CR at the same time is not supported", err.Error())
```
caseydavenport, line 154: This is a user facing message, so:
- We shouldn't abbreviate CR
- We should spell out the full API names (ManagementClusterConnection rather than ManagementCluster).

e.g., "Only one of Authentication and ManagementClusterConnection may be specified"

pkg/controller/authentication/authentication_controller.go
```diff
@@ -0,0 +1,269 @@
...
+		return reconcile.Result{}, err
+	} else if !errors.IsNotFound(err) {
+		log.Error(err, "Error querying ManagementCluster")
+		r.status.SetDegraded("Error querying ManagementCluster", err.Error())
```
caseydavenport, line 158: Same here and throughout in user-facing messages (and honestly logs too) - should be ManagementClusterConnection instead of ManagementCluster

api/v1/authentication_types.go
rene-dekker: Looking back it seems we would deprecate the field.

It is mentioned in the comments of https://docs.google.com/document/d/1TNaoURVG-6zM2ReBZ2Ky-p1IhbpAjR8a8V_SoZDuhDU/edit# that this should be deprecated, not removed. I will rectify this.

pkg/controller/authentication/authentication_controller.go
```diff
@@ -0,0 +1,269 @@
...
+	idpSecret = render.CopySecrets(render.DexNamespace, idpSecret)[0]
+
+	dexSecret := &corev1.Secret{}
+	if err := r.client.Get(ctx, types.NamespacedName{Name: render.DexObjectName, Namespace: render.OperatorNamespace()}, dexSecret); err != nil {
+		if errors.IsNotFound(err) {
+			// We need to render a new one.
+			dexSecret = render.CreateDexClientSecret()
+		} else {
+			log.Error(err, "Failed to read dex secret")
+			r.status.SetDegraded("Failed to read dex secret", err.Error())
```
caseydavenport, line 193: "failed to read secret tigera-operator/tigera-dex" will be more useful - spells out exactly the namespace / name of the object that caused the error.

pkg/controller/authentication/authentication_controller.go
```diff
@@ -0,0 +1,269 @@
...
+			return reconcile.Result{}, err
+		}
+	}
+
+	// Extract the connector to be configured for Dex.
```
caseydavenport, line 198: I think this comment is a copy/paste error?

pkg/controller/authentication/authentication_controller.go
```diff
@@ -0,0 +1,269 @@
...
+		if errors.IsNotFound(err) {
+			r.status.OnCRNotFound()
+		}
+		return reconcile.Result{}, err
+	}
+	r.status.OnCRFound()
+	reqLogger.V(2).Info("Loaded config", "config", authentication)
+
+	// Query for the installation object.
+	network, err := installation.GetInstallation(context.Background(), r.client)
```
caseydavenport, line 133: This looks like a copy/paste from some old code from before the Installation object was called "Installation" (it used to be called network!)

I'd s/network/installation here to make the names agree.

pkg/controller/authentication/authentication_controller.go
```diff
@@ -0,0 +1,269 @@
...
+
+	if authentication.Spec.OIDC != nil {
+		secretName = render.OIDCSecretName
+	} else {
+		secretName = render.OpenshiftSecretName
+	}
+
+	secret := &corev1.Secret{}
+	if err := client.Get(ctx, types.NamespacedName{Name: secretName, Namespace: render.OperatorNamespace()}, secret); err != nil {
+		return nil, fmt.Errorf("missing secret %s: %w", secretName, err)
```
caseydavenport, line 254: generally good to include namespace when identifying namespaced objects in logs / errors.

pkg/controller/installation/core_controller.go
```diff
@@ -709,7 +709,7 @@ func (r *ReconcileInstallation) Reconcile(request reconcile.Request) (reconcile.
 		}
 
 		authentication, err = utils.GetAuthentication(ctx, r.client)
-		if err != nil {
+		if err != nil && !apierrors.IsNotFound(err) {
```
caseydavenport, line 712: Hm, interesting. For my own understanding, why was it OK before to return an error when not found and not any more?

pkg/render/render.go
caseydavenport: Gotcha - those files were minimized in my GH - I guess since the PR is so large :) 

I'm feeling like this needs to be scoped within a dex package, maybe along with all of the other dex utilities in dex_config.go?

Alternatively we could rename from K8sAttributes -> DexK8sAttributes or something like that, but it starts to be suspicious when a package needs to start prepending words to scope its structs / functions / etc. Usually a good sign that they deserve their own package.

pkg/render/compliance.go
caseydavenport: Redundant isn't necessarily a bad thing - I'd rather be explicit and verbose rather than implicit and harder to understand. 

I'd suggest, if it isn't required for this functionality, let's keep it out of this PR. Makes it easier to focus on the important changes.

pkg/render/render.go
Brian-McM: The K8sAttributes interface could be used in a variety of different ways, not just with dex. I think there are other components that could benefit from this. Elasticsearch is one that comes to mind, I previously used a decorator like pattern for this but it doesn't work as well as this.

Voltron could be another candidate, as I think similar set up needs to be done in multiple places, i.e. setting annotations, setting common env variables, setting volume mounts and volumes. Regardless, the point is that this really is a generic interface that anything component could take advantage of.

> it starts to be suspicious when a package needs to start prepending words to scope its structs / functions / etc. Usually a good sign that they deserve their own package

Frankly I think that ship has long since sailed with the render package, in my opinion that's quite a few packages mixed into one.

api/v1/manager_types.go
rene-dekker: I added deprecation. Thanks for bringing this up.

pkg/controller/authentication/authentication_controller.go
rene-dekker: I thought we did that because installation collides with the package name.

pkg/controller/authentication/authentication_controller.go
rene-dekker: That was a mistake. Thanks for catching it.

api/v1/authentication_types.go
rene-dekker: I added deprecation. Thanks for bringing this up.

matthewdupre (approved):
We should merge this to fix the bug because of its absence and get some testing.  This doesn't mark the end of making review markups for this PR.

pkg/controller/authentication/authentication_controller.go
Brian-McM: Ya it would be great to not use a name that conflicts, maybe we can just change them all in another PR. You can't use anything from the installation package after this point if network was changed to installation, I wonder if we can enable a linter or something to catch this.

api/v1/authentication_types.go
caseydavenport: @rene-dekker do you mean you added a release note? or something else?

api/v1/authentication_types.go
```diff
@@ -60,28 +65,31 @@ type AuthenticationOIDC struct {
 	// +required
 	UsernameClaim string `json:"usernameClaim"`
 
-	// RequestedScopes is a list of scopes to request from the OIDC provider. If not provided, all the available scopes
-	// are requested.
+	// RequestedScopes is a list of scopes to request from the OIDC provider. If not provided, the following scopes are
+	// requested: ["openid", "email", "profile", "groups", "offline_access"].
```
caseydavenport, line 69: These aren't the user-facing values, right?

We should change them to match what the user would put in:
- OpenID
- Email
- Profile
- Groups
- OfflineAccess

api/v1/manager_types.go
```diff
@@ -29,7 +29,7 @@ type ManagerSpec struct {
 
 // ManagerStatus defines the observed state of the Calico Enterprise manager GUI.
 type ManagerStatus struct {
-	// The last successfully applied authentication configuration.
+	// Deprecated. Please use the Authentication CR for configuring authentication.
```
caseydavenport, line 32: We need some form of status to be here in order to meet openshift certification critera. 

if this one is deprecated we'll need to introduce something else - maybe just a "Ready" field like the others?

pkg/render/render.go
caseydavenport: This isn't a review of the entire render package, this is a review of a new interface whose comments specifically call out "dex" and whose name is _very_ generic. Just because the render packages is broad doesn't mean we should put every line of code into it. Let's acknowledge that it is too large and take steps to fix it!

This is just not a well defined interface at the moment, and it could easily be. Can we either:
- Remove the references to dex and re-brand this?
- More strongly associate this with dex built its own package, renaming it to be more specific?

caseydavenport: > The idea is to keep this generic and make the constructors of render objects smaller across the project.

I'm more than a little worried we're ending up with too many ways to do the same thing in the operator code. As it is, none of the other components take advantage of this and so rather than intorduce some new unifying simplicity, we've added additional complexity. 

if we hadn't already merged this, I would ask that either we revert this change and to it as a separate PR to _all of the components_ so that we achieve consistency, or that we achieve that consistency in this PR itself. 

Are you planning a follow-on PR to make sure this interface is used consistently?

pkg/controller/authentication/authentication_controller.go
```diff
@@ -0,0 +1,268 @@
...
+	ctx := context.Background()
+
+	// Fetch the Authentication spec. If present, we deploy dex in the cluster
+	authentication, err := utils.GetAuthentication(ctx, r.client)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			r.status.OnCRNotFound()
+		}
+		return reconcile.Result{}, err
+	}
```
caseydavenport, line 128: Do we ever perform validation on the queried Authentication CR? I'm not seeing it, but maybe I'm just overlooking something.

pkg/controller/compliance/compliance_controller.go
```diff
@@ -274,13 +280,30 @@ func (r *ReconcileCompliance) Reconcile(request reconcile.Request) (reconcile.Re
 		return reconcile.Result{}, err
 	}
 
+	// Fetch the Authentication spec. If present, we use it to configure dex as an authentication proxy.
+	authentication, err := utils.GetAuthentication(ctx, r.client)
```
caseydavenport, line 284: In downstream controllers that rely on dex being properly configured, I believe we should be blocking them based on the Authentication.Status.Ready field that we set only after Dex is properly configured.

pkg/controller/utils/utils.go
```diff
@@ -274,62 +271,28 @@ func GetAmazonCloudIntegration(ctx context.Context, client client.Client) (*oper
...
-func GetAuthentication(ctx context.Context, client client.Client) (interface{}, error) {
-	var authenticationConfig interface{}
+// GetAuthentication finds the authentication CR in your cluster.
+func GetAuthentication(ctx context.Context, cli client.Client) (*operatorv1.Authentication, error) {
+	authentication := &operatorv1.Authentication{}
+	err := cli.Get(ctx, DefaultTSEEInstanceKey, authentication)
+	if err != nil {
+		return nil, err
+	}
+	if authentication.Spec.OIDC != nil && authentication.Spec.Openshift != nil {
```
caseydavenport, line 281: Ah, I think this is the validation I was looking for. 

Please see this similar comment about the pattern we've established for defaulting / validation in the operator. I think it mostly applies here as well: https://github.com/tigera/operator/pull/941#discussion_r509548336

Basically:

- Query the object
- Fill in defaults for missing fields.
- Validate the object
- Write it back.
- Render components
- Update status

Any other controllers that depend on Dex / the Authentication resource should monitor for the .Status.Ready field, which indicates all of this logic has been completed.

pkg/controller/authentication/authentication_controller.go
caseydavenport: I found it

pkg/controller/authentication/authentication_controller.go
```diff
@@ -0,0 +1,268 @@
...
+			log.Error(err, "Failed to read tigera-operator/tigera-dex-tls secret")
+			r.status.SetDegraded("Failed to read tigera-operator/tigera-dex-tls secret", err.Error())
+			return reconcile.Result{}, err
+		}
+	}
+
+	// Dex will be configured with the contents of this secret, such as clientID and clientSecret.
+	idpSecret, err := getIdpSecret(ctx, r.client, authentication)
+	if err != nil {
+		log.Error(err, "Invalid or missing idp secret")
```
caseydavenport, line 178: These are user facing error messages, so we should be spelling out "identity provider" and also providing the name and namespace of the secret that is missing.

The intention is to make it easy for the user to spot what they have misconfigured or what dependencies they have failed to provide.

pkg/controller/authentication/authentication_controller.go
```diff
@@ -0,0 +1,268 @@
...
+	if err := client.Get(ctx, types.NamespacedName{Name: secretName, Namespace: render.OperatorNamespace()}, secret); err != nil {
+		return nil, fmt.Errorf("missing secret %s/%s: %w", render.OperatorNamespace(), secretName, err)
+	}
+
+	if len(secret.Data[render.ClientIDSecretField]) == 0 {
+		return nil, fmt.Errorf("clientID is a required field for secret %v", secret.Name)
+	}
+
+	if len(secret.Data[render.ClientSecretSecretField]) == 0 {
+		return nil, fmt.Errorf("clientSecret is a required field for secret %v", secret.Name)
```
caseydavenport, line 261: these errors here should have secret namespace as well as secret name to fully qualify them.

api/v1/authentication_types.go
rene-dekker: This was a pre-existing field. To be honest I do not think we need it anymore at all and I have no plans of using it in the instructions for the new create-user-login.md page. I believe we could deprecate it, but when I made the design, I did not yet come to that realization.

That being said, the way I think it works and used to work is that the list provided would be passed into Kibana without modification. I realize that this is a violation of the standards for CRDs, where enum values have their first char capitalized.

The reason why I think it can be deprecated is because the manager uses a fixed set of scopes as well and those are limited to a subset of the current defaults. If this would not suffice, then people would not be able to log into the manager.

cc: @Brian-McM What are your thoughts on this?

api/v1/manager_types.go
rene-dekker: Do you mean to do it like this?
```
// IntrusionDetectionStatus defines the observed state of Tigera intrusion detection capabilities.
type IntrusionDetectionStatus struct {
	// State provides user-readable status.
	State string `json:"state,omitempty"`
}
```

I see that while the ManagerStatus.Auth is deprecated here, the Manager.Spec.Auth is not. I should rectify this.

api/v1/authentication_types.go
caseydavenport: > This was a pre-existing field.

D'oh, right. Missed that! 

> list provided would be passed into Kibana without modification

Dang.

> The reason why I think it can be deprecated is because the manager uses a fixed set of scopes

Hm, does that effectively mean that the current code doesn't actually support changing this?

api/v1/manager_types.go
caseydavenport: > Do you mean to do it like this?

Yeah, all those resources that just have "State" is a way of meeting this requirement.

And yeah, we should have the spec and corresponding state field be the same deprecation status :)

api/v1/authentication_types.go
rene-dekker: This CRD was created for the Kibana OIDC feature in - I believe - v3.2. I think in that design Brian was very thorough in keeping all the options open for having a very flexible spec. While the code still uses this field to override the Kibana settings, it is not used for the manager settings. 

The manager just has a startup shell script that hard-codes [email profile groups]. 

If we think that we will need these requested scopes for the manager, we need to think about how to implement this. This would be a new feature that we did not have before. All that we really want from the identity provider are groups and a username and I believe with the current scopes we should be able to satisfy this.

I suppose we should take this back and discuss how to move forward with this field before making changes.

cc: @matthewdupre

pkg/render/render.go
Brian-McM: I think I should have been more clear with my comment on the package as a whole, and why the whole package would need to be tackled at once, in short it's because we would get cyclic imports. We likely want each component to be a sub package in render, however, there is common code that most every until uses that's in the render package itself. So if we move dex related stuff to it's own sub package it would refer to items in the base render package, but compliance, log storage, and manager would refer back to the dex package. I think this is a problem even if it's not sub packages. 

I had thought about this long ago with log storage and just came to this conclusion, so this isn't just a comment from me saying "oh it's already a problem everywhere else why fix it here", it's just seems like it is an all or nothing thing (at least for the easiest path)

As for the "well defined ness" of the interface, what's Dex specific about it? If you see something there then ya let's change it to be truly generic. 

I'm also wondering what you mean about "too many ways to do the same thing". I thought the only patterns we had was no pattern (pass every resource as it's own parameter) and the Elasticsearch decorator pattern. I'm all for following up and removing the decorator pattern and exchange it for this (I'm happy to take that on myself). I was more thinking there was no pattern like this in the render code so why not add this.

I'd like to admit as well that it was a real error in my judgement to not think about getting your input first before going through with this, and for that I am sorry, adding a new pattern should have been approved by you and or Erik.

caseydavenport: >  what's Dex specific about it?

```
	// RequiredSecrets returns secrets that you need to render for dex.
```

```
	// RequiredVolumes returns volumes that are related to dex.
```

```
	// RequiredEnv returns env that is used to configure pods with dex options.
```
 etc

Brian-McM: Right the comments, I didn't notice them. @rene-dekker could you update those please.

Was there anything dex specific about the interface itself?

caseydavenport: There was nothing dex specific about the lines of code that get compiled, no. But to be clear the comments are the interface - they define what the interface is and does. Without them it could pretty much do anything.

caseydavenport: > I'm also wondering what you mean about "too many ways to do the same thing".

This is probably an overly dramatic way to state what I was thinking. But essentially, if we're trying to introduce a new simplifying paradigm here, we need to roll it out everywhere. Otherwise the next person who comes along is just going to implement their own thing - either because they don't see this option because it's only used in one case (most likely) or because they're going to think "there's no established pattern so I will try to establish one that works for me"

I really appreciate the thought and effort going in to this, and don't have any problems with the actual implementation. But I really feel like we need to do this 100% or not at all, and right now I don't think this is completed.

Brian-McM: Ya I can definitely understand and agree with a) making sure it's visible to the next developer making changes and b) rolling it out completely. Not everything must use this pattern, but I think we should at least get rid of that previous Elasticsearch decorator pattern and replace it with this, it solved the problem of rendering the same secrets but not the problem of having to pass everything individually. 

So if you're OK with this I'm happy to take on the responsibility of removing that pattern as well as look into more specific package management after this release is through, should we agree with what we should do going forward. I'm also definitely OK with making the interface dex specific until then if that's what you want. WDYT?

rene-dekker: This is a great discussion. We have some real pain points in the code and we did an attempt to address it. For now I will rename the interface and move it to the dex_config.go file.

I am sure we now have a better perspective on this issue and an example of how to de-clutter that render function, so let's pick up this discussion elsewhere on a more suitable/interactive medium (video  call or dm) separate from this PR.

Brian-McM: That sounds good, thanks @rene-dekker.

PR #948, opened 2020-10-16
Bump resources for ECK operator

rene-dekker:
Bump ECK memory to prevent OOM.


pkg/render/logstorage.go
```diff
@@ -988,11 +988,11 @@ func (es elasticsearchComponent) eckOperatorStatefulSet() *appsv1.StatefulSet {
 						Resources: corev1.ResourceRequirements{
 							Limits: corev1.ResourceList{
 								"cpu":    resource.MustParse("1"),
-								"memory": resource.MustParse("150Mi"),
+								"memory": resource.MustParse("512Mi"),
```
bcreane, line 991: Is ECK operator written in Visual Basic? Never mind, I know it's golang, but ... wth? You can run the entire Linux kernel in half this space.

bcreane (approved):

PR #958, opened 2020-10-23
[release-v.12] Fix missing entry in enterprise.go

rene-dekker:
DexComponent was missing in the enterprise.go file.


matthewdupre (approved):
LGTM but I think we need an update to make gen-versions as well so it does this for us in future.

tmjd:
Yeah like Matt said, updates should have been made to `hack/gen-versions/enterprise.go.tpl` and `config/enterprise-version.yml`. Then `make gen-versions` should be used to update the file that was updated.

PR #959, opened 2020-10-23
Add dex to enterprise_versions.yml and tpl.

rene-dekker:
Additions for making make gen-files work with the newest dex additions.


tmjd (approved):
Looks good. I assume you did a `make gen-versions` and it produced the same changes you made previously by hand.

rene-dekker:
> Looks good. I assume you did a `make gen-versions` and it produced the same changes you made previously by hand.

I did and it did :)

PR #963, opened 2020-10-27
Changes for auto MTU detection

caseydavenport:
## Description

- spec.CalicoNetwork.MTU == nil means "auto-detect".
- will no longer set FELIX_ env vars unless explicitly set in Installation.
- Adds MTU to the installation status so we can see the value Calico picked.

The operator deployment will need the /var/lib/calico dir mounted in for this change. We could remove that requirement by making the MTU status field "optional". 

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


pkg/controller/installation/core_controller.go
```diff
@@ -879,14 +884,43 @@ func (r *ReconcileInstallation) Reconcile(request reconcile.Request) (reconcile.
 	}
 
 	// Everything is available - update the CRD status.
+	mtu, err := readMTUFile()
+	if err != nil {
+		r.SetDegraded("error reading network MTU", err, reqLogger)
+		return reconcile.Result{}, err
+	}
+	if mtu == 0 {
```
caseydavenport, line 892: We could change this so that instead of blocking the controller, we just don't write the status field.

That would remove dependence on MTU for other controllers to start, and also dependence on the `/var/lib/calico/mtu` file being mounted properly.

pkg/render/node.go
```diff
@@ -928,16 +929,13 @@ func (c *nodeComponent) nodeEnvVars() []v1.EnvVar {
 
 	// Determine MTU to use. If specified explicitly, use that. Otherwise, set defaults based on an overall
 	// MTU of 1460.
-	ipipMtu := "1440"
```
tmjd, line 931: I have some slight hesitation that we're going from some defaulted values to auto-detection, the concern being if some how the auto-detected values are not correct and causes things to break.
Since we defaulted to fairly small MTUs, most likely auto-detected value will be higher if anything. If those higher values are incorrect then once new pods are started could that cause traffic to not function as expected?

My concerns are mostly FUD, I don't have any concerns based on fact.

pkg/controller/installation/core_controller.go
caseydavenport: Yeah, I guess it comes down to whether or not OpenShift will handle the MTU status field changing gracefully or not (I hope it does).

pkg/render/node.go
caseydavenport: Yeah, you're right that this might change MTUs, but that is sorta the intention of this feature :)

pkg/controller/installation/core_controller.go
```diff
@@ -828,47 +868,23 @@ func (r *ReconcileInstallation) Reconcile(request reconcile.Request) (reconcile.
...
-		if instance.Spec.CalicoNetwork != nil && instance.Spec.CalicoNetwork.MTU != nil {
-			// If specified in the spec, then use the value provided by the user.
-			// This is what the rendering code will have populated into the created resources.
-			openshiftConfig.Status.ClusterNetworkMTU = int(*instance.Spec.CalicoNetwork.MTU)
-		} else if instance.Spec.CalicoNetwork != nil {
-			// If not specified, then use the value for Calico VXLAN networking. This is the smallest
-			// value, so might not perform the best but will work everywhere.
-			openshiftConfig.Status.ClusterNetworkMTU = 1410
-		}
+		openshiftConfig.Status.ClusterNetworkMTU = statusMTU
```
tmjd, line 880: 0 is not really a valid MTU, should we not set the status in that case?
I guess it would be zero by default.... should the field be switched to a pointer.

pkg/controller/installation/core_controller.go
```diff
@@ -878,15 +894,40 @@ func (r *ReconcileInstallation) Reconcile(request reconcile.Request) (reconcile.
 		return reconcile.Result{RequeueAfter: 30 * time.Second}, nil
 	}
 
-	// Everything is available - update the CRD status.
+	// Write updated status.
+	instance.Status.MTU = int32(statusMTU)
```
tmjd, line 898: Same question about not setting the MTU here.

tmjd (commented):
Last question, should we not skip setting MTU in the status/Openshift when it is zero (unset)?

pkg/controller/installation/core_controller.go
tmjd: After chatting about this, sounds like we're good. It'll be hidden because of omitempty on the field.

tmjd (approved):
LGTM, once CI is happy

PR #964, opened 2020-10-27
sort bird mounts alphabetically to prevent reconcile loop

ozdanborne:
## Description

A customer recently reported that their calico-node daemonsets were constantly being updated, and were on their ~50 thousandth generation, when using custom bird configs.

Turns out the bird configs were being added as volume mounts to calico-node in a random order (since map iteration in golang is random). Each calico-node daemonset render was different, which causes new reconcile events to happen.

This PR fixes the problem by sorting the bird volumes alphabetically by key before creating them, ensuring their order is consistent.

Custom BGP templates are unusable with operator until this is merged!

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/node.go
+++ b/pkg/render/node.go
@@ -17,6 +17,7 @@ package render
 import (
 	"fmt"
 	"net"
+	"sort"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


caseydavenport (approved):
Wonder if there are any other maps that we're doing this for today.

Also wonder if there is a good test to write?

ozdanborne:
I looked through the render package and the only other place that iterates through maps is the tls annotations for manager. However, they are also stored in a map. And the problem only occurs when we are appending them to lists.

As for a test, I was thinking we could render the same config twice and assert they're exactly equal, but don't have the time to add a toplevel test like that right now.

PR #966, opened 2020-10-27
[master] Dex PR follow-up

rene-dekker:
A few items were called out in the dex PR after it was merged. This PR is an attempt to address the most urgent ones:
- Use core controller pattern of fetching->filling->updating->...-> rather than abusing the GetAuthentication for this
- Fix issue where MCM was blocked by secret
- Deprecate Manager.Auth, but add Status.State
- Log error when fetching installation leads to an error (core controller)
- Pull secrets were not copied over to dex namespace

### DIFF
--- a/pkg/controller/authentication/authentication_controller.go
+++ b/pkg/controller/authentication/authentication_controller.go
@@ -118,16 +118,30 @@ func (r *ReconcileAuthentication) Reconcile(request reconcile.Request) (reconcil
 
 	ctx := context.Background()
 
-	// Fetch the Authentication spec. If present, we deploy dex in the cluster
+	// Fetch the Authentication spec. If present, we deploy dex in the cluster.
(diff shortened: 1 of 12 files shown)


pkg/controller/authentication/authentication_controller.go
```diff
@@ -118,17 +118,31 @@ func (r *ReconcileAuthentication) Reconcile(request reconcile.Request) (reconcil
...
 		return reconcile.Result{}, err
 	}
 	r.status.OnCRFound()
 	reqLogger.V(2).Info("Loaded config", "config", authentication)
 
+	// Set defaults for backwards compatibility.
+	updateAuthenticationWithDefaults(authentication)
+
+	// Validate the configuration
+	err = validateAuthentication(authentication)
```
caseydavenport, line 137: looks like we're missing error handling on this call

caseydavenport (approved):

PR #972, opened 2020-10-29
[master] Create dex namespace and avoid mutual controller dependency

rene-dekker:


pkg/controller/authentication/authentication_controller.go
```diff
@@ -122,10 +124,50 @@ func (r *ReconcileAuthentication) Reconcile(request reconcile.Request) (reconcil
...
 	authentication, err := utils.GetAuthentication(ctx, r.client)
 	if err != nil {
 		if errors.IsNotFound(err) {
+			if installationCR != nil && installationCR.Status.Variant == oprv1.TigeraSecureEnterprise {
+				// Make sure that the dex namespace exists before policies are applied.
+				dexNs := &corev1.Namespace{}
+				if err := r.client.Get(ctx, client.ObjectKey{Name: render.DexObjectName}, dexNs); err != nil {
+					if errors.IsNotFound(err) {
+						// We need to create the namespace and attach it to the installation resource.
+						dexNs = &corev1.Namespace{
```
Brian-McM, line 148: I think it might be better create a namespace component for this, i.e. DexNamespaceComponent and create the handler for the installation object. I know this is likely temporary, but it would good to still keep the regular flow of component rendering / creation the same. This code would turn into something like:
```
installationHndlr := utils.NewComponentHandler(log, r.client, r.scheme, installationCR)
if err := installationHndlr.CreateOrUpdate(context.Background, render.DexNamespace(), r.status()); err != nil {
....
}
```

This handles updating the namespace if necessary as well, so you can just leverage all the thought behind the CreateOrUpdate function that's already there.

pkg/controller/authentication/authentication_controller.go
```diff
@@ -122,10 +124,50 @@ func (r *ReconcileAuthentication) Reconcile(request reconcile.Request) (reconcil
...
+								Annotations: map[string]string{},
+							},
+						}
+						// We add installation as an ownerRef so garbage collection will take place when our product is deleted.
+						if err := controllerutil.SetControllerReference(installationCR, dexNs, r.scheme); err != nil {
+							log.Error(err, "Error creating namespace tigera-dex")
+							return reconcile.Result{}, err
+						}
+						if err := r.client.Create(ctx, dexNs); err != nil {
+							log.Error(err, "Error creating namespace tigera-dex")
```
Brian-McM, line 162: Could you change this to "Failed to create tigera-dex namespace" please.

pkg/controller/authentication/authentication_controller.go
```diff
@@ -134,6 +176,17 @@ func (r *ReconcileAuthentication) Reconcile(request reconcile.Request) (reconcil
 	r.status.OnCRFound()
 	reqLogger.V(2).Info("Loaded config", "config", authentication)
 
+	// Verify the presence and status of the installation resource.
+	if installationCR == nil {
```
Brian-McM, line 180: I think this will be removed if you follow through with my suggestion to return an error if the installationCR is not found

pkg/controller/authentication/authentication_controller.go
```diff
@@ -134,6 +176,17 @@ func (r *ReconcileAuthentication) Reconcile(request reconcile.Request) (reconcil
...
+		log.Error(err, "Installation not found")
+		r.status.SetDegraded("Installation not found", err.Error())
+		return reconcile.Result{}, err
+	}
+	if installationCR.Status.Variant != oprv1.TigeraSecureEnterprise {
```
Brian-McM, line 185: I think you may as well just move this up before any of the authentication CR stuff so it's all together.

Brian-McM (changes requested):

tmjd (commented):
I think maybe here we should try to isolate the specialized code that creates the dex namespace with Installation as a reference if the Auth CR is not present. That should allow us to have the remainder of the code as we would typically expect:
* Auth is the first CR retrieved. The special case for the hack is that if it is not present but Installation is we do something.
* Other CRs that Auth depends on are retrieved after the Auth CR is attempted. This allows us to write the code as we would normally expect for retrieving Installation and reporting any associated errors.
I think one thing that I'm trying to suggest here is that we should have two different paths where we retrieve Installation CR in this code. 1 in the normal path that happens after Auth CR is successfully retrieved and a 2nd in the case that the Auth CR is not found.
I do like the idea to use a ComponentHandler for creating the namespace.

Psuedo code:
```
auth, err := GetAuth()
if err != nil {
  if errors.IsNotFound(err) {
    r.status.OnCRNotFound()
    // Till the end of this if block is all the 'special' code for creating the dex namespace
    // when Auth isn't present. Also this should include a big comment about why this is
    // being done this way. (Because operator doesn't handle policy yet and to keep from
    // needing to split the dex policy into a separate file.)
    i, err := GetInstallation()
    // if error we don't SetDegraded, we only log an error and return
    // if variant Enterprise continue, otherwise return
    h := NewComponentHandler(...)
    r, err := render.DexNS()
    h.CreateOrUpdate(r)
  }
}
i, err := GetInstallation()
// normal err handling, including SetDegraded when appropriate.
```

pkg/render/render_test.go
```diff
@@ -95,13 +95,13 @@ var _ = Describe("Rendering tests", func() {
 		// For this scenario, we expect the basic resources plus the following for Tigera Secure:
 		// - X Same as default config
 		// - 1 Service to expose calico/node metrics.
-		// - 1 ns (tigera-prometheus)
+		// - 2 ns (tigera-prometheus & tigera-dex)
```
tmjd, line 98: This comment isn't right as far as I can tell. I think `tigera-prometheus` is created by a manifest. I think the sum should be `(6 + 4 + 2 + 7 + 5 + 1 + 1) + 1 +1` Note the 2nd `+1` in the parenthesized section (it was added above but no here.

pkg/controller/authentication/authentication_controller.go
```diff
@@ -168,6 +168,19 @@ func (r *ReconcileAuthentication) Reconcile(request reconcile.Request) (reconcil
...
 	}
 
+	// Make sure the tigera-dex namespace exists, before rendering any objects there.
+	if err := r.client.Get(ctx, client.ObjectKey{Name: render.DexObjectName}, &corev1.Namespace{}); err != nil {
+		if errors.IsNotFound(err) {
+			log.Error(err, "Waiting for namespace tigera-dex to be created")
+			r.status.SetDegraded("Waiting for namespace tigera-dex to be created", err.Error())
+			return reconcile.Result{RequeueAfter: 10 * time.Second}, nil
+		} else {
+			log.Error(err, "Error querying tigera-dex namespace")
```
Brian-McM, line 178: Nit, but can you change this to "Failed to retrieve the tigera-dex namespace"?

pkg/controller/installation/core_controller.go
```diff
@@ -718,10 +718,6 @@ func (r *ReconcileInstallation) Reconcile(request reconcile.Request) (reconcile.
 			r.status.SetDegraded("An error occurred retrieving the authentication configuration", err.Error())
 			return reconcile.Result{}, err
 		}
-		if authentication != nil && authentication.Status.State != operator.TigeraStatusReady {
```
Brian-McM, line 721: Was this something that shouldn't have been here in the first place? I'm just wondering how the current changes in for this PR prompted removing this.

Brian-McM (commented):
This looks pretty good to me, I'm leaving the approval of this PR to @tmjd though since he's active in this PR.

pkg/controller/installation/core_controller.go
rene-dekker: That's right. It's in the title of this PR but most of the discussion about it was had in dm.

pkg/controller/authentication/authentication_controller.go
rene-dekker: I'd rather leave it like this. Every other resource that is queried/retrieved this way has the exact same message

tmjd (approved):
LGTM

PR #975, opened 2020-10-30
Add new flag to Authentication spec in order to support OKTA

rene-dekker:
Add ability to skip email verification check by dex.


api/v1/authentication_types.go
```diff
@@ -81,6 +81,12 @@ type AuthenticationOIDC struct {
 	// Deprecated. Please use Authentication.Spec.GroupsPrefix instead.
 	// +optional
 	GroupsPrefix string `json:"groupsPrefix,omitempty"`
+
+	// Some providers do not include the claim "email_verified" when there is no verification in the user enrollment
+	// process or if they are acting as a proxy for another IdP. By default those tokens are deemed invalid.
+	// To skip this check, set the value to true.
+	// +optional
+	InsecureSkipEmailVerified *bool `json:"insecureSkipEmailVerified,omitempty"`
```
tmjd, line 89: I don't think this should be a bool to match K8s API conventions.
Maybe `Email: [Verify|SkipInsecure]`... would that make sense?

tmjd: I feel like this is something that should be defaulted and written back to the datastore also.

pkg/render/dex_config_test.go
```diff
@@ -0,0 +1,46 @@
...
+			ManagerDomain: "https://example.com",
+			OIDC: &operatorv1.AuthenticationOIDC{
+				IssuerURL:                 "https://example.com",
+				UsernameClaim:             "email",
+				GroupsClaim:               "group",
+				RequestedScopes:           []string{"scope"},
+				InsecureSkipEmailVerified: &t,
+			},
+		},
+	}, nil, nil, &corev1.Secret{
```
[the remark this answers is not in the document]
rene-dekker: It is not according to `goimports`, but it does look weird. I added some newlines so my upcoming commit will make it look more normal.

api/v1/authentication_types.go
caseydavenport: Shouldn't this be `Ensure` rather than `Insure`? 

Actually, couldn't it just be `SkipEmailVerified`?

Are there other similar fields we might introduce in the future? should we have a sub-structure of some sort to group these?

caseydavenport: Ignore my previous comment, I am still waking up.

api/v1/authentication_types.go
```diff
@@ -81,6 +88,14 @@ type AuthenticationOIDC struct {
 	// Deprecated. Please use Authentication.Spec.GroupsPrefix instead.
 	// +optional
 	GroupsPrefix string `json:"groupsPrefix,omitempty"`
+
+	// Some providers do not include the claim "email_verified" when there is no verification in the user enrollment
+	// process or if they are acting as a proxy for another IdP. By default those tokens are deemed invalid.
```
caseydavenport, line 93: IdP -> Identity provider

caseydavenport (approved):
Lookin' good

PR #986, opened 2020-11-05
Bugfix for checking SyslogLogIDSEvents in syslog.LogTypes field

stevegaossou:
## Description
- Fixes bug in check for SyslogLogIDSEvents value in syslog.LogTypes field for managed cluster scenario.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/logcollector/logcollector_controller.go
+++ b/pkg/controller/logcollector/logcollector_controller.go
@@ -258,8 +258,8 @@ func (r *ReconcileLogCollector) Reconcile(request reconcile.Request) (reconcile.
 
 			// Try to grab the ManagementClusterConnection CR because we need it for some
 			// validation with respect to Syslog.logTypes.
-			managementClusterConnection, mccErr := utils.GetManagementClusterConnection(ctx, r.client)
-			if mccErr != nil {
+			managementClusterConnection, err := utils.GetManagementClusterConnection(ctx, r.client)
+			if err != nil {
 				// Not finding a ManagementClusterConnection CR is not an error, as only a managed cluster will
 				// have this CR available, but we should communicate any other kind of error that we encounter.
 				if !errors.IsNotFound(err) {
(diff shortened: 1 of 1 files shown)


pkg/controller/logcollector/logcollector_controller.go
```diff
@@ -276,13 +276,13 @@ func (r *ReconcileLogCollector) Reconcile(request reconcile.Request) (reconcile.
...
 			// are only forwarded within a non-managed cluster (where LogStorage is present).
 			if syslog.LogTypes != nil {
-				if mccErr == nil && managementClusterConnection != nil {
+				if err == nil && managementClusterConnection != nil {
 					for _, l := range syslog.LogTypes {
 						// Set status to degraded to warn user and let them fix the issue themselves.
 						if l == v1.SyslogLogIDSEvents {
 							r.status.SetDegraded(
 								"IDSEvents option is not supported for Syslog config in a managed cluster",
-								err.Error(),
```
stevegaossou, line 285: This was the bug.

pkg/controller/logcollector/logcollector_controller.go
```diff
@@ -258,8 +258,8 @@ func (r *ReconcileLogCollector) Reconcile(request reconcile.Request) (reconcile.
 
 			// Try to grab the ManagementClusterConnection CR because we need it for some
 			// validation with respect to Syslog.logTypes.
-			managementClusterConnection, mccErr := utils.GetManagementClusterConnection(ctx, r.client)
```
stevegaossou, line 261: No real reason to name the error mccErr, so switching it back to standard `err` name we always use.

matthewdupre: Either this change is wrong, or the existing code was.  Note that in line 265 we look at err (before), so this is an actual behavioral change.

caseydavenport (approved):

pkg/controller/logcollector/logcollector_controller.go
caseydavenport: I believe that the old code was wrong, based on my reading of it.

matthewdupre: This changes what we return in line 287.  I think that might be OK, but it's not just cosmetic?

stevegaossou: This ultimately did not cause the bug that was caught. However it did not match with the remaining references to the error in this section (which were using `err`). So either way the name had to get reconciled, either change all the references to `mccErr` or to `err`. 

I chose the latter.

marvin-tigera:

marvin-tigera:

PR #995, opened 2020-11-10
Allow overrides of 'default' installation resource through second 'overrides' installation resource

ozdanborne:
## Description

This PR allows overlays of 'default' installation resource through second 'overlay' installation resource.

Changes (separated by commit):
- Updates `render` package to rely on `InstallationSpec` instead of a full `Installation` resource
- Update the LogStorage renderer to use the ManagementClusterconnection as it's OwnerRef component in a managed cluster.
- Updates core controller to check for second Installation resource named 'overlay' and merge it's Spec with the 'default' installation spec. It stores the combined result into the Status field of the 'default' installation, and passes that final spec into the render code.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


pkg/controller/installation/core_controller.go
```diff
@@ -590,6 +590,22 @@ func (r *ReconcileInstallation) Reconcile(request reconcile.Request) (reconcile.
...
+		}
+		reqLogger.V(5).Info("no 'overrides' installation found")
+	} else {
+		reqLogger.V(5).Info("merging overrides")
+		instance.Spec = overrideInstallationSpec(instance.Spec, overrides.Spec)
+	}
+
 	reqLogger.V(2).Info("Loaded config", "config", instance)
 
 	// Validate the configuration.
```
caseydavenport, line 611: I think we should validate pre-override and post so we can clearly indicate if the issue is with the base config or the overridden config.

pkg/controller/installation/core_controller.go
```diff
@@ -241,12 +241,26 @@ type ReconcileInstallation struct {
...
+	// launch pods until the core controller has completed it's transition to enterprise, which is indicated via the variant field
+	// in the status resource.
+	// Note that we don't need to merge data from the spec field because it basically already is for our purposes, since
+	// spec.computed isn't written until Installation has made it to that point.
+	// However we do run a nil check on Computed only because historically, this function never should return nil
+	// TODO: change to return reference instead of pointer
+	if instance.Status.Computed != nil {
+		return instance.Status.Computed, nil
+	}
+	return &operator.InstallationSpec{}, nil
```
ozdanborne, line 263: @tmjd this bit of code still has me unsure. The `Spec` field was always a literal. But because `Status.Computed` is a pointer, we have introduced a new code path where there was no error getting the Installation but it's status is nil.

You'll notice in the original code, there's no way to receive `nil, nil`. So I've replicated that here by returning an empty spec even when there's no status.

tmjd: What would you think of returning an IsNotFound error if Status.Computed is not set?

pkg/controller/installation/core_controller.go
```diff
@@ -600,20 +614,45 @@ func (r *ReconcileInstallation) Reconcile(request reconcile.Request) (reconcile.
...
+			reqLogger.Error(err, "An error occurred when querying the 'overrides' Installation resource")
+			return reconcile.Result{}, err
+		}
+		reqLogger.V(5).Info("no 'overrides' installation found")
+	} else {
+		reqLogger.V(5).Info("merging overrides")
+		instance.Spec = overrideInstallationSpec(instance.Spec, overrides.Spec)
+	}
+
+	reqLogger.V(2).Info("Loaded config", "config", instance)
```
tmjd, line 637: With this message we'll be logging out twice "Loaded config" even in cases without an overlay.

pkg/controller/installation/core_controller.go
```diff
@@ -600,20 +614,45 @@ func (r *ReconcileInstallation) Reconcile(request reconcile.Request) (reconcile.
...
+
+	// Validate the configuration.
+	if err := validateCustomResource(instance); err != nil {
```
tmjd, line 640: I think we should do this validation only if there is an override and we should use a different message so it is obvious when it is with the merged Installation.

pkg/controller/installation/core_controller.go
```diff
@@ -600,20 +614,45 @@ func (r *ReconcileInstallation) Reconcile(request reconcile.Request) (reconcile.
...
 	r.migrationChecked = true
 
 	// A status is needed at this point for operator scorecard tests.
-	// status.variant is written later but for some tests the reconciliation
+	// status.computedSpec.variant is written later but for some tests the reconciliation
 	// does not get to that point.
 	if reflect.DeepEqual(status, operator.InstallationStatus{}) {
-		instance.Status = operator.InstallationStatus{}
+		instance.Status = operator.InstallationStatus{
+			Computed: &operator.InstallationSpec{},
```
tmjd, line 654: This seems unnecessary.

pkg/render/node.go
```diff
@@ -57,29 +55,32 @@ var (
 // Node creates the node daemonset and other resources for the daemonset to operate normally.
 func Node(
 	k8sServiceEp K8sServiceEndpoint,
-	cr *operator.Installation,
+	cr operator.InstallationSpec,
```
tmjd, line 58: I'm not sure about switching these from pointers. Doesn't that mean a shallow copy will be done.

tmjd (commented):
It looks like the changes are a little inconsistent because some places you've switched the render from using a pointer to a value, I think we should keep them as pointers so we're not causing shallow copies.

pkg/render/node.go
```diff
@@ -494,20 +495,18 @@ func (c *nodeComponent) nodeDaemonset(cniCfgMap *v1.ConfigMap) *apps.DaemonSet {
...
-		annotations["prometheus.io/port"] = fmt.Sprintf("%d", *c.cr.Spec.NodeMetricsPort)
+		annotations["prometheus.io/port"] = fmt.Sprintf("%d", *c.cr.NodeMetricsPort)
 	}
 
 	// check tech preview annotation for calico-node apparmor profile
-	a := c.cr.GetObjectMeta().GetAnnotations()
-	if val, ok := a[techPreviewFeatureSeccompApparmor]; ok {
-		annotations["container.apparmor.security.beta.kubernetes.io/calico-node"] = val
-
+	if c.nodeAppArmorProfile != "" {
```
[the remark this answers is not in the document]
ozdanborne: You're right this is a slight behavior change but it is not a valid use case to specify an empty string. 

As per https://kubernetes.io/docs/tutorials/clusters/apparmor/

>The profile_ref can be one of:
>- `runtime/default` to apply the runtime's default profile
>- `localhost/<profile_name>` to apply the profile loaded on the host with the name <profile_name>
>- `unconfined` to indicate that no profiles will be loaded

pkg/controller/installation/core_controller.go
```diff
@@ -600,11 +602,32 @@ func (r *ReconcileInstallation) Reconcile(request reconcile.Request) (reconcile.
...
+	overlay := operator.Installation{}
+	if err := r.client.Get(ctx, utils.OverlayInstanceKey, &overlay); err != nil {
+		if !apierrors.IsNotFound(err) {
+			reqLogger.Error(err, "An error occurred when querying the 'overlay' Installation resource")
+			return reconcile.Result{}, err
+		}
+		reqLogger.V(5).Info("no 'overlay' installation found")
+	} else {
+		instance.Spec = overrideInstallationSpec(instance.Spec, overlay.Spec)
+		reqLogger.V(2).Info("loaded final override config", "config", instance)
```
tmjd, line 622:
```suggestion
		reqLogger.V(2).Info("loaded computed config", "config", instance)
```
I don't think we refer any place else to this as being an override so we should be consistent and use overlay or computed and I think computed is probably best here.

pkg/controller/installation/core_controller.go
```diff
@@ -600,11 +602,32 @@ func (r *ReconcileInstallation) Reconcile(request reconcile.Request) (reconcile.
...
+
+		// Validate the configuration.
+		if err := validateCustomResource(instance); err != nil {
+			r.SetDegraded("Invalid override config", err, reqLogger)
```
tmjd, line 626:
```suggestion
			r.SetDegraded("Invalid computed config", err, reqLogger)
```

pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -499,7 +499,7 @@ func (r *ReconcileLogStorage) Reconcile(request reconcile.Request) (reconcile.Re
 
 	component := render.LogStorage(
 		ls,
-		installationCR,
+		*installationCR.Status.Computed,
```
tmjd, line 502: Another place where we switched to a value rather than pointer.

tmjd (approved):
A few more minor things but if you fix them I don't need to see it again before merging.

PR #997, opened 2020-11-11
Automated cherry pick of #963: Changes for auto MTU detection

caseydavenport:
Cherry pick of #963 on release-v1.13.

#963: Changes for auto MTU detection


pkg/render/node.go
```diff
@@ -337,11 +337,19 @@ func (c *nodeComponent) nodeCNIConfigMap() *v1.ConfigMap {
...
+	var mtu int32 = 0
 	if m := getMTU(c.cr); m != nil {
 		mtu = *m
 	}
 
+	// Calico Enterprise doesn't do auto-MTU detection yet.
+	// If not specified, default it to the smallest value (wireguard) so that
+	// it will work in the most environments.
+	if c.cr.Spec.Variant == operatorv1.TigeraSecureEnterprise && mtu == 0 {
+		mtu = 1400
```
tmjd, line 350: The default for this was 1410 before. Should it have been 1400 and we were just lucky?

pkg/controller/installation/core_controller.go
```diff
@@ -828,47 +868,23 @@ func (r *ReconcileInstallation) Reconcile(request reconcile.Request) (reconcile.
...
-		if instance.Spec.CalicoNetwork != nil && instance.Spec.CalicoNetwork.MTU != nil {
-			// If specified in the spec, then use the value provided by the user.
-			// This is what the rendering code will have populated into the created resources.
-			openshiftConfig.Status.ClusterNetworkMTU = int(*instance.Spec.CalicoNetwork.MTU)
-		} else if instance.Spec.CalicoNetwork != nil {
-			// If not specified, then use the value for Calico VXLAN networking. This is the smallest
-			// value, so might not perform the best but will work everywhere.
-			openshiftConfig.Status.ClusterNetworkMTU = 1410
-		}
+		openshiftConfig.Status.ClusterNetworkMTU = statusMTU
```
tmjd, line 880: I think if enterprise and statusMTU is zero then this needs to be set to 1410 (or should it be 1400).

tmjd (commented):
A couple things I think. One might just be fixing something that was wrong before.

pkg/render/node.go
caseydavenport: Yeah, it should have been 1400 to account for wireguard potentially being enabled.

tmjd (approved):
LGTM

PR #999, opened 2020-11-16
Adjust route table range for GKE CNI plugin

caseydavenport:
## Description

When installed with the GKE CNI plugin, configure Felix so that it
doesn't use the first route table. This will prevent Felix from
attempting to modify / remove routes programmed by the GKE CNI plugin.

## For PR author

- [x] Tests for change.

## For PR reviewers

- [x] Appropriate labels:


fasaxc (approved):
LGTM.  I guess ideally we'd use a felixconfig but, since this is only in GKE it's hard to see it conflict.

caseydavenport:
> I guess ideally we'd use a felixconfig 

Yeah, very much agree. But that needs some more structural work to support and is a larger undertaking than just this var. Hopefully we can do that work soon!

PR #1000, opened 2020-11-18
add linux nodeselector to elastic, kibana, and ids job

ozdanborne:
## Description

The 1000th PR! 🎉 🎆 what an honor!

This PR adds the linux OS nodeselector to elastic, kibana, and ids jobs which were missing them, causing them to sometimes be scheduled on windows nodes in multi-os clusters.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/utils/component.go
+++ b/pkg/controller/utils/component.go
@@ -291,40 +291,36 @@ func ensureOSSchedulingRestrictions(obj runtime.Object, osType render.OSType) {
 		return
 	}
 
-	var podSpec *v1.PodSpec
+	var podSpecs []*v1.PodSpec
 	switch obj.(type) {
 	case *apps.Deployment:
-		podSpec = &obj.(*apps.Deployment).Spec.Template.Spec
+		podSpecs = []*v1.PodSpec{&obj.(*apps.Deployment).Spec.Template.Spec}
 	case *apps.DaemonSet:
-		podSpec = &obj.(*apps.DaemonSet).Spec.Template.Spec
+		podSpecs = []*v1.PodSpec{&obj.(*apps.DaemonSet).Spec.Template.Spec}
 	case *apps.StatefulSet:
-		podSpec = &obj.(*apps.StatefulSet).Spec.Template.Spec
+		podSpecs = []*v1.PodSpec{&obj.(*apps.StatefulSet).Spec.Template.Spec}
 	case *batchv1beta.CronJob:
-		podSpec = &obj.(*batchv1beta.CronJob).Spec.JobTemplate.Spec.Template.Spec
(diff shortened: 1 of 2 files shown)


pkg/controller/utils/component.go
```diff
@@ -301,6 +301,21 @@ func ensureOSSchedulingRestrictions(obj runtime.Object, osType render.OSType) {
...
+	case *kbv1.Kibana:
+		podSpec = &obj.(*kbv1.Kibana).Spec.PodTemplate.Spec
+	case *esv1.Elasticsearch:
+		// elasticsearch resource describes multiple nodeSets which each have a nodeSelector.
+		nodeSets := obj.(*esv1.Elasticsearch).Spec.NodeSets
+		for i, ns := range nodeSets {
+			if ns.PodTemplate.Spec.NodeSelector == nil {
+				nodeSets[i].PodTemplate.Spec.NodeSelector = make(map[string]string)
+			}
+			nodeSets[i].PodTemplate.Spec.NodeSelector["kubernetes.io/os"] = string(osType)
```
ozdanborne, line 315: Because nodeSets are a list, it doesn't fit well into this function which handles a single nodeSelector. I opted not to modify the function to handle multiple nodeSelectors as I thought it might be more confusing than just handling this one exception separately. But I'm open to input!

Brian-McM: I think I'd prefer this changed slightly, as the definition of the function doesn't need to change. This switch just grabs the pod template in a consistent manner than sets the selector below, so if a component can have multiple pod templates (which I hadn't thought of) then I'd prefer it if we changed `var podSpec *v1.PodSpec` to `var podSpecs[] *v1.PodSpec` and then change
```
if podSpec.NodeSelector == nil {
    podSpec.NodeSelector = make(map[string]string)
}

podSpec.NodeSelector["kubernetes.io/os"] = string(osType)
```
to
```
for _, podSpec := range podSpecs {
    if podSpec.NodeSelector == nil {
        podSpec.NodeSelector = make(map[string]string)
    }

    podSpec.NodeSelector["kubernetes.io/os"] = string(osType)
}
```

With this implementation we cover the case that other components could have multiple pod templates and we protect against missing updating a duplicate of the pod template validation / selector assignment logic should we need change this logic.

Brian-McM (changes requested):

pkg/controller/utils/component.go
ozdanborne: I thought wrapping the podSpec sets in `[]*v1.PodSpec{&obj.(*<type>).Spec.Template.Spec}` was going to look uglier than it ended up looking. Implemented as you suggested and it looks good and is working fine :+1:

Brian-McM (approved):

PR #1001, opened 2020-11-19
Add migration check for nodes when typha is deployed

tmjd:
Make sure enough nodes exist for the current typha and new typha

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/migration/convert/typha.go
+++ b/pkg/controller/migration/convert/typha.go
@@ -1,0 +1,42 @@
+package convert
+
+import (
+	"fmt"
+
+	corev1 "k8s.io/api/core/v1"
+
+	operatorv1 "github.com/tigera/operator/api/v1"
+	"github.com/tigera/operator/pkg/controller/utils"
+)
+
+func checkTypha(c *components, _ *operatorv1.Installation) error {
+	if c.typha == nil {
+		return nil
+	}
+	if c.typha.Spec.Replicas == nil {
+		return nil
+	}
+
+	curReplicas := int(*c.typha.Spec.Replicas)
+
+	if curReplicas == 0 {
+		return nil
+	}
+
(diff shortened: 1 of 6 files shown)


pkg/controller/migration/convert/typha.go
```diff
@@ -0,0 +1,43 @@
...
+	"fmt"
+
+	corev1 "k8s.io/api/core/v1"
+
+	operatorv1 "github.com/tigera/operator/api/v1"
+	"github.com/tigera/operator/pkg/controller/utils"
+)
+
+func checkTypha(c *components, _ *operatorv1.Installation) error {
+	if c.typha != nil {
```
ozdanborne, line 13: nit, you can de-indent this if you return
```suggestion
	if c.typha == nil {
	  return nil
	}
```

pkg/controller/migration/convert/typha.go
```diff
@@ -0,0 +1,43 @@
...
+		}
+
+		curReplicas := int(*c.typha.Spec.Replicas)
+
+		if curReplicas == 0 {
+			return nil
+		}
+
+		nodes := corev1.NodeList{}
+		err := c.client.List(ctx, &nodes)
```
ozdanborne, line 25: wild idea: should we avoid this client request by checking the status of the existing calico-node daemonset to see how many instances it's running?

ozdanborne (approved):
couple optional nits but lgtm

pkg/controller/migration/convert/typha.go
tmjd: That is an interesting idea. Though I think there would be issues if operator restarted near when there were only a few of the 'old' calico-node pods left.

PR #1002, opened 2020-11-23
increase resiliency of new overlay feature

ozdanborne:
## Description

The new overlay feature was implemented such that other controllers pull the computed installation spec from the status field. However, when run on systems which do not have the updated CRD definition, this was always nil.

This PR addresses this problem by refactoring `GetInstallation` (the primary function used by other controllers to get current installation state) to compute the final installation spec instead of pulling it from Status.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/logstorage/logstorage_controller.go
+++ b/pkg/controller/logstorage/logstorage_controller.go
@@ -318,7 +318,7 @@ func (r *ReconcileLogStorage) Reconcile(request reconcile.Request) (reconcile.Re
 		r.status.OnCRFound()
 	}
 
-	installationCR, err := installation.GetInstallation(context.Background(), r.client)
+	variant, install, err := installation.GetInstallation(context.Background(), r.client)
 	if err != nil {
 		if errors.IsNotFound(err) {
 			r.status.SetDegraded("Installation not found", err.Error())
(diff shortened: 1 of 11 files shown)


pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -318,7 +318,7 @@ func (r *ReconcileLogStorage) Reconcile(request reconcile.Request) (reconcile.Re
 		r.status.OnCRFound()
 	}
 
-	installationCR, err := installation.GetInstallation(context.Background(), r.client)
+	variant, installationCR, err := installation.GetInstallation(context.Background(), r.client)
```
tmjd, line 321: Same here, `installationCR` isn't a correct name anymore.

pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -466,7 +466,7 @@ func (r *ReconcileLogStorage) Reconcile(request reconcile.Request) (reconcile.Re
 	if ls != nil {
 		hdler = utils.NewComponentHandler(log, r.client, r.scheme, ls)
 	} else {
-		hdler = utils.NewComponentHandler(log, r.client, r.scheme, installationCR)
+		hdler = utils.NewComponentHandler(log, r.client, r.scheme, managementClusterConnection)
```
tmjd, line 469: Are you getting input from Brian or Rene about using this for the owner ref?

pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -501,7 +501,7 @@ func (r *ReconcileLogStorage) Reconcile(request reconcile.Request) (reconcile.Re
 
 	component := render.LogStorage(
 		ls,
-		installationCR.Status.Computed,
+		installationCR,
```
tmjd, line 504: Name again

tmjd (commented):
A few nits on some names.
I wouldn't be upset if you normalized all the names we use for the installation/network/installationCR variable name we use in the different controllers. But my request is to at least update the ones called `installationCR`.

pkg/controller/logstorage/logstorage_controller.go
Brian-McM: This seems like the right thing to do.

tmjd (approved):
LGTM, please squash before merging.

PR #1007, opened 2020-12-01
Move RootCA to alpine trusted cert location

rene-dekker:
This PR comes with a few advantages:
- By moving the location of the cert, it now enables the rootCA not only for tigera-openshift-credentials, but also for OIDC.
- This allows privately hosted OIDC connectors with certs that are not issued by public authorities. We need this for our own e2e, but we know that there are users who will need this.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -33,7 +33,6 @@ require (
 	github.com/r3labs/diff/v2 v2.8.0
 	github.com/robfig/cron v1.2.0 // indirect
 	gopkg.in/yaml.v2 v2.3.0
-	k8s.io/klog/v2 v2.3.0 // indirect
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -33,7 +33,6 @@ require (
 	github.com/r3labs/diff/v2 v2.8.0
 	github.com/robfig/cron v1.2.0 // indirect
 	gopkg.in/yaml.v2 v2.3.0
-	k8s.io/klog/v2 v2.3.0 // indirect
```
rene-dekker, line 36: This is automatically done by running the CI, so I committed it.

Brian-McM (approved):

caseydavenport (approved):

PR #1014, opened 2020-12-08
Propagate the KUBERNETES_SERVICE_XXX variables to kube-controllers.

fasaxc:
## Description

This should allow kube-controllers to function before kube-proxy starts.

## For PR author

- [x] Tests for change.
- [ ] If changing pkg/apis/, run `make gen-files`
- [ ] If changing versions, run `make gen-versions`

## For PR reviewers

A note for code reviewers - all pull requests must have the following:

- [ ] Milestone set according to targeted release.
- [ ] Appropriate labels:
  - `kind/bug` if this is a bugfix.
  - `kind/enhancement` if this is a a new feature.
  - `enterprise` if this PR applies to Calico Enterprise only.

PR #1015, opened 2020-12-08
Propagate the KUBERNETES_SERVICE_XXX variables to API server.

fasaxc:
## Description

Similarly to #1014, propagate the env vars to the API server.

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/apiserver/apiserver_controller.go
+++ b/pkg/controller/apiserver/apiserver_controller.go
@@ -21,12 +21,6 @@ import (
 
 	v1 "k8s.io/api/core/v1"
 
-	operatorv1 "github.com/tigera/operator/api/v1"
-	"github.com/tigera/operator/pkg/controller/installation"
-	"github.com/tigera/operator/pkg/controller/options"
-	"github.com/tigera/operator/pkg/controller/status"
-	"github.com/tigera/operator/pkg/controller/utils"
-	"github.com/tigera/operator/pkg/render"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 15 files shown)


pkg/controller/apiserver/apiserver_controller.go
```diff
@@ -36,28 +30,37 @@ import (
...
+	"github.com/tigera/operator/pkg/controller/utils"
+	"github.com/tigera/operator/pkg/render"
 )
 
 var log = logf.Log.WithName("controller_apiserver")
 
 // Add creates a new APIServer Controller and adds it to the Manager. The Manager will set fields on the Controller
 // and Start it when the Manager is Started.
-func Add(mgr manager.Manager, opts options.AddOptions) error {
+func Add(k8sServiceEp k8sapi.ServiceEndpoint, mgr manager.Manager, opts options.AddOptions) error {
```
tmjd, line 47: I don't think there is any need to pass the k8sServiceEp through here. We didn't in the core_controller.

I'm half tempted to suggest we don't even pass it into the render functions and just have the renders use it directly. But I think I like it as an input the renders, I think...

fasaxc: I like having it as an input to rendering; makes it easier to test.

pkg/controller/apiserver/apiserver_controller.go
```diff
@@ -47,17 +49,18 @@ func Add(mgr manager.Manager, opts options.AddOptions) error {
 		// No need to start this controller.
 		return nil
 	}
-	return add(mgr, newReconciler(mgr, opts.DetectedProvider, opts.AmazonCRDExists))
+	return add(mgr, newReconciler(k8sapi.Endpoint, mgr, opts.DetectedProvider, opts.AmazonCRDExists))
```
tmjd, line 52: Also on core_controller we didn't pass the Endpoint here

pkg/controller/apiserver/apiserver_controller.go
```diff
@@ -47,17 +49,18 @@ func Add(mgr manager.Manager, opts options.AddOptions) error {
...
 }
 
 // newReconciler returns a new reconcile.Reconciler
-func newReconciler(mgr manager.Manager, provider operatorv1.Provider, amazonCRDExists bool) *ReconcileAPIServer {
+func newReconciler(k8sServiceEp k8sapi.ServiceEndpoint, mgr manager.Manager, provider operatorv1.Provider, amazonCRDExists bool) *ReconcileAPIServer {
```
tmjd, line 56: Or here.

pkg/controller/apiserver/apiserver_controller.go
```diff
@@ -47,17 +49,18 @@ func Add(mgr manager.Manager, opts options.AddOptions) error {
...
 	r := &ReconcileAPIServer{
 		client:          mgr.GetClient(),
 		scheme:          mgr.GetScheme(),
 		provider:        provider,
 		amazonCRDExists: amazonCRDExists,
 		status:          status.New(mgr.GetClient(), "apiserver"),
+		k8sServiceEp:    k8sServiceEp,
```
tmjd, line 63: Or store it in the Reconcile object either.

tmjd (commented):
A few more comments similar to the last one. Do you think there is any benefit to pass and storing the Endpoint as part of the Reconcile struct?

tmjd (approved):
LGTM

PR #1018, opened 2020-12-08
Propagate k8s API env var to CNI plugin.

fasaxc:
## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/k8sapi/k8s-endpoint.go
+++ b/pkg/controller/k8sapi/k8s-endpoint.go
@@ -15,7 +15,9 @@
 package k8sapi
 
 import (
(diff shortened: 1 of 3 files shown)


fasaxc:
OK, I've tested this on a kind cluster.  Firstly, it continues to work with this change in place.  I can start busybox pods as normal.  Secondly, if I manually hack the `10-calico.conflist` to set the URL incorrectly then the CNI plugin fails, so that means it must be using the parameter as expected.

tmjd (approved):
LGTM

PR #1026, opened 2020-12-09
Switch Updates after defaulting to use Patch

tmjd:
Use Patch call to update defaults to ensure fields the operator
doesn't know about (like due to new fields during upgrade) are not
deleted by an 'old' already running operator.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/manager/manager_controller.go
+++ b/pkg/controller/manager/manager_controller.go
@@ -154,7 +154,7 @@ type ReconcileManager struct {
 
 // GetManager returns the default manager instance with defaults populated.
 func GetManager(ctx context.Context, cli client.Client) (*operatorv1.Manager, error) {
-	// Fetch the manager instance. We only support a single instance named "default".
+	// Fetch the manager instance. We only support a single instance named "tigera-secure".
 	instance := &operatorv1.Manager{}
 	err := cli.Get(ctx, utils.DefaultTSEEInstanceKey, instance)
 	if err != nil {
@@ -189,12 +189,6 @@ func (r *ReconcileManager) Reconcile(request reconcile.Request) (reconcile.Resul
 	}
 	reqLogger.V(2).Info("Loaded config", "config", instance)
 	r.status.OnCRFound()
-
-	// Write the manager back to the datastore.
-	if err = r.client.Update(ctx, instance); err != nil {
-		r.status.SetDegraded("Failed to write defaults", err.Error())
-		return reconcile.Result{}, err
-	}
 
 	if !utils.IsAPIServerReady(r.client, reqLogger) {
 		r.status.SetDegraded("Waiting for Tigera API server to be ready", "")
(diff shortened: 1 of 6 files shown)


pkg/controller/utils/patch.go
```diff
@@ -0,0 +1,42 @@
...
+		return nil, err
+	}
+
+	patch, err := strategicpatch.CreateTwoWayMergePatch(ogjson, newjson, og)
+	if err != nil {
+		return nil, err
+	}
+
+	return client.RawPatch(types.MergePatchType, patch), nil
+}
```
ozdanborne, line 42: this looks awfully similar to MergeFrom functionality in controller-runtime https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/client/patch.go#L100-L114

Could we use this https://godoc.org/sigs.k8s.io/controller-runtime/pkg/client#MergeFrom?

pkg/controller/manager/manager_controller.go
```diff
@@ -190,12 +190,6 @@ func (r *ReconcileManager) Reconcile(request reconcile.Request) (reconcile.Resul
 	reqLogger.V(2).Info("Loaded config", "config", instance)
 	r.status.OnCRFound()
 
-	// Write the manager back to the datastore.
-	if err = r.client.Update(ctx, instance); err != nil {
```
ozdanborne, line 194: i'm curious how this could be removed without causing problems?

test/patch_test.go
```diff
@@ -0,0 +1,177 @@
...
+	AfterEach(func() {
+		// Clean up Calico data that might be left behind.
+		err := cleanInstallationResources(c)
+		Expect(err).ToNot(HaveOccurred())
+
+	})
+
+	Describe("Test patching Installation resource", func() {
+
+		table.DescribeTable("Blah",
```
ozdanborne, line 57: got a stray `blah` here

test/patch_test.go
```diff
@@ -0,0 +1,177 @@
...
+				},
+			),
+		)
+	})
+})
+
+func cleanInstallationResources(c client.Client) error {
+	ctx := context.Background()
+	instList := &operator.InstallationList{}
+	err := c.List(ctx, instList)
```
ozdanborne, line 165: just curious, what made you go for the `fv` test instead of a ut with fake clients? you know how much i like fake clients :)

ozdanborne (changes requested):

pkg/controller/manager/manager_controller.go
tmjd: I think this was needed when there was Auth stuff in Manager but that has now been moved to the Authentication CR. GetManager now checks that some fields are not set on Manager and reports they should be set on Authentication CR instead.

test/patch_test.go
tmjd: My recollection was that the fake clients don't handle Patching the resource. I'll try with a fake client though.

pkg/controller/utils/patch.go
tmjd: Thanks we can.

tmjd:
I think we should move forward with this PR with the known issue that if we added any fields to IPPool (or any array/slice where we set a default field) that utilizing the new field would/could be removed.

ozdanborne (approved):

PR #1028, opened 2020-12-14
Fix dex bug where svc.cluster.local was hardcoded.

rene-dekker:
Move method from logstorage_controller to separate file and re-use it for dex-related configuration.


pkg/render/dex_test.go
```diff
@@ -89,5 +105,16 @@ var _ = Describe("dex rendering tests", func() {
 				ExpectResource(resources[i], expectedRes.name, expectedRes.ns, expectedRes.group, expectedRes.version, expectedRes.kind)
 			}
 		})
+
+		It("should render the cluster dns properly", func() {
```
lmm, line 109: This test is testing the dex key validator config is rendering properly

lmm: It might be nice if we had a test for the non-default local search path case since we switch on that in the dex baseCfg

lmm (approved):
Left some minor comments, otherwise LGTM

pkg/render/dex_test.go
rene-dekker: There is currently no way to configure the resolv.conf path in `operator.tigera.io`. As discussed I moved a test to dns_test.go where this is tested instead.

pkg/common/common_suite_test.go
```diff
@@ -0,0 +1,30 @@
+// Copyright (c) 2019 Tigera, Inc. All rights reserved.
```
Brian-McM, line 1: Could you move this to a folder not named common, it's sort of a go anti pattern. I'd just add this to a package called dns and all dns related items could go there.

Brian-McM (changes requested):
Small thing, but can you rename the common package to dns or something a little more specific?  Common and util packages, although still used in some of our projects, is a golang antipattern.

pkg/dns/dns.go
```diff
@@ -0,0 +1,60 @@
+// Copyright (c) 2020 Tigera, Inc. All rights reserved.
+
+// Licensed under the Apache License, Version 2.0 (the "License");
```
Brian-McM, line 3: I don't think we need the Apache license, you may want to confirm with somebody else though.

Brian-McM (approved):
Looks good!

PR #1032, opened 2020-12-17
Add some checks for dirty and formating to ci

tmjd:
## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -83,7 +83,9 @@ LOCAL_USER_ID?=$(shell id -u $$USER)
 GO_BUILD_VER?=v0.45
 CALICO_BUILD?=calico/go-build:$(GO_BUILD_VER)
 SRC_FILES=$(shell find ./pkg -name '*.go')
-SRC_FILES+=$(shell find ./cmd -name '*.go')
+SRC_FILES+=$(shell find ./api -name '*.go')
+SRC_FILES+=$(shell find ./controllers -name '*.go')
+SRC_FILES+=main.go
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -303,6 +305,22 @@ static-checks:
...
 
+.PHONY: format-check
+format-check:
+	@$(CONTAINERIZED) \
+	sh -c '$(GIT_CONFIG_SSH) \
+	files=$$(gofmt -l ./pkg ./controllers ./api ./test); \
+	[ "$$files" = "" ] && exit 0; \
+	echo The following files need a format update:; \
+	echo $$files; \
+	echo Try running \"make fmt\" and committing any changes; \
```
doublek, line 316: I think in other repos we prefer using goimports. Could we use that instead?

tmjd: Those aren't the same things it seems. I tried running goimports on the operator and it changed several imports and they looked incorrect.

doublek: Perhaps a difference in go version between containerized and a local non-containerized builds?

doublek: To clarify, I did expect that goimports would do the work of `go fmt` as well as clean up imports. However if the imports look wrong, then my above comment could be one reason for this behavior.

tmjd: Ok yeah looks like if I containerize and use go-build to run goimports (which already exists under the `fix` make target) it doesn't change a bunch of stuff incorrectly.

tmjd: Is it ok using the target we already had that is called `fix` which runs goimport?

doublek: Awesome. I think that works.

doublek (approved):
LGTM

PR #1034, opened 2020-12-18
Make apiserver run as UID 0

tmjd:
## Description

Also fix PodSecurityPolicy to match what we have documented. This also
matches what it should be based on the settings we use.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/apiserver.go
+++ b/pkg/render/apiserver.go
@@ -665,8 +665,11 @@ func (c *apiServerComponent) apiServerContainer() corev1.Container {
 		Args:  c.startUpArgs(),
 		Env:   env,
 		// Needed for permissions to write to the audit log
-		SecurityContext: &corev1.SecurityContext{Privileged: &isPrivileged},
-		VolumeMounts:    volumeMounts,
+		SecurityContext: &corev1.SecurityContext{
+			Privileged: &isPrivileged,
+			RunAsUser:  Int64(0),
+		},
+		VolumeMounts: volumeMounts,
 		LivenessProbe: &corev1.Probe{
 			Handler: corev1.Handler{
 				HTTPGet: &corev1.HTTPGetAction{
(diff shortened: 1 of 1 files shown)


pkg/render/apiserver.go
```diff
@@ -1100,8 +1103,8 @@ func (c *apiServerComponent) tigeraNetworkAdminClusterRole() *rbacv1.ClusterRole
 func (c *apiServerComponent) apiServerPodSecurityPolicy() *policyv1beta1.PodSecurityPolicy {
 	psp := basePodSecurityPolicy()
 	psp.GetObjectMeta().SetName("tigera-apiserver")
-	psp.Spec.Privileged = true
-	psp.Spec.AllowPrivilegeEscalation = Bool(true)
+	psp.Spec.Privileged = false
+	psp.Spec.AllowPrivilegeEscalation = Bool(false)
 	psp.Spec.Volumes = append(psp.Spec.Volumes, policyv1beta1.HostPath)
 	psp.Spec.RunAsUser.Rule = policyv1beta1.RunAsUserStrategyRunAsAny
```
doublek, line 1109: Should this be set to run as root or UID 0 as well?

doublek: Talked offline about this and looks like we do this because
> We also set the PSPs to RunAsAny when we needed root since they operate over an entire pod so if you wanted to have an init container be something else, you don't want to force root in the PSP

doublek (approved):
LGTM

pkg/render/apiserver.go
tmjd: The decision was that it should not because it would force all containers to run as UID 0 even though we are only interested in the apiserver container (not queryserver) running as UID 0.

PR #1029, opened 2020-12-15
Use FQDN service references

lmm:
## Description

Windows nodes don't support PQDN resolution so this PR updates service name references to use the fully qualified version.

~~TODO: rebase on master once https://github.com/tigera/operator/pull/1028 is merged~~

## For PR author

- [ ] Tests for change.
- [ ] If changing pkg/apis/, run `make gen-files`
- [ ] If changing versions, run `make gen-versions`

## For PR reviewers

A note for code reviewers - all pull requests must have the following:

- [ ] Milestone set according to targeted release.
- [ ] Appropriate labels:
  - `kind/bug` if this is a bugfix.
  - `kind/enhancement` if this is a a new feature.
  - `enterprise` if this PR applies to Calico Enterprise only.

PR #1035, opened 2020-12-21
Follow-up fixes to cert SANs

lmm:
## Description

This PR follows up on https://github.com/tigera/operator/pull/1029
Using only FQDN service names for cert SANs breaks things in a few places. Notably, kube-apiserver proxies to API services using the `.svc` name.

- Refactor dns.GetLocalDNSName to return the cluster domain instead of `svc.<cluster_domain>` 
- Update certs to contain the service's qualified names in its SANs

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/dns/dns.go
+++ b/pkg/dns/dns.go
@@ -19,7 +19,6 @@ import (
 	"fmt"
 	"os"
 	"regexp"
-	"strings"
 )
 
 const (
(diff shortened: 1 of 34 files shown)


pkg/dns/dns.go
```diff
@@ -19,42 +19,66 @@ import (
...
+
+	// The FQDN service name will be: <svc_name>.<ns>.svc.<cluster-domain>
+	// There is no guarantee the cluster-domain will be dotted (kubelet does not
+	// validate the cluster domain).
+	if len(all) < 4 {
+		return nil, fmt.Errorf("failed to split FQDN service name %q", fqdnServiceName)
+	}
+
+	// We return:
+	// - <svc_name>.<ns>
```
caseydavenport, line 76: Do we need to handle the local-namespace access case as well? e.g., just `<svc_name>`?

pkg/dns/dns.go
```diff
@@ -19,42 +19,66 @@ import (
...
 	}
 
-	return localDNSName, nil
+	return clusterDomain, nil
+}
+
+// GetServiceDNSNames parses the fully-qualified domain service name nd
+// returns a list of its qualified service names.
+func GetServiceDNSNames(fqdnServiceName string, clusterDomain string) ([]string, error) {
+	all := strings.Split(fqdnServiceName, ".")
```
caseydavenport, line 66: Feels weird that we're doing a `fmt.Sprintf(` to build `fqdnServiceName` where we're calling this, and then immediately splitting it here and validating a certain number of elements.

I think this function would be more user-friendly if it just took the service name and namespace as separate string arguments instead of as a combined fqdn.

pkg/dns/dns.go
lmm: We don't need it but I'll add it for completeness.

pkg/dns/dns.go
lmm: If you're okay with it I think I'd like to leave this as-is and refactor in a follow-up PR I need to make

lmm: Fixed in 6ade567

caseydavenport (approved):

PR #1036, opened 2020-12-21
Disable Elasticsearch features based on license type

Suraiya-Hameed:
## Description

Disable features based on elastic search license type.

Kibana OIDC and ML jobs are not supported in ES basic license. Add watches to elastic-licensing configmap and disable these features if license type is basic.

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/logstorage/logstorage_controller.go
+++ b/pkg/controller/logstorage/logstorage_controller.go
@@ -167,6 +167,10 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 		return fmt.Errorf("log-storage-controller failed to watch the ConfigMap resource: %w", err)
 	}
 
+	if err = utils.AddConfigMapWatch(c, render.ECKLicenseConfigMapName, render.ECKOperatorNamespace); err != nil {
+		return fmt.Errorf("log-storage-controller failed to watch the ConfigMap resource: %w", err)
+	}
+
 	if err := utils.AddServiceWatch(c, render.ElasticsearchServiceName, render.ElasticsearchNamespace); err != nil {
 		return fmt.Errorf("log-storage-controller failed to watch the Service resource: %w", err)
 	}
@@ -351,6 +355,7 @@ func (r *ReconcileLogStorage) Reconcile(request reconcile.Request) (reconcile.Re
 
 	var elasticsearchSecrets, kibanaSecrets, curatorSecrets []*corev1.Secret
 	var clusterConfig *render.ElasticsearchClusterConfig
+	var elasticLicenseType render.ElasticLicenseType
 	applyTrial := false
 
 	if managementClusterConnection == nil {
@@ -394,6 +399,16 @@ func (r *ReconcileLogStorage) Reconcile(request reconcile.Request) (reconcile.Re
 		if err != nil {
 			r.status.SetDegraded("Failed to get eck trial license", err.Error())
 			return reconcile.Result{}, err
+		}
+
+		if elasticLicenseType, err = utils.GetElasticLicenseType(ctx, r.client, reqLogger); err != nil {
+			// If ECKLicenseConfigMapName is not found, it means ECK operator is not running yet, log the information and proceed
+			if errors.IsNotFound(err) {
+				log.Info("%s ConfigMap not found yet", render.ECKLicenseConfigMapName)
+			} else {
+				r.status.SetDegraded("Failed to get elastic license", err.Error())
+				return reconcile.Result{}, err
+			}
 		}
 	}
 
@@ -468,6 +483,7 @@ func (r *ReconcileLogStorage) Reconcile(request reconcile.Request) (reconcile.Re
 		r.clusterDomain,
 		applyTrial,
 		dexCfg,
+		elasticLicenseType,
 	)
 
 	if err := hdler.CreateOrUpdate(ctx, component, r.status); err != nil {
--- a/pkg/controller/utils/utils.go
+++ b/pkg/controller/utils/utils.go
@@ -299,3 +299,29 @@ func GetExpectedTyphaScale(nodes int) int {
 	}
 	return typhas
 }
+
+// GetElasticLicenseType returns the license type from elastic-licensing ConfigMap that ECK operator keeps updated.
+func GetElasticLicenseType(ctx context.Context, cli client.Client, logger logr.Logger) (render.ElasticLicenseType, error) {
+	cm := &corev1.ConfigMap{}
+	err := cli.Get(ctx, client.ObjectKey{Name: render.ECKLicenseConfigMapName, Namespace: render.ECKOperatorNamespace}, cm)
+	if err != nil {
+		return render.ElasticLicenseTypeUnknown, err
+	}
+	license, ok := cm.Data["eck_license_level"]
+	if !ok {
+		return render.ElasticLicenseTypeUnknown, fmt.Errorf("eck_license_level not available.")
+	}
+
+	return StrToElasticLicenseType(license, logger), nil
+}
+
+// StrToElasticLicenseType maps Elasticsearch license to one of the known and expected value.
+func StrToElasticLicenseType(license string, logger logr.Logger) render.ElasticLicenseType {
+	if license == string(render.ElasticLicenseTypeEnterprise) ||
+		license == string(render.ElasticLicenseTypeBasic) ||
+		license == string(render.ElasticLicenseTypeEnterpriseTrial) {
+		return render.ElasticLicenseType(license)
+	}
+	logger.V(3).Info("Elasticsearch license %s is unexpected", license)
+	return render.ElasticLicenseTypeUnknown
+}
--- a/pkg/render/intrusion_detection.go
+++ b/pkg/render/intrusion_detection.go
@@ -50,28 +50,31 @@ func IntrusionDetection(
 	pullSecrets []*corev1.Secret,
 	openshift bool,
 	clusterDomain string,
+	elasticLicenseType ElasticLicenseType,
 ) Component {
 	return &intrusionDetectionComponent{
-		lc:               lc,
-		esSecrets:        esSecrets,
-		kibanaCertSecret: kibanaCertSecret,
-		installation:     installation,
-		esClusterConfig:  esClusterConfig,
-		pullSecrets:      pullSecrets,
-		openshift:        openshift,
-		clusterDomain:    clusterDomain,
+		lc:                 lc,
+		esSecrets:          esSecrets,
+		kibanaCertSecret:   kibanaCertSecret,
+		installation:       installation,
+		esClusterConfig:    esClusterConfig,
+		pullSecrets:        pullSecrets,
+		openshift:          openshift,
+		clusterDomain:      clusterDomain,
+		elasticLicenseType: elasticLicenseType,
 	}
 }
 
 type intrusionDetectionComponent struct {
-	lc               *operatorv1.LogCollector
-	esSecrets        []*corev1.Secret
-	kibanaCertSecret *corev1.Secret
-	installation     *operator.InstallationSpec
-	esClusterConfig  *ElasticsearchClusterConfig
-	pullSecrets      []*corev1.Secret
-	openshift        bool
-	clusterDomain    string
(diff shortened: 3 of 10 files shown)


pkg/controller/utils/utils.go
```diff
@@ -299,3 +300,22 @@ func GetExpectedTyphaScale(nodes int) int {
...
 	return typhas
 }
+
+// GetElasticLicenseType returns the Elasticsearch license type from elastic-licensing configmap that ECK operator keeps updated.
+// If elastic-licensing configmap doesn't exists, it must be a fresh install, return "basic" license.
+func GetElasticLicenseType(ctx context.Context, cli client.Client) (render.ElasticLicenseType, error) {
+	cm := &corev1.ConfigMap{}
+	err := cli.Get(ctx, client.ObjectKey{Name: render.ECKLicenseConfigMapName, Namespace: render.ECKOperatorNamespace}, cm)
+	if err != nil {
+		if errors.IsNotFound(err) {
```
Brian-McM, line 310: Does the ECK operator create a basic license config map if there's no applied config map? If it does, then you should return "Unknown" here, if it doesn't, then ignore this comment.

Brian-McM:
I'm not 100% sure, but doesn't the intrusion detection job create some required indices and dashboards? It may be that the controller creates certain indices, but https://github.com/tigera/intrusion-detection/blob/master/install/config.yaml#L51 is used for the dashboards.

pkg/controller/utils/utils.go
Suraiya-Hameed: For fresh install of caliEnt, elasticsearch license will be basic by default, plus during the fresh install we won't have any existing `elastic-licensing` configmap until ECK operator is running, so we return "basic" if config map is not found.
Once ECK operator is running, it always creates this config map irrespective of license type. Only possible scenario where config map is not found is during fresh install.

Suraiya-Hameed:
> I'm not 100% sure, but doesn't the intrusion detection job create some required indices and dashboards? It may be that the controller creates certain indices, but https://github.com/tigera/intrusion-detection/blob/master/install/config.yaml#L51 is used for the dashboards.

Thanks for pointing this :+1: 
I've made changes to pass the Elasticsearch license type as env variable to intrusion detection job, there if the license type is basic it will just apply the kibana dashboard related stuff else proceed as usual.

pkg/controller/utils/utils.go
Brian-McM: > Once ECK operator is running, it always creates this config map irrespective of license type

With this in mind I think we should not proceed on anything that requires having knowledge of the license, let's wait until the ECK operator creates the license.

When we have the option to we should avoid assuming assuming configuration for third party software, in this case we just need to wait for ECK to tell us what configuration it will apply. If it changes in the future we won't have a scenario where we temporarily use a basic license then immediately switch to a different one.

pkg/controller/utils/utils.go
```diff
@@ -299,3 +300,23 @@ func GetExpectedTyphaScale(nodes int) int {
...
 	return typhas
 }
+
+// GetElasticLicenseType returns the license type from elastic-licensing ConfigMap that ECK operator keeps updated.
+func GetElasticLicenseType(ctx context.Context, cli client.Client, logger logr.Logger) (render.ElasticLicenseType, error) {
+	cm := &corev1.ConfigMap{}
+	err := cli.Get(ctx, client.ObjectKey{Name: render.ECKLicenseConfigMapName, Namespace: render.ECKOperatorNamespace}, cm)
+	if err != nil {
+		// If ConfigMap is not available, it means ECK operator is not running yet, so log the information and proceed
+		if errors.IsNotFound(err) {
```
Brian-McM, line 310: You don't need / shouldn't have a check here for is not found, you need to return an error so that we don't progress, so just return the error as is

Brian-McM (changes requested):

pkg/controller/utils/utils.go
Suraiya-Hameed: This function is called in two controllers and we need this check to differentiate 2 scenario.
- in intrusion-detection, if elastic-licensing is not found or if there is an from this function, we don't install ml jobs
- in logstorage, if elastic-licensing is not found we need to proceed with rendering ECK operator related resource, if there is any other error in `GetElasticLicenseType` we don't progress

Brian-McM: I think you should just check if the error is a not found one in the operator and continue, and just leave this simple. As far as this function is concerned we should consider not found an error, and if the caller would like to ignore that error then it can do so but checking if it's a not found error.

The code you have of `err != nil || elasticLicenseType == render.ElasticLicenseTypeUnknown` should really just be `err != nil && errors.IsNotFound(err)` (always return Unkown if there's an error).

pkg/controller/utils/utils.go
```diff
@@ -299,3 +300,23 @@ func GetExpectedTyphaScale(nodes int) int {
...
+			return render.ElasticLicenseTypeUnknown, nil
+		}
+		return render.ElasticLicenseTypeUnknown, err
+	}
+	license, ok := cm.Data["eck_license_level"]
+	if !ok {
+		return render.ElasticLicenseTypeUnknown, fmt.Errorf("eck_license_level not available.")
+	}
+
+	return render.ElasticLicenseType(license), nil
```
Brian-McM, line 321: I think that you should write a function `render.StrToElasticLicenseType` to guarantee this only returns types that we want to consider. For instance, if don't want to consider somebody who put in a platinum license then this would return unknown if the type was platinum, then we just check if the type is unknown. 

This gives the Unknown type a real meaning besides just what to return for an error, it's Unknown to us but not necessarily an error, but a caller _could_ treat it as an error if it requires a known license type.

There's almost no point in using a specific type if you're just going to cast every string value as the ElasticLicenseType since it doesn't validate that it's a type you've defined.

Brian-McM (changes requested):

Brian-McM (approved):

pkg/render/logstorage.go
```diff
@@ -330,7 +341,7 @@ func (es *elasticsearchComponent) Objects() ([]runtime.Object, []runtime.Object)
 		)
 	}
 
-	if es.dexCfg != nil {
+	if es.dexCfg != nil && es.elasticLicenseType != ElasticLicenseTypeBasic {
```
tmjd, line 344: To be clear about what would happen here if elasticLicenseType was Unknown and then went to TypeBasic (and dex was configured).
While the type was unknown we would be configuring dex integration. Would ES just ignore the config or it would error with the config provided?
Once the type was available and Basic the configuration would be updated and ES would run normal.

pkg/render/intrusion_detection.go
```diff
@@ -191,6 +195,10 @@ func (c *intrusionDetectionComponent) intrusionDetectionJobContainer() v1.Contai
 				Name:  "CLUSTER_NAME",
 				Value: c.esClusterConfig.ClusterName(),
 			},
+			{
+				Name:  "ELASTIC_LICENSE_TYPE",
+				Value: string(c.elasticLicenseType),
```
tmjd, line 200: Does it matter how Intrusion treats an empty ELASTIC_LICENSE_TYPE?
Is there any problem if initially this uses the wrong License?
It seems likely that this can be ran with the wrong value since it is a job. Since the controller will launch this if the license does not exist, I'm concerned on the impact if it is wrong.
What I'm kind of getting at is, should the controller require the license config?

pkg/render/logstorage.go
Suraiya-Hameed: > While the type was unknown we would be configuring dex integration. Would ES just ignore the config or it would error with the config provided?

Ideally it should never be unknown as we are reading license values written by ECK operator, if ECK adds support to new license type we will have to catch that during ECK version update and change this condition. Among the licenses that ECK now supports, basic is only type that doesn't support SSO.
If license type is unknown and dex is configured, and say if ES doesn't support OIDC then, ES will work but it will log errors about OIDC and also Kibana will throw license related error.

pkg/render/intrusion_detection.go
tmjd: Nevermind, I'm wrong. I see the controller bails if the license type is not found.

pkg/render/logstorage.go
tmjd: Would it be better to ensureLicenseType is Enterprise or EnterpriseTrial? Then we'll be sure we have read the value and won't ever configure Dex without it being a type that supports OIDC?
Is there a downside to do that?

pkg/render/intrusion_detection.go
Suraiya-Hameed: If ELASTIC_LICENSE_TYPE is set to basic, we skip installing ml jobs in es-job-installer. If it is empty or has a unknown value, it will continue as usual, that is install ml jobs and apply kibana dashboards. So it is basically the same behaviour as before, except we don't install ml jobs if license is basic.

>  Is there any problem if initially this uses the wrong License?

Say if we switch from basic to enterprise, since we are watching `elastic-licensing` configmap, this job will get restarted and ml jobs will get installed, shouldn't be any problem.
For other way around, enterprise to basic, as xpack is not available anymore ml jobs will not run.

I haven't changed anything in the intrusion-detection-controller, it will continue watching for ml jobs and create reports/alerts if job is available, else log error. It does seem to be flood with error, I'll add logic in operator to set `DISABLE_ANOMALY=yes` if license type is basic to avoid all the errors.

pkg/render/logstorage.go
Suraiya-Hameed: yes, I've updated to check for Enterprise or EnterpriseTrial.

pkg/render/intrusion_detection.go
```diff
@@ -357,6 +365,17 @@ func (c *intrusionDetectionComponent) deploymentPodTemplate() *corev1.PodTemplat
 		}
 	}
 
+	container := ElasticsearchContainerDecorateIndexCreator(
+		ElasticsearchContainerDecorate(c.intrusionDetectionControllerContainer(), c.esClusterConfig.ClusterName(), ElasticsearchIntrusionDetectionUserSecret, c.clusterDomain),
+		c.esClusterConfig.Replicas(), c.esClusterConfig.Shards())
+
+	if c.elasticLicenseType == ElasticLicenseTypeBasic {
+		envVars := []corev1.EnvVar{
+			{Name: "DISABLE_ANOMALY", Value: "yes"},
```
tmjd, line 374: Since I don't know what this does can you get someone to review that understands the ML job?

pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -395,6 +400,16 @@ func (r *ReconcileLogStorage) Reconcile(request reconcile.Request) (reconcile.Re
 			r.status.SetDegraded("Failed to get eck trial license", err.Error())
 			return reconcile.Result{}, err
 		}
+
+		if elasticLicenseType, err = utils.GetElasticLicenseType(ctx, r.client, reqLogger); err != nil {
+			// If ECKLicenseConfigMapName is not found, it means ECK operator is not running yet, log the information and proceed
```
Brian-McM, line 405: I'm just commenting for the sake of it, but maybe it would make sense for separate controllers for the ECK operator and Elasticsearch?

Please don't consider this for this PR, I'm just putting this thought out inter the ether.

Brian-McM (approved):

pkg/render/intrusion_detection.go
Brian-McM: I'm comfortable with this change since watchers aren't supported, but do we also need to set DISABLE_ALERTS to true as well?

Brian-McM (changes requested):
Could you please check if we need to set DISABLE_ALERTS to yes for intrusion detection as well, I think the alerts has a watcher which means with a basic license it will spit out errors.

tmjd (approved):
LGTM
Please make sure you squash before or when you merge.
One minor thing the Copyright in utils_test.go, should be updated to 2021.

marvin-tigera:

marvin-tigera:

Brian-McM (approved):

marvin-tigera:

PR #1037, opened 2020-12-22
Add certificate management for calico and minimal enterprise clusters.

rene-dekker:
First draft of certificate management


api/v1/installation_types.go
```diff
@@ -492,3 +496,49 @@ type InstallationList struct {
...
+	KeyAlgorithmRSAWithSize8192   = "RSAWithSize8192"
+	KeyAlgorithmECDSAWithCurve256 = "ECDSAWithCurve256"
+	KeyAlgorithmECDSAWithCurve384 = "ECDSAWithCurve384"
+	KeyAlgorithmECDSAWithCurve521 = "ECDSAWithCurve521"
+)
+
+// Signature algorithm for certificate signing requests.
+type SignatureAlgorithm string
+
+const (
```
rene-dekker, line 536: The string values in here are equivalent to the enum in the x509 package.

api/v1/installation_types.go
```diff
@@ -100,6 +100,10 @@ type InstallationSpec struct {
 	// ComponentResources can be used to customize the resource requirements for each component.
 	// +optional
 	ComponentResources []ComponentResource `json:"componentResources,omitempty"`
+
+	// This is a temporary placeholder until the design is ready.
+	//
+	CertificateManagement *CertificateManagement
```
rene-dekker, line 106: Current working name for the feature.

bcreane:
@rene-dekker as discussed, k8s CSR integration for Calico pods should be OSS first and then Enterprise to follow.

asincu:
@rene-dekker You might wanna check the build :

```
et: pkg/render/typha_test.go:228:80: too few arguments in call to render.Typha
```

pkg/render/apiserver.go
```diff
@@ -47,30 +47,33 @@ func APIServer(k8sServiceEndpoint k8sapi.ServiceEndpoint, installation *operator
...
-			APIServerTLSSecretName,
-			APIServerSecretKeyName,
-			APIServerSecretCertName,
-			DefaultCertificateDuration,
-			nil,
-			svcDNSNames...,
-		)
-		if err != nil {
-			return nil, err
+	if installation.CertificateManagement == nil {
```
asincu, line 50:
```suggestion
	if installation.CertificateManagement == nil && tlsKeyPair == nil {
```

pkg/render/csr.go
```diff
@@ -0,0 +1,122 @@
...
+	mountName string,
+	commonName string,
+	keyName string,
+	certName string,
+	dnsNames []string,
+	registerApiserver bool) corev1.Container {
+	return corev1.Container{
+		Name:            "key-cert-provisioner",
+		Image:           components.GetReference(components.ComponentCSRInitContainer, "docker.io/", installation.ImagePath), //todo: Set the right registry.
+		ImagePullPolicy: "Always",                                                                                            //todo: delete this line.
```
asincu, line 44: We should have the same policy as the rest of our containers.

asincu: And resolve the //todo:s

pkg/render/csr.go
```diff
@@ -0,0 +1,122 @@
...
+		},
+	}
+}
+
+// csrClusterRole returns a role with the necessary permissions to create certificate signing requests.
+func csrClusterRole() runtime.Object {
+	return &rbacv1.ClusterRole{
+		TypeMeta: metav1.TypeMeta{Kind: "ClusterRole", APIVersion: "rbac.authorization.k8s.io/v1"},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:   "tigera-csr-creator",
```
asincu, line 86: We should extract the names as constants. We can use them as validation in tests and in reference to other resources (role bindings)

pkg/render/csr.go
```diff
@@ -0,0 +1,122 @@
...
+		},
+	}
+}
+
+// csrClusterRoleBinding returns a role binding with the necessary permissions to create certificate signing requests.
+func csrClusterRoleBinding(name, namespace string) *rbacv1.ClusterRoleBinding {
+	crb := &rbacv1.ClusterRoleBinding{
+		TypeMeta: metav1.TypeMeta{Kind: "ClusterRoleBinding", APIVersion: "rbac.authorization.k8s.io/v1"},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:   fmt.Sprintf("%s-%s:csr-creator", namespace, name),
```
asincu, line 105: Why do we need the namespace since this is a cluster role ?

pkg/render/node.go
```diff
@@ -486,16 +495,29 @@ func (c *nodeComponent) clusterAdminClusterRoleBinding() *rbacv1.ClusterRoleBind
...
+	var initContainers []v1.Container
 
 	annotations := make(map[string]string)
 	if len(c.birdTemplates) != 0 {
 		annotations[birdTemplateHashAnnotation] = AnnotationHash(c.birdTemplates)
 	}
 	annotations[typhaCAHashAnnotation] = AnnotationHash(c.typhaNodeTLS.CAConfigMap.Data)
-	annotations[nodeCertHashAnnotation] = AnnotationHash(c.typhaNodeTLS.NodeSecret.Data)
+	if c.cr.CertificateManagement == nil {
+		annotations[nodeCertHashAnnotation] = AnnotationHash(c.typhaNodeTLS.NodeSecret.Data)
```
asincu, line 509: :thinking: If we are no longer using annotations to manage certificates ... Do we delete older annotations once you apply a CertificateManagement CR ?

pkg/render/node.go
```diff
@@ -610,6 +631,19 @@ func (c *nodeComponent) nodeVolumes() []v1.Volume {
 	fileOrCreate := v1.HostPathFileOrCreate
 	dirOrCreate := v1.HostPathDirectoryOrCreate
 
+	var felixVolume v1.VolumeSource
```
asincu, line 634: I see this being code duplicated in multiple files. How about we define a function that creates a generic volume, like 

```
func CertificateVolume(certMgmt *CertificateManagement, secretName string) v1.VolumeSource {
var volume v1.VolumeSource
if cerMgmt != nil {
	volume = v1.VolumeSource{
		EmptyDir: &v1.EmptyDirVolumeSource{},
	}
} else {
	volume = v1.VolumeSource{
			SecretName: secretName
		},
	}
return volume
}


```

pkg/render/csr.go
```diff
@@ -0,0 +1,122 @@
...
+	keyName string,
+	certName string,
+	dnsNames []string,
+	registerApiserver bool) corev1.Container {
+	return corev1.Container{
+		Name:            "key-cert-provisioner",
+		Image:           components.GetReference(components.ComponentCSRInitContainer, "docker.io/", installation.ImagePath), //todo: Set the right registry.
+		ImagePullPolicy: "Always",                                                                                            //todo: delete this line.
+		VolumeMounts: []corev1.VolumeMount{
+			{MountPath: "/secret", Name: mountName, ReadOnly: false},
```
asincu, line 46: :thinking: Why does the secret need to writable ?

api/v1/installation_types.go
```diff
@@ -492,3 +498,28 @@ type InstallationList struct {
 func init() {
 	SchemeBuilder.Register(&Installation{}, &InstallationList{})
 }
+
+// CertificateManagement configures pods to submit a CertificateSigningRequest to API certificates.k8s.io/v1beta1 in order
+// to obtain TLS certificates. This feature requires that you bring your own CSR signing and approval process, otherwise
+// pods will be stuck during initialization.
+type CertificateManagement struct {
+	// Root CA certificate of the certificate authority that signs the certificate requests in PEM format.
+	RootCA string `json:"rootCA"`
```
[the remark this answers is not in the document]
rene-dekker: It can. You think it would be better to name it CACert?

pkg/render/csr.go
rene-dekker: An init container will be added to the pod. That init container will submit a CSR and then write the results (private key + cert) to this location. The main container will then read from this location.

pkg/render/apiserver.go
rene-dekker: I think that will change the result in ways we do not want.
There are three possibilities:
- cert management enabled  => do nothing
- cert management disabled and no tls key pair exists => create key pair and then copy key pair
- cert management disabled and key pair exists => copy key pair

api/v1/installation_types.go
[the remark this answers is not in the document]
asincu: I suggest we rename it.

pkg/render/csr.go
asincu: Hm ... Can we make the secret read only and use a different path to write the certificates ?

pkg/render/apiserver.go
```diff
@@ -354,6 +365,18 @@ func (c *apiServerComponent) apiServiceAccountClusterRole() *rbacv1.ClusterRole
 		})
 	}
 
+	if c.installation.CertificateManagement != nil {
```
Brian-McM: So our API server was never registered as an API server with k8s?

pkg/render/csr.go
Brian-McM: > Hm ... Can we make the secret read only and use a different path to write the certificates ?
The init container and main container are separate and the init container creates a tls cert that the main container needs  so there _must_ be a shared volume that the init container can write to.

I'm not a huge fan of the directory name though, secret is pretty vague.

pkg/render/node.go
Brian-McM: The operator already takes care of this, the node daemonset would be deployed and the previous annotations would be overwritten (e.g. if the new dameonset does not have an annotation it will not be there in the deployed daemonset).

pkg/render/csr.go
rene-dekker: "/secret" is just a mount path name and has no relationship to kubernetes secrets. It is the location where the in-memory volume is mounted. For the initcontainer it needs write permissions, and the main container has read permissions.

pkg/render/csr.go
rene-dekker: I suppose it is not really necessary, I'll remove it.

pkg/render/csr.go
rene-dekker: Lemme change that name then.

pkg/render/apiserver.go
rene-dekker: This comment seems a bit out of place, but I can tell that you responded to - now - outdated code. The tigera-apiserver _was_ and _is_ registered by the operator as an apiService. This is a special object for aggregated apiservers.

That being said there was a brief moment where this PR was not registering it for Cert management, but that change has been rolled back as we discussed earlier this week.

pkg/render/csr.go
asincu: Sounds good, let's change it.

asincu (approved):

config/calico_versions.yml
```diff
@@ -9,3 +9,6 @@ components:
     version: master
   flexvol:
     version: master
+  rdtigera/init-container:
```
caseydavenport, line 12: what's the `rd` stand for?

rene-dekker: My initials. That is my personal docker.io. I plan on changing these next week when we have a hooked up the repo to our build processes.

hack/gen-versions/components.go
```diff
@@ -36,6 +36,7 @@ var defaultImages = map[string]string{
 	"eck-kibana":              "tigera/kibana",
 	"guardian":                "tigera/guardian",
 	"tigera-cni":              "tigera/cni",
+	"rdtigera/init-container": "rdtigera/init-container",
```
caseydavenport, line 39: I think the key in this map is just a name for the image, so can be `tigera/init-container`.

Also, I think we should change the name of this container. `init-container` is a very vague term and doesn't describe what the container does. The container should be called something like `tigera/cert-requester` or something like that.

pkg/controller/status/status.go
```diff
@@ -409,6 +430,15 @@ func (m *statusManager) syncState() bool {
 		}
 	}
 
+	for csrLabel := range m.certificatestatusrequests {
+		pending, err := hasPendingCSR(context.TODO(), m.client, csrLabel)
+		if err != nil {
+			log.WithValues("error", err).Info(fmt.Sprintf("Unable to poll for CertificateSigningRequest(s) with label value %s", csrLabel))
```
caseydavenport, line 436: Should this be an error level log?

pkg/controller/status/status.go
```diff
@@ -639,3 +669,35 @@ func (m *statusManager) degradedReason() string {
...
+	for _, csr := range csrs.Items {
+		if len(csr.Status.Conditions) == 0 {
+			// No conditions means status is pending
+			return true, nil
+		}
+		// no condition approved, means it is pending.
+		for _, condition := range csr.Status.Conditions {
+			if condition.Status == v1.ConditionUnknown {
+				return true, nil
+			} else if condition.Type == v1beta1.CertificateApproved && csr.Status.Certificate == nil {
```
caseydavenport, line 697: Not sure I full understand this - what is the series of states that a CSR goes through? Could we simply say that if `csr.Status.Certificate` is nil that the signing procedure hasn't completed yet?

pkg/controller/status/status.go
```diff
@@ -216,6 +223,13 @@ func (m *statusManager) AddCronJobs(cjs []types.NamespacedName) {
 	}
 }
 
+// AddCertificateSigningRequests tells the status manager to monitor the health of the given CertificateSigningRequests.
```
caseydavenport, line 226: Would be worth extending the comment here to explain that the "label" argument is the value of the `k8s-app` label on the CSR itself. Alternatively, to make this code more generic, we could pass both the key and value to this function.

As it is, I believe there is some tribal knowledge here, since the current implementation of the status manager code is highly-dependent on the logic of the CSR creation code (the two need to be in-sync on labels).

pkg/render/typha.go
```diff
@@ -308,9 +316,23 @@ func (c *typhaComponent) typhaDeployment() *apps.Deployment {
...
+	var initContainers []v1.Container
 	annotations := make(map[string]string)
 	annotations[typhaCAHashAnnotation] = AnnotationHash(c.typhaNodeTLS.CAConfigMap.Data)
-	annotations[typhaCertHashAnnotation] = AnnotationHash(c.typhaNodeTLS.TyphaSecret.Data)
+	if c.installation.CertificateManagement == nil {
+		annotations[typhaCertHashAnnotation] = AnnotationHash(c.typhaNodeTLS.TyphaSecret.Data)
+	} else {
+		annotations[typhaCertHashAnnotation] = AnnotationHash(c.installation.CertificateManagement.CACert)
+		initContainers = append(initContainers, CreateCSRInitContainer(
+			c.installation,
```
caseydavenport, line 327: Do we need to pass both the `installation` as well as a subfield of the installation? (`installation.CertificateManagement`)

Maybe we could just remove the second argument?

pkg/render/typha.go
```diff
@@ -434,6 +453,25 @@ func (c *typhaComponent) typhaResources() v1.ResourceRequirements {
...
+		}
+	} else {
+		cnEnv = v1.EnvVar{
+			Name: "TYPHA_CLIENTCN", ValueFrom: &v1.EnvVarSource{
+				SecretKeyRef: &v1.SecretKeySelector{
+					LocalObjectReference: v1.LocalObjectReference{
+						Name: NodeTLSSecretName,
+					},
+					Key:      CommonName,
+					Optional: Bool(true),
```
caseydavenport, line 469: Why does this need `optional: true` on it?

caseydavenport: I guess it used to have it before. Hm. I wonder why.

caseydavenport (approved):

hack/gen-versions/components.go
rene-dekker: We discussed this offline, a follow-up will change this to tigera/key-cert-provisioner

pkg/controller/status/status.go
rene-dekker: Excellent question. I think my docs change is the best answer to your question as I made visible what the pod statuses look like: https://docs.tigera.io/master/security/comms/certificate-management

TLDR: two separate endpoints are necessary to go from pending to completely done:
- Add certificate
- Approve
While the docs state that it should be first approved before signing, in practice this can happen in any order.

pkg/controller/status/status.go
rene-dekker: While this is code-wise more succinct I agree that it would be better to stick to the conventional data structures here. I will update my PR accordingly.

pkg/controller/status/status.go
rene-dekker: Lastly, to give a better answer to your question It is possible that a CSR is rejected/denied or if there is another non-approved condition. In that case there is not a certificate and at the same time, the CSR is not pending (nor should the tigerastatus). The init container will then crash, delete the CSR and try again on next init until it is approved and signed.

bcreane:
merging based on @caseydavenport and @asincu review.
/merge-when-ready delete-branch

PR #1041, opened 2020-12-23
Remove the validatingwebhookconfiguration for elasticsearch.

rene-dekker:
The webhookconfiguration does us more harm than good, so this PR is removing it from the operator.

@caseydavenport 
Does this mean I can also delete this? I haven't included it yet. (logstorage.go)
```
	if es.installation.KubernetesProvider == operatorv1.ProviderEKS &&
		es.installation.CNI.Type == operatorv1.PluginCalico {
		// Workaround the fact that webhooks don't work for non-host-networked pods
		// when in this networking mode on EKS, because the control plane nodes don't run
		// Calico.
		hostNetwork = true

		// Adjust DNS policy so we can access in-cluster services.
		dnsPolicy = corev1.DNSClusterFirstWithHostNet
	}
```


matthewdupre (approved):
LGTM

PR #1043, opened 2020-12-23
Nodeselectors tolerations

ozdanborne:
## Description

**Background / Use Case**

A lot of users have been asking for the ability to set custom nodeSelectors and tolerations on resources launched by the Operator. The initial request was to be able to set it individually for each individual component. But when we asked about their use case, it became clear that users are mainly using nodeSelectors and tolerations to place infrastructure components (like calico) on separate, more persistent infrastructure nodes, while running their business logic workloads on worker nodes.

This use case closely matches what the (arguably poorly named) controlPlaneNodeSelector field was designed for. However, controlPlaneNodeSelector only is applied to the OS product.

**Changes**

This PR does the following:
- Extends `controlPlaneNodeSelector` to all other components, with a few exceptions:
  - It still does not apply to Typha, as it's use case is different.
  - It still does not apply to daemonsets like fluentd, calico-node, or compliance-benchmarker.
- Introduce a new `controlPlaneToleration` which applies to all components (unless they already tolerate all taints).

Note that the LogStorage CR already has a `DataNodeSelector` field. As part of this PR, that field will now override the value set in `controlPlaneNodeSelector`.

To accomplish this, I've added some helper methods to be able to more easily group our tolerations together. 
I also removed the 'critical-pod' annotation since it has been deprecated in favor of the toleration since v1.16.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/fluentd.go
+++ b/pkg/render/fluentd.go
@@ -709,6 +709,7 @@ func (c *fluentdComponent) eksLogForwarderDeployment() *appsv1.Deployment {
 					Annotations: annots,
 				},
 				Spec: corev1.PodSpec{
+					Tolerations:        c.installation.ControlPlaneTolerations,
 					ServiceAccountName: eksLogForwarderName,
 					ImagePullSecrets:   getImagePullSecretReferenceList(c.pullSecrets),
 					InitContainers: []corev1.Container{ElasticsearchContainerDecorateENVVars(corev1.Container{
--- a/pkg/render/manager_test.go
+++ b/pkg/render/manager_test.go
@@ -21,6 +21,7 @@ import (
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	appsv1 "k8s.io/api/apps/v1"
+	v1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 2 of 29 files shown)


tmjd (commented):
My concern is the change on fluentd.

pkg/render/fluentd.go
```diff
@@ -273,7 +273,7 @@ func (c *fluentdComponent) daemonset() *appsv1.DaemonSet {
 		},
 		Spec: ElasticsearchPodSpecDecorate(corev1.PodSpec{
 			NodeSelector:                  map[string]string{},
-			Tolerations:                   c.tolerations(),
+			Tolerations:                   tolerateAll(),
```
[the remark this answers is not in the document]
ozdanborne: Yeah, I agree, I've restored it back to just the two effects. 

I was assuming that since all our other components tolerate those two plus CriticalAddonsOnly, it's safe to do the same here, but lets keep functionality the same in this PR :+1:

pkg/render/manager_test.go
```diff
@@ -334,6 +335,49 @@ var _ = Describe("Tigera Secure Manager rendering tests", func() {
...
+	It("should apply controlPlaneTolerations", func() {
+		t := corev1.Toleration{
+			Key:      "foo",
+			Operator: corev1.TolerationOpEqual,
+			Value:    "bar",
+		}
+		deployment := renderManager(&operator.InstallationSpec{
+			ControlPlaneTolerations: []corev1.Toleration{t},
+		})
+		Expect(deployment.Spec.Template.Spec.Tolerations).To(ContainElement(t))
```
tmjd, line 379: Should include master and CriticalAddonsOnly here

pkg/render/render_utils_test.go
```diff
@@ -91,3 +92,30 @@ func ExpectEnv(env []v1.EnvVar, key, value string) {
 	}
 	Expect(false).To(BeTrue(), fmt.Sprintf("Missing expected environment variable %s", key))
 }
+
+var (
+	tolerateMaster = corev1.Toleration{
```
tmjd, line 97: Is it necessary and beneficial to re-define these here?

tmjd (approved):
I'm not sure if I like the common.go tolerations as variables. I wonder if they should be functions so they can't be modified. This isn't critical so LGTM, we can change it later if we want.

pkg/render/render_utils_test.go
ozdanborne: This test is a separate `_test` package, so the 'private' variables in the render package aren't accessible. I didn't want to  make the toleration variables external (by capitalizing) so I redefined.

I'm open to input on how to better set this up

diranged:
@ozdanborne Can you elaborate on why the typha use case is different? We want to run Typha on our "system" nodes, but but we need to be able to set the toleration/selector on it..

PR #1046, opened 2020-12-29
expose option to set typha affinity

ozdanborne:
## Description

This PR exposes a new field on the Installation resource which allows users to specify an Affinity for Typha. 

This new field uses custom types so that only the NodeAffinity (not PodAffinity) and only PreferredDuringSchedulingIgnoredDuringExecution (not RequiredDuringSchedulingIgnoredDuringExecution) are exposed. This is because Typha is crucial to Calico-Node functioning, and if the pool of nodes which meet the affinity are exhausted, we'd rather typha get scheduled elsewhere than take the whole cluster offline.

This PR also includes migration logic so that if this field is set on a non-operator-managed cluster, the field will be detected and populated instead of ignored.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


pkg/render/typha.go
```diff
@@ -342,6 +342,7 @@ func (c *typhaComponent) typhaDeployment() *apps.Deployment {
 				},
 				Spec: v1.PodSpec{
 					Tolerations:                   c.tolerations(),
+					Affinity:                      toAffinity(c.installation.TyphaAffinity),
```
ozdanborne, line 345: Is there a better way to convert from our custom `TyphaAffinity` type to the near-identical `Affinity` type? I have all the subfields with matching names and pointer types, but am not able to cast it. I think this is because you cannot cast unless the structs are identical, and in this case our struct is a subset of the other.

If that's true, I was considering adding a `ToAffinity` receiver method off of the `TyphaAffinity` struct. But I haven't seen any other examples of receiver methods being added (besides the `DeepCopy` autogenerated methods.

caseydavenport: I think a helper like this is fine, you could do a receiver on our API type if you wanted. I think we've done that elsewhere in some of the elastic stuff maybe?

As for implementation, this seems fine

pkg/controller/migration/convert/core.go
```diff
@@ -303,11 +303,28 @@ func handleNodeSelectors(c *components, install *operatorv1.Installation) error
...
 	// check typha nodeSelectors
 	if c.typha != nil {
-		if c.typha.Spec.Template.Spec.Affinity != nil {
-			return ErrIncompatibleCluster{
-				err:       "node affinity not supported for typha deployment",
-				component: ComponentTypha,
-				fix:       "remove the affinity",
+		// we can migrate typha affinities provided they are a NodeAffinity for Preferred.
+		if aff := c.typha.Spec.Template.Spec.Affinity; aff != nil {
+			if aff.PodAffinity != nil || aff.PodAntiAffinity != nil {
```
caseydavenport, line 308: What if the pod anti-affinity matches the one that we're going to install ourselves?

ozdanborne: The pod anti-affinity you're referring to is applied during the namespace migration, which occurs long after this code here is completed. This code here should never be run when a namespace migration is in progress, so we shouldn't ever block our own migration from happening :+1:

caseydavenport (approved):

PR #1048, opened 2020-12-30
avoid virtual nodes in aks

ozdanborne:
## Description

AKS has a feature called [virtual node](https://docs.microsoft.com/en-us/azure/aks/virtual-nodes) which is an adapter that allows use of Azure Container Service (ACS) in a Kubernetes cluster. Pods running there have many limitations, most notably cannot volume mount to the host, resulting in this error:

>Pod calico-node requires volume which is unsupported type.

The node is tainted to prevent pods like typha and kube-controllers from running there, but Calico-node tolerates all taints. As such, they prevent calico-node from running on it using an Affinity.

```yaml
affinity:
  nodeAffinity:
    requiredDuringSchedulingIgnoredDuringExecution:
      nodeSelectorTerms:
      - matchExpressions:
        - key: kubernetes.azure.com/cluster
          operator: Exists
        - key: type
          operator: NotIn
          values:
          - virtual-kubelet
```

This PR prevents calico-node from running on these nodes by setting an affinity to block type=virtual-kubelet if kubernetesProvider=AKS.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/migration/convert/core.go
+++ b/pkg/controller/migration/convert/core.go
@@ -281,10 +281,24 @@ func removeExpectedAnnotations(existing, ignore map[string]string) map[string]st
 func handleNodeSelectors(c *components, install *operatorv1.Installation) error {
 	// check calico-node nodeSelectors
 	if c.node.Spec.Template.Spec.Affinity != nil {
-		return ErrIncompatibleCluster{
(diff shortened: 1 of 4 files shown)


tmjd (approved):
LGTM

PR #1050, opened 2021-01-05
Draft: CRD changes for Certificate Management

rene-dekker:


api/v1/certificatemanagement_types.go
```diff
@@ -0,0 +1,103 @@
+// Copyright (c) 2020 Tigera, Inc. All rights reserved.
```
tmjd, line 1: its 2021 :grinning:

api/v1/certificatemanagement_types.go
```diff
@@ -0,0 +1,103 @@
...
+
+func init() {
+	SchemeBuilder.Register(&CertificateManagement{}, &CertificateManagementList{})
+}
+
+// Apply this to your cluster only if an automatic certificate signing and approving process is in place for the cluster.
+// Rather than creating self-signed certificates, pods created by this operator will issue a CSR. These pods will not
+// come up until their CSRs are approved.
+type CertificateManagementSpec struct {
+	// Root CA of the certificate authority that signs the certificate requests.
```
tmjd, line 59: Does this comment make sense? Wouldn't it expand to `Root Certificate Authority of the certificate authority that signs the certificate requests?

api/v1/certificatemanagement_types.go
```diff
@@ -0,0 +1,103 @@
...
+	RootCA string `json:"rootCA"`
```
tmjd, line 60: Is this the root cert (possible cert chain) from the CA? It feels a little odd to me that this needs to be provided and doesn't come with the signed certificate. But maybe this is so the signed cert can be verified it came from the right source?
I'm not suggesting change but mentioning it seems odd to me. Though I certainly am not super knowledgeable about cert/signing.

PR #1057, opened 2021-01-07
Support Elasticsearch native users

Suraiya-Hameed:
## Description

- Create a new ConfigMap `tigera-known-oidc-users` and a Secrete `tigera-known-oidc-users-credentials`, if external idP is configured and elasticsearch uses basic license. This ConfigMap will be used by tigera-manager to add end-user information needed to create elasticsearch native users.
- Delete ConfigMap `tigera-known-oidc-users` and the Secrete `tigera-known-oidc-users-credentials`, if external idP is not configured or elasticsearch is not in basic license
- Add permissions to manager and kube-controller to access the new ConfigMap and Secret

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/logstorage/logstorage_controller.go
+++ b/pkg/controller/logstorage/logstorage_controller.go
@@ -192,6 +192,14 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 		return fmt.Errorf("log-storage-controller failed to watch primary resource: %w", err)
 	}
 
+	if err = utils.AddConfigMapWatch(c, render.OIDCUsersConfigMapName, render.ElasticsearchNamespace); err != nil {
+		return fmt.Errorf("log-storage-controller failed to watch the ConfigMap resource: %w", err)
+	}
+
+	if err = utils.AddSecretsWatch(c, render.OIDCUsersSecreteName, render.ElasticsearchNamespace); err != nil {
+		return fmt.Errorf("log-storage-controller failed to watch the Secret resource: %w", err)
+	}
+
 	return nil
 }
 
@@ -355,7 +363,9 @@ func (r *ReconcileLogStorage) Reconcile(request reconcile.Request) (reconcile.Re
 
 	var elasticsearchSecrets, kibanaSecrets, curatorSecrets []*corev1.Secret
 	var clusterConfig *render.ElasticsearchClusterConfig
-	var elasticLicenseType render.ElasticLicenseType
+	var esLicenseType render.ElasticsearchLicenseType
+	var oidcUserConfigMap *corev1.ConfigMap
+	var oidcUserSecret *corev1.Secret
 	applyTrial := false
 
 	if managementClusterConnection == nil {
@@ -401,7 +411,7 @@ func (r *ReconcileLogStorage) Reconcile(request reconcile.Request) (reconcile.Re
 			return reconcile.Result{}, err
 		}
 
-		if elasticLicenseType, err = utils.GetElasticLicenseType(ctx, r.client, reqLogger); err != nil {
+		if esLicenseType, err = utils.GetElasticLicenseType(ctx, r.client, reqLogger); err != nil {
 			// If ECKLicenseConfigMapName is not found, it means ECK operator is not running yet, log the information and proceed
 			if errors.IsNotFound(err) {
 				log.Info("%s ConfigMap not found yet", render.ECKLicenseConfigMapName)
@@ -409,6 +419,16 @@ func (r *ReconcileLogStorage) Reconcile(request reconcile.Request) (reconcile.Re
 				r.status.SetDegraded("Failed to get elastic license", err.Error())
 				return reconcile.Result{}, err
 			}
+		}
+
+		if oidcUserConfigMap, err = r.getOIDCUserConfigMap(ctx); err != nil {
+			r.status.SetDegraded("Failed to read oid user configmap", err.Error())
+			return reconcile.Result{}, err
+		}
+
+		if oidcUserSecret, err = r.oidcUserSecret(ctx); err != nil {
+			r.status.SetDegraded("Failed to read oid user secret", err.Error())
+			return reconcile.Result{}, err
 		}
 	}
 
@@ -483,7 +503,9 @@ func (r *ReconcileLogStorage) Reconcile(request reconcile.Request) (reconcile.Re
 		r.clusterDomain,
 		applyTrial,
 		dexCfg,
-		elasticLicenseType,
+		esLicenseType,
+		oidcUserConfigMap,
+		oidcUserSecret,
 	)
 
 	if err := hdler.CreateOrUpdate(ctx, component, r.status); err != nil {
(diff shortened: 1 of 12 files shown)


pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -652,6 +658,45 @@ func (r *ReconcileLogStorage) getKibanaService(ctx context.Context) (*corev1.Ser
...
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      render.OIDCUsersSecreteName,
+				Namespace: render.ElasticsearchNamespace,
+			},
+		},
+	}
+
+	cmErr := r.client.Get(ctx, types.NamespacedName{Name: render.OIDCUsersConfigMapName, Namespace: render.ElasticsearchNamespace}, &corev1.ConfigMap{})
+	secretErr := r.client.Get(ctx, types.NamespacedName{Name: render.OIDCUsersSecreteName, Namespace: render.ElasticsearchNamespace}, &corev1.Secret{})
+	if (cmErr != nil || secretErr != nil) && addUsersConfigMap {
```
Brian-McM, line 687: The render package allows you to return objects that need to be either created or deleted, this isn't the way to remove items, the second array in the declaration of `Objects() ([]runtime.Object, []runtime.Object)` is for object to delete.

The correct pattern for this (correct as in how the operator team decided to do this) would be look up the config map and secret and pass them to the render code. The render code can then decide whether or not to create the object or delete it. If these object don't exist then you would be passing nil to the render code, in which case the render code can take that as a sign to initialise those config maps.

Brian-McM (changes requested):

pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -465,6 +465,12 @@ func (r *ReconcileLogStorage) Reconcile(request reconcile.Request) (reconcile.Re
 		dexCfg = render.NewDexRelyingPartyConfig(authentication, dexTLSSecret, dexSecret, r.clusterDomain)
 	}
 
+	var oidcUserConfigMapAvailable, oidcUserSecretAvailable bool
+	if managementClusterConnection == nil {
+		oidcUserConfigMapAvailable = r.oidcUserConfigMapAvailable(ctx)
```
Brian-McM, line 470: Could you please follow the pattern that the other components use and pass the component to render and let it decide what to do if the component exists or not.

For instance, the config map might exist but in future iterations there may be upgrades that we have to make to the config map or changes to make, like adding annotations or labels or completely restructuring the resource.

The general pattern has been to grab a resource, if it exists re render it anyway because you may have updates to it or somebody may have changed it outside of the operator and we want to reconcile it. With the config map and secret we do want the Data edited, but the operator can make sure that config map a secret meta data that should not be changed is always accurate.

Also to note, if the data is changed while the operator is reconciling the request to update the config map will fail and be re attempted later because the revision version will be different, so this is a safe pattern.

pkg/render/kube-controllers.go
```diff
@@ -197,6 +197,18 @@ func (c *kubeControllersComponent) controllersRole() *rbacv1.ClusterRole {
 				Resources: []string{"clusterroles", "clusterrolebindings"},
 				Verbs:     []string{"watch", "list", "get"},
 			},
+			{
+				APIGroups:     []string{""},
```
Brian-McM, line 201: After chatting about this with @doublek we may need to make sure that we're restricting access appropriate, so that kube controllers can only access these resources in the tigera-elasticsearch namespace.

I think you're going to need to create a regular role in the tigera-elasticsearch namespace and bind kube controllers to that. Please give that a though and start a conversation with @doublek and I.

pkg/render/logstorage.go
```diff
@@ -323,6 +331,21 @@ func (es *elasticsearchComponent) Objects() ([]runtime.Object, []runtime.Object)
 			toCreate = append(toCreate, es.elasticEnterpriseTrial())
 		}
 
+		// OIDCUsersConfigMapName and OIDCUsersSecreteName should be created only once
```
Brian-McM, line 334: I added comments in the controller about this, we should be passing the config map and secrets to render to be re rendered so that in the future if there's changes we need to make to any meta data it will be easily applied, and there's no worry about overriding the config map data.

Brian-McM (changes requested):

pkg/render/kube-controllers.go
Suraiya-Hameed: yes, creating a roles in tigera-elasticsearch namespace and binding it to manager and kube-controller is definitely better option than clusterrole.

pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -465,6 +465,13 @@ func (r *ReconcileLogStorage) Reconcile(request reconcile.Request) (reconcile.Re
 		dexCfg = render.NewDexRelyingPartyConfig(authentication, dexTLSSecret, dexSecret, r.clusterDomain)
 	}
 
+	var oidcUserConfigMap *corev1.ConfigMap
+	var oidcUserSecret *corev1.Secret
+	if managementClusterConnection == nil {
+		oidcUserConfigMap = r.getOIDCUserConfigMapAvailable(ctx)
```
tmjd, line 471: Is there a watch for this resource?

pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -465,6 +465,13 @@ func (r *ReconcileLogStorage) Reconcile(request reconcile.Request) (reconcile.Re
...
+		oidcUserSecret = r.oidcUserSecretAvailable(ctx)
```
tmjd, line 472: Same here.

pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -652,6 +661,34 @@ func (r *ReconcileLogStorage) getKibanaService(ctx context.Context) (*corev1.Ser
 	return &svc, nil
 }
 
+func (r *ReconcileLogStorage) getOIDCUserConfigMapAvailable(ctx context.Context) *corev1.ConfigMap {
```
tmjd, line 664: I think the name should just be `getOIDCUserConfigMap`.
Also I think it should return an error. We need to be able to tell if the error was because the resource doesn't exist or there was an error with the query.

pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -652,6 +661,34 @@ func (r *ReconcileLogStorage) getKibanaService(ctx context.Context) (*corev1.Ser
...
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      render.OIDCUsersConfigMapName,
+				Namespace: render.ElasticsearchNamespace,
+			},
+		}
+	}
+	return cm
+}
+
+func (r *ReconcileLogStorage) oidcUserSecretAvailable(ctx context.Context) *corev1.Secret {
```
tmjd, line 678: Same comments on this function.

tmjd (changes requested):

pkg/controller/logstorage/logstorage_controller.go
Suraiya-Hameed: There is no watch on this ConfigMap or Secret, they should be created or deleted depending on es license type and dex config, we already have watch on them, also data in this ConfigMap & Secret will be updated by other components (es-proxy and kube-controller) so we don't have to watch for changes.

tmjd: I think we still need watches for them.
Unless it is guaranteed that the ConfigMap and Secret will be updated before the license type and dex config are updated. If not then it would be possible that the controller gets an update (from the watch) for ES license type and dex config, logstorage controller Reconciles, then the ConfigMap and Secret are updated. Logstorage would have missed the updates to the ConfigMap and Secret.

tmjd (approved):
LGTM

marvin-tigera:

Brian-McM (approved):

PR #1062, opened 2021-01-11
Add fluentd component for Windows

lmm:
## Description

This PR adds support for rendering the fluentd component for Windows. It uses the `tigera/fluentd-windows` multi-arch image (soon to be a thing).

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


pkg/controller/logcollector/logcollector_controller.go
```diff
@@ -369,13 +373,40 @@ func (r *ReconcileLogCollector) Reconcile(request reconcile.Request) (reconcile.
...
 		return reconcile.Result{}, err
 	}
 
+	// Render a fluentd component for Windows if the cluster has Windows nodes.
+	hasWindowsNodes, err := hasWindowsNodes(r.client)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	if hasWindowsNodes {
```
caseydavenport, line 390: Any reason to not just always render a Windows daemonset? If there are no windows nodes it just won't run pods anyway, right?

I'm thinking it would be easier than this approach, because for this to work we need to monitor nodes to see if windows nodes are ever added to the cluster after initial provisioning.

caseydavenport: Node updates happen frequently, and so will cause unnecessary reconciliation in a lot of scenarios. Probably not an terrible issue, but certainly not ideal.

lmm: Hmm... I was thinking it might be weird to have the daemonset even running on windows. That is a good point about the node update frequency that I didn't consider though.

pkg/render/fluentd.go
```diff
@@ -154,7 +220,10 @@ func (c *fluentdComponent) Objects() ([]runtime.Object, []runtime.Object) {
 			c.eksLogForwarderSecret(),
 			c.eksLogForwarderDeployment())
 	}
-	if c.installation.KubernetesProvider != operatorv1.ProviderOpenShift {
+
+	// Windows does not support allowedHostPaths yet.
```
mgleung, line 224: I think this comment is supposed to be about PSPs

pkg/render/fluentd.go
```diff
@@ -121,10 +131,64 @@ type fluentdComponent struct {
...
+}
+
+func (c *fluentdComponent) fluentdNodeName() string {
+	if c.osType == OSTypeWindows {
+		return fluentdNodeWindowsName
+	}
+	return fluentdNodeName
+}
+
+func (c *fluentdComponent) eksLogForwarderName() string {
```
mgleung, line 155: we don't deploy this for windows right? Do we need to do the name handling for this in that case?

pkg/controller/logcollector/logcollector_controller.go
caseydavenport: it wouldn't run on non-windows nodes though, right? We just slap a node selector on each so that the linux daemonset only runs on linux nodes, and the windows daemonset only runs on windows nodes (if any exist).

If the cluster has no windows nodes, then the windows daemonset pod won't run at all.

pkg/render/fluentd.go
lmm: Added a clarification

pkg/render/fluentd.go
lmm: The log forwarder should just work on Windows too (the eks forwarder binary is built for windows and added to the fluentd image)

pkg/render/fluentd.go
```diff
@@ -144,7 +208,9 @@ func (c *fluentdComponent) Objects() ([]runtime.Object, []runtime.Object) {
 		objs = append(objs, c.filtersConfigMap())
 	}
 	if c.eksConfig != nil {
-		if c.installation.KubernetesProvider != operatorv1.ProviderOpenShift {
+		// Windows PSP does not support allowedHostPaths yet.
```
mgleung, line 211: Sorry, this one isn't just PSP right? This is for EKS specifically or am I misunderstanding?

pkg/controller/logcollector/logcollector_controller.go
lmm: I should have been more specific and said `it might be weird to have the daemonset even running at all`.
I did try this and I got a status degraded which makes sense:
```
  - lastTransitionTime: "2021-01-12T02:38:35Z"
    message: DaemonSet "tigera-fluentd/fluentd-node-windows" is not yet scheduled
      on any nodes
    reason: Not all pods are ready
```

I poked around but didn't see a good place to workaround. I looked at status manager but it only keeps track of namespaced names.

pkg/render/fluentd.go
mgleung: nvm I confused myself.

mgleung (approved):
LGTM

pkg/controller/logcollector/logcollector_controller.go
lmm: Casey, we might merge this for now, just to help unblock some testing. I can address any follow-up/review you have afterwards.

pkg/render/fluentd.go
```diff
@@ -182,6 +187,20 @@ func (c *fluentdComponent) livenessCmd() []string {
 	return []string{"sh", "-c", "/bin/liveness.sh"}
 }
 
+func (c *fluentdComponent) probeTimeout() int32 {
+	if c.osType == OSTypeWindows {
+		return ProbeTimeoutSeconds
```
song-jiang, line 192: This line seems to be wrong.

song-jiang, line 192:
```suggestion
		return ProbeTimeoutSecondsWindows
```

pkg/render/fluentd.go
```diff
@@ -121,10 +136,78 @@ type fluentdComponent struct {
...
+		return []string{`c:\ruby26\msys64\usr\bin\bash.exe`, `-lc`, `/c/bin/liveness.sh`}
+	}
+	return []string{"sh", "-c", "/bin/liveness.sh"}
+}
+
+func (c *fluentdComponent) probeTimeout() int32 {
+	if c.osType == OSTypeWindows {
+		return ProbeTimeoutSeconds
+	}
+	return ProbeTimeoutSecondsWindows
```
song-jiang, line 194:
```suggestion
	return ProbeTimeoutSeconds
```

song-jiang (approved):
LGTM

song-jiang:
I'm going to merge this PR. @lmm Please follow up any updates required. Thanks.

pkg/controller/logcollector/logcollector_controller.go
```diff
@@ -100,6 +100,10 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 		}
 	}
 
+	err = c.Watch(&source.Kind{Type: &corev1.Node{}}, &handler.EnqueueRequestForObject{})
```
tmjd, line 103: @lmm I was just noticing some reconciles happening because of (I assume) node updates so I got looking at why we have the watch (leading me here) and I think the watch could be limited to Node Creation and Deletion. It isn't like a node's OS type can change. WDYT?

PR #1064, opened 2021-01-12
Imageset

tmjd:
## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


pkg/controller/utils/imageset.go
```diff
@@ -0,0 +1,132 @@
...
+func AddImageSetWatch(c controller.Controller) error {
+	return c.Watch(&source.Kind{Type: &operator.ImageSet{}}, &handler.EnqueueRequestForObject{})
+}
+
+func getSetName(v operator.ProductVariant) string {
+	setName := fmt.Sprintf("calico-%s", components.CalicoRelease)
+	if v == operator.TigeraSecureEnterprise {
+		setName = fmt.Sprintf("enterprise-%s", components.EnterpriseRelease)
+	}
+	return setName
```
ozdanborne, line 69: nit, but i couldn't help myself
```suggestion
	if v == operator.TigeraSecureEnterprise {
		return fmt.Sprintf("enterprise-%s", components.EnterpriseRelease)
	}
	return fmt.Sprintf("calico-%s", components.CalicoRelease)
```

pkg/controller/utils/imageset.go
```diff
@@ -0,0 +1,132 @@
...
+}
+
+// getImageSet finds the ImageSet CR for specified variant and for the correct .
```
ozdanborne, line 72:
```suggestion
// getImageSet finds the ImageSet for the specified variant.
```

pkg/controller/utils/imageset.go
```diff
@@ -0,0 +1,132 @@
...
+func getImageSet(ctx context.Context, cli client.Client, v operator.ProductVariant) (*operator.ImageSet, error) {
+	isl := &operator.ImageSetList{}
+
+	// List the ImageSets because if any exist then we will require the
+	// existence of one for the expected version of the operator running
+	// and the variant configured.
```
ozdanborne, line 78: Instead of a List, could we do a Get using this expected name?

Or is there some more detailed implication where if any imageset exists, we need to always use imagesets? If so, I adjusted the comment a bit because that wasn't clear to me on first read
```suggestion
	// List the ImageSets to see if any exist. If at least one exists, then require
	// one which corresponds to the current version / variant. This ensures that
	// a user upgrading between versions of Operator won't accidentally move off
	// of ImageSets if they accidentally didn't create an ImageSet with the new expected name.
```

ozdanborne (changes requested):
Mostly nits. I do wish we didn't have to put near-identical looking code in each component's `ResolveImage` function but I failed to come up with a way to do it myself so I guess it's fine as is

config/samples/operator_v1_imageset.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: operator.tigera.io/v1
+kind: ImageSet
+metadata:
+  name: calico-v3.4.0
+spec:
+  images:
+  - image: "calico/node"
+    digest: "sha256:0deadbeef1234567890"
```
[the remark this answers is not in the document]
tmjd: They'll change with each release though. Its probably better to have someone use these dummy ones and fail rather than deploy with some really old images (I'm assuming they won't get updated).
WDYT?

pkg/controller/utils/imageset.go
```diff
@@ -0,0 +1,132 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/source"
+
+	operator "github.com/tigera/operator/api/v1"
+	"github.com/tigera/operator/pkg/components"
+	"github.com/tigera/operator/pkg/render"
+)
+
+// ApplyImageSet gets the appropriate ImageSet and applies it to
+// the components passed in.
+func ApplyImageSet(ctx context.Context, c client.Client, v operator.ProductVariant, comps ...render.Component) error {
```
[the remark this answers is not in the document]
tmjd: Should that be in `pkg/controller/utils/imageset`, `pkg/controller/imageset` or somewhere else?

pkg/controller/utils/imageset.go
tmjd: yeah the list is so that we know if any ImageSets exist. I figured it might as well use what is returned in the list instead of doing a separate Get. I'll include the comment update

pkg/render/fluentd.go
```diff
@@ -137,6 +138,29 @@ type fluentdComponent struct {
 	installation    *operatorv1.InstallationSpec
 	clusterDomain   string
 	osType          OSType
+	linuxImage      string
+	windowsImage    string
```
ozdanborne, line 142: could this just be `image` and then set it to the correct one based on osType inside of ResolveImages?

ozdanborne (approved):
one last comment but lgtm

api/v1/imageset_types.go
```diff
@@ -0,0 +1,76 @@
...
+	Image string `json:"image"`
+
+	// Digest is the image identifier that will be used for the Image.
+	// The field should not include a leading `@` and must be prefixed with `sha256:`, if other
+	// digest prefixes are needed please submit a feature request.
+	Digest string `json:"digest"`
+}
+
+// ImageSetStatus defines the observed state of ImageSet
+type ImageSetStatus struct {
```
caseydavenport, line 44: No status?

pkg/controller/utils/imageset/imageset.go
```diff
@@ -0,0 +1,154 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/source"
+
+	operator "github.com/tigera/operator/api/v1"
+	"github.com/tigera/operator/pkg/components"
+	"github.com/tigera/operator/pkg/render"
+)
+
+// ApplyImageSet gets the appropriate ImageSet, validates the ImageSet, and applies it to
+// the components passed in.
+func ApplyImageSet(ctx context.Context, c client.Client, v operator.ProductVariant, comps ...render.Component) error {
```
caseydavenport, line 34: what does "apply it to the components passed in" mean? Does it mean it's going to modify any images within the components to use new images?

pkg/controller/clusterconnection/clusterconnection_controller.go
```diff
@@ -178,6 +183,12 @@ func (r *ReconcileConnection) Reconcile(request reconcile.Request) (reconcile.Re
 		tunnelSecret,
 	)
 
+	if err = imageset.ApplyImageSet(ctx, r.Client, variant, component); err != nil {
```
caseydavenport, line 186: I'm just skimming, but why did you decide to go with a pattern of modifying the component after creating it rather than passing the image set to the constructor?

api/v1/imageset_types.go
tmjd: What status would we have here? There wasn't anything in the design.
We do have a status on the Installation about which ImageSet is being used.

pkg/controller/clusterconnection/clusterconnection_controller.go
tmjd: I think the main reason was I wanted it to be obvious that a Component needs to have the ResolveImages functionality.
(I wasn't confident on this pattern and is why I put out a draft PR of this last week.)

pkg/controller/utils/imageset/imageset.go
tmjd: I should change it to call ResolveImages on each of the components passed in (because that is what it means).

api/v1/imageset_types.go
caseydavenport: I think it's fine to have no status if there's nothing meaningful to put there, but we can just delete this struct then.

pkg/controller/clusterconnection/clusterconnection_controller.go
caseydavenport: I don't think I really have a problem with it, it just wasn't my first thought. I was imagining that we'd just add an argument to `GetReference` and that would be obvious enough that an imageset is required anywhere that an image is used.

tmjd: Hmmm, my reasoning for not doing that was because of where we were calling GetReference and changing it like that could cause an error and didn't want to change Objects signature. So I wanted that resolution to happen sooner (not in Objects()), but I guess I've changed that behavior already by moving those calls into ResolveImages.

pkg/render/render.go
```diff
@@ -41,6 +41,7 @@ var (
 )
 
 type Component interface {
+	ResolveImages(is *operator.ImageSet) error
```
caseydavenport, line 44: This should have a comment explaining what the expected behavior of this function is.

caseydavenport (approved):

PR #1065, opened 2021-01-12
Respect scopes for Google, change default scopes, don't respect scope…

Brian-McM:
…s for Open Shift

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/dex_config.go
+++ b/pkg/render/dex_config.go
@@ -217,10 +217,10 @@ func (d *dexBaseCfg) ClientSecret() []byte {
 }
 
 func (d *dexBaseCfg) RequestedScopes() []string {
-	if d.connectorType == connectorTypeOIDC && d.authentication.Spec.OIDC.RequestedScopes != nil {
+	if d.authentication.Spec.OIDC.RequestedScopes != nil {
 		return d.authentication.Spec.OIDC.RequestedScopes
 	}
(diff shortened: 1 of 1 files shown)


pkg/render/dex_config.go
```diff
@@ -451,7 +451,7 @@ func (d *dexConfig) Connector() map[string]interface{} {
 		RootCASecretField: rootCASecretLocation,
 	}
 
-	if len(d.RequestedScopes()) > 0 {
+	if d.connectorType != connectorTypeOpenshift && len(d.RequestedScopes()) > 0 {
```
rene-dekker, line 454: I think it would be better to specify when we *do* want the scopes, so we don't make a mistake once we add more connectors, so :
`if type == google | type ==  oidc]`

rene-dekker (approved):
lgtm

pkg/render/dex_config.go
rene-dekker: Actually, the specs other than oidc (& google) do not have the requestedscopes field, so I think you can omit the whole change.

Brian-McM: As we discussed this does need to be there for google as the scope does work, and I'm not sure if this will break for oidc which is why I added it this check.

lmm (approved):

tmjd (approved):
LGTM

PR #1072, opened 2021-01-19
Add Elasticsearch version and Kibana URL to Manager environment variables

Brian-McM:
## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/manager.go
+++ b/pkg/render/manager.go
@@ -485,7 +485,10 @@ func (c *managerComponent) managerEsProxyContainer() corev1.Container {
 
 	env := []v1.EnvVar{
 		{Name: "ELASTIC_LICENSE_TYPE", Value: string(c.esLicenseType)},
+		{Name: "ELASTIC_VERSION", Value: components.ComponentEckElasticsearch.Version},
+		{Name: "ELASTIC_KIBANA_ENDPOINT", Value: fmt.Sprintf(KibanaHTTPSEndpoint, c.clusterDomain)},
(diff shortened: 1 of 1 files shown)


pkg/render/manager.go
```diff
@@ -484,8 +484,12 @@ func (c *managerComponent) managerEsProxyContainer() corev1.Container {
 	}
 
 	env := []v1.EnvVar{
+		{Name: "LOG_LEVEL", Value: "INFO"},
```
tmjd, line 487: Is this needed? I'd expect this to be the default and we should avoid setting env vars if not needed.

Brian-McM: It's not needed but it's intentional. I always forget what the log level env name is "LOGLEVEL vs LOG_LEVEL" and it's nice to be able to tweak the deployment env variable value instead of adding a new line. I can remove it if you don't want it there though.

tmjd: Yeah I think it should be removed. We hope to have a real API for this at some point, ideally not in the operator API, and having this set would conflict/override.
If we don't have this documented we should fix that instead of setting it here.

tmjd (approved):
LGTM

pkg/render/manager.go
Brian-McM: I've removed this, but I don't know if this is something we should document, because there's no way for a user to set this with using the unsupported tag.

PR #1080, opened 2021-01-22
recreate ES/kibana certs when DNS names change

lmm:
## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/logstorage/logstorage_controller.go
+++ b/pkg/controller/logstorage/logstorage_controller.go
@@ -574,25 +574,25 @@ func (r *ReconcileLogStorage) elasticsearchSecrets(ctx context.Context) ([]*core
 	var secrets []*corev1.Secret
 	svcDNSNames := dns.GetServiceDNSNames(render.ElasticsearchServiceName, render.ElasticsearchNamespace, r.clusterDomain)
 
-	// Get the key secret - might be nil
-	keySecret, err := utils.GetCertificateSecret(ctx, r.client, render.TigeraElasticsearchCertSecret, render.OperatorNamespace())
+	// Get the secret - might be nil
+	secret, err := utils.GetCertificateSecret(ctx, r.client, render.TigeraElasticsearchCertSecret, render.OperatorNamespace())
 	if err != nil {
 		return nil, err
 	}
+
+	// Ensure that secret is valid.
+	secret, err = utils.EnsureCertificateKeySecret(ctx, render.TigeraElasticsearchCertSecret, secret, svcDNSNames...)
+	if err != nil {
+		return nil, err
+	}
+
+	secrets = append(secrets, secret, render.CopySecrets(render.ElasticsearchNamespace, secret)[0])
 
 	// Get the pub secret - might be nil
 	pubSecret, err := utils.GetCertificateSecret(ctx, r.client, render.ElasticsearchPublicCertSecret, render.ElasticsearchNamespace)
 	if err != nil {
 		return nil, err
 	}
-
-	// Ensure that key secret is valid.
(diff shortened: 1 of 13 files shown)


pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -570,29 +573,78 @@ func (r *ReconcileLogStorage) Reconcile(request reconcile.Request) (reconcile.Re
...
-	if err := r.client.Get(ctx, types.NamespacedName{Name: render.TigeraElasticsearchCertSecret, Namespace: render.OperatorNamespace()}, secret); err != nil {
-		if errors.IsNotFound(err) {
-			svcDNSNames := dns.GetServiceDNSNames(render.ElasticsearchServiceName, render.ElasticsearchNamespace, r.clusterDomain)
-			secret, err = render.CreateOperatorTLSSecret(nil,
-				render.TigeraElasticsearchCertSecret, "tls.key", "tls.crt",
-				render.DefaultCertificateDuration, nil, svcDNSNames...,
-			)
-		} else {
+
+	createCert := func() (*corev1.Secret, error) {
```
Brian-McM, line 579: I don't think we should use an anonymous function here, if it's useful to do this here it'll be useful else where and should be a function that can be accessed by all.

pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -570,29 +573,78 @@ func (r *ReconcileLogStorage) Reconcile(request reconcile.Request) (reconcile.Re
 	return reconcile.Result{}, nil
 }
 
-func (r *ReconcileLogStorage) elasticsearchSecrets(ctx context.Context) ([]*corev1.Secret, error) {
-	var secrets []*corev1.Secret
+func (r *ReconcileLogStorage) getCertificateKeySecret(ctx context.Context, keySecretName string, keySecretNs string, pubSecretName string, pubSecretNs string, svcDNSNames ...string) (*corev1.Secret, error) {
```
Brian-McM, line 576: I don't think this function belongs to the logstorage reconciler, could you please put this in a suitable common location.

Brian-McM (changes requested):

pkg/controller/utils/elasticsearch.go
```diff
@@ -20,8 +20,10 @@ import (
 	"crypto/tls"
```
Brian-McM, line 20: I guess this is a bit nitpicky, but I don't think this has anything to do with elasticsearch so I don't think it should be in this file.

pkg/controller/utils/elasticsearch.go
```diff
@@ -373,3 +376,75 @@ func getTotalEsDisk(ls *operatorv1.LogStorage) int64 {
...
+	pemBlock, _ := pem.Decode(certBytes)
+	cert, err := x509.ParseCertificate(pemBlock.Bytes)
+	if err != nil {
+		return false, err
+	}
+
+	return reflect.DeepEqual(cert.DNSNames, expectedDNSNames), nil
+}
+
+func GetCertificateSecret(ctx context.Context, client client.Client, name string, ns string) (*corev1.Secret, error) {
```
Brian-McM, line 441: This doesn't really have anything to do with specifically getting a secret that contains a certificate, it's just shortcutting the code to get a secret in general.

lmm: Moved to its own file

pkg/controller/utils/elasticsearch.go
lmm: Makes sense. Moved to a new file

doublek (approved):
LGTM

lmm:
After discussing with @doublek , I'll add a follow-up PR tomorrow

PR #1082, opened 2021-01-22
LDAP support for enterpise w/ Authentication CRD

rene-dekker:


pkg/render/dex.go
```diff
@@ -175,8 +174,8 @@ func (c *dexComponent) deployment() runtime.Object {
 					ImagePullSecrets:   getImagePullSecretReferenceList(c.pullSecrets),
 					Containers: []corev1.Container{
 						{
-							Name:            DexObjectName,
-							Image:           components.GetReference(components.ComponentDex, c.installation.Registry, c.installation.ImagePath),
+							Name:  DexObjectName,
+							Image: "gcr.io/tigera-dev/cnx/tigera/dex:rene", ImagePullPolicy: "Always", //components.GetReference(components.ComponentDex, c.installation.Registry, c.installation.ImagePath),
```
rene-dekker, line 178: reminder: revert

api/v1/authentication_types.go
```diff
@@ -105,6 +109,71 @@ type AuthenticationOpenshift struct {
...
+	// Following list contains field pairs that are used to match a user to a group. It adds an additional
+	// requirement to the filter that an attribute in the group must match the user's
+	// attribute value.
+	// +required
+	UserMatchers []UserMatch `json:"userMatchers"`
+}
+
+// UserMatch when the value of a UserAttribute and a GroupAttribute match, a user belongs to the group.
+type UserMatch struct {
+	UserAttribute  string `json:"userAttribute"`
```
tmjd, line 173: I'd like to see required or optional for this and GroupAttribute.

pkg/controller/authentication/authentication_controller.go
```diff
@@ -330,19 +339,27 @@ func updateAuthenticationWithDefaults(authentication *oprv1.Authentication) {
...
 
 // validateAuthentication makes sure that the authentication spec is ready for use.
 func validateAuthentication(authentication *oprv1.Authentication) error {
-	// We support using only one connector at once.
-	if authentication.Spec.OIDC != nil && authentication.Spec.Openshift != nil {
-		return fmt.Errorf("multiple identity provider connectors were specified, but only 1 is allowed in the Authentication spec")
-	} else if authentication.Spec.OIDC == nil && authentication.Spec.Openshift == nil {
+	oidc := authentication.Spec.OIDC != nil
+	ldp := authentication.Spec.LDAP
+	numConnectors := render.BoolToInt(oidc) + render.BoolToInt(authentication.Spec.Openshift != nil) + render.BoolToInt(ldp != nil)
```
tmjd, line 354: I'm not sure I like this for a couple reasons.
1. oidc is a bool, ldp is a reference, and Openshift doesn't have a short form.
2. The BoolToInt function feels odd as a way to count the "true"s, would this better be served by a more direct function of `func countTrues(b ...bool) int`.

pkg/controller/authentication/authentication_controller_test.go
```diff
@@ -12,24 +12,25 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package authentication
+package authentication_test
```
tmjd, line 15: Do you feel it is important to switch this to *_test?
I probably recently changed this to without the _test because I think testing of atleast one of the other controllers needed access to internals and so I was trying to make them all consistent to the _test omitted form.
Is there a downside to the non-_test form that we should avoid it?

tmjd:
I've changed the milestone for this to v1.16 because for this to be in v1.14.1 it will need to be picked into the release-v1.14 branch and that Pick should get the v1.14.1 milestone.

pkg/controller/authentication/authentication_controller_test.go
rene-dekker: There is a convenience method for these internals. I guessed you did not know when you added that test and I refactored that. I'm good either way.

rene-dekker: I'll revert.

Brian-McM: So I'm not completely opposed to having the same package for the tests, but the downside is specifically that you are allowed to access the internals, and you lose the black box testing. The best unit tests usually only use interfaces that are exposed outside of the package, mimicing how the package would actually be used.

pkg/render/dex_config.go
```diff
@@ -28,6 +28,7 @@ const (
 	connectorTypeOIDC      = "oidc"
```
Brian-McM, line 28: Could you please add some tests for these changes?

Brian-McM (changes requested):

pkg/controller/authentication/authentication_controller.go
Brian-McM: Oh I commented on how I thought it was weird to have a  "countTrues" function exported from the utilites, I think "countConnectors" in this function or just having that code inline would be better.

Brian-McM (approved):

tmjd (approved):
LGTM

PR #1085, opened 2021-01-23
Updating controller-runtime

tmjd:
## Description

Some CI tests hit cases where the operator does not see updates to the LicenseKey but getting the LicenseKey with kubectl shows that the resource exists. This seems to be an issue with the client cache that the controller-runtime sets up. I even verified this by disabling the cache completely and no longer hit this problem. I also found that the controller-runtime had an update with a new feature for the cache where the cache can be disabled for some resources.
This PR updates the controller-runtime version and disables the cache for LicenseKey resources.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


main.go
```diff
@@ -145,6 +146,11 @@ func main() {
 		Port:               9443,
 		LeaderElection:     enableLeaderElection,
 		LeaderElectionID:   "operator-lock",
+		// We should test this again in the future to see if the problem with LicenseKey updates
```
caseydavenport, line 149: Is there a link to a GitHub issue that we can include here? This comment is not going to be useful in 6 months time without a reference.

tmjd: I added more to the comment and also a link to the controller-runtime issue.

caseydavenport (approved):

PR #1091, opened 2021-01-25
Tigeracustom Disable ES

tmjd:
## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/common/common.go
+++ b/pkg/common/common.go
@@ -27,7 +27,7 @@ const (
 )
 
 type Installation struct {
-	Spec                *operator.InstallationSpec
-	TigeraCustom        bool
-	NodeAppArmorProfile string
+	Spec                    *operator.InstallationSpec
+	CalicoCloudControlPlane bool
+	NodeAppArmorProfile     string
 }
--- a/pkg/common/common.go
+++ b/pkg/common/common.go
@@ -14,6 +14,10 @@
 
 package common
 
+import (
+	operator "github.com/tigera/operator/api/v1"
+)
+
 const (
 	CalicoNamespace     = "calico-system"
 	TyphaDeploymentName = "calico-typha"
@@ -21,3 +25,9 @@ const (
 
 	TigeraPrometheusNamespace = "tigera-prometheus"
 )
+
+type InstallationInternal struct {
+	Spec                *operator.InstallationSpec
+	TigeraCustom        bool
+	NodeAppArmorProfile string
+}
--- a/pkg/common/common.go
+++ b/pkg/common/common.go
@@ -26,7 +26,7 @@ const (
 	TigeraPrometheusNamespace = "tigera-prometheus"
 )
 
-type InstallationInternal struct {
+type Installation struct {
 	Spec                *operator.InstallationSpec
 	TigeraCustom        bool
 	NodeAppArmorProfile string
--- a/pkg/controller/amazoncloudintegration/amazoncloudintegration_controller.go
+++ b/pkg/controller/amazoncloudintegration/amazoncloudintegration_controller.go
@@ -168,7 +168,7 @@ func (r *ReconcileAmazonCloudIntegration) Reconcile(ctx context.Context, request
 		return reconcile.Result{}, nil
 	}
 
-	pullSecrets, err := utils.GetNetworkingPullSecrets(network, r.client)
+	pullSecrets, err := utils.GetNetworkingPullSecrets(network.Spec, r.client)
 	if err != nil {
 		r.SetDegraded("Error retrieving pull secrets", err, reqLogger)
 		return reconcile.Result{}, err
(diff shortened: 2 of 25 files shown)


pkg/controller/amazoncloudintegration/amazoncloudintegration_controller.go
```diff
@@ -187,7 +187,7 @@ func (r *ReconcileAmazonCloudIntegration) Reconcile(request reconcile.Request) (
 
 	// Render the desired objects from the CRD and create or update them.
 	reqLogger.V(3).Info("rendering components")
-	component, err := render.AmazonCloudIntegration(instance, network, awsCredential, pullSecrets, r.provider == operatorv1.ProviderOpenShift)
+	component, err := render.AmazonCloudIntegration(instance, network.Spec, awsCredential, pullSecrets, r.provider == operatorv1.ProviderOpenShift)
```
ozdanborne, line 190: Seems like you've created a new type `InstallationInternal` which allows you to pass both the spec and the additional annotation options via a single struct. I like this idea, but I think it belongs in the render package as a `RenderAmazonCloudOptions` struct. And we could even move all these other arguments into it too, over time. I've always thought this could work well for us since if you look in our tests, we usually pass the default empty values for most of these anyways (false, nil, "", etc.).

Alternatively, we could follow the existing pattern by just passing those additional options (tigeracustom / nodeapparmor) as additional args to the render functions?

ozdanborne (changes requested):

pkg/render/fluentd.go
```diff
@@ -598,6 +602,21 @@ func (c *fluentdComponent) envvars() []corev1.EnvVar {
 		}
 	}
 
+	// For the TigeraCustom flag we disable log forwarding to Elasticsearch.
+	// This means the logs for this cluster are not stored anywhere.
+	if c.installation.TigeraCustom {
+		// Disable only for management clusters
+		if c.managementCluster != nil {
```
ozdanborne, line 609: nit?
```suggestion
	// Disable only for management clusters
	if c.installation.TigeraCustom && c.managementCluster != nil {	
```

pkg/render/fluentd.go
```diff
@@ -126,23 +129,24 @@ type EksCloudwatchLogConfig struct {
...
-	splkCredential  *SplunkCredential
-	filters         *FluentdFilters
-	eksConfig       *EksCloudwatchLogConfig
-	pullSecrets     []*corev1.Secret
-	installation    *operatorv1.InstallationSpec
-	clusterDomain   string
-	osType          OSType
-	image           string
+	lc                *operatorv1.LogCollector
+	managementCluster *operatorv1.ManagementCluster
```
ozdanborne, line 133: Is this value used for anything more than triggering static code? 
```suggestion
	isManagementCluster bool
```

pkg/common/common.go
```diff
@@ -14,10 +14,20 @@
...
 const (
 	CalicoNamespace     = "calico-system"
 	TyphaDeploymentName = "calico-typha"
 	NodeDaemonSetName   = "calico-node"
 
 	TigeraPrometheusNamespace = "tigera-prometheus"
 )
+
+type InstallationInternal struct {
+	Spec                *operator.InstallationSpec
```
ozdanborne, line 30: One last idea - would it make sense to use struct inheritence so we don't have to pass `.Spec` everywhere?
```suggestion
	*operator.InstallationSpec
```

ozdanborne (approved):
couple of nits bug lgtm

pkg/controller/amazoncloudintegration/amazoncloudintegration_controller.go
tmjd: We chatted about this in slack and I think we're good with the InstallationInternal for now.

pkg/common/common.go
tmjd: I could make this change but for all the render calls that take an InstallationSpec we would need to pass the embedded struct so embedding it would not make a difference.
For example
```
x := common.InstallationInternal

// Render call taking operator.InstallationSpec
render.Example(x.InstallationSpec)
```

caseydavenport: Is the idea here to wrap an installation spec with additional fields?

Aren't `TigeraCustom` and `NodeAppArmorProfile` derivable from the "real" Installation API? I'm not sure I understand the purpose of this structure.

pkg/controller/installation/core_controller.go
```diff
@@ -244,10 +245,12 @@ type ReconcileInstallation struct {
 	clusterDomain        string
 }
 
+const TigeraCustom = "unsupported.operator.tigera.io/custom"
```
caseydavenport, line 248: This annotation needs a comment explaining what it's purpose is. "custom" is very vague, and doesn't make it obvious what setting this annotation does.

Ideally, the name of the annotation would describe it's purpose.

pkg/common/common.go
tmjd: We stopped passing the full Installation resource around because really renders should (ideally) only need the Spec. Also with the overlay functionality we only needed to merge the InstallationSpec so it made since to stop passing the full Installation around.

caseydavenport: Right, but this PR is re-introducing the need to pass more than just the spec to render.

It feels wrong to then say "Render only needs spec, but there's also this other field that we want to pass, so we'll create a new resource that just tacks on the new field to the Installation spec".

I'd rather remove this structure altogether, and then either:
- Pass the whole Installation again, because that is what is needed.
- Extend the render functions to take an additional argument wherever needed, passing in the relevant fields explicitly.

caseydavenport (approved):
LGTM. Hopefully we can pull this out in the near future.

PR #1110, opened 2021-01-28
Use 'skipInterface: ^br-.*' for Docker Enterprise ipv4 autodetection …

Brian-McM:
…method

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/installation/core_controller.go
+++ b/pkg/controller/installation/core_controller.go
@@ -455,10 +455,18 @@ func fillDefaults(instance *operator.Installation) error {
 				v4pool.NodeSelector = operator.NodeSelectorDefault
 			}
 			if instance.Spec.CalicoNetwork.NodeAddressAutodetectionV4 == nil {
-				// Default IPv4 address detection to "first found" if not specified.
-				t := true
-				instance.Spec.CalicoNetwork.NodeAddressAutodetectionV4 = &operator.NodeAddressAutodetection{
-					FirstFound: &t,
+				if instance.Spec.KubernetesProvider == operator.ProviderDockerEE {
+					// firstFound finds the Docker Enterprise interface prefixed with br-, which is unusable for the
+					// node address, so instead skip the interface br-.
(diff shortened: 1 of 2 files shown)


pkg/controller/installation/core_controller.go
```diff
@@ -455,10 +455,18 @@ func fillDefaults(instance *operator.Installation) error {
...
 			if instance.Spec.CalicoNetwork.NodeAddressAutodetectionV4 == nil {
-				// Default IPv4 address detection to "first found" if not specified.
-				t := true
-				instance.Spec.CalicoNetwork.NodeAddressAutodetectionV4 = &operator.NodeAddressAutodetection{
-					FirstFound: &t,
+				if instance.Spec.KubernetesProvider == operator.ProviderDockerEE {
+					// firstFound finds the Docker Enterprise interface prefixed with br-, which is unusable for the
+					// node address, so instead skip the interface br-.
+					instance.Spec.CalicoNetwork.NodeAddressAutodetectionV4 = &operator.NodeAddressAutodetection{
+						SkipInterface: "^br-.*",
```
caseydavenport, line 462: I worry that we also need to skip a lot of other interfaces as well.

I think we still need a regex that also matches these: https://github.com/projectcalico/node/blob/904c449d84f3d9e0bcfde849f3076869e1a2a4ec/pkg/startup/startup_linux.go#L29-L33

WDYT? I'm not 100% up to speed on Docker enterprise - will it ever have any of those other interfaces on it as well?

Brian-McM: That's a really good point.... I'll double check if it's needed and add it if it is.

Brian-McM: @caseydavenport https://github.com/projectcalico/node/blob/master/pkg/startup/startup.go#L815-L828, looks like we still use those interfaces to skip.

matthewdupre (approved):

pkg/controller/installation/core_controller.go
caseydavenport: Awesome, :shipit:

PR #1115, opened 2021-01-29
Until the cert initialization is available for OS treat it as private

tmjd:
## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/components/enterprise.go
+++ b/pkg/components/enterprise.go
@@ -152,6 +152,11 @@ var (
 		Image:   "tigera/cloud-controllers",
 	}
 
+	ComponentCSRInitContainer = component{
+		Version: "release-calient-v3.5",
+		Image:   "tigera/key-cert-provisioner",
+	}
+
 	EnterpriseComponents = []component{
 		ComponentAPIServer,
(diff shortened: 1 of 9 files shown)


rene-dekker (approved):
lgtm

pkg/components/enterprise.go
```diff
@@ -180,5 +185,6 @@ var (
 		ComponentTigeraTypha,
 		ComponentTigeraCNI,
 		ComponentCloudControllers,
+		ComponentCSRInitContainer,
```
rene-dekker, line 188: Won't this make it use the wrong registry?

doublek (approved):
LGTM

pkg/components/enterprise.go
tmjd: This doesn't have any effect on the registry used. In components/references.go is what controls which registry it comes from. This PR removes it from the switch case that removes it from the Calico registry (where the OS components are) so the default case hits which is the Tigera registry (quay.io or gcr.io for testing/hashreleases).

PR #1116, opened 2021-01-29
Disable cert management for variant=calico for v1.14 only

rene-dekker:
Disable cert management for variant=calico for v1.14 only


tmjd (approved):
LGTM

PR #1117, opened 2021-01-29
Need to ensure internal manager TLS secret is updated

lmm:
## Description

This PR fixes an issue in management clusters where the internal manager TLS secret is not updated.

This PR puts the rendering of the internal manager TLS secret in the controller, where we are tending to move the TLS secret creation logic. But for now I've left the other secrets in render/render.go alone.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/installation/core_controller.go
+++ b/pkg/controller/installation/core_controller.go
@@ -790,7 +790,7 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 		certDur := 825 * 24 * time.Hour // 825days*24hours: Create cert with a max expiration that macOS 10.15 will accept
 
 		managerInternalTLSSecret, err = utils.EnsureCertificateSecret(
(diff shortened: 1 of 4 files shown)


lmm:
@Brian-McM looks like kubecontrollers will be recreated with the updated secret so I believe we don't need any changes there: https://github.com/tigera/operator/blob/release-v1.14/pkg/render/render.go#L268

Brian-McM (approved):
Looks good please double check with a operator dev though before merging

tmjd (approved):
I think we know this is kind of the wrong direction to remove secret creation from the Render because as it is results in user provided secrets having OwnerReferences added but it might not matter for this one and I guess it will be consistent with the others this way.
So I'll reluctantly approve. :grinning:

lmm:
I need to merge the changes after https://github.com/tigera/operator/pull/1117 is merged.

Once this is done I'll pick into master

tmjd (approved):
LGTM but I think you should do a rebase and clean up the history so there isn't that extra merge in it.

PR #1118, opened 2021-01-30
Move managing OIDC elsticsearch user configmap and secret from operator

Suraiya-Hameed:
## Description

Move managing of OIDC Elsticsearch users ConfigMap and Secret from operator to kube-controller. 
ConfigMap `tigera-known-oidc-users` and Secret `tigera-oidc-users-elasticsearch-credentials`, needs to updated by few other internal components and doesn't fit well into tigera-operator design.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/logstorage/logstorage_controller.go
+++ b/pkg/controller/logstorage/logstorage_controller.go
@@ -372,8 +372,6 @@ func (r *ReconcileLogStorage) Reconcile(ctx context.Context, request reconcile.R
 	var elasticsearchSecrets, kibanaSecrets, curatorSecrets []*corev1.Secret
 	var clusterConfig *render.ElasticsearchClusterConfig
 	var esLicenseType render.ElasticsearchLicenseType
-	var oidcUserConfigMap *corev1.ConfigMap
-	var oidcUserSecret *corev1.Secret
 	applyTrial := false
 
 	if managementClusterConnection == nil {
@@ -429,15 +427,6 @@ func (r *ReconcileLogStorage) Reconcile(ctx context.Context, request reconcile.R
 			}
 		}
 
-		if oidcUserConfigMap, err = r.getOIDCUserConfigMap(ctx); err != nil {
-			r.status.SetDegraded("Failed to read oid user configmap", err.Error())
-			return reconcile.Result{}, err
-		}
-
-		if oidcUserSecret, err = r.oidcUsersEsSecret(ctx); err != nil {
-			r.status.SetDegraded("Failed to read oid user secret", err.Error())
-			return reconcile.Result{}, err
-		}
 	}
 
 	elasticsearch, err := r.getElasticsearch(ctx)
@@ -512,8 +501,6 @@ func (r *ReconcileLogStorage) Reconcile(ctx context.Context, request reconcile.R
 		applyTrial,
 		dexCfg,
 		esLicenseType,
-		oidcUserConfigMap,
-		oidcUserSecret,
 	)
 
 	if err = imageset.ApplyImageSet(ctx, r.client, variant, component); err != nil {
@@ -551,6 +538,16 @@ func (r *ReconcileLogStorage) Reconcile(ctx context.Context, request reconcile.R
 			reqLogger.Error(err, "failed to create or update Elasticsearch lifecycle policies")
 			r.status.SetDegraded("Failed to create or update Elasticsearch lifecycle policies", err.Error())
 			return reconcile.Result{}, err
+		}
+
+		// kube-controller creates the ConfigMap and Secret needed for SSO into Kibana.
+		// If elastisearch uses basic license, degrade logstorage if the ConfigMap and Secret
+		// needed for logging user into Kibana is not available.
+		if esLicenseType == render.ElasticsearchLicenseTypeBasic {
+			if err = r.checkOIDCUsersEsResource(ctx); err != nil {
+				r.status.SetDegraded("Failed to get oidc user Secret and ConfigMap", err.Error())
+				return reconcile.Result{}, err
+			}
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


asincu:
@Suraiya-Hameed Why is this change needed ?

pkg/render/logstorage.go
```diff
@@ -1476,13 +1459,13 @@ func (es elasticsearchComponent) oidcUserRole() []client.Object {
 					APIGroups:     []string{""},
 					Resources:     []string{"configmaps"},
 					ResourceNames: []string{OIDCUsersConfigMapName},
-					Verbs:         []string{"get", "list", "watch"},
+					Verbs:         []string{"*"},
```
asincu, line 1462: Why are we increasing access to all verbs if we are no longer watching the config map ?

pkg/render/logstorage.go
```diff
@@ -1476,13 +1459,13 @@ func (es elasticsearchComponent) oidcUserRole() []client.Object {
...
 				},
 				{
 					APIGroups:     []string{""},
 					Resources:     []string{"secrets"},
 					ResourceNames: []string{OIDCUsersEsSecreteName},
-					Verbs:         []string{"update", "get"},
+					Verbs:         []string{"*"},
```
asincu, line 1468: Why are we increasing access to all verbs ?

pkg/render/logstorage.go
Suraiya-Hameed: We are moving the code to manage this ConfigMap and Secret over to kube-controller, I've added updated this PRs description and added more details in the kube-controller PR.

asincu: I see we have create, delete, get & watch. Sounds good.

pkg/render/logstorage.go
asincu: Same as above.

asincu (approved):

pkg/controller/utils/elasticsearch.go
```diff
@@ -211,15 +211,15 @@ func (es *esClient) createOrUpdatePolicies(ctx context.Context, listPolicy map[s
...
-				es.lock.Unlock()
-			} else if elastic.IsNotFound(err) {
+			if elastic.IsNotFound(err) {
 				// If policy doesn't exist, create one
 				return applyILMPolicy(ctx, es.client, indexName, pd.policy)
 			}
+			// Reset the Elasticsearch client for all other errors.
+			// This will handle cases where Elasticsearch gets recreated or Elasticsearch credential gets recreated.
+			es.lock.Lock()
+			es.client = nil
```
rene-dekker, line 221: There are places where the es.client is used directly, like ` es.client.XPackIlmGetLifecycle().Policy(policyName).Do(ctx)`. You probably want to put access to that client in a method or it could turn into a panic.

rene-dekker, line 221: `es.client.XPackIlmGetLifecycle().Policy(policyName).Do(ctx)` still tries to access client directly. I think you should turn it into `es.client()...` to make sure you can never use a method while es == nil.

tmjd:
What uses the ConfigMap and Secret? I'm wondering if we need an annotation on a pod so that when they change/update the pod gets restarted to pick up those changes.
Should the operator continue to watch these resources and block logstorage if they don't exist?

tmjd:
Do the ILM changes need to be in this PR? It seems like the changes are not related and it would be easier to review if they were separate.
If it is necessary they are in this PR could you explain why since I didn't understand.

rene-dekker:
@tmjd That is partially my fault. I saw some inefficiencies with locking and suggested to clean that up. That led to the refactoring.

Suraiya-Hameed:
> What uses the ConfigMap and Secret? I'm wondering if we need an annotation on a pod so that when they change/update the pod gets restarted to pick up those changes.

This ConfigMap and Secret are managed by kube-controller, it watches and re-creates it as needed, and es-proxy writes into the ConfigMap and reads from Secret. We are already passing the Elasticsearch license type to kube-controller, and the kube-controller creates/deletes this ConfigMap and Secret based on that environment value. 

> Should the operator continue to watch these resources and block logstorage if they don't exist?

That's a good idea, consequence of not having this ConfigMap and Secret is that SSO-like functionality will not work when moving from ManagerUI to Kibana. It does make sense to block logstorage till they are available.

Suraiya-Hameed:
> Do the ILM changes need to be in this PR? It seems like the changes are not related and it would be easier to review if they were separate.
If it is necessary they are in this PR could you explain why since I didn't understand.

The changes to elasticsearch file can be a separate PR, it is mostly refactoring and has a fix to handle scenario where Elasticsearch password changes in the background.

pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -552,6 +539,13 @@ func (r *ReconcileLogStorage) Reconcile(ctx context.Context, request reconcile.R
 			r.status.SetDegraded("Failed to create or update Elasticsearch lifecycle policies", err.Error())
 			return reconcile.Result{}, err
 		}
+
+		if esLicenseType == render.ElasticsearchLicenseTypeBasic {
+			if err = r.checkOIDCUsersEsResource(ctx); err != nil {
```
tmjd, line 544: I'm not sure anymore if we need to block LogStorage on the OIDC resources. Is them not being present going to be a problem with LogStorage or just for the manager? Or are they used by Kibana which is launched by LogStorage.
You mentioned that going from Manager UI to Kibana wouldn't work when using SSO, but it isn't even possible to get to Kibana until Manager is available anyway.
But I guess if SSO is changed then LogStorage should be Degraded if there is a known resource missing that Kibana would need.

Assuming we should keep this check, can you add a comment here summarizing the reason why we're checking and degrading LogStorage if these resources don't exist.

pkg/render/logstorage.go
```diff
@@ -1479,13 +1462,13 @@ func (es elasticsearchComponent) oidcUserRole() []client.Object {
 					APIGroups:     []string{""},
 					Resources:     []string{"configmaps"},
 					ResourceNames: []string{OIDCUsersConfigMapName},
-					Verbs:         []string{"get", "list", "watch"},
+					Verbs:         []string{"*"},
```
tmjd, line 1465: We should specify the Verbs explicitly instead of using the wildcard. We haven't been very consistent about this so far but that is what we should be striving for.

pkg/render/logstorage.go
```diff
@@ -1479,13 +1462,13 @@ func (es elasticsearchComponent) oidcUserRole() []client.Object {
...
 				},
 				{
 					APIGroups:     []string{""},
 					Resources:     []string{"secrets"},
 					ResourceNames: []string{OIDCUsersEsSecreteName},
-					Verbs:         []string{"update", "get"},
+					Verbs:         []string{"*"},
```
tmjd, line 1471: Same here the verbs should be explicit.

tmjd (commented):
Sorry a few more comments and requests

pkg/controller/logstorage/logstorage_controller.go
Suraiya-Hameed: Yes, right, if those resources are missing SSO to Kibana will be affected, and Kibana is part of LogStorage so it makes sense to degrade it.

tmjd (approved):

marvin-tigera:

PR #1120, opened 2021-01-30
Check that cert is managed by operator by checking cert issuer

lmm:
## Description

Previous PRs added logic to check whether a cert is operator managed or supplied by the user by checking the certificate secret's ownerreference which is incorrect.

Instead this PR uses the fact that the self-signed certs issued by the operator have used the same CA signer name prefix.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


pkg/controller/manager/manager_controller_test.go
```diff
@@ -200,18 +196,14 @@ var _ = Describe("Manager controller tests", func() {
 		})
 
 		It("should reconcile if user supplied a manager TLS cert", func() {
-			// Create a manager cert secret with invalid DNS name
+			// Create a manager cert secret.
 			dnsNames := []string{"manager.example.com", "192.168.10.22"}
+			testCA := test.MakeTestCA("manager-test")
 			secret, err := render.CreateOperatorTLSSecret(
-				nil, render.ManagerTLSSecretName, render.ManagerSecretKeyName, render.ManagerSecretCertName, render.DefaultCertificateDuration, nil, dnsNames...)
+				testCA, render.ManagerTLSSecretName, render.ManagerSecretKeyName, render.ManagerSecretCertName, render.DefaultCertificateDuration, nil, dnsNames...)
```
tmjd, line 203: Can you add a test to this down below to verify that the secret created here matches the one after the Render?
Maybe we should even include a second Reconcile call here because that would have caught the previous problem.

tmjd: A second Reconcile would really only be to specifically catch if we put on owner references and then used them for something so a 2nd reconcile won't do much. So I wouldn't add a 2nd.

lmm:
Closing in favor of https://github.com/tigera/operator/pull/1121

PR #1134, opened 2021-02-02
Update so CI will push images for cloud-* tags

tmjd:
## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


ozdanborne (approved):
I have two concerns but we resolved them both offline so approving.

First question was if we're going to cut github tags for cloud operator releases. Sounds like we are, so this is needed.

Second concern was that we might accidentally overwrite an official release if we create a new branch off of an existing tag. but not only is this change not going into master, it also will be 1 commit unique as soon as it's merged.

so lgtm!

PR #1135, opened 2021-02-03
Move render common and help functions to common package

Brian-McM:
## Description

This is step one in refactoring the render package into a more manageable structure. the render/common package contains utility packages responsible for rendering k8s objects. For instance, the render/common/meta package contains utilities for creating annotations and getting namespace names, and the render/common/elasticsearch package contains decorators to add elasticsearch specific items to all the components that interact with elasticsearch.

The next step is to move every component in the render package into it's own subpacakge, i.e. render/node, render/logstorage, and break cross component access.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


caseydavenport (commented):
Nothing here looks too controversial. I would be interested in seeing the final package structure spelled out a bit more so that I understand what this is building towards a bit better.

The only other thing I noticed was this:

> 	rutil "github.com/tigera/operator/pkg/render/common"

"common" is a bit of a flag that this package will be a dumping ground for stuff that we don't know where to put anywhere else. Sometimes that's necessary, but maybe we could do better given the purpose of this is to reorganize the package structure to have better-defined packages?

Brian-McM:
> "common" is a bit of a flag that this package will be a dumping ground for stuff that we don't know where to put anywhere else.

Brian-McM:
@caseydavenport maybe we should hold this for a minute, I'll try to come up with the complete structure instead of a partial one before moving forward.

pkg/render/common/data/secrets.go
```diff
@@ -0,0 +1,92 @@
+package data
```
tmjd, line 1: I'm not a fan of `common/data`, data to me suggests that it just contains some common "data".
I don't feel like the Secret/TLS creation and resource copying fits, as "data". If we want to keep everything under data in one package it feels like a "render/utils" would be most appropriate for what is in "data" currently. I think if we can't come up with a better package name then I think "utils" is more descriptive than "data".
Saying that I think perhaps we should have a `render/common/secret` package and then any secret manipulation (including CreateTLSSecret) can go in there.

pkg/render/common/data/convenience.go
```diff
@@ -0,0 +1,46 @@
...
+			if cr.ComponentName == name && cr.ResourceRequirements != nil {
+				return *cr.ResourceRequirements
+			}
+		}
+	}
+	return v1.ResourceRequirements{}
+}
+
+// EnvVarSourceFromSecret returns an EnvVarSource using the given secret name and key.
+func EnvVarSourceFromSecret(secretName, key string, optional bool) *v1.EnvVarSource {
```
tmjd, line 31: If we have a `render/common/secret` package I think this should go in that package.

pkg/render/common/data/security.go
```diff
@@ -0,0 +1,77 @@
+package data
```
tmjd, line 1: I think we should switch these into their own package of `render/common/security`. I'd even consider splitting the functions into their own packages of `render/common/podsecuritypolicy` and `render/common/securitycontext`.

pkg/render/compliance.go
```diff
@@ -30,6 +30,9 @@ import (
 	v3 "github.com/tigera/api/pkg/apis/projectcalico/v3"
 	operatorv1 "github.com/tigera/operator/api/v1"
 	"github.com/tigera/operator/pkg/components"
+	rdata "github.com/tigera/operator/pkg/render/common/data"
+	relasticsearch "github.com/tigera/operator/pkg/render/common/elasticsearch"
```
tmjd, line 34: WDYT about `res` for this? `relasticsearch` feels a bit long and it reads like a typo to me but that could just be me, personally I'd rather it be an acronym than something that reads "odd".
I think this is just preference so no change necessary.

tmjd (commented):
I didn't do a thorough review but here are a few thoughts.

pkg/controller/amazoncloudintegration/amazoncloudintegration_controller.go
```diff
@@ -19,6 +19,8 @@ import (
 	"fmt"
 	"time"
 
+	rmeta "github.com/tigera/operator/pkg/render/common/meta"
```
caseydavenport, line 22: Do we need to alias this? We don't have any other packages named meta imported here?

pkg/controller/utils/elasticsearch.go
```diff
@@ -98,15 +102,15 @@ func ElasticsearchSecrets(ctx context.Context, userSecretNames []string, cli cli
 	return append(esUserSecrets, esCertSecret), nil
 }
 
-// GetElasticsearchClusterConfig retrieves the config map containing the elasticsearch configuration values, such as the
+// Getrelasticsearch.ClusterConfig retrieves the config map containing the elasticsearch configuration values, such as the
```
caseydavenport, line 105: I think this function should stay the same name, or be moved into an ES specific package, otherwise it's too vague for a "utils" package. 

Also the comment seems a little wonky now?

pkg/controller/logcollector/logcollector_controller.go
```diff
@@ -40,6 +40,9 @@ import (
 	"github.com/tigera/operator/pkg/controller/utils"
 	"github.com/tigera/operator/pkg/controller/utils/imageset"
 	"github.com/tigera/operator/pkg/render"
+	relasticsearch "github.com/tigera/operator/pkg/render/common/elasticsearch"
```
caseydavenport, line 43: same here - why do we need to alias everything with an "r"?

pkg/render/compliance.go
caseydavenport: Why not just "elasticsearch", or if we really feel we need to alias it do "es"? I'm not sure we need the alias, though.

I don't like "res" because that's a common abbreviation for "resource".

caseydavenport (commented):
nifty

pkg/controller/utils/elasticsearch.go
Brian-McM: Ya I did some global search and replace, I missed a few of these.

pkg/controller/amazoncloudintegration/amazoncloudintegration_controller.go
Brian-McM: We don't _need_ to, I just thought having a consistent package name would be nice to have.

pkg/controller/logcollector/logcollector_controller.go
Brian-McM: The "r" is "render", I thought it would be nicer to have that the name consistent and differentiate between the elasticsearch package which actually does stuff with elasticsarch. It's kind of like how we alias the k8s core package from /v1 to corev1 even if there's nothing conflicting.

pkg/controller/compliance/compliance_controller_test.go
```diff
@@ -173,11 +178,14 @@ var _ = Describe("Compliance controller tests", func() {
 
 		By("replacing the cert with one that has the wrong DNS names")
 		Expect(c.Delete(ctx, &corev1.Secret{ObjectMeta: metav1.ObjectMeta{Name: render.ComplianceServerCertSecret,
-			Namespace: render.OperatorNamespace()}})).NotTo(HaveOccurred())
+			Namespace: rmeta.OperatorNamespace()}})).NotTo(HaveOccurred())
 
 		oldDNSName := "compliance.tigera-compliance.svc"
-		newSecret, err := render.CreateOperatorTLSSecret(nil,
-			render.ComplianceServerCertSecret, render.ComplianceServerKeyName, render.ComplianceServerCertName, render.DefaultCertificateDuration, nil, oldDNSName,
+		ca, err := tls.MakeCA(rmeta.DefaultOperatorCASignerName())
```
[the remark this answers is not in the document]
Brian-McM: Because I don't think we should be generating tls certificates in the render package, I'd like to separate these concerns. We can achieve the same thing by having a convenience function in the controller that creates the certificate and passes it to the render function, but I just chose to do this instead since it since it was just an extra line. I can make a convenience function if you'd like in the controllers package.

pkg/controller/amazoncloudintegration/amazoncloudintegration_controller.go
caseydavenport: gotcha - maybe we should keep our own packages with the name that we give them, and alias any externally conflicting packages? 

I'm guessing this is conflicting with the k8s meta package, which is normally imported as `metav1` IIRC?

pkg/render/common/data/secrets.go
Brian-McM: I agree, I'm not the hugest fan of data either, I just don't like "utils" either. Maybe we could just have a secrets package, but I'm unsure about adding a package for everything else that's in here (some constants not secret related, config map stuff, pod security).

What do you think, should I just rename it to utils or create a bunch of separate packages that may not have much in them? I'm a little torn...

Brian-McM: I just saw your comment below being in favour of multiple packages. I'll change that.

pkg/render/common/data/security.go
Brian-McM: Will do

pkg/render/compliance.go
Brian-McM: It doesn't feel too long to me, but I think I'm more OK with longer names than most developers. @caseydavenport I'd like to keep the prefix so that it is clear that this is for rendering Elasticsearch components and not working with Elasticsearch, even if there isn't a package collision. It just keeps things consistent among all packages.

pkg/controller/amazoncloudintegration/amazoncloudintegration_controller.go
Brian-McM: Ya that's a good idea, but for Elasticsearch we have multiple packages of our own. I wanted to keep the original name for the package that actually does stuff with elasticsearch, and alias the rendering package.

caseydavenport (commented):
This LGTM - @tmjd should have the final say

pkg/controller/compliance/compliance_controller_test.go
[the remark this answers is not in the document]
tmjd: I'm talking about why do MakeCA when CreateTLSSecret will do the call for us if we pass nil CreateTLSSecret.
I don't understand how directly doing MakeCA here is a step in the direction you're suggesting of moving cert creation out of render.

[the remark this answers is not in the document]
Brian-McM: I've reverted this change as per our conversation on zoom and we'll circle back on how to handle this later.

pkg/render/common/secret/secrets.go
```diff
@@ -0,0 +1,127 @@
...
+	"github.com/tigera/operator/pkg/tls"
+
+	"github.com/openshift/library-go/pkg/crypto"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// CreateOperatorTLSSecret Creates a new TLS secret with the information passed
```
tmjd, line 18: nit: Doesn't match function name.

tmjd (commented):
One nit but LGTM. I don't need to see it again.

pkg/render/common/secret/secrets.go
Brian-McM: updated

tmjd (approved):

PR #1136, opened 2021-02-03
Fix the version verification and add separate image dump argument

tmjd:
## Description

Also include a comment in the code around needing to update the Makefile check if the line changes again.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/main.go
+++ b/main.go
@@ -79,7 +79,7 @@ func main() {
 	// kubeconfig but should use the in-cluster service account
 	var urlOnlyKubeconfig string
 	var showVersion bool
-	var dumpImages string
+	var printImages string
 	var sgSetup bool
 	flag.BoolVar(&enableLeaderElection, "enable-leader-election", true,
 		"Enable leader election for controller manager. "+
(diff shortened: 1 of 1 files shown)


ozdanborne (commented):
When I first saw this error, I thought it was a bug that we were printing all three of the Operator, Calico, and Enterprise versions on the same line. Are you sure we don't want to split that over three lines instead?

tmjd:
I kept it at one line because I thought for purposes of using that output to get the images it would be good to have it all on one line.
That being said maybe we should add a new option for dumping the image list separately, and then splitting those among a few lines wouldn't matter.

tmjd:
Could we do that additional change as a separate PR? That way I can pick this one to the release-v1.14 and cloud-dev and not make "unnecessary" changes on top.

tmjd:
Ok I changed my mind about the "unnecessary" changes and updated this PR.
WDYT?

ozdanborne (approved):

PR #1137, opened 2021-02-03
Add eck-operator memory requests/limits to LogStorage CR.

vberezny:
## Description

[Bug Fix] Exposes the elastic-operator pod memory requests/limits in the LogStorage CR. This was done to allow the end user to easily get around issues with elastic-operator pod being OOM killed due to memory usage.

Tested on the local kind cluster as well as in a banzai provisioned gcp-kubeadm environment.

https://tigera.atlassian.net/browse/SAAS-1250

Corresponding Docs PR:
https://github.com/tigera/calico-private/pull/2986

## For PR author

- [x ] Tests for change.
- [x ] If changing pkg/apis/, run `make gen-files`

## For PR reviewers

- [ ] Appropriate labels:


api/v1/installation_types.go
```diff
@@ -141,6 +141,7 @@ const (
 	ComponentNameNode            ComponentName = "Node"
 	ComponentNameTypha           ComponentName = "Typha"
 	ComponentNameKubeControllers ComponentName = "KubeControllers"
+	ComponentNameECKOperator     ComponentName = "ECKOperator"
```
tmjd, line 144: You'll need to include below in the validation that ECKOperator is allowed.
I think also on Line 111 above you need to add a note that "Node, Typha, and KubeControllers" are supported.
Corresponding to that in `pkg/controller/installation/core_controller.go` should have validation that only those 3 are present on its componentResources.

pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -273,6 +275,22 @@ func fillDefaults(opr *operatorv1.LogStorage) {
...
+		opr.Spec.ComponentResources = []operatorv1.ComponentResource{
+			{
+				ComponentName: operatorv1.ComponentNameECKOperator,
+				ResourceRequirements: &corev1.ResourceRequirements{
+					Limits:   limits,
+					Requests: requests,
+				},
+			},
+		}
+	}
```
tmjd, line 293: Not in fillDefaults here but I think where we call fillDefaults you should add a new function to verify the LogStorage spec and that should include that ComponentResources only has a ComponentName of ECKOperator.

tmjd (changes requested):
Looking good, I think a few additions are needed though.
In addition to some validation (see my comments) I'd also like to see UTs for the validation.
Also make sure you run `make gen-files` to update the CRDs and deepcopy, I pulled your branch and ran it and saw a CRD update.

pkg/render/logstorage_test.go
```diff
@@ -617,6 +617,105 @@ var _ = Describe("Elasticsearch rendering tests", func() {
 			oidcUserSecret := GetResource(createResources, render.OIDCUsersEsSecreteName, render.ElasticsearchNamespace, "", "v1", "Secret")
 			Expect(oidcUserSecret).ShouldNot(BeNil())
 		})
+
+		Context("ECKOperator memory requests/limits", func() {
+			When("LogStorage Spec contains an entry for ECKOperator in ComponentResources", func() {
+				It("should set matching memory requests/limits in the elastic-operator StatefulSet Spec, manager container", func() {
```
tmjd, line 623: I think this test should not include all those other resource checks. I would prefer it to be focused on the minimum test necessary to see that the memory requests/limits are set, because with all this other stuff then if any of them change or are removed this test which is unrelated has to be updated.

pkg/controller/installation/core_controller.go
```diff
@@ -559,6 +563,26 @@ func fillDefaults(instance *operator.Installation) error {
...
+
+	validComponentNames := map[operator.ComponentName]struct{}{
+		operator.ComponentNameKubeControllers: {},
+		operator.ComponentNameNode:            {},
+		operator.ComponentNameTypha:           {},
+	}
+
+	for _, resource := range installation.Spec.ComponentResources {
+		if _, ok := validComponentNames[resource.ComponentName]; !ok {
+			return fmt.Errorf("installation spec componentResources contains an invalid componentName %+v", resource)
```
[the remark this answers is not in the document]
vberezny: Made the adjustment to match the other validation error messages but rather than `installation.spec.componentResources` I made it `Installation spec.ComponentResources` since that is the pattern I'm seeing in older error messages. if the `installation.spec.componentResources` format is preferable I can adjust my additions along with the older messages to match this format.

pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -273,6 +279,71 @@ func fillDefaults(opr *operatorv1.LogStorage) {
...
+}
+
+func validateLogStorageSpec(spec *operatorv1.LogStorageSpec) error {
+	if spec.Retention == nil {
+		return fmt.Errorf("LogStorage Spec missing Retention %+v", spec)
```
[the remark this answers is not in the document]
vberezny: Removed the verification and replaced with a direct test for fillDefaults() in `logstorage_controller_test.go`.

vberezny:
Looking into build failures right now.

tmjd (approved):
LGTM

PR #1138, opened 2021-02-03
Only merge metadata if it's not set on the desired state

caseydavenport:
## Description

If the controller, for whatever reason, wants to assert a certain
revision or UID for an object, then the merge logic should respect that.

Most objects will come through without revision information, in which
case it will be copied from the live state on the cluster.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/utils/component.go
+++ b/pkg/controller/utils/component.go
@@ -195,10 +195,16 @@ func mergeState(desired client.Object, current runtime.Object) client.Object {
 	currentMeta := current.(metav1.ObjectMetaAccessor).GetObjectMeta()
 	desiredMeta := desired.(metav1.ObjectMetaAccessor).GetObjectMeta()
 
-	// Merge common metadata fields.
-	desiredMeta.SetResourceVersion(currentMeta.GetResourceVersion())
-	desiredMeta.SetUID(currentMeta.GetUID())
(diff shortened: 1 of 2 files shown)


Brian-McM (approved):

pkg/controller/utils/component.go
```diff
@@ -195,10 +195,16 @@ func mergeState(desired client.Object, current runtime.Object) client.Object {
 	currentMeta := current.(metav1.ObjectMetaAccessor).GetObjectMeta()
 	desiredMeta := desired.(metav1.ObjectMetaAccessor).GetObjectMeta()
 
-	// Merge common metadata fields.
-	desiredMeta.SetResourceVersion(currentMeta.GetResourceVersion())
-	desiredMeta.SetUID(currentMeta.GetUID())
-	desiredMeta.SetCreationTimestamp(currentMeta.GetCreationTimestamp())
+	// Merge common metadata fields if not present on the desired state.
+	if desiredMeta.GetResourceVersion() != "" {
```
tmjd, line 199: Shouldn't these checks be if `== ""`

caseydavenport: d'oh

Brian-McM: wow can't believe I missed that too lol

PR #1142, opened 2021-02-03
include OS node selector in PodTemplate resource types

robbrockbank:
## Description

Compliance defines a pod template - looks like the pods template is not included in the os-node-selector updates - it should be,

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/utils/component.go
+++ b/pkg/controller/utils/component.go
@@ -301,6 +301,8 @@ func ensureOSSchedulingRestrictions(obj client.Object, osType render.OSType) {
 
 	var podSpecs []*v1.PodSpec
 	switch obj.(type) {
+	case *v1.PodTemplate:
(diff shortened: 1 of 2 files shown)


lmm (approved):
LGTM

pkg/controller/utils/component.go
```diff
@@ -301,6 +301,8 @@ func ensureOSSchedulingRestrictions(obj client.Object, osType render.OSType) {
 
 	var podSpecs []*v1.PodSpec
 	switch obj.(type) {
+	case *v1.PodTemplate:
```
Brian-McM, line 304: I'm a little confused, how does a PodTemplate end up getting here, these should be renderable items, is a pod template being passed as a resource that should be rendered?

Brian-McM: Nm @robbrockbank answer this and it's a valid type that you can apply.

Brian-McM (approved):

PR #1148, opened 2021-02-09
Install compliance and intrusion detection if features are active

asincu:
## Description

Compliance and intrusion detection controller will read the LicenseKeyStatus that exposes a new field, called Features. If their respective feature names are active on the license, the installation will continue. Otherwise, the component will not installed. 

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/compliance/compliance_controller.go
+++ b/pkg/controller/compliance/compliance_controller.go
@@ -23,6 +23,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 
 	operatorv1 "github.com/tigera/operator/api/v1"
+	"github.com/tigera/operator/pkg/common"
 	"github.com/tigera/operator/pkg/controller/installation"
 	"github.com/tigera/operator/pkg/controller/options"
 	"github.com/tigera/operator/pkg/controller/status"
@@ -186,8 +187,19 @@ func (r *ReconcileCompliance) Reconcile(ctx context.Context, request reconcile.R
 		return reconcile.Result{}, err
 	}
 
-	if err = utils.CheckLicenseKey(ctx, r.client); err != nil {
-		r.status.SetDegraded("License not found", err.Error())
+	license, err := utils.FetchLicenseKey(ctx, r.client)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			r.status.SetDegraded("License not found", err.Error())
+			return reconcile.Result{RequeueAfter: 10 * time.Second}, nil
+		}
+		r.status.SetDegraded("Error querying license", err.Error())
+		return reconcile.Result{RequeueAfter: 10 * time.Second}, nil
+	}
+
+	if !utils.IsFeatureActive(license, common.ComplianceFeature) {
+		log.V(4).Info("Compliance is not activated as part of this license")
+		r.status.SetDegraded("Feature is not active", "License does not support this feature")
 		return reconcile.Result{RequeueAfter: 10 * time.Second}, nil
 	}
 
--- a/pkg/controller/intrusiondetection/intrusiondetection_controller.go
+++ b/pkg/controller/intrusiondetection/intrusiondetection_controller.go
@@ -22,6 +22,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 
 	operatorv1 "github.com/tigera/operator/api/v1"
+	"github.com/tigera/operator/pkg/common"
 	"github.com/tigera/operator/pkg/controller/installation"
 	"github.com/tigera/operator/pkg/controller/logcollector"
 	"github.com/tigera/operator/pkg/controller/options"
@@ -180,8 +181,19 @@ func (r *ReconcileIntrusionDetection) Reconcile(ctx context.Context, request rec
 		return reconcile.Result{}, err
 	}
 
-	if err = utils.CheckLicenseKey(ctx, r.client); err != nil {
-		r.status.SetDegraded("License not found", err.Error())
+	license, err := utils.FetchLicenseKey(ctx, r.client)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			r.status.SetDegraded("License not found", err.Error())
+			return reconcile.Result{RequeueAfter: 10 * time.Second}, nil
+		}
+		r.status.SetDegraded("Error querying license", err.Error())
+		return reconcile.Result{RequeueAfter: 10 * time.Second}, nil
+	}
+
+	if !utils.IsFeatureActive(license, common.ThreatDefenseFeature) {
+		log.V(4).Info("IntrusionDetection is not activated as part of this license")
+		r.status.SetDegraded("Feature is not active", "License does not support this feature")
 		return reconcile.Result{RequeueAfter: 10 * time.Second}, nil
 	}
 
--- a/pkg/controller/logcollector/logcollector_controller.go
+++ b/pkg/controller/logcollector/logcollector_controller.go
@@ -184,8 +184,12 @@ func (r *ReconcileLogCollector) Reconcile(ctx context.Context, request reconcile
 		return reconcile.Result{}, nil
 	}
 
-	if err = utils.CheckLicenseKey(ctx, r.client); err != nil {
-		r.status.SetDegraded("License not found", err.Error())
+	if _, err := utils.FetchLicenseKey(ctx, r.client); err != nil {
+		if errors.IsNotFound(err) {
+			r.status.SetDegraded("License not found", err.Error())
+			return reconcile.Result{RequeueAfter: 10 * time.Second}, nil
+		}
+		r.status.SetDegraded("Error querying license", err.Error())
 		return reconcile.Result{RequeueAfter: 10 * time.Second}, nil
 	}
 
--- a/pkg/controller/manager/manager_controller.go
+++ b/pkg/controller/manager/manager_controller.go
@@ -207,8 +207,13 @@ func (r *ReconcileManager) Reconcile(ctx context.Context, request reconcile.Requ
 		return reconcile.Result{}, nil
 	}
 
-	if err = utils.CheckLicenseKey(ctx, r.client); err != nil {
-		r.status.SetDegraded("License not found", err.Error())
+	license, err := utils.FetchLicenseKey(ctx, r.client)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			r.status.SetDegraded("License not found", err.Error())
+			return reconcile.Result{RequeueAfter: 10 * time.Second}, nil
+		}
+		r.status.SetDegraded("Error querying license", err.Error())
 		return reconcile.Result{RequeueAfter: 10 * time.Second}, nil
 	}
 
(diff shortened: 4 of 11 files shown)


pkg/controller/compliance/compliance_controller.go
```diff
@@ -184,11 +184,17 @@ func (r *ReconcileCompliance) Reconcile(ctx context.Context, request reconcile.R
...
 	}
 
-	if err = utils.CheckLicenseKey(ctx, r.client); err != nil {
+	license, err := utils.CheckLicenseKey(ctx, r.client)
+	if err != nil {
 		r.status.SetDegraded("License not found", err.Error())
 		return reconcile.Result{RequeueAfter: 10 * time.Second}, nil
 	}
 
+	if !utils.IsComplianceFeatureActive(license) {
```
rene-dekker, line 193: If compliance feature is not active, my guess would be that:
- we probably want to throw an error to make people remove compliance
- we set r.status to degraded

pkg/controller/utils/utils.go
```diff
@@ -179,9 +179,20 @@ func LogStorageExists(ctx context.Context, cli client.Client) (bool, error) {
 // CheckLicenseKey checks if a license has been installed. It's useful
 // to prevent rollout of TSEE components that might require it.
 // It will return an error if the license is not installed, and nil otherwise.
-func CheckLicenseKey(ctx context.Context, cli client.Client) error {
+func CheckLicenseKey(ctx context.Context, cli client.Client) (v3.LicenseKey, error) {
```
rene-dekker, line 182: Maybe you could rename this to GetLicenseKey?

pkg/controller/compliance/compliance_controller.go
```diff
@@ -184,11 +185,17 @@ func (r *ReconcileCompliance) Reconcile(ctx context.Context, request reconcile.R
...
-	if err = utils.CheckLicenseKey(ctx, r.client); err != nil {
+	license, err := utils.FetchLicenseKey(ctx, r.client)
+	if err != nil {
 		r.status.SetDegraded("License not found", err.Error())
 		return reconcile.Result{RequeueAfter: 10 * time.Second}, nil
 	}
 
+	if !utils.IsFeatureActive(license, common.ComplianceFeature) {
+		log.Info("Compliance is not activated as part of this license")
+		return reconcile.Result{RequeueAfter: 10 * time.Second}, nil
```
caseydavenport, line 196: Shouldn't we be calling SetDegraded here? I thought that was the design.

pkg/controller/intrusiondetection/intrusiondetection_controller.go
```diff
@@ -178,11 +179,17 @@ func (r *ReconcileIntrusionDetection) Reconcile(ctx context.Context, request rec
...
-	if err = utils.CheckLicenseKey(ctx, r.client); err != nil {
+	license, err := utils.FetchLicenseKey(ctx, r.client)
+	if err != nil {
 		r.status.SetDegraded("License not found", err.Error())
 		return reconcile.Result{RequeueAfter: 10 * time.Second}, nil
 	}
 
+	if !utils.IsFeatureActive(license, common.ThreatDefenseFeature) {
+		log.Info("IntrusionDetection is not activated as part of this license")
+		return reconcile.Result{RequeueAfter: 10 * time.Second}, nil
```
caseydavenport, line 190: Same here, if the CR is provided but the license doesn't enable it, we should mark the user's intent as degraded.

pkg/controller/utils/utils.go
```diff
@@ -176,12 +176,24 @@ func LogStorageExists(ctx context.Context, cli client.Client) (bool, error) {
 	return true, nil
 }
 
-// CheckLicenseKey checks if a license has been installed. It's useful
+// FetchLicenseKey checks if a license has been installed. It's useful
```
caseydavenport, line 179: > It will return an error if the license is not installed, and nil otherwise.

Comment is subtly wrong now since this has two returns now

pkg/controller/intrusiondetection/intrusiondetection_controller.go
asincu: @caseydavenport I added the change

```
kubectl get tigerastatus compliance -o yaml 

kind: TigeraStatus
metadata:
  creationTimestamp: "2021-02-10T19:16:11Z"
  generation: 1
  managedFields:
    fieldsType: FieldsV1
    fieldsV1:
      f:spec: {}
      f:status:
        .: {}
        f:conditions: {}
    manager: operator
    operation: Update
    time: "2021-02-10T19:16:11Z"
  name: compliance
  resourceVersion: "197120"
  selfLink: /apis/operator.tigera.io/v1/tigerastatuses/compliance
  uid: e140454f-1e28-4d7f-a3b0-61dc08cd5d9a
spec: {}
status:
  conditions:
  - lastTransitionTime: "2021-02-10T19:16:11Z"
    message: License does not support this feature
    reason: Feature is not active
    status: "True"
    type: Degraded
  - lastTransitionTime: "2021-02-10T19:16:16Z"
    status: "False"
    type: Available
  - lastTransitionTime: "2021-02-10T19:16:21Z"
    status: "False"
    type: Progressing

```

pkg/controller/compliance/compliance_controller.go
asincu:
```
kubectl get tigerastatus intrusion-detection -o yaml

kind: TigeraStatus
metadata:
  creationTimestamp: "2021-02-10T19:16:11Z"
  generation: 1
  managedFields:
    fieldsType: FieldsV1
    fieldsV1:
      f:spec: {}
      f:status:
        .: {}
        f:conditions: {}
    manager: operator
    operation: Update
    time: "2021-02-10T19:16:11Z"
  name: intrusion-detection
  resourceVersion: "196988"
  selfLink: /apis/operator.tigera.io/v1/tigerastatuses/intrusion-detection
  uid: bc96282e-5e05-41e5-9461-7ea2c9ac0615
spec: {}
status:
  conditions:
  - lastTransitionTime: "2021-02-10T19:16:11Z"
    status: "False"
    type: Progressing
  - lastTransitionTime: "2021-02-10T19:16:11Z"
    message: License does not support this feature
    reason: Feature is not active
    status: "True"
    type: Degraded
  - lastTransitionTime: "2021-02-10T19:16:16Z"
    status: "False"
    type: Available

```

pkg/controller/compliance/compliance_controller.go
```diff
@@ -184,11 +185,18 @@ func (r *ReconcileCompliance) Reconcile(ctx context.Context, request reconcile.R
...
 
-	if err = utils.CheckLicenseKey(ctx, r.client); err != nil {
+	license, err := utils.FetchLicenseKey(ctx, r.client)
+	if err != nil {
 		r.status.SetDegraded("License not found", err.Error())
 		return reconcile.Result{RequeueAfter: 10 * time.Second}, nil
 	}
 
+	if !utils.IsFeatureActive(license, common.ComplianceFeature) {
+		log.Info("Compliance is not activated as part of this license")
```
caseydavenport, line 195: This will potentially be a very chatty log, if we're logging it every 10 seconds. Might want to make it a debug log?

pkg/controller/intrusiondetection/intrusiondetection_controller.go
```diff
@@ -178,11 +179,18 @@ func (r *ReconcileIntrusionDetection) Reconcile(ctx context.Context, request rec
...
 
-	if err = utils.CheckLicenseKey(ctx, r.client); err != nil {
+	license, err := utils.FetchLicenseKey(ctx, r.client)
+	if err != nil {
 		r.status.SetDegraded("License not found", err.Error())
 		return reconcile.Result{RequeueAfter: 10 * time.Second}, nil
 	}
 
+	if !utils.IsFeatureActive(license, common.ThreatDefenseFeature) {
+		log.Info("IntrusionDetection is not activated as part of this license")
```
caseydavenport, line 189: Same for this guy - should probably be debug.

pkg/controller/manager/manager_controller.go
```diff
@@ -271,19 +272,23 @@ func (r *ReconcileManager) Reconcile(ctx context.Context, request reconcile.Requ
 		}
 	}
 
-	// Check that compliance is running.
-	compliance, err := compliance.GetCompliance(ctx, r.client)
-	if err != nil {
-		if errors.IsNotFound(err) {
-			r.status.SetDegraded("Compliance not found", err.Error())
+	var installCompliance = utils.IsFeatureActive(license, common.ComplianceFeature)
```
caseydavenport, line 275: What happens when my license doesn't have this capability, but then I later apply a new license that has the feature enabled? How will we detect this and update the installed components?

Are we just relying on the compliance_controller's 10s retry timer to detect and update the status, thus triggering the manager as well?

I suppose that will probably be good enough, in the absence of a true watch for the License itself.

caseydavenport (approved):

pkg/controller/manager/manager_controller.go
asincu: @caseydavenport Yes, it will install the component if a license is being installed that has that feature.

pkg/controller/logcollector/logcollector_controller.go
```diff
@@ -181,7 +181,7 @@ func (r *ReconcileLogCollector) Reconcile(ctx context.Context, request reconcile
 		return reconcile.Result{}, nil
 	}
 
-	if err = utils.CheckLicenseKey(ctx, r.client); err != nil {
+	if _, err := utils.FetchLicenseKey(ctx, r.client); err != nil {
```
rene-dekker, line 184: If there is an error, we cannot conclude it is because the license was not found. It could be any error. Therefore I think we should improved the degraded message to something like: "Unable to fetch License".
There are multiple occurrences of this in the PR.

caseydavenport: We can check the specific type of error - we do it elsewhere.

i.e., if it's a "NotFound" error it's different than a connection error. We should probably do that here.

pkg/controller/manager/manager_controller.go
```diff
@@ -271,19 +272,23 @@ func (r *ReconcileManager) Reconcile(ctx context.Context, request reconcile.Requ
...
+
+	if installCompliance {
```
rene-dekker, line 277: There are two if installCompliance blocks, you could add all logic into one.

pkg/controller/logcollector/logcollector_controller.go
asincu: @rene-dekker I actually see that we set `r.status.SetDegraded("License not found", err.Error())`

This means that on the status of the tigera status we set something like. 
  - lastTransitionTime: "2021-02-10T19:16:11Z"
    message: License not found
    reason: (ks8 API timeout | license not found)
    status: "True"

pkg/controller/manager/manager_controller.go
rene-dekker: Alternatively, we have this block with a helper method, which could be copied for compliance.

```
	if !utils.IsAPIServerReady(r.client, reqLogger) {
		r.status.SetDegraded("Waiting for Tigera API server to be ready", "")
	}
```

asincu: Sure, I can combine them. I might be mistaken, but I do not think the sample you provided applies to this PR ...

pkg/controller/logcollector/logcollector_controller.go
rene-dekker: In other places we change the message as well, based on the type of error. I'm not super picky about it, but for consistency it might be better.

pkg/controller/manager/manager_controller.go
rene-dekker:
```
if installCompliance && !utils.isComplianceReady (r.client, reqLogger) {
		r.status.SetDegraded("Waiting for Compliance to be ready", "")
	
}
```

asincu: I would rather not create a function that required a logger as a parameter and keep the logic in this controller.

PR #1151, opened 2021-02-10
Use FQDN elasticsearch and kibana endpoints only for Windows

lmm:
## Description

For components running on Linux, FQDN endpoints for ES and Kibana aren't required so revert that change. This smooths out the upgrade for users of custom TLS certs.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/utils/utils.go
+++ b/pkg/controller/utils/utils.go
@@ -197,18 +197,12 @@ func IsFeatureActive(license v3.LicenseKey, featureName string) bool {
 	return false
 }
 
-// ValidateCertPair validates the cert pair secret in the tigera-operator namespace.
-func ValidateCertPair(client client.Client, certPairSecretName, keyName, certName string) (*corev1.Secret, error) {
-	return ValidateCertPairInNamespace(client, rmeta.OperatorNamespace(), certPairSecretName, keyName, certName)
-}
-
-// ValidateCertPairInNamespace checks if the given secret exists in the given
+// ValidateCertPair checks if the given secret exists in the given
 // namespace and if so that it contains key and cert fields. If a secret exists then it is returned.
 // If there is an error accessing the secret (except NotFound) or the cert
 // does not have both a key and cert field then an appropriate error is returned.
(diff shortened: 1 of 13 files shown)


pkg/controller/utils/utils.go
```diff
@@ -184,17 +184,11 @@ func CheckLicenseKey(ctx context.Context, cli client.Client) error {
...
-	return ValidateCertPairInNamespace(client, render.OperatorNamespace(), certPairSecretName, keyName, certName)
-}
-
 // ValidateCertPairInNamespace checks if the given secret exists in the given
 // namespace and if so that it contains key and cert fields. If a secret exists then it is returned.
 // If there is an error accessing the secret (except NotFound) or the cert
 // does not have both a key and cert field then an appropriate error is returned.
 // If no secret exists then nil, nil is returned to represent that no cert is valid.
-// TODO: Replace this with version of ValidateCertPair taking in a ns.
 func ValidateCertPairInNamespace(client client.Client, namespace, certPairSecretName, keyName, certName string) (*corev1.Secret, error) {
```
Brian-McM, line 192: I think you could / maybe should keep the old name, I don't think you need to describe that this is for a namespace when it's the only cert validation function.

lmm: Good point, fixed!

pkg/render/common/kibana/service.go
```diff
@@ -0,0 +1,20 @@
+package kibana
```
Brian-McM, line 1: I think we can keep elasticsearch and kibana in the same package for now, they're pretty tightly coupled together right now so this could cause circular dependencies.

We should definitely evaluate if this should still be the case once the render refactoring is done.

Brian-McM: I retract this comment, there is no kibana stuff in the elasticsearch package and this is absolutely the right thing to do.

Brian-McM (approved):
Looks good! Please get an operator dev to approve this though.

caseydavenport (approved):

PR #1163, opened 2021-02-17
Copy ComponentResources struct to eliminate CRD confusion

vberezny:
## Description
[Bug Fix] Copy ComponentResources struct to make generated CRDs accurately reflect supported components and their resource requirements.

## For PR author

- [x ] Tests for change.
- [x ] If changing pkg/apis/, run `make gen-files`

## Issue
https://tigera.atlassian.net/browse/SAAS-1250


api/v1/logstorage_types.go
```diff
@@ -142,6 +142,22 @@ type Retention struct {
...
+
+const (
+	ComponentNameECKOperator LogStorageComponentName = "ECKOperator"
+)
+
+// The ComponentResource struct associates a ResourceRequirements with a component by name
+type LogStorageComponentResource struct {
+	// ComponentName is an enum which identifies the component
+	// +kubebuilder:validation:Enum=ECKOperator
+	LogStorageComponentName LogStorageComponentName `json:"componentName"`
```
tmjd, line 156: I think we should just call this ComponentName, WDYT?
Either way this name and the comment above should match.

tmjd (commented):
Looks good, one small naming thing to fix.

api/v1/logstorage_types.go
vberezny: Yeah good call, changed it to `ComponentName`.

tmjd (approved):
LGTM

PR #1164, opened 2021-02-17
Add linuxDataplane option and defaulting/extra mounts for BPF

fasaxc:
## Description

* Add a new linuxDataplane field to the operator.  
* When set to "BPF":
  * Add the extra mount point required by BPF mode.  I made this a "must exist" mount point so that the roll out will fail on a cluster that doesn't have the needed FS mounted.
  * Validate against incompatible fields.
  * Set FELIX_BPFENABLED=true
* Change defaulting for IP pool; only create a default IP pool if the IPAM is Calico IPAM.  This prevents an empty calicoNetwork section from triggering IP pool creation.

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


pkg/controller/installation/validation.go
```diff
@@ -260,6 +264,11 @@ func validateCustomResource(instance *operatorv1.Installation) error {
 			if instance.Spec.CNI.Type != operatorv1.PluginCalico {
 				return fmt.Errorf("spec.calicoNetwork.hostPorts is supported only for Calico CNI")
 			}
+
+			if instance.Spec.CalicoNetwork.LinuxDataplane != nil && *instance.Spec.CalicoNetwork.LinuxDataplane == operatorv1.LinuxDataplaneBPF {
```
caseydavenport, line 268: Presumably if host ports is explicitly "Disabled" that would be an acceptable config?

fasaxc: I guess, but the check above didn't seem to make that exception?  Happy to fix up both if that makes sense

caseydavenport: I think the difference is that if the CNI plugin isn't Calico, we actually can't control if host ports is enabled or disabled, since we're not actually installing and configuring the CNI plugin, so accepting either option (Enabled or Disabled) would be wrong, since it's out of our control.

If the CNI plugin is Calico, but the data plane mode is BPF, we still control whether the hostport plugin is actually installed / configured, so we could choose to accept the values for that field that are compatible. 

It's a minor thing though. Worth checking if we actually configure hostPorts if that field is nil, though.

fasaxc: Maybe it's different for non-Calico CNI; the validation there is saying "we're not even in control so even Disabled makes no sense".

fasaxc: Crossed in the post(!) I adjusted it to ignore disabled only for BPF mode.

api/v1/installation_types.go
```diff
@@ -248,8 +248,22 @@ const (
...
+const (
+	LinuxDataplaneIptables LinuxDataplaneOption = "Iptables"
+	LinuxDataplaneBPF      LinuxDataplaneOption = "BPF"
+)
+
 // CalicoNetworkSpec specifies configuration options for Calico provided pod networking.
 type CalicoNetworkSpec struct {
+	// LinuxDataplane is used to select the dataplane used for Linux nodes. In particular, it
+	// causes the operator to add required mounts and environment variables for the particular dataplane.
+	// If not specified, iptables mode is used.
```
tmjd, line 262: We should default this field in `fillDefaults` in the `pkg/controller/installation/core_controller.go`, when CalicoNetwork is exists.

tmjd (commented):
Is there any additional defaulting that should be done when the dataplane is BPF is set? would we want to default the IPPools differently if BPF is set?
I think we should have validation that NodeAddressAutodetection is set even if the value has no subfield set.

fasaxc:
> would we want to default the IPPools differently if BPF is set?

BPF with Calico networking should do the usual Calico thing, I think.  However, maybe we should skip creating IP pools if not suing Calico CNI (even if calicoNetwork is non-nil)?

fasaxc:
> would we want to default the IPPools differently if BPF is set?

I've changed this to depend on the IPAM plugin in use, does that make sense?

tmjd (commented):
I'm thinking we should always default a CalicoNetwork if it is nil.
Now that we only create IPPools with Calico CNI, it should be fine to always initialize BGP since BGP default already takes into account the CNI type, and it seems like we'd always want to default the dataplane config too.

I thought you were considering to wait for the node rollout to finish and then once that happened update FelixConfig to have BPF mode enabled. That way there wouldn't be a mixed dataplane cluster (at least for a much shorter period).

pkg/render/node.go
```diff
@@ -695,6 +696,12 @@ func (c *nodeComponent) nodeVolumes() []v1.Volume {
 		},
 	}
 
+	if c.cr.CalicoNetwork != nil &&
+		c.cr.CalicoNetwork.LinuxDataplane != nil &&
+		*c.cr.CalicoNetwork.LinuxDataplane == operatorv1.LinuxDataplaneBPF {
+		volumes = append(volumes, v1.Volume{Name: "bpffs", VolumeSource: v1.VolumeSource{HostPath: &v1.HostPathVolumeSource{Path: "/sys/fs/bpf", Type: &dirMustExist}}})
```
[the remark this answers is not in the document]
fasaxc: My idea was that, since the user is opting in to BPF, we can be more strict here (and allow progress towards a non-privileged calico-node); doing it this way should halt the roll out if the bpffs isn't mounted, which seems like a good compromise and at least gives some feedback to the user.

fasaxc:
>I thought you were considering to wait for the node rollout to finish and then once that happened update FelixConfig to have BPF mode enabled. That way there wouldn't be a mixed dataplane cluster (at least for a much shorter period).

I thought Casey cautioned that we don't have great support for that in operator just yet so I figured let's get this increment banked and maybe we can improve it later.

tmjd:
> > I thought you were considering to wait for the node rollout to finish and then once that happened update FelixConfig to have BPF mode enabled. That way there wouldn't be a mixed dataplane cluster (at least for a much shorter period).
>
> I thought Casey cautioned that we don't have great support for that in operator just yet so I figured let's get this increment banked and maybe we can improve it later.

@fasaxc that seems a change of concerns compared to when we first talked about enabling BPF in operator. As the concern with having operator monitor FelixConfig and update the volume mount when BPF was enabled in FelixConfig was that there would be a temporary disruption of traffic but maybe the key difference was that policy would temporarily be dropped when switching over. That switch sounds much less disruptive than what this PR is currently implementing.

After taking a look, I don't think this would be very hard because we already have the check `if !r.status.IsAvailable()` at https://github.com/tigera/operator/blob/master/pkg/controller/installation/core_controller.go#L1124, seems like we could just include a check after that if BPF is enabled that the operator could patch the felixconfig to enable BPF.
I guess we would also need a corresponding check, before rendering, to disable BPF in felixconfig if the Installation did not have BPF enabled.
@caseydavenport WDYT?

fasaxc:
The issue I'm trying to solve is that restarting calico/node completely breaks the BPF dataplane if this mount point is not in place.  It effectively:

* Flushes conntrack (so existing flows all get dropped)
* Removes the policy from host endpoints, so they end up allowing all traffic.
* Removes the policy from workload endpoints, so they drop all traffic.

When calico-node comes back up it fixes the latter two but all existing flows remain broken.

This change causes disruption during the switch over to BPF but only for node ports.  That's not great, but I can document around it if need be.

fasaxc:
> I'm thinking we should always default a CalicoNetwork if it is nil

Done

caseydavenport:
> After taking a look, I don't think this would be very hard because we already have the check if !r.status.IsAvailable() at https://github.com/tigera/operator/blob/master/pkg/controller/installation/core_controller.go#L1124, seems like we could just include a check after that if BPF is enabled that the operator could patch the felixconfig to enable BPF.

@tmjd I think it is a little bit more complicated than that (although maybe not too much more). If we're going from an already avaialble state with linux mode, to eBPF, that `r.status.isAvailable()` call is going to race with rolling out the new config. We may pass that check simply because the status manager hasn't had time to update between applying the new resources and the `IsAvailable` test. 

Unless I've missed something, we don't currently have a mechanism to say "we're rolling out a new state, so clear the available bit until that rollout is over". If we had that, then I think it would be doable like you said. I'm not sure how we do that though - we don't currently have a way of comparing cluster state with the desired state - we just always blindly apply the desired state.

caseydavenport (approved):
This LGTM - there is a potential improvement to the eBPF enabling workflow so that we only switch to eBPF after the volume mounts are in place. 

This should be possible, but IMO is fine to do in a second PR. We just need to note that there will be some potential disruption when switching data planes.

tmjd (approved):
LGTM

fasaxc:
Thanks, :crossed_fingers:

PR #1165, opened 2021-02-17
Standardize intallation enum documentation format

caseydavenport:
## Description

Some of the enums didn't list the valid values, resulting in docs that were incomplete, like this: https://docs.projectcalico.org/reference/installation/api#operator.tigera.io/v1.BGPOption

This PR standardizes on a format for enums that should render nicely in the docs. 

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/api/v1/installation_types.go
+++ b/api/v1/installation_types.go
@@ -135,7 +135,9 @@ type PreferredNodeAffinity struct {
 	PreferredDuringSchedulingIgnoredDuringExecution []v1.PreferredSchedulingTerm `json:"preferredDuringSchedulingIgnoredDuringExecution,omitempty"`
 }
 
-// ComponentName CRD enum
+// ComponentName represents a single component.
+//
+// One of: Node, Typha, KubeControllers
 type ComponentName string
 
 const (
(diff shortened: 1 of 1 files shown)


tmjd:
Do we have any in this format already that I can see what this looks like in the API reference?

api/v1/installation_types.go
```diff
@@ -135,7 +135,9 @@ type PreferredNodeAffinity struct {
 	PreferredDuringSchedulingIgnoredDuringExecution []v1.PreferredSchedulingTerm `json:"preferredDuringSchedulingIgnoredDuringExecution,omitempty"`
 }
 
-// ComponentName CRD enum
+// ComponentName represents a single component.
+//
+// One of: Node, Typha, KubeControllers, ECKOperator
```
tmjd, line 140: This is wrong now because ECKOperator is now in its own type.

caseydavenport: According to the enum itself it's still valid (see line 147 below) - should I remove that?

marvin-tigera:

caseydavenport:
> Do we have any in this format already that I can see what this looks like in the API reference?

Nope, nothing yet. But it renders basically as you'd expect (I built it locally).

tmjd (approved):
LGTM

PR #1174, opened 2021-02-19
Adding checks for export-logs

asincu:
## Description

Add a license check for feature export - logs before setting up sending of data to 3rd party systems.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/common/common.go
+++ b/pkg/common/common.go
@@ -21,10 +21,10 @@ const (
 
 	TigeraPrometheusNamespace = "tigera-prometheus"
 
-	// Compliance feature name
+	// ComplianceFeature name
 	ComplianceFeature = "compliance-reports"
-	// Threat defence feature name
+	// ThreatDefenseFeature feature name
 	ThreatDefenseFeature = "threat-defense"
-	// Export logs to 3rd party systems feature name
-	ExportLogs = "export-logs"
+	// ExportLogsFeature to 3rd party systems feature name
+	ExportLogsFeature = "export-logs"
 )
--- a/pkg/common/common.go
+++ b/pkg/common/common.go
@@ -23,6 +23,8 @@ const (
 
 	// Compliance feature name
 	ComplianceFeature = "compliance-reports"
-	// Threat defene feature name
+	// Threat defence feature name
 	ThreatDefenseFeature = "threat-defense"
+	// Export logs to 3rd party systems feature name
+	ExportLogs = "export-logs"
 )
--- a/pkg/controller/compliance/compliance_controller.go
+++ b/pkg/controller/compliance/compliance_controller.go
@@ -369,9 +369,10 @@ func (r *ReconcileCompliance) Reconcile(ctx context.Context, request reconcile.R
 	}
 
 	reqLogger.V(3).Info("rendering components")
+	var hasNoLicense = !utils.IsFeatureActive(license, common.ComplianceFeature)
 	openshift := r.provider == operatorv1.ProviderOpenShift
 	// Render the desired objects from the CRD and create or update them.
-	component, err := render.Compliance(esSecrets, managerInternalTLSSecret, network, complianceServerCertSecret, esClusterConfig, pullSecrets, openshift, managementCluster, managementClusterConnection, dexCfg, r.clusterDomain)
+	component, err := render.Compliance(esSecrets, managerInternalTLSSecret, network, complianceServerCertSecret, esClusterConfig, pullSecrets, openshift, managementCluster, managementClusterConnection, dexCfg, r.clusterDomain, hasNoLicense)
 	if err != nil {
 		log.Error(err, "error rendering Compliance")
 		r.status.SetDegraded("Error rendering Compliance", err.Error())
@@ -384,22 +385,15 @@ func (r *ReconcileCompliance) Reconcile(ctx context.Context, request reconcile.R
 		return reconcile.Result{}, err
 	}
 
-	if !utils.IsFeatureActive(license, common.ComplianceFeature) {
-		log.V(4).Info("Compliance is not activated as part of this license")
-		if r.status.IsAvailable() {
-			if err := handler.Delete(context.Background(), component, r.status); err != nil {
-				r.status.SetDegraded("Error deleting resource", err.Error())
-				return reconcile.Result{RequeueAfter: 30 * time.Second}, nil
-			}
-		}
-
-		r.status.SetDegraded("Feature is not active", "License does not support this feature")
-		return reconcile.Result{}, nil
+	if err := handler.CreateOrUpdateOrDelete(ctx, component, r.status); err != nil {
+		r.status.SetDegraded("Error creating / updating / deleting resource", err.Error())
+		return reconcile.Result{}, err
 	}
 
-	if err := handler.CreateOrUpdate(ctx, component, r.status); err != nil {
-		r.status.SetDegraded("Error creating / updating resource", err.Error())
-		return reconcile.Result{}, err
+	if hasNoLicense {
+		log.V(4).Info("Compliance is not activated as part of this license")
+		r.status.SetDegraded("Feature is not active", "License does not support this feature")
+		return reconcile.Result{}, nil
 	}
 
 	// Clear the degraded bit if we've reached this far.
--- a/pkg/controller/compliance/compliance_controller.go
+++ b/pkg/controller/compliance/compliance_controller.go
@@ -22,6 +22,7 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/kubernetes"
 
 	operatorv1 "github.com/tigera/operator/api/v1"
 	"github.com/tigera/operator/pkg/common"
@@ -64,7 +65,13 @@ func Add(mgr manager.Manager, opts options.AddOptions) error {
 		return err
 	}
 
-	go utils.WaitToAddLicenseKeyWatch(controller, reconciler.client, log, reconciler.ready)
+	k8sClient, err := kubernetes.NewForConfig(mgr.GetConfig())
+	if err != nil {
+		log.Error(err, "Failed to establish a connection to k8s")
+		return err
+	}
+
+	go utils.WaitToAddLicenseKeyWatch(controller, k8sClient, log, reconciler.ready)
 
 	return add(mgr, controller)
 }
@@ -78,7 +85,6 @@ func newReconciler(mgr manager.Manager, provider operatorv1.Provider, clusterDom
 		status:        status.New(mgr.GetClient(), "compliance"),
 		clusterDomain: clusterDomain,
 		ready:         make(chan bool),
-		wg:            sync.WaitGroup{},
 	}
 	r.status.Run()
 	return r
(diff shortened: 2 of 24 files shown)


pkg/common/common.go
```diff
@@ -23,6 +23,8 @@ const (
 
 	// Compliance feature name
 	ComplianceFeature = "compliance-reports"
-	// Threat defene feature name
+	// Threat defence feature name
```
rene-dekker, line 26: I'd  go w/ `defense`, since we use the US spelling elsewhere mostly.

rene-dekker (approved):
lgtm

tmjd (commented):
As discussed in slack, I think the issue with this is that if the `export-logs` feature goes from enabled to disabled the feature change would not be detected since there is no watch on the License resource and no requeue of the reconcile.

pkg/controller/compliance/compliance_controller.go
```diff
@@ -187,6 +200,25 @@ func (r *ReconcileCompliance) Reconcile(ctx context.Context, request reconcile.R
 		return reconcile.Result{}, err
 	}
 
+	if !r.hasLicenseWatch {
+		r.wg.Add(1)
+
+		go func() {
+			defer close(r.ready)
+			for {
```
caseydavenport, line 208: Why do you need to run this in a goroutine if we're going to wait for it to complete in the Reconcile function anyway? Couldn't we just remove the `go func() {}` wrapper and just wait for the license watch to complete?

I was originally imagining that we wouldn't block Reconcile(), though - rather, we'd just have a check that would set degraded and return if not ready, and another goroutine which would establish the watch.

pkg/controller/utils/utils.go
```diff
@@ -114,6 +115,34 @@ func AddServiceWatch(c controller.Controller, name, namespace string) error {
...
+	lic := &v3.LicenseKey{
+		TypeMeta: metav1.TypeMeta{Kind: "LicenseKey"},
+	}
+	return c.Watch(&source.Kind{Type: lic}, &handler.EnqueueRequestForObject{})
+}
+
+// WaitToAddLicenseKeyWatch will check if the API server is available and if so, it will add a watch for LicenseKey
+// The completion of this operation will be signaled on a ready channel
+func WaitToAddLicenseKeyWatch(controller controller.Controller, client client.Client, log logr.Logger, ready chan bool) {
+	ticker := time.NewTicker(10 * time.Minute)
```
caseydavenport, line 128: This is a long time to wait - if we miss the initial watch on this couldn't it delay rollout of some components an additional 10 minutes?

Is there a way we could be more responsive without doing tons of GET requests? Maybe by watching the apiserver CR status?

pkg/controller/compliance/compliance_controller.go
caseydavenport: I guess the problem is that the interface for a Reconciler doesn't expose a way to mark as ready, and so we need to do the "marking as ready" part within the Reconciler itself?

I think we could still launch this goroutine at the same time we launch the `WaitToAddLicenseKeyWatch` routine, though.It could mark one or more ready bits on the reconciler.

And then, the `Reconcile()` function just needs to do something like this:

```go
if reason, ready := r.IsReady(); !ready {
  r.SetDegraded(reason)
  return
}
```

pkg/controller/compliance/compliance_controller.go
```diff
@@ -356,6 +392,19 @@ func (r *ReconcileCompliance) Reconcile(ctx context.Context, request reconcile.R
 		return reconcile.Result{}, err
 	}
 
+	if !utils.IsFeatureActive(license, common.ComplianceFeature) {
+		log.V(4).Info("Compliance is not activated as part of this license")
+		if r.status.IsAvailable() {
+			if err := handler.Delete(context.Background(), component, r.status); err != nil {
```
caseydavenport, line 398: I think we already have a mechanism to delete objects that aren't needed.

Here's an example: https://github.com/tigera/operator/blob/master/pkg/render/kube-controllers.go#L156

Would it work for this to have something like this in render?

```go
func Objects() ([]client.Object, []client.Object) {
. . .

  if licenseActive {
    // Create everything
    return objs, nil
  } else {
    // Delete everything.
    return nil, objs
  }
}
```

pkg/controller/utils/component.go
```diff
@@ -192,6 +193,52 @@ func (c componentHandler) CreateOrUpdate(ctx context.Context, component render.C
 	return nil
 }
 
+func (c componentHandler) Delete(ctx context.Context, component render.Component, status status.StatusManager) error {
+	// Before deleting the component, make sure that it is ready. This provides a hook to do
+	// dependency checking for the component.
+	cmpLog := c.log.WithValues("component", reflect.TypeOf(component))
+	cmpLog.V(2).Info("Checking if component is ready")
+	if !component.Ready() {
```
caseydavenport, line 201: I think there might be some nuance here that I'm not understanding about this function vs. the `CreateOrUpdate` function above (besides the fact that CreateOrUpdate should probably be called CreateOrUpdateOrDelete)

pkg/controller/utils/utils.go
```diff
@@ -114,6 +116,41 @@ func AddServiceWatch(c controller.Controller, name, namespace string) error {
...
+		select {
+		case <-ticker.C:
+			duration = duration * 2
+			if duration >= maxDuration {
+				duration = maxDuration
+			}
+			ticker.Reset(duration)
+			if areCalicoAPIsReady(client) {
+				err := addLicenseWatch(controller)
+				if err != nil {
```
caseydavenport, line 143: I wonder if we want a way to propagate this error back to TigeraStatus? Probably not required but might be a useful enhancement.

pkg/render/fluentd.go
```diff
@@ -497,7 +500,7 @@ func (c *fluentdComponent) envvars() []corev1.EnvVar {
 			)
 		}
 		syslog := c.lc.Spec.AdditionalStores.Syslog
-		if syslog != nil {
+		if c.exportLogsToAdditionalStorages && syslog != nil {
```
caseydavenport, line 503: Shouldn't it be an error if `exportLogcsToAdditionalStores` is false and `syslog != nil`? 

I think this means that the user intent is not actionable, and so rather than silently ignore it we should shout loudly.

pkg/controller/logcollector/logcollector_controller.go
```diff
@@ -236,8 +281,9 @@ func (r *ReconcileLogCollector) Reconcile(ctx context.Context, request reconcile
 		return reconcile.Result{}, err
 	}
 
+	var exportLogs = utils.IsFeatureActive(license, common.ExportLogsFeature)
 	var s3Credential *render.S3Credential
-	if instance.Spec.AdditionalStores != nil {
+	if exportLogs && instance.Spec.AdditionalStores != nil {
```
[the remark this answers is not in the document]
asincu: This one is a bit tricky since the component is installed, bu we limit the ability to export logs to a 3rd party system based on the type of license used.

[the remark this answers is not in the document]
caseydavenport: But the concept is basically "You asked for exporting to this external system, but your license doesn't allow for that" - so that should show up as a configuration error.

pkg/controller/compliance/compliance_controller.go
asincu: From what I noticed, `objectsToDelete` is nil and not populated by the renderer (for compliance and intrusion detection). I think that we should delete all the objects we are creating.

caseydavenport: Yes, exactly - I am suggesting that we populate that list with all of the rendered objects rather than leave it nil.

pkg/controller/utils/component.go
asincu: Delete returns an error if the component is not installed. I added a check to see if the component is marked as ready (installation was successful), then I can delete objects that reside within the tigera-*<component>* namespace.

pkg/render/fluentd.go
asincu: I think the error should be in the controller in this case and set the status as degraded.

caseydavenport: correct, but in that case we don't need this additional check here and probably don't need to pass an additional boolean to the render function?

i.e., the controller determines if the given intent is actionable. If it is, it proceeds to render. If not actionable, render is not called.

asincu: I would agree, but I see in render we do not use that way. Seems we are checking the installation spec instead of the object created in the controller.

https://github.com/tigera/operator/blob/master/pkg/render/fluentd.go#L470
https://github.com/tigera/operator/blob/master/pkg/render/fluentd.go#L499-L500
https://github.com/tigera/operator/blob/master/pkg/render/fluentd.go#L550-L551

pkg/controller/compliance/compliance_controller.go
asincu: In compliance, we are not actually respecting this contract. In MCM, we actually delete compliance server, but keep the rest of the components. I actually need to delete all created objects.

https://github.com/tigera/operator/blob/master/pkg/render/compliance.go#L200-L214.

pkg/controller/logcollector/logcollector_controller.go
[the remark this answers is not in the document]
asincu: Setting the status to degraded if you specify additional storage and do not have the correct license,

pkg/controller/compliance/compliance_controller.go
asincu: Update this flow as per offline discussions.

pkg/render/fluentd.go
asincu: I added a check in the controller. We should not get into this state.

pkg/controller/utils/component.go
asincu: Does not apply anymore.

pkg/controller/intrusiondetection/intrusiondetection_controller.go
```diff
@@ -181,20 +212,18 @@ func (r *ReconcileIntrusionDetection) Reconcile(ctx context.Context, request rec
 		return reconcile.Result{}, err
 	}
 
+	if !r.IsReady() {
+		return reconcile.Result{RequeueAfter: 10 * time.Second}, nil
```
caseydavenport, line 216: Do we want to set degraded here?

pkg/controller/logcollector/logcollector_controller.go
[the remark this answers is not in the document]
caseydavenport: I think we can remove the `exportLogs` part of the test here, because we've already checked it in the if statement above.

caseydavenport (commented):
This LGTM - I think @tmjd should give the final approval though.

tmjd (commented):
Looks good, I feel like adding IsReady and MarkAsReady functions to those components is a little confusing since they aren't really functions that apply to the whole component. I wonder if we should have a ReadyFlag struct that has a bool and RWMutex, then the components can have an instance of it. That ReadyFlag could be passed in to the newReconciler and then to the WaitToAddLicenseKeyWatch.
I know this PR has been in progress for a while now, so if this has been considered just let me know.

tmjd (approved):
LGTM
We need to make sure the commits are squashed before or during merge.

tmjd (approved):

PR #1180, opened 2021-02-23
Disable IDS job installer within managed clusters

vberezny:
## Description
This PR disables the intrusion detection job installer in managed clusters.

## Issue
https://tigera.atlassian.net/browse/SAAS-1328

## Related PRs
Intrusion Detection: https://github.com/tigera/intrusion-detection/pull/295
Kube Controllers: https://github.com/tigera/kube-controllers-private/pull/242

## Testing
Validated in Standalone, Management and Managed clusters.

PR #1182, opened 2021-02-25
Automated cherry pick of #1180: Disable IDS job installer within managed clusters

vberezny:
Cherry pick of #1180 on release-v1.14.

#1180: Disable IDS job installer within managed clusters


tmjd:
I've added the "hold merge" label since this pick should be held until we need to do an enterprise v3.5.1 release as there are component updates that need to go along with the change.

stevegaossou:
Just to add a note to this, the related component is intrusion-detection. We would need to put in a pick for that component to move the changes into v3.5 release branch before this PR can be merged. 

There is also a separate operator fix that would need to be picked into release-v1.14: 
https://github.com/tigera/operator/pull/1201/files

PR #1186, opened 2021-02-26
Set env variables for ES for Standalone cluster

asincu:
## Description

KubeControllers need the following env variables in order to connect to ES. This configuration is needed for a Standalone and Management cluster.

- `ELASTIC_HOST`
- `ELASTIC_PORT`
- `ELASTIC_USERNAME` from secret `tigera-ee-kube-controllers-elasticsearch-access`
- `ELASTIC_PASSWORD` from secret `tigera-ee-kube-controllers-elasticsearch-access`
- `ELASTIC_CA` using cert from secret `tigera-secure-es-http-certs-public` at path `/etc/ssl/elastic/ca.pem`

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/kube-controllers.go
+++ b/pkg/render/kube-controllers.go
@@ -134,7 +134,7 @@ func (c *kubeControllersComponent) Objects() ([]client.Object, []client.Object) 
 			secret.CopyToNamespace(common.CalicoNamespace, c.elasticsearchSecret)...)...)
 	}
 
-	if c.managementCluster != nil && c.elasticsearchUserSecret != nil {
+	if !c.isManagedCluster() && c.elasticsearchUserSecret != nil {
 		kubeControllerObjects = append(kubeControllerObjects, secret.ToRuntimeObjects(c.elasticsearchUserSecret)...)
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/render/kube-controllers.go
```diff
@@ -134,7 +134,7 @@ func (c *kubeControllersComponent) Objects() ([]client.Object, []client.Object)
 			secret.CopyToNamespace(common.CalicoNamespace, c.elasticsearchSecret)...)...)
 	}
 
-	if c.managementCluster != nil && c.elasticsearchUserSecret != nil {
+	if !c.isManagedCluster() && c.elasticsearchUserSecret != nil {
```
rene-dekker, line 137: If the contents of the secrets change, is there an annotation to make sure the pod restarts?

asincu: @rene-dekker Yup, we should have watched added for them.

Brian-McM (approved):

pkg/render/kube-controllers.go
tmjd: In addition to the watch we also annotate the podTemplate with a hash for this secret so that ensures it will be restarted in the case of change.

tmjd:
Looks like a test failure.

asincu:
@tmjd Yes,the calico installation was failing.  We also need to check for the logstorage to be present before mounting the volume.

PR #1192, opened 2021-03-01
Create Elasticsearch client on the fly instead of caching it

Brian-McM:
This commit changes logstorage to create the Elasticsearch client for every reconcile. This has been changed because we could be reconciling Elasticsearch credentials being recreated, in which case the cached client wouldn't work.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/utils/elasticsearch.go
+++ b/pkg/controller/utils/elasticsearch.go
@@ -22,7 +22,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
-	"sync"
 	"time"
 
 	"github.com/olivere/elastic/v7"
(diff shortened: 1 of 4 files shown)


pkg/controller/utils/elasticsearch.go
```diff
@@ -113,69 +112,55 @@ func GetElasticsearchClusterConfig(ctx context.Context, cli client.Client) (*rel
...
+			break
 		}
+		log.Error(err, "Elastic connect failed, retrying")
+		time.Sleep(retryInterval)
+	}
 
-		var eserr error
-		var esClient *elastic.Client
-		for i := 0; i < ElasticConnRetries; i++ {
+	return &esClient{client: esCli}, nil
```
Suraiya-Hameed, line 157: I think this case needs to be handled - say if the `esCli` is never created after `ElasticConnRetries`, we will have to return error.

Brian-McM: oh right woops thanks for catching that

Brian-McM: fixed

Suraiya-Hameed (approved):

tmjd (approved):
LGTM

PR #1194, opened 2021-03-01
Validate only operator-managed certificates

lmm:
## Description

Before v1.14.0, we did not validate custom TLS cert DNS names. In v1.14.0 we changed the DNS names in certain certs, but also imposed a requirement of custom TLS certs to contain the full range of service DNS names.

This PR changes the validation of cert DNS names to only validate operator-managed certificates. Custom TLS certs are just passed through, as with versions < v1.14.0.

This is the first step; the next step is to add back in the validation of custom TLS certs. In order to do so, we'll need to add a way to detect whether we're running hybrid clusters - the logstorage controller does a simple watch on the nodes and checks if we have Windows nodes or not. We'll need something that will work in other controllers too.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/utils/certs.go
+++ b/pkg/controller/utils/certs.go
@@ -21,7 +21,6 @@ import (
 	"errors"
 	"fmt"
 	"regexp"
-	"strings"
 	"time"
 
 	rmeta "github.com/tigera/operator/pkg/render/common/meta"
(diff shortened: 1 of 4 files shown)


pkg/controller/utils/certs.go
```diff
@@ -74,28 +73,30 @@ func EnsureCertificateSecret(secretName string, secret *corev1.Secret, keyName s
 		)
 	}
 
+	operatorManaged, err := IsOperatorManaged(secret, certName)
```
rene-dekker, line 76: I understand from the other code, that if a cert is operator managed and invalid, we want to throw it away. That lgtm.

But what I do not understand is that if it is not operator managed we skip checking for dns names. Won't our components run into ussues if they run at  https://tigera-server.tigera-ns.svc.cluster.local, but don't have that subject (alt) name?

lmm: We talked about this and I think you're right: I'll add some logging and set the controller degraded if the user-provided certs are not valid.

lmm: Actually... the reason for reverting the cert DNS validation for custom certs is because we don't have a good way to detect that we have a hybrid cluster.

rene-dekker (approved):
lgtm

PR #1201, opened 2021-03-02
Disable waiting on IDS job secret in managed clusters

vberezny:
## Description
[Fix] This PR disables waiting on the IDS job secret in managed clusters. This is done to reflect the changes that disable the creation of the secret in managed clusters.

### DIFF
--- a/pkg/controller/intrusiondetection/intrusiondetection_controller_test.go
+++ b/pkg/controller/intrusiondetection/intrusiondetection_controller_test.go
@@ -126,10 +126,6 @@ var _ = Describe("IntrusionDetection controller tests", func() {
 				Namespace: "tigera-operator"}})).NotTo(HaveOccurred())
 		Expect(c.Create(ctx, &corev1.Secret{
 			ObjectMeta: metav1.ObjectMeta{
-				Name:      render.ElasticsearchIntrusionDetectionJobUserSecret,
-				Namespace: "tigera-operator"}})).NotTo(HaveOccurred())
-		Expect(c.Create(ctx, &corev1.Secret{
-			ObjectMeta: metav1.ObjectMeta{
 				Name:      render.ElasticsearchADJobUserSecret,
 				Namespace: "tigera-operator"}})).NotTo(HaveOccurred())
 		Expect(c.Create(ctx, &corev1.Secret{
@@ -149,6 +145,13 @@ var _ = Describe("IntrusionDetection controller tests", func() {
 	})
 
 	Context("image reconciliation", func() {
+		BeforeEach(func() {
+			Expect(c.Create(ctx, &corev1.Secret{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      render.ElasticsearchIntrusionDetectionJobUserSecret,
+					Namespace: "tigera-operator"}})).NotTo(HaveOccurred())
+		})
+
 		It("should use builtin images", func() {
 			_, err := r.Reconcile(ctx, reconcile.Request{})
 			Expect(err).ShouldNot(HaveOccurred())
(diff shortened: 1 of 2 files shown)


pkg/controller/intrusiondetection/intrusiondetection_controller_test.go
```diff
@@ -274,6 +274,34 @@ var _ = Describe("IntrusionDetection controller tests", func() {
...
+		It("should wait on tigera-ee-installer-elasticsearch-access secret when in a management cluster", func() {
+			Expect(c.Create(ctx, &operatorv1.ManagementCluster{
+				ObjectMeta: metav1.ObjectMeta{Name: "tigera-secure"},
+				Spec: operatorv1.ManagementClusterSpec{
+					Address: "127.0.0.1:12345",
+				},
+			})).ToNot(HaveOccurred())
+
+			result, err := r.Reconcile(ctx, reconcile.Request{})
+			Expect(err).ShouldNot(HaveOccurred())
```
Brian-McM, line 300: Shouldn't we expect an error to be returned if it's waiting for a secret?

Brian-McM (changes requested):

pkg/controller/intrusiondetection/intrusiondetection_controller_test.go
vberezny: I'll need to rethink these tests, realized the first one doesn't fail when it is forced to check for the IDS job secret.

vberezny: I updated the tests to make sure the change is properly covered. I am now deleting the IDS job secret and expecting the management cluster to call `status.SetDegraded` in response to not finding it in `utils.ElasticSearch`.

pkg/controller/intrusiondetection/intrusiondetection_controller_test.go
```diff
@@ -274,6 +276,58 @@ var _ = Describe("IntrusionDetection controller tests", func() {
...
+			// Delete the secret to ensure that utils.ElasticSearch correctly looks for secrets relevant to managed
+			// clusters only and doesn't attempt to find the IDS job secret.
+			idsSecret := &corev1.Secret{
+				TypeMeta: metav1.TypeMeta{Kind: "Secret", APIVersion: "v1"},
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      render.ElasticsearchIntrusionDetectionJobUserSecret,
+					Namespace: rmeta.OperatorNamespace(),
+				},
+			}
+			err := c.Delete(ctx, idsSecret)
```
Brian-McM, line 297: I think instead of doing this you should change the before each for the context above this context to not add the installer secret and put in a before each for the sibling contexts that need.

Brian-McM (changes requested):

pkg/controller/intrusiondetection/intrusiondetection_controller_test.go
vberezny: Done

pkg/controller/intrusiondetection/intrusiondetection_controller_test.go
```diff
@@ -74,6 +74,7 @@ var _ = Describe("IntrusionDetection controller tests", func() {
 		mockStatus.On("IsAvailable").Return(true)
 		mockStatus.On("OnCRFound").Return()
 		mockStatus.On("ClearDegraded")
+		mockStatus.On("SetDegraded", mock.Anything, mock.Anything).Return()
```
Brian-McM, line 77: Doesn't this mean that tests the previously weren't setting a degraded status are now expected to, i.e. all the tests in the context "image reconciliation"? That seems like we shouldn't be changing that expectation...

vberezny: You're right, I'll throw them in a before each only for the tests that need `SetDegraded`

pkg/controller/intrusiondetection/intrusiondetection_controller_test.go
```diff
@@ -274,6 +277,39 @@ var _ = Describe("IntrusionDetection controller tests", func() {
...
+				ObjectMeta: metav1.ObjectMeta{Name: "tigera-secure"},
+				Spec: operatorv1.ManagementClusterSpec{
+					Address: "127.0.0.1:12345",
+				},
+			})).ToNot(HaveOccurred())
+
+			_, err := r.Reconcile(ctx, reconcile.Request{})
+			Expect(err).ShouldNot(HaveOccurred())
+			// The missing secret should force utils.ElasticSearch to return a NotFound error which triggers r.status.SetDegraded.
+			Expect(mockStatus.AssertNumberOfCalls(nil, "SetDegraded", 1)).To(BeTrue())
```
Brian-McM, line 309: Instead of doing this it would probably be better to add an "AfterEach" in the top level context and use the AssertExpectations. That means that if you set "On" for a mock and that function wasn't called it would error out. I've been doing this for other projects, I think this was one of the first projects I started mocking with so I think I just forgot.

Brian-McM (approved):

PR #1204, opened 2021-03-03
Handle case where there are more Typhas than there are nodes

caseydavenport:
## Description

When migrating from manifest -> operator, sometimes there are a combined
total number of Typhas greater than the cluster can handle (since max 1
typha per node can run).

Previously, we just failed in this scenario. This PR attempts to let us
progress by doing a rolling update from one Typha deployment to the
other using nodeSelectors to ensure that a max of one typha is run on
each node.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


pkg/controller/migration/namespace_migration.go
```diff
@@ -384,19 +447,24 @@ func (m *CoreNamespaceMigration) ensureKubeSysNodeDaemonSetHasNodeSelectorAndIsR
...
 		}
-		if ds.Status.DesiredNumberScheduled != ds.Status.NumberReady || ds.Status.ObservedGeneration != ds.ObjectMeta.Generation {
-			return false, fmt.Errorf("not all pods are ready yet: %d/%d", ds.Status.NumberReady, ds.Status.DesiredNumberScheduled)
+		if ds.Status.ObservedGeneration != ds.ObjectMeta.Generation {
+			log.Info(fmt.Sprintf("waiting for observed generation (%d) to match object generation (%d)", ds.Status.ObservedGeneration, ds.ObjectMeta.Generation))
+			return false, nil
 		}
+		if ds.Status.DesiredNumberScheduled != ds.Status.NumberReady {
+			log.Info(fmt.Sprintf("waiting for kube-system/calico-node to have %d replicas, currently at %d", ds.Status.DesiredNumberScheduled, ds.Status.NumberReady))
+			return false, nil
```
caseydavenport, line 456: We used to return an error here, which would go all the way back to the Reconcile() retry. I switched it to return false/nil to simply retry in the PollImmediate - is there a reason this is no good? It means less work overall, and fewer scary looking log messages during an "expected" transition.

pkg/controller/migration/namespace_migration.go
```diff
@@ -21,8 +21,8 @@ import (
 	"strings"
 	"time"
 
-	"github.com/cloudflare/cfssl/log"
```
caseydavenport, line 24: We were using two logging libraries - switched it to one.

pkg/controller/migration/namespace_migration.go
```diff
@@ -220,18 +220,22 @@ func (m *CoreNamespaceMigration) Run(ctx context.Context, log logr.Logger) error
...
 	if err := m.labelUnmigratedNodes(ctx); err != nil {
 		return fmt.Errorf("failed to label unmigrated nodes: %s", err.Error())
 	}
 	log.V(1).Info("All unmigrated nodes labeled")
-	if err := m.ensureKubeSysNodeDaemonSetHasNodeSelectorAndIsReady(ctx); err != nil {
+	if err := m.ensureKubeSysNodeDaemonSetHasNodeSelectorAndIsReady(ctx, log); err != nil {
 		return fmt.Errorf("the kube-system node DaemonSet is not ready with the updated nodeSelector: %s", err.Error())
 	}
 	log.V(1).Info("Node selector added to kube-system node DaemonSet")
+	if err := m.ensureTyphaRoom(ctx, log); err != nil {
```
caseydavenport, line 231: I still need to delete / disable any auto-scalers running against the old cluster before calling this function.

pkg/controller/utils/utils.go
```diff
@@ -297,23 +297,6 @@ func GetAuthentication(ctx context.Context, cli client.Client) (*operatorv1.Auth
 	return authentication, nil
 }
 
-// GetTyphaScaleCount will return the number of Typhas needed for the number of nodes.
-func GetExpectedTyphaScale(nodes int) int {
```
caseydavenport, line 301: I had to move this in order to resolve a circular dependency.

pkg/render/typha.go
```diff
@@ -60,24 +60,24 @@ func Typha(
...
-		installation:       installation,
-		typhaNodeTLS:       tnTLS,
-		amazonCloudInt:     aci,
-		namespaceMigration: migrationNeeded,
-		clusterDomain:      clusterDomain,
+		k8sServiceEp:    k8sServiceEp,
+		installation:    installation,
+		typhaNodeTLS:    tnTLS,
+		amazonCloudInt:  aci,
+		migrationNeeded: migrationNeeded,
```
caseydavenport, line 67: changed the field name to be consistent with pkg/render/node.go

pkg/controller/migration/namespace_migration.go
```diff
@@ -220,18 +220,22 @@ func (m *CoreNamespaceMigration) Run(ctx context.Context, log logr.Logger) error
...
+		return fmt.Errorf("unable to ensure room for enough typhas: %s", err.Error())
+	}
+	log.V(1).Info("Ensured room for Typha deployments")
+	if err := m.waitForOperatorTyphaDeploymentReady(ctx, log); err != nil {
```
tmjd, line 235: With the Operator Typha now being subject to the migration label, won't this always work (assuming no restarting/errors) because there will be no Operator Typha's running at this point.

pkg/controller/migration/namespace_migration.go
tmjd: I want to say it was so that if any new nodes showed up that they would get labeled but I'm not seeing why that would be necessary here.

tmjd (commented):
I'm wondering if we need an update or 2 to `migrateEachNode`.
* Either check if typha is ready as part of its checks before switching the label on the next node (at least one will be healthy if node is healthy).
* Do we need to trigger the typha auto-scaler to update as we flip the label on each node? (currently the update timer is 2 minutes)

I think things will be fine without updates to `migrateEachNode` but wanted to see what you thought.

pkg/controller/migration/convert/typha.go
```diff
@@ -30,13 +30,16 @@ func checkTypha(c *components, _ *operatorv1.Installation) error {
...
+	if nodeCount >= curReplicas+common.GetExpectedTyphaScale(nodeCount) {
 		return nil
 	}
 
-	return ErrIncompatibleCluster{
-		err: fmt.Sprintf("Not enough nodes available for typha deployment. Have %d nodes with %d typha replicas currently deployed and %d additional typhas are needed during migration",
-			nodeCount, curReplicas, utils.GetExpectedTyphaScale(nodeCount)),
-		component: ComponentTypha,
-	}
+	// CASEY: HACK: Allow Typha migration to continue even if we don't have enough nodes.
```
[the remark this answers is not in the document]
caseydavenport: whoops, this was a comment from early on to remind me to clean this up. I think we can just delete this entire handler now.

pkg/controller/installation/typha_autoscaler.go
```diff
@@ -213,6 +212,10 @@ func (t *typhaAutoscaler) getNodeCounts() (int, int, error) {
 		if n.Spec.Unschedulable {
 			continue
 		}
+		if n.GetObjectMeta().GetAnnotations()["projectcalico.org/operator-node-migration"] == "pre-operator" {
```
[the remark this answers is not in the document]
caseydavenport: I think we want the expected typhas to be "expected once the rollout is complete", and this function is only used by the auto-scaler to get the number of typhas needed _now_. Or maybe I'm misunderstanding?

pkg/controller/migration/namespace_migration.go
caseydavenport: In practice, I think one will get scheduled anyway before the auto-scaler kicks in, but otherwise yeah I think you're right it's not a super useful check here. It would be better to do this as part of migrateEachNode.

pkg/controller/installation/typha_autoscaler.go
```diff
@@ -157,15 +198,18 @@ func (t *typhaAutoscaler) autoscaleReplicas() error {
...
 	if linuxNodes < expectedReplicas {
 		return fmt.Errorf("not enough linux nodes to schedule typha pods on, require %d and have %d", expectedReplicas, linuxNodes)
 	}
 
-	err = t.updateReplicas(int32(expectedReplicas))
-
-	if err != nil && !apierrors.IsNotFound(err) {
-		return fmt.Errorf("could not scale Typha deployment: %w", err)
+	log.V(1).Info("Checking if we need to scale typha", "expectedReplicas", expectedReplicas, "currentReplicas", t.activeReplicas)
+	if int32(expectedReplicas) != t.activeReplicas {
```
caseydavenport, line 208: This check allows us to skip the Get() call within updateReplicas if we don't actually need to update the replicas. We'll spot when the number of replicas changes because we're instead watching the typha deployment.

pkg/controller/installation/typha_autoscaler.go
```diff
@@ -156,15 +198,18 @@ func (t *typhaAutoscaler) autoscaleReplicas() error {
 	if err != nil {
 		return fmt.Errorf("could not get number of nodes: %w", err)
 	}
+	log.V(1).Info("Number of nodes to consider for typha autoscaling", "all", allSchedulableNodes, "linux", linuxNodes)
```
tmjd, line 201: I'm assuming these are just for testing as getting this message every 10 seconds seems a bit verbose.

tmjd (commented):
Love the typha_autoscaler changes :heart:

pkg/controller/installation/typha_autoscaler.go
caseydavenport: yeah these were just test logs

pkg/controller/installation/typha_autoscaler.go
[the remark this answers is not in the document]
tmjd: Yeah ok I see that now. I think there are some cases where the migration can calculate more typhas than will really be deployed but I don't think there is an issue with that.

pkg/controller/migration/namespace_migration.go
```diff
@@ -456,6 +540,14 @@ func (m *CoreNamespaceMigration) migrateEachNode(ctx context.Context, log logr.L
 				log.WithValues("error", err).V(1).Info("Error checking for new healthy pods")
 				time.Sleep(10 * time.Second)
 			}
+
+			// Wait for the operator-managed Typha deployment to be ready.
+			if err := m.waitForOperatorTyphaDeploymentReady(ctx, log); err != nil {
+				return fmt.Errorf("failed to wait for operator typha deployment to be ready: %s", err.Error())
+			}
+			log.V(1).Info("calico-system/calico-typha is running with expected replica count after migrating node")
```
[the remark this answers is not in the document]
tmjd: If it would then I think you should increase the level for this one because I don't think for normal operation we need 2 different messages on each node.

tmjd (approved):
Only a couple small things so I'm approving. LGTM

pkg/controller/installation/typha_autoscaler_test.go
```diff
@@ -16,59 +16,119 @@ package installation
...
 	It("should get the correct number of nodes", func() {
 		n1 := createNode(c, "node1", map[string]string{"kubernetes.io/os": "linux"})
 		_ = createNode(c, "node2", map[string]string{"kubernetes.io/os": "linux"})
 
-		ta := newTyphaAutoscaler(c, statusManager)
-		schedulableNodes, linuxNodes, err := ta.getNodeCounts()
-		Expect(err).To(BeNil())
-		Expect(schedulableNodes).To(Equal(2))
-		Expect(linuxNodes).To(Equal(2))
+		ta := newTyphaAutoscaler(c, nlw, tlw, statusManager)
```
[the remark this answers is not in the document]
caseydavenport: not really, since we are pre-loading the data before calling `ta.start()`, although maybe this does apply to the second update. Either way, I didn't see this test taking very long. I'll look at it.

pkg/controller/migration/namespace_migration.go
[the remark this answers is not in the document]
caseydavenport: V(1) is not included in the default logging level

PR #1209, opened 2021-03-05
Fix validation with HostLocal IPAM

tmjd:
* Allow BGP to be disabled with HostLocal IPAM 
* Only default Encapsulation to IPIP when IPAMCalico is being used, otherwise default to No encapsulation (for IPv4).
* Validate encapsulation and BGP when not using Calico IPAM

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/installation/validation.go
+++ b/pkg/controller/installation/validation.go
@@ -130,22 +130,34 @@ func validateCustomResource(instance *operatorv1.Installation) error {
 			}
 
 			if instance.Spec.CNI.Type == operatorv1.PluginCalico {
-				// Verify the specified encapsulation type is valid.
-				switch v4pool.Encapsulation {
-				case operatorv1.EncapsulationIPIP, operatorv1.EncapsulationIPIPCrossSubnet:
-					// IPIP currently requires BGP to be running in order to program routes.
-					if instance.Spec.CalicoNetwork.BGP == nil || *instance.Spec.CalicoNetwork.BGP == operatorv1.BGPDisabled {
-						return fmt.Errorf("IPIP encapsulation requires that BGP is enabled")
+				switch instance.Spec.CNI.IPAM.Type {
+				case operatorv1.IPAMPluginCalico:
+					// Verify the specified encapsulation type is valid.
+					switch v4pool.Encapsulation {
+					case operatorv1.EncapsulationIPIP, operatorv1.EncapsulationIPIPCrossSubnet:
+						// IPIP currently requires BGP to be running in order to program routes.
+						if instance.Spec.CalicoNetwork.BGP == nil || *instance.Spec.CalicoNetwork.BGP == operatorv1.BGPDisabled {
+							return fmt.Errorf("IPIP encapsulation requires that BGP is enabled")
+						}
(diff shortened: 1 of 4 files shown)


pkg/controller/installation/validation.go
```diff
@@ -130,22 +130,37 @@ func validateCustomResource(instance *operatorv1.Installation) error {
...
+							v4pool.Encapsulation, strings.Join(operatorv1.EncapsulationTypesString, ","))
 					}
-				case operatorv1.EncapsulationVXLAN, operatorv1.EncapsulationVXLANCrossSubnet:
-				case operatorv1.EncapsulationNone:
-					// Unencapsulated currently requires BGP to be running in order to program routes.
-					if instance.Spec.CalicoNetwork.BGP == nil || *instance.Spec.CalicoNetwork.BGP == operatorv1.BGPDisabled {
-						return fmt.Errorf("Unencapsulated IP pools require that BGP is enabled")
+				} else {
+					// Verify the specified encapsulation type is None for non Calico IPAM.
+					if v4pool.Encapsulation != operatorv1.EncapsulationNone {
```
tmjd, line 153: What do you think about these 2 validations for Calico CNI and non-Calico IPAM?
I've got a slight concern that even though I do not think there is a use case we support that there might be a config out there that has set VXLAN with HostLocal or something like that.

caseydavenport: The Calico code itself doesn't really work for VXLAN with HostLocal - we haven't implemented allocation of tunnel addresses (i.e., it would work for some traffic paths but not pod<->host).

pkg/controller/installation/validation.go
```diff
@@ -130,22 +130,37 @@ func validateCustomResource(instance *operatorv1.Installation) error {
...
+						return fmt.Errorf("%s is invalid for ipPool.encapsulation with %s CNI and %s IPAM",
+							v4pool.Encapsulation,
+							instance.Spec.CNI.Type,
+							instance.Spec.CNI.IPAM.Type)
+					}
+					if instance.Spec.CalicoNetwork.BGP != nil && *instance.Spec.CalicoNetwork.BGP != operatorv1.BGPDisabled {
```
caseydavenport, line 159: This is the validation that I think is not correct - now that I think about it, I am certain that at least one major user is running in this mode (Calico CNI, host-local IPAM, and BGP to distribute routes)

caseydavenport (commented):
One minor comment

pkg/controller/installation/validation.go
tmjd: What about Encapsulation? If we're supporting BGP with HostLocal then shouldn't we support any encapsulation?

caseydavenport: IPIP should work, VXLAN won't without a bit more coding.

tmjd:
I've dropped the defaulting encapsulation change I had included, see my last commit for that specific change. I think maybe a bigger change should be done for the defaulting but isn't really related to the validation change needed.

caseydavenport (approved):
LGTM, but there's a test failure

PR #1211, opened 2021-03-05
Add Elasticsearch / Fluentd metrics monitoring

Brian-McM:
This commit does two things:
1) Adds the metrics port for fluentd so we can set up a Prometheus Pod Monitor to scrape metrics from fluentd
2) Adds the elasticsearch-metrics deployment / service to pull metrics from Elasticsearch and present them to Prometheus for scraping

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/logstorage/logstorage_controller.go
+++ b/pkg/controller/logstorage/logstorage_controller.go
@@ -33,6 +33,7 @@ import (
 	relasticsearch "github.com/tigera/operator/pkg/render/common/elasticsearch"
 	rmeta "github.com/tigera/operator/pkg/render/common/meta"
 	rsecret "github.com/tigera/operator/pkg/render/common/secret"
+	"github.com/tigera/operator/pkg/render/logstorage/esmetrics"
 
 	apps "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
@@ -409,13 +410,13 @@ func (r *ReconcileLogStorage) Reconcile(ctx context.Context, request reconcile.R
 		return reconcile.Result{}, err
 	}
 
-	var elasticsearchSecrets, kibanaSecrets, curatorSecrets []*corev1.Secret
+	var esCertSecret, esCertSecretESCopy, esPubCertSecret, esAdminUserSecret *corev1.Secret
+	var kibanaSecrets, curatorSecrets []*corev1.Secret
 	var clusterConfig *relasticsearch.ClusterConfig
 	var esLicenseType render.ElasticsearchLicenseType
 	applyTrial := false
 
 	if managementClusterConnection == nil {
-
 		var flowShards = calculateFlowShards(ls.Spec.Nodes, defaultElasticsearchShards)
 		clusterConfig = relasticsearch.NewClusterConfig(render.DefaultElasticsearchClusterName, ls.Replicas(), defaultElasticsearchShards, flowShards)
 
@@ -433,7 +434,13 @@ func (r *ReconcileLogStorage) Reconcile(ctx context.Context, request reconcile.R
 			return reconcile.Result{}, nil
 		}
 
-		if elasticsearchSecrets, err = r.elasticsearchSecrets(ctx); err != nil {
+		// Get the secret - might be nil
+		esAdminUserSecret, err = utils.GetSecret(ctx, r.client, render.ElasticsearchAdminUserSecret, render.ElasticsearchNamespace)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+
+		if esCertSecretESCopy, esPubCertSecret, esAdminUserSecret, err = r.getElasticsearchCertificateSecrets(ctx); err != nil {
 			reqLogger.Error(err, err.Error())
 			r.status.SetDegraded("Failed to create elasticsearch secrets", err.Error())
 			return reconcile.Result{}, err
@@ -530,7 +537,7 @@ func (r *ReconcileLogStorage) Reconcile(ctx context.Context, request reconcile.R
 		elasticsearch,
 		kibana,
 		clusterConfig,
-		elasticsearchSecrets,
+		[]*corev1.Secret{esCertSecret, esCertSecretESCopy, esPubCertSecret, esAdminUserSecret},
 		kibanaSecrets,
 		pullSecrets,
 		r.provider,
@@ -595,6 +602,33 @@ func (r *ReconcileLogStorage) Reconcile(ctx context.Context, request reconcile.R
 				return reconcile.Result{}, err
 			}
 		}
+
+		esMetricsSecret, err := utils.GetSecret(context.Background(), r.client, esmetrics.ElasticsearchMetricsSecret, rmeta.OperatorNamespace())
+		if err != nil {
+			r.status.SetDegraded("Failed to retrieve Elasticsearch metrics user secret.", err.Error())
+			return reconcile.Result{}, err
+		} else if esMetricsSecret == nil {
+			r.status.SetDegraded("Waiting for elasticsearch metrics secrets to become available", "")
+			return reconcile.Result{}, nil
+		}
+
+		if esPubCertSecret == nil {
+			r.status.SetDegraded("Waiting for elasticsearch public cert secret to become available", "")
+			return reconcile.Result{}, nil
+		}
+
+		esMetricsComponent := esmetrics.ElasticsearchMetrics(install, clusterConfig, esMetricsSecret, esPubCertSecret, r.clusterDomain)
+		if err = imageset.ApplyImageSet(ctx, r.client, variant, esMetricsComponent); err != nil {
+			reqLogger.Error(err, "Error with images from ImageSet")
+			r.status.SetDegraded("Error with images from ImageSet", err.Error())
+			return reconcile.Result{}, err
+		}
+
+		if err := hdler.CreateOrUpdate(ctx, esMetricsComponent, r.status); err != nil {
+			reqLogger.Error(err, err.Error())
+			r.status.SetDegraded("Error creating / updating resource", err.Error())
+			return reconcile.Result{}, err
+		}
 	}
 
 	r.status.ClearDegraded()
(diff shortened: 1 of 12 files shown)


pkg/render/logstorage/components/elasticsearch_metrics.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+
+	if len(errMsgs) != 0 {
+		return fmt.Errorf(strings.Join(errMsgs, ","))
+	}
+	return nil
+}
+
+func (e *elasticsearchMetrics) Objects() (objsToCreate, objsToDelete []client.Object) {
+	toCreate := []client.Object{e.metricsService(), e.metricsDeployment()}
```
tmjd, line 68: I think we should have these objects in the order they should be created.
I think that should be:
1. Secret
2. Deployment
3. Service

Brian-McM: Makes sense, I'll do that.

pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -617,6 +651,50 @@ func (r *ReconcileLogStorage) deleteInvalidECKManagedPublicCertSecret(ctx contex
...
+				}
+			}
+		}
+
+		pubSecret = rsecret.CopyToNamespace(rmeta.OperatorNamespace(), pubSecret)[0]
+	}
+
+	return certSecret, certSecretESCopy, pubSecret, nil
+}
+
```
tmjd, line 697: Is the following function still in use since you removed the call to it at L436?

pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -617,6 +651,50 @@ func (r *ReconcileLogStorage) deleteInvalidECKManagedPublicCertSecret(ctx contex
...
+	certSecretESCopy := rsecret.CopyToNamespace(render.ElasticsearchNamespace, certSecret)[0]
+
+	// Get the pub secret - might be nil
+	pubSecret, err := utils.GetSecret(ctx, r.client, relasticsearch.PublicCertSecret, render.ElasticsearchNamespace)
+	if err != nil {
+		return nil, nil, nil, err
+	}
+
+	if pubSecret != nil {
+		operatorManaged, err := utils.IsOperatorManaged(certSecret, "tls.crt")
```
tmjd, line 678: Should this be using `pubSecret` instead of `certSecret`

pkg/render/fluentd.go
```diff
@@ -455,6 +455,10 @@ func (c *fluentdComponent) container() corev1.Container {
 		StartupProbe:    c.startup(),
 		LivenessProbe:   c.liveness(),
 		ReadinessProbe:  c.readiness(),
+		Ports: []corev1.ContainerPort{{
```
tmjd, line 458: Is there or should there be a service that is targeting this?

pkg/render/logstorage/components/elasticsearch_metrics.go
```diff
@@ -0,0 +1,151 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	operatorv1 "github.com/tigera/operator/api/v1"
+	"github.com/tigera/operator/pkg/components"
+	"github.com/tigera/operator/pkg/render"
+	rmeta "github.com/tigera/operator/pkg/render/common/meta"
+	corev1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"k8s.io/apimachinery/pkg/util/intstr"
```
tmjd, line 21: I've been trying to organize our imports into the 3 categories:
1. The golang imports (fmt, strings)
2. The non-operator imports (basically everything that isn't github.com/tigera/operator)
3. The github.com/tiger/operator imports

pkg/render/logstorage/components/elasticsearch_metrics.go
```diff
@@ -0,0 +1,151 @@
...
+)
+
+func ElasticsearchMetrics(
+	installation *operatorv1.InstallationSpec,
+	clusterConfig *relasticsearch.ClusterConfig,
```
tmjd, line 26: I'm not a fan of the name clusterConfig. There are too many possible "clusters", I think esConfig would be much more obvious.

pkg/render/logstorage/components/elasticsearch_metrics.go
```diff
@@ -0,0 +1,151 @@
...
+
+func (e *elasticsearchMetrics) ResolveImages(is *operatorv1.ImageSet) error {
+	var err error
+	var errMsgs []string
+
+	reg := e.installation.Registry
+	path := e.installation.ImagePath
+
+	e.esMetricsImage, err = components.GetReference(components.ComponentElasticsearchMetrics, reg, path, is)
+	if err != nil {
```
tmjd, line 57: I think this code was just copied, and I'm hoping I'm not contradicting what I did in other places, but I think if this is error then this can return, no need for the append.
It should be this way if there are multiple images but that isn't the case here.

pkg/render/logstorage/components/elasticsearch_metrics.go
```diff
@@ -0,0 +1,151 @@
...
+	}
+
+	if len(errMsgs) != 0 {
+		return fmt.Errorf(strings.Join(errMsgs, ","))
+	}
+	return nil
+}
+
+func (e *elasticsearchMetrics) Objects() (objsToCreate, objsToDelete []client.Object) {
+	var toCreate []client.Object
```
tmjd, line 68: There is nothing dynamic here, should it just be `return []client.Object{ <secret>, e.metricsService(), e.metricsDeployment()}, nil`.
No biggie if you think it should remain the way you've got it.

pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -617,6 +651,50 @@ func (r *ReconcileLogStorage) deleteInvalidECKManagedPublicCertSecret(ctx contex
 	return r.client.Delete(ctx, secret)
 }
 
+func (r *ReconcileLogStorage) getElasticsearchCertificateSecrets(ctx context.Context) (*corev1.Secret, *corev1.Secret, *corev1.Secret, error) {
```
tmjd, line 654: Can you add a function comment to list the secrets and order that are returned?

tmjd (commented):
Looking good, I think there was only one error that had to be fixed.

pkg/render/fluentd.go
Brian-McM: No I'm going to use a pod selector from prometheus.

pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -530,7 +537,7 @@ func (r *ReconcileLogStorage) Reconcile(ctx context.Context, request reconcile.R
 		elasticsearch,
 		kibana,
 		clusterConfig,
-		elasticsearchSecrets,
+		[]*corev1.Secret{esCertSecret, esCertSecretESCopy, esPubCertSecret, esAdminUserSecret},
```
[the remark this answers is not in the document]
Brian-McM: I think I made sure there wouldn't be, I fixed up the copy functions to ignore nil but I'll double check.

pkg/controller/logstorage/logstorage_controller.go
Brian-McM: Sure thing

pkg/controller/logstorage/logstorage_controller.go
Brian-McM: It's actually supposed to be certSecret. certSecret is what the user would create and the ECK operator creates pubSecret. What this is saying is that if this is operator managed and the dns name is wrong, delete the public secret so it can get recreated.

pkg/render/logstorage/components/elasticsearch_metrics.go
Brian-McM: If don't think the errors from GetReference should be handled then I don't believe GetReference should return errors. I'd rather not ignore a returned error here.

pkg/render/logstorage/components/elasticsearch_metrics.go
Brian-McM: I had just confused myself, the reason I did this was because you `secret.ToRuntimeObject` returns multiple secrets objects, so I usually just like to use the array expansion append uses. I can just assign toCreate though to the result of ToRunitimeObjects.

pkg/controller/logstorage/logstorage_controller.go
tmjd: Could you include a comment on the if block that is the intention here?

Also how does this work then when a user provides the certSecret (after first using the auto-generated one)? Is it possible that we need to delete the pubSecret so it gets recreated?

pkg/render/logstorage/components/elasticsearch_metrics.go
tmjd: Sorry I wasn't clear here, I wasn't trying to suggest ignoring the error.
I was trying to say that after the `..., err = components.GetReference ...` we can just `return err`.

Brian-McM: Ohh right ya this is over kill.

pkg/controller/logstorage/logstorage_controller.go
Brian-McM: I think you've pointed out an existing bug in the code. This is the logic that was there originally so I'm going to leave it as is, but I've created this ticket https://tigera.atlassian.net/browse/SAAS-1365 to address this issue.

tmjd: I would still like a comment on why in this pubSecret if block that we're checking if the certSecret "IsOperatorManaged".

tmjd (approved):
Please add the requested comment then I think this LGTM.

pkg/controller/logstorage/logstorage_controller.go
Brian-McM: I added the comment to the wrong section, just moved it.

marvin-tigera:

PR #1217, opened 2021-03-09
Add Typha prometheus config

tmjd:
## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/migration/convert/typha.go
+++ b/pkg/controller/migration/convert/typha.go
@@ -1,10 +1,52 @@
 package convert
 
 import (
+	"fmt"
+	"strconv"
+	"strings"
+
 	operatorv1 "github.com/tigera/operator/api/v1"
+)
+
+const (
+	containerTypha = "calico-typha"
 )
 
 func checkTypha(c *components, _ *operatorv1.Installation) error {
 	// No validation required.
+	return nil
+}
+
+// handleTyphaMetrics is a migration handler which detects custom prometheus settings for typha and
+// carries those options forward via the TyphaMetricsPort field.
+func handleTyphaMetrics(c *components, install *operatorv1.Installation) error {
+	if c.typha == nil {
+		return nil
+	}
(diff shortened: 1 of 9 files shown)


caseydavenport (approved):

pkg/controller/migration/convert/typha.go
```diff
@@ -1,10 +1,52 @@
...
+func handleTyphaMetrics(c *components, install *operatorv1.Installation) error {
+	if c.typha == nil {
+		return nil
+	}
+	metricsEnabled, err := getEnv(ctx, c.client, c.typha.Spec.Template.Spec, ComponentTypha, containerTypha, "TYPHA_PROMETHEUSMETRICSENABLED")
+	if err != nil {
+		return err
+	}
+	if metricsEnabled != nil && strings.ToLower(*metricsEnabled) == "true" {
+		var _9091 int32 = 9091
```
stevegaossou, line 31: Hey @tmjd, 

I remember coming across this in the Prometheus docs regarding picking port numbers: 
https://prometheus.io/docs/instrumenting/writing_exporters/#port-numbers

Seems like 9091 is one of the [default ports used by Prometheus](https://github.com/prometheus/prometheus/wiki/Default-port-allocations#core-components-starting-at-9090) (pushgateway). Might be better to avoid that port range for our Prometheus endpoint (just to be safe).

tmjd: I used that port because that is what our docs say the default value is for the port. Even worse that port is actually the same default as the node metrics port.
I didn't change that default value because in the migration case the user has already been using that port successfully so it should continue to work.

stevegaossou: Ah right, I vaguely recall seeing that we were using both 9091 (prometheusMetricsPort ) and 9092 (prometheusReporterPort ) ports for our defaults. I didn't do anything at the time, because I was adding a new endpoint (BGP) at the time, so I was able to avoid the well-known ports when choosing my default value (which was more important, since I wasn't exposing the BGP port for user override).

ozdanborne (approved):

PR #1227, opened 2021-03-12
inherit nodeReporterPort from felixconfig

ozdanborne:
## Description

Some of our users are setting the PrometheusReporterPort in FelixConfiguration (Enterprise) to prevent port conflicts with other services on their hosts. That felixconfig option gets overwritten by the env var Operator sets.

This PR enables Operator to check if the value is set in the FelixConfiguration and use that value instead of its hardcoded default.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


pkg/controller/installation/core_controller.go
```diff
@@ -913,6 +925,26 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
...
+	// Operator is responsible for creating a service which maps to that port.
+	// Here, we'll check the default felixconfiguration to see if the user is specifying
+	// a non-default port, and use that value if they are.
+	nodeReporterMetricsPort := defaultNodeReporterPort
+	if instance.Spec.Variant == operator.TigeraSecureEnterprise {
+		// Query the FelixConfiguration object. We'll use this to help configure felix.
+		felixConfiguration := &crdv1.FelixConfiguration{}
+		err = r.client.Get(ctx, types.NamespacedName{Name: "default"}, felixConfiguration)
+		if err != nil && !apierrors.IsNotFound(err) {
+			r.SetDegraded("Unable to ready FelixConfiguration", err, reqLogger)
```
tmjd, line 938: typo: ready should be read

pkg/render/node.go
```diff
@@ -1086,7 +1086,7 @@ func (c *nodeComponent) nodeEnvVars() []v1.EnvVar {
 		// Add in Calico Enterprise specific configuration.
 		extraNodeEnv := []v1.EnvVar{
 			{Name: "FELIX_PROMETHEUSREPORTERENABLED", Value: "true"},
-			{Name: "FELIX_PROMETHEUSREPORTERPORT", Value: fmt.Sprintf("%d", nodeReporterPort)},
+			{Name: "FELIX_PROMETHEUSREPORTERPORT", Value: fmt.Sprintf("%d", c.nodeReporterMetricsPort)},
```
tmjd, line 1089: Should we check that the port is not zero? I think it would probably be a problem if it was. Should we set the above line to false if it is zero?

tmjd: Or maybe we should check that it is not zero in the controller and degrade if it is? :shrug:

tmjd (commented):
Looking good, maybe one validation to add to the controller?

pkg/controller/installation/core_controller.go
```diff
@@ -913,11 +925,37 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
...
+		}
+
+		// Determine the port to use for nodeReporter metrics.
+		if felixConfiguration.Spec.PrometheusReporterPort != nil {
+			nodeReporterMetricsPort = *felixConfiguration.Spec.PrometheusReporterPort
+		}
+
+		if nodeReporterMetricsPort == 0 {
+			err := errors.New("felixConfiguration prometheusReporterPort=0 not supported")
+			r.SetDegraded("", err, reqLogger)
```
tmjd, line 949: I think we need to have something in the 1st field because that is the Degraded "reason", I think saying "Invalid metrics port" would work and you can keep the error message as you have it.

tmjd (approved):
LGTM

PR #1233, opened 2021-03-18
[v1.14] Only enable groups params if secrets are present (google oidc)

rene-dekker:
Cannot be picked.

Only enable groups params if secrets are present (google oidc)

### DIFF
--- a/pkg/render/dex_config.go
+++ b/pkg/render/dex_config.go
@@ -458,7 +458,7 @@ func (d *dexConfig) Connector() map[string]interface{} {
 	}
 
 	//Google specific.
-	if d.idpSecret.Data[serviceAccountSecretField] != nil || d.idpSecret.Data[adminEmailSecretField] == nil {
+	if d.idpSecret.Data[serviceAccountSecretField] != nil && d.idpSecret.Data[adminEmailSecretField] != nil {
 		config["serviceAccountFilePath"] = serviceAccountSecretLocation
 		config[adminEmailSecretField] = fmt.Sprintf("$%s", googleAdminEmailEnv)
 	}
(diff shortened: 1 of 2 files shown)


pkg/render/dex_config.go
```diff
@@ -458,7 +458,7 @@ func (d *dexConfig) Connector() map[string]interface{} {
 	}
 
 	//Google specific.
-	if d.idpSecret.Data[serviceAccountSecretField] != nil || d.idpSecret.Data[adminEmailSecretField] == nil {
+	if d.idpSecret.Data[serviceAccountSecretField] != nil && d.idpSecret.Data[adminEmailSecretField] != nil {
 		config["serviceAccountFilePath"] = serviceAccountSecretLocation
```
stevegaossou, line 462: Minor: Any reason `"serviceAccountFilePath"` is hard-coded string instead of a constant like the other values? 

The string is only used once, so there's that. But, it still feels more consistent to have it defined in a const like the others.

stevegaossou (approved):
LGTM

pkg/render/dex_config.go
```diff
@@ -458,8 +459,8 @@ func (d *dexConfig) Connector() map[string]interface{} {
 	}
 
 	//Google specific.
-	if d.idpSecret.Data[serviceAccountSecretField] != nil || d.idpSecret.Data[adminEmailSecretField] == nil {
-		config["serviceAccountFilePath"] = serviceAccountSecretLocation
+	if d.idpSecret.Data[serviceAccountSecretField] != nil && d.idpSecret.Data[adminEmailSecretField] != nil {
```
tmjd, line 462: The comment above this says it is google specific and in the master branch it is only done for the connectorTypeGoogle. Should we include `connectorType == connectorTypeGoogle` in the `if` statement?

rene-dekker: We could, but it does not really matter for two reasons:
- The other connectors in dex don't read this config, so it will be ignored when the dex code parses the config.
- There is no way a user can get their hands on this config if they are not using google.
I think we can decide to leave it the way it is.

PR #1234, opened 2021-03-18
Check for nil replicas before accessing

caseydavenport:
## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/migration/namespace_migration.go
+++ b/pkg/controller/migration/namespace_migration.go
@@ -270,8 +270,10 @@ func (m *CoreNamespaceMigration) ensureTyphaRoom(ctx context.Context, log logr.L
 		// No Typha running - we're good to go.
 		return nil
 	}
-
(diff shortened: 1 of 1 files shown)


tmjd (approved):
LGTM

pkg/controller/migration/namespace_migration.go
```diff
@@ -270,8 +270,10 @@ func (m *CoreNamespaceMigration) ensureTyphaRoom(ctx context.Context, log logr.L
 		// No Typha running - we're good to go.
```
caseydavenport, line 270: I think this check is bugged somehow - maybe it's not returning nil but instead just an empty struct?

tmjd: I would think we would hit this in more places than just here. Yeah it seems odd.

PR #1238, opened 2021-03-18
Add OIDC Prompt type to Authentication spec

rene-dekker:
Add prompt type so certain idps can now be supported.


api/v1/authentication_types.go
```diff
@@ -100,8 +100,30 @@ type AuthenticationOIDC struct {
 	// +optional
 	// +kubebuilder:validation:Enum=Verify;InsecureSkip
 	EmailVerification *EmailVerificationType `json:"emailVerification,omitempty"`
+
+	// PromptTypes is an optional list of string values that specifies whether the identity provider prompts the end user
+	// for re-authentication and consent. See the RFC for more information on prompt types:
+	// https://openid.net/specs/openid-connect-core-1_0.html. The defined values are:
+	// ["None", "Login", "Consent", "SelectAccount"].
```
tmjd, line 107: Do we need to have explicit validation on PromptTypes? I can't remember if we automatically get validation that only the strings matching a PromptType will be allowed.
If not then I think a a user could specify anything and it would be passed through.

api/v1/authentication_types.go
```diff
@@ -100,8 +100,30 @@ type AuthenticationOIDC struct {
...
+	// +optional
+	PromptTypes []PromptType `json:"promptTypes,omitempty"`
 }
 
+// PromptType is a value that specifies whether the identity provider prompts the end user for re-authentication and
+// consent. See the RFC for more information on prompt types: https://openid.net/specs/openid-connect-core-1_0.html
+type PromptType string
```
tmjd, line 114: Looking at what is rendered in the API Reference, we should list the allowed values in the comment on this type.

api/v1/authentication_types.go
rene-dekker: Adding the kubebuilder magic to PromptType fixed it, thanks for pointing that out!
```
spec.oidc.promptTypes: Unsupported value: "oanasd": supported values: "None", "Login", "Consent", "SelectAccount"
```

api/v1/authentication_types.go
```diff
@@ -100,8 +100,32 @@ type AuthenticationOIDC struct {
 	// +optional
 	// +kubebuilder:validation:Enum=Verify;InsecureSkip
 	EmailVerification *EmailVerificationType `json:"emailVerification,omitempty"`
+
+	// PromptTypes is an optional list of string values that specifies whether the identity provider prompts the end user
+	// for re-authentication and consent. See the RFC for more information on prompt types:
+	// https://openid.net/specs/openid-connect-core-1_0.html. The defined values are:
+	// ["None", "Login", "Consent", "SelectAccount"].
```
stevegaossou, line 107: Minor: Since you already explain the allowable values and the RFC further down with the PromptType declaration, you don't need to also explain it here. 

I think the first sentence you have adequate, though it may be worth mentioning what the default behaviour is (if the user does not define PromptTypes).

pkg/render/dex_config.go
```diff
@@ -479,6 +479,25 @@ func (d *dexConfig) Connector() map[string]interface{} {
...
+					prompts[i] = "select_account"
+				case oprv1.PromptTypeLogin:
+					prompts[i] = "login"
+				case oprv1.PromptTypeConsent:
+					prompts[i] = "consent"
+				}
+			}
+			// RFC specifies space delimited case sensitive list: https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest
+			config["promptType"] = strings.Join(prompts, " ")
+		}
```
stevegaossou, line 500: Minor: The Connector function is starting to grow in size. Might be easier to read to split case into a separate function (at some point).

stevegaossou (approved):
LGTM

pkg/render/dex_config.go
rene-dekker: I'd like to defer this for now.

api/v1/authentication_types.go
tmjd: I still don't see the allowed values in this comment.

rene-dekker: Added. I used the "One of: ..." way for showing the options. This seems to have been used a lot in the recent past.

tmjd: Ok, just for my info, does the api reference do anything special with that?

tmjd (approved):
LGTM

PR #1247, opened 2021-03-29
Add rights to access licenkeys for kube-controller

asincu:
## Description

Adding rights to list/watch/get licensekeys from a management cluster and create/update from a managed cluster.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/kube-controllers.go
+++ b/pkg/render/kube-controllers.go
@@ -242,6 +242,12 @@ func (c *kubeControllersComponent) controllersRole() *rbacv1.ClusterRole {
 				Verbs:     []string{"watch", "list", "get"},
 			},
 			{
+				// Needed to validate the license
+				APIGroups: []string{"projectcalico.org"},
+				Resources: []string{"licensekeys"},
+				Verbs:     []string{"get", "watch", "list"},
+			},
+			{
 				// calico-kube-controllers requires tiers create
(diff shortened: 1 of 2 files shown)


rene-dekker (approved):
lgtm

pkg/render/kube-controllers.go
```diff
@@ -283,6 +289,14 @@ func (c *kubeControllersComponent) controllersRole() *rbacv1.ClusterRole {
 				Verbs:     []string{"create"},
 			})
 		}
+
+		if c.managementClusterConnection != nil {
+			role.Rules = append(role.Rules, rbacv1.PolicyRule{
```
tmjd, line 294: How does this work since the actual create/update comes from the management cluster?
Does guardian impersonate kube-controller in the managed cluster, so operator in the managed cluster needs to give kube-controller permissions to create/update the licensekeys even though the kube-controller in the managed cluster won't be doing anything with licenses.

asincu: Yes, all API requests that pass through the tunnel use impersonation.

rene-dekker: That is right. Guardian impersonates ckc and for this the role should be present.

PR #1251, opened 2021-04-05
Cert management for Elasticsearch

rene-dekker:
Add certificate management for Elasticsearch.

### DIFF
--- a/pkg/controller/logstorage/logstorage_controller.go
+++ b/pkg/controller/logstorage/logstorage_controller.go
@@ -348,11 +348,6 @@ func (r *ReconcileLogStorage) Reconcile(ctx context.Context, request reconcile.R
 		return reconcile.Result{}, err
 	}
 
-	if ls != nil && install.CertificateManagement != nil {
-		r.status.SetDegraded("Certificate Management is not yet supported for clusters with LogStorage, please remove the setting from your Installation resource.", "")
-		return reconcile.Result{}, fmt.Errorf("certificate management is not yet supported for clusters with LogStorage, please remove the setting from your Installation resource")
-	}
-
 	managementCluster, err := utils.GetManagementCluster(ctx, r.client)
 	if err != nil {
 		reqLogger.Error(err, "Error reading ManagementCluster")
@@ -443,13 +438,13 @@ func (r *ReconcileLogStorage) Reconcile(ctx context.Context, request reconcile.R
 			esAdminUserSecret = rsecret.CopyToNamespace(rmeta.OperatorNamespace(), esAdminUserSecret)[0]
 		}
 
-		if esCertSecret, esCertSecretESCopy, esPubCertSecret, err = r.getElasticsearchCertificateSecrets(ctx); err != nil {
+		if esCertSecret, esCertSecretESCopy, esPubCertSecret, err = r.getElasticsearchCertificateSecrets(ctx, install); err != nil {
 			reqLogger.Error(err, err.Error())
 			r.status.SetDegraded("Failed to create elasticsearch secrets", err.Error())
 			return reconcile.Result{}, err
 		}
 
-		if kibanaSecrets, err = r.kibanaSecrets(ctx); err != nil {
+		if kibanaSecrets, err = r.kibanaSecrets(ctx, install); err != nil {
 			reqLogger.Error(err, err.Error())
 			r.status.SetDegraded("Failed to create kibana secrets", err.Error())
 			return reconcile.Result{}, err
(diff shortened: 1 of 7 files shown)


pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -660,51 +655,71 @@ func (r *ReconcileLogStorage) deleteInvalidECKManagedPublicCertSecret(ctx contex
...
 	if err != nil {
 		return nil, nil, nil, err
 	}
 
-	// If the provided certificate secret (secret) is managed by the operator we need to check if the secret that
-	// Elasticsearch creates from that given secret (pubSecret) has the expected DNS name. If it doesn't, delete the
-	// public secret so it can get recreated.
-	if pubSecret != nil {
-		operatorManaged, err := utils.IsOperatorManaged(certSecret, "tls.crt")
+	// If Certificate management is enabled, we only want to trust the CA cert and let private key generation being handled from an init container.
```
Brian-McM, line 678: Nit "let private key generation being handled from an init container" => "let the init container handle private key generation"

pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -660,51 +655,71 @@ func (r *ReconcileLogStorage) deleteInvalidECKManagedPublicCertSecret(ctx contex
...
-			if err == utils.ErrInvalidCertDNSNames {
-				if err := r.deleteInvalidECKManagedPublicCertSecret(ctx, pubSecret); err != nil {
-					return nil, nil, nil, err
+		operatorManaged := utils.IsOperatorIssued(oprKeyCertIssuer)
+
+		// If the provided certificate secret (secret) is managed by the operator we need to check if the secret that
+		// Elasticsearch creates from that given secret (pubSecret) has the expected DNS name. If it doesn't, delete the
+		// public secret so it can get recreated.
+		if pubSecret != nil {
+
```
Brian-McM, line 708: Nit rogue space

pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -660,51 +655,71 @@ func (r *ReconcileLogStorage) deleteInvalidECKManagedPublicCertSecret(ctx contex
...
+	// If Certificate management is enabled, we only want to trust the CA cert and let private key generation being handled from an init container.
+	if instl.CertificateManagement != nil {
+		cmCa := instl.CertificateManagement.CACert
+		cmIssuer, err := utils.GetCertificateIssuer(cmCa)
+		if err != nil {
+			return nil, nil, nil, err
+		}
+
+		if cmIssuer != oprKeyCertIssuer && !utils.IsOperatorIssued(oprKeyCertIssuer) {
+			return nil, nil, nil, fmt.Errorf("certificate management does not support custom Elastic secrets, please delete secret %s/%s or disable certificate management", oprKeyCert.Namespace, oprKeyCert.Name)
```
Brian-McM, line 687: This is something the user will likely see so could you please say Elasticsearch and not Elastic.

Also, I'm not sure "custom Elastic secrets" is right, you mean "customer Elasticsearch certificates" right?

Brian-McM: I'm also wondering if it's a fair assumption to just overwrite the certificate or delete if they own it if they have enabled certification management. Enabling certificate management sort of singles user intent that they expect this certificate to be managed by the system they've configured. I'd run this by @tmjd though first.

pkg/render/logstorage.go
```diff
@@ -586,6 +614,113 @@ func (es elasticsearchComponent) podTemplate() corev1.PodTemplateSpec {
...
+	var autoMountToken bool
+	if es.installation.CertificateManagement != nil {
+
+		// If certificate management is used, we need to override a mounting options for this init container.
+		initFSName := "elastic-internal-init-filesystem"
+		initFSContainer := corev1.Container{
+			Name:  initFSName,
+			Image: es.esImage,
+			SecurityContext: &corev1.SecurityContext{
+				Privileged: ptr.BoolToPtr(true),
```
Brian-McM, line 626: Can you double check with the security team that we can do this?

pkg/render/logstorage.go
```diff
@@ -586,6 +614,113 @@ func (es elasticsearchComponent) podTemplate() corev1.PodTemplateSpec {
...
+				RunAsUser:  ptr.Int64ToPtr(0),
+			},
+			Args: []string{"bash", "-c", "/mnt/elastic-internal/scripts/prepare-fs.sh"},
```
Brian-McM, line 629: Does the configuration get merged with the default init container that's there, or is the init container overwritten completely by this? I ask because if it's overwritten completely you're missing the resource limits, but if it's merged then these arguments are the same as before so you probably don't need to use include them.

pkg/controller/logstorage/logstorage_controller.go
rene-dekker: I mean the "custom" secrets.

> I'm also wondering if it's a fair assumption to just overwrite the certificate or delete if they own it if they have enabled certification management.
I thought about this quite a bit. I asked CS, Tesla and Solutions and it seems like nobody ever used the BYO secrets feature, which is a good thing. For the hypothetical scenario in which a customer did bring his own certificate, I am hesitant to delete it, as we cannot judge how hard it was to obtain that certificate for them. They may be disappointed to see being deleted. Logging an error seemed like the appropriate thing to do.

If @Brian-McM or @tmjd or anyone else reviewing this feel that deleting it is better, I can be pusuaded.

pkg/render/logstorage.go
rene-dekker: They are aware. This is not new. The init container has always existed and has always been privileged. For this reason we maintain a table/spreadsheet somewhere in the docs to show our progress on privileged containers where we have an entry for Elasticsearch.

pkg/render/logstorage.go
rene-dekker: I'm astounded by this insight, really nice. I've noticed that some things they merge, others they don't. I will verify this and deal with it if necessary before resolving this thread.

pkg/render/logstorage.go
rene-dekker: Actually, it works without it being privileged, so I will address this, thanks.

pkg/render/logstorage.go
rene-dekker: Added the resources, thanks!

pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -660,51 +655,70 @@ func (r *ReconcileLogStorage) deleteInvalidECKManagedPublicCertSecret(ctx contex
...
 			}
-		}
 
-		pubSecret = rsecret.CopyToNamespace(rmeta.OperatorNamespace(), pubSecret)[0]
+			oprCert = rsecret.CopyToNamespace(rmeta.OperatorNamespace(), pubSecret)[0]
+		}
 	}
 
-	return certSecret, certSecretESCopy, pubSecret, nil
+	return oprKeyCert, esKeyCert, oprCert, err
```
asincu, line 721: I think we should rename the third variable to pubSecret. I find having both `oprKeyCert` and `oprCert` confusing.

pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -721,7 +735,22 @@ func (r *ReconcileLogStorage) shouldApplyElasticTrialSecret(ctx context.Context)
 	return false, nil
 }
 
-func (r *ReconcileLogStorage) kibanaSecrets(ctx context.Context) ([]*corev1.Secret, error) {
+// createCaCert is a convenience method for creating a secret that contains the ca to trust.
+func createCaCertSecret(caPem []byte, secretName string, namespace string) *corev1.Secret {
```
asincu, line 739: I suggest moving this into a utils package.

pkg/render/logstorage.go
```diff
@@ -852,9 +997,14 @@ func (es elasticsearchComponent) nodeSetTemplate(pvcTemplate corev1.PersistentVo
 			"op.issuer":                   fmt.Sprintf("%s/dex", es.dexCfg.ManagerURI()),
 			"op.authorization_endpoint":   fmt.Sprintf("%s/dex/auth", es.dexCfg.ManagerURI()),
 			"ssl.certificate_authorities": []string{"/usr/share/elasticsearch/config/dex/tls-dex.crt"},
+			// todo: Add certificate management for dex. This planned already and will be a separate PR.
```
asincu, line 1000: Let's just keep this todo mentally.

asincu (commented):
Looks good! I added some small comments.

pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -660,51 +655,70 @@ func (r *ReconcileLogStorage) deleteInvalidECKManagedPublicCertSecret(ctx contex
...
-		operatorManaged, err := utils.IsOperatorManaged(certSecret, "tls.crt")
+	// If Certificate management is enabled, we only want to trust the CA cert andlet the init container handle private key generation.
+	if instl.CertificateManagement != nil {
+		cmCa := instl.CertificateManagement.CACert
+		cmIssuer, err := utils.GetCertificateIssuer(cmCa)
+		if err != nil {
+			return nil, nil, nil, err
+		}
+
+		if cmIssuer != oprKeyCertIssuer && !utils.IsOperatorIssued(oprKeyCertIssuer) {
```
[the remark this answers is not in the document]
rene-dekker: There are three possible issuers: certManagement, operator, customer.
- If the cert is issued by the customer and at the same time certmanagement is enabled, we want to throw an error.
- CertManagement issuer is not the same as the operator issuer.
- There is an edge case possibility that certManagement and customer are the same issuer, but in this case we should be good to proceed without error.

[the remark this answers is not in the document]
rene-dekker: I will add a comment in the code to make it more clear.

pkg/controller/logstorage/logstorage_controller.go
rene-dekker: I have already done this in the next PR that is about to follow. Please ignore for now.

asincu (approved):

pkg/controller/logstorage/logstorage_controller.go
tmjd: I think erroring is the correct thing to do if the custom certs were provided and cert management is enabled.

Brian-McM: Naw I'm good with the error, it seems like the right thing to do.

tmjd (commented):
Some questions

pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -732,11 +767,20 @@ func (r *ReconcileLogStorage) kibanaSecrets(ctx context.Context) ([]*corev1.Secr
...
 	if err != nil {
 		return nil, err
 	}
 
+	if instl.CertificateManagement != nil {
+		return []*corev1.Secret{
+			secret,
+			rsecret.CopyToNamespace(render.KibanaNamespace, secret)[0],
+			createCaCertSecret(instl.CertificateManagement.CACert, relasticsearch.PublicCertSecret, render.KibanaNamespace),
+			createCaCertSecret(instl.CertificateManagement.CACert, render.KibanaPublicCertSecret, rmeta.OperatorNamespace()),
```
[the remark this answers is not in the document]
rene-dekker: Other controllers will already mount this secret when they need to trust Kibana, so it will work automagically and keep the code simpler.

pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -732,11 +767,20 @@ func (r *ReconcileLogStorage) kibanaSecrets(ctx context.Context) ([]*corev1.Secr
...
+	secret, err = utils.EnsureCertificateSecret(render.TigeraKibanaCertSecret, secret, corev1.TLSPrivateKeyKey, corev1.TLSCertKey, rmeta.DefaultCertificateDuration, svcDNSNames...)
 	if err != nil {
 		return nil, err
 	}
 
+	if instl.CertificateManagement != nil {
+		return []*corev1.Secret{
+			secret,
+			rsecret.CopyToNamespace(render.KibanaNamespace, secret)[0],
+			createCaCertSecret(instl.CertificateManagement.CACert, relasticsearch.PublicCertSecret, render.KibanaNamespace),
```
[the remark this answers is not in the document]
rene-dekker: Yes, PublicCertSecret is the Elasticsearch cert that Kibana should trust, whereas KibanaPublicCertSecret is the kibana cert that other servers should trust when connecting to kibana.

pkg/render/logstorage_test.go
```diff
@@ -1509,7 +1609,7 @@ var deleteLogStorageTests = func(managementCluster *operatorv1.ManagementCluster
 }
 
 func compareResources(resources []client.Object, expectedResources []resourceTestObj) {
-	Expect(len(resources)).To(Equal(len(expectedResources)))
+	// First compare item-by-item, as it provides clear error messages.
```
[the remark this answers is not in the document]
rene-dekker: I considered that, but then I thought that panic gives me about the same information the len() check would. I'll revert, I can always temporarily do it while running my ut suite.

tmjd (approved):
LGTM
Please make sure the test plan includes explicitly testing with a basic license, not Trial since they will be common in Calico Cloud.

PR #1253, opened 2021-04-06
Use certificate/v1 instead when kubernetes version is 1.19 onward

xiumozhan:
In Operator, we are watching CSRs using package certificates.k8s.io/v1beta1 which has been deprecated since v1.19 and will be unavailable since v1.22. For any cluster running K8S v1.19+, we need to use package certificates.k8s.io/v1 instead.

### DIFF
--- a/main.go
+++ b/main.go
@@ -213,8 +213,8 @@ func main() {
 
 	kubernetesVersion, err := common.GetKubernetesVersion(clientset)
 	if err != nil {
-		log.Error(err, fmt.Sprintf("Unable to resolve Kubernetes version"))
-		os.Exit(1)
+		log.Error(err, fmt.Sprintf("Unable to resolve Kubernetes version, defaulting to v1.18"))
+		kubernetesVersion = &common.VersionInfo{Major: 1, Minor: 18}
 	}
 
 	options := options.AddOptions{
--- a/main.go
+++ b/main.go
@@ -42,6 +42,7 @@ import (
 	"github.com/tigera/operator/controllers"
 	"github.com/tigera/operator/pkg/apis"
 	"github.com/tigera/operator/pkg/awssgsetup"
+	"github.com/tigera/operator/pkg/common"
 	"github.com/tigera/operator/pkg/components"
 	"github.com/tigera/operator/pkg/controller/options"
 	"github.com/tigera/operator/pkg/controller/utils"
@@ -210,11 +211,18 @@ func main() {
 		log.Error(err, fmt.Sprintf("Couldn't find the cluster domain from the resolv.conf, defaulting to %s", clusterDomain))
 	}
 
+	kubernetesVersion, err := common.GetKubernetesVersion(clientset)
+	if err != nil {
+		log.Error(err, fmt.Sprintf("Unable to resolve Kubernetes version"))
+		os.Exit(1)
+	}
+
 	options := options.AddOptions{
 		DetectedProvider:    provider,
 		EnterpriseCRDExists: enterpriseCRDExists,
 		AmazonCRDExists:     amazonCRDExists,
 		ClusterDomain:       clusterDomain,
+		KubernetesVersion:   kubernetesVersion,
 	}
 
 	err = controllers.AddToManager(mgr, options)
--- a/pkg/common/common.go
+++ b/pkg/common/common.go
@@ -28,10 +28,3 @@ const (
 	// ExportLogsFeature to 3rd party systems feature name
 	ExportLogsFeature = "export-logs"
 )
-
-// VersionInfo contains information about the version of Kubernetes API the cluster is using
-// Major and Minor fields map to the v<Major>.<Minor>+ part of the version string
-type VersionInfo struct {
-	Major int
-	Minor int
-}
--- a/pkg/common/common.go
+++ b/pkg/common/common.go
@@ -29,6 +29,8 @@ const (
 	ExportLogsFeature = "export-logs"
 )
 
+// VersionInfo contains information about the version of Kubernetes API the cluster is using
+// Major and Minor fields map to the v<Major>.<Minor>+ part of the version string
 type VersionInfo struct {
 	Major int
 	Minor int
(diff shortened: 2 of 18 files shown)


pkg/common/common.go
```diff
@@ -28,3 +28,8 @@ const (
 	// ExportLogsFeature to 3rd party systems feature name
 	ExportLogsFeature = "export-logs"
 )
+
+type VersionInfo struct {
```
rene-dekker, line 32: Please add a comment with a succinct description to every exported type.

pkg/common/kubernetes_version.go
```diff
@@ -0,0 +1,45 @@
...
+package common
+
+import (
+	"fmt"
+	"k8s.io/client-go/kubernetes"
+	"strconv"
+	"strings"
+)
+
+func GetKubernetesVersion(clientset kubernetes.Interface) (*VersionInfo, error) {
```
rene-dekker, line 24: I think that we could change the return type from *VersionInfo -> *version.Info. Thiis is the type returned by the ServerVersion() call. However, please keep the checks that the minor and major are actually present in the response.

pkg/controller/status/status.go
```diff
@@ -17,6 +17,7 @@ package status
 import (
 	"context"
 	"fmt"
+	"github.com/tigera/operator/pkg/common"
```
rene-dekker, line 20: Please organize imports.

pkg/controller/status/status_test.go
```diff
@@ -16,27 +16,31 @@ package status
 
 import (
 	"context"
-
```
rene-dekker, line 19: Please revert.

pkg/controller/status/status.go
```diff
@@ -670,8 +674,44 @@ func (m *statusManager) degradedReason() string {
...
+		return hasPendingCSRUsingCertV1(ctx, m.client, labelMap)
+	}
+	return hasPendingCSRUsingCertV1beta1(ctx, m.client, labelMap)
+}
+
+func hasPendingCSRUsingCertV1(ctx context.Context, cli client.Client, labelMap map[string]string) (bool, error) {
+	csrs := &certV1.CertificateSigningRequestList{}
+	selector := labels.SelectorFromSet(labelMap)
+	err := cli.List(ctx, csrs, &client.ListOptions{LabelSelector: selector})
+	if err != nil {
```
rene-dekker, line 688: From this point onwards the code is duplicate from the other func if you consider that constant `certV1beta1.CertificateApproved` == `certV1.CertificateApproved`. I think you can turn the two funcs into one  if you add a bool param for `useV1Api` or something like that.

xiumozhan: But this `csr` is package-specific, so are `csr.Status`, `csr.Status.Conditions` and `condition.Type`. Unless we define a common interface for both `v1.CertificateSigningRequest` and `v1beta1.CertificateSigningRequest`, combining them into one function will make the code very ugly.

pkg/common/kubernetes_version.go
xiumozhan: `version.Info` has a lot of other fields whereas at this point we only care about `Major` and `Minor`. Besides those fields are of string type which means every time we want to use `Major` and `Minor`, we need to convert them to integers another time.

rene-dekker: Fair enough.

rene-dekker (approved):
lgtm

pkg/common/kubernetes_version.go
```diff
@@ -0,0 +1,45 @@
...
+	minor, err := strconv.Atoi(strings.TrimSuffix(v.Minor, "+"))
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse k8s minor version: %s", v.Minor)
+	}
+
+	return &VersionInfo{
+		Major: major,
+		Minor: minor,
+	}, nil
+}
```
tmjd, line 45: I think we should add functions to the VersionInfo struct for checking if something is available with a specific version.
Like we should add a `ProvidesCertificateV1API`

main.go
```diff
@@ -210,11 +211,18 @@ func main() {
 		log.Error(err, fmt.Sprintf("Couldn't find the cluster domain from the resolv.conf, defaulting to %s", clusterDomain))
 	}
 
+	kubernetesVersion, err := common.GetKubernetesVersion(clientset)
+	if err != nil {
+		log.Error(err, fmt.Sprintf("Unable to resolve Kubernetes version"))
+		os.Exit(1)
```
tmjd, line 217: I'm a little leery of exiting on this failure. Did you consider logging and continuing?
I'm not sure if I would suggest defaulting to v1.19 or v1.18. I think v1.18 is probably the safer option.

tmjd (commented):
A couple comments.

pkg/controller/status/status.go
```diff
@@ -670,8 +674,44 @@ func (m *statusManager) degradedReason() string {
 	return strings.Join(reasons, "; ")
 }
 
-func hasPendingCSR(ctx context.Context, cli client.Client, labelMap map[string]string) (bool, error) {
-	csrs := &v1beta1.CertificateSigningRequestList{}
+func hasPendingCSR(ctx context.Context, m *statusManager, labelMap map[string]string) (bool, error) {
+	if m.kubernetesVersion.Major > 1 || (m.kubernetesVersion.Major == 1 && m.kubernetesVersion.Minor >= 19) {
+		return hasPendingCSRUsingCertV1(ctx, m.client, labelMap)
+	}
+	return hasPendingCSRUsingCertV1beta1(ctx, m.client, labelMap)
```
caseydavenport, line 681: We should have a comment here explaining that we can remove this code once we stop supporting v1.18, so that future developers know when this can / cannot be removed.

main.go
tmjd: To be more specific. I think if this errors we should not fail and exit but instead log and default to v1.18.

tmjd (approved):
LGTM

tmjd:
These commits should be squashed before or when merging.

PR #1255, opened 2021-04-13
Added a watch for the kubernetes_services_endpoint configmap.

sridhartigera:
Added a watch for kubernetes_service_endpoints and operator reads the configmap and pushes the changes to calico-node ds, typha and apiserver deployments.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/apiserver/apiserver_controller.go
+++ b/pkg/controller/apiserver/apiserver_controller.go
@@ -34,7 +34,6 @@ import (
 
 	operatorv1 "github.com/tigera/operator/api/v1"
 	"github.com/tigera/operator/pkg/controller/installation"
-	"github.com/tigera/operator/pkg/controller/k8sapi"
 	"github.com/tigera/operator/pkg/controller/options"
 	"github.com/tigera/operator/pkg/controller/status"
 	"github.com/tigera/operator/pkg/controller/utils"
@@ -90,6 +89,10 @@ func add(mgr manager.Manager, r *ReconcileAPIServer) error {
 
 	if err = utils.AddSecretsWatch(c, render.APIServerTLSSecretName, rmeta.OperatorNamespace()); err != nil {
 		return fmt.Errorf("apiserver-controller failed to watch the Secret resource: %v", err)
+	}
+
+	if err = utils.AddConfigMapWatch(c, render.K8sSvcEndpointConfigMapName, rmeta.OperatorNamespace()); err != nil {
+		return fmt.Errorf("tigera-installation-controller failed to watch ConfigMap %s: %w", render.K8sSvcEndpointConfigMapName, err)
 	}
 
 	for _, namespace := range []string{rmeta.OperatorNamespace(), render.APIServerNamespace} {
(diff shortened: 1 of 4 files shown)


pkg/controller/utils/utils.go
```diff
@@ -280,6 +281,28 @@ func ValidateCertPair(client client.Client, namespace, certPairSecretName, keyNa
...
+	cmName := render.K8sSvcEndpointConfigMapName
+	cm := &corev1.ConfigMap{}
+	cmNamespacedName := types.NamespacedName{
+		Name:      cmName,
+		Namespace: rmeta.OperatorNamespace(),
+	}
+	if err := client.Get(context.Background(), cmNamespacedName, cm); err != nil {
+		// If the configmap is unavailable, do not return error
+		statErr, ok := err.(*kerrors.StatusError)
+		if ok && statErr.ErrStatus.Reason == metav1.StatusReasonNotFound {
```
rene-dekker, line 294: Please use if errors.IsNotFound(err) {}

pkg/controller/apiserver/apiserver_controller.go
```diff
@@ -92,6 +91,10 @@ func add(mgr manager.Manager, r *ReconcileAPIServer) error {
 		return fmt.Errorf("apiserver-controller failed to watch the Secret resource: %v", err)
 	}
 
+	if err = utils.AddConfigMapWatch(c, render.K8sSvcEndpointConfigMapName, rmeta.OperatorNamespace()); err != nil {
+		return fmt.Errorf("tigera-installation-controller failed to watch ConfigMap %s: %w", render.K8sSvcEndpointConfigMapName, err)
```
rene-dekker, line 95: "tigera-installation-controller"->"apiserver-controller"

pkg/controller/utils/utils.go
```diff
@@ -280,6 +281,28 @@ func ValidateCertPair(client client.Client, namespace, certPairSecretName, keyNa
 	return secret, nil
 }
 
+func GetK8sServiceEndPoint(client client.Client) (*k8sapi.ServiceEndpoint, error) {
```
rene-dekker, line 284: We try and add a doc string to every exported type to describe what it is used for succinctly.

pkg/controller/installation/core_controller.go
```diff
@@ -914,6 +918,13 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 		return reconcile.Result{}, err
 	}
 
+	k8sEndpoint, err := utils.GetK8sServiceEndPoint(r.client)
```
rene-dekker, line 921: Maybe I am missing context, but what is responsible for creating this endpoint? This is the first controller that reconciles networking components for calico/tigera, so if the endpoint is not present, networking fails to be installed. This will cascade and everything fails to be installed. Does this mean it will break installations in new clusters?

pkg/controller/apiserver/apiserver_controller.go
```diff
@@ -259,12 +262,19 @@ func (r *ReconcileAPIServer) Reconcile(ctx context.Context, request reconcile.Re
 		}
 	}
 
+	k8sEndpoint, err := utils.GetK8sServiceEndPoint(r.client)
+	if err != nil {
+		log.Error(err, "Error reading services endpoint configmap")
+		r.status.SetDegraded("Error reading services endpoint configmap", err.Error())
```
fasaxc, line 268: Do you need to unset degraded after a successful retry?  (I genuinely don't know due to not knowing the operator very well.)

fasaxc (commented):
Is there anything we can/should do to make the operator itself pick up the change for its own connection to the API server?

pkg/controller/apiserver/apiserver_controller.go
tmjd: No need to do that. It is cleared when every thing is ready.

tmjd (approved):
We may at some point decide that the operator needs to reload if the configmap changes to pick up the values for itself but we can do that in the future. If the operator can read the configmap then its current HOST:PORT values are working so why restart. Hopefully if it stops working the operator would exit/get restarted and then should pick up the new values.

PR #1256, opened 2021-04-13
[hold merge] Make node containers nonprivileged in iptables mode

mgleung:
## Description

Make all the containers in calico-node run nonprivileged while running with the iptables dataplane.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/node.go
+++ b/pkg/render/node.go
@@ -1353,9 +1353,14 @@ func (c *nodeComponent) nodePodSecurityPolicy() *policyv1beta1.PodSecurityPolicy
 	psp.GetObjectMeta().SetName(common.NodeDaemonSetName)
 	psp.Spec.Privileged = true
 	psp.Spec.AllowPrivilegeEscalation = ptr.BoolToPtr(true)
-	psp.Spec.AllowedCapabilities = []v1.Capability{
-		v1.Capability("NET_ADMIN"),
-		v1.Capability("NET_RAW"),
+	if !c.bpfDataplaneEnabled() {
+		// Do not need privileged permissions when running in iptables mode.
+		psp.Spec.Privileged = false
+		psp.Spec.AllowPrivilegeEscalation = ptr.BoolToPtr(false)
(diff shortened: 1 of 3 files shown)


pkg/render/node.go
```diff
@@ -825,14 +827,23 @@ func (c *nodeComponent) cniEnvvars() []v1.EnvVar {
...
+		Name:      "calico-node",
+		Image:     c.nodeImage,
+		Resources: c.nodeResources(),
+		SecurityContext: &v1.SecurityContext{
+			Privileged: ptr.BoolToPtr(false),
+			RunAsUser:  ptr.Int64ToPtr(0),
+			Capabilities: &v1.Capabilities{
+				Add: []v1.Capability{
+					v1.Capability("NET_RAW"),
+					v1.Capability("NET_ADMIN"),
```
fasaxc, line 839: Kernel v5.8 adds CAP_BPF.  BPF mode also needs to adjust resource limits; probably needs CAP_SYS_RESOURCE?

mgleung: Did a bit of experimenting. It appears that we still need `privileged` permissions to write to `bpffs`. We can probably adjust the permissions required if we're able to change where `tc` writes to but a quick test of setting `TC_BPF_MNT` proved unsuccessful for now. We'll probably need to tackle a reduced set of permissions for BPF at a later time.

mgleung: After some more testing, it seems that it should be possible to run BPF as non-privileged with some of the newer kernel versions (5.8+). There are a few more issues around mounts that need to be worked out, but I don't think we should block the removal of `privileged` permissions on the case anymore. We can remove privileged on BPF later.

tmjd:
If this PR needs to go into v1.19 for the up coming enterprise release then once this is merged to master you'll need to cherry-pick into the release-v1.19 branch and add the Milestone of v1.19.0 to that PR.

PR #1257, opened 2021-04-14
FELIX_BPFEXTTOSERVICECONNMARK should default to 0x80 in EKS

tomastigera:
0x80 is the default connmark for routing nodeport traffic to the main
ENI. BPF mode needs to set the flag to match EKS setting.

Setting this in BPF mode is a no-op.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/node.go
+++ b/pkg/render/node.go
@@ -968,6 +968,10 @@ func (c *nodeComponent) nodeEnvVars() []v1.EnvVar {
 		}},
 	}
 
+	if c.cr.CNI != nil && c.cr.CNI.Type == operator.PluginAmazonVPC {
+		nodeEnv = append(nodeEnv, v1.EnvVar{Name: "FELIX_BPFEXTTOSERVICECONNMARK", Value: "0x80"})
+	}
+
 	// If there are no IP pools specified, then configure no default IP pools.
 	if c.cr.CalicoNetwork == nil || len(c.cr.CalicoNetwork.IPPools) == 0 {
 		nodeEnv = append(nodeEnv, v1.EnvVar{Name: "NO_DEFAULT_POOLS", Value: "true"})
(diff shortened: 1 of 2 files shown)


pkg/render/node.go
```diff
@@ -968,6 +968,10 @@ func (c *nodeComponent) nodeEnvVars() []v1.EnvVar {
 		}},
 	}
 
+	if c.cr.KubernetesProvider == operator.ProviderEKS {
```
fasaxc, line 971: I think you'd want to check the CNI that's in use rather than the provider.  You can run EKS with Calico CNI or Kops with Amazon CNI, for example.  It's actually the CNI that does the wacky routing.

tmjd:
Is this something that the user may need to set differently when using AmazonVPC plugin?
Is this setting available to be set in the FelixConfiguration resource?

tmjd:
Is it possible to set that in the FelixConfiguration resource?

tomastigera:
yes https://github.com/projectcalico/libcalico-go/blob/master/lib/apis/v3/felixconfig.go#L328

pkg/render/node.go
```diff
@@ -968,6 +968,10 @@ func (c *nodeComponent) nodeEnvVars() []v1.EnvVar {
 		}},
 	}
 
+	if c.cr.CNI != nil && c.cr.CNI.Type == operator.PluginAmazonVPC {
+		nodeEnv = append(nodeEnv, v1.EnvVar{Name: "FELIX_BPFEXTTOSERVICECONNMARK", Value: "0x80"})
```
tmjd, line 972: Should this only been done when using BPF mode? Or maybe it doesn't matter in non-bpf modes.

tmjd (approved):
Assuming it doesn't matter if we set the env var when in non-bpf mode I'll go ahead and approve this.
We'll tackle more correctly setting this in FelixConfiguration in a future update.

pkg/render/node.go
tomastigera: It does not matter in non-BPF modes.

PR #1260, opened 2021-04-20
Add certificate management for compliance, dex, manager

rene-dekker:
Add cert management for the remaining components: manager, compliance, dex

### DIFF
--- a/pkg/controller/authentication/authentication_controller.go
+++ b/pkg/controller/authentication/authentication_controller.go
@@ -20,7 +20,6 @@ import (
 	"time"
 
 	"github.com/go-ldap/ldap"
-	"github.com/tigera/operator/pkg/render/common/secret"
 
 	oprv1 "github.com/tigera/operator/api/v1"
 	"github.com/tigera/operator/pkg/controller/installation"
@@ -216,22 +215,6 @@ func (r *ReconcileAuthentication) Reconcile(ctx context.Context, request reconci
 				r.status.SetDegraded("Failed to read tigera-operator/tigera-dex-tls secret", err.Error())
 				return reconcile.Result{}, err
 			}
-		} else if tlsSecret.Type == corev1.SecretTypeTLS {
-			// If the tls secret exists but has the type TLS, we want to replace it, because secret type is an immutable field.
-			// Ever since certificate management is introduced, we want to move away from secret type TLS because we no
-			// longer want to store private keys in the secrets.
-			dexNsTLSSecret := secret.CopyToNamespace(render.DexNamespace, tlsSecret)[0]
-			if err := r.client.Delete(ctx, dexNsTLSSecret); err != nil && !errors.IsNotFound(err) {
-				log.Error(err, "Failed to replace outdated secret tigera-dex/tigera-dex-tls secret")
-				r.status.SetDegraded("Failed to replace outdated secret tigera-dex/tigera-dex-tls secret", err.Error())
-				return reconcile.Result{}, err
-			}
-			if err := r.client.Delete(ctx, tlsSecret); err != nil && !errors.IsNotFound(err) {
-				log.Error(err, "Failed to replace outdated secret tigera-operator/tigera-dex-tls secret")
-				r.status.SetDegraded("Failed to replace outdated secret tigera-operator/tigera-dex-tls secret", err.Error())
-				return reconcile.Result{}, err
-			}
-			tlsSecret = render.CreateDexTLSSecret(fmt.Sprintf(render.DexCNPattern, r.clusterDomain))
 		}
 	}
 
@@ -263,7 +246,7 @@ func (r *ReconcileAuthentication) Reconcile(ctx context.Context, request reconci
 	}
 
 	// DexConfig adds convenience methods around dex related objects in k8s and can be used to configure Dex.
-	dexCfg := render.NewDexConfig(install, authentication, tlsSecret, dexSecret, idpSecret, r.clusterDomain)
+	dexCfg := render.NewDexConfig(install.CertificateManagement, authentication, tlsSecret, dexSecret, idpSecret, r.clusterDomain)
 
 	// Create a component handler to manage the rendered component.
 	hlr := utils.NewComponentHandler(log, r.client, r.scheme, authentication)
--- a/pkg/controller/authentication/authentication_controller.go
+++ b/pkg/controller/authentication/authentication_controller.go
@@ -209,7 +209,7 @@ func (r *ReconcileAuthentication) Reconcile(ctx context.Context, request reconci
 		tlsSecret = &corev1.Secret{}
 		if err := r.client.Get(ctx, types.NamespacedName{Name: render.DexTLSSecretName, Namespace: rmeta.OperatorNamespace()}, tlsSecret); err != nil {
 			if errors.IsNotFound(err) {
-				tlsSecret = render.CreateDexTLSSecret(fmt.Sprintf(render.DexCN, r.clusterDomain))
+				tlsSecret = render.CreateDexTLSSecret(fmt.Sprintf(render.DexCNPattern, r.clusterDomain))
 			} else {
 				log.Error(err, "Failed to read tigera-operator/tigera-dex-tls secret")
 				r.status.SetDegraded("Failed to read tigera-operator/tigera-dex-tls secret", err.Error())
--- a/pkg/controller/authentication/authentication_controller.go
+++ b/pkg/controller/authentication/authentication_controller.go
@@ -209,9 +209,7 @@ func (r *ReconcileAuthentication) Reconcile(ctx context.Context, request reconci
 		tlsSecret = &corev1.Secret{}
 		if err := r.client.Get(ctx, types.NamespacedName{Name: render.DexTLSSecretName, Namespace: rmeta.OperatorNamespace()}, tlsSecret); err != nil {
 			if errors.IsNotFound(err) {
-				if install.CertificateManagement == nil {
-					tlsSecret = render.CreateDexTLSSecret(fmt.Sprintf(render.DexCN, r.clusterDomain))
-				}
+				tlsSecret = render.CreateDexTLSSecret(fmt.Sprintf(render.DexCN, r.clusterDomain))
 			} else {
 				log.Error(err, "Failed to read tigera-operator/tigera-dex-tls secret")
 				r.status.SetDegraded("Failed to read tigera-operator/tigera-dex-tls secret", err.Error())
--- a/pkg/controller/authentication/authentication_controller.go
+++ b/pkg/controller/authentication/authentication_controller.go
@@ -20,6 +20,7 @@ import (
 	"time"
 
 	"github.com/go-ldap/ldap"
+	"github.com/tigera/operator/pkg/render/common/secret"
 
 	oprv1 "github.com/tigera/operator/api/v1"
 	"github.com/tigera/operator/pkg/controller/installation"
@@ -215,6 +216,22 @@ func (r *ReconcileAuthentication) Reconcile(ctx context.Context, request reconci
 				r.status.SetDegraded("Failed to read tigera-operator/tigera-dex-tls secret", err.Error())
 				return reconcile.Result{}, err
 			}
+		} else if tlsSecret.Type == corev1.SecretTypeTLS {
+			// If the tls secret exists but has the type TLS, we want to replace it, because secret type is an immutable field.
+			// Ever since certificate management is introduced, we want to move away from secret type TLS because we no
+			// longer want to store private keys in the secrets.
+			dexNsTLSSecret := secret.CopyToNamespace(render.DexNamespace, tlsSecret)[0]
+			if err := r.client.Delete(ctx, dexNsTLSSecret); err != nil && !errors.IsNotFound(err) {
+				log.Error(err, "Failed to replace outdated secret tigera-dex/tigera-dex-tls secret")
+				r.status.SetDegraded("Failed to replace outdated secret tigera-dex/tigera-dex-tls secret", err.Error())
+				return reconcile.Result{}, err
+			}
+			if err := r.client.Delete(ctx, tlsSecret); err != nil && !errors.IsNotFound(err) {
+				log.Error(err, "Failed to replace outdated secret tigera-operator/tigera-dex-tls secret")
+				r.status.SetDegraded("Failed to replace outdated secret tigera-operator/tigera-dex-tls secret", err.Error())
+				return reconcile.Result{}, err
+			}
+			tlsSecret = render.CreateDexTLSSecret(fmt.Sprintf(render.DexCNPattern, r.clusterDomain))
 		}
 	}
 
--- a/pkg/controller/authentication/authentication_controller.go
+++ b/pkg/controller/authentication/authentication_controller.go
@@ -29,8 +29,6 @@ import (
 	"github.com/tigera/operator/pkg/controller/utils/imageset"
 	"github.com/tigera/operator/pkg/render"
 	rmeta "github.com/tigera/operator/pkg/render/common/meta"
-	"github.com/tigera/operator/pkg/render/common/secret"
-
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -48,9 +46,6 @@ var log = logf.Log.WithName("controller_authentication")
 
 const (
 	controllerName = "authentication-controller"
-
-	// Common name to add to the Dex TLS secret.
-	dexCN = "tigera-dex.tigera-dex.svc.%s"
 
 	defaultNameAttribute string = "uid"
 )
@@ -101,7 +96,7 @@ func add(mgr manager.Manager, r *ReconcileAuthentication) error {
 
 	for _, namespace := range []string{rmeta.OperatorNamespace(), render.DexNamespace} {
 		for _, secretName := range []string{
-			render.DexTLSSecretName, render.OIDCSecretName, render.OpenshiftSecretName, render.DexObjectName,
+			render.DexTLSSecretName, render.DexCertSecretName, render.OIDCSecretName, render.OpenshiftSecretName, render.DexObjectName,
 		} {
 			if err = utils.AddSecretsWatch(c, secretName, namespace); err != nil {
 				return fmt.Errorf("%s failed to watch the secret '%s' in '%s' namespace: %w", controllerName, secretName, namespace, err)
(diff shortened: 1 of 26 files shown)


pkg/render/dex.go
```diff
@@ -35,46 +36,70 @@ import (
...
+	DexObjectName     = "tigera-dex"
+	DexPort           = 5556
+	DexCertSecretName = "tigera-dex-tls-crt"
+	DexTLSSecretName  = "tigera-dex-tls"
 
 	// Constants related to Dex configurations
 	DexClientId = "tigera-manager"
+
+	// Common name to add to the Dex TLS secret.
+	DexCN = "tigera-dex.tigera-dex.svc.%s"
```
asincu, line 49: Cosmetic: I would suggest DexCN pattern,

pkg/render/crypto_utils.go
```diff
@@ -67,7 +67,6 @@ func CreateDexTLSSecret(dexCommonName string) *corev1.Secret {
 			Name:      DexTLSSecretName,
 			Namespace: rmeta.OperatorNamespace(),
 		},
-		Type: corev1.SecretTypeTLS,
```
asincu, line 70: Why are we removing the secret type ?

pkg/render/dex_config.go
```diff
@@ -461,6 +480,19 @@ func (d *dexRelyingPartyConfig) UserInfoURI() string {
 	return fmt.Sprintf(userInfoURI, d.clusterDomain)
 }
 
+// CreateCertSecret creates the secret containing the certificate that others should mount in order to trust dex.
+func (d *dexConfig) CreateCertSecret() *corev1.Secret {
+	var certBytes []byte
+
+	if d.installation.CertificateManagement != nil {
```
asincu, line 487:
```suggestion
	if d.installation != nil && d.installation.CertificateManagement != nil {
```

pkg/render/manager.go
```diff
@@ -94,15 +95,20 @@ func Manager(
...
+		tlsAnnotations[tlsSecretHashAnnotation] = rmeta.AnnotationHash(tlsKeyPair.Data)
+	} else {
+		tlsAnnotations[tlsSecretHashAnnotation] = rmeta.AnnotationHash(installation.CertificateManagement.CACert)
+	}
 
 	if dexCfg != nil {
 		tlsSecrets = append(tlsSecrets, dexCfg.RequiredSecrets(ManagerNamespace)...)
 		tlsAnnotations = dexCfg.RequiredAnnotations()
 	}
-	tlsAnnotations[tlsSecretHashAnnotation] = rmeta.AnnotationHash(tlsKeyPair.Data)
```
asincu, line 105: I would rather we keep the annotation grouped together.

pkg/render/manager.go
```diff
@@ -94,15 +95,20 @@ func Manager(
 	clusterDomain string,
 	esLicenseType ElasticsearchLicenseType,
 ) (Component, error) {
-	tlsSecrets := []*corev1.Secret{tlsKeyPair}
-	tlsSecrets = append(tlsSecrets, secret.CopyToNamespace(ManagerNamespace, tlsKeyPair)...)
+	var tlsSecrets []*corev1.Secret
 	tlsAnnotations := make(map[string]string)
+	if installation.CertificateManagement == nil {
```
asincu, line 100:
```suggestion
	if installation != nil && installation.CertificateManagement == nil {
```

pkg/controller/manager/manager_controller.go
```diff
@@ -305,6 +305,10 @@ func (r *ReconcileManager) Reconcile(ctx context.Context, request reconcile.Requ
 		}
 	}
 
+	if certOperatorManaged && installation.CertificateManagement != nil {
+		tlsSecret = nil
```
asincu, line 309: I do not see where this secret is being used after this. Why do we need to se this to nil ?

pkg/render/manager.go
```diff
@@ -94,15 +95,20 @@ func Manager(
...
+		tlsSecrets = append(tlsSecrets, tlsKeyPair)
```
asincu, line 101: I do not like the fact that we are setting this secret to nil in the controller and checking on different condition here.

pkg/controller/manager/manager_controller.go
```diff
@@ -288,7 +288,7 @@ func (r *ReconcileManager) Reconcile(ctx context.Context, request reconcile.Requ
 	// If the secret does not exist, then create one.
 	// If the secret exists but is operator managed, then check that it has the
 	// right DNS names and update it if necessary.
-	if tlsSecret == nil || certOperatorManaged {
+	if (tlsSecret == nil || certOperatorManaged) && installation.CertificateManagement == nil {
```
asincu, line 291:
```suggestion
	if  installation.CertificateManagement == nil {
	    if tlsSecret == nil || certOperatorManaged {
	    }
	} else {
	  if certOperatorManaged {
	     tlsSecret = nil
	  }
	}
```

pkg/controller/manager/manager_controller.go
```diff
@@ -305,6 +305,10 @@ func (r *ReconcileManager) Reconcile(ctx context.Context, request reconcile.Requ
 		}
 	}
 
+	if certOperatorManaged && installation.CertificateManagement != nil {
```
asincu, line 308: Is tlsSecret an user provided certificate ? From what I remember it should not be. Isn't it always generated by operator. This secret should be used within the k8s cluster for communication.

pkg/render/crypto_utils.go
rene-dekker: We are removing the type because it is not allowed to create a TLS secret without a key and it is not allowed to edit the secret type. This type was a bit of a headache case and we want to avoid that going forward with certificate management.

pkg/controller/manager/manager_controller.go
rene-dekker: It is, it is the only secret that we strongly encourage to byo independent of the features you use: https://docs.tigera.io/security/comms/manager-tls

pkg/render/manager.go
rene-dekker: It is impossible for installation to be nil. This is true for everywhere installation is used in the render package.

pkg/render/dex_config.go
rene-dekker: Installation can never be nil in any of the render package. All controllers will terminate if it  is not present.

asincu (approved):

pkg/render/dex_config.go
```diff
@@ -119,27 +121,28 @@ type DexRelyingPartyConfig interface {
 
 func NewDexRelyingPartyConfig(
 	authentication *oprv1.Authentication,
-	tlsSecret *corev1.Secret,
+	caSecret *corev1.Secret,
```
tmjd, line 124: Internally I think this is be called certSecret, should this be the same?
Could you add function definitions with some details on the arguments since this function and the NewDexKeyValidatorConfig have had this field changed but NewDexConfig has not changed.

tmjd (commented):
I'm not sure I'm done reviewing but wanted to submit my comments so far.

pkg/controller/authentication/authentication_controller.go
```diff
@@ -101,7 +96,7 @@ func add(mgr manager.Manager, r *ReconcileAuthentication) error {
 
 	for _, namespace := range []string{rmeta.OperatorNamespace(), render.DexNamespace} {
 		for _, secretName := range []string{
-			render.DexTLSSecretName, render.OIDCSecretName, render.OpenshiftSecretName, render.DexObjectName,
+			render.DexTLSSecretName, render.DexCertSecretName, render.OIDCSecretName, render.OpenshiftSecretName, render.DexObjectName,
```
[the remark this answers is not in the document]
rene-dekker: TLS Secret has a key and a cert and is to be mounted only by Dex itself. CertSecret is for other servers who are to trust Dex.

pkg/controller/compliance/compliance_controller.go
```diff
@@ -306,32 +306,33 @@ func (r *ReconcileCompliance) Reconcile(ctx context.Context, request reconcile.R
...
-		log.Error(err, fmt.Sprintf("failed to retrieve / validate %s", render.ComplianceServerCertSecret))
-		r.status.SetDegraded(fmt.Sprintf("failed to retrieve / validate  %s", render.ComplianceServerCertSecret), err.Error())
-		return reconcile.Result{}, err
-	}
+	var complianceServerCertSecret *corev1.Secret
+	if network.CertificateManagement == nil {
+		complianceServerCertSecret, err = utils.ValidateCertPair(r.client,
+			rmeta.OperatorNamespace(),
+			render.ComplianceServerCertSecret,
+			corev1.TLSCertKey,
```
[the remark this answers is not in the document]
rene-dekker: I personally thought using these constants would go without saying, they are placed in corev1 secret objects and integral to k8s. Do you have a prefence for reverting those?

pkg/render/dex.go
```diff
@@ -91,8 +116,14 @@ func (c *dexComponent) Objects() ([]client.Object, []client.Object) {
 		c.configMap(),
 	}
 	objs = append(objs, secret.ToRuntimeObjects(c.dexConfig.RequiredSecrets(rmeta.OperatorNamespace())...)...)
+	objs = append(objs, c.dexConfig.CreateCertSecret())
```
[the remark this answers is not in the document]
rene-dekker: This is creating the secret that is used in tigera-operator and copied over to other namespaces if they connect to Dex.

pkg/render/dex_config.go
```diff
@@ -182,21 +187,25 @@ func baseCfg(
...
 	}
 }
 
 type dexBaseCfg struct {
+	installation   *oprv1.InstallationSpec
 	authentication *oprv1.Authentication
 	tlsSecret      *corev1.Secret
 	idpSecret      *corev1.Secret
 	dexSecret      *corev1.Secret
+	certSecret     *corev1.Secret
```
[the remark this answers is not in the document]
rene-dekker: Dex itself should mount the TLS Secret, whereas the servers connecting to Dex should mount that certSecret.

pkg/render/dex_config.go
```diff
@@ -234,8 +243,12 @@ func (d *dexBaseCfg) RequestedScopes() []string {
 }
 
 func (d *dexBaseCfg) RequiredSecrets(namespace string) []*corev1.Secret {
-	secrets := []*corev1.Secret{
-		secret.CopyToNamespace(namespace, d.tlsSecret)[0],
+	var secrets []*corev1.Secret
+	if d.tlsSecret != nil {
```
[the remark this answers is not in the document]
rene-dekker: Before these changes, whether the rendered component was Dex itself, or simply a server that should trust dex, it would always mount the secret that contained the private key. Now since these changes, Dex will mount the tls secret, and the other components will ount the certSecret in stead.

pkg/render/dex_config.go
```diff
@@ -370,10 +389,10 @@ func (d *dexConfig) RequiredVolumes() []corev1.Volume {
...
 	return []corev1.Volume{
 		{
-			Name: DexTLSSecretName,
+			Name: DexCertSecretName,
 			VolumeSource: corev1.VolumeSource{
 				Secret: &corev1.SecretVolumeSource{
-					SecretName: DexTLSSecretName,
+					SecretName: DexCertSecretName,
 					Items: []corev1.KeyToPath{
 						{Key: "tls.crt", Path: "tls-dex.crt"},
```
tmjd, line 397: should the corev1.TLSCertKey be used here?

pkg/render/dex_config.go
```diff
@@ -387,10 +406,10 @@ func (d *dexKeyValidatorConfig) RequiredVolumes() []corev1.Volume {
...
```
tmjd, line 414: Same here for the corev1.TLSCertKey

pkg/render/logstorage.go
```diff
@@ -895,9 +895,9 @@ func (es elasticsearchComponent) nodeSets() []esv1.NodeSet {
 	var nodeSets []esv1.NodeSet
 	if nodeConfig.NodeSets == nil || len(nodeConfig.NodeSets) < 1 {
 		nodeSet := es.nodeSetTemplate(pvcTemplate)
-		nodeSet.Name = nodeSetName(pvcTemplate)
-		nodeSet.Count = int32(nodeConfig.Count)
 		nodeSet.PodTemplate = es.podTemplate()
+		nodeSet.Name = nodeSetName(nodeSet.PodTemplate)
```
tmjd, line 899: The changes here in logstorage look un-related to the certificate management changes. Why are they in here?

pkg/render/logstorage_test.go
```diff
@@ -1609,7 +1609,7 @@ var deleteLogStorageTests = func(managementCluster *operatorv1.ManagementCluster
 }
 
 func compareResources(resources []client.Object, expectedResources []resourceTestObj) {
-	Expect(len(resources)).To(Equal(len(expectedResources)))
+	// First compare item-by-item, as it provides clear error messages.
```
tmjd, line 1612: The test will probably panic if expectedResources is longer than resources. How is a panic better than 1st a message that the len of the 2 is different. If we want to see the items in the lists when the len fails then please add to output resources and expectedResources on failure of the len equality check.

pkg/controller/manager/manager_controller.go
```diff
@@ -285,24 +285,28 @@ func (r *ReconcileManager) Reconcile(ctx context.Context, request reconcile.Requ
...
+			)
+
+			if err != nil {
+				r.status.SetDegraded(fmt.Sprintf("Error ensuring manager TLS certificate %q exists and has valid DNS names", render.ManagerTLSSecretName), err.Error())
+				return reconcile.Result{}, err
+			}
 		}
+	} else if certOperatorManaged {
+		tlsSecret = nil
 	}
```
tmjd, line 310: I think there should be an else case here for if there is a user provided cert and CertificateManagement defined, it should result in a degraded state.

pkg/render/manager.go
```diff
@@ -94,16 +95,26 @@ func Manager(
...
 ) (Component, error) {
-	tlsSecrets := []*corev1.Secret{tlsKeyPair}
-	tlsSecrets = append(tlsSecrets, secret.CopyToNamespace(ManagerNamespace, tlsKeyPair)...)
-	tlsAnnotations := make(map[string]string)
+	var tlsSecrets []*corev1.Secret
+	tlsAnnotations := map[string]string{
+		KibanaTLSHashAnnotation: rmeta.SecretsAnnotationHash(kibanaSecrets...),
+	}
+	// If the TLS Secret was passed it means we want to use it, regardless of the certificate management feature.
+	useCertificateManagement := tlsKeyPair == nil && installation.CertificateManagement != nil
```
tmjd, line 103: I don't think we should get here if both tlsKeyPair and installation.CertificateManagement are non-nil?

pkg/render/dex_config.go
```diff
@@ -119,27 +121,28 @@ type DexRelyingPartyConfig interface {
...
-	tlsSecret *corev1.Secret,
+	caSecret *corev1.Secret,
 	clusterDomain string) DexKeyValidatorConfig {
-	return &dexKeyValidatorConfig{baseCfg(authentication, tlsSecret, nil, nil, clusterDomain)}
+	return &dexKeyValidatorConfig{baseCfg(nil, authentication, nil, nil, nil, caSecret, clusterDomain)}
 }
 
 // Create a new DexConfig.
 func NewDexConfig(
+	installation *oprv1.InstallationSpec,
```
tmjd, line 139: Did you consider passing in the CertificateManagement instead of the full installation? I'm not sure if that would be better.

tmjd (changes requested):
Please let me know when you're done with updates for me to re-review.

pkg/controller/manager/manager_controller.go
rene-dekker: For the time being the behaviour is that if the user brings his own TLS secret, we default to the old behaviour and mount that BYO secret. This is also part of the test cycles of the feature.

Currently we have not budgetted the time to improve this, but ideally we allow for the CRD to let a user define what SANs we should add to the TLS secret of the manager. This should be picked up in a next iteration on cert management.

pkg/controller/manager/manager_controller.go
```diff
@@ -327,8 +331,8 @@ func (r *ReconcileManager) Reconcile(ctx context.Context, request reconcile.Requ
 		complianceServerCertSecret, err = utils.ValidateCertPair(r.client,
 			rmeta.OperatorNamespace(),
 			render.ComplianceServerCertSecret,
-			render.ComplianceServerCertName,
-			render.ComplianceServerKeyName,
+			"", // We don't need the key.
```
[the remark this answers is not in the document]
rene-dekker: ValidateCertPair first checks the key, then the cert. In my last commit I -hopefully- caught all usages of this function to make sure.

pkg/render/crypto_utils.go
rene-dekker: You two are right. Things evolved while working on this ticket. While the SecretType is a nuisance, I will revert this change.

pkg/render/manager.go
rene-dekker: If the user brings its own manager-tls, we will use it. For now that is their only way to add extra SANs to the secret. Currently, adding extra SANs is not yet part of the spec.

If the manager-tls is not user-provided while certificatemanagement is enabled, the tlsKeyPair variable will be nil.

So in any case that the manager-tls is provided, we want to use it.

pkg/controller/utils/utils.go
```diff
@@ -249,10 +249,13 @@ func IsFeatureActive(license v3.LicenseKey, featureName string) bool {
 }
 
 // ValidateCertPair checks if the given secret exists in the given
-// namespace and if so that it contains key and cert fields. If a secret exists then it is returned.
-// If there is an error accessing the secret (except NotFound) or the cert
-// does not have both a key and cert field then an appropriate error is returned.
-// If no secret exists then nil, nil is returned to represent that no cert is valid.
+// namespace and if so that it contains key and cert fields. If an
+// empty string is passed for the keyName argument, If a secret
```
tmjd, line 253: I think `If a secret` is a typo and was suppose to be `and a secret`

pkg/controller/manager/manager_controller.go
tmjd: Was this covered in a design doc that I'm forgetting about? What I'm talking about is the user facing API behavior.
>Currently we have not budgetted the time to improve this, but ideally we allow for the CRD to let a user define what SANs we should add to the TLS secret of the manager. This should be picked up in a next iteration on cert management.

I didn't understand the relevance at first, but I think it is because the user may want to use CertificateManagement but they need to be able to provide SANs for the manager but that isn't possible so we're allowing specifying a TLS secret and CertificateManagement?
It seems like a partially implemented feature then.

I don't like this because that means we'll need to change behavior when we have this fully implemented where SANs can be specified for CertificateManagement. My assumption is that when we have the feature implemented fully, we will want to degrade if both the manager TLS secret and CertificateManagement are specified. That is not how this is currently implemented though.
I think my preference would be to go ahead and degrade in this case and say if you need to specify SANs for the manager then you cannot use CertificateManagement yet.

pkg/render/manager.go
tmjd: I understand why this is so I'll resolve this comment but expect further discussion in my related comment on the manager_controller

pkg/controller/manager/manager_controller.go
```diff
@@ -305,7 +305,10 @@ func (r *ReconcileManager) Reconcile(ctx context.Context, request reconcile.Requ
 				return reconcile.Result{}, err
 			}
 		}
-	} else if certOperatorManaged {
+	} else if !certOperatorManaged {
+		r.status.SetDegraded(fmt.Sprintf("self provided secret %s/%s is not supported when certificate management is not supported", render.ManagerNamespace, render.ManagerTLSSecretName), "")
```
tmjd, line 309: I think this wording is a bit off. I think maybe it should be 
> user provided secret %s/%s is not supported when certificate management is enabled

Maybe changing it to something like this? WDYT?
```
err := fmt.Errorf("user provided secret %s/%s is not supported when certificate management is enabled", render.ManagerNamespace, render.ManagerTLSSecretName)
r.status.SetDegraded("Invalid certificate configuration", errMsg)
```

tmjd (commented):
Mostly nit and one wording change request.

pkg/controller/manager/manager_controller_test.go
```diff
@@ -273,6 +277,42 @@ var _ = Describe("Manager controller tests", func() {
...
+				Expect(c.Get(ctx, types.NamespacedName{Name: "tigera-manager", Namespace: render.ManagerNamespace}, deploy)).ShouldNot(HaveOccurred())
+
+				if certificateManagementEnabled {
+					Expect(deploy.Spec.Template.Spec.InitContainers).To(HaveLen(1))
+				} else {
+					Expect(deploy.Spec.Template.Spec.InitContainers).To(HaveLen(0))
+				}
+			}
+		},
+			Entry("Regular flow", false, false, false),
```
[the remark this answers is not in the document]
rene-dekker: We discussed this, I appreciate the code style, but they went paired with IDE warnings, so I won't do it this time.

tmjd (approved):
LGTM
Thanks for your effort here.

PR #1262, opened 2021-04-20
Refactor groups claim and make it cooperate w/ dex

rene-dekker:


pkg/render/dex.go
```diff
@@ -188,6 +188,7 @@ func (c *dexComponent) deployment() client.Object {
 					Containers: []corev1.Container{
 						{
 							Name:            DexObjectName,
+							ImagePullPolicy: "Always", //todo: revert.
```
ozdanborne, line 191: Don't forget to remove this!
```suggestion
```

rene-dekker, line 191: I gotta revert this..

pkg/render/dex_config.go
```diff
@@ -291,7 +281,7 @@ func (d *dexKeyValidatorConfig) RequiredEnv(prefix string) []corev1.EnvVar {
 		{Name: fmt.Sprintf("%sDEX_JWKS_URL", prefix), Value: fmt.Sprintf(jwksURI, d.clusterDomain)},
 		{Name: fmt.Sprintf("%sDEX_CLIENT_ID", prefix), Value: DexClientId},
 		{Name: fmt.Sprintf("%sDEX_USERNAME_CLAIM", prefix), Value: d.UsernameClaim()},
-		{Name: fmt.Sprintf("%sDEX_GROUPS_CLAIM", prefix), Value: d.GroupsClaim()},
+		{Name: fmt.Sprintf("%sDEX_GROUPS_CLAIM", prefix), Value: DefaultGroupsClaim},
```
ozdanborne, line 284: So the claim mapping is only set if you specify a non-default groups claim. However, this env var `%sDEX_GROUPS_CLAIM` is always set to DefaultGroupsClaim, even if you specify a non-default groups claim?

Am I understanding correctly that this is because `DEX_GROUPS_CLAIM` controls where dex places claims in the tokens it issues, whereas the claimMapping controls where it reads claims from the tokens issued to it?

ozdanborne (changes requested):
1 line to remove and 1 question but then if all checks out I'll approve :heavy_check_mark:

pkg/render/dex_config.go
rene-dekker: Dex config has three use cases:
- Dex itself needs config to read the actual IDP
- Relying parties need to get stuff from dex
- "Validators" need to unwrap JWT's and verify the signature
Like you already were leading to is that this is a validator with always gets groups from the "groups" claim of a token issued by dex.
While dex will put data from the IDP issued token from the <user-supplied-groups-claim> claim and put it in a Dex issued token with the claim "groups".

ozdanborne (approved):

PR #1271, opened 2021-04-23
Add init container to mount BPF filesystem.

fasaxc:
## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/node.go
+++ b/pkg/render/node.go
@@ -589,7 +589,7 @@ func (c *nodeComponent) nodeDaemonset(cniCfgMap *v1.ConfigMap) *apps.DaemonSet {
 		initContainers = append(initContainers, c.flexVolumeContainer())
 	}
 
-	if c.bpfDataplane() {
+	if c.bpfDataplaneEnabled() {
 		initContainers = append(initContainers, c.bpffsInitContainer())
 	}
 
@@ -700,7 +700,7 @@ func (c *nodeComponent) nodeVolumes() []v1.Volume {
 		},
 	}
 
-	if c.bpfDataplane() {
+	if c.bpfDataplaneEnabled() {
 		volumes = append(volumes,
 			// Volume for the containing directory so that the init container can mount the child bpf directory if needed.
 			v1.Volume{Name: "sys-fs", VolumeSource: v1.VolumeSource{HostPath: &v1.HostPathVolumeSource{Path: "/sys/fs", Type: &dirOrCreate}}},
@@ -753,7 +753,7 @@ func (c *nodeComponent) nodeVolumes() []v1.Volume {
 	return volumes
 }
 
-func (c *nodeComponent) bpfDataplane() bool {
+func (c *nodeComponent) bpfDataplaneEnabled() bool {
 	return c.cr.CalicoNetwork != nil &&
 		c.cr.CalicoNetwork.LinuxDataplane != nil &&
 		*c.cr.CalicoNetwork.LinuxDataplane == operatorv1.LinuxDataplaneBPF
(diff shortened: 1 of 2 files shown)


pkg/render/node.go
```diff
@@ -746,6 +753,12 @@ func (c *nodeComponent) nodeVolumes() []v1.Volume {
 	return volumes
 }
 
+func (c *nodeComponent) bpfDataplane() bool {
```
tmjd, line 756: For what this function does, I would prefer for the function name to be more of a question, something like `bpfDataplaneEnabled` or `isBpfDataplaneEnabled`. WDYT? I feel like it could improve readability slightly but not excessively so if you don't think that makes sense then feel free to leave it as it is.

pkg/render/node.go
```diff
@@ -589,6 +589,10 @@ func (c *nodeComponent) nodeDaemonset(cniCfgMap *v1.ConfigMap) *apps.DaemonSet {
 		initContainers = append(initContainers, c.flexVolumeContainer())
 	}
 
+	if c.bpfDataplane() {
```
tmjd, line 592: We only need this with AKS so we should only add this container when using ProviderAKS is set.

pkg/render/node.go
```diff
@@ -696,10 +700,13 @@ func (c *nodeComponent) nodeVolumes() []v1.Volume {
...
-	if c.cr.CalicoNetwork != nil &&
-		c.cr.CalicoNetwork.LinuxDataplane != nil &&
-		*c.cr.CalicoNetwork.LinuxDataplane == operatorv1.LinuxDataplaneBPF {
-		volumes = append(volumes, v1.Volume{Name: "bpffs", VolumeSource: v1.VolumeSource{HostPath: &v1.HostPathVolumeSource{Path: "/sys/fs/bpf", Type: &dirMustExist}}})
+	if c.bpfDataplane() {
+		volumes = append(volumes,
+			// Volume for the containing directory so that the init container can mount the child bpf directory if needed.
+			v1.Volume{Name: "sys-fs", VolumeSource: v1.VolumeSource{HostPath: &v1.HostPathVolumeSource{Path: "/sys/fs", Type: &dirMustExist}}},
+			// Volume for the bpffs itself, used by the main node container.
+			v1.Volume{Name: "bpffs", VolumeSource: v1.VolumeSource{HostPath: &v1.HostPathVolumeSource{Path: "/sys/fs/bpf", Type: &dirMustExist}}},
```
tmjd, line 708: Isn't the `dirMustExist` going to be a problem in the AKS case where it doesn't exist initially? Or does that only come into play when the main node container tries to use the volume?

pkg/render/node.go
```diff
@@ -696,10 +700,13 @@ func (c *nodeComponent) nodeVolumes() []v1.Volume {
...
 	}
 
-	if c.cr.CalicoNetwork != nil &&
-		c.cr.CalicoNetwork.LinuxDataplane != nil &&
-		*c.cr.CalicoNetwork.LinuxDataplane == operatorv1.LinuxDataplaneBPF {
-		volumes = append(volumes, v1.Volume{Name: "bpffs", VolumeSource: v1.VolumeSource{HostPath: &v1.HostPathVolumeSource{Path: "/sys/fs/bpf", Type: &dirMustExist}}})
+	if c.bpfDataplane() {
+		volumes = append(volumes,
+			// Volume for the containing directory so that the init container can mount the child bpf directory if needed.
+			v1.Volume{Name: "sys-fs", VolumeSource: v1.VolumeSource{HostPath: &v1.HostPathVolumeSource{Path: "/sys/fs", Type: &dirMustExist}}},
```
tmjd, line 706: I feel like it would be good practice to only add this if ProviderAKS is set also.

pkg/render/node.go
fasaxc: I was worried about that so I tested it and I was relived to discover that the init container gets to run before this would become a problem.  So, it all works as we'd hoped.

pkg/render/node.go
fasaxc: I think it's more broadly useful than that so I'd prefer to leave it in.  On Ubuntu 20.04 and Bottlerocket, it's a no-op but it covers Ubuntu 18.04, which is still pretty common on-prem (and maybe in other clouds that we haven't tested with operator yet).

tmjd (commented):
LGTM.
I'll leave final approval to Matt or I'll come back and approve after he takes a look.

mgleung (approved):
LGTM

PR #1272, opened 2021-04-23
Support BGP layout ConfigMap for BGP multi-rack setups

neiljerram:
## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


pkg/render/node.go
```diff
@@ -67,6 +72,7 @@ func Node(
 	nodeAppArmorProfile string,
 	clusterDomain string,
 	nodeReporterMetricsPort int,
+	bgpLayoutHash string,
```
tmjd, line 75: This feels odd to me to pass in the hash. I think usually the component that needs a hash is calculating it but there is no particular reason for that. So I think passing this in is fine. I just wanted to mention my take in case you had an reservations about passing the hash.
tl;dr, this is not a request to change anything.

tmjd (commented):
I think ideally I would like to have some unit tests on the render/node.go changes, but I would not object if that doesn't happen since you already have the mainline tests.

pkg/render/node.go
neiljerram: I did think about this and went through a few iterations before arriving here.  I think this works well, even though it's a different place than other rmeta.AnnotationHash calls.

Makefile
```diff
@@ -269,6 +271,7 @@ cluster-create: kubectl
 	cp ~/.kube/kind-config-kind $(KUBECONFIG)
 	$(MAKE) deploy-crds
 	$(MAKE) create-tigera-operator-namespace
+	$(MAKE) create-tigera-pull-secret
```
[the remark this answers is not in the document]
neiljerram: I've reverted this for now, and have added UTs instead.

tmjd (approved):
LGTM

PR #1274, opened 2021-04-26
Install elasticsearch with basic license by default

Suraiya-Hameed:
## Description

Do not create trial license in ECK namespace, this will allow ECK operator to install Elasticsearch with basic license.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/logstorage/logstorage_controller.go
+++ b/pkg/controller/logstorage/logstorage_controller.go
@@ -409,7 +409,6 @@ func (r *ReconcileLogStorage) Reconcile(ctx context.Context, request reconcile.R
 	var kibanaSecrets, curatorSecrets []*corev1.Secret
 	var clusterConfig *relasticsearch.ClusterConfig
 	var esLicenseType render.ElasticsearchLicenseType
-	applyTrial := false
 
 	if managementClusterConnection == nil {
 		var flowShards = calculateFlowShards(ls.Spec.Nodes, defaultElasticsearchShards)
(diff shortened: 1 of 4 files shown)


pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -456,12 +455,6 @@ func (r *ReconcileLogStorage) Reconcile(ctx context.Context, request reconcile.R
 			return reconcile.Result{}, err
 		}
 
-		applyTrial, err = r.shouldApplyElasticTrialSecret(ctx)
-		if err != nil {
-			r.status.SetDegraded("Failed to get eck trial license", err.Error())
-			return reconcile.Result{}, err
-		}
-
 		if esLicenseType, err = utils.GetElasticLicenseType(ctx, r.client, reqLogger); err != nil {
```
rene-dekker, line 458: Basically we still need this function for clusters that are upgraded from a lower version within a month of installing the old version, right?

Is it possible to just remove it and always use the 'basic auth' login trick for kibana? That would make the possibly upcoming changes a lot easier. If you prefer to defer this one more release, I can understand.

Suraiya-Hameed: There is no plan to remove that check yet, there might be customers with paid Elasticsearch license, if so they will continue to use Kibana's OIDC.

tmjd (approved):
LGTM

PR #1278, opened 2021-04-27
Add reading ConfigMap for extra Kibana config

tmjd:
To avoid changing LogStorage render the kibana config is a public
variable of render/logstorage so the controller can set the field
without passing in an additional variable.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/logstorage/logstorage_controller.go
+++ b/pkg/controller/logstorage/logstorage_controller.go
@@ -16,6 +16,7 @@ package logstorage
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 
 	cmnv1 "github.com/elastic/cloud-on-k8s/pkg/apis/common/v1"
@@ -208,6 +209,11 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 		if err = utils.AddConfigMapWatch(c, name, render.ElasticsearchNamespace); err != nil {
 			return fmt.Errorf("log-storage-controller failed to watch the ConfigMap resource: %w", err)
 		}
+	}
+
+	// Cloud modifications
+	if err = utils.AddConfigMapWatch(c, "cloud-kibana-config", rmeta.OperatorNamespace()); err != nil {
+		return fmt.Errorf("log-storage-controller failed to watch the ConfigMap resource: %w", err)
 	}
 
 	return nil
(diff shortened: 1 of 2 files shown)


pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -532,6 +538,20 @@ func (r *ReconcileLogStorage) Reconcile(ctx context.Context, request reconcile.R
 		dexCfg = render.NewDexRelyingPartyConfig(authentication, dexTLSSecret, dexSecret, r.clusterDomain)
 	}
 
+	// Cloud modifications
+	kbCm := &corev1.ConfigMap{}
+	key := types.NamespacedName{Name: "cloud-kibana-config", Namespace: rmeta.OperatorNamespace()}
```
ozdanborne, line 543: Is the `cloud-` prefix necessary? Seems like this is a generic feature of "setting custom kibana config"

tmjd: Certainly not necessary but I don't expect that this will be pushed into the main operator branch so I like keeping it obvious that this is a cloud operator only change and if anyone sees the config map the name will hopefully make that obvious.

pkg/render/logstorage.go
```diff
@@ -148,6 +148,8 @@ echo "Keystore initialization successful."
 
 var log = logf.Log.WithName("render")
 
+var CloudKibanaConfig = map[string]interface{}{}
```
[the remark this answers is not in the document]
tmjd: I want to keep the Cloud to keep it clear that it is a Cloud modification. That will ensure merges in the future are obvious about where the field came from.

pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -532,6 +538,20 @@ func (r *ReconcileLogStorage) Reconcile(ctx context.Context, request reconcile.R
...
+	if err = r.client.Get(ctx, key, kbCm); err != nil {
+		if !errors.IsNotFound(err) {
+			return reconcile.Result{}, fmt.Errorf("Failed to read cloud-kibana-config ConfigMap: %s", err.Error())
+		}
+	} else {
+		if err = json.Unmarshal([]byte(kbCm.Data["config"]), &render.CloudKibanaConfig); err != nil {
```
ozdanborne, line 549: FYI - ISTR that configmaps have a field `.BinaryData` for accessing data in byte form, removing the need to cast the string as a byte array.

But I also STR that there are cases where that won't be filled in, so it's only worth switching to it if you can quickly verify it works as expected. So fine leaving this change out.

ozdanborne (approved):
couple small nits but LGTM

pkg/controller/logstorage/logstorage_controller.go
tmjd: It looks like the `.BinaryData` is actually a whole different field ... I thought the same as you until I took a look at the ConfigMap. 
```
    // BinaryData contains the binary data.
    // Each key must consist of alphanumeric characters, '-', '_' or '.'.
    // BinaryData can contain byte sequences that are not in the UTF-8 range.
    // The keys stored in BinaryData must not overlap with the ones in
    // the Data field, this is enforced during validation process.
    // Using this field will require 1.10+ apiserver and
    // kubelet.
    // +optional
    BinaryData map[string][]byte `json:"binaryData,omitempty" protobuf:"bytes,3,rep,name=binaryData"`
```

PR #1279, opened 2021-04-29
Get manager up and running with OIDC type = Tigera

rene-dekker:


Brian-McM:
Replaced with https://github.com/tigera/operator/pull/1293

PR #1286, opened 2021-05-03
Allow users to select Typha affinity type

wkrysmann:
## Description

This PR allows users to select Typha node affinity type freely. 

We should not limit users from selecting different affinity model, since for some clusters, especially with multiple node-pools with different configurations original reasoning that we try to schedule Typha no matter what doesn't hold up. 

Scenario: 
Cluster is running two types of nodes with label "A" and "B". Pool "A" is preferred for Typha, since only those nodes have firewall rules to allow federation with another clusters. Typha is being deployed but there is temporarily no capacity on pool "A" so Typha is scheduled on pool "B". The effect is that scheduler is happy but Typha is degraded. In this scenario it would be better to wait till there is available capacity and schedule Typha on right pool ("A"). 

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


tmjd:
Thank you for this submission but this is intended behavior that we want to enforce. If pool A is where cluster admins want to ensure typha will always be scheduled then they should define affinity for the pool and ensure there is always node capacity, this includes ensuring additional nodes exist in pool A so if any of the nodes go down the typha pods can be scheduled to other nodes in the pool.
The reason for this is that typha is critical to the cluster and should always be able to be scheduled.

wkrysmann:
Thanks for the replay @tmjd! 

>  If pool A is where cluster admins want to ensure Typha will always be scheduled then they should define affinity for the pool and ensure there is always node capacity

This is understood, the problem is that `preferredDuringSchedulingIgnoredDuringExecution` is not deterministic, to give some more examples you potentially don't want Typha to be scheduled on:
- nodes with GPUs
- nodes which holds "decommission" label
- control plane nodes
- or other way around - you want them to be scheduled on control plane nodes, which are getting special attention
- nodes with certain firewall configuration
- other nodes then Typha dedicated nodes

>  additional nodes exist in pool A so if any of the nodes go down the Typha pods can be scheduled to other nodes in the pool.

100% agree on that statement, but this doesn't determine that non-deterministic behavior of scheduler helps here. Scheduling Typha on a node which can't fulfill it's requirements makes more harm then good and taking away configuration option from users makes it harder to manage.

caseydavenport:
>  the problem is that preferredDuringSchedulingIgnoredDuringExecution is not deterministic

@wkrysmann could you provide a bit more detail here? What about that field is non-deterministic? 

I definitely agree with your examples for not wanting Typha scheduled on certain nodes, but I'm surprised to hear that preferredDuringSchedulingIgnoredDuringExecution doesn't actually work.

wkrysmann:
@caseydavenport I was not very precise, sorry about that. I was trying to say that that with `preferredDuringSchedulingIgnoredDuringExecution` if we have 10 node-pools and only few of them can handle Typha and if preferred node-pool can't schedule Typha, we can't predict where Typha will be scheduled - that's the non-deterministic bit I had in mind. 
There is nothing wrong with this scheduling method, it's just that due to it's nature, it may not be addressing the intended safeguard for Typha.

api/v1/installation_types.go
```diff
@@ -129,17 +129,7 @@ type InstallationSpec struct {
...
-}
-
-// PreferredNodeAffinity is similar to an affinity except it only exposes the Preferred scheduling option,
-// which ensures a pod will still be scheduled even when all the nodes it matches are unschedulable.
-type PreferredNodeAffinity struct {
-	// The scheduler will prefer to schedule pods to nodes that satisfy
-	// the affinity expressions specified by this field, but it may choose
-	// a node that violates one or more of the expressions.
-	// +optional
-	PreferredDuringSchedulingIgnoredDuringExecution []v1.PreferredSchedulingTerm `json:"preferredDuringSchedulingIgnoredDuringExecution,omitempty"`
```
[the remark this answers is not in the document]
tmjd: @caseydavenport do you agree that instead of explicitly putting the RequiredDuring for ProviderAKS, now we should default the field in the Installation resource?

tmjd (changes requested):
After some discussions we're willing to move forward with this change though we want to make sure that we're as prominent as possible with warnings about using `RequiredDuring...` for affinity.
To add this, then in addition to the other comments I made I think we also need a few unit tests that the render/typha is correct.
I think we'd want a few unit tests:
1. Just simply setting the field results in the proper setting on the typha deployment
2. Setting the field on the Installation and having Provider = ProviderAKS results in the specified field being set on the typha deployment (ProviderAKS no longer explicitly means that the virtual node `RequiredDuring...` affinity will be set.

Ideally with this change we would also update the typha autoscaler code to be able to evaluate the available nodes to see that there are enough so the operator can explicitly report when there are not enough nodes that match the required affinity so the best error can be reported. This is not a requirement though because the operator will report degraded if there are less typhas deployed than requested but just will not have the information that the reason for degradation is because there are not enough nodes.

api/v1/installation_types.go
```diff
@@ -129,17 +129,7 @@ type InstallationSpec struct {
 type TyphaAffinity struct {
 	// NodeAffinity describes node affinity scheduling rules for typha.
 	// +optional
-	NodeAffinity *PreferredNodeAffinity `json:"nodeAffinity,omitempty"`
```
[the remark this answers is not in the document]
wkrysmann: Make sense. Updated.

wkrysmann:
> After some discussions we're willing to move forward with this change though we want to make sure that we're as prominent as possible with warnings about using `RequiredDuring...` for affinity.
> To add this, then in addition to the other comments I made I think we also need a few unit tests that the render/typha is correct.
> I think we'd want a few unit tests:
> 
> 1. Just simply setting the field results in the proper setting on the typha deployment
> 2. Setting the field on the Installation and having Provider = ProviderAKS results in the specified field being set on the typha deployment (ProviderAKS no longer explicitly means that the virtual node `RequiredDuring...` affinity will be set.
> 
> Ideally with this change we would also update the typha autoscaler code to be able to evaluate the available nodes to see that there are enough so the operator can explicitly report when there are not enough nodes that match the required affinity so the best error can be reported. This is not a requirement though because the operator will report degraded if there are less typhas deployed than requested but just will not have the information that the reason for degradation is because there are not enough nodes.

Added more unit tests to cover all the possible scenarios. All tests are passing when I'm running them locally, I see that  `ci/semaphoreci/pr: Initial Pipeline` failed, but I'm getting 404 when accessing the details, so it's hard to say what's wrong.

api/v1/installation_types.go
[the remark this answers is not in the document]
caseydavenport: > do you agree that instead of explicitly putting the RequiredDuring for ProviderAKS, now we should default the field in the Installation resource?

Yeah, I think that is the correct way to do this.

tmjd:
The failure looked unrelated to your changes, it was a failure downloading a utility, so I've kicked off CI again.

deploy/scripts/create_kind_cluster.sh
```diff
@@ -35,7 +35,8 @@ echo
 echo "Download kind executable with dual stack support"
 # We need to replace kind executable and node image
 # with official release once dual stack is fully supported by upstream.
-curl -L https://github.com/song-jiang/kind/releases/download/dualstack-1.17.0/kind -o ${KIND}
+#curl -L https://github.com/song-jiang/kind/releases/download/dualstack-1.17.0/kind -o ${KIND}
```
tmjd, line 38: This change is causing the CI issue.

wkrysmann: Thanks @tmjd I made that change to try to run `make test` on my workstation (I had to use kind binary build for mac) and forget to revert it 🤦 Pushing a fix.

tmjd: That's what I was guessing had probably happened.

api/v1/installation_types.go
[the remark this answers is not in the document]
wkrysmann: I don't think it's a good idea to move this to default, since if we do that, then user is able to override this, which in turn means that if someone will override it in Azure, it will re-enable scheduling on those nodes which cannot be scheduled. Current behavior will be better - since this override will take precedence.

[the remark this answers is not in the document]
wkrysmann: @tmjd I moved it to default. I think it's ready for re-review.

api/v1/installation_types.go
```diff
@@ -129,17 +129,30 @@ type InstallationSpec struct {
...
 	// the affinity expressions specified by this field, but it may choose
 	// a node that violates one or more of the expressions.
 	// +optional
 	PreferredDuringSchedulingIgnoredDuringExecution []v1.PreferredSchedulingTerm `json:"preferredDuringSchedulingIgnoredDuringExecution,omitempty"`
+
+	// WARNING: Please note that if the affinity requirements specified by this field are not met at
+	// scheduling time, the pod will NOT be scheduled onto the node.
+	// There is no fallback to another affinity rules with this setting.
+	// This may cause networking disruption or even catastrophic failure!
+	// PreferredDuringSchedulingIgnoredDuringExecution should be a default choice if
```
tmjd, line 147: I do not think this sentence would be very clear to someone that was not familiar with what we are trying to convey. How about 
```
PreferredDuringSchedulingIgnoredDuringExecution should be used for affinity unless there is a specific well understood reason to use RequiredDuringSchedulingIgnoredDuringExecution and you can guarantee that the RequiredDuringSchedulingIgnoredDuringExecution will always have sufficient nodes to satisfy the requirement.
```

pkg/render/typha.go
```diff
@@ -620,43 +620,18 @@ func (c *typhaComponent) typhaPodSecurityPolicy() *policyv1beta1.PodSecurityPoli
...
+		aff = &v1.Affinity{NodeAffinity: &v1.NodeAffinity{}}
+		// only set if specified to not override the defaults.
+		if c.installation.TyphaAffinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution != nil {
+			aff.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution = c.installation.TyphaAffinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution
+		}
 		// this ensures we still return nil if neither condition is hit.
-		if aff == nil {
-			aff = &v1.Affinity{NodeAffinity: &v1.NodeAffinity{}}
+		if len(c.installation.TyphaAffinity.NodeAffinity.PreferredDuringSchedulingIgnoredDuringExecution) > 0 {
+			aff.NodeAffinity.PreferredDuringSchedulingIgnoredDuringExecution = c.installation.TyphaAffinity.NodeAffinity.PreferredDuringSchedulingIgnoredDuringExecution
```
tmjd, line 633: The `ensure we still return nil` comment is wrong currently and with the moving of the block above for ProviderAKS `RequiredDuring...` being changed to a default, I think perhaps this code in this block should be adjusted a bit.
I think maybe it should change to
```
// this ensures we return nil if no affinity is specified.
if c.installation.TyphaAffinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution == nil && len(c.installation.TyphaAffinity.NodeAffinity.PreferredDuringSchedulingIgnoredDuringExecution) == 0 {
    return nil
}
aff = &v1.Affinity{NodeAffinity: &v1.NodeAffinity{
        RequiredDuringSchedulingIgnoredDuringExecution: c.installation.TyphaAffinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution,
        PreferredDuringSchedulingIgnoredDuringExecution: c.installation.TyphaAffinity.NodeAffinity.PreferredDuringSchedulingIgnoredDuringExecution
    }
}
```
Just to be clear it should still include the `if c.installation.TyphaAffinity != nil && c.installation.TyphaAffinity.NodeAffinity != nil {` around the highlighted block.

tmjd (commented):
Looking good. Two small adjustments and then I think I'll be ready to merge this.
@caseydavenport did you want to take a look at this?

pkg/render/typha.go
wkrysmann: Hey, thanks for comment @tmjd! Addressed as requested.

api/v1/installation_types.go
```diff
@@ -129,17 +129,31 @@ type InstallationSpec struct {
...
 	// NodeAffinity describes node affinity scheduling rules for typha.
 	// +optional
-	NodeAffinity *PreferredNodeAffinity `json:"nodeAffinity,omitempty"`
+	NodeAffinity *TigeraNodeAffinty `json:"nodeAffinity,omitempty"`
 }
 
-// PreferredNodeAffinity is similar to an affinity except it only exposes the Preferred scheduling option,
-// which ensures a pod will still be scheduled even when all the nodes it matches are unschedulable.
-type PreferredNodeAffinity struct {
+// TigeraNodeAffinty is similar to *v1.NodeAffinity, but allows us to limit available shcedulers.
```
caseydavenport, line 135: typo: shcedulers

wkrysmann: Fixed.

tmjd (approved):
LGTM

PR #1289, opened 2021-05-04
don't migrate if the daemonset and deployments are deleting

paulgmiller:
## Description
If calico is deployed with addonmanager there is actuall a race where addon manager will purge before operator comes up. Most of the time this seems okay. Howver if a pod is slow to terminate then the operator will migrate the deleting daemonset and nodes can get stuck pretty easily.  

https://github.com/projectcalico/calico/issues/4525#issuecomment-830115368

Even if it's not the whole solution having operator ignore deleting daemonsets might be better than current behavior. 

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


caseydavenport (approved):
This looks pretty reasonable to me - nice fix!

PR #1293, opened 2021-05-06
Allow Implicit OIDC flow

Brian-McM:
## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/authentication/authentication_controller.go
+++ b/pkg/controller/authentication/authentication_controller.go
@@ -29,6 +29,7 @@ import (
 	"github.com/tigera/operator/pkg/controller/utils/imageset"
 	"github.com/tigera/operator/pkg/render"
 	rmeta "github.com/tigera/operator/pkg/render/common/meta"
+
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -219,7 +220,7 @@ func (r *ReconcileAuthentication) Reconcile(ctx context.Context, request reconci
 	}
 
 	// Dex will be configured with the contents of this secret, such as clientID and clientSecret.
-	idpSecret, err := getIdpSecret(ctx, r.client, authentication)
+	idpSecret, err := utils.GetIdpSecret(ctx, r.client, authentication)
 	if err != nil {
 		log.Error(err, "Invalid or missing identity provider secret")
 		r.status.SetDegraded("Invalid or missing identity provider secret", err.Error())
@@ -245,6 +246,11 @@ func (r *ReconcileAuthentication) Reconcile(ctx context.Context, request reconci
 		return reconcile.Result{}, err
 	}
 
+	disableDex := false
+	if authentication.Spec.OIDC != nil && authentication.Spec.OIDC.Type == oprv1.OIDCTypeTigera {
+		disableDex = true
+	}
+
 	// DexConfig adds convenience methods around dex related objects in k8s and can be used to configure Dex.
 	dexCfg := render.NewDexConfig(install.CertificateManagement, authentication, tlsSecret, dexSecret, idpSecret, r.clusterDomain)
 
@@ -259,6 +265,7 @@ func (r *ReconcileAuthentication) Reconcile(ctx context.Context, request reconci
 		install,
 		dexCfg,
 		r.clusterDomain,
+		disableDex,
 	)
 
 	if err = imageset.ApplyImageSet(ctx, r.client, variant, component); err != nil {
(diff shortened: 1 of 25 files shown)


main.go
```diff
@@ -156,11 +156,12 @@ func main() {
...
-		MetricsBindAddress: metricsAddr(),
-		Port:               9443,
-		LeaderElection:     enableLeaderElection,
-		LeaderElectionID:   "operator-lock",
+		Scheme:                  scheme,
+		MetricsBindAddress:      metricsAddr(),
+		Port:                    9443,
+		LeaderElection:          enableLeaderElection,
+		LeaderElectionID:        "operator-lock",
+		LeaderElectionNamespace: "tigera-operator",
```
Brian-McM, line 164: Needs to be removed

pkg/controller/authentication/authentication_controller.go
```diff
@@ -276,9 +283,13 @@ func (r *ReconcileAuthentication) Reconcile(ctx context.Context, request reconci
 	// Clear the degraded bit if we've reached this far.
 	r.status.ClearDegraded()
 
-	if !r.status.IsAvailable() {
-		// Schedule a kick to check again in the near future.
-		return reconcile.Result{RequeueAfter: 30 * time.Second}, nil
+	// TODO <before PR is closed> figure out a better way to handle the authentication resource not being "available"
```
ozdanborne, line 286: What's the status of this TODO?

pkg/controller/installation/core_controller.go
```diff
@@ -864,6 +864,14 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 		}
 	}
 
+	// If the Elasticsearch license type is basic OR the there is an Authentication CR with OIDC type Tigera, then
```
ozdanborne, line 867: caught a small typo
```suggestion
	// If the Elasticsearch license type is basic OR there is an Authentication CR with OIDC type Tigera, then
```

pkg/controller/utils/auth.go
```diff
@@ -0,0 +1,98 @@
...
+func GetKeyValidatorConfig(ctx context.Context, cli client.Client, authenticationCR *operatorv1.Authentication, clusterDomain string) (rauth.KeyValidatorConfig, error) {
+	var keyValidatorConfig rauth.KeyValidatorConfig
+	if authenticationCR != nil {
+		idpSecret, err := GetIdpSecret(ctx, cli, authenticationCR)
+		if err != nil {
+			return nil, err
+		}
+
+		oidc := authenticationCR.Spec.OIDC
+		if authenticationCR.Spec.OIDC != nil && authenticationCR.Spec.OIDC.Type == operatorv1.OIDCTypeTigera {
```
ozdanborne, line 31: You created this nice alias but then you hardly used it!
```suggestion
		if oidc != nil && oidc.Type == operatorv1.OIDCTypeTigera {
```

ozdanborne (changes requested):
A couple nits but overall looking good

pkg/controller/authentication/authentication_controller.go
Brian-McM: I think I know what I'm going to do, just need to confirm with Casey. I think we may just have another function for the status manager to say there's no resources to monitor.

pkg/controller/authentication/authentication_controller.go
```diff
@@ -259,6 +265,7 @@ func (r *ReconcileAuthentication) Reconcile(ctx context.Context, request reconci
 		install,
 		dexCfg,
 		r.clusterDomain,
+		disableDex,
```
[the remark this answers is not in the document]
Brian-McM: That would be nice, but the problem is we need to remove the dex components, and the easiest way to do that right now is to just create all the components with the dex config and delete them instead of creating them.

I'm not the _biggest_ fan of this solution, but other alternatives are either uglier with no benefit over this or are nicer but may require a little more refactoring of Dex.

pkg/render/common/authentication/types.go
```diff
@@ -0,0 +1,97 @@
...
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"strings"
+
+	corev1 "k8s.io/api/core/v1"
+)
+
+type KeyValidatorConfig interface {
```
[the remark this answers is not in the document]
Brian-McM: Yes we absolutely could, and should do something like this. I haven't done it yet just bring this PR done to exactly what needs to change to get this working. Thanks for reminding me of this.

pkg/render/common/authentication/tigera/key_validator_config/key_validator_config.go
```diff
@@ -0,0 +1,133 @@
...
+
+func (kvc *KeyValidatorConfig) ClientID() string {
+	return kvc.clientID
+}
+
+func (kvc *KeyValidatorConfig) RequiredEnv(prefix string) []corev1.EnvVar {
+	return []corev1.EnvVar{
+		{Name: fmt.Sprintf("%sOIDC_AUTH_ENABLED", prefix), Value: strconv.FormatBool(true)},
+		{Name: fmt.Sprintf("%sOIDC_AUTH_ISSUER", prefix), Value: kvc.Issuer()},
+		{Name: fmt.Sprintf("%sOIDC_AUTH_URL", prefix), Value: kvc.Issuer()},
```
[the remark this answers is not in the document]
Brian-McM: I'll double check this but I'm hesitant to remove this since this is what dex has been using, it may be that they can be the same but their not always?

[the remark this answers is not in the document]
Brian-McM: It actually doesn't look like this is an env variable for any project (i.e. voltron, es-proxy...)....

pkg/render/manager.go
```diff
@@ -495,13 +513,19 @@ func (c *managerComponent) managerContainer() corev1.Container {
...
-			{Name: "CNX_WEB_OIDC_AUTHORITY", Value: fmt.Sprintf("%s/dex", c.dexCfg.ManagerURI())},
-			{Name: "CNX_WEB_OIDC_CLIENT_ID", Value: DexClientId}}
+			{Name: "CNX_WEB_OIDC_CLIENT_ID", Value: c.keyValidatorConfig.ClientID()}}
+
+		switch c.keyValidatorConfig.(type) {
+		case *DexKeyValidatorConfig:
+			envs = append(envs, corev1.EnvVar{Name: "CNX_WEB_OIDC_AUTHORITY", Value: c.keyValidatorConfig.Issuer()})
+		case *tigerakvc.KeyValidatorConfig:
+			envs = append(envs, corev1.EnvVar{Name: "CNX_WEB_OIDC_AUTHORITY", Value: ""})
+		}
```
ozdanborne, line 528: I think this could be pretty easy to whip up a test for if you think it'd be valuable, given what you told me about how easy it is in go to accidentally match a struct with the wrong interface if one interface is a subset of another.

the test would create a managerComponent with each keyValidatorConfig, then call `managerOauth2EnvVars()` and assert that `CNX_WEB_OIDC_AUTHORITY` is called correctly

PR #1297, opened 2021-05-11
Handle no resources to monitor in the status manager

Brian-McM:
This commit adds the ReadyToMonitor function to the StatusManager, and doesn't return false for IsAvailable if "progressing" and "failing" or nil. The purpose of ReadyToMonitor is to tell the status manager that it's OK to start reporting the available state of the resources it's been told about. If it hasn't been told about any resources, then the state is available (so long as there's no explicit degraded reason).

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/status/status.go
+++ b/pkg/controller/status/status.go
@@ -74,6 +74,7 @@ type StatusManager interface {
 	IsAvailable() bool
 	IsProgressing() bool
 	IsDegraded() bool
+	ReadyToMonitor()
 }
 
 type statusManager struct {
@@ -96,6 +97,11 @@ type statusManager struct {
 	// Keep track of currently calculated status.
 	progressing []string
 	failing     []string
+
+	// readyToMonitor tells the status manager that it's ready to monitor the resources that it's been told to monitor,
+	// if there are any, and report statuses based on the state of those resources.
+	readyToMonitor bool
+	hasSynced      bool
 }
 
 func New(client client.Client, component string, kubernetesVersion *common.VersionInfo) StatusManager {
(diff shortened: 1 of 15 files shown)


pkg/controller/status/status.go
```diff
@@ -312,8 +337,7 @@ func (m *statusManager) IsProgressing() bool {
 	m.lock.Lock()
 	defer m.lock.Unlock()
 
-	if m.progressing == nil || m.failing == nil {
```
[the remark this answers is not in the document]
caseydavenport: `if m.progressing == nil || m.failing == nil || !m.readyToMonitor`

The way the code is right now, it probably doesn't matter, since syncState will always complete. But the function contract doesn't guarantee that will always be the case (or has always been the case).

pkg/controller/status/status.go
```diff
@@ -324,14 +348,14 @@ func (m *statusManager) IsDegraded() bool {
...
 
-	// Controllers can explicitly set us degraded.
+	// Controllers can explicitly set us degraded, which can be set even before we tell the status manager that it
+	// should start monitoring resources.
 	if m.degraded {
 		return true
 	}
 
 	// Otherwise, we might be degraded due to failing pods.
-	if m.progressing == nil || m.failing == nil {
```
caseydavenport, line 333: and here as well

pkg/controller/status/status.go
```diff
@@ -443,21 +467,9 @@ func (m *statusManager) syncState() bool {
 		}
 	}
 
-	if len(m.deployments)+len(m.daemonsets)+len(m.statefulsets)+len(m.cronjobs) > 0 {
```
caseydavenport, line 446: Oh haha, I see that you changed this to align with the other changes. That makes more sense.

pkg/controller/utils/component.go
```diff
@@ -190,6 +190,10 @@ func (c componentHandler) CreateOrUpdateOrDelete(ctx context.Context, component
 	}
 
 	cmpLog.V(1).Info("Done reconciling component")
+	// TODO Get each controller to explicitly call ReadyToMonitor on the status manager instead of doing it here.
```
caseydavenport, line 193: Is this really a TODO? I think it's fine to do it here.

pkg/controller/status/status.go
Brian-McM: > That is effectively a "Has syncState() completed?" check

If that's what we want then an _explicit_ "has sync run" field feels more appropriate. This isn't easily understood by reading this, a dev wouldn't know easily that progressing and failing being false is a signal that this hasn't run. It also means that a change to this behaviour in the sync function breaks everything, so it's a bit of a fragile signal.

caseydavenport: > ideally the other functions wouldn't need to know about the internals of how syncState works to know that it's complete, but it is what it is.

Yep, this is what I was hinting at with this vague comment. Totally agree.

caseydavenport (approved):

PR #1298, opened 2021-05-12
Adjust fluentd probe timeouts

lmm:
## Description

This PR fixes some issues with the fluentd probes, mostly for Windows
- The startup probe on Windows can fail due to slow startup. This PR doubles the startup probe period and timeout to 20s.
- The readiness probe period was not set so it defaults to 10s while the timeout is 5s.
- Sets the failure threshold explicitly
- Adds some tests

I've tested these changes on a cluster with somewhat underprovisioned Windows nodes.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/fluentd.go
+++ b/pkg/render/fluentd.go
@@ -62,8 +62,14 @@ const (
 	SplunkFluentdDefaultCertDir              = "/etc/ssl/splunk/"
 	SplunkFluentdDefaultCertPath             = SplunkFluentdDefaultCertDir + SplunkFluentdSecretCertificateKey
 
-	ProbeTimeoutSeconds = 5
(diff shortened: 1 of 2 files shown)


pkg/render/fluentd.go
```diff
@@ -103,6 +109,13 @@ func Fluentd(
 	clusterDomain string,
 	osType rmeta.OSType,
 ) Component {
+	timeout := int32(probeTimeoutSeconds)
```
rene-dekker, line 112: The int32() seems redundant here.

rene-dekker (approved):
lgtm

Brian-McM (approved):

Issue #1246, opened 2021-03-29
Support for arm64 architecture


cguertin14:
## Expected Behavior

Tigera-operator behaving normally and running on arm64 clusters as intended.

## Current Behavior

Cannot use tigera-operator on arm64 cluster because of incompatibility with arm64.

## Possible Solution

The solution would be to deploy a tigera-operator arm64 capable Docker image to the https://quay.io/repository/tigera/operator registry.

## Steps to Reproduce (for bugs)

1. Download Tigera Operator's YAML configuration: https://docs.projectcalico.org/manifests/tigera-operator.yaml
2. Apply the manifests on an arm64 server/computer
3. Notice the operator not working because of incompatibility.

## Context

I recently started a Raspberry Pi cluster at home to play with k3s and bare-metal Kubernetes. For my CNI, I chose Calico because I heard great things about it, and then tigera-operator failed to run on my cluster. But, Calico still works and i'm not sure if i need tigera operator at all, but the guide I followed (https://docs.projectcalico.org/getting-started/kubernetes/k3s/quickstart) featured it.

## Your Environment

* Operating System and version: Ubuntu Server 20.04, running on 4 different Raspberry Pi 4b
* Link to your project (optional): https://github.com/cguertin14/infra

tmjd:
Thank you for your interest in Calico and the operator. It would be great if the operator could install other architectures. If you are interested in submitting changes to be able to build an arm64 version I would be happy to review.

For your purposes I think you will be fine using the "Manifest" tab under "Install Calico".

cguertin14:
> Thank you for your interest in Calico and the operator. It would be great if the operator could install other architectures. If you are interested in submitting changes to be able to build an arm64 version I would be happy to review.
> 
> 
> 
> For your purposes I think you will be fine using the "Manifest" tab under "Install Calico".

Sweet! I'll look into it and submit a PR probably this week then 😀

jsievenpiper:
@tmjd I took a very crude stab at providing support -- would love your input when you get the time!

HaveFun83:
Looks like you already have done something with arm64 images according to your makefile?
https://github.com/tigera/operator/blob/a3756a531ee51666cf7a76934f7ee049d8734868/Makefile#L33

frozenprocess:
I believe https://github.com/tigera/operator/pull/1464 and https://github.com/tigera/operator/pull/1469 address this issue.

An early operator image can be find here for testing and hopefully endup in the next release.
```
[...]

PR #1310, opened 2021-05-19
Make Docker Image Multi-Arch Compatible

jsievenpiper:
## Description

Addresses #1246 This PR attempts to take a first crack at making the docker image for `tigera/operator` multi-arch compatible. This is my first time in the community here (👋🏾 hi, by the way!), so I don't have a lot of insight into how things work around here!

I took the path of least resistance: since we're building a simple `FROM scratch` image for the most part, I took the `Dockerfile` and made just a single one with an ARCH arg for building. Then, instead of deriving `ARCHES` from the host machine, I put a static list of "supported" (i.e. just the ones I was interested in...) arches. `make images-all` then produces two tagged images: `tigera/operator:latest-amd64` and `tigera/operator:latest-arm64`.

There's a couple issues I need some feedback on:
- I don't have a lot of experience with `manifest-tool`: will it properly set the architecture for the images? Without heavy refactoring to use `docker buildx`, I don't have a ton of experience in how to make `docker build` write out the `arm64` image as actually being for `arm64` (even if the binary itself build from go _is_).
- How would you like the final manifest to be built? I removed the architecture-less generic tag from the `Makefile` in preparation of making the fat manifest, but that's where my inexperience with `manifest-tool` comes in.

Thanks!

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -205,7 +205,7 @@ $(BINDIR)/operator-$(ARCH): $(SRC_FILES)
 	mkdir -p $(BINDIR)
 	$(CONTAINERIZED) \
 	sh -c '$(GIT_CONFIG_SSH) \
-	go build -v -i -o $(BINDIR)/operator-$(ARCH) -ldflags "-X $(PACKAGE_NAME)/version.VERSION=$(GIT_VERSION) -s -w" ./main.go'
+	GOARCH=$(ARCH) go build -v -i -o $(BINDIR)/operator-$(ARCH) -ldflags "-X $(PACKAGE_NAME)/version.VERSION=$(GIT_VERSION) -s -w" ./main.go'
 
 .PHONY: image
 image: build $(BUILD_IMAGE)
--- a/Makefile
+++ b/Makefile
@@ -216,9 +216,6 @@ sub-image-%:
 $(BUILD_IMAGE): $(BUILD_IMAGE)-$(ARCH)
 $(BUILD_IMAGE)-$(ARCH): $(BINDIR)/operator-$(ARCH)
 	docker build --pull -t $(BUILD_IMAGE):latest-$(ARCH) --build-arg GIT_VERSION=$(GIT_VERSION) --build-arg ARCH=$(ARCH) -f ./build/Dockerfile .
-ifeq ($(ARCH),amd64)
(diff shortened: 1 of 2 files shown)


frozenprocess:
Hey @jsievenpiper,
I might be able to help with this PR if you are interested.

``` Don't quote me on explanations :D ```
> I don't have a lot of experience with manifest-tool: will it properly set the architecture for the images?

Yes, by using `platform` it can.
https://docs.docker.com/registry/spec/manifest-v2-2/#manifest-list
If you like to know more about it I suggest https://integratedcode.us/2016/04/22/a-step-towards-multi-platform-docker-images/

> Without heavy refactoring to use docker buildx, I don't have a ton of experience in how to make docker build write out the arm64 image as actually being for arm64 (even if the binary itself build from go is).

I think docker buildx might not be the solution here since it can complicate CI/CD compilations. Judging by other projects (Calico Node, go-build ...) I think we should work toward implementing `qemu-static-user` with your approach. 

https://github.com/tigera/operator/blob/95578c812912d420d6c18eb11959f7e1fe33c55d/Makefile#L7

I'm a big fan of shortcuts * *Cough cheating* * . let's take look at another project that already got this implemented.
https://github.com/projectcalico/go-build/blob/5c653ae9f5b76aab0c3d0574c33aeed57b64a87f/Makefile#L84-L88


> How would you like the final manifest to be built? I removed the architecture-less generic tag from the Makefile in preparation for making the fat manifest, but that's where my inexperience with manifest-tool comes in.

Judging by other projects (calico-node,go-build ...) I would say duplicate the `dockerfile.amd64` and replace the suffix with target architecture `dockerfile.arm64` adjust the images and that would be it.

frozenprocess:
Hey @jsievenpiper,
I took stab at this problem and seems like it is working. If you are still intersted all the changes can be found @ https://github.com/tigera/operator/pull/1464

jsievenpiper:
Hey @frozenprocess great to hear! I ended up installing on my cluster manually without the operator and this fell off my radar for a long time. But glad to see this will help people in the future! I'll go ahead and close this one out.

PR #1322, opened 2021-05-25
updates for cloud-dev portal

ozdanborne:
## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/manager/manager_controller.go
+++ b/pkg/controller/manager/manager_controller.go
@@ -138,6 +138,10 @@ func add(mgr manager.Manager, c controller.Controller) error {
 		return fmt.Errorf("compliance-controller failed to watch the ConfigMap resource: %w", err)
 	}
 
+	if err = utils.AddConfigMapWatch(c, render.CloudManagerConfigOverrideName, rmeta.OperatorNamespace()); err != nil {
+		return fmt.Errorf("manager-controller failed to watch the ConfigMap resource: %v", err)
+	}
+
 	if err = utils.AddNetworkWatch(c); err != nil {
 		return fmt.Errorf("manager-controller failed to watch Network resource: %w", err)
 	}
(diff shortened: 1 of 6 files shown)


pkg/controller/manager/manager_controller.go
```diff
@@ -476,6 +481,20 @@ func (r *ReconcileManager) Reconcile(ctx context.Context, request reconcile.Requ
 		}
 	}
 
+	// Cloud modifications
+	mgrCfg := &corev1.ConfigMap{}
+	key := types.NamespacedName{Name: "cloud-manager-config", Namespace: rmeta.OperatorNamespace()}
```
tmjd, line 486: Should use `render.CloudManagerConfigOverrideName` here.

pkg/controller/manager/manager_controller.go
```diff
@@ -476,6 +481,20 @@ func (r *ReconcileManager) Reconcile(ctx context.Context, request reconcile.Requ
...
+	if err = r.client.Get(ctx, key, mgrCfg); err != nil {
+		if !errors.IsNotFound(err) {
+			return reconcile.Result{}, fmt.Errorf("Failed to read cloud-manager-config ConfigMap: %s", err.Error())
+		}
+	} else {
+		if err = json.Unmarshal([]byte(mgrCfg.Data["config"]), &render.CloudManagerConfigOverrides); err != nil {
```
tmjd, line 492: I don't think there is any reason to involve a json.Unmarshall here. We should just store the portalAPIURL directly in the ConfigMap and then get it with
```
render.CloudManagerConfigOverrides = mgrCfg.Data["portalAPIURL"]
```
and then render.CloudManagerConfigOverrides is a string.
I think you copied this from the kibana config but that was json because for kibana we needed map. (now that I think about it, maybe I should have handled that differently for kibana and avoided json)

pkg/render/manager.go
```diff
@@ -528,13 +537,20 @@ func (c *managerComponent) managerOAuth2EnvVars() []v1.EnvVar {
 	} else {
 		envs = []corev1.EnvVar{
 			{Name: "CNX_WEB_AUTHENTICATION_TYPE", Value: "OIDC"},
-			{Name: "CNX_WEB_OIDC_CLIENT_ID", Value: c.keyValidatorConfig.ClientID()}}
+			{Name: "CNX_WEB_OIDC_CLIENT_ID", Value: c.keyValidatorConfig.ClientID()},
+
+			// todo: remove this once manager correctly reads well-known-config from root of local domain
```
tmjd, line 542: Is this todo and the one below expected to come in before the next release?

pkg/controller/manager/manager_controller.go
ozdanborne: Oh! Perfect, will do. thanks

pkg/render/manager.go
ozdanborne: I'm not sure. It's not a blocker so UI team may punt to after CC release

tmjd (approved):
LGTM

PR #1327, opened 2021-05-26
Hopefully more consistent gen-versions check

tmjd:
Run gen-versions and do dirty-check instead of trying to determine if
only one of the version files has been modified but not the other.

## Description

The previous method to check if the version files were out of sync didn't work for release branches, that could have been fixed but I think simply running `gen-versions` and then checking for dirty files will be more likely to catch future changes.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


hack/gen-versions/validate.sh
```diff
@@ -1,40 +0,0 @@
...
-# See the License for the specific language governing permissions and
-# limitations under the License.
-
-set -e
-
-git fetch origin
-
-function modded() {
-    file=$1
-    git diff --name-only HEAD..$(git merge-base HEAD origin/master) $file
```
ozdanborne, line 22: I think your changes are correct to remove this file and rely on dirty checks. But I just remembered why we were comparing head and wanted to explain for posterity.

This check assumes that any commit which modifies the versions.yml must also modify the rendered components, and vice versa. So it it checks that both files were _changed_. 

However, this doesn't  ensure that they're up to date. A user could, for example, make an unrelated change to each file but this test would still pass. The render-then-check-dirty you've implemented does, so 💯

Makefile
```diff
@@ -338,10 +338,11 @@ foss-checks:
 ###############################################################################
 .PHONY: ci
 ## Run what CI runs
-ci: clean format-check images test $(BINDIR)/gen-versions validate-gen-versions dirty-check
+ci: clean format-check images test dirty-check validate-gen-versions
```
ozdanborne, line 341: It seems a little redundant to have `dirty-check` here and also inside `validate-gen-versions` but I don't know enough about makefile execution order to suggest a working alternative so lgtm

ozdanborne (approved):

Makefile
tmjd: Yeah it is certainly redundant in the ci target but I thought it would be better to dirty-check specifically in validate-gen-versions so it would be obvious that the dirty was due to the gen-versions. It also allows the validate-gen-versions target to be a standalone check so if ci was changed to remove the dirty-check the validate-gen-versions would still catch a problem.

PR #1336, opened 2021-06-02
Allow calico-node to update packetcaptures

asincu:
## Description

Enhance calico-node clusterrole to update packetures CRs with their status.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/node.go
+++ b/pkg/render/node.go
@@ -384,6 +384,14 @@ func (c *nodeComponent) nodeRole() *rbacv1.ClusterRole {
 				},
 				Verbs: []string{"create"},
 			},
+			{
+				// Tigera Secure updates status for packet captures.
+				APIGroups: []string{"crd.projectcalico.org"},
+				Resources: []string{
+					"packetcaptures",
+				},
+				Verbs: []string{"update"},
+			},
 		}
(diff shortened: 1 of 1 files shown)


rene-dekker (approved):
lgtm

pkg/render/node.go
```diff
@@ -384,6 +384,14 @@ func (c *nodeComponent) nodeRole() *rbacv1.ClusterRole {
 				},
 				Verbs: []string{"create"},
 			},
+			{
+				// Tigera Secure updates status for packet captures.
+				APIGroups: []string{"crd.projectcalico.org"},
+				Resources: []string{
+					"packetcaptures",
```
tmjd, line 391: Would `packetcaptures/status` work (like `pod/status` at Line 281)?

asincu: Not really. I tried using only packetcapture/status, but it did not work.

tmjd: As you pointed out in slack, libcalico-go doesn't provide a Status only update mechanism so I think that is why this isn't possible. Bummer.

tmjd (approved):
LGTM

PR #1337, opened 2021-06-03
Always manage CNI config from calico/node

caseydavenport:
## Description

Part of the fix for https://github.com/projectcalico/node/issues/391

We'll also need to update manifests.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/node.go
+++ b/pkg/render/node.go
@@ -943,6 +943,10 @@ func (c *nodeComponent) nodeVolumeMounts() []v1.VolumeMount {
 		nodeVolumeMounts = append(nodeVolumeMounts, cniLogMount)
 	}
 
+	if c.cr.CNI.Type == operator.PluginCalico {
+		nodeVolumeMounts = append(nodeVolumeMounts, v1.VolumeMount{MountPath: "/host/etc/cni/net.d", Name: "cni-net-dir"})
+	}
+
 	if c.birdTemplates != nil {
 		// create a volume mount for each bird template, but sort them alphabetically first,
 		// otherwise, since map iteration is random, they'll be added to the list of volumes in a random order,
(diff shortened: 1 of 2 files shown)


pkg/render/node.go
```diff
@@ -929,6 +929,7 @@ func (c *nodeComponent) nodeVolumeMounts() []v1.VolumeMount {
 		{MountPath: "/var/run/nodeagent", Name: "policysync"},
 		{MountPath: "/typha-ca", Name: "typha-ca", ReadOnly: true},
 		{MountPath: "/felix-certs", Name: "felix-certs", ReadOnly: true},
+		{MountPath: "/host/etc/cni/net.d", Name: "cni-net-dir"},
```
lmm, line 932: Do we want to add this only if Calico CNI is enabled, e.g. https://github.com/tigera/operator/blob/02072527228c00328410e747856922dfce67b845/pkg/render/node.go#L414

pkg/render/node.go
```diff
@@ -1021,6 +1022,7 @@ func (c *nodeComponent) nodeEnvVars() []v1.EnvVar {
 		{Name: "WAIT_FOR_DATASTORE", Value: "true"},
 		{Name: "CLUSTER_TYPE", Value: clusterType},
 		{Name: "CALICO_DISABLE_FILE_LOGGING", Value: "true"},
+		{Name: "CALICO_MANAGE_CNI", Value: "true"},
```
lmm, line 1025: Do we need to handle clusters migrating from where `CALICO_MANAGE_CNI` is false?
I guess if the Calico CNI monitor-token loop was running, it would could just run and not do anything.

lmm (commented):
I'm not super familiar with the `BoundServiceAccountTokenVolume` stuff but everything looks reasonable. I just added some questions.

pkg/render/node.go
caseydavenport: yeah good call

pkg/render/node.go
caseydavenport: yeah, migration should be fine and shouldn't have an impact.

lmm (approved):

PR #1341, opened 2021-06-10
Implement Operator logic to install ES Gateway

vberezny:
## Description
This PR enables Operator to install ES Gateway. The ES Gateway sits behind the `tigera-secure-es-http` and `tigera-secure-kb-http` services (previously created by eckOperator/kibana, now managed by the Operator directly) and proxies traffic to the `tigera-secure-internal-es-http` and `tigera-secure-internal-kb-http` services (the services that are now created by eckOperator/kibana).

## TODO

- [x] Unit tests.
- [x] Docs PR for network policy update.
- [x] Propagating the cert to a Managed cluster when there is version skew.
- [x] Readiness Probe

### DIFF
--- a/pkg/components/enterprise.go
+++ b/pkg/components/enterprise.go
@@ -82,6 +82,11 @@ var (
 		Image:   "tigera/es-proxy",
 	}
 
+	ComponentESGateway = component{
+		Version: "master",
+		Image:   "tigera/es-gateway",
+	}
+
 	ComponentFluentd = component{
 		Version: "master",
 		Image:   "tigera/fluentd",
@@ -204,5 +209,6 @@ var (
 		ComponentTigeraCNI,
 		ComponentCloudControllers,
 		ComponentElasticsearchMetrics,
+		ComponentESGateway,
 	}
 )
--- a/pkg/controller/installation/core_controller.go
+++ b/pkg/controller/installation/core_controller.go
@@ -905,11 +905,9 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 		}
 	}
 
-	var esAdminSecret *v1.Secret
+	var kubeControllersGatewaySecret *v1.Secret
 	if instance.Spec.Variant == operator.TigeraSecureEnterprise {
-		// Kube controllers needs the admin secret copied to it's namespace as it has administrative tasks to run on
-		// Elasticsearch.
-		esAdminSecret, err = utils.GetSecret(ctx, r.client, render.ElasticsearchAdminUserSecret, rmeta.OperatorNamespace())
+		kubeControllersGatewaySecret, err = utils.GetSecret(ctx, r.client, render.ElasticsearchKubeControllersUserSecret, rmeta.OperatorNamespace())
 		if err != nil {
 			return reconcile.Result{}, err
 		}
@@ -1058,7 +1056,7 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 		nodeAppArmorProfile,
 		r.clusterDomain,
 		enableESOIDCWorkaround,
-		esAdminSecret,
+		kubeControllersGatewaySecret,
 		kubeControllersMetricsPort,
 		nodeReporterMetricsPort,
 		bgpLayout,
--- a/pkg/controller/intrusiondetection/intrusiondetection_controller.go
+++ b/pkg/controller/intrusiondetection/intrusiondetection_controller.go
@@ -133,7 +133,7 @@ func add(mgr manager.Manager, c controller.Controller) error {
 	for _, secretName := range []string{
 		relasticsearch.PublicCertSecret, render.ElasticsearchIntrusionDetectionUserSecret,
 		render.ElasticsearchIntrusionDetectionJobUserSecret, render.ElasticsearchADJobUserSecret,
-		render.KibanaPublicCertSecret, render.ManagerInternalTLSSecretName,
+		render.ManagerInternalTLSSecretName,
 	} {
 		if err = utils.AddSecretsWatch(c, secretName, rmeta.OperatorNamespace()); err != nil {
 			return fmt.Errorf("intrusiondetection-controller failed to watch the Secret resource: %v", err)
@@ -141,6 +141,11 @@ func add(mgr manager.Manager, c controller.Controller) error {
 	}
 
 	if err = utils.AddSecretsWatch(c, render.ManagerInternalTLSSecretName, render.IntrusionDetectionNamespace); err != nil {
+		return fmt.Errorf("intrusiondetection-controller failed to watch the Secret resource: %v", err)
+	}
+
+	// These watches are here to catch a modification to the resources we create in reconcile so the changes would be corrected.
+	if err = utils.AddSecretsWatch(c, relasticsearch.PublicCertSecret, render.IntrusionDetectionNamespace); err != nil {
 		return fmt.Errorf("intrusiondetection-controller failed to watch the Secret resource: %v", err)
 	}
 
--- a/pkg/controller/logstorage/logstorage_controller.go
+++ b/pkg/controller/logstorage/logstorage_controller.go
@@ -18,9 +18,12 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/tigera/operator/pkg/crypto"
+
 	cmnv1 "github.com/elastic/cloud-on-k8s/pkg/apis/common/v1"
 	esv1 "github.com/elastic/cloud-on-k8s/pkg/apis/elasticsearch/v1"
 	kbv1 "github.com/elastic/cloud-on-k8s/pkg/apis/kibana/v1"
+	"golang.org/x/crypto/bcrypt"
 
 	operatorv1 "github.com/tigera/operator/api/v1"
 	"github.com/tigera/operator/pkg/controller/options"
@@ -32,6 +35,7 @@ import (
 	relasticsearch "github.com/tigera/operator/pkg/render/common/elasticsearch"
 	rmeta "github.com/tigera/operator/pkg/render/common/meta"
 	rsecret "github.com/tigera/operator/pkg/render/common/secret"
+	"github.com/tigera/operator/pkg/render/logstorage/esgateway"
 	"github.com/tigera/operator/pkg/render/logstorage/esmetrics"
 
 	apps "k8s.io/api/apps/v1"
@@ -167,10 +171,19 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 	// Watch all the secrets created by this controller so we can regenerate any that are deleted
 	for _, secretName := range []string{
 		render.TigeraElasticsearchCertSecret, render.TigeraKibanaCertSecret,
-		render.OIDCSecretName, render.DexObjectName} {
+		render.OIDCSecretName, render.DexObjectName, relasticsearch.PublicCertSecret} {
 		if err = utils.AddSecretsWatch(c, secretName, rmeta.OperatorNamespace()); err != nil {
 			return fmt.Errorf("log-storage-controller failed to watch the Secret resource: %w", err)
 		}
+	}
+
+	// Catch if something modifies the certs that this controller creates.
+	if err = utils.AddSecretsWatch(c, relasticsearch.PublicCertSecret, render.ElasticsearchNamespace); err != nil {
+		return fmt.Errorf("log-storage-controller failed to watch the Secret resource: %w", err)
+	}
+
+	if err = utils.AddSecretsWatch(c, render.TigeraElasticsearchInternalCertSecret, render.ElasticsearchNamespace); err != nil {
+		return fmt.Errorf("log-storage-controller failed to watch the Secret resource: %w", err)
 	}
 
 	if err = utils.AddConfigMapWatch(c, relasticsearch.ClusterConfigConfigMapName, rmeta.OperatorNamespace()); err != nil {
@@ -182,6 +195,10 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 	}
 
 	if err := utils.AddServiceWatch(c, render.ElasticsearchServiceName, render.ElasticsearchNamespace); err != nil {
+		return fmt.Errorf("log-storage-controller failed to watch the Service resource: %w", err)
+	}
+
+	if err := utils.AddServiceWatch(c, esgateway.ServiceName, render.ElasticsearchNamespace); err != nil {
 		return fmt.Errorf("log-storage-controller failed to watch the Service resource: %w", err)
 	}
 
@@ -404,7 +421,8 @@ func (r *ReconcileLogStorage) Reconcile(ctx context.Context, request reconcile.R
 		return reconcile.Result{}, err
 	}
 
-	var esCertSecret, esCertSecretESCopy, esPubCertSecret, esAdminUserSecret *corev1.Secret
+	var gatewayCertSecret, publicCertSecret *corev1.Secret
+	var esInternalCertSecret, esAdminUserSecret, esCertSecret *corev1.Secret
 	var kibanaSecrets, curatorSecrets []*corev1.Secret
 	var clusterConfig *relasticsearch.ClusterConfig
 	var esLicenseType render.ElasticsearchLicenseType
@@ -427,6 +445,12 @@ func (r *ReconcileLogStorage) Reconcile(ctx context.Context, request reconcile.R
 			return reconcile.Result{}, nil
 		}
 
+		if gatewayCertSecret, publicCertSecret, err = r.getESGatewayCertificateSecrets(ctx, install); err != nil {
+			reqLogger.Error(err, err.Error())
+			r.status.SetDegraded("Failed to create Elasticsearch Gateway secrets", err.Error())
+			return reconcile.Result{}, err
+		}
+
 		// Get the admin user secret to copy to the operator namespace.
 		esAdminUserSecret, err = utils.GetSecret(ctx, r.client, render.ElasticsearchAdminUserSecret, render.ElasticsearchNamespace)
 		if err != nil {
@@ -436,13 +460,13 @@ func (r *ReconcileLogStorage) Reconcile(ctx context.Context, request reconcile.R
 			esAdminUserSecret = rsecret.CopyToNamespace(rmeta.OperatorNamespace(), esAdminUserSecret)[0]
 		}
 
-		if esCertSecret, esCertSecretESCopy, esPubCertSecret, err = r.getElasticsearchCertificateSecrets(ctx, install); err != nil {
+		if esCertSecret, esInternalCertSecret, err = r.getElasticsearchCertificateSecrets(ctx, install); err != nil {
 			reqLogger.Error(err, err.Error())
-			r.status.SetDegraded("Failed to create elasticsearch secrets", err.Error())
+			r.status.SetDegraded("Failed to create Elasticsearch secrets", err.Error())
 			return reconcile.Result{}, err
 		}
 
-		if kibanaSecrets, err = r.kibanaSecrets(ctx, install); err != nil {
+		if kibanaSecrets, err = r.kibanaInternalSecrets(ctx, install); err != nil {
 			reqLogger.Error(err, err.Error())
 			r.status.SetDegraded("Failed to create kibana secrets", err.Error())
 			return reconcile.Result{}, err
@@ -526,7 +550,7 @@ func (r *ReconcileLogStorage) Reconcile(ctx context.Context, request reconcile.R
 		elasticsearch,
 		kibana,
 		clusterConfig,
-		[]*corev1.Secret{esCertSecret, esCertSecretESCopy, esPubCertSecret, esAdminUserSecret},
+		[]*corev1.Secret{esCertSecret, esInternalCertSecret, esAdminUserSecret},
 		kibanaSecrets,
 		pullSecrets,
 		r.provider,
@@ -559,6 +583,42 @@ func (r *ReconcileLogStorage) Reconcile(ctx context.Context, request reconcile.R
 		if kibana == nil || kibana.Status.AssociationStatus != cmnv1.AssociationEstablished {
 			r.status.SetDegraded("Waiting for Kibana cluster to be operational", "")
 			return reconcile.Result{}, nil
+		}
+
+		kibanaInternalCertSecret, err := utils.GetSecret(ctx, r.client, render.KibanaInternalCertSecret, rmeta.OperatorNamespace())
+		if err != nil {
+			reqLogger.Error(err, err.Error())
+			r.status.SetDegraded("Waiting for internal Kibana tls certificate secret to be available", "")
+			return reconcile.Result{}, err
+		}
+
+		kubeControllersGatewaySecret, kubeControllersVerificationSecret, kubeControllersSecureUserSecret, err := r.createKubeControllersSecrets(ctx, esAdminUserSecret)
+		if err != nil {
+			reqLogger.Error(err, err.Error())
+			r.status.SetDegraded("Failed to create kube-controllers secrets for Elasticsearch gateway", "")
+			return reconcile.Result{}, err
+		}
+
+		esGatewayComponent := esgateway.EsGateway(
+			install,
+			pullSecrets,
+			[]*corev1.Secret{gatewayCertSecret, publicCertSecret},
+			[]*corev1.Secret{kubeControllersGatewaySecret, kubeControllersVerificationSecret, kubeControllersSecureUserSecret},
+			kibanaInternalCertSecret,
+			esInternalCertSecret,
+			r.clusterDomain,
+		)
+
+		if err = imageset.ApplyImageSet(ctx, r.client, variant, esGatewayComponent); err != nil {
+			reqLogger.Error(err, "Error with images from ImageSet")
+			r.status.SetDegraded("Error with images from ImageSet", err.Error())
+			return reconcile.Result{}, err
+		}
+
+		if err := hdler.CreateOrUpdateOrDelete(ctx, esGatewayComponent, r.status); err != nil {
+			reqLogger.Error(err, err.Error())
+			r.status.SetDegraded("Error creating / updating resource", err.Error())
+			return reconcile.Result{}, err
 		}
 
 		if len(curatorSecrets) == 0 {
@@ -600,12 +660,16 @@ func (r *ReconcileLogStorage) Reconcile(ctx context.Context, request reconcile.R
 			return reconcile.Result{}, nil
 		}
 
-		if esPubCertSecret == nil {
+		publicCertSecretESCopy, err := utils.GetSecret(context.Background(), r.client, relasticsearch.PublicCertSecret, render.ElasticsearchNamespace)
+		if err != nil {
+			r.status.SetDegraded("Failed to retrieve Elasticsearch public cert secret.", err.Error())
+			return reconcile.Result{}, err
+		} else if publicCertSecretESCopy == nil {
 			r.status.SetDegraded("Waiting for elasticsearch public cert secret to become available", "")
 			return reconcile.Result{}, nil
 		}
 
-		esMetricsComponent := esmetrics.ElasticsearchMetrics(install, pullSecrets, clusterConfig, esMetricsSecret, esPubCertSecret, r.clusterDomain)
+		esMetricsComponent := esmetrics.ElasticsearchMetrics(install, pullSecrets, clusterConfig, esMetricsSecret, publicCertSecretESCopy, r.clusterDomain)
 		if err = imageset.ApplyImageSet(ctx, r.client, variant, esMetricsComponent); err != nil {
 			reqLogger.Error(err, "Error with images from ImageSet")
 			r.status.SetDegraded("Error with images from ImageSet", err.Error())
@@ -639,25 +703,92 @@ func (r *ReconcileLogStorage) deleteInvalidECKManagedPublicCertSecret(ctx contex
 	return r.client.Delete(ctx, secret)
 }
 
-// getElasticsearchCertificateSecrets retrieves Elasticsearch certificate secrets needed for Elasticsearch to run or for
-// components to communicate with Elasticsearch. The order of the secrets returned are:
-// 1) The certificate secret needed for Elasticsearch (in the operator namespace). If the user didn't create this it is
-//    created.
-// 2) The certificate secret needed for Elasticsearch (in the Elasticsearch namespace).
-// 3) The certificate mounted by other clients that connect to Elasticsearch.
-func (r *ReconcileLogStorage) getElasticsearchCertificateSecrets(ctx context.Context, instl *operatorv1.InstallationSpec) (oprKeyCert *corev1.Secret, esKeyCert *corev1.Secret, certSecret *corev1.Secret, err error) {
-	svcDNSNames := dns.GetServiceDNSNames(render.ElasticsearchServiceName, render.ElasticsearchNamespace, r.clusterDomain)
-
-	// Get the secret - might be nil
-	oprKeyCert, err = utils.GetSecret(ctx, r.client, render.TigeraElasticsearchCertSecret, rmeta.OperatorNamespace())
+// createKubeControllersSecrets checks for the existence of the secrets necessary for Kube controllers to access Elasticsearch through ES gateway and
+// creates them if they are missing. Kube controllers no longer uses admin credentials to make requests directly to Elasticsearch. Instead, gateway credentials
+// are generated and stored in the user secret, a hashed version of the credentials is stored in the tigera-elasticsearch namespace for ES Gateway to retrieve and use to compare
+// the gateway credentials, and a secret containing real admin level credentials is created and stored in the tigera-elasticsearch namespace to be swapped in once
+// ES Gateway has confirmed that the gateway credentials match.
+func (r *ReconcileLogStorage) createKubeControllersSecrets(ctx context.Context, esAdminUserSecret *corev1.Secret) (*corev1.Secret, *corev1.Secret, *corev1.Secret, error) {
+	kubeControllersGatewaySecret, err := utils.GetSecret(ctx, r.client, render.ElasticsearchKubeControllersUserSecret, rmeta.OperatorNamespace())
 	if err != nil {
 		return nil, nil, nil, err
+	}
+	if kubeControllersGatewaySecret == nil {
+		password := crypto.GeneratePassword(16)
+		kubeControllersGatewaySecret = &corev1.Secret{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      render.ElasticsearchKubeControllersUserSecret,
+				Namespace: rmeta.OperatorNamespace(),
+			},
+			Data: map[string][]byte{
+				"username": []byte(render.ElasticsearchKubeControllersUserName),
+				"password": []byte(password),
+			},
+		}
+	}
+	hashedPassword, err := bcrypt.GenerateFromPassword(kubeControllersGatewaySecret.Data["password"], bcrypt.MinCost)
+	if err != nil {
+		return nil, nil, nil, err
+	}
+
+	kubeControllersVerificationSecret, err := utils.GetSecret(ctx, r.client, render.ElasticsearchKubeControllersVerificationUserSecret, render.ElasticsearchNamespace)
+	if err != nil {
+		return nil, nil, nil, err
+	}
+	if kubeControllersVerificationSecret == nil {
+		kubeControllersVerificationSecret = &corev1.Secret{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      render.ElasticsearchKubeControllersVerificationUserSecret,
+				Namespace: render.ElasticsearchNamespace,
+			},
+			Data: map[string][]byte{
+				"username": []byte(render.ElasticsearchKubeControllersUserName),
+				"password": hashedPassword,
+			},
+		}
+	}
+
+	kubeControllersSecureUserSecret, err := utils.GetSecret(ctx, r.client, render.ElasticsearchKubeControllersSecureUserSecret, render.ElasticsearchNamespace)
+	if err != nil {
+		return nil, nil, nil, err
+	}
+	if kubeControllersSecureUserSecret == nil {
+		kubeControllersSecureUserSecret = &corev1.Secret{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      render.ElasticsearchKubeControllersSecureUserSecret,
+				Namespace: render.ElasticsearchNamespace,
+			},
+			Data: map[string][]byte{
+				"username": []byte("elastic"),
+				"password": esAdminUserSecret.Data["elastic"],
+			},
+		}
+	}
+
+	return kubeControllersGatewaySecret, kubeControllersVerificationSecret, kubeControllersSecureUserSecret, nil
+}
+
+// getESGatewayCertificateSecrets retrieves certificate secrets needed for ES Gateway to run or for
+// components to communicate with Elasticsearch/Kibana through ES Gateway. The order of the secrets returned are:
+// 1) The certificate/key secret to be mounted by ES Gateway and used to authenticate requests before
+// proxying to Elasticsearch/Kibana (in the operator namespace). If the user didn't create this secret, it is created.
+// 2) The certificate mounted by other clients that connect to Elasticsearch/Kibana through ES Gateway (in the operator namespace).
+func (r *ReconcileLogStorage) getESGatewayCertificateSecrets(ctx context.Context, instl *operatorv1.InstallationSpec) (*corev1.Secret, *corev1.Secret, error) {
+	var publicCertSecret *corev1.Secret
+
+	svcDNSNames := dns.GetServiceDNSNames(render.ElasticsearchServiceName, render.ElasticsearchNamespace, r.clusterDomain)
+	svcDNSNames = append(svcDNSNames, dns.GetServiceDNSNames(esgateway.ServiceName, render.ElasticsearchNamespace, r.clusterDomain)...)
+
+	// Get the secret - might be nil
+	oprKeyCert, err := utils.GetSecret(ctx, r.client, render.TigeraElasticsearchCertSecret, rmeta.OperatorNamespace())
+	if err != nil {
+		return nil, nil, err
 	}
 
 	// Ensure that cert is valid.
 	oprKeyCert, err = utils.EnsureCertificateSecret(render.TigeraElasticsearchCertSecret, oprKeyCert, corev1.TLSPrivateKeyKey, corev1.TLSCertKey, rmeta.DefaultCertificateDuration, svcDNSNames...)
 	if err != nil {
-		return nil, nil, nil, err
+		return nil, nil, err
 	}
 
 	// Three different certificate issuers are possible:
@@ -666,7 +797,85 @@ func (r *ReconcileLogStorage) getElasticsearchCertificateSecrets(ctx context.Con
 	// - The issuer that is provided through the certificate management feature.
 	keyCertIssuer, err := utils.GetCertificateIssuer(oprKeyCert.Data[corev1.TLSCertKey])
 	if err != nil {
-		return nil, nil, nil, err
+		return nil, nil, err
+	}
+	customerProvidedCert := !utils.IsOperatorIssued(keyCertIssuer)
+
+	// If Certificate management is enabled, we only want to trust the CA cert and let the init container handle private key generation.
+	if instl.CertificateManagement != nil {
+		cmCa := instl.CertificateManagement.CACert
+		cmIssuer, err := utils.GetCertificateIssuer(cmCa)
+		if err != nil {
+			return nil, nil, err
+		}
+
+		// If the issuer of the current secret is not the same as the certificate management issuer and also is not
+		// issued by the tigera-operator, it means that it is added to this cluster by the customer. This is not supported
+		// in combination with certificate management.
+		if customerProvidedCert && cmIssuer != keyCertIssuer {
+			return nil, nil, fmt.Errorf("certificate management does not support custom Elasticsearch secrets, please delete secret %s/%s or disable certificate management", oprKeyCert.Namespace, oprKeyCert.Name)
+		}
+
+		oprKeyCert.Data[corev1.TLSCertKey] = instl.CertificateManagement.CACert
+		publicCertSecret = render.CreateCertificateSecret(instl.CertificateManagement.CACert, relasticsearch.PublicCertSecret, rmeta.OperatorNamespace())
+	} else {
+		// Get the es gateway pub secret - might be nil
+		publicCertSecret, err = utils.GetSecret(ctx, r.client, relasticsearch.PublicCertSecret, rmeta.OperatorNamespace())
+		if err != nil {
+			return nil, nil, err
+		}
+
+		if publicCertSecret != nil {
+			// If the provided certificate secret (secret) is managed by the operator we need to check if the secret has the expected DNS names.
+			// If it doesn't, delete the public secret so it can get recreated.
+			if !customerProvidedCert {
+				err = utils.SecretHasExpectedDNSNames(publicCertSecret, corev1.TLSCertKey, svcDNSNames)
+				if err == utils.ErrInvalidCertDNSNames {
+					if err := r.deleteInvalidECKManagedPublicCertSecret(ctx, publicCertSecret); err != nil {
+						return nil, nil, err
+					}
+					publicCertSecret = render.CreateCertificateSecret(oprKeyCert.Data[corev1.TLSCertKey], relasticsearch.PublicCertSecret, rmeta.OperatorNamespace())
+				}
+			}
+		} else {
+			publicCertSecret = render.CreateCertificateSecret(oprKeyCert.Data[corev1.TLSCertKey], relasticsearch.PublicCertSecret, rmeta.OperatorNamespace())
+		}
+	}
+
+	return oprKeyCert, publicCertSecret, nil
+}
+
+// getElasticsearchCertificateSecrets retrieves Elasticsearch certificate secrets needed for Elasticsearch to run or for
+// ES gateway to communicate with Elasticsearch. The order of the secrets returned are:
+// 1) The certificate secret needed for Elasticsearch (in the Elasticsearch namespace). If the user didn't create this it is
+//    created.
+// 2) The certificate mounted by ES gateway to connect to Elasticsearch.
+func (r *ReconcileLogStorage) getElasticsearchCertificateSecrets(ctx context.Context, instl *operatorv1.InstallationSpec) (*corev1.Secret, *corev1.Secret, error) {
+	var esKeyCert, certSecret *corev1.Secret
+	svcDNSNames := dns.GetServiceDNSNames(render.ElasticsearchServiceName, render.ElasticsearchNamespace, r.clusterDomain)
+
+	// Get the secret - might be nil
+	esKeyCert, err := utils.GetSecret(ctx, r.client, render.TigeraElasticsearchInternalCertSecret, render.ElasticsearchNamespace)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// Ensure that cert is valid.
+	esKeyCert, err = utils.EnsureCertificateSecret(render.TigeraElasticsearchInternalCertSecret, esKeyCert, corev1.TLSPrivateKeyKey, corev1.TLSCertKey, rmeta.DefaultCertificateDuration, svcDNSNames...)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// Override the Operator namespace set by utils.EnsureCertificateSecret.
+	esKeyCert.Namespace = render.ElasticsearchNamespace
+
+	// Three different certificate issuers are possible:
+	// - The operator self-signed certificate
+	// - A user's BYO keypair for Elastic (uncommon)
+	// - The issuer that is provided through the certificate management feature.
+	keyCertIssuer, err := utils.GetCertificateIssuer(esKeyCert.Data[corev1.TLSCertKey])
+	if err != nil {
+		return nil, nil, err
 	}
 	customerProvidedCert := !utils.IsOperatorIssued(keyCertIssuer)
 
@@ -675,49 +884,46 @@ func (r *ReconcileLogStorage) getElasticsearchCertificateSecrets(ctx context.Con
 		cmCa := instl.CertificateManagement.CACert
 		cmIssuer, err := utils.GetCertificateIssuer(cmCa)
 		if err != nil {
-			return nil, nil, nil, err
+			return nil, nil, err
 		}
 
 		// If the issuer of the current secret is not the same as the certificate management issuer and also is not
 		// issued by the tigera-operator, it means that it is added to this cluster by the customer. This is not supported
 		// in combination with certificate management.
 		if customerProvidedCert && cmIssuer != keyCertIssuer {
-			return nil, nil, nil, fmt.Errorf("certificate management does not support custom Elasticsearch secrets, please delete secret %s/%s or disable certificate management", oprKeyCert.Namespace, oprKeyCert.Name)
+			return nil, nil, fmt.Errorf("certificate management does not support custom Elasticsearch secrets, please delete secret %s/%s or disable certificate management", esKeyCert.Namespace, esKeyCert.Name)
 		}
 
-		oprKeyCert.Data[corev1.TLSCertKey] = instl.CertificateManagement.CACert
-		esKeyCert = rsecret.CopyToNamespace(render.ElasticsearchNamespace, oprKeyCert)[0]
-		certSecret = render.CreateCertificateSecret(instl.CertificateManagement.CACert, relasticsearch.PublicCertSecret, rmeta.OperatorNamespace())
-
+		esKeyCert.Data[corev1.TLSCertKey] = instl.CertificateManagement.CACert
+		certSecret = render.CreateCertificateSecret(instl.CertificateManagement.CACert, relasticsearch.InternalCertSecret, render.ElasticsearchNamespace)
 	} else {
-		esKeyCert = rsecret.CopyToNamespace(render.ElasticsearchNamespace, oprKeyCert)[0]
-		// Get the pub secret - might be nil
-		pubSecret, err := utils.GetSecret(ctx, r.client, relasticsearch.PublicCertSecret, render.ElasticsearchNamespace)
+		// Get the internal public cert secret - might be nil
+		internalSecret, err := utils.GetSecret(ctx, r.client, relasticsearch.InternalCertSecret, render.ElasticsearchNamespace)
 		if err != nil {
-			return nil, nil, nil, err
+			return nil, nil, err
 		}
 
-		if pubSecret != nil {
+		if internalSecret != nil {
 			// If the provided certificate secret (secret) is managed by the operator we need to check if the secret that
 			// Elasticsearch creates from that given secret (pubSecret) has the expected DNS name. If it doesn't, delete the
 			// public secret so it can get recreated.
 			if !customerProvidedCert {
-				err = utils.SecretHasExpectedDNSNames(pubSecret, corev1.TLSCertKey, svcDNSNames)
+				err = utils.SecretHasExpectedDNSNames(internalSecret, corev1.TLSCertKey, svcDNSNames)
 				if err == utils.ErrInvalidCertDNSNames {
-					if err := r.deleteInvalidECKManagedPublicCertSecret(ctx, pubSecret); err != nil {
-						return nil, nil, nil, err
+					if err := r.deleteInvalidECKManagedPublicCertSecret(ctx, internalSecret); err != nil {
+						return nil, nil, err
 					}
 				}
 			}
-
-			certSecret = rsecret.CopyToNamespace(rmeta.OperatorNamespace(), pubSecret)[0]
+		} else {
+			certSecret = render.CreateCertificateSecret(esKeyCert.Data[corev1.TLSCertKey], relasticsearch.InternalCertSecret, render.ElasticsearchNamespace)
 		}
 	}
 
-	return oprKeyCert, esKeyCert, certSecret, err
+	return esKeyCert, certSecret, err
 }
 
-func (r *ReconcileLogStorage) kibanaSecrets(ctx context.Context, instl *operatorv1.InstallationSpec) ([]*corev1.Secret, error) {
+func (r *ReconcileLogStorage) kibanaInternalSecrets(ctx context.Context, instl *operatorv1.InstallationSpec) ([]*corev1.Secret, error) {
 
 	var secrets []*corev1.Secret
 	svcDNSNames := dns.GetServiceDNSNames(render.KibanaServiceName, render.KibanaNamespace, r.clusterDomain)
@@ -738,21 +944,21 @@ func (r *ReconcileLogStorage) kibanaSecrets(ctx context.Context, instl *operator
 		return []*corev1.Secret{
 			secret,
 			rsecret.CopyToNamespace(render.KibanaNamespace, secret)[0],
-			render.CreateCertificateSecret(instl.CertificateManagement.CACert, relasticsearch.PublicCertSecret, render.KibanaNamespace),
-			render.CreateCertificateSecret(instl.CertificateManagement.CACert, render.KibanaPublicCertSecret, rmeta.OperatorNamespace()),
+			render.CreateCertificateSecret(instl.CertificateManagement.CACert, relasticsearch.InternalCertSecret, render.KibanaNamespace),
+			render.CreateCertificateSecret(instl.CertificateManagement.CACert, render.KibanaInternalCertSecret, rmeta.OperatorNamespace()),
 		}, nil
 	}
 
 	secrets = append(secrets, secret, rsecret.CopyToNamespace(render.KibanaNamespace, secret)[0])
 
 	// Get the pub secret - might be nil
-	pubSecret, err := utils.GetSecret(ctx, r.client, render.KibanaPublicCertSecret, render.KibanaNamespace)
+	internalSecret, err := utils.GetSecret(ctx, r.client, render.KibanaInternalCertSecret, render.KibanaNamespace)
 	if err != nil {
 		return nil, err
 	}
 
-	if pubSecret == nil {
-		log.Info(fmt.Sprintf("Public cert secret %q not found yet", render.KibanaPublicCertSecret))
+	if internalSecret == nil {
+		log.Info(fmt.Sprintf("Internal cert secret %q not found yet", render.KibanaInternalCertSecret))
 		return secrets, nil
 	}
 
@@ -762,15 +968,15 @@ func (r *ReconcileLogStorage) kibanaSecrets(ctx context.Context, instl *operator
 	}
 
 	if utils.IsOperatorIssued(issuer) {
-		err = utils.SecretHasExpectedDNSNames(pubSecret, corev1.TLSCertKey, svcDNSNames)
+		err = utils.SecretHasExpectedDNSNames(internalSecret, corev1.TLSCertKey, svcDNSNames)
 		if err == utils.ErrInvalidCertDNSNames {
-			if err := r.deleteInvalidECKManagedPublicCertSecret(ctx, pubSecret); err != nil {
+			if err := r.deleteInvalidECKManagedPublicCertSecret(ctx, internalSecret); err != nil {
 				return nil, err
 			}
 		}
 	}
 	// If the cert was not deleted, copy the valid cert to operator namespace.
-	secrets = append(secrets, rsecret.CopyToNamespace(rmeta.OperatorNamespace(), pubSecret)...)
+	secrets = append(secrets, rsecret.CopyToNamespace(rmeta.OperatorNamespace(), internalSecret)...)
 
 	return secrets, nil
 }
--- a/pkg/render/common/elasticsearch/service.go
+++ b/pkg/render/common/elasticsearch/service.go
@@ -7,8 +7,8 @@ import (
 )
 
 const (
-	httpsEndpoint     = "https://tigera-secure-es-http.tigera-elasticsearch.svc:9200"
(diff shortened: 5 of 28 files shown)


pkg/render/logstorage.go
```diff
@@ -1705,6 +1751,168 @@ func (es elasticsearchComponent) oidcUserRoleBinding() []client.Object {
...
+			ServiceAccountName: EsGatewayServiceAccountName,
+			ImagePullSecrets:   secret.GetReferenceList(es.pullSecrets),
+			InitContainers:     initContainers,
+			Volumes:            volumes,
+			Containers: []corev1.Container{
+				{
+					Name:          EsGatewayName,
+					Image:         es.esGatewayImage,
+					Env:           envVars,
+					LivenessProbe: complianceLivenessProbe,
```
stevegaossou, line 1829: This looks like a typo. 

We will have a readiness probe but we probably won't define a liveness probe.

pkg/render/logstorage.go
```diff
@@ -1705,6 +1773,184 @@ func (es elasticsearchComponent) oidcUserRoleBinding() []client.Object {
 	}
 }
 
+func (es elasticsearchComponent) esGatewayDeployment() *appsv1.Deployment {
```
Brian-McM, line 1776: I don't think this should be added to the this component, I think it should be it's own component in the logstorage package. You can render multiple components per controller, this component is bloated and we should add anything to it that is elasticsearch specific (i.e. that actually configures Elasticsearch).

Brian-McM (changes requested):

api/v1/logstorage_types.go
```diff
@@ -52,6 +52,10 @@ type LogStorageSpec struct {
 	// Only ECKOperator is supported for this spec.
 	// +optional
 	ComponentResources []LogStorageComponentResource `json:"componentResources,omitempty"`
+
+	// EsGatewayReplicaCount can be used to adjust the replica count for ES Gateway.
+	// +optional
+	EsGatewayReplicaCount *int32 `json:"esGatewayReplicaCount,omitempty"`
```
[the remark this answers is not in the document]
tmjd: I don't know, `IngestReplicaCount`?

pkg/controller/intrusiondetection/intrusiondetection_controller.go
```diff
@@ -144,6 +145,14 @@ func add(mgr manager.Manager, c controller.Controller) error {
 		return fmt.Errorf("intrusiondetection-controller failed to watch the Secret resource: %v", err)
 	}
 
+	if err = utils.AddSecretsWatch(c, esgateway.EsGatewayElasticPublicCertSecret, render.IntrusionDetectionNamespace); err != nil {
```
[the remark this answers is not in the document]
vberezny: > I think the only reason is to add a watch for these in the IntrusionDetectionNamespace is to make sure if the secrets are modified they would be "corrected" by the triggered Reconcile

That was my thinking here, if we don't actually need them I can remove the watches.

api/v1/logstorage_types.go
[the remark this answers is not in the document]
vberezny: I'll change it to `IngestReplicaCount`

pkg/render/logstorage/esgateway/esgateway.go
```diff
@@ -0,0 +1,427 @@
...
+	EsGatewayElasticServiceName      = "tigera-secure-es-http"
+	EsGatewayKibanaServiceName       = "tigera-secure-kb-http"
+	EsGatewayKibanaPublicCertSecret  = "tigera-secure-kb-http-certs-public"
+	EsGatewayElasticPublicCertSecret = "tigera-secure-es-http-certs-public"
+	EsGatewayElasticUserSecret       = "tigera-es-gateway-elasticsearch-access"
+	EsGatewayTLSSecret               = "tigera-es-gateway-tls"
+	EsGatewayPortName                = "es-gateway-port"
+	EsGatewayPort                    = 5554
+
+	ElasticsearchHTTPSEndpoint = "https://tigera-secure-internal-es-http.tigera-elasticsearch.svc:9200"
```
tmjd, line 39: I'm concerned with this Endpoint change. Because the documentation calls out that if providing the `tigera-secure-elasticsearch-cert` that it needs to contain `tigera-secure-es-http.tigera-elasticsearch.svc`, is this PR breaking that?

vberezny: I wasn't aware of this but after taking a look yes, this PR would break that. I can put up a docs PR to change the requirement from `tigera-secure-es-http.tigera-elasticsearch.svc` to `tigera-secure-internal-es-http.tigera-elasticsearch.svc`. Would that be an acceptable solution? Is there a concern that when upgrading a customer would be forced to generate/provide a new cert?

pkg/render/common/esgateway/esgateway.go
```diff
@@ -0,0 +1,9 @@
+package esgateway
+
+const (
+	EsGatewayElasticServiceName      = "tigera-secure-es-http"
```
Brian-McM, line 4: Nit, Could you please remove the `EsGateway` prefix here. The package name is `esgateway` so `esgateway.EsGateway` is redundant.

Brian-McM: Nit, please don't shorten `Elasticsearch` to `Elastic`, `Elastic` has a meanings outside of `Elasticsearch` in computing. I'd just say Elasticsearch.

pkg/render/logstorage/esgateway/esgateway.go
stevegaossou: Just want to make sure we're on the same page here. 

We're talking about this user doc, right @tmjd: 
https://docs.tigera.io/master/security/comms/log-storage-tls

Basically we give the user the option to provide their own cert/key, instead of our self-signed version, which should be used with the `tigera-secure-es-http.tigera-elasticsearch.svc` domain. 

With the addition of ES Gateway, it's the gateway that has adopted the `tigera-secure-es-http.tigera-elasticsearch.svc` domain. So what should happen here is that the user provided cert/key should be used by ES Gateway, not the Elasticsearch pod. 

The Elasticsearch pod should continue to use the new domain `tigera-secure-internal-es-http.tigera-elasticsearch.svc`. 

Does that make sense to everyone?

pkg/render/common/elasticsearch/decorator.go
```diff
@@ -139,7 +140,7 @@ func DefaultVolume() corev1.Volume {
 		Name: "elastic-ca-cert-volume",
 		VolumeSource: v1.VolumeSource{
 			Secret: &v1.SecretVolumeSource{
-				SecretName: PublicCertSecret,
+				SecretName: esgateway.EsGatewayElasticPublicCertSecret,
```
Brian-McM, line 143: Using the esgateway package directly here makes me a little wary for two reasons:
1. I don't know that this decorator needs to know about a gateway
2. Importing this means you can't import in the other direction (i.e. import elasticsearch items into the gateway package) because you'd have a cyclic dependency

I'd personally strive for low coupling and use dependency injection to pass the name through, or just redefine the default value here.

pkg/render/logstorage/esgateway/esgateway.go
```diff
@@ -0,0 +1,427 @@
...
+)
+
+const (
+	EsGatewayName                    = "tigera-es-gateway"
+	EsGatewayServiceAccountName      = "tigera-es-gateway"
+	EsGatewayRole                    = "tigera-es-gateway"
+	EsGatewaySecretKeyName           = "key"
+	EsGatewaySecretCertName          = "tls.crt"
+	EsGatewayVolumeName              = "tigera-es-gateway-certs"
+	EsGatewayElasticServiceName      = "tigera-secure-es-http"
```
Brian-McM, line 30: Moving these constants here make it so that every component now needs to know of the existence of the gateway, when they really don't. Logically the gateway is an invisible component, and the gateway url and certs should really be exposed to them as Elasticsearch items. 

The names are the same, the _only_ item that should know that there is a gateway is the logstorage controller, and the Gateway and Elasticsearch render items should be configurable enough the the logstorage controller can configure them accordingly.

I would keep these parameters in the package that they were in, and pass the values into the gateway and elasticsearch components. That way we can still keep all the other components blind to the fact that there is a gateway, to them there's only Elasticsearch.

pkg/render/logstorage/esgateway/esgateway.go
```diff
@@ -0,0 +1,427 @@
...
+			Data: map[string][]byte{
+				"username": []byte("elastic"),
+				"password": esAdminUserSecret.Data["elastic"],
+			},
+		}
+		esGatewayTLSSecrets = append(esGatewayTLSSecrets, esGatewayElasticUserSecret)
+	}
+
+	// tigera-es-gateway-tls
+	esGatewayTLSSecrets = append(esGatewayTLSSecrets, esGatewayTLSSecret)
```
Brian-McM, line 76: We need to make sure that we don't write back resources that can be create outside of the render function. This TLS secret may be created by a customer, or something other than the operator, so we shouldn't write back to it (it may change the owner reference).

You can still copy these secrets to other namespace though.

Brian-McM (changes requested):

pkg/render/logstorage/esgateway/esgateway.go
vberezny: > With the addition of ES Gateway, it's the gateway that has adopted the tigera-secure-es-http.tigera-elasticsearch.svc domain. So what should happen here is that the user provided cert/key should be used by ES Gateway, not the Elasticsearch pod.

So this means we should allow the user to create/update the `tigera-es-gateway-tls` secret instead of the `tigera-secure-elasticsearch-cert` correct? That would involve making some changes in this PR as well as opening a docs PR to update https://docs.tigera.io/master/security/comms/log-storage-tls.

Brian-McM: @stevegaossou I think that makes sense, Elasticsearch now becomes strictly internal, so we could just use our own internally generated secrets.

This reminds though that we need to bring @rene-dekker into this so that we don't break certificate management. @rene-dekker could you please review the relevant bits here when you have time?

stevegaossou: I think what this has made me realize is that `tigera-es-gateway-tls` == `tigera-secure-elasticsearch-cert`. So we don't actually need both of them.

Is there any other use of `tigera-secure-elasticsearch-cert` (referred to as [`TigeraElasticsearchCertSecret` in the Operator code](https://github.com/tigera/operator/search?q=TigeraElasticsearchCertSecret)) other than this? It doesn't look like it. 

Your options are: 

1. Use the existing `tigera-secure-elasticsearch-cert` for ES gateway
- Pro: 
  - This preserves what is already documented to the user and is less confusing.
- Con: 
  - Maybe it's a little less intuitive that ES Gateway's secret is named `tigera-secure-elasticsearch-cert` (although IMO ES Gateway is tied to ES so it's not that bad of a name).

2. Use the new `tigera-es-gateway-tls` for ES gateway and refactor the old refers to `tigera-secure-elasticsearch-cert`
- Pro: 
  - Basically the opposite of the Con for Option 1 (but again I don't think this is a very strong benefit)
- Con: 
  - You have to refactor all the references to `tigera-secure-elasticsearch-cert` in code and user docs
  - User docs change from what they mentioned before (so there's a break in consistency)

My vote is for keeping `tigera-secure-elasticsearch-cert` and getting rid of `tigera-es-gateway-tls`.

Brian-McM: > My vote is for keeping tigera-secure-elasticsearch-cert and getting rid of tigera-es-gateway-tls.

Mine too, nothing really needs to know about the gateway besides the controller, I don't think we should expose any configuration and mention it as little as possible in places that just care about Elasticsearch (and not an invisible intermediary).

api/v1/logstorage_types.go
```diff
@@ -52,6 +52,10 @@ type LogStorageSpec struct {
 	// Only ECKOperator is supported for this spec.
 	// +optional
 	ComponentResources []LogStorageComponentResource `json:"componentResources,omitempty"`
+
+	// IngestReplicaCount can be used to adjust the replica count for ES Gateway.
+	// +optional
+	IngestReplicaCount *int32 `json:"ingestReplicaCount,omitempty"`
```
rene-dekker, line 58: I feel a bit conflicted about this field. If it is used, are we then supposed to not have Nodes, Indices, etc, since ES is outsourced to an external cluster? I would expect to see validation to ensure this is the case.

I also feel that the name IngestReplicaCount can easily lead to confusions, since ES also has a concept of ingestion nodes. I would be in favor of adding an es-gateway struct, which then has more clarification around it's setup (replica's in this case), to clearly divide regular logstorage from the es-gateway scenario.

pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -454,6 +477,40 @@ func (r *ReconcileLogStorage) Reconcile(ctx context.Context, request reconcile.R
 			return reconcile.Result{}, err
 		}
 
+		// Check that if the ES Gateway certpair secret exists that it is valid (has key and cert fields)
+		// If it does not exist then this function returns a nil secret but no error and a self-signed
+		// certificate will be generated when rendering below.
+		tlsSecret, err = utils.ValidateCertPair(r.client,
```
rene-dekker, line 483: I'd pick a more specific secret name to indicate it is for es gateway.

pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -454,6 +477,40 @@ func (r *ReconcileLogStorage) Reconcile(ctx context.Context, request reconcile.R
...
+		)
+
+		// An error is returned in case the read cannot be performed of the secret does not match the expected format
+		// In case the secret is not found, the error and the secret will be nil.
+		if err != nil {
+			r.status.SetDegraded("Error validating ES Gateway TLS certificate", err.Error())
+			return reconcile.Result{}, err
+		}
+
+		if tlsSecret == nil {
```
rene-dekker, line 497: This is not certificate management compatible. You may want to throw an error if es-gateway and certificate management are used at the same time.

rene-dekker: Does this code mean that every cluster will have an es-gateway unless it is a managed cluster?

api/v1/logstorage_types.go
Brian-McM: @rene-dekker, @tmjd had request Vlad change this. @rene-dekker is right that ingest is an overloaded term here, and I would like to reserve it for the specific elasticsearch node type, should we decide to optimise the Elasticsearch clusters and allow users to configure it.

@tmjd what was the reasoning you don't want to expose that there's a gateway in these settings? I'd agree that if there was a way to deduce the number of replicas for the gateway that we shouldn't expose it, but that's not possible, so I would argue that it should be clear that the setting changes the gateway replica count (Gateway section with ReplicaCount). The resource will be called a gateway in the cluster, no reason to hide a user facing setting for it.

pkg/render/logstorage/esgateway/esgateway.go
vberezny: > My vote is for keeping tigera-secure-elasticsearch-cert and getting rid of tigera-es-gateway-tls.

I agree with this approach. I'll proceed with the refactor.

pkg/controller/logstorage/logstorage_controller.go
vberezny: > Does this code mean that every cluster will have an es-gateway unless it is a managed cluster?

Yes exactly.

> This is not certificate management compatible. You may want to throw an error if es-gateway and certificate management are used at the same time.

I plan to revisit this now that we've decided to re-purpose `tigera-secure-elasticsearch-cert` to be the gateway secret and I'll make sure that this is cert management compatible.

pkg/render/logstorage/esgateway/esgateway.go
stevegaossou: @rene-dekker you had some comments about the direction we're heading with respect to BYO certs, but it wasn't visible to everyone (namely Erik).

> We currently have no customer at all that uses BYO certs. After much of the certificate management discussions, we were moving in the direction of removing this feature over time altogether, not letting the code BYO them. If they want better certs, we wanna expose the SANs through an API and use certificate management. Therefore I would propose not to allow them at all and use certificate management.

Since we're just going to reuse the `tigera-secure-elasticsearch-cert` cert and keep it consistent, this change shouldn't really push us any further down the BYO cert path. 

Rene, do you have any qualms with Vlad make this PR adopt the cert/key found in `tigera-secure-elasticsearch-cert` for ES Gateway, if a user provides that Secret?

pkg/components/enterprise.go
```diff
@@ -186,5 +191,6 @@ var (
 		ComponentTigeraCNI,
 		ComponentCloudControllers,
 		ComponentElasticsearchMetrics,
+		ComponentEsGateway,
```
stevegaossou, line 194: Minor: I think for acronyms the Go convention is to use all caps. So `ComponentESGateway`.

pkg/render/logstorage/esgateway/esgateway.go
tmjd: Sorry I didn't chime in yesterday, but I think where this has landed is what I think we should do. The gateway changes should require no user facing API changes (even though no one is using it, BYO cert configuration should not change at this time).

As far as dropping the BYO cert capability, that sounds good, but that needs to be explicitly planned work.

api/v1/logstorage_types.go
tmjd: I don't want to expose this to the user to because I believe it is something that user shouldn't specifically be aware of. For example, if we moved this functionality to voltron then a Gateway setting would be a misnomer then.

Have we considered deducing the number of replicas needed for the gateway? You say it isn't possible but I think there are several ways we could address that. What if we exposed an endpoint on the gateway that the operator queried and scaled the gateway up when a certain threshold was reached.
Another option would be to use node counts to determine its scale, though we'd need to add some functionality to get node counts for managed clusters.
Another option is to scale with ES nodes, we know ES will have to be able to handle everything going through the gateway so if we know 1 gateway can saturate one ES node then that's an easy scaling.

Also we don't know how much this component will need to scale, I think we should deploy at least 2 for redundancy/HA so those 2 may be enough to scale.

Originally I believe there was no plan to expose an option for gateway scaling until @Thakurvaibhav requested it, so we had it for internal use, if the expectation is still that this would be for internal use then since we don't have a good design for this configuration I'd prefer not to expose it generally.

vberezny: I've extracted this field to a separate branch for now and we can continue to assess how we want to go about scaling the gateway after performance testing it.

Brian-McM: > Also we don't know how much this component will need to scale, I think we should deploy at least 2 for redundancy/HA so those 2 may be enough to scale.

Ya let's do this, that's a good plan.

Regarding everything else you said, I think you're right. I'm forgetting that this really is just a simple proxy, if it couldn't handle 100 times the load of Elasticsearch then the implementation of the gateway is extremely flawed, so 2 replicas for 3 nodes (which is the max anybody has) is _**very**_ generous.

> I've extracted this field to a separate branch for now and we can continue to assess how we want to go about scaling the gateway after performance testing it.

@vberezny if performance testing shows that the gateway is the bottleneck here then I would be very hesitant to say the way forward would be to add more gateways, I'd say the implementation should be scrutinised and figure out why it's the bottleneck and not Elasticsearch.

pkg/render/logstorage/esgateway/esgateway.go
```diff
@@ -0,0 +1,352 @@
...
+			ValueFrom: secret.GetEnvVarSource(ElasticsearchUserSecret, "password", false),
+		},
+	}
+
+	volumes := []corev1.Volume{
+		{
+			Name: VolumeName,
+			VolumeSource: corev1.VolumeSource{
+				Secret: &corev1.SecretVolumeSource{
+					SecretName: render.TigeraElasticsearchCertSecret,
```
rene-dekker, line 195: If I understand well, this is the secret containing the key and cert for encrypting traffic. I believe you should still add an annotation for this secret to the spec.

pkg/render/logstorage/esgateway/esgateway.go
```diff
@@ -0,0 +1,352 @@
...
+
+	envVars := []corev1.EnvVar{
+		{Name: "ES_GATEWAY_LOG_LEVEL", Value: "INFO"},
+		{Name: "ES_GATEWAY_ELASTIC_ENDPOINT", Value: ElasticsearchHTTPSEndpoint},
+		{Name: "ES_GATEWAY_KIBANA_ENDPOINT", Value: KibanaHTTPSEndpoint},
+		{Name: "ES_GATEWAY_HTTPS_CERT", Value: "/certs/https/tls.crt"},
+		{Name: "ES_GATEWAY_HTTPS_KEY", Value: "/certs/https/tls.key"},
+		{
+			Name:      "ES_GATEWAY_ELASTIC_USERNAME",
+			ValueFrom: secret.GetEnvVarSource(ElasticsearchUserSecret, "username", false),
```
rene-dekker, line 182: I believe you still need an annotation for this spec, so the pod is replaced if the value expires.

pkg/render/logstorage/esgateway/esgateway.go
```diff
@@ -0,0 +1,352 @@
...
+		},
+		{
+			Name:      "ES_GATEWAY_ELASTIC_PASSWORD",
```
rene-dekker, line 185: I believe you still need an annotation for this spec, so the pod is replaced if the value expires.

pkg/render/logstorage/esgateway/esgateway.go
```diff
@@ -0,0 +1,352 @@
...
+			ValueFrom: secret.GetEnvVarSource(ElasticsearchUserSecret, "password", false),
```
rene-dekker, line 186: I believe you still need an annotation for this spec, so the pod is replaced if the value expires.

pkg/render/logstorage.go
```diff
@@ -59,22 +59,25 @@ const (
 
 	ElasticsearchNamespace = "tigera-elasticsearch"
 
-	TigeraElasticsearchCertSecret = "tigera-secure-elasticsearch-cert"
+	TigeraElasticsearchCertSecret         = "tigera-secure-elasticsearch-cert"
+	TigeraElasticsearchInternalCertSecret = "tigera-secure-internal-elasticsearch-cert"
```
rene-dekker, line 63: This line can be removed from the code. It is never used.

pkg/controller/installation/core_controller.go
```diff
@@ -906,10 +906,17 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
...
-		esAdminSecret, err = utils.GetSecret(ctx, r.client, render.ElasticsearchAdminUserSecret, rmeta.OperatorNamespace())
+		// Kube controllers no longer uses the admin secret to run it's administrative tasks on Elasticsearch.
+		// Instead a username/password combo is randomly generated and stored in tigera-ee-kube-controllers-elasticsearch-access
+		// and passed to the ES Gateway deployment to be used to detect requests from kube-controllers and swap in admin
+		// level Elasticsearch credentials.
+		kubeControllersUserSecret, err = utils.GetSecret(ctx, r.client, render.ElasticsearchKubeControllersUserSecret, common.CalicoNamespace)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+		esAdminSecret, err = utils.GetSecret(ctx, r.client, render.ElasticsearchAdminUserSecret, render.ElasticsearchNamespace)
```
tmjd, line 919: Reading the comment on kubeControllersUserSecret, do we still need this?

pkg/controller/installation/core_controller.go
```diff
@@ -241,12 +241,12 @@ func add(mgr manager.Manager, r *ReconcileInstallation) error {
 			return fmt.Errorf("tigera-installation-controller failed to watch Secret %s: %w", render.ElasticsearchAdminUserSecret, err)
 		}
 
-		if err = utils.AddSecretsWatch(c, relasticsearch.PublicCertSecret, rmeta.OperatorNamespace()); err != nil {
-			return fmt.Errorf("tigera-installation-controller failed to watch Secret '%s' in '%s' namespace: %w", relasticsearch.PublicCertSecret, rmeta.OperatorNamespace(), err)
+		if err = utils.AddSecretsWatch(c, relasticsearch.InternalCertSecret, rmeta.OperatorNamespace()); err != nil {
```
tmjd, line 244: You've changed these watches but in the Reconcile you're still getting the Public version of this and the Kibana CertSecrets.

pkg/controller/intrusiondetection/intrusiondetection_controller.go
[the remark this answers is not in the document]
tmjd: Can you add a comment on these watches that they are only here to catch a modification to the resources we create in reconcile so they changes would be corrected? Thanks

pkg/render/logstorage/esgateway/esgateway.go
```diff
@@ -0,0 +1,352 @@
...
+	DeploymentName          = "tigera-es-gateway"
+	ServiceAccountName      = "tigera-es-gateway"
+	RoleName                = "tigera-es-gateway"
+	VolumeName              = "tigera-es-gateway-certs"
+	ElasticsearchUserSecret = "tigera-es-gateway-elasticsearch-access"
+	TLSSecret               = "tigera-es-gateway-tls"
+	PortName                = "es-gateway-port"
+	Port                    = 5554
+
+	ElasticsearchHTTPSEndpoint = "https://tigera-secure-internal-es-http.tigera-elasticsearch.svc:9200"
```
rene-dekker, line 32: Adding a new thread for the same line. Although the thread appears to be resolved, nowhere in this PR do I see the creation of a TLS certificate with this SAN. Am I missing anything or will there be TLS issues the way this PR is right now?

rene-dekker: Never mind, ECK will create those by taking "tigera-secure-internal" and add the suffix "-es-http", right?

vberezny: Yes exactly, everything from service names to certificates are generated by ECK Operator like you said. I've tested this in a fresh cluster and everything works as expected.

pkg/controller/installation/core_controller.go
vberezny: Yes we do, I forgot to update the comment since we wen't back and forth on this a few times. Initially we planned to eliminate the admin user secret for kube controllers entirely and have a special case in the gateway that would swap in the admin credentials that the gateway already has access to on all kube-controllers Elasticsearch requests. But later we decided to eliminate the special case and have the same dummy credential -> real credential swap for kube-controllers as we do for other components. So we still need to create an admin user secret for kube-controllers, the only difference being that we create it in the tigera-elasticsearch namespace now instead of calico-system. 

I will update the comment to reflect this.

pkg/controller/installation/core_controller.go
vberezny: Ah right that was likely a find + replace mistake, I'll revert.

pkg/controller/installation/core_controller.go
vberezny: Actually I will remove this comment and write up something similar in kube-controllers.go

pkg/render/logstorage.go
```diff
@@ -660,10 +661,10 @@ func (es elasticsearchComponent) podTemplate() corev1.PodTemplateSpec {
 			es.installation.CertificateManagement,
 			es.csrImage,
 			csrVolumeNameHTTP,
-			ElasticsearchServiceName,
+			ElasticsearchInternalServiceName,
```
rene-dekker, line 664: This change will have some consequences for our users that upgrade elasticsearch. What will happen is that alongside the current Elasticsearch CR, a new CR will be created with this new name. That means that the user will have two clusters alongside, one with the old data and one that is empty. I believe this cluster will stay empty, no current data will be ported over. Also, the user needs to have enough disks for both clusters. We need to come up with a solution if we want to keep this change, which might involve adding extra upgrade steps to the docs.

pkg/render/logstorage.go
```diff
@@ -61,20 +61,22 @@ const (
 
 	TigeraElasticsearchCertSecret = "tigera-secure-elasticsearch-cert"
 
-	ElasticsearchName                     = "tigera-secure"
+	ElasticsearchName                     = "tigera-secure-internal"
```
rene-dekker, line 64: ElasticsearchName

Brian-McM: After chatting with @rene-dekker it seems like the service name uses this name, so it makes this sort of awkward. @stevegaossou I don't think changing this name is an option, it's a very risky operation that will like result in this cluster being deleted. I think this has to go back to the drawing board, we can't change the service name because of this, so we need to figure out if that's acceptable or not to roll out to customers outside of calico cloud

Brian-McM (changes requested):

pkg/render/logstorage.go
Brian-McM: @rene-dekker had a good idea about modifying the existing service in a certain way (I'll let him explain it), I just wanted to chime in to say whatever you do can you make sure the ECK operator won't trounce over it, please review what you come up with against the service controller here https://github.com/elastic/cloud-on-k8s/blob/1.3/pkg/controller/elasticsearch/services/services.go

stevegaossou: @rene-dekker @Brian-McM thanks for the feedback. 

@vberezny is going to look into seeing if there is a supported way to override the service name that is generated by ECK operator (which is `tigera-secure-es-http`). 

We unfortunately need to change that name, so that es-gateway can use `tigera-secure-es-http`. This is important for Calico Cloud specifically, because we cannot require customers upgrade their managed clusters. So components need to continue using `tigera-secure-es-http`, and that service needs to be backed by es-gateway instead. 

I don't think we want to go down a path where some components (from some clusters) use `tigera-secure-es-http` and some use a new service (e.g. `tigera-secure-es-gateway-http`). 

> I think this has to go back to the drawing board, we can't change the service name because of this, so we need to figure out if that's acceptable or not to roll out to customers outside of calico cloud

I think we want to have es-gateway in place for Calico Enterprise because it has security benefits for MCM setups in general (not just Calico Cloud). But I'll leave this open and check in with Matt D and Garwood.

rene-dekker: Brian and I were thinking if we configure Elasticsearch (CR) with HTTP.Service.Spec to point to label k8s-app=es-gateway and create a new service with k8s-app=elasticsearch (or something like that), we can keep the existing name for the CR and hopefully everything works. Now there are two caveats:
- It only works if ECK respects that service configuration
- When requests are sent to a service with label k8s-app=elasticsearch during an ECK change between two node sets due to a spec change or whatever, I cannot predict what is going to happen.

stevegaossou: @vberezny I think you can override it like this: 

```
	elasticsearch := &esv1.Elasticsearch{
		TypeMeta: metav1.TypeMeta{Kind: "Elasticsearch", APIVersion: "elasticsearch.k8s.elastic.co/v1"},
			Name:      ElasticsearchName,
			Namespace: ElasticsearchNamespace,
			Annotations: map[string]string{
				"common.k8s.elastic.co/controller-version": components.ComponentElasticsearchOperator.Version,
			},
		},
		Spec: esv1.ElasticsearchSpec{
			Version: components.ComponentEckElasticsearch.Version,
			Image:   es.esImage,
			HTTP: cmnv1.HTTPConfig{
				TLS: cmnv1.TLSOptions{
					Certificate: cmnv1.SecretRef{
						SecretName: TigeraElasticsearchCertSecret,
					},
				},
                                 // -------------------------- ADD THE BELOW--------------------------
				Service: cmnv1.ServiceTemplate{
						Name: "tigera-secure-internal-es-http",
					},
				},
				// ----------------------------------------------------------------------
			},
			NodeSets: es.nodeSets(),
		},
	}

```

pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -641,23 +678,31 @@ func (r *ReconcileLogStorage) deleteInvalidECKManagedPublicCertSecret(ctx contex
...
 // getElasticsearchCertificateSecrets retrieves Elasticsearch certificate secrets needed for Elasticsearch to run or for
 // components to communicate with Elasticsearch. The order of the secrets returned are:
-// 1) The certificate secret needed for Elasticsearch (in the operator namespace). If the user didn't create this it is
+// 1) The certificate secret needed for ES Gateway (in the operator namespace). If the user didn't create this it is
 //    created.
-// 2) The certificate secret needed for Elasticsearch (in the Elasticsearch namespace).
-// 3) The certificate mounted by other clients that connect to Elasticsearch.
-func (r *ReconcileLogStorage) getElasticsearchCertificateSecrets(ctx context.Context, instl *operatorv1.InstallationSpec) (oprKeyCert *corev1.Secret, esKeyCert *corev1.Secret, certSecret *corev1.Secret, err error) {
+// 2) The certificate secret needed for ES Gateway (in the Elasticsearch namespace).
+// 3) The certificate mounted by other clients that connect to Elasticsearch (through ES Gateway).
```
tmjd, line 684: Shouldn't this be the same as 2? Because other clients should be using the Public cert, which is what a user would provide/create.

pkg/render/logstorage.go
rene-dekker: ObjectMeta cannot be customized according to the ECK docs. Only the spec.

pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -738,21 +809,21 @@ func (r *ReconcileLogStorage) kibanaSecrets(ctx context.Context, instl *operator
...
+			render.CreateCertificateSecret(instl.CertificateManagement.CACert, relasticsearch.InternalCertSecret, render.KibanaNamespace),
+			render.CreateCertificateSecret(instl.CertificateManagement.CACert, render.KibanaInternalCertSecret, rmeta.OperatorNamespace()),
 		}, nil
 	}
 
 	secrets = append(secrets, secret, rsecret.CopyToNamespace(render.KibanaNamespace, secret)[0])
 
 	// Get the pub secret - might be nil
-	pubSecret, err := utils.GetSecret(ctx, r.client, render.KibanaPublicCertSecret, render.KibanaNamespace)
+	pubSecret, err := utils.GetSecret(ctx, r.client, render.KibanaInternalCertSecret, render.KibanaNamespace)
```
tmjd, line 820: assignment and secret name do not match.

pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -666,61 +711,87 @@ func (r *ReconcileLogStorage) getElasticsearchCertificateSecrets(ctx context.Con
...
 
-	return oprKeyCert, esKeyCert, certSecret, err
+	return oprKeyCert, esKeyCert, elasticsearchCertSecret, kibanaCertSecret, err
 }
 
 func (r *ReconcileLogStorage) kibanaSecrets(ctx context.Context, instl *operatorv1.InstallationSpec) ([]*corev1.Secret, error) {
 
 	var secrets []*corev1.Secret
-	svcDNSNames := dns.GetServiceDNSNames(render.KibanaServiceName, render.KibanaNamespace, r.clusterDomain)
+	svcDNSNames := dns.GetServiceDNSNames(render.KibanaInternalServiceName, render.KibanaNamespace, r.clusterDomain)
```
tmjd, line 794: I think you should adjust the name of this function if it is suppose to be handling "Internal" secret configuration.

pkg/render/logstorage.go
stevegaossou: That's sad. 😞 

Vlad mentioned the next option you guys are trying (of using ECK operator to setup the service to be backed by ES gateway pods) and then create the other HTTP service outside of ECK operator and have that be backed by ES pods. 

A problem might be how Kibana pods are configured to talk to ES pods (which ECK operator configures.

tmjd (commented):
Some more comments/questions.

pkg/controller/installation/core_controller.go
vberezny: we are only giving kube controllers the dummy set, the admin set is being dropped into the tigera-elasticsearch namespace and only accessed by es-gateway during the swapping.

pkg/render/logstorage.go
Brian-McM: Kibana shouldn't have any problem @stevegaossou, Kibana just needs the url and the certificate for Elastcisearch, and those won't change.

pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -167,12 +168,21 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 	// Watch all the secrets created by this controller so we can regenerate any that are deleted
 	for _, secretName := range []string{
 		render.TigeraElasticsearchCertSecret, render.TigeraKibanaCertSecret,
-		render.OIDCSecretName, render.DexObjectName} {
+		render.OIDCSecretName, render.DexObjectName, relasticsearch.PublicCertSecret,
+		render.KibanaPublicCertSecret} {
```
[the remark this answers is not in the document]
vberezny: TigeraElasticsearchCertSecret and TigeraKibanaCertSecret hold the cert/key pair while their public counterparts hold just the cert.

> I think since TigeraElasticsearchCertSecret (and TigeraKibanaCertSecret) are not something we need to let the user set we should not store those in the operator namespace.

Until we phase out BYO certs (@rene-dekker created a task for this I believe), we still need these secrets in the operator namespace.

pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -181,7 +191,7 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 		return fmt.Errorf("log-storage-controller failed to watch the ConfigMap resource: %w", err)
 	}
 
-	if err := utils.AddServiceWatch(c, render.ElasticsearchServiceName, render.ElasticsearchNamespace); err != nil {
+	if err := utils.AddServiceWatch(c, render.ElasticsearchInternalServiceName, render.ElasticsearchNamespace); err != nil {
```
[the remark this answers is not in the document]
vberezny: I think this should now be both since they are created in this controller now. I've added watches for ElasticsearchServiceName and  esgateway.ServiceName now that we've gone away from the `internal` naming.

pkg/controller/logstorage/logstorage_controller.go
vberezny: I've changed this up quite a bit, let me know if the new function makes sense.

pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -639,25 +687,33 @@ func (r *ReconcileLogStorage) deleteInvalidECKManagedPublicCertSecret(ctx contex
...
+// components to communicate with Elasticsearch/Kibana through ES Gateway. The order of the secrets returned are:
+// 1) The certificate secret needed for ES Gateway (in the operator namespace). If the user didn't create this it is
 //    created.
-// 2) The certificate secret needed for Elasticsearch (in the Elasticsearch namespace).
-// 3) The certificate mounted by other clients that connect to Elasticsearch.
-func (r *ReconcileLogStorage) getElasticsearchCertificateSecrets(ctx context.Context, instl *operatorv1.InstallationSpec) (oprKeyCert *corev1.Secret, esKeyCert *corev1.Secret, certSecret *corev1.Secret, err error) {
+// 2) The certificate secret needed for ES Gateway (in the Elasticsearch namespace).
+// 3) The certificate mounted by other clients that connect to Elasticsearch/Kibana through ES Gateway (in the operator namespace).
+// 4) The certificate mounted by es curator and es metrics to connect to Elasticsearch/Kibana through ES Gateway (in the Elasticsearch namespace).
+func (r *ReconcileLogStorage) getESGatewayCertificateSecrets(ctx context.Context, instl *operatorv1.InstallationSpec) (
```
rene-dekker, line 697: From the looks of it, the newly introduced 4th return variable is redundant. It is used in two places:
- to calculate a hash at a place where the 3rd variable would give an identical hash
- to pass to the elasticsearch_metrics.go, where the secret is already copied to the es namespace

pkg/render/kube-controllers.go
```diff
@@ -54,55 +61,93 @@ func KubeControllers(
...
+	// the user secret, a hashed version of the credentials is stored in the tigera-elasticsearch namespace for ES Gateway to retrieve and use to compare
+	// the dummy credentials, and a secret containing real admin level credentials is created and stored in the tigera-elasticsearch namespace to be swapped in once
+	// ES Gateway has confirmed that the dummy credentials match.
+	hashedUserSecret := kubeControllersVerificationSecret
+	if hashedUserSecret == nil {
+		hashedPassword, _ := bcrypt.GenerateFromPassword(kubeControllersUserSecret.Data["password"], bcrypt.MinCost)
+		hashedUserSecret = &v1.Secret{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      ElasticsearchKubeControllersVerificationUserSecret,
+				Namespace: ElasticsearchNamespace,
```
rene-dekker, line 93: If I get it right, the kube controllers repo will be changed to monitor this secret, rather than the tigera-secure-es-user (or something like that). If that assumption is correct:
- The secret used by kube controllers should be in the calico-system namespace
- The render function of kube controllers should not be responsible for creating anything in the elasticsearch namespace. (Is the existence of that namespace even guaranteed here?) Normally, it copies things to the tigera-operator namespace.
- We should restrict rbac such that it can no longer access secrets that it does not need.

pkg/render/kube-controllers.go
```diff
@@ -558,3 +609,10 @@ func kubeControllersVolumes(defaultMode int32, managerSecret *v1.Secret) []v1.Vo
 
 	return []v1.Volume{}
 }
+
+func randomString(length int) string {
```
rene-dekker, line 613: An identical func already exists in rand.go

rene-dekker: Or even better, one that doesnt need to be urlencoded in crypto.go.

I guess you can use that one and remove the other two functions as they are unused.

pkg/render/logstorage.go
```diff
@@ -67,14 +67,14 @@ const (
...
-	KibanaName             = "tigera-secure"
-	KibanaNamespace        = "tigera-kibana"
-	KibanaPublicCertSecret = "tigera-secure-kb-http-certs-public"
-	TigeraKibanaCertSecret = "tigera-secure-kibana-cert"
-	KibanaDefaultCertPath  = "/etc/ssl/kibana/ca.pem"
-	KibanaBasePath         = "tigera-kibana"
-	KibanaServiceName      = "tigera-secure-kb-http"
+	KibanaName               = "tigera-secure"
+	KibanaNamespace          = "tigera-kibana"
+	KibanaPublicCertSecret   = "tigera-secure-es-gateway-http-certs-public"
```
rene-dekker, line 72: I suppose we can simplify the secret name to `tigera-es-gateway`.

pkg/render/logstorage/esgateway/esgateway.go
```diff
@@ -0,0 +1,324 @@
...
+		TypeMeta: metav1.TypeMeta{Kind: "Role", APIVersion: "rbac.authorization.k8s.io/v1"},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      RoleName,
+			Namespace: render.ElasticsearchNamespace,
+		},
+		Rules: []rbacv1.PolicyRule{
+			{
+				APIGroups:     []string{""},
+				Resources:     []string{"secrets"},
+				ResourceNames: []string{},
```
rene-dekker, line 146: It would be a good practice to explicitly list the secrets you would like to watch, if possible.

pkg/render/logstorage/esgateway/esgateway.go
```diff
@@ -0,0 +1,324 @@
...
+	"github.com/tigera/operator/pkg/render/common/secret"
+)
+
+const (
+	DeploymentName          = "tigera-es-gateway"
+	ServiceAccountName      = "tigera-es-gateway"
+	RoleName                = "tigera-es-gateway"
+	VolumeName              = "tigera-es-gateway-certs"
+	ElasticsearchUserSecret = "tigera-es-gateway-elasticsearch-access"
+	ServiceName             = "tigera-secure-es-gateway-http"
```
rene-dekker, line 28: We could just call it `tigera-es-gateway`

pkg/render/manager.go
```diff
@@ -21,9 +21,8 @@ import (
 
 	tigerakvc "github.com/tigera/operator/pkg/render/common/authentication/tigera/key_validator_config"
 
-	"github.com/tigera/operator/pkg/render/common/authentication"
-
 	ocsv1 "github.com/openshift/api/security/v1"
+	"github.com/tigera/operator/pkg/render/common/authentication"
```
rene-dekker, line 25: I was not sure where to place this comment, but I don't really see any change that makes components reach kibana through voltron. Does this change exist solely in the voltron pod? Shouldn't there be some env set in here to dictate this?

pkg/controller/intrusiondetection/intrusiondetection_controller.go
```diff
@@ -144,6 +144,15 @@ func add(mgr manager.Manager, c controller.Controller) error {
 		return fmt.Errorf("intrusiondetection-controller failed to watch the Secret resource: %v", err)
 	}
 
+	// These watches are here to catch a modification to the resources we create in reconcile so the changes would be corrected.
```
rene-dekker, line 147: I am just commenting in a random line. I do not see anything in this PR that routes Kibana requests through es-gateway. I was under the impression that this would be part of the work.

vberezny: The components that need to reach Kibana can do so through the new `tigera-secure-es-gateway-http` service. I changed the endpoint in `pkg/render/common/kibana/service.go` which is what components use to set their KIBANA_HOST value.

pkg/render/kube-controllers.go
vberezny: The ElasticsearchKubeControllersVerificationUserSecret stores the hashed password from ElasticsearchKubeControllersUserSecret and is used by ES gateway to authenticate the dummy credentials that kube-controllers uses in its requests. It needs to live in the elasticsearch namespace so that the gateway can access it. 

> If I get it right, the kube controllers repo will be changed to monitor this secret, rather than the tigera-secure-es-user.

Kube controllers will still use the same secret name as before, the difference is that it will contain dummy credentials now created in the lines above this comment. The admin level credentials will be stored in the Elasticsearch namespace and es gateway will swap them in once it authenticates the dummy credentials.

> The render function of kube controllers should not be responsible for creating anything in the elasticsearch namespace. (Is the existence of that namespace even guaranteed here?) Normally, it copies things to the tigera-operator namespace.

I'm not sure how to get around this, I guess one option would be to move this responsibility to the esgateway component? Have kube-controllers just generate the secret with the dummy credentials (ElasticsearchKubeControllersUserSecret) and then use that to create ElasticsearchKubeControllersVerificationUserSecret and ElasticsearchKubeControllersSecureUserSecret in esgateway rendering.

pkg/render/logstorage/esgateway/esgateway.go
Brian-McM: I think we wanted all the names to be `tigera-secure`, not just `tigera`. Maybe that's changed though.

pkg/controller/installation/core_controller.go
Brian-McM: The kube controllers controller should not be responsible for copying to the elasticsearch namespace, that should be handled by the logstorage controller.

pkg/render/kube-controllers.go
Brian-McM: @rene-dekker is right (I commented on this earlier as well), kube controllers shouldn't create anything in the elasticsearch namespace.

These secrets aren't generated by the core controller, there's no need to pass them down to be copied here. The Logstorage controller should be responsible for copying these secrets to the elasticseach namespace, the only secret this controller should be responsible for is the `ElasticsearchKubeControllersVerificationUserSecret`, because that's the only one it actually uses.

Brian-McM (changes requested):

pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -639,25 +687,33 @@ func (r *ReconcileLogStorage) deleteInvalidECKManagedPublicCertSecret(ctx contex
...
+	oprKeyCert *corev1.Secret,
```
Brian-McM, line 698: Could you please remove the named return types and keep things explicit, we already have explicit returns (I don't think this should have been here in the first place). 

The named returns, although valid in golang, cause more problems then they solve. With an explicit return you know at a glance what's being returned unlike with the named return types, and if you want to null out other values and return an error, you either need to remember to null out the values before returning or still do an explicit return (as is done here) and you've gained very little with this mix and match.

pkg/controller/logstorage/logstorage_controller.go
Brian-McM: @vberezny, extending on this I think you should only need to pass back the operator name spaced secrets, and then just get the Elasticsearch controller to copy those secrets to the Elasticsearch namespace explicitly. 

I think you should also just hash on the secrets put in the elasticsearch namespace (this is an improvement to what we have). We don't want to roll over Elasticsearch when the operator secrets change, it's a little too preemptive Elasticsearch doesn't actually use those secrets. We want Elasticsearch to roll over only when the secrets it actually uses (the copies) change.

As for es metrics, just pass the name of the secret down, or use the hard coded value.

pkg/render/kube-controllers.go
```diff
@@ -54,55 +61,93 @@ func KubeControllers(
...
+	kubeControllersUserSecret *v1.Secret,
+	kubeControllersVerificationSecret *v1.Secret,
+	esAdminUserSecret *v1.Secret,
 	metricsPort int,
 ) *kubeControllersComponent {
-	var elasticsearchUserSecret *v1.Secret
-	if esAdminSecret != nil {
-		elasticsearchUserSecret = &v1.Secret{
+	var kubeControllersSecureUserSecret *v1.Secret
+	if kubeControllersUserSecret == nil {
```
Brian-McM, line 70: I think it would make sense to get the logstorage controller to create this secret.

Thinking out loud here, this isn't really a secret for kube controllers, it's the "dummy" representation of the admin secret, so maybe it should be named accordingly.

pkg/render/kube-controllers.go
```diff
@@ -140,8 +185,14 @@ func (c *kubeControllersComponent) Objects() ([]client.Object, []client.Object)
 			secret.CopyToNamespace(common.CalicoNamespace, c.elasticsearchSecret)...)...)
 	}
 
-	if !c.isManagedCluster() && c.elasticsearchUserSecret != nil {
-		objectsToCreate = append(objectsToCreate, secret.ToRuntimeObjects(c.elasticsearchUserSecret)...)
+	if !c.isManagedCluster() && c.alternateUserSecret != nil {
```
Brian-McM, line 188: Are we removing the "hashed" and "alternate" naming scheme? Didn't we replace that?

pkg/render/logstorage.go
Brian-McM: I think all of them should have the `secure` in it.

pkg/render/logstorage.go
```diff
@@ -411,9 +410,7 @@ func (es *elasticsearchComponent) Objects() ([]client.Object, []client.Object) {
 	} else {
 		toCreate = append(toCreate,
 			createNamespace(ElasticsearchNamespace, es.installation.KubernetesProvider),
-			createNamespace(KibanaNamespace, es.installation.KubernetesProvider),
```
Brian-McM, line 414: What's creating the kibana namespace now? Was this a mistake to remove this?

pkg/render/logstorage/esgateway/esgateway.go
```diff
@@ -0,0 +1,324 @@
...
+		installation:   installation,
+		pullSecrets:    pullSecrets,
+		secrets:        secrets,
+		tlsAnnotations: tlsAnnotations,
+		clusterDomain:  clusterDomain,
+	}
+}
+
+type esGateway struct {
+	logStorage     *operatorv1.LogStorage
```
Brian-McM, line 84: I don't see any reference to logStorage, could you please remove it if it's not used.

pkg/render/logstorage/esgateway/esgateway.go
```diff
@@ -0,0 +1,324 @@
...
+	// tigera-es-gateway-elasticsearch-access
+	if esAdminUserSecret != nil {
+		elasticsearchUserSecret := &corev1.Secret{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      ElasticsearchUserSecret,
+				Namespace: render.ElasticsearchNamespace,
+			},
+			Data: map[string][]byte{
+				"username": []byte("elastic"),
+				"password": esAdminUserSecret.Data["elastic"],
```
Brian-McM, line 61: I think this same secret is renamed a few times, maybe you should just create a single copy of the admin secret in the Elasticsearch namespace and anything that needs that secret can just mount that.

Brian-McM (changes requested):

pkg/render/common/elasticsearch/service.go
```diff
@@ -7,8 +7,8 @@ import (
 )
 
 const (
-	httpsEndpoint     = "https://tigera-secure-es-http.tigera-elasticsearch.svc:9200"
-	httpsFQDNEndpoint = "https://tigera-secure-es-http.tigera-elasticsearch.svc.%s:9200"
+	httpsEndpoint     = "https://tigera-secure-es-gateway-http.tigera-elasticsearch.svc:9200"
```
Brian-McM, line 10: Does this mean that there was some agreement reached that we're going to require customers with managed clusters to upgrade those clusters when they upgrade the management cluster?

vberezny: The customers won't need to upgrade the managed clusters, they will still have the old code running meaning that their components will still be configured to hit the old service. The new certificate passed down to these components will contain `tigera-secure-es-http` in the SANs field in order to support these older clusters/service names. Voltron will be configured to forward requests to the new es-gateway service. I've verified that this works by testing on 3.7 managed cluster connected to a custom management cluster running this operator image.

pkg/render/logstorage/esgateway/esgateway.go
vberezny: Looks like this was missed after some changes to es-gateway no longer require it to have access to admin credentials. We will now only need one secret with these credentials (for kube-controllers in the elasticsearch namespace).

pkg/render/logstorage.go
vberezny: This code path is for Managed clusters only and they do not need the kibana namespace or service any longer since there are no components in Managed clusters that make calls to Kibana. Please let me know if that's not the case but I'm fairly certain that it is.

pkg/render/kube-controllers.go
vberezny: That's right, I forgot to update this after we decided on the naming scheme change, will update now.

pkg/render/logstorage/esgateway/esgateway.go
vberezny: I'm not sure that will be possible given that each component will have 2 secrets each living in this namespace, and we can have any amount of managed cluster components connected, each of them also having 2 secrets per component.

Certainly for the management/standalone secrets we can explicitly list every components secret, but things get tricky with managed clusters since their secrets follow the format `component-name-cluster-name-elasticsearch-access-gateway` or `component-name-cluster-name-gateway-verification-credentials`. Is it worth looking into a way to list every connected managed cluster and to build a list of secrets that es-gateway is allowed to access explicitly? If so, maybe that can be a follow up task?

pkg/controller/intrusiondetection/intrusiondetection_controller.go
```diff
@@ -144,6 +144,15 @@ func add(mgr manager.Manager, c controller.Controller) error {
...
+	if err = utils.AddSecretsWatch(c, relasticsearch.PublicCertSecret, render.IntrusionDetectionNamespace); err != nil {
```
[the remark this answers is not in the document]
vberezny: I found that strange as well, must have been missed previously.

rene-dekker (approved):
The PR looks good to me now, provided that we address certificate management before the upcoming release in a separate PR as discussed. 

Please also have another reviewer approve it before merging. It is quite an important set of changes.

pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -167,12 +171,17 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 	// Watch all the secrets created by this controller so we can regenerate any that are deleted
 	for _, secretName := range []string{
 		render.TigeraElasticsearchCertSecret, render.TigeraKibanaCertSecret,
-		render.OIDCSecretName, render.DexObjectName} {
+		render.OIDCSecretName, render.DexObjectName, relasticsearch.PublicCertSecret} {
```
tmjd, line 174: Will `relasticsearch.PublicCertSecret` exist in the operator namespace? My current assumption is that it has just the cert portion of TigeraElasticsearchCertSecret, based on previous comments.

pkg/controller/logstorage/logstorage_controller.go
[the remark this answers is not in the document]
tmjd: I'm marking this resolved because I've got a new comment on this.

pkg/controller/logstorage/logstorage_controller.go
vberezny: That is correct, it will contain just the cert potion and will exist in the Operator namespace. From there it will be copied to any namespace that has components that need to communicate with the gateway.

pkg/render/common/elasticsearch/service.go
Brian-McM: OK that sounds like a good plan.

pkg/render/logstorage.go
Brian-McM: No you're right, I just didn't realise this was for a managed cluster. Thanks for cleaning that up.

pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -639,25 +709,74 @@ func (r *ReconcileLogStorage) deleteInvalidECKManagedPublicCertSecret(ctx contex
...
+			},
+		}
+	}
+
+	return kubeControllersVerificationSecret, kubeControllersSecureUserSecret, nil
+}
+
+// getESGatewayCertificateSecrets retrieves certificate secrets needed for ES Gateway to run or for
+// components to communicate with Elasticsearch/Kibana through ES Gateway. The order of the secrets returned are:
+// 1) The certificate secret needed for ES Gateway (in the operator namespace). If the user didn't create this it is created.
```
tmjd, line 762: the 2 comment is very clear. Could you include the same clarify for 1, by specifying where this is mounted and what it is used for? (I think, this secret is mounted into es-gateway for it to connect to ES/Kibana)

tmjd (commented):
I'm handing final operator maintainer review of this over to @Brian-McM.
But before I remove myself from this PR though can you @vberezny add a comment to this PR about what, if any, changes will be necessary as far as user provided secrets? If not changes in the operator what about in the docs and will there be a need for any release notes.
Sorry I haven't kept in my head the last state that BYO certs will be in once this PR goes in. Thanks

vberezny:
> I'm handing final operator maintainer review of this over to @Brian-McM.
> But before I remove myself from this PR though can you @vberezny add a comment to this PR about what, if any, changes will be necessary as far as user provided secrets? If not changes in the operator what about in the docs and will there be a need for any release notes.
> Sorry I haven't kept in my head the last state that BYO certs will be in once this PR goes in. Thanks

Sure no problem. If a customer chooses to bring their own certs, they can still follow the existing instructions here https://docs.tigera.io/security/comms/log-storage-tls. The name of the secret has not changed and the only difference is that now they cert/key is used by es-gateway not the Elasticsearch cluster. As far as release notes go, I can't think of anything that we would need to include.

pkg/controller/logstorage/logstorage_controller.go
vberezny: Updated the comment.

pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -561,6 +575,58 @@ func (r *ReconcileLogStorage) Reconcile(ctx context.Context, request reconcile.R
...
+		}
+
+		esInternalCertSecret, err = utils.GetSecret(ctx, r.client, relasticsearch.InternalCertSecret, render.ElasticsearchNamespace)
+		if err != nil {
+			reqLogger.Error(err, err.Error())
+			r.status.SetDegraded("Waiting for internal Elasticsearch cert secret to be available", "")
+			return reconcile.Result{}, err
+		}
+
+		kubeControllersGatewaySecret, err := utils.GetSecret(ctx, r.client, render.ElasticsearchKubeControllersUserSecret, common.CalicoNamespace)
```
Brian-McM, line 592: So I think this logic is a little backwards, kubecontrollers should not be creating this secret, the logstorage controller should be creating this secret. It should put it in the operator namespace, and kubecontrollers should copy it over to it's namespace if it exists.

Previously that's how this was done, except it wasn't just a simple copy, we'd create a new secret with same values of the admin secret.

pkg/render/kube-controllers.go
```diff
@@ -54,55 +57,55 @@ func KubeControllers(
...
 	enabledESOIDCWorkaround bool,
 	clusterDomain string,
-	esAdminSecret *v1.Secret,
+	kubeControllersGatewaySecret *v1.Secret,
 	metricsPort int,
 ) *kubeControllersComponent {
-	var elasticsearchUserSecret *v1.Secret
-	if esAdminSecret != nil {
-		elasticsearchUserSecret = &v1.Secret{
+	if kubeControllersGatewaySecret == nil {
```
Brian-McM, line 63: I made a comment in the logstorage controller about how this shouldn't be created here. This also introduces an problem for the operator when used in OS, where there is no Elasticsearch (i.e. don't created create elasticsearch secrets if there's no elasticsearch).

Brian-McM (changes requested):

pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -561,6 +575,58 @@ func (r *ReconcileLogStorage) Reconcile(ctx context.Context, request reconcile.R
 			return reconcile.Result{}, nil
 		}
 
+		kibanaInternalCertSecret, err := utils.GetSecret(ctx, r.client, render.KibanaInternalCertSecret, rmeta.OperatorNamespace())
+		if err != nil {
+			reqLogger.Error(err, err.Error())
+			r.status.SetDegraded("Waiting for internal Kibana cert secret to be available", "")
```
Brian-McM, line 581: Please change "cert" to "tls certificate", as this is customer facing.

pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -561,6 +575,58 @@ func (r *ReconcileLogStorage) Reconcile(ctx context.Context, request reconcile.R
...
+			return reconcile.Result{}, err
+		}
+
+		esInternalCertSecret, err = utils.GetSecret(ctx, r.client, relasticsearch.InternalCertSecret, render.ElasticsearchNamespace)
+		if err != nil {
+			reqLogger.Error(err, err.Error())
+			r.status.SetDegraded("Waiting for internal Elasticsearch cert secret to be available", "")
```
Brian-McM, line 588: Please change "cert" to "tls certificate", as this is customer facing.

pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -561,6 +575,58 @@ func (r *ReconcileLogStorage) Reconcile(ctx context.Context, request reconcile.R
...
+		}
+		if kubeControllersGatewaySecret == nil {
+			r.status.SetDegraded("Waiting for kube-controllers gateway user secret to be available", "")
+			return reconcile.Result{}, nil
+		}
+
+		kubeControllersVerificationSecret, kubeControllersSecureUserSecret, err := r.createKubeControllersSecrets(ctx, kubeControllersGatewaySecret, esAdminUserSecret)
+		if err != nil {
+			reqLogger.Error(err, err.Error())
+			r.status.SetDegraded("Failed to create kube-controllers secrets for ES gateway", "")
```
Brian-McM, line 604: customer facing message, ES => Elasticsearch.

pkg/render/logstorage.go
```diff
@@ -837,15 +834,8 @@ func (es elasticsearchComponent) elasticsearchCluster(secureSettings bool) *esv1
 			},
 		},
 		Spec: esv1.ElasticsearchSpec{
-			Version: components.ComponentEckElasticsearch.Version,
-			Image:   es.esImage,
-			HTTP: cmnv1.HTTPConfig{
```
Brian-McM, line 842: This might warrant further discussion, are we removing BYO certs for Elasticsearch right now?

I'm not sure if this also affects certificate management. Taking this out means that the ECK operator will generate a public certificate that for Elasticsearch (based off the private certificate it creates and stores in a secret), and we can't change this (or else the ECK operator will change it back).

This means that certificate management will be broken, since it can't override the public certificate, or you'd have to mount the default secret generated by elasticsearch if cert management wasn't enabled, and a different one if it was, which makes this a very uncomfortable configuration situation.

We always need to be the one to generate the certificate, ECK will copy the one created here to the "default" secret name instead of generating it. All default certificates must be generated by us. Please either change this back, or generate a differently named certificate secret that we generate by default.

CC @rene-dekker, is there something I'm missing here?

Brian-McM (dismissed):

Brian-McM (changes requested):
meant to request changes not approve

pkg/render/logstorage/esgateway/esgateway.go
```diff
@@ -0,0 +1,313 @@
...
+	clusterDomain string,
+) render.Component {
+	var certSecretsESCopy []*corev1.Secret
+	secrets := certSecrets
+
+	// Copy the Operator namespaced cert secrets to the Elasticsearch namespace.
+	for _, s := range certSecrets {
+		certSecretsESCopy = append(certSecretsESCopy, secret.CopyToNamespace(render.ElasticsearchNamespace, s)...)
+	}
+	tlsAnnotations := map[string]string{render.ElasticsearchTLSHashAnnotation: rmeta.SecretsAnnotationHash(append(certSecretsESCopy, esInternalCertSecret)...)}
```
Brian-McM, line 54: Isn't esInternalCertSecret already in the namespace? If so, please don't re copy it back.

pkg/render/logstorage/esgateway/esgateway_test.go
```diff
@@ -0,0 +1,99 @@
...
+			)
+
+			createResources, _ := component.Objects()
+			compareResources(createResources, expectedResources)
+		})
+
+	})
+})
+
+func compareResources(resources []client.Object, expectedResources []resourceTestObj) {
```
Brian-McM, line 85: I thought there was already a function like this?

pkg/render/manager.go
```diff
@@ -559,6 +558,7 @@ func (c *managerComponent) managerProxyContainer() corev1.Container {
 		{Name: "VOLTRON_KIBANA_CA_BUNDLE_PATH", Value: "/certs/kibana/tls.crt"},
 		{Name: "VOLTRON_ENABLE_MULTI_CLUSTER_MANAGEMENT", Value: strconv.FormatBool(c.managementCluster != nil)},
 		{Name: "VOLTRON_TUNNEL_PORT", Value: defaultTunnelVoltronPort},
+		{Name: "VOLTRON_DEFAULT_FORWARD_SERVER", Value: "tigera-secure-es-gateway-http.tigera-elasticsearch.svc:9200"},
```
Brian-McM, line 561: Completely forgot I put that in voltron, I thought you'd have to make some voltron changes. Yay for preemptive "seemingly useless" configuration options lol.

pkg/render/logstorage/esgateway/esgateway.go
vberezny: It is but I'm not copying it here, just adding it to the tls annotation so that gateway is restarted if that secret changes.

pkg/render/logstorage.go
rene-dekker: I think everything you say is essentially correct, @Brian-McM.
As for certificate management, there are many ways to get esgw to trust elasticsearch:
- Doing as Brian suggests would be the least number of code changes probably.
- Configure esgw to read es cert from /path/to/ca.crt from the init container (which is a bit hacky)
- Configure esgw to put  /path/to/ca.crt from the init container into the OS trusted cert directory.
I would be ok with any of them personally, but maybe for the sake of getting this PR ready, the first option would be a good choice.

Brian-McM: I'm not OK with that last two options until certificate management is done like that for all components. Currently cert management work by creating a public certificate secret from the private one created in the pod, so that's how it should stay for all components until a new design has been put forward to change this and has been approved.

pkg/render/logstorage/esgateway/esgateway.go
Brian-McM: Oh right derp, sorry, didn't realise this was for the annotation.

pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -666,58 +797,133 @@ func (r *ReconcileLogStorage) getElasticsearchCertificateSecrets(ctx context.Con
...
-			return nil, nil, nil, err
+			return nil, nil, err
 		}
 
-		if pubSecret != nil {
+		if internalSecret != nil {
 			// If the provided certificate secret (secret) is managed by the operator we need to check if the secret that
 			// Elasticsearch creates from that given secret (pubSecret) has the expected DNS name. If it doesn't, delete the
 			// public secret so it can get recreated.
 			if !customerProvidedCert {
```
Brian-McM, line 910: Could you add a follow up task to evaluate if this route is possible, I don't think it is since you've taken away BYO. Think about it later though, not now, let's try to get this through

Brian-McM (approved):

pkg/controller/logstorage/logstorage_controller.go
vberezny: Sure will do

vberezny: https://tigera.atlassian.net/browse/SAAS-1586

PR #1356, opened 2021-06-21
Add deletecollection permissions for deleting ES secrets

rene-dekker:
Adds the clusterrole binding to the kube controllers, so it can clean up unnecessary secrets that it has created previously.


Brian-McM (approved):

tmjd (approved):
LGTM

PR #1366, opened 2021-06-29
Add a packet capture container to tigera-guardian

asincu:
## Description

Add a new container for packet capture on guardian side. In a
multicluster setup, the management cluster will proxy a request via the
tunnel to Guardian. Guardian will forward the request to the packet
capture container. This API will retrieve the files generated by a
packet capture via a pod/exec command. 

Tigera-guardian will be used as a service account to perform all the needed
k8s API calls:
- get packetcaptures
- list fluentd pods
- exec in fluentd pods

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/guardian.go
+++ b/pkg/render/guardian.go
@@ -17,6 +17,11 @@
 package render
 
 import (
+	operatorv1 "github.com/tigera/operator/api/v1"
+	"github.com/tigera/operator/pkg/components"
+	rmeta "github.com/tigera/operator/pkg/render/common/meta"
+	"github.com/tigera/operator/pkg/render/common/podsecuritycontext"
+	"github.com/tigera/operator/pkg/render/common/secret"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	v1 "k8s.io/api/core/v1"
@@ -24,12 +29,6 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-
-	operatorv1 "github.com/tigera/operator/api/v1"
-	"github.com/tigera/operator/pkg/components"
-	rmeta "github.com/tigera/operator/pkg/render/common/meta"
-	"github.com/tigera/operator/pkg/render/common/podsecuritycontext"
-	"github.com/tigera/operator/pkg/render/common/secret"
 )
 
 // The names of the components related to the Guardian related rendered objects.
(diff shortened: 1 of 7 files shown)


pkg/render/guardian.go
```diff
@@ -220,8 +233,11 @@ func (c *GuardianComponent) deployment() client.Object {
 					ServiceAccountName: GuardianServiceAccountName,
 					Tolerations:        append(c.installation.ControlPlaneTolerations, rmeta.TolerateMaster, rmeta.TolerateCriticalAddonsOnly),
 					ImagePullSecrets:   secret.GetReferenceList(c.pullSecrets),
-					Containers:         c.container(),
-					Volumes:            c.volumes(),
+					Containers: []corev1.Container{
```
Brian-McM, line 236: From an architecture perspective I'm very hesitant about adding this to the guardian pod, since at first glance packet capturing has nothing to do with the guardian / voltron tunnel so it doesn't seem to make sense to couple it to this.

Can you explain why it needs to be part of the guardian pod?

Brian-McM (approved):
I've approved this, but I just want to leave a couple comments:

The packet capture API is now tied to both the manager and guardian. This has a few reprocussions:
1. We have two separate deployment paths / code paths for this API
2. We can't scale the manager, tigera, or the packet capture api alone
3. This API can only access through either guardian or the manager

We could avoid all of these issue by just having the packet capture contain in it's own pod, and use it as a separate service. The only downside is that we need to deal with certificates, which we do for all our components.

If we decide to do this later we may have upgrade issues (although maybe not, maybe it could be mitigated).

There's time constraints at play here so putting this through is understandable, but I just want to make sure this is understood.

pkg/render/guardian.go
```diff
@@ -17,19 +17,18 @@
...
+	"github.com/tigera/operator/pkg/render/common/secret"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	v1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-
-	operatorv1 "github.com/tigera/operator/api/v1"
```
tmjd, line 28: This is the way we've decided to format our includes, the project specific includes should come in a section after the external includes. Please put these back.

tmjd:
I think @Brian-McM makes some really good points, so much so that if we can hold off on providing an MCM method to pull these packet captures than I think we should.

> We can't scale the manager, tigera, or the packet capture api alone

This really resonates with me as I can see needing to be able to scale the manager or voltron independently and this change is going the wrong direction by adding another component to a pod with 2 components that already probably shouldn't be tied together.

asincu:
Closing this in favour of: https://github.com/tigera/operator/pull/1404

PR #1374, opened 2021-07-02
Enhance Monitor render unit tests

hjiawei:
Test rendered object Specs in Monitor render unit tests.

This is a unit test only change.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/monitor_test.go
+++ b/pkg/render/monitor_test.go
@@ -20,7 +20,9 @@ import (
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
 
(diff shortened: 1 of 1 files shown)


pkg/render/monitor_test.go
```diff
@@ -71,12 +73,83 @@ var _ = Describe("monitor rendering tests", func() {
 
 			if alertmanagerObj, ok := obj.(*monitoringv1.Alertmanager); ok {
 				com := components.ComponentPrometheusAlertmanager
-				expectedImage := fmt.Sprintf("%s%s:%s", components.PrometheusRegistry, com.Image, com.Version)
```
rene-dekker, line 74: Rather than iterating over all objects and trying to cast them, we have some logic to help you find the object you are looking for. This is how we do it in other tests, I would suggest you use that behaviour as well. For instance when you want to find a certain deployment from the rendered objects:
```
//dex_test.go:150
d := rtest.GetResource(resources, render.DexObjectName, render.DexNamespace, "apps", "v1", "Deployment").(*appsv1.Deployment)
Expect(...).To(...)
```

hjiawei: As suggested, moved to a separate test case. Thank you!

rene-dekker (approved):
lgtm

PR #1382, opened 2021-07-08
Separate out all the elasticsearch related kube-controllers into another deployment

xiumozhan:
We now want to have 2 kube-controllers pods running in the calico-system namespace (both in the same namespace because it's breaking change in a version skew scenario). The old calico-kube-controllers runs everything that doesn't relate to elasticsearch and another es-calico-kube-controllers running all the elasticsearch related controllers.


Brian-McM:
Could you please update the title and description to be more explicit about about what this PR is for.

Brian-McM (changes requested):
I don't think the logstorage_render code should be rendering kube controllers, I think there's going to be a lot of duplicated logic there. I think the kube controllers render code should be generic made generic enough that you could configure it to be either the kube controller for the calico-system namespace, or the elasticsearch namespace.

I also think that this may also be the time to refactor the kube controllers render code to be in it's own subpackage (take a look at the logstorage subpackage in render).

These things are probably best talked about offline, so ping me when you'd like to discuss it.

pkg/render/kubecontrollers/kube-controllers.go
```diff
@@ -99,36 +113,54 @@ func (c *kubeControllersComponent) SupportedOSType() rmeta.OSType {
 }
 
 func (c *kubeControllersComponent) Objects() ([]client.Object, []client.Object) {
+	var kubeControllerServiceAccountName, roleName, roleBindingName, kubeControllerName string
+	if c.cfg.Installation.Variant == operatorv1.TigeraSecureEnterprise && c.cfg.LogStorageExists &&
```
Brian-McM, line 117: This seems like an overly complicated way of signalling which kube-controller this is. I think that if this is going to be a single render component, then you should probably have all this figured out in the constructor so you don't a barrage of if statements riddled through out this component (when I'd mentioned doing it this way I said it only made sense to do it this way if we didn't have to have a bunch of if statements everywhere to decide which controller this is being rendered for).

What I would try to do is see if it makes sense to have two constructors for this component, and store the configuration in the component. One constructore would be NewElasticsearchKubeControllers, and the other NewSomethingElseKubecontrollers. Then you can set the controller specific information in the object and not have to figure out at different points in this code which controller this is.

DM me if you need some clarification on this.

pkg/controller/logstorage/eskubecontrollers.go
```diff
@@ -0,0 +1,136 @@
...
+		if err != nil {
+			r.status.SetDegraded(fmt.Sprintf("Error ensuring internal manager TLS certificate %q exists and has valid DNS names", render.ManagerInternalTLSSecretName), err.Error())
+			return reconcile.Result{}, false, err
+		}
+	}
+	kubeControllersCfg := kubecontrollers.KubeControllersConfiguration{
+		K8sServiceEp:                 k8sapi.Endpoint,
+		Installation:                 install,
+		ManagementCluster:            managementCluster,
+		ManagementClusterConnection:  managementClusterConnection,
```
Brian-McM, line 98: This component should never be rendered if there's a managementClusterConnection, as that means it's a managed cluster.

If you're only passing this information because the regular kube-controllers sets it, I don't think you should do it. Just don't set it.

Brian-McM (changes requested):

pkg/render/kubecontrollers/kube-controllers.go
```diff
@@ -290,57 +352,40 @@ func (c *kubeControllersComponent) controllersRole() *rbacv1.ClusterRole {
...
 	}
 
 	env = append(env, c.cfg.K8sServiceEp.EnvVars(false, c.cfg.Installation.KubernetesProvider)...)
 
-	enabledControllers := []string{"node"}
-	if c.cfg.Installation.Variant == operatorv1.TigeraSecureEnterprise {
-		enabledControllers = append(enabledControllers, "service", "federatedservices")
+	enabledControllers := make([]string, 0)
+	if c.kubeControllerName == KubeController {
+		enabledControllers = append(enabledControllers, "node")
```
Brian-McM, line 372: I think I'd like to see the decision of what controllers are enabled set in the constructors. Then you can render set them here.

pkg/render/kubecontrollers/kube-controllers.go
```diff
@@ -244,22 +269,47 @@ func (c *kubeControllersComponent) controllersRole() *rbacv1.ClusterRole {
...
-			},
-			{
-				APIGroups: []string{"rbac.authorization.k8s.io"},
-				Resources: []string{"clusterroles", "clusterrolebindings"},
-				Verbs:     []string{"watch", "list", "get"},
-			},
+		}
+
+		if c.kubeControllerName == KubeController {
+			extraRules = append(extraRules,
```
Brian-McM, line 275: Can you make this decision in the constructors as well please, and set them to a rules variable. Then you can just render the rules outright.

pkg/render/kubecontrollers/kube-controllers.go
```diff
@@ -106,29 +144,33 @@ func (c *kubeControllersComponent) Objects() ([]client.Object, []client.Object)
...
 		objectsToCreate = append(objectsToCreate, secret.ToRuntimeObjects(
 			secret.CopyToNamespace(common.CalicoNamespace, c.cfg.ManagerInternalSecret)...)...)
 	}
 
-	if c.cfg.ElasticsearchSecret != nil {
-		objectsToCreate = append(objectsToCreate, secret.ToRuntimeObjects(
-			secret.CopyToNamespace(common.CalicoNamespace, c.cfg.ElasticsearchSecret)...)...)
-	}
+	if c.kubeControllerName == EsKubeController {
+		if c.cfg.ElasticsearchSecret != nil {
```
Brian-McM, line 153: I wonder if for these situations we can still make the decision of whether or not the es secret is rendered in the constructor. You could add a boolean field `renderElasticsearchSecret`, and set that to true in the `NewElasticsearchKubeControllers` if `ElasticsearchSecret` is not nil. Then this becomes:

```
if c.renderElasticsearchSecret {
    objectsToCreate = append(objectsToCreate, secret.ToRuntimeObjects(
				secret.CopyToNamespace(common.CalicoNamespace, c.cfg.ElasticsearchSecret)...)...)
}
```

My preference would be to follow this pattern and remove the `kubeControllerConfigName` variable.

Brian-McM (changes requested):

Brian-McM (approved):
A couple of small personal preference nits (so you don't need to do them) but looks good, just DM me when you want me to merge it.

pkg/render/kubecontrollers/kube-controllers.go
```diff
@@ -69,8 +84,54 @@ type KubeControllersConfiguration struct {
...
 }
 
-func KubeControllers(cfg *KubeControllersConfiguration) *kubeControllersComponent {
-	return &kubeControllersComponent{cfg: cfg}
+func NewCalicoKubeControllers(cfg *KubeControllersConfiguration) *kubeControllersComponent {
+	kubeControllerRolePolicyRules := kubeControllersRoleCommonRules(cfg, KubeController)
+	enabledControllers := []string{"node"}
+	if cfg.Installation.Variant == operatorv1.TigeraSecureEnterprise {
+		kubeControllerRolePolicyRules = append(kubeControllerRolePolicyRules, kubeControllersRoleEnterpriseCommonRules(cfg)...)
+		kubeControllerRolePolicyRules = append(kubeControllerRolePolicyRules, calicoKubeControllersRoleSpecificRules()...)
```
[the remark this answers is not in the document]
xiumozhan: I assume you're talking about `calicoKubeControllersRoleSpecificRules` and `elasticsearchKubeControllersRoleSpecificRules` because other rules are common and shared by both constructors.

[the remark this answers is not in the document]
Brian-McM: Yes I was

PR #1389, opened 2021-07-16
Move Prometheus ClusterRole to Role

hjiawei:
Move `ClusterRole`s in API group `monitoring.coreos.com` to `tigera-prometheus`
namespaced `Role`s. Note that we still keep `list` and `watch` to be cluster
scoped due to the fact that watches are by default non-namespaced in
`controller-runtime`.

Tested on GCP kubernetes and AWs Openshift clusters.

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/monitor.go
+++ b/pkg/render/monitor.go
@@ -19,6 +19,7 @@ import (
 	"strings"
 
 	corev1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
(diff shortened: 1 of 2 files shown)


pkg/render/monitor.go
```diff
@@ -268,3 +273,58 @@ func (mc *monitorComponent) serviceMonitorElasicsearch() *monitoringv1.ServiceMo
 		},
 	}
 }
+
+func (mc *monitorComponent) role() *rbacv1.Role {
+	// Even though list and watch verbs are listed here, they are still cluster scoped
```
matthewdupre, line 278: Would this comment be better where the actual Cluster scoped Role is?  (I assume that still exists somewhere else?)

matthewdupre: I see - they're in the other repo.  I guess it was annoying to change kube-controllers to be more specific in its watch?

hjiawei:
Closing as `Role` and `RoleBinding` will be defined in `tigera-operator.yaml` file.

PR #1391, opened 2021-07-20
Enable FELIX_WIRGEUARDHOSTENCRYPTIONENABLED for AKS/EKS

aaaaaaaalex:
## Description

Felix needs to explicitly add host IP's to wireguard's trusted IP's list in order to work properly on AKS, and also on EKS so that it can properly encrypt cross-host traffic when making requests through services/nodeports.

This change toggles `FELIX_WIREGUARDHOSTENCRYPTIONENABLED` flag when using AKS, or when using EKS with CalicoCNI (EKS CNI still a ways off)

Related Felix PR: https://github.com/projectcalico/felix/pull/2897

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/node.go
+++ b/pkg/render/node.go
@@ -1240,6 +1240,8 @@ func (c *nodeComponent) nodeEnvVars() []v1.EnvVar {
 			// We also need to configure a non-default trusted DNS server, since there's no kube-dns.
 			nodeEnv = append(nodeEnv, v1.EnvVar{Name: "FELIX_DNSTRUSTEDSERVERS", Value: "k8s-service:openshift-dns/dns-default"})
 		}
+	case operator.ProviderAKS:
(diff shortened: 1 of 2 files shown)


aaaaaaaalex:
Open question for the operator devs...

If I were to add a felixconfig var that would force [this behaviour](https://github.com/tigera/operator/pull/1391/files#diff-d4d31d15e69aa9a0bad7776e758dfbd50b8e0e1a969fd16719f56c470cac4e7bR1243) to "true"/"false", I expect the docstring for it would be messy - There might end up being a lot of `defaults to "true" on X and "false" on Y` talk, which seems messy compared to the otherwise clear documentation for other felixconfig vars.  Would that be an unacceptable var to have? Thoughts on alternatives?

Brian-McM (approved):

Brian-McM:
@aaaaaaaalex I think what you did is correct, I don't think we want this configurable through the felixconfig since this isn't something that a user should be able to override right now. This can always be revisited, but this looks correct to me.

PR #1392, opened 2021-07-21
Set FELIX_ROUTETABLERANGE to 31-250 for AWS CNI

mikestephen:
AWS CNI will create routing tables with indexes from 2-15, so this
configuration will instruct Felix to no longer manage routing tables
in the range AWS CNI is using.

## Description

This change fixes this [bug](https://tigera.atlassian.net/browse/APP-95).

I'm unsure yet whether this change should be applied only for EKS with AWS CNI, or for all AWS CNI environments. I suspect this will be a problem on all AWS CNI environments.

I have not tested this fix yet, just putting it up for early feedback.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/node.go
+++ b/pkg/render/node.go
@@ -1246,6 +1246,8 @@ func (c *nodeComponent) nodeEnvVars() []v1.EnvVar {
 	case operator.PluginAmazonVPC:
 		nodeEnv = append(nodeEnv, v1.EnvVar{Name: "FELIX_INTERFACEPREFIX", Value: "eni"})
 		nodeEnv = append(nodeEnv, v1.EnvVar{Name: "FELIX_IPTABLESMANGLEALLOWACTION", Value: "Return"})
+		// Don't conflict with the AWS CNI plugin's routes.
+		nodeEnv = append(nodeEnv, v1.EnvVar{Name: "FELIX_ROUTETABLERANGE", Value: "31-250"})
 	case operator.PluginGKE:
 		// The GKE CNI plugin uses its own interface prefix.
 		nodeEnv = append(nodeEnv, v1.EnvVar{Name: "FELIX_INTERFACEPREFIX", Value: "gke"})
--- a/pkg/render/node_test.go
+++ b/pkg/render/node_test.go
@@ -1043,6 +1043,7 @@ var _ = Describe("Node rendering tests", func() {
 			}},
 			{Name: "FELIX_INTERFACEPREFIX", Value: "eni"},
 			{Name: "FELIX_IPTABLESMANGLEALLOWACTION", Value: "Return"},
+			{Name: "FELIX_ROUTETABLERANGE", Value: "31-250"},
 			{Name: "FELIX_ROUTESOURCE", Value: "WorkloadIPs"},
 			{Name: "FELIX_BPFEXTTOSERVICECONNMARK", Value: "0x80"},
 		}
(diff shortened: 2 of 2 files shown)


aaaaaaaalex:
From what I can tell, we only need to worry about EKS clusters here (AWS doesn't seem to support using this CNI outside of EKS). This is hinted at by banzai, which wont let you provision an AWS/Kubeadm cluster with anything other than CalicoCNI

aaaaaaaalex:
Deployed this fix to an EKS cluster. Prior to patch, I observed these ip rule deletions:
```
Deleted 1536:   from 192.168.4.34 to 192.168.0.0/16 lookup 2 
Deleted 1536:   from 192.168.20.113 to 192.168.0.0/16 lookup 2 
Deleted 1536:   from 192.168.2.97 to 192.168.0.0/16 lookup 2 
Deleted 1536:   from 192.168.1.17 to 192.168.0.0/16 lookup 2 
Deleted 1536:   from 192.168.19.223 to 192.168.0.0/16 lookup 2 
Deleted 1536:   from 192.168.17.251 to 192.168.0.0/16 lookup 2
```

Post-patch, no-such deletions.

petercork:
This is important to get in I think as right now without this, EKS is simply broken once felix starts programming overlapping routing tables with EKS such as wireguard, egress, per-host proxy etc. The downside with this approach is that with upgrades there will be orphaned tables to be manually cleaned up. I am not sure if there is a better solution. Right now wireguard cannot work on EKS with AWS CNI with this bug. 

I would be happy to go with this for now but up to operator maintainers cc @tmjd

fasaxc:
Afraid I'm not an Operator maintainer; you need Erik or Casey

tmjd:
What i don't understand is the impact of this change on currently functioning clusters? This will change a value that was previously the default, is it (possibly) going to break or cause issues on an already existing cluster when they update to the operator with this change? I don't think saying people that upgrade will need to clean up orphaned tables is good enough.

Should we consider an alternative of adding to the documentation when enabling wireguard, egress, per-host proxy to include setting the configuration through FelixConfiguration. I'd rather have an issue when people are taking a deliberate action to enable a feature to adjust the configuration rather than it be a change in behavior on upgrade and potentially cause issues then.

mikestephen:
What I see from testing is that without FELIX_ROUTETABLERANGE set to something like 31-250, as soon as they exceed 5-6 pods, the node will get a second ENI assigned by AWS CNI, and it will use a table index of 2. Immediately I'm seeing that Felix will consider that this change must be removed since it did not sanction it. This is without even using Wireguard at all.
I'm surprised that this hasn't been raised by customers though?

Erik I see what you're saying, is there any way in the operator to make a change like in this PR, but have it only applied on install of new clusters rather than for existing ones?

Without this change we'd need to update the docs to advise setting the configuration through FelixConfiguration for all AWS CNI clusters, not just those using Wireguard. I completely see your point though, at least if they're setting it manually they can do it when it suits them, and are prepared to do any required cleanup.

tmjd:
There isn't a way to make this happen for only new clusters. We would have to add it to the operator Installation API and the only action we could do would be to set a default but since the field wouldn't exist on current installations we would default it on upgrade also.

I think someone mentioned possibly adding something to Felix to clean up old Felix tables, is that a real possibility?

mikestephen:
I was asking @fasaxc about Felix cleaning up the tables and it would be pretty messy and error-prone, since Felix no longer knows the table indexes to remove it would have to try and identify the rules and tables by other means.

fasaxc:
> What I see from testing is that without FELIX_ROUTETABLERANGE set to something like 31-250, as soon as they exceed 5-6 pods, the node will get a second ENI assigned by AWS CNI, and it will use a table index of 2. Immediately I'm seeing that Felix will consider that this change must be removed since it did not sanction it.

I think this is the swinger for me.  It seems that it's broken already!  I'm inclined to get the fix out there with a release note.  I think the vast majority of EKS operator users are ahead of us.  Changing this now will inconvenience a few (who had broken clusters anyway) but it'll save the vast majority of users who haven't switched top operator yet.

PR #1393, opened 2021-07-21
[SAAS-1454] Ignore all kubectl generated annotation during the Installation conversion

xiumozhan:
When the operator is first applied on a Calico cluster which is not managed by tigera-operator before, it will try to convert the cluster into an installation managed by the operator. One of the steps is it checks if the cluster's components has any incompatible annotation. During this check we need to ignore those annotations generated by kubectl, i.e. all the annotations start with `kubernetes.kubectl.io`.

### DIFF
--- a/pkg/controller/migration/convert/core.go
+++ b/pkg/controller/migration/convert/core.go
@@ -12,6 +12,7 @@ import (
 )
 
 var toBeIgnoredAnnotationKeyRegExps []*regexp.Regexp
+
 func init() {
 	log.Info("Compiling regular expressions for annotation keys to be ignored...")
 	toBeIgnoredAnnotationKeyRegExps = make([]*regexp.Regexp, 0)
--- a/pkg/controller/migration/convert/core.go
+++ b/pkg/controller/migration/convert/core.go
@@ -218,7 +218,7 @@ func addResources(install *operatorv1.Installation, compName operatorv1.Componen
 // since Operator does not support setting custom annotations on components, these annotations
 // would otherwise be dropped.
 func handleAnnotations(c *components, _ *operatorv1.Installation) error {
-	kubectlGenerated := []string {"^kubectl\\.kubernetes\\.io"}
+	kubectlGenerated := []string{"^kubectl\\.kubernetes\\.io"}
 	if a := removeExpectedAnnotations(c.node.Annotations, map[string]string{}, kubectlGenerated); len(a) != 0 {
 		return ErrIncompatibleAnnotation(a, ComponentCalicoNode)
 	}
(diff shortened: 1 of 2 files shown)


pkg/controller/migration/convert/core.go
```diff
@@ -255,20 +257,36 @@ func handleAnnotations(c *components, _ *operatorv1.Installation) error {
...
 			key == "deployment.kubernetes.io/revision" ||
 			key == "scheduler.alpha.kubernetes.io/critical-pod" {
 			delete(a, key)
 			continue
 		}
 
-		if v, ok := ignore[key]; ok && v == val {
+		if v, ok := ignoreWithValue[key]; ok && v == val {
 			delete(a, key)
 		}
```
stevegaossou, line 282: Seems like we could add a `continue` here as well, similar to the first if-block, since there's a new 3rd block that we can skip (if we have entered this second if-block).

pkg/controller/migration/convert/core.go
```diff
@@ -217,7 +218,8 @@ func addResources(install *operatorv1.Installation, compName operatorv1.Componen
 // since Operator does not support setting custom annotations on components, these annotations
 // would otherwise be dropped.
 func handleAnnotations(c *components, _ *operatorv1.Installation) error {
-	if a := removeExpectedAnnotations(c.node.Annotations, map[string]string{}); len(a) != 0 {
+	kubectlGenerated := []string{"^kubectl\\.kubernetes\\.io"}
```
stevegaossou, line 221: Minor: Might be more appropriate to have a generic name for this (e.g. regexIgnoreList). Since kubectlGenerated is just one specific example of an expected pattern that we want to remove from the annotations.

stevegaossou:
Hey @xiumozhan, were you able to test this out against a cluster?

pkg/controller/migration/convert/core.go
```diff
@@ -255,19 +257,36 @@ func handleAnnotations(c *components, _ *operatorv1.Installation) error {
...
 	a := existing
+	annotationKeyRegexps := make([]*regexp.Regexp, 0)
+	for _, annotationKey := range ignoreNoMatterWhatValue {
+		annotationKeyRegexp, err := regexp.Compile(annotationKey)
+		if err != nil {
+			log.Error(fmt.Errorf("%s is not a valid regular expression", annotationKey), "Error when removing expected annotations")
+			continue
+		}
+		annotationKeyRegexps = append(annotationKeyRegexps, annotationKeyRegexp)
+	}
```
stevegaossou, line 270: Minor: Seems like `removeExpectedAnnotations(...)` is called in various places (i.e. multiple times). It would be slightly more efficient to build these reg exps (`regexp.Compile(...)`) outside of this function, so they only need to be compiled once. Then they just need to be passed around. 

Going all the way out, perhaps instead of doing them within the scope of these functions, we could have an `init(...)` in this package that loads the regexes into a package level variable.

stevegaossou (approved):
LGTM

Brian-McM:
Can you please add a proper title and description.

PR #1394, opened 2021-07-21
Operator changes for process path collection in flowLogs

sridhartigera:
## Description
Process path and arguments are collected in flowLogs. In order to read the process path, arguments, felix will read /proc/pid/cmdline. This requires hostPID to be set to true in calico-node daemonset, set FELIX_FLOWLOGSCOLLECTPROCESSPATH env var. To enable/disable this feature, a new section (`collectProcessPath`) is added under LogCollector.

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


pkg/controller/logcollector/logcollector_controller.go
```diff
@@ -158,6 +158,7 @@ func GetLogCollector(ctx context.Context, cli client.Client) (*operatorv1.LogCol
 		return nil, err
 	}
 
+	fillDefaults(instance)
```
tmjd, line 161: I don't think we can do the fillDefaults in this function because we need the "raw" instance in the Reconcile to save as preDefaultPatchFrom. Otherwise the reconcile will not see the that we're adding the CollectProcessPath and NOT patch the resource with the default.

tmjd: I also think in the Reconcile we need to move the patching of the LogCollector resource earlier, especially now since the Core controller will be reading the resource.
Please also as this move I think you will need to move https://github.com/tigera/operator/pull/1394/files#diff-2b8c59da3e46730154a614a6d0ea55cfc27b17106a1e4bcc3fc8c12f794e8571R351-R358 the defaulting of  instance.Spec.AdditionalStores.Syslog.LogTypes into the fillDefaults function.
Hopefully this all makes sense, let me know if it doesn't.

tmjd (commented):
Looking good. I think we need a final decision on if "privilegedFeatures" is needed in the API.
Also I think we should adjust the defaulting of LogCollector since the core controller depends on LogCollector now.

pkg/controller/logcollector/logcollector_controller.go
```diff
@@ -171,11 +170,77 @@ func GetLogCollector(ctx context.Context, cli client.Client) (*operatorv1.LogCol
...
+							return false, nil, err
+						}
+					}
+				}
+			}
+
+			// Special case: For users that have a Syslog config and are upgrading from an older release
+			//  where logTypes field did not exist, we will auto-populate default values for
+			// them. This should only happen on upgrade, since logTypes is a required field.
+			if syslog.LogTypes == nil || len(syslog.LogTypes) == 0 {
```
tmjd, line 226: Only the block here should have been moved into fillDefaults. The above retrieving of the managementClusterConnection and the validation that IDSEvents is not set should be left where it was in the Reconcile.

pkg/controller/logcollector/logcollector_controller.go
```diff
@@ -171,11 +170,77 @@ func GetLogCollector(ctx context.Context, cli client.Client) (*operatorv1.LogCol
...
+				// option was not available prior to the logTypes field being introduced). This ensures
+				// existing users continue to get the same expected behavior for Syslog forwarding.
+				instance.Spec.AdditionalStores.Syslog.LogTypes = []v1.SyslogLogType{
+					v1.SyslogLogAudit,
+					v1.SyslogLogDNS,
+					v1.SyslogLogFlows,
+				}
+
+				// Mark LogCollector as changed so we know to save it
+				isModified = true
```
tmjd, line 237: I don't think we need this flag. In the core_controller we always patch so I think that is safe to do.

pkg/controller/logcollector/logcollector_controller.go
```diff
@@ -226,7 +286,10 @@ func (r *ReconcileLogCollector) Reconcile(ctx context.Context, request reconcile
 		r.status.SetDegraded("Error querying license", err.Error())
 		return reconcile.Result{RequeueAfter: 10 * time.Second}, nil
 	}
-
+	isModified, modifiedFields, err := r.fillDefaults(ctx, instance)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
```
tmjd, line 292: Please move patching the instance to after this so we can Patch as soon as possible.

tmjd (changes requested):

pkg/controller/logcollector/logcollector_controller.go
```diff
@@ -170,18 +170,47 @@ func GetLogCollector(ctx context.Context, cli client.Client) (*operatorv1.LogCol
 	return instance, nil
 }
 
+func fillDefaults(instance *operatorv1.LogCollector) []string {
```
tmjd, line 173: Please add a comment on this function that includes what the slice returned contains.

pkg/controller/logcollector/logcollector_controller_test.go
```diff
@@ -136,6 +136,13 @@ var _ = Describe("LogCollector controller tests", func() {
 	})
 
 	Context("image reconciliation", func() {
+		It("CollectProcessPath should be enabled if not set", func() {
```
tmjd, line 139: This test seems to be badly organized with almost everything in this "image reconciliation" context, but could you create a separate context to put this new `It` block in that is more appropriate. Maybe a new Context block of "defaulting logic".

tmjd (commented):
Looking good. A couple of minor things, I think we're mainly waiting on final decision on the API and if we need the `privilegedFeatures` sub-section.

tmjd (commented):
Has this been tested on a cluster that needs PodSecurityPolicies? https://kubernetes.io/docs/concepts/policy/pod-security-policy/#what-is-a-pod-security-policy suggests that access to hostPID is controlled with PodSecurityPolicy.

sridhartigera:
> Has this been tested on a cluster that needs PodSecurityPolicies? https://kubernetes.io/docs/concepts/policy/pod-security-policy/#what-is-a-pod-security-policy suggests that access to hostPID is controlled with PodSecurityPolicy.

yes. Tested with pod-security-policy enabled and made changes to operator to fix the same.

pkg/render/node.go
```diff
@@ -1388,6 +1405,10 @@ func (c *nodeComponent) nodePodSecurityPolicy() *policyv1beta1.PodSecurityPolicy
 	psp.Spec.AllowPrivilegeEscalation = ptr.BoolToPtr(true)
 	psp.Spec.Volumes = append(psp.Spec.Volumes, policyv1beta1.HostPath)
 	psp.Spec.HostNetwork = true
+	// CollectProcessPath feature in logCollectorSpec requires access to hostPID
+	// Hence setting hostPID to true in the calico-node PSP, for this feature
+	// to work with PSP turned on
+	psp.Spec.HostPID = true
```
tmjd, line 1411: I think this should be gated on `c.collectProcessPathEnabled()`.

tmjd (approved):
LGTM, definitely need to squash this one on merge :grinning:

PR #1397, opened 2021-07-22
Fix Elasticsearch service name for Managed clusters

vberezny:
This PR closes https://tigera.atlassian.net/browse/SAAS-1617.

### DIFF
--- a/pkg/render/logstorage.go
+++ b/pkg/render/logstorage.go
@@ -65,6 +65,7 @@ const (
 
 	ElasticsearchName                     = "tigera-secure"
 	ElasticsearchServiceName              = "tigera-secure-es-http"
+	ESGatewayServiceName                  = "tigera-secure-es-gateway-http"
 	ElasticsearchSecureSettingsSecretName = "tigera-elasticsearch-secure-settings"
 	ElasticsearchOperatorUserSecret       = "tigera-ee-operator-elasticsearch-access"
 	ElasticsearchAdminUserSecret          = "tigera-secure-es-elastic-user"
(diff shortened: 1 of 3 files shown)


pkg/render/logstorage.go
```diff
@@ -450,20 +451,6 @@ func (es elasticsearchComponent) elasticsearchExternalService() *corev1.Service
...
-			Name:      KibanaServiceName,
-			Namespace: KibanaNamespace,
-		},
-		Spec: corev1.ServiceSpec{
-			Type:         corev1.ServiceTypeExternalName,
-			ExternalName: fmt.Sprintf("%s.%s.svc.%s", GuardianServiceName, GuardianNamespace, es.clusterDomain),
-		},
-	}
-}
-
```
stevegaossou, line 466: So this function is dead code and this is just to clean it up now, right?

pkg/render/logstorage.go
```diff
@@ -450,20 +451,6 @@ func (es elasticsearchComponent) elasticsearchExternalService() *corev1.Service
 	}
 }
 
-func (es elasticsearchComponent) kibanaExternalService() *corev1.Service {
```
vberezny, line 453: This should have been removed earlier as well since it is no longer used given that Managed clusters don't have any components configured to hit this service.

stevegaossou (commented):
LGTM, but you'll need to get a Tigera operator dev team member to approve.

stevegaossou:
@vberezny looks like the build failed.

pkg/render/logstorage.go
vberezny: Yeah exactly

Brian-McM (approved):

pkg/render/logstorage.go
```diff
@@ -440,7 +441,7 @@ func (es elasticsearchComponent) elasticsearchExternalService() *corev1.Service
 	return &corev1.Service{
 		TypeMeta: metav1.TypeMeta{Kind: "Service", APIVersion: "v1"},
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      ElasticsearchServiceName,
+			Name:      ESGatewayServiceName,
```
Brian-McM, line 444: I don't really like that there's a reference in the Elasticsearch component specifically for the gateway here, but I think it might be more of a hassle to change this now. I guess when we refactor the logstorage render components we can tackle this, maybe separating the external service (and possibly any other managed cluster specific components) into their own component.

Brian-McM (approved):

PR #1399, opened 2021-07-22
Remove BYO Elasticsearch cert code cruft

vberezny:
This PR closes https://tigera.atlassian.net/browse/SAAS-1586.


Brian-McM:
Just to double check this is just taking away BYO for elasticsearch, but are we still allowing BYO for the gateway itself?

vberezny:
> Just to double check this is just taking away BYO for elasticsearch, but are we still allowing BYO for the gateway itself?

Yeah that's correct, BYO for gateway only with the possibility of phasing that out in the future.

Brian-McM (approved):

PR #1404, opened 2021-07-23
Extract packet capture as a service

asincu:
## Description

PacketCapture will be exposed as its own service and deployment.
Rendering these new resources will be triggered by the API server
installation resource. The Api server controller will create a new
namespace, service account, service, cluster role, clusterrolebinding,
and deployment. The service will be accesible via HTTPS. Generating
certificates will be done either via certificate management init
container or by providing a certificate in new added secret.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/apiserver/apiserver_controller.go
+++ b/pkg/controller/apiserver/apiserver_controller.go
@@ -19,6 +19,8 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/tigera/operator/pkg/dns"
+
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -120,6 +122,10 @@ func add(mgr manager.Manager, r *ReconcileAPIServer) error {
 		return fmt.Errorf("apiserver-controller failed to watch the Secret resource: %v", err)
 	}
 	if err = utils.AddSecretsWatch(c, "tigera-apiserver-certs", rmeta.OperatorNamespace()); err != nil {
+		return fmt.Errorf("apiserver-controller failed to watch the Secret resource: %v", err)
+	}
+
+	if err = utils.AddSecretsWatch(c, render.PacketCaptureCertSecret, rmeta.OperatorNamespace()); err != nil {
 		return fmt.Errorf("apiserver-controller failed to watch the Secret resource: %v", err)
 	}
 
@@ -284,6 +290,7 @@ func (r *ReconcileAPIServer) Reconcile(ctx context.Context, request reconcile.Re
 
 	// Render the desired objects from the CRD and create or update them.
 	reqLogger.V(3).Info("rendering components")
+	var components []render.Component
 	component, err := render.APIServer(k8sapi.Endpoint, network, false, managementCluster, managementClusterConnection, amazon, tlsSecret, pullSecrets, r.provider == operatorv1.ProviderOpenShift,
 		tunnelCASecret, r.clusterDomain)
 	if err != nil {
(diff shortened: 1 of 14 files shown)


tmjd:
A couple notes from reading the description:
* Does packet capture depend on any other resource existing (like IDS or Compliance)? If it does then I think this service should be deployed based on them, if not then the API Server is probably the correct resource to trigger on. (Is there any concern for a minimal, networking and apiserver, install? I don't think there would be.)
* For the certificates you don't seem to include the auto-generated certificates that we do for all our certificates currently. Was that on purpose because I think we need to keep that default certificate behavior until we have something better.

asincu:
@tmjd It does not rely on either IDS or Compliance, not is is related to those features in any way. This service is part of a default installation of Calico Enterprise. I first tried to tie the controller with the Installation CR, but after talking with @caseydavenport that seemed too aggressive. 

```
For the certificates you don't seem to include the auto-generated certificates that we do for all our certificates currently. Was that on purpose because I think we need to keep that default certificate behavior until we have something better.
```
I am not sure I follow this statement. I added support to generate certificates via CSR (we configure an init csr container that generates certificates) and also offering users the possibility to replace the certficates with their own (until we deprecate this from our product, I think we still need to support it). What other auto-generated certificate do we currently support ? If I missed something, please point me to an example of how we are doing that.

tmjd:
@asincu and I chatted and my concerns are resolved.

pkg/render/packetcapture_api_test.go
```diff
@@ -0,0 +1,373 @@
...
+					Name:  "PACKETCAPTURE_API_OIDC_AUTH_ISSUER",
+					Value: "https://127.0.0.1/dex",
+				},
+				{
+					Name:  "PACKETCAPTURE_API_OIDC_AUTH_JWKSURL",
+					Value: "https://tigera-dex.tigera-dex.svc.cluster.local:5556/dex/keys",
+				},
+				{
+					Name:  "PACKETCAPTURE_API_OIDC_AUTH_CLIENT_ID",
+					Value: "tigera-manager",
```
asincu, line 144: @rene-dekker Is this value correct ? This seem what is being generated for a default dex installation.

pkg/controller/apiserver/apiserver_controller.go
```diff
@@ -284,25 +290,78 @@ func (r *ReconcileAPIServer) Reconcile(ctx context.Context, request reconcile.Re
...
+		}
+
+		keyValidatorConfig, err := utils.GetKeyValidatorConfig(ctx, r.client, authenticationCR, r.clusterDomain)
+		if err != nil {
+			log.Error(err, "Failed to process the authentication CR.")
+			r.status.SetDegraded("Failed to process the authentication CR.", err.Error())
+			return reconcile.Result{}, err
+		}
+
+		var pc = render.PacketCaptureAPI(pullSecrets, r.provider == operatorv1.ProviderOpenShift, network, keyValidatorConfig, serverCertSecret, r.clusterDomain)
```
Brian-McM, line 349: I'm not sure that this should be managed by the apiserver controller. What the reasoning that it should be handled by the apiserver controller and not it's own? 

The biggest red flag for me is that it's not even in the same namespace as the apiserver, so what does it have to do with it?

pkg/render/packet_capture_api.go
```diff
@@ -0,0 +1,340 @@
+// Copyright (c) 2019-2020 Tigera, Inc. All rights reserved.
```
Brian-McM, line 1: Copyright needs updating.

pkg/render/packet_capture_api.go
```diff
@@ -0,0 +1,340 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package render
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/tigera/operator/pkg/dns"
```
Brian-McM, line 21: nit but could you please keep the operator imports together.

pkg/controller/apiserver/apiserver_controller.go
asincu: I do not think we want to introduce a new installation CR for this. This service should be part of the default installation of Calico Enterprise.  Creating PacketCaptures is part of the default installation. The service that allows you to download capture files should also be part of the default installation. We will also have a default UI integration which needs this service to be present.

pkg/render/packet_capture_api.go
asincu: Sure, will do. I normally run a `make fix` on the PR. Are we using different rules for imports in operator ?

pkg/render/packetcapture_api_test.go
rene-dekker: Yes, that is right

rene-dekker: See: https://github.com/tigera/operator/blob/fb4e8f655c0676b1f6f47076dcf7cefd0ff3650c/pkg/render/dex.go#L44

pkg/controller/clusterconnection/clusterconnection_controller.go
```diff
@@ -165,20 +170,38 @@ func (r *ReconcileConnection) Reconcile(ctx context.Context, request reconcile.R
 	tunnelSecret := &corev1.Secret{}
 	err = r.Client.Get(ctx, types.NamespacedName{Name: render.GuardianSecretName, Namespace: rmeta.OperatorNamespace()}, tunnelSecret)
 	if err != nil {
-		r.status.SetDegraded("Error copying secrets to the guardian namespace", err.Error())
+		r.status.SetDegraded("Error retrieving secrets to the guardian namespace", err.Error())
```
rene-dekker, line 173: retrieving ... from

pkg/render/manager.go
```diff
@@ -783,12 +755,10 @@ func managerClusterRole(managementCluster, managedCluster, openshift bool) *rbac
 		)
 	}
 
-	if !managedCluster {
+	if managementCluster {
```
rene-dekker, line 758: Just a double check; Standalone cluster do not need this, but management does?

pkg/render/packetcapture_api_test.go
rene-dekker: Please use the convenience method. If something is missing, you can either add it or rectify in the other repo.
```
func (d *dexKeyValidatorConfig) RequiredEnv(prefix string) []corev1.EnvVar {
	return []corev1.EnvVar{
		{Name: fmt.Sprintf("%sDEX_ENABLED", prefix), Value: strconv.FormatBool(true)},
		{Name: fmt.Sprintf("%sDEX_ISSUER", prefix), Value: fmt.Sprintf("%s/dex", d.ManagerURI())},
		{Name: fmt.Sprintf("%sDEX_URL", prefix), Value: fmt.Sprintf("https://tigera-dex.tigera-dex.svc.%s:5556/", d.clusterDomain)},
		{Name: fmt.Sprintf("%sDEX_JWKS_URL", prefix), Value: fmt.Sprintf(jwksURI, d.clusterDomain)},
		{Name: fmt.Sprintf("%sDEX_CLIENT_ID", prefix), Value: DexClientId},
		{Name: fmt.Sprintf("%sDEX_USERNAME_CLAIM", prefix), Value: d.UsernameClaim()},
		{Name: fmt.Sprintf("%sDEX_GROUPS_CLAIM", prefix), Value: defaultGroupsClaim},
		{Name: fmt.Sprintf("%sDEX_USERNAME_PREFIX", prefix), Value: d.authentication.Spec.UsernamePrefix},
		{Name: fmt.Sprintf("%sDEX_GROUPS_PREFIX", prefix), Value: d.authentication.Spec.GroupsPrefix},
	}
}
```

pkg/controller/apiserver/apiserver_controller.go
```diff
@@ -284,25 +290,78 @@ func (r *ReconcileAPIServer) Reconcile(ctx context.Context, request reconcile.Re
...
+		if network.CertificateManagement == nil {
+			serverCertSecret, err = utils.ValidateCertPair(r.client,
+				rmeta.OperatorNamespace(),
+				render.PacketCaptureCertSecret,
+				v1.TLSPrivateKeyKey,
+				v1.TLSCertKey,
+			)
+			if err != nil {
+				log.Error(err, fmt.Sprintf("failed to retrieve / validate %s", render.PacketCaptureCertSecret))
+				r.status.SetDegraded(fmt.Sprintf("failed to retrieve / validate  %s", render.PacketCaptureCertSecret), err.Error())
```
Brian-McM, line 315: Nit, but this is a user facing message so please capitalise the first let of the sentence.

pkg/controller/apiserver/apiserver_controller.go
```diff
@@ -284,25 +290,78 @@ func (r *ReconcileAPIServer) Reconcile(ctx context.Context, request reconcile.Re
...
 	if err != nil {
 		log.Error(err, "Error rendering APIServer")
 		r.status.SetDegraded("Error rendering APIServer", err.Error())
 		return reconcile.Result{}, err
 	}
+	components = append(components, component)
+
+	if variant == operatorv1.TigeraSecureEnterprise {
+
+		var serverCertSecret *v1.Secret
```
Brian-McM, line 305: could you please prepend this variable with "packetCapture" just so it's clear this is specifically for the packet capture component.

pkg/render/fluentd.go
```diff
@@ -395,8 +395,8 @@ func (c *fluentdComponent) packetCaptureApiRoleBinding() *rbacv1.RoleBinding {
 		Subjects: []rbacv1.Subject{
 			{
 				Kind:      "ServiceAccount",
-				Name:      ManagerServiceAccount,
```
Brian-McM, line 398: Just to double check, this was only here because the packet capture was originally attached to the manager right?

pkg/render/guardian.go
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2020 Tigera, Inc. All rights reserved.
```
Brian-McM, line 1: Does guardian have defaults for the certificate location and service name hard coded?

Brian-McM (approved):
Couple of nits / a question about if guardian has default configuration for the packet capture api name / the certificate secret, but otherwise look good.

pkg/render/fluentd.go
asincu: Yes, that is correct.

pkg/render/guardian.go
asincu: Yes, they are `https://tigera-packetcapture.tigera-packetcapture.svc.cluster.local` and `/certs/packetcapture/tls.crt`

PR #1407, opened 2021-07-27
Fix! Render secret in operator namespace

asincu:
## Description

Render `tigera-packetcapture-server-tls` in the operator namespace.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/packet_capture_api.go
+++ b/pkg/render/packet_capture_api.go
@@ -58,7 +58,7 @@ type packetCaptureApiComponent struct {
 	image              string
 	csrInitImage       string
 	keyValidatorConfig authentication.KeyValidatorConfig
-	serverCertSecret   *corev1.Secret
+	serverCertSecrets  []*corev1.Secret
 	clusterDomain      string
 }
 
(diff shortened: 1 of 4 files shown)


pkg/render/packet_capture_api.go
```diff
@@ -68,12 +68,16 @@ func PacketCaptureAPI(pullSecrets []*corev1.Secret, openshift bool,
 	serverCertSecret *corev1.Secret,
 	clusterDomain string) Component {
 
+	var serverCertSecrets []*corev1.Secret
+	serverCertSecrets = append(serverCertSecrets, serverCertSecret)
```
tmjd, line 72: I don't think we should be returning this secret in the objs to create. Because if the user had provided this secret the operator will now place an OwnerReference for itself on the secret.
I realize you maybe copied this from another controller because I think we incorrectly do that in some other controllers.

asincu: Ok, then when/how are the auto-generated secrets actually created  ?

tmjd: Just to summarize what was discussed elsewhere.
A secret (or probably any resource) passed in to a render should not return that same resource to be created.
The secret passed in should be created somehow by the Reconcile.

asincu: @tmjd https://github.com/tigera/operator/pull/1407/commits/a0456e4fea2267c06f7a3e591db2dc707fb39ff2

tmjd (approved):
LGTM

PR #1413, opened 2021-07-29
Update ES Gateway render code to avoid re-rendering a customer provid…

vberezny:
Currently if a customer provides a cert secret for es-gateway, the render code overwrites the OwnerReference on that secret. This PR adds a flag that can be used to check if the cert is provided by the customer, and if it is, then it will not be re-rendered in the Operator namespace (only copied to the Elasticsearch namespace).


Brian-McM (changes requested):
Instead of doing this could you please make sure to create the certificate outside of render, it should be created in the controller. You can use Alina's PR for an example of what I mean https://github.com/tigera/operator/commit/0e9b93585205856638f3d0e0be57acacf7f27814

Brian-McM (approved):

PR #1415, opened 2021-07-30
Add VPP support

AloysAugustin:
## Description

This PR adds support for VPP in the operator. We hope that it will enable the operator to become the primary way of managing Calico installations with the VPP dataplane enabled.
It should have no impact when VPP-specific configuration is not added. It changes the Installation object to add `VPP` as an option for `linuxDataplane`, and adds all the VPP-specific options in a `vpp` object under the `CalicoNetwork` object. 

It has been tested manually on VM, baremetal and EKS clusters, which are the 3 deployments currently officially supported by VPP.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/api/v1/installation_types.go
+++ b/api/v1/installation_types.go
@@ -302,6 +302,7 @@ type LinuxDataplaneOption string
 const (
 	LinuxDataplaneIptables LinuxDataplaneOption = "Iptables"
 	LinuxDataplaneBPF      LinuxDataplaneOption = "BPF"
+	LinuxDataplaneVPP      LinuxDataplaneOption = "VPP"
 )
 
 // CalicoNetworkSpec specifies configuration options for Calico provided pod networking.
@@ -311,7 +312,7 @@ type CalicoNetworkSpec struct {
 	// If not specified, iptables mode is used.
 	// Default: Iptables
 	// +optional
-	// +kubebuilder:validation:Enum=Iptables;BPF
+	// +kubebuilder:validation:Enum=Iptables;BPF;VPP
 	LinuxDataplane *LinuxDataplaneOption `json:"linuxDataplane,omitempty"`
 
 	// BGP configures whether or not to enable Calico's BGP capabilities.
--- a/pkg/controller/installation/validation.go
+++ b/pkg/controller/installation/validation.go
@@ -249,6 +249,9 @@ func validateCustomResource(instance *operatorv1.Installation) error {
 
 		// VPP specific validation
 		if instance.Spec.CalicoNetwork.LinuxDataplane != nil && *instance.Spec.CalicoNetwork.LinuxDataplane == operatorv1.LinuxDataplaneVPP {
+			if instance.Spec.Variant != operatorv1.Calico {
+				return fmt.Errorf("The VPP dataplane only supports the Calico variant (configured: %s)", instance.Spec.Variant)
+			}
 			if instance.Spec.CNI.Type != operatorv1.PluginCalico {
 				return fmt.Errorf("The VPP dataplane only supports the Calico CNI (configured: %s)", instance.Spec.CNI.Type)
 			}
--- a/pkg/controller/installation/validation.go
+++ b/pkg/controller/installation/validation.go
@@ -247,6 +247,16 @@ func validateCustomResource(instance *operatorv1.Installation) error {
 			}
 		}
 
+		// VPP specific validation
+		if instance.Spec.CalicoNetwork.LinuxDataplane != nil && *instance.Spec.CalicoNetwork.LinuxDataplane == operatorv1.LinuxDataplaneVPP {
+			if instance.Spec.CNI.Type != operatorv1.PluginCalico {
+				return fmt.Errorf("The VPP dataplane only supports the Calico CNI (configured: %s)", instance.Spec.CNI.Type)
+			}
+			if instance.Spec.CalicoNetwork.BGP == nil || *instance.Spec.CalicoNetwork.BGP == operatorv1.BGPDisabled {
+				return fmt.Errorf("VPP requires BGP to be enabled")
+			}
+		}
+
 		if bpfDataplane && instance.Spec.CalicoNetwork.NodeAddressAutodetectionV4 == nil {
 			return fmt.Errorf("spec.calicoNetwork.nodeAddressAutodetectionV4 is required for the BPF dataplane")
 		}
--- a/pkg/controller/installation/validation.go
+++ b/pkg/controller/installation/validation.go
@@ -258,6 +258,9 @@ func validateCustomResource(instance *operatorv1.Installation) error {
 			if instance.Spec.CalicoNetwork.BGP == nil || *instance.Spec.CalicoNetwork.BGP == operatorv1.BGPDisabled {
 				return fmt.Errorf("VPP requires BGP to be enabled")
 			}
+			if instance.Spec.CalicoNetwork.HostPorts != nil && *instance.Spec.CalicoNetwork.HostPorts == operatorv1.HostPortsDisabled {
+				return fmt.Errorf("VPP doesn't support disabling HostPorts")
+			}
 		}
 
 		if bpfDataplane && instance.Spec.CalicoNetwork.NodeAddressAutodetectionV4 == nil {
--- a/pkg/render/node.go
+++ b/pkg/render/node.go
@@ -1260,9 +1260,6 @@ func (c *nodeComponent) nodeEnvVars() []corev1.EnvVar {
 			nodeEnv = append(nodeEnv, corev1.EnvVar{
 				Name:  "FELIX_AWSSRCDSTCHECK",
 				Value: "Disable",
-			}, corev1.EnvVar{
-				Name:  "IP_AUTODETECTION_METHOD",
-				Value: "interface=eth0",
 			})
 		}
 	}
@@ -1301,6 +1298,7 @@ func (c *nodeComponent) nodeEnvVars() []corev1.EnvVar {
 	} else {
 		// BGP is enabled.
 		if c.vppDataplaneEnabled() {
+			// VPP comes with its own BGP daemon, so bird should be disabled
 			nodeEnv = append(nodeEnv, corev1.EnvVar{Name: "CALICO_NETWORKING_BACKEND", Value: "none"})
 		} else {
 			nodeEnv = append(nodeEnv, corev1.EnvVar{Name: "CALICO_NETWORKING_BACKEND", Value: "bird"})
--- a/pkg/render/node.go
+++ b/pkg/render/node.go
@@ -488,6 +488,15 @@ func (c *nodeComponent) nodeCNIConfigMap() *corev1.ConfigMap {
 		k8sAPIRoot = fmt.Sprintf("\n          \"k8s_api_root\":\"%s\",", apiRoot)
 	}
 
+	var externalDataplane string = ""
+	if c.vppDataplaneEnabled() {
+		externalDataplane = `,
+      "dataplane_options": {
+        "type": "grpc",
+        "socket": "unix:///var/run/calico/cni-server.sock"
+      }`
+	}
+
 	// Build the CNI configuration json.
 	var config = fmt.Sprintf(`{
   "name": "k8s-pod-network",
@@ -509,14 +518,14 @@ func (c *nodeComponent) nodeCNIConfigMap() *corev1.ConfigMap {
       },
       "kubernetes": {%s
           "kubeconfig": "__KUBECONFIG_FILEPATH__"
-      }
+      }%s
     },
     {
       "type": "bandwidth",
       "capabilities": {"bandwidth": true}
     }%s
   ]
-}`, mtu, nodenameFileOptional, ipam, ipForward, k8sAPIRoot, portmap)
+}`, mtu, nodenameFileOptional, ipam, ipForward, k8sAPIRoot, externalDataplane, portmap)
 
 	return &corev1.ConfigMap{
 		TypeMeta: metav1.TypeMeta{Kind: "ConfigMap", APIVersion: "v1"},
@@ -787,6 +796,13 @@ func (c *nodeComponent) nodeVolumes() []corev1.Volume {
 		)
 	}
 
+	if c.vppDataplaneEnabled() {
+		volumes = append(volumes,
+			// Volume that contains the felix dataplane binary
+			corev1.Volume{Name: "felix-plugins", VolumeSource: corev1.VolumeSource{HostPath: &corev1.HostPathVolumeSource{Path: "/var/lib/calico/felix-plugins"}}},
+		)
+	}
+
 	// If needed for this configuration, then include the CNI volumes.
 	if c.cfg.Installation.CNI.Type == operatorv1.PluginCalico {
 		// Determine directories to use for CNI artifacts based on the provider.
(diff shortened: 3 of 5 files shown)


pkg/render/node.go
```diff
@@ -1484,11 +1893,20 @@ func GetIPv6Pool(pools []operatorv1.IPPool) *operatorv1.IPPool {
...
+		*cr.CalicoNetwork.LinuxDataplane == operatorv1.LinuxDataplaneVPP &&
+		cr.CalicoNetwork.VPP != nil
+}
+
 // bgpEnabled returns true if the given Installation enables BGP, false otherwise.
 func bgpEnabled(instance *operatorv1.InstallationSpec) bool {
 	return instance.CalicoNetwork != nil &&
 		instance.CalicoNetwork.BGP != nil &&
-		*instance.CalicoNetwork.BGP == operatorv1.BGPEnabled
+		*instance.CalicoNetwork.BGP == operatorv1.BGPEnabled &&
```
caseydavenport, line 1908: I think this change is a bit confusing - instance.CalicoNetwork.BGP is meant to indicate whether or not BGP should be enabled - since VPP bundles its own BGP daemon, BGP is actually enabled, it's just not using BIRD.

Maybe it's as simple as renaming this function to `birdEnabled`? 

We probably also want to update validation to ensure that if VPP is enabled, instance.CalicoNetwork.BGP is also enabled - unless VPP data plane also supports running without BGP?

pkg/render/node.go
```diff
@@ -1484,11 +1893,20 @@ func GetIPv6Pool(pools []operatorv1.IPPool) *operatorv1.IPPool {
 	return nil
 }
 
+// vppDataplaneEnabled returns true if the given Installation enables VPP
+func vppDataplaneEnabled(cr *operatorv1.InstallationSpec) bool {
+	return cr.CalicoNetwork != nil &&
+		cr.CalicoNetwork.LinuxDataplane != nil &&
+		*cr.CalicoNetwork.LinuxDataplane == operatorv1.LinuxDataplaneVPP &&
```
caseydavenport, line 1900: It's a bit of a red flag to me that we're relying on two API configs to determine if vpp is enabled. the LinuxDataplane option should be the deciding factor as to whether or not the VPP dataplane is being used.

The cr.CalicoNetwork.VPP should probably be defaulted based on the LinuxDataplane option, and validation should be used to make sure that the two are consistent with each other before ever reaching this function.

caseydavenport: In other words - cr.CalicoNetwork.VPP should never be nil at this point in the code if LinuxDataplane == VPP

pkg/render/node.go
```diff
@@ -706,6 +798,46 @@ func (c *nodeComponent) nodeDaemonset(cniCfgMap *corev1.ConfigMap) *appsv1.Daemo
 	return &ds
 }
 
+// vppDaemonset creates the vpp daemonset.
+func (c *nodeComponent) vppDaemonset() *appsv1.DaemonSet {
```
caseydavenport, line 802: I think there is probably a case to made for this DaemonSet and supporting functions being in its own rendered - it's not really part of calico/node. @tmjd WDYT?

pkg/render/node.go
```diff
@@ -255,137 +289,180 @@ func (c *nodeComponent) nodeRoleBinding() *rbacv1.ClusterRoleBinding {
...
-				Verbs: []string{"create", "update"},
+		},
+	}
+	return crb
+}
+
+// nodeRole creates the clusterrole containing policy rules that allow the node daemonset to operate normally.
+func (c *nodeComponent) nodeRole(forVpp bool) *rbacv1.ClusterRole {
+	var role *rbacv1.ClusterRole
+	if !forVpp {
```
caseydavenport, line 319: I'm not a fan of this approach for re-using the ClusterRole struct between two different cluster role. I think we should just have two separate cluster roles.

pkg/render/node.go
```diff
@@ -143,6 +147,20 @@ func (c *nodeComponent) ResolveImages(is *operatorv1.ImageSet) error {
 		errMsgs = append(errMsgs, err.Error())
 	}
 
+	if vppDataplaneEnabled(c.cfg.Installation) {
+		c.vppImage, err = components.GetReference(components.ComponentCalicoVpp, reg, path, prefix, is)
+	}
+	if err != nil {
```
caseydavenport, line 153: This error check should be within the `vppDataplaneEnabled` branch above where the error is generated

pkg/render/node.go
```diff
@@ -143,6 +147,20 @@ func (c *nodeComponent) ResolveImages(is *operatorv1.ImageSet) error {
...
+		errMsgs = append(errMsgs, err.Error())
+	}
+
+	if vppDataplaneEnabled(c.cfg.Installation) {
+		c.vppAgentImage, err = components.GetReference(components.ComponentCalicoVppAgent, reg, path, prefix, is)
+	}
+	if err != nil {
```
caseydavenport, line 160: same for this guy

api/v1/installation_types.go
```diff
@@ -345,6 +350,76 @@ type CalicoNetworkSpec struct {
...
+
+	// TunQueueSize can be used to configure the queue size of the uplink interface
+	// Default: 1024
+	// +optional
+	TunQueueSize int `json:"tunQueueSize,omitempty"`
+
+	// IPsec can be used to enable IPsec encryption of IPIP tunnels
+	// Accepted values: "Disabled" (default), "Enabled"
+	// +optional
+	IPsec string `json:"ipsec,omitempty"`
```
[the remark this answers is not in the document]
caseydavenport: Similar to SerivcePrefix, wondering if this should ultimately be outside of this struct and enforced using validation. In theory, other dataplanes could add IPSec support in the future.

caseydavenport (changes requested):
@AloysAugustin really sorry for the delay on this! To expedite, I think we should do what we talked about on the call and do this in two parts. I think it will also help clarify some of the logic for me as well.

1. Add the new LinuxDataplaneVPP option, which configures calico/node and CNI to use the external dataplane. The VPP daemonset itself can be deployed out-of-band.
2. Add the VPPConfiguration fields in a second PR, allowing deployment of VPP itself via the operator.

Hope that's not too much of a burden. 

I did review the VPPConfiguration section - I am still chewing on some of the details. Most of my comments were around aligning with [Kubernetes API conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md). 

The other area I'm thinking about is validation - I don't see much of it in this PR as it stands, but we definitely will need some - especially doing cross-field validation that the entire Installation is consistent. The pattern we follow in the operator is that the render code shouldn't have to make decisions for the user - the defaulting and validation code should ensure that an actionable, self-consistent configuration is passed to render. 

As an example, if VPP is enabled but BGP is disabled, what is the user asking for? It's not clear, so we should validate against it.

api/v1/installation_types.go
```diff
@@ -345,6 +350,76 @@ type CalicoNetworkSpec struct {
...
+
+	// Hugepages can be set to "Enabled" if hugepages are available on all nodes
+	// to allow more efficient drivers to be used
+	// +optional
+	Hugepages string `json:"hugepages,omitempty"`
+
+	// DefaultGateway can be set to configure a default route in VPP.
+	// Accepted formats are "ipv4", "ipv6" or "ipv4,ipv6"
+	// +optional
+	DefaultGateway string `json:"defaultGateway,omitempty"`
```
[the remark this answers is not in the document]
AloysAugustin: We actually expect one or two IP addresses, looks like the comment is misleading :)

api/v1/installation_types.go
```diff
@@ -345,6 +350,76 @@ type CalicoNetworkSpec struct {
...
+	// Workers can be set to configure the number of VPP workers. With 0 workers, VPP will run with
+	// a single main thread processing API requests and network traffic. With 1+ workers, VPP will
+	// run with one main thread processing API requests, and the configured number of workers
+	// processing container traffic. Default: 0
+	// +optional
+	Workers int `json:"workers,omitempty"`
+
+	// HookType is set to specify node distribution, only "systemctl" is supported for now
+	// +optional
+	HookType string `json:"hookType,omitempty"`
```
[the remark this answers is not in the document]
AloysAugustin: It can also be empty - but like a few other fields, we should make it a proper enum, and default to something like "None"

AloysAugustin:
Thanks @caseydavenport for the review! I simplified this PR to include only the changes to calico-node as discussed. We'll open a new PR for the VPP specific bits, while taking your comments into account.

I've already added some level of config validation in this PR to ensure that the config is consistent (if VPP is enabled, the CNI should be calico and BGP should be enabled), and a few tests as well.

pkg/render/node.go
```diff
@@ -1220,6 +1245,27 @@ func (c *nodeComponent) nodeEnvVars() []corev1.EnvVar {
...
+		}, corev1.EnvVar{
+			Name:  "FELIX_XDPENABLED",
+			Value: "false",
+		})
+		if c.cfg.Installation.KubernetesProvider == operatorv1.ProviderEKS {
+			nodeEnv = append(nodeEnv, corev1.EnvVar{
+				Name:  "FELIX_AWSSRCDSTCHECK",
+				Value: "Disable",
+			}, corev1.EnvVar{
+				Name:  "IP_AUTODETECTION_METHOD",
```
caseydavenport, line 1264: IP_AUTODETECTION_METHOD is configurable via the Installation API, so we can't ignore whatever the user has specified there. If VPP requires this, then we should set this in defaulting and validation rather than deep in the render code so that the Installation API is consistent with the rendered objects.

pkg/render/node.go
```diff
@@ -1220,6 +1245,27 @@ func (c *nodeComponent) nodeEnvVars() []corev1.EnvVar {
...
+		}, corev1.EnvVar{
+			Name:  "FELIX_DATAPLANEDRIVER",
+			Value: "/usr/local/bin/felix-plugins/felix-api-proxy",
+		}, corev1.EnvVar{
+			Name:  "FELIX_XDPENABLED",
+			Value: "false",
+		})
+		if c.cfg.Installation.KubernetesProvider == operatorv1.ProviderEKS {
+			nodeEnv = append(nodeEnv, corev1.EnvVar{
+				Name:  "FELIX_AWSSRCDSTCHECK",
```
caseydavenport, line 1261: Hm, I was expecting us to already be setting this but it doesn't look like it. @tmjd do you recall whether we discussed this in the past?

pkg/render/node.go
```diff
@@ -1254,7 +1300,11 @@ func (c *nodeComponent) nodeEnvVars() []corev1.EnvVar {
 		}
 	} else {
 		// BGP is enabled.
-		nodeEnv = append(nodeEnv, corev1.EnvVar{Name: "CALICO_NETWORKING_BACKEND", Value: "bird"})
+		if c.vppDataplaneEnabled() {
+			nodeEnv = append(nodeEnv, corev1.EnvVar{Name: "CALICO_NETWORKING_BACKEND", Value: "none"})
```
caseydavenport, line 1304: Probably worth a comment here saying that BGP is enabled, but that the VPP data plane has its own BGP implementation and so we disabled BIRD anyway. Otherwise this code might be not-obvious for future readers.

pkg/controller/installation/validation.go
```diff
@@ -247,6 +247,16 @@ func validateCustomResource(instance *operatorv1.Installation) error {
 			}
 		}
 
+		// VPP specific validation
+		if instance.Spec.CalicoNetwork.LinuxDataplane != nil && *instance.Spec.CalicoNetwork.LinuxDataplane == operatorv1.LinuxDataplaneVPP {
```
caseydavenport, line 251: We should probably also verify that the `spec.Variant` field is set to `Calico` since we don't support this for Calico Enterprise just yet. 

@tmjd any other validations you think we should be checking here?

caseydavenport:
@AloysAugustin thanks for reworking this guy - it's looking very close! I just added a couple thoughts.

pkg/render/node.go
tmjd: I don't remember talking about that one before.
I'd be concerned to default it because of this from the [reference docs](https://projectcalico.docs.tigera.io/reference/felix/configuration) 
> Check IAM role and profile configuration for setting the necessary permission for this setting to work. [Default: DoNothing]

pkg/controller/installation/validation.go
tmjd: Are there any IPv4 and IPv6 limitations?
Any encapsulation limitations?
Are there any HostPorts or ContainerIPForwarding limitations?
I don't really know what limitations there would be, I was just looking at the fields in the Installation resource.

AloysAugustin: Thanks, good questions :) 
IPv4 / IPv6 / dual stack is supported.
In terms of encap, we support flat IP / IPIP / VXLAN and wireguard so I think it should be alright.
HostPorts have been added recently, and ContainerIPForwarding should be supported but I don't think we've tested it, I'll make a note to check it.

pkg/render/node.go
AloysAugustin: For this one, we've had a lot fewer issues by enabling it on EKS, and EKS nodes have the required IAM policy by default (as stated on [this page](https://projectcalico.docs.tigera.io/reference/resources/felixconfig#aws-iam-rolepolicy-for-source-destination-check-configuration)). So it would be nice to have a way to enable it. Would you prefer to expose this variable in the Installation spec and default it to `Disable` when using EKS + VPP?

AloysAugustin:
Thanks @caseydavenport  and @tmjd for the review, most comments should be addressed by the latest commit, looks like the last sticking point is the FELIX_AWSSRCDSTCHECK discussion.

pkg/controller/installation/validation.go
caseydavenport: > HostPorts have been added recently,

Perhaps a nitty point, but does VPP support _disabling_ hostports? e.g., if `hostPorts: Disabled` is set, will VPP respect that? If not, we need to validate that hostPorts is never set to Disabled while in VPP mode.

pkg/render/node.go
caseydavenport: I think it's OK to leave this as it is here - if we ever do add a config option for it, we can default that option based on existing behavior.

The only consideration would be if we want to set this via `FelixConfiguration` datastore API instead of an env variable - I believe that's the path we want to be going but I can't recall if we have the right render infrastructure set up to make that easy to do.

tmjd: Sounds good to me to leave this.
As discussed in the Calico Community meeting we don't have the proper infrastructure for the FelixConfiguration to allow/understand if the users overwrote/set the field.

pkg/controller/installation/validation.go
caseydavenport: @AloysAugustin just one more question above, otherwise I think this is good to go

AloysAugustin: We don't support disabling `HostPorts`, so I added a validation check for that.

caseydavenport:
@tmjd CI looks to be failing with a CRD issue - does this just need rebase to fix?

tmjd:
Yes, I merged a PR that should fix that earlier in the day.

AloysAugustin:
Thanks, I just rebased the branch.

AloysAugustin:
Looks like the CI failed again but I can't see the logs, any chance someone could take a look?

caseydavenport:
@AloysAugustin was just a test flake - all good now! :rocket:

PR #1447, opened 2021-08-10
Default FelixConfiguration RouteTableRange when using AWS CNI

tmjd:
## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/apis/crd.projectcalico.org/v1/felixconfig.go
+++ b/pkg/apis/crd.projectcalico.org/v1/felixconfig.go
@@ -213,7 +213,7 @@ type FelixConfigurationSpec struct {
 	// set to false. This reduces the number of metrics reported, reducing Prometheus load. [Default: true]
 	PrometheusProcessMetricsEnabled *bool `json:"prometheusProcessMetricsEnabled,omitempty"`
 	// PrometheusReporterPort specifies the TCP port on which to report denied packet metrics.
-	PrometheusReporterPort *int `json:"prometheusReporterPort"`
+	PrometheusReporterPort *int `json:"prometheusReporterPort,omitempty"`
 
 	// FailsafeInboundHostPorts is a comma-delimited list of UDP/TCP ports that Felix will allow incoming traffic to host endpoints
 	// on irrespective of the security policy. This is useful to avoid accidentally cutting off a host with incorrect configuration. Each
--- a/pkg/controller/installation/core_controller.go
+++ b/pkg/controller/installation/core_controller.go
@@ -215,6 +215,12 @@ func add(mgr manager.Manager, r *ReconcileInstallation) error {
 		return fmt.Errorf("tigera-installation-controller failed to watch KubeControllersConfiguration resource: %w", err)
 	}
 
+	// Watch for changes to FelixConfiguration.
+	err = c.Watch(&source.Kind{Type: &crdv1.FelixConfiguration{}}, &handler.EnqueueRequestForObject{})
+	if err != nil {
+		return fmt.Errorf("tigera-installation-controller failed to watch FelixConfiguration resource: %w", err)
+	}
+
 	if r.enterpriseCRDsExist {
 		// Watch for changes to primary resource ManagementCluster
 		err = c.Watch(&source.Kind{Type: &operator.ManagementCluster{}}, &handler.EnqueueRequestForObject{})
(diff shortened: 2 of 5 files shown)


tmjd:
This will also need a docs PR to add `create` permissions to `felixconfigurations`. Operator already has patch, list, get, and watch.

petercork (commented):
LGTM (spending some time testing too)

pkg/apis/crd.projectcalico.org/v1/felixconfig.go
```diff
@@ -213,7 +213,7 @@ type FelixConfigurationSpec struct {
 	// set to false. This reduces the number of metrics reported, reducing Prometheus load. [Default: true]
 	PrometheusProcessMetricsEnabled *bool `json:"prometheusProcessMetricsEnabled,omitempty"`
 	// PrometheusReporterPort specifies the TCP port on which to report denied packet metrics.
-	PrometheusReporterPort *int `json:"prometheusReporterPort"`
+	PrometheusReporterPort *int `json:"prometheusReporterPort,omitempty"`
```
tmjd, line 216: Without this the operator was getting errors when trying to create the FelixConfiguration with no PrometheusReporterPort value.

tmjd: This also seems to match what is latest in [tigera/api](https://github.com/tigera/api/blob/master/pkg/apis/projectcalico/v3/felixconfig.go#L311)

mikestephen:
The operator sets FELIX_ROUTETABLERANGE for GKE to 10-250 here, I wonder will GKE need the same treatment that AWS CNI gets in this PR?

https://github.com/tigera/operator/blob/5cfdc2734a474fce64b284240dba9894c8f2433c/pkg/render/node.go#L1286

pkg/controller/installation/core_controller.go
```diff
@@ -1354,6 +1359,37 @@ func (r *ReconcileInstallation) validateTyphaCAConfigMap() (*corev1.ConfigMap, e
...
+// otherwise a patch will be performed.
+func (r *ReconcileInstallation) setDefaultsOnFelixConfiguration(ctx context.Context, install *operator.Installation, fc *crdv1.FelixConfiguration, create bool, log logr.Logger) error {
+	// If we're using the AWS CNI plugin we need to ensure the route tables that calico-node
+	// uses do not conflict with the ones the AWS CNI plugin uses so default them
+	// in the FelixConfiguration if they are not already set.
+	if install.Spec.CNI.Type == operator.PluginAmazonVPC {
+		if fc.Spec.RouteTableRange == nil {
+			patchFrom := client.MergeFrom(fc.DeepCopy())
+			fc.Spec.RouteTableRange = &crdv1.RouteTableRange{
+				Min: 31,
```
robbrockbank, line 1373: Is this range correct - curious to understand where these numbers are derived from.

Depending on which version of the VPC CNI plugin I'm looking at, I see table indexes being assigned based on:
- The ENI device number + 1   (and I think we can have up to 4x15 devices for p4d.24xlarge)
- The VLAN table ID + 100

robbrockbank: ^^^ it's also possible that my code searching in the plugin was flawed... so I could be totally out there!

robbrockbank: Also, worth including any details in the comment here about this range.

mikestephen (approved):
LGTM!

pkg/controller/installation/core_controller.go
```diff
@@ -1002,19 +1002,24 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 		}
 	}
 
+	// Fetch any existing default FelixConfiguration object.
+	felixConfiguration := &crdv1.FelixConfiguration{}
+	fcErr := r.client.Get(ctx, types.NamespacedName{Name: "default"}, felixConfiguration)
```
caseydavenport, line 1007: nit, but can we just use `err` here? It's confusing to have multiply named error vars when logically they aren't needed (unless for some reason it is needed here?)

caseydavenport: Just a reminder we'll need to update the OSS RBAC since I don't believe it has felixconfig Get in it right now.

pkg/controller/installation/core_controller.go
```diff
@@ -1354,6 +1359,59 @@ func (r *ReconcileInstallation) validateTyphaCAConfigMap() (*corev1.ConfigMap, e
...
+			// - The VLAN table ID + 100 (there is doubt if this is true)
+			fc.Spec.RouteTableRange = &crdv1.RouteTableRange{
+				Min: 65,
+				Max: 99,
+			}
+		}
+	case operator.PluginGKE:
+		if fc.Spec.RouteTableRange == nil {
+			updated = true
+			// Don't conflict with the GKE CNI plugin's routes.
```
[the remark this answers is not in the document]
caseydavenport: Ah I see, this was just replacing what had previously been done via an env var for compatibility.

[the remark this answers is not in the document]
tmjd: I don't know. I was just copying what I moved from the previous value that was being added as an Env Var.

caseydavenport (approved):
We should probably move the felixconfiguraiton watch outside of the EE only branch.

Don't need to re-review though.

PR #1448, opened 2021-08-10
Refactor Calico renderer

caseydavenport:
## Description

This is probably not the ideal end state, but I think it's better than
what we have today.

Main goal is to remove some of the extra layering in the Calico
rendering logic, and to make ever so slightly a more clean split between responsibilities
of the controller and those of the renderer.

Should make it easier to add / remove arguments to the render code as
well without needing to change a ton of different places.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


pkg/controller/installation/core_controller.go
```diff
@@ -931,12 +935,37 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 
 	var typhaNodeTLS *render.TyphaNodeTLS
 	if instance.Spec.CertificateManagement == nil {
+		// First, attempt to load TLS secrets from the cluster, if any exist.
```
caseydavenport, line 938: I need to revisit this since I don't think I fully understand how it was working before, but it seemed like we had a strange split of TLS secret generation between the controller and the render package.

The approach I'm trying to move towards would to have the controller be responsible for generating the secrets when needed, and caching them in its own namespace. It can pass them as inputs to the render package, which can then modify / move them to the correct destination namespace and return them for installation into the cluster.

pkg/controller/utils/utils.go
```diff
@@ -51,6 +51,7 @@ const (
 )
 
 var DefaultInstanceKey = client.ObjectKey{Name: "default"}
+var CalicoEnterpriseInstsanceKey = client.ObjectKey{Name: "calico-enterprise"}
```
caseydavenport, line 54: Whoops, this shouldn't be here.

pkg/render/apiserver.go
```diff
@@ -249,7 +249,7 @@ func (c *apiServerComponent) Objects() ([]client.Object, []client.Object) {
 		globalObjects = append(globalObjects, c.apiServiceRegistration(c.tlsSecrets[0].Data[APIServerSecretCertName]))
 	} else {
 		namespacedObjects = append(namespacedObjects, c.apiServiceRegistration(c.installation.CertificateManagement.CACert))
-		globalObjects = append(globalObjects, CsrClusterRoleBinding(csrRolebindingName(c.installation.Variant), rmeta.APIServerNamespace(c.installation.Variant)))
+		globalObjects = append(globalObjects, CSRClusterRoleBinding(csrRolebindingName(c.installation.Variant), rmeta.APIServerNamespace(c.installation.Variant)))
```
caseydavenport, line 252: Just a nit cleanup to fix capitalization.

pkg/render/render_test.go
```diff
@@ -38,6 +38,41 @@ import (
 	rmeta "github.com/tigera/operator/pkg/render/common/meta"
 )
 
+// allCalicoComponents takes the given configuration and returns all the components
+// associated with installing Calico's core, similar to how the core_controller behaves.
+func allCalicoComponents(
```
caseydavenport, line 43: This is a dummy test function that mimics what we used to do for `render.Calico` - I think we can probably just remove this test code after these changes, but wanted to keep it around for testing purposes.

tmjd: Another option would be to move these to be tests in core_controller_test and have the tests be more end-to-end (because it is also through the reconcile). Just a thought.

tmjd (commented):
I'm not a huge fan of structs to pass args but this definitely will make testing better.

I kind of liked the render/render 'wrapper' since it bundled the whole calico component but do like cutting out that middle man.

I'm definitely a fan of what you suggested for the Secrets.

caseydavenport:
> I'm not a huge fan of structs to pass args but this definitely will make testing better.

It will make adding new config substantially easier as well - adding a new field to pass to the Node rendered required touching a ton of different places (the intermediate function, the intermediate struct, the node function, the node struct, the controller).

A struct will cut out a ton of those.

pkg/render/node.go
```diff
@@ -61,61 +65,64 @@ var (
...
-	clusterDomain           string
-	nodeImage               string
-	cniImage                string
-	flexvolImage            string
-	certSignReqImage        string
-	nodeReporterMetricsPort int
-	bgpLayoutHash           string
-	logCollector            *operator.LogCollector
+	// Input configuration from the controller.
+	cfg *NodeConfiguration
```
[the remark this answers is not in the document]
caseydavenport: we probably could do something like that, but not sure what it would buy us?

pkg/controller/installation/core_controller.go
```diff
@@ -1097,7 +1093,51 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
...
+		LogCollector:            logCollector,
+		BirdTemplates:           birdTemplates,
+		TLS:                     typhaNodeTLS,
+		ClusterDomain:           r.clusterDomain,
+		NodeReporterMetricsPort: nodeReporterMetricsPort,
+		BGPLayouts:              bgpLayout,
+		NodeAppArmorProfile:     nodeAppArmorProfile,
+		MigrateNamespaces:       needNsMigration,
+	}
+	components = append(components, render.Node(&nodeCfg))
```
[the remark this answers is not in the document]
caseydavenport: I don't think it matters, but happy to put Typha first anyway.

pkg/controller/installation/core_controller.go
Brian-McM: Agreed, I think this is where we decided to go after that operator meeting like 2 months ago.

pkg/controller/installation/core_controller.go
```diff
@@ -1048,43 +1077,10 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
...
-	if err != nil {
-		log.Error(err, "Error with rendering Calico")
-		r.SetDegraded("Error with rendering Calico resources", err, reqLogger)
-		return reconcile.Result{}, err
-	}
-
+	// Render namespaces for Calico.
 	components := []render.Component{}
+	components = append(components, render.Namespaces(&instance.Spec, pullSecrets))
+
```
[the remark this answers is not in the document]
tmjd: I think that also is what is the cause of the failing tests.

pkg/render/render_test.go
caseydavenport: These tests are super horrible and brittle, I am finding out. They have a lot of assertions of the form "expected 30 to be 29" without any indication of which resource was extra / missing.

I think they probably need a heavy re-write, but for now I am trying to keep them in place and get them passing.

pkg/render/render_test.go
```diff
@@ -267,6 +396,10 @@ func componentCount(components []render.Component) int {
 	for _, c := range components {
 		objsToCreate, _ := c.Objects()
 		count += len(objsToCreate)
+		glog.Printf("Component: %s\n", reflect.TypeOf(c))
```
caseydavenport, line 399: Makes it easier to debug these tests when they fail - prints out each component and the resources emitted.

caseydavenport:
@tmjd @Brian-McM I think this one is good to go now - got all the tests passing, just a few minimal adjustments to the tests themselves to pass more realistic data and adjust expectation ordering now that some things are created in a different order - e.g., rather than all secrets and configmaps coming first, the ones in `tigera-operator` come first and the ones in `calico-system` now get created with the components that require them.

tmjd (commented):
One thing, not a big deal if you don't agree.
I think the current implementation has the correct behavior.

pkg/render/kube-controllers_test.go
```diff
@@ -167,9 +183,14 @@ var _ = Describe("kube-controllers rendering tests", func() {
 		}
 
 		instance.Variant = operator.TigeraSecureEnterprise
-
-		component := render.KubeControllers(k8sServiceEp, instance, true, nil, nil, &internalManagerTLSSecret, &elasticsearchSecret, &kibanaSecret, nil,
-			true, dns.DefaultClusterDomain, &kubeControllersUserSecret, 9094)
+		cfg.LogStorageExists = true
```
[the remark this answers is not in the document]
caseydavenport: lol yeah, it was a real pain trying to go through each of these tests and determine which config options were being set and which weren't.

pkg/controller/installation/core_controller.go
```diff
@@ -931,12 +935,37 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
...
+		// First, attempt to load TLS secrets from the cluster, if any exist.
+		typhaNodeTLS, err = r.GetTyphaNodeTLSConfig()
 		if err != nil {
 			log.Error(err, "Error with Typha/Felix secrets")
 			r.SetDegraded("Error with Typha/Felix secrets", err, reqLogger)
 			return reconcile.Result{}, err
 		}
+
+		if typhaNodeTLS.CAConfigMap == nil || typhaNodeTLS.TyphaSecret == nil || typhaNodeTLS.NodeSecret == nil {
+			// Unable to find at least one necessary bit of TLS config. Generate new ones ourselves.
```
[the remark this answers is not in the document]
caseydavenport: I think I agree with you, but I'd also rather save this for another rainy day if possible. 

I don't really like the way the existing code combines fetching the existing resources with validation, but it's a bit much for me to unpick right now!

> Maybe the validation should be pulled out of the GetTyphaNodeTLSConfig and the new function could return if the TLS info needs to be created and an err on the validation result?

:100: - or something like this. Split out the querying from the validation.

Brian-McM:
This looks pretty good to me but @tmjd has been doing the fine grained review so I'll leave it for his approval. 

The only comment I had, which is just a comment and not something that we should do anything about (since you already have this amazing work completed), is a limitation of the config struct solution. We still have required variables, and in general I think it's nice when required variables are part of the function definition (it acts a a validation that it's set). The major problem was with "optional" arguments, of which there were many. This is obviously 1000 times better then what we had, and this is a very very minor thing that's not worth addressing since you have this.

caseydavenport:
> We still have required variables, and in general I think it's nice when required variables are part of the function definition (it acts a a validation that it's set). The major problem was with "optional" arguments, of which there were many

Yeah, I definitely spotted this problem myself when implementing this. Something I'll think about building on in the future!

tmjd (approved):
LGTM

PR #1453, opened 2021-08-12
Fix! Enable authentication for packet capture

asincu:
## Description

- For a multicluster setup, tigera-guardian service account needs to be
allowed to create authentication reviews inside the managed cluster
- For an OIDC setup, api controller needs to be notified when
authentication CR is created
- Reduce initial delay time for readiness probe

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/guardian.go
+++ b/pkg/render/guardian.go
@@ -167,6 +167,10 @@ func (c *GuardianComponent) clusterRole() client.Object {
 				Verbs:     []string{"impersonate"},
 			},
 			{
+				// Guardian forwards its token when sending the request to other services
+				// PacketCapture will authenticate the token from the request and check if impersonation is allowed
+				// AuthenticationReview API can authenticate a bearer token from the request if the token can create
+				// authenticationreviews
 				APIGroups: []string{"projectcalico.org"},
 				Resources: []string{"authenticationreviews"},
 				Verbs:     []string{"create"},
--- a/pkg/render/guardian.go
+++ b/pkg/render/guardian.go
@@ -166,6 +166,11 @@ func (c *GuardianComponent) clusterRole() client.Object {
 				Resources: []string{"users", "groups", "serviceaccounts"},
 				Verbs:     []string{"impersonate"},
 			},
+			{
+				APIGroups: []string{"projectcalico.org"},
+				Resources: []string{"authenticationreviews"},
+				Verbs:     []string{"create"},
+			},
 		},
 	}
 }
--- a/pkg/controller/apiserver/apiserver_controller.go
+++ b/pkg/controller/apiserver/apiserver_controller.go
@@ -115,6 +115,12 @@ func add(mgr manager.Manager, r *ReconcileAPIServer) error {
 				return fmt.Errorf("apiserver-controller failed to watch the Secret resource: %v", err)
 			}
 		}
+
+		// Watch for changes to authentication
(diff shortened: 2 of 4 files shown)


pkg/render/guardian.go
```diff
@@ -166,6 +166,11 @@ func (c *GuardianComponent) clusterRole() client.Object {
 				Resources: []string{"users", "groups", "serviceaccounts"},
 				Verbs:     []string{"impersonate"},
 			},
+			{
+				APIGroups: []string{"projectcalico.org"},
```
rene-dekker, line 170: I need to refresh my memory a little bit here...

For MCM a user makes a request in management cluster for packetcapture in **management** cluster:
- Voltron routes to PacketCapture pod
- PacketCapture pod does authentication review for the bearer token to extract authenticated user
- Then uses authenticated user for RBAC purposes
- Then does whatever PacketCapture does.

For MCM a user makes a request in management cluster for packetcapture in **managed** cluster
- Voltron sees x-cluster-id header and realizes it has to be proxied to managed cluster.
- It does auth check and extracts the user
- It then sends the request through the tunnel and on the other end, guardian will impersonate this user.
- Guardian will send the request to PacketCapture pod
- PacketCapture pod does authentication review for the bearer token to extract authenticated user, which is guardian (because the authenticationreviews api is not smart enough to extract impersonation headers. Please correct me if I am wrong)
- Then PacketCapture still extracts the original user and does RBAC checks for that original user.

Is this summary right? If so, should we maybe add a ticket to improve on the flow in a later occasion? Also, we probably want to comment/document why we need this RBAC rule.

asincu:
```
PacketCapture pod does authentication review for the bearer token to extract authenticated user, which is guardian (because the authenticationreviews api is not smart enough to extract impersonation headers. Please correct me if I am wrong)
```

Yes, that is correct. I would say that the ticket needs to be in lma because it does authenticationreviews using the bearer token and checks if that token can authenticate itself.

rene-dekker: So the reason why guardian needs authentication reviews permission now and never because is because PacketCapture is the first server to live inside the managed cluster. ES proxy, compliance, etc all live inside the management cluster and so they are dealing with the original bearer tokens and not an impersonator. 

The result is that authentication review inside the managed cluster is authenticating the service account and not the original user. 

Given our architecture, we should check incoming requests for whether they are impersonated. If they are, we should check if the impersonator is authenticated and allowed to impersonate. And for the impersonated, we need to check if it has access for the packetcapture requests. Inside the managed cluster, no OIDCAuthenticator is needed with this flow.

AuthenticationReviews is one way of checking if the impersonator is authenticated. The other is TokenReviews. TokenReviews is probably a better api for it, but we can leave the code as-is. So I retract what I say about improving the flow. On second glance I don't think that is necessary.

rene-dekker: Still, I would recommend capturing some of this discussion as a comment in the ticket.

rene-dekker (approved):
lgtm

pkg/render/guardian.go
asincu: Yes, we already have the logic implemented to check if the service account can actually impersonate. I think we moved away from TokenReviews because of missing featured for OIDC for managed provisioners (GKE etc). The changes should be our internal library for authentication / review the authenticationreviews API.

tmjd:
LGTM

PR #1459, opened 2021-08-17
Only set kubernetes server/port env vars on host-networked pods.

fasaxc:
## Description

In MKE clusters, the API server is only reachable via the MKE proxy from the host namespace because it is implemented as a DNS entry in /etc/hosts on the host.

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/node.go
+++ b/pkg/render/node.go
@@ -944,7 +944,7 @@ func (c *nodeComponent) cniEnvvars() []v1.EnvVar {
 		},
 	}
 
-	envVars = append(envVars, c.cfg.K8sServiceEp.EnvVars()...)
+	envVars = append(envVars, c.cfg.K8sServiceEp.EnvVars(true, c.cfg.Installation.KubernetesProvider)...)
 
 	if c.cfg.Installation.Variant == operator.TigeraSecureEnterprise {
 		if c.cfg.Installation.CalicoNetwork != nil && c.cfg.Installation.CalicoNetwork.MultiInterfaceMode != nil {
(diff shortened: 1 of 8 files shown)


tmjd:
This change seems to be very broad and will affect more than MKE clusters.
Specifically, I think there is a "minimal" setup that does not use Kubernetes services that this change would break since it is no longer setting the K8s env vars on several components.

Should this behavior be a special case when the provider is "DockerEnterprise" (which I think turned into MKE).

fasaxc:
I was wary of making it specific to one environment since that makes the environments more different.  However, the "minimal" setup changes the calculus.  I'll condition on DockerEE.

fasaxc:
@tmjd I've made it conditional on DockerEE and the actual value that is a special case.  I pushed the logic down into the shared code, thought that made it cleaner.

tmjd (approved):
I like that this also checks for the special `proxy.local`.

pkg/render/node.go
```diff
@@ -944,7 +944,7 @@ func (c *nodeComponent) cniEnvvars() []v1.EnvVar {
 		},
 	}
 
-	envVars = append(envVars, c.cfg.K8sServiceEp.EnvVars()...)
+	envVars = append(envVars, c.cfg.K8sServiceEp.EnvVars(true, c.cr.KubernetesProvider)...)
```
tmjd, line 947: I think you need to rebase on master. I think this will need to change to c.cfg.Installation.KubernetesProvider.

marvin-tigera:

PR #1466, opened 2021-08-18
Refactor esgateway component and logstorage controller to enable easi…

vberezny:


pkg/controller/logstorage/logstorage.go
```diff
@@ -0,0 +1,288 @@
...
+	operatorv1 "github.com/tigera/operator/api/v1"
+	"github.com/tigera/operator/pkg/controller/utils"
+	"github.com/tigera/operator/pkg/controller/utils/imageset"
+	"github.com/tigera/operator/pkg/render"
+	relasticsearch "github.com/tigera/operator/pkg/render/common/elasticsearch"
+	rmeta "github.com/tigera/operator/pkg/render/common/meta"
+	rsecret "github.com/tigera/operator/pkg/render/common/secret"
+)
+
+func (r *ReconcileLogStorage) createLogStorage(
```
Brian-McM, line 32: I think all of these createXXX functions need to return a Reconcile object and an error, and return that to the caller of the controller. I say this because there are 3 types of errors:
1. A real error happened that we weren't expecting, and we expect an immediate retry to fix it (hopefully)
2. A temporary error, like something not being ready, and we want to retry in a specific amount of time
3. A temporary error, like a k8s resource not being ready, and we want to wait for a notification from the k8s API telling us that the state of a resource has changed

In order, the logging / what should be returned to the caller of the controller is as follows:
1. Log the error, set degraded, return an empty reconcile and an error
2. Log the error, set degraded, return a reconcile with `Requeue` or `RequeueAfter` set and a nil error
3. Log the error, set degraded, return an empty Reconcile and a nil error (we'll re reconcile when k8s notifies us of a change)

I commented on one the tests regarded 3, it's important we don't return an error otherwise we end up in a tight reconcile loop when we know there's nothing we can do until k8s notifies us about a change.

Brian-McM: One other thing, I think the createXXX functions need to either return a 3rd parameter saying we should continue, or return a special error that we can switch on. The caller wouldn't be able to tell if Reconcile is empty and the error is nil because everything is OK or because of the 3rd case above.

Brian-McM: I've asked @tmjd about his preference about returning a special error or a bool, as this is likely setting a precedent.

pkg/controller/logstorage/logstorage.go
```diff
@@ -0,0 +1,288 @@
...
+) error {
+	var esInternalCertSecret, esAdminUserSecret, esCertSecret *corev1.Secret
+	var kibanaSecrets, curatorSecrets []*corev1.Secret
+	var esLicenseType render.ElasticsearchLicenseType
+
+	if managementClusterConnection == nil {
+		// Get the admin user secret to copy to the operator namespace.
+		esAdminUserSecret, err := utils.GetSecret(ctx, r.client, render.ElasticsearchAdminUserSecret, render.ElasticsearchNamespace)
+		if err != nil {
+			reqLogger.Error(err, err.Error())
```
Brian-McM, line 54: You'll need to set degraded here, and you should also set a better error message.

I think we should also be careful about doing `reqLogger.Error(err, err.Error())`, the second parameter is used to add additional context to the error. We set `reqLogger.Error(err, err.Error())` when there actually isn't any specific error, it's just a "not found" or we've decided we can't continue and want to add some context to the logs, but `reqLogger.Error` requires that an error is set.

pkg/controller/logstorage/logstorage.go
```diff
@@ -0,0 +1,288 @@
...
+		}
+		r.esAdminUserSecret = esAdminUserSecret
+
+		// Check if there is a StorageClass available to run Elasticsearch on.
+		if err := r.client.Get(ctx, client.ObjectKey{Name: ls.Spec.StorageClassName}, &storagev1.StorageClass{}); err != nil {
+			if errors.IsNotFound(err) {
+				err := fmt.Errorf("couldn't find storage class %s, this must be provided", ls.Spec.StorageClassName)
+				reqLogger.Error(err, err.Error())
+				r.status.SetDegraded("Failed to get storage class", err.Error())
+				return err
```
Brian-McM, line 68: I've added comments at the beginning of this file, but wanted to point one out in particular, this cannot return an error or we'll end up in a tight Reconcile loop.

pkg/controller/logstorage/logstorage.go
```diff
@@ -0,0 +1,288 @@
...
+		// Get the admin user secret to copy to the operator namespace.
+		esAdminUserSecret, err := utils.GetSecret(ctx, r.client, render.ElasticsearchAdminUserSecret, render.ElasticsearchNamespace)
+		if err != nil {
+			reqLogger.Error(err, err.Error())
+			return err
+		}
+		if esAdminUserSecret != nil {
+			esAdminUserSecret = rsecret.CopyToNamespace(rmeta.OperatorNamespace(), esAdminUserSecret)[0]
+		}
+		r.esAdminUserSecret = esAdminUserSecret
```
Brian-McM, line 60: We really shouldn't set this to the context of this struct. This information should only exist in the context of the Reconcile function, it needs to be re retrieved every call. This is analogous to setting a global variables after startup.

Brian-McM (changes requested):

pkg/controller/logstorage/logstorage.go
vberezny: For now I am checking to see if `Requeue` is set on the Reconcile object. I can add the third return value instead if that's what we end up going with.

PR #1474, opened 2021-08-23
Add support for DeepPacketInspection

Suraiya-Hameed:
## Description

- `intrusiondetections.operator.tigera.io` allows ResourceRequirements for component DeepPacketInspection
- Operator watches for `deeppacketinspection.projectcalico.org` and `intrusiondetections.operator.tigera.io` resource
- If there is DPI resource, create DPI daemonset if it doesn’t already exist
- If there are no DPI resources, delete/remove the DPI daemoset
- When creating DPI daemonset use default resource requirements if they are not explicitly specified in IDS resource
- If the resource requirement in IDS resource changes for DPI, update the DPI daemonset with newer resource requirements

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


api/v1/intrusiondetection_types.go
```diff
@@ -17,11 +17,16 @@ limitations under the License.
...
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 // IntrusionDetectionSpec defines the desired state of Tigera intrusion detection capabilities.
 type IntrusionDetectionSpec struct {
+	// ComponentResources can be used to customize the resource requirements for each component.
+	// Only DeepPacketInspection is supported for this spec.
+	// +optional
+	ComponentResources []IntrusionDetectionComponentResource `json:"componentResources,omitempty"`
```
asincu, line 29: Do we support only 1 component for intrusion detection ?

Suraiya-Hameed: Only DPI supports customising resource requirements for now.

pkg/render/intrusion_detection.go
```diff
@@ -892,3 +950,243 @@ func (c *intrusionDetectionComponent) intrusionDetectionAnnotations() map[string
...
+		{Name: "DPI_TYPHACERTFILE", Value: fmt.Sprintf("/node-certs/%s", TLSSecretCertName)},
+		{Name: "DPI_TYPHAKEYFILE", Value: fmt.Sprintf("/node-certs/%s", TLSSecretKeyName)},
+		cnEnv,
+		{Name: "DPI_TYPHAURISAN", ValueFrom: &v1.EnvVarSource{
+			SecretKeyRef: &v1.SecretKeySelector{
+				LocalObjectReference: v1.LocalObjectReference{
+					Name: TyphaTLSSecretName,
+				},
+				Key:      URISAN,
+				Optional: ptr.BoolToPtr(true),
```
asincu, line 1084: When do we need to make this optional ?

pkg/render/intrusion_detection.go
```diff
@@ -892,3 +950,243 @@ func (c *intrusionDetectionComponent) intrusionDetectionAnnotations() map[string
...
+	return []v1.VolumeMount{
+		{MountPath: "/typha-ca", Name: "typha-ca", ReadOnly: true},
+		{MountPath: "/node-certs", Name: "node-certs", ReadOnly: true},
+	}
+}
+
+func (c *intrusionDetectionComponent) dpiReadinessProbes() *v1.Probe {
+	return &v1.Probe{
+		Handler: v1.Handler{
+			HTTPGet: &v1.HTTPGetAction{
```
asincu, line 1100: What about protocol ? Is this using HTTP or HTTPS ?

pkg/render/intrusion_detection.go
```diff
@@ -892,3 +950,243 @@ func (c *intrusionDetectionComponent) intrusionDetectionAnnotations() map[string
...
+				Host: "localhost",
+				Path: "/liveness",
+				Port: intstr.FromInt(9097),
+			},
+		},
+		TimeoutSeconds: 10,
```
asincu, line 1106: What about `InitialDelaySeconds` and `PeriodSeconds` ?

pkg/render/intrusion_detection.go
```diff
@@ -892,3 +950,243 @@ func (c *intrusionDetectionComponent) intrusionDetectionAnnotations() map[string
...
+	return v1.Container{
+		Name:         DeepPacketInspectionName,
+		Image:        c.dpiImage,
+		Resources:    *c.ids.Spec.ComponentResources[0].ResourceRequirements,
+		Env:          c.dpiEnvVars(),
+		VolumeMounts: c.dpiVolumeMounts(),
+		SecurityContext: &corev1.SecurityContext{
+			Privileged: &privileged,
+		},
+		LivenessProbe: c.dpiReadinessProbes(),
```
asincu, line 1022: What about the readiness probe ?

pkg/render/intrusion_detection.go
```diff
@@ -892,3 +950,243 @@ func (c *intrusionDetectionComponent) intrusionDetectionAnnotations() map[string
...
+	return &v1.ServiceAccount{
+		TypeMeta: metav1.TypeMeta{Kind: "ServiceAccount", APIVersion: "v1"},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      DeepPacketInspectionName,
+			Namespace: DeepPacketInspectionNamespace,
+		},
+	}
+}
+
+func (c *intrusionDetectionComponent) dpiRoleBinding() *rbacv1.ClusterRoleBinding {
```
asincu, line 1120: [Cosmetic] This function returns a clusterrolebinding.

pkg/render/intrusion_detection.go
```diff
@@ -892,3 +950,243 @@ func (c *intrusionDetectionComponent) intrusionDetectionAnnotations() map[string
...
+			{
+				Kind:      "ServiceAccount",
+				Name:      DeepPacketInspectionName,
+				Namespace: DeepPacketInspectionNamespace,
+			},
+		},
+	}
+}
+
+func (c *intrusionDetectionComponent) dpiRole() *rbacv1.ClusterRole {
```
asincu, line 1142: [Cosmetic] This returns a clusterrole.

pkg/render/intrusion_detection.go
```diff
@@ -155,11 +190,33 @@ func (c *intrusionDetectionComponent) Objects() ([]client.Object, []client.Objec
 			c.intrusionDetectionPSPClusterRoleBinding())
 	}
 
+	if c.dpiDaemonSet {
+		c.intrusionDetectionComponentResources()
```
asincu, line 194: Is this still needed here ? We are actually appending https://github.com/tigera/operator/pull/1474/files#diff-41a87360d53438a77f76d9f2a90345ded1d71cb682d4db44041ee2f16a78a504R216

pkg/render/intrusion_detection.go
```diff
@@ -155,11 +190,33 @@ func (c *intrusionDetectionComponent) Objects() ([]client.Object, []client.Objec
...
+
 	if c.hasNoLicense {
 		return nil, objs
 	}
 
-	return objs, nil
+	// If IDS resource needs to be updated with default resourceRequirements,
+	// append it after license check is done to avoid deleting IDS CR.
+	if c.dpiDaemonSet {
+		objs = append(objs, c.ids)
```
asincu, line 216: Could we have some built-in defaults in the API instead of trying to populate them in our render ?

pkg/render/intrusion_detection.go
```diff
@@ -892,3 +950,243 @@ func (c *intrusionDetectionComponent) intrusionDetectionAnnotations() map[string
 	}
 	return nil
 }
+
+// intrusionDetectionComponentResources sets the default requirements value for DPI in IDS resource if it is not set.
+func (c *intrusionDetectionComponent) intrusionDetectionComponentResources() {
```
asincu, line 955: [Cosmetic] This function should contain the word `set` or `default` to signal that is cause side effects.

pkg/render/intrusion_detection.go
```diff
@@ -892,3 +950,243 @@ func (c *intrusionDetectionComponent) intrusionDetectionAnnotations() map[string
...
+		VolumeMounts: c.dpiVolumeMounts(),
+		SecurityContext: &corev1.SecurityContext{
+			Privileged: &privileged,
+		},
+		LivenessProbe: c.dpiReadinessProbes(),
+	}
+}
+
+func (c *intrusionDetectionComponent) dpiVolumes() []v1.Volume {
+	return []v1.Volume{
```
asincu, line 1027: Where are setting annotations for these volumes ?

pkg/render/intrusion_detection_test.go
```diff
@@ -38,15 +39,7 @@ var _ = Describe("Intrusion Detection rendering tests", func() {
...
-		component := render.IntrusionDetection(
-			nil,
-			nil,
-			&corev1.Secret{ObjectMeta: metav1.ObjectMeta{Name: render.TigeraKibanaCertSecret}},
-			&operatorv1.InstallationSpec{Registry: "testregistry.com/"},
-			esConfigMap, nil, notOpenshift, dns.DefaultClusterDomain, render.ElasticsearchLicenseTypeUnknown, notManagedCluster,
-			false,
-			&internalManagerTLSSecret,
-		)
+		component := render.IntrusionDetection(nil, nil, nil, &corev1.Secret{ObjectMeta: metav1.ObjectMeta{Name: render.TigeraKibanaCertSecret}}, &operatorv1.InstallationSpec{Registry: "testregistry.com/"}, esConfigMap, nil, notOpenshift, dns.DefaultClusterDomain, render.ElasticsearchLicenseTypeUnknown, notManagedCluster, false, &internalManagerTLSSecret, false, nil, nil, nil)
```
asincu, line 42: [Cosmetic] Could we keep the same formatting ? It is harder to review this since we are exceeding the 80 chars limit per line.

pkg/render/intrusion_detection_test.go
```diff
@@ -207,18 +192,78 @@ var _ = Describe("Intrusion Detection rendering tests", func() {
...
+
+		Expect(len(resources)).To(Equal(len(expectedResources)))
+
+		for i, expectedRes := range expectedResources {
+			rtest.ExpectResource(resources[i], expectedRes.name, expectedRes.ns, expectedRes.group, expectedRes.version, expectedRes.kind)
+		}
+
+		// IDS resource should be updated with default resource values for DPI
+		idsResp := rtest.GetResource(resources, "tigera-secure", "", "", "v1", "IntrusionDetection").(*operatorv1.IntrusionDetection)
+		Expect(len(idsResp.Spec.ComponentResources)).Should(Equal(1))
```
asincu, line 255: We should add checks for clusterroles, clusterrolebindings, volumes, volume mounts, daemonset annotations.

pkg/render/intrusion_detection_test.go
```diff
@@ -281,33 +326,25 @@ var _ = Describe("Intrusion Detection rendering tests", func() {
...
-			false,
-			nil,
-		)
+		component := render.IntrusionDetection(nil, nil, nil, &corev1.Secret{ObjectMeta: metav1.ObjectMeta{Name: render.TigeraKibanaCertSecret}}, &operatorv1.InstallationSpec{
+			ControlPlaneNodeSelector: map[string]string{"foo": "bar"},
+		}, &relasticsearch.ClusterConfig{}, nil, false, dns.DefaultClusterDomain, render.ElasticsearchLicenseTypeUnknown, notManagedCluster, false, nil, false, nil, nil, nil)
 		resources, _ := component.Objects()
 		idc := rtest.GetResource(resources, "intrusion-detection-controller", render.IntrusionDetectionNamespace, "", "v1", "Deployment").(*apps.Deployment)
 		job := rtest.GetResource(resources, render.IntrusionDetectionInstallerJobName, render.IntrusionDetectionNamespace, "batch", "v1", "Job").(*batchv1.Job)
 		Expect(idc.Spec.Template.Spec.NodeSelector).To(Equal(map[string]string{"foo": "bar"}))
```
asincu, line 335: Does this apply to the daemon set in any way ? Do you mind adding some checks to state that is does not ?

pkg/render/intrusion_detection_test.go
```diff
@@ -207,18 +192,78 @@ var _ = Describe("Intrusion Detection rendering tests", func() {
...
+		Expect(len(resources)).To(Equal(len(expectedResources)))
+
+		for i, expectedRes := range expectedResources {
+			rtest.ExpectResource(resources[i], expectedRes.name, expectedRes.ns, expectedRes.group, expectedRes.version, expectedRes.kind)
+		}
+
+		// IDS resource should be updated with default resource values for DPI
+		idsResp := rtest.GetResource(resources, "tigera-secure", "", "", "v1", "IntrusionDetection").(*operatorv1.IntrusionDetection)
+		Expect(len(idsResp.Spec.ComponentResources)).Should(Equal(1))
+		Expect(idsResp.Spec.ComponentResources[0].ComponentName).Should(Equal(operatorv1.ComponentNameDeepPacketInspection))
```
asincu, line 256: Do you mind adding a test when a user define the resource components values ?

pkg/render/intrusion_detection.go
```diff
@@ -892,3 +950,243 @@ func (c *intrusionDetectionComponent) intrusionDetectionAnnotations() map[string
...
+				ConfigMap: &v1.ConfigMapVolumeSource{
+					LocalObjectReference: v1.LocalObjectReference{
+						Name: TyphaCAConfigMapName,
+					},
+				},
+			},
+		},
+		{
+			Name:         "node-certs",
+			VolumeSource: certificateVolumeSource(c.installation.CertificateManagement, NodeTLSSecretName),
```
asincu, line 1040: Are we generating new client certificates for the nodes for dpi or simply binding to existing ones ?

api/v1/intrusiondetection_types.go
asincu: Yes, but I assume we can have multiple DPI components in the future. Not just DeepPacketInspection.

pkg/controller/intrusiondetection/intrusiondetection_controller.go
```diff
@@ -157,6 +160,17 @@ func add(mgr manager.Manager, c controller.Controller) error {
 		return fmt.Errorf("intrusiondetection-controller failed to watch the ConfigMap resource: %v", err)
 	}
 
+	// Watch for intrusiondetection resource and also secretes that needs to be mounted for dpi daemonset
+	if err = c.Watch(&source.Kind{Type: &v3.DeepPacketInspection{
```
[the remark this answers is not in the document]
Suraiya-Hameed: If api server is not available behaviour will be same as the current operator, intrusion-detection will be degraded as it can't get license.

pkg/render/intrusion_detection.go
```diff
@@ -892,3 +950,243 @@ func (c *intrusionDetectionComponent) intrusionDetectionAnnotations() map[string
...
+		},
+	}
+}
+
+func (c *intrusionDetectionComponent) dpiEnvVars() []v1.EnvVar {
+	var cnEnv v1.EnvVar
+	if c.installation.CertificateManagement != nil {
+		cnEnv = v1.EnvVar{
+			Name: "DPI_TYPHACN", Value: TyphaCommonName,
```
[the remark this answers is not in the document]
Suraiya-Hameed: This is the format expected by typha client
https://github.com/projectcalico/typha/blob/fffa7dc45a63bb6403b66b0c80173974b1ded268/pkg/syncclientutils/config.go#L47-L51

pkg/controller/intrusiondetection/intrusiondetection_controller.go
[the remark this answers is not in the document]
asincu: I would have assumed this watch would fail since we do have the API group installed.

pkg/render/intrusion_detection.go
```diff
@@ -892,3 +950,243 @@ func (c *intrusionDetectionComponent) intrusionDetectionAnnotations() map[string
...
+		}
+	} else {
+		cnEnv = v1.EnvVar{
+			Name: "DPI_TYPHACN", ValueFrom: &v1.EnvVarSource{
+				SecretKeyRef: &v1.SecretKeySelector{
+					LocalObjectReference: v1.LocalObjectReference{
+						Name: TyphaTLSSecretName,
+					},
+					Key:      CommonName,
+					Optional: ptr.BoolToPtr(true),
```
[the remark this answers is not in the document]
Suraiya-Hameed: No, typha certs should be present, core_controller creates it if it is not available or degrades if it not able to create one.

pkg/render/intrusion_detection.go
```diff
@@ -892,3 +950,243 @@ func (c *intrusionDetectionComponent) intrusionDetectionAnnotations() map[string
...
+	}
+	return []v1.EnvVar{
+		// TODO: check if we need DPI_DATASTORE_TYPE and DPI_WAIT_FOR_DATASTORE
+		{
+			Name: "NODENAME",
+			ValueFrom: &v1.EnvVarSource{
+				FieldRef: &v1.ObjectFieldSelector{FieldPath: "spec.nodeName"},
+			},
+		},
+		{Name: "DPI_TYPHAK8SNAMESPACE", Value: common.CalicoNamespace},
```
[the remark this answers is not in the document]
Suraiya-Hameed: I'll add this to the other repository. 

and we can't split the name as it is expected by typha client https://github.com/tigera/operator/pull/1474#discussion_r694295476

pkg/controller/intrusiondetection/intrusiondetection_controller.go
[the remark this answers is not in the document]
Suraiya-Hameed: There is mixture of both those failures

```
E0823 23:42:58.691965       1 reflector.go:127] pkg/mod/k8s.io/client-go@v0.19.2/tools/cache/reflector.go:156: Failed to watch *v3.LicenseKey: failed to list *v3.LicenseKey: the server is currently unable to handle the request (get licensekeys.projectcalico.org)
E0823 23:42:59.699225       1 reflector.go:127] pkg/mod/k8s.io/client-go@v0.19.2/tools/cache/reflector.go:156: Failed to watch *v3.DeepPacketInspection: failed to list *v3.DeepPacketInspection: the server is currently unable to handle the request (get deeppacketinspections.projectcalico.org)

```

pkg/render/intrusion_detection.go
Suraiya-Hameed: We are re-using the existing certificates, got an okay from Neil and Manoj on that.

pkg/controller/intrusiondetection/intrusiondetection_controller.go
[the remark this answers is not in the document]
asincu: What happens after the connection is re-established in case of the DPI resource ? Is the watch ever added for the DPI resource in this case ?

pkg/render/intrusion_detection.go
asincu: Could we bind the certificates as the volume source explicitly ? `certificateVolumeSource` has logic for certificate generation to create an empty directory in order to be populated by an init container. If this not needed, I would rather be more declarative with this code.

pkg/render/intrusion_detection.go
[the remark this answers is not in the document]
asincu: :( Ok, hard to read.

pkg/render/intrusion_detection.go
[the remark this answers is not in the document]
asincu: Do we need to mark this as optional ?

pkg/render/intrusion_detection.go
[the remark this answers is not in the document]
asincu: Yup, I am ok with the naming.

pkg/controller/intrusiondetection/intrusiondetection_controller.go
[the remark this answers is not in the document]
Suraiya-Hameed: Yes, when apiserver is back and available watches start working and intrusion-detection status becomes available

pkg/render/intrusion_detection.go
Suraiya-Hameed: Tried setting defaults using kubebuilder and json tags on the crd, it doesn't reflect on the intrusiondetections resource (I didn't delete and recreate the resource after applying updated crd), I'll leave this be.

pkg/render/intrusion_detection.go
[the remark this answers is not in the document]
Suraiya-Hameed: I've added comment about this in the code below. It needs to be optional, typha client needs either TYPHACN or TYPHAURISAN.

pkg/render/intrusion_detection.go
Suraiya-Hameed: Yes, added a comment about it.

pkg/render/intrusion_detection.go
Suraiya-Hameed: I've updated this to be a readiness probe (I'll make corresponding changes to DPI repository), there is no liveness probe, dpi pod panics if it encounters an irrecoverable issue so it does not necessarily need a liveness probe.

pkg/render/intrusion_detection.go
Suraiya-Hameed: Yes, the resource requirements needs to be set before appending, as those values are needed for creating daemonset. I've added a comment in the code too.

api/v1/intrusiondetection_types.go
Brian-McM, line 29: Was this structure agreed on already?

I don't think this should be an arbitrary array of resources, there are going to be a finite number of intrusion detection components, and not many of them, I would we would just want specific component fields, maybe nested under the Resource field.

pkg/render/intrusion_detection.go
```diff
@@ -19,6 +19,12 @@ import (
 	"strings"
```
Brian-McM, line 19: Could you please create a separate render component for this, we're trying to achieve a better separation of concerns. In future iterations we'll likely separate out the installer and the ids controller from this component.

You can take a look in the logstorage controller for some examples of how we combine multiple components.

The preference would also be to create a subpackage, named intrusion detection, and put this as a subpackage in that folder. You can look at the logstorage folder if you need a reference.

Brian-McM: We're also moving away from a massive list of function arguments and using configuration structs instead, there should be a few examples of this in the render package now.

Brian-McM (changes requested):

api/v1/intrusiondetection_types.go
Suraiya-Hameed: Yes, this was the agreed structure. It is not a new structure, it is already used by installation and logstorage component to set resource requirements
https://github.com/tigera/operator/blob/271d1b851db73b758c6c986f7f2f5170e253b724/api/v1/logstorage_types.go#L54
https://github.com/tigera/operator/blob/271d1b851db73b758c6c986f7f2f5170e253b724/api/v1/installation_types.go#L132

Brian-McM: Ah OK great.

pkg/controller/intrusiondetection/intrusiondetection_controller.go
```diff
@@ -356,6 +375,55 @@ func (r *ReconcileIntrusionDetection) Reconcile(ctx context.Context, request rec
...
+	typhaCAConfigMap := &corev1.ConfigMap{}
+	dpiList := &v3.DeepPacketInspectionList{}
+	if err := r.client.List(ctx, dpiList); err != nil {
+		r.status.SetDegraded("Failed to retrieve DeepPacketInspection resource", err.Error())
+		return reconcile.Result{}, err
+	}
+	if len(dpiList.Items) > 0 {
+		nodeTLSSecret, err = utils.GetSecret(ctx, r.client, render.NodeTLSSecretName, rmeta.OperatorNamespace())
+		if err != nil {
+			r.status.SetDegraded("Failed to retrieve node cert secret", err.Error())
```
Brian-McM, line 388: Could you please not abbreviate in the degraded messages, they're messages to the customer.

You should also log all errors so they're also available when viewing the operator logs.

pkg/controller/intrusiondetection/intrusiondetection_controller.go
```diff
@@ -356,6 +375,55 @@ func (r *ReconcileIntrusionDetection) Reconcile(ctx context.Context, request rec
...
+			return reconcile.Result{}, err
+		}
+
+		typhaTLSSecret, err = utils.GetSecret(ctx, r.client, render.TyphaTLSSecretName, rmeta.OperatorNamespace())
```
Brian-McM, line 392: GetSecret returns no error if the secret is not found. Is it OK if typhaTLSSecret is nil?

Brian-McM: Same comment goes for the other secrets.

pkg/render/intrusiondetection/dpi/dpi.go
```diff
@@ -0,0 +1,391 @@
...
+	DeepPacketInspectionNamespace = "tigera-dpi"
+	DeepPacketInspectionName      = "tigera-dpi"
+	DefaultMemoryLimitDPI         = "1Gi"
+	DefaultMemoryRequestDPI       = "100Mi"
+	DefaultCPULimitDPI            = "1"
+	DefaultCPURequestDPI          = "100m"
+)
+
+func DPI(
+	ids *operatorv1.IntrusionDetection,
```
Brian-McM, line 48: Casey has started a new pattern for passing arguments to render which I think you should adopt here https://github.com/tigera/operator/blob/master/pkg/render/typha.go#L61.

pkg/render/intrusiondetection/dpi/dpi.go
```diff
@@ -0,0 +1,391 @@
...
+	typhaTLSSecret   *corev1.Secret
+	typhaCAConfigMap *corev1.ConfigMap
+	pullSecrets      []*corev1.Secret
+	openshift        bool
+	hasNoLicense     bool
+	dpiImage         string
+}
+
+func (d *dpi) ResolveImages(is *operatorv1.ImageSet) error {
+	reg := d.installation.Registry
```
Brian-McM, line 82: Nit, but is there any point in putting these in variables as opposed to just using them directly?

pkg/render/intrusiondetection/dpi/dpi.go
```diff
@@ -0,0 +1,391 @@
...
+		return err
+	}
+	return nil
+}
+
+func (d *dpi) Objects() (objsToCreate, objsToDelete []client.Object) {
+	var toDelete []client.Object
+	var toCreate []client.Object
+	// Add DeepPacketInspection related objects if all necessary bit of TLS config are set.
+	addDPIDaemonSet := d.nodeTLSSecret != nil && d.typhaTLSSecret != nil && d.typhaCAConfigMap != nil
```
Brian-McM, line 97: If any of these are nil we don't want to create the dpi component, but I would think we'd expect them to eventually be created. I think this would warrant not rendering this component until they're available, and degrading. If somebody doesn't have typha correctly configured and there is DPI resource then this is an error, and we should degrade.

pkg/render/intrusiondetection/dpi/dpi.go
```diff
@@ -0,0 +1,391 @@
...
+			d.dpiServiceAccount(),
+			d.dpiClusterRole(),
+			d.dpiClusterRoleBinding(),
+			d.dpiDaemonset(),
+		)
+		return toCreate, nil
+	}
+
+	toDelete = append(toDelete, d.dpiNamespace())
+	return nil, toDelete
```
Brian-McM, line 121: Just to add a little more from the comment above, deleting the namespace because one of the secrets doesn't exist doesn't seem correct, it could be a temporary issue (maybe it was deleting on purpose so it would be regenerated).

pkg/render/intrusiondetection/dpi/dpi.go
```diff
@@ -0,0 +1,391 @@
...
+}
+
+func (d *dpi) SupportedOSType() meta.OSType {
+	return rmeta.OSTypeLinux
+}
+
+// defaultIntrusionDetectionComponentResources sets the default requirements value for DPI in IDS resource if it is not set.
+func (d *dpi) defaultIntrusionDetectionComponentResources() {
+	if d.ids.Spec.ComponentResources == nil {
+		d.ids.Spec.ComponentResources = []operatorv1.IntrusionDetectionComponentResource{
```
Brian-McM, line 135: We should not be modifying the ids resource in here, if there's defaults to set they should be set in the controller and written back to the k8s API.

pkg/render/intrusiondetection/dpi/dpi.go
```diff
@@ -0,0 +1,391 @@
...
+	// Add DeepPacketInspection related objects if all necessary bit of TLS config are set.
+	addDPIDaemonSet := d.nodeTLSSecret != nil && d.typhaTLSSecret != nil && d.typhaCAConfigMap != nil
+
+	// toCreate is populated only if DPI daemonset needs to be created and there is valid license,
+	// for all other combinations delete the DPI namespace.
+	if addDPIDaemonSet && !d.hasNoLicense {
+		// Update the IDS resource with the default values for DPI ComponentResources if it doesn't exist,
+		// these values are used for setting container resource requirements in DPI daemonset.
+		d.defaultIntrusionDetectionComponentResources()
+		toCreate = append(toCreate, d.ids)
```
Brian-McM, line 105: I think you probably pulled this from logstorage, but I don't think this is the correct way to write back the defaults, even more so because this could be done by the intrusion detection render code.

If this is returned as an object from the render code the owner reference will be changed.

Brian-McM (changes requested):

pkg/controller/utils/utils.go
```diff
@@ -471,3 +423,59 @@ func StrToElasticLicenseType(license string, logger logr.Logger) render.Elastics
 	logger.V(3).Info("Elasticsearch license %s is unexpected", license)
 	return render.ElasticsearchLicenseTypeUnknown
 }
+
+// WaitToAddResourceWatch will check if projectcalico.org APIs are available and if so, it will add a watch for resource
+// The completion of this operation will be signaled on a ready channel
+func WaitToAddResourceWatch(controller controller.Controller, client kubernetes.Interface, log logr.Logger, flag *ReadyFlag, resourceKind string) {
```
Brian-McM, line 429: I think that if we're going to make this more generic then we could just go all the way and have the ability to run an arbitrary function based off of an arbitrary condition that can be set by the caller. For instance, a crude implementation would be:

```
func WaitToExecute(f1 func(), f2 func() bool, done chan struct{}) {
	maxDuration := 30 * time.Second
	duration := 1 * time.Second
	ticker := time.NewTicker(duration)
	defer ticker.Stop()
	for {
		select {
		case <-ticker.C:
			duration = duration * 2
			if duration >= maxDuration {
				duration = maxDuration
			}
			ticker.Reset(duration)

			if f2() {
				f1()
				close(done)
			}
		}
	}
}
```

Then you could just have an exported `IsResourceReady`, and in your controller you would just run:
```
done := make(chan struct{})
	go utils.WaitToExecute(func() {
		controller.Watch(&source.Kind{Type: &v3.DeepPacketInspection{TypeMeta: metav1.TypeMeta{Kind: v3.KindDeepPacketInspection}}},
			&handler.EnqueueRequestForObject{})
	}, func() bool {
		return utils.IsResourceReady(client, v3.KindDeepPacketInspection)
	}, done)
```

It can be used in any situation where we need to delay watching based off of any arbitrary signal, it's very readable what's happening in the controller.

pkg/controller/intrusiondetection/intrusiondetection_controller.go
Suraiya-Hameed: It shouldn't be nil, thanks for pointing that, I'll update it.

pkg/render/intrusiondetection/dpi/dpi.go
Suraiya-Hameed: Okay, so I guess the default values should be set by intrusion detection controller before objects returned by intrusion detection render are created?

Brian-McM: Yes exactly, set the default before anything is rendered and write them back. I think the core controller has a good example of this.

pkg/controller/utils/utils.go
Suraiya-Hameed: I think I'll leave it the way it is for now, probably better to do it a separate PR.

Brian-McM: Sounds good!

pkg/controller/intrusiondetection/intrusiondetection_controller.go
```diff
@@ -204,6 +220,13 @@ func (r *ReconcileIntrusionDetection) Reconcile(ctx context.Context, request rec
 	r.status.OnCRFound()
 	reqLogger.V(2).Info("Loaded config", "config", instance)
 
+	// Set defaults for IntrusionDetection resource
+	if err := r.setDefaultsOnIntrusionDetection(ctx, instance); err != nil {
+		reqLogger.V(3).Info("failed to set defaults on IntrusionDetection CR", "err", err)
```
Brian-McM, line 225: I think this should just be an error, if we failed to set the defaults then a real error occurred and should be logged (as opposed to we're waiting for something to become available and we degrade).

pkg/controller/intrusiondetection/intrusiondetection_controller.go
```diff
@@ -19,6 +19,12 @@ import (
 	"fmt"
 	"time"
 
+	"k8s.io/apimachinery/pkg/api/resource"
```
Brian-McM, line 22: Can you please organise these imports in the same manner as here https://github.com/tigera/operator/blob/271d1b851db73b758c6c986f7f2f5170e253b724/pkg/controller/installation/core_controller.go#L17-L75.

pkg/controller/intrusiondetection/intrusiondetection_controller.go
```diff
@@ -356,6 +384,67 @@ func (r *ReconcileIntrusionDetection) Reconcile(ctx context.Context, request rec
...
+
+		typhaTLSSecret, err = utils.GetSecret(ctx, r.client, render.TyphaTLSSecretName, rmeta.OperatorNamespace())
+		if err != nil {
+			reqLogger.Error(err, fmt.Sprintf("Failed to retrieve %s secret", render.TyphaTLSSecretName))
+			r.status.SetDegraded(fmt.Sprintf("Failed to retrieve %s secret", render.TyphaTLSSecretName), err.Error())
+			return reconcile.Result{}, err
+		}
+
+		// If TLS secrets are not available, degrade.
+		if nodeTLSSecret == nil || typhaTLSSecret == nil {
```
Brian-McM, line 410: Nit, I think we normally do the checks separately and degrade, i.e. check nodeTLSSecret immediately after you check if there was an error then degrade saying that specific secret is not available.

pkg/controller/intrusiondetection/intrusiondetection_controller.go
```diff
@@ -356,6 +384,67 @@ func (r *ReconcileIntrusionDetection) Reconcile(ctx context.Context, request rec
...
+			reqLogger.Error(err, fmt.Sprintf("Waiting for both %s and %s secrets to be available", render.NodeTLSSecretName, render.TyphaTLSSecretName))
+			r.status.SetDegraded(fmt.Sprintf("Waiting for both %s and %s secrets to be available", render.NodeTLSSecretName, render.TyphaTLSSecretName), "")
+			return reconcile.Result{}, err
+		}
+
+		err = r.client.Get(ctx, types.NamespacedName{Name: render.TyphaCAConfigMapName, Namespace: rmeta.OperatorNamespace()}, typhaCAConfigMap)
```
Brian-McM, line 416: Don't you need to check if the typha config map is nil?

It's also surprising that there isn't a GetConfigMap function, like the GetSecret function.

pkg/controller/intrusiondetection/intrusiondetection_controller.go
```diff
@@ -356,6 +384,67 @@ func (r *ReconcileIntrusionDetection) Reconcile(ctx context.Context, request rec
...
+		if err != nil {
+			reqLogger.Error(err, fmt.Sprintf("Failed to retrieve %s configmap", render.TyphaCAConfigMapName))
+			r.status.SetDegraded(fmt.Sprintf("Failed to retrieve %s configmap", render.TyphaCAConfigMapName), err.Error())
+			return reconcile.Result{}, err
+		}
+	}
+
+	dpiCfg := dpi.DPIConfig{
```
Brian-McM, line 424: Nit but I'd just be tempted to not declare the extra variable and do this inline.

pkg/controller/utils/utils.go
```diff
@@ -471,3 +423,59 @@ func StrToElasticLicenseType(license string, logger logr.Logger) render.Elastics
...
+				} else {
+					flag.MarkAsReady()
+					return
+				}
+			}
+		}
+	}
+}
+
+func addResourceWatch(c controller.Controller, resourceKind string) error {
```
Brian-McM, line 455: I think that instead of using the resource kind and having to switch here, just have `WaitToAddResourceWatch` to accept the object you want to watch. 

Anytime somebody wants to watch a new resource using this they'll have to modify the switch. You've made this slightly more generic in the sense that you can call the same function to watch only two  different resource types, but I think this might be worse then have specific `WaitToAddLicenseKeyWatch` and `WaitToAddDPIKeyWatch` because the developer would have to dig into the internals to figure out that you can only watch these two resources with this.

pkg/render/intrusiondetection/dpi/dpi.go
```diff
@@ -0,0 +1,359 @@
...
+	Openshift          bool
+	HasNoLicense       bool
+}
+
+func DPI(cfg *DPIConfig) render.Component {
+	return &dpiComponent{cfg: cfg}
+}
+
+type dpiComponent struct {
+	// Given configuration.
```
Brian-McM, line 62: Nit, I love comments, but only when they add some kind of value. I think the fact that this has "Config" in the name that this is some sort of "given configuration". If you wanted to describe this as "the configuration passed in by the caller", I think that would add some value.

pkg/render/intrusiondetection/dpi/dpi.go
```diff
@@ -0,0 +1,359 @@
...
+	cfg *DPIConfig
+
+	// Generated internal config, built from the given configuration.
```
Brian-McM, line 65: Is comment correct?

Brian-McM (changes requested):

pkg/controller/intrusiondetection/intrusiondetection_controller.go
Suraiya-Hameed: If the configmap is not available client throws error here. I was expecting same behaviour from GetSecret, but it returns nil if secret is not available, only that needs a nil check.

Brian-McM: Oh right right, the GetSecret uses the client in this same way but swallows the not found error. Forget this comment.

pkg/controller/intrusiondetection/intrusiondetection_controller.go
```diff
@@ -76,18 +80,21 @@ func Add(mgr manager.Manager, opts options.AddOptions) error {
 
 	go utils.WaitToAddLicenseKeyWatch(controller, k8sClient, log, licenseAPIReady)
 
+	go utils.WaitToAddResourceWatch(controller, k8sClient, log, dpiAPIReady, &v3.DeepPacketInspection{TypeMeta: metav1.TypeMeta{Kind: v3.KindDeepPacketInspection}})
```
Brian-McM, line 83: Nit, but this is a long line, could you put the last parameter on a new line please?

Brian-McM (approved):

marvin-tigera:

PR #1475, opened 2021-08-24
Fix tigera-secure-es-http-certs-public upgrade issue

vberezny:


pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -746,6 +746,11 @@ func (r *ReconcileLogStorage) getElasticsearchCertificateSecrets(ctx context.Con
 					return nil, nil, err
 				}
 			}
+
+			newInternalSecret := render.CreateCertificateSecret(esKeyCert.Data[corev1.TLSCertKey], relasticsearch.InternalCertSecret, render.ElasticsearchNamespace)
+			if rmeta.SecretsAnnotationHash([]*corev1.Secret{internalSecret}...) != rmeta.SecretsAnnotationHash([]*corev1.Secret{newInternalSecret}...) {
+				certSecret = newInternalSecret
+			}
```
stevegaossou, line 753: I wonder why this wasn't already part of this code flow for handling the CA cert. It shouldn't have any negative impact to having it here and is a good catch all to cover our bases.  

@vberezny do you have any insight into why we didn't have this before? 

I am guessing we just never thought the CA cert would ever need to be changed (other than when the DNS name list is missing an expected domain).

Might be worth pinging either Rene or Brian just to be sure.

tmjd: I'm guessing we didn't have this before because, like suggested, we didn't consider that the CA would change.

I think maybe instead of this we should just do kind of like @stevegaossou suggested and check that `bytes.Equal(esKeyCert.Data[corev1.TLSCertKey], internal.Data[corev1.TLSCertKey])`, that does kind use knowledge of the internals of the secrets though. But it leaves out the Hashing which isn't necessary.

Also instead of directly creating the Secret perhaps we should do like we do if the DNS name isn't correct and delete the Secret, then let the "normal" reconcile create the secret.

stevegaossou: I'm also wondering why @hjiawei wasn't able to reproduce this during upgrade tests, after adding the `watch kubectl get tigerastatus/apiserver` step. 

This seems unrelated to the tigera-policies / packet capture namespace not found issue. The cert being stale seems like it should happen 100% of the time for upgrades (switching ECK versions should guarantee that the CA cert value would change).

PR #1500, opened 2021-08-31
Move misc Prometheus resources to Tigera Operator

hjiawei:
## Description

Move the following Prometheus resources to Operator:
* 99-alertmanager-secret.yaml
* 99-alertmanager-service.yaml

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/monitor.go
+++ b/pkg/render/monitor.go
@@ -49,6 +49,22 @@ const (
 
 	PrometheusHTTPAPIServiceName = "prometheus-http-api"
 	PrometheusDefaultPort        = 9090
+
+	// Write your alertmanager configuration file based on
+	// https://prometheus.io/docs/alerting/configuration/
+	alertmanagerConfig = `global:
+  resolve_timeout: 5m
+route:
+  group_by: ['job']
+  group_wait: 30s
+  group_interval: 1m
+  repeat_interval: 5m
+  receiver: 'webhook'
+receivers:
+- name: 'webhook'
+  webhook_configs:
+  - url: 'http://calico-alertmanager-webhook:30501/'
+`
 )
 
 func Monitor(
@@ -105,6 +121,8 @@ func (mc *monitorComponent) Objects() ([]client.Object, []client.Object) {
 		mc.role(),
 		mc.roleBinding(),
 		mc.alertmanager(),
+		mc.alertmanagerSecret(),
+		mc.alertmanagerService(),
 		mc.prometheus(),
 		mc.prometheusRule(),
 		mc.serviceMonitorCalicoNode(),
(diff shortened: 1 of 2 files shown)


hjiawei:
calico-private PR: https://github.com/tigera/calico-private/pull/3580

pkg/render/monitor.go
```diff
@@ -142,6 +160,44 @@ func (mc *monitorComponent) alertmanager() *monitoringv1.Alertmanager {
 	}
 }
 
+func (mc *monitorComponent) alertmanagerSecret() *corev1.Secret {
+	// This secret will be mounted as the Alertmanager configuration file.
+	// TODO: We should use the native CR support for this instead of using a secret for it.
```
hjiawei, line 165: This TODO comment is copied from the yaml file. This PR will not address this TODO item.

hjiawei:
Tested on a gcp-kubeadm cluster with alertmanager secret and service removed. They will be reconciled by the operator after this change.

pkg/render/monitor.go
```diff
@@ -142,6 +164,44 @@ func (mc *monitorComponent) alertmanager() *monitoringv1.Alertmanager {
...
+	return &corev1.Secret{
+		TypeMeta: metav1.TypeMeta{Kind: "Secret", APIVersion: "v1"},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "alertmanager-calico-node-alertmanager",
+			Namespace: common.TigeraPrometheusNamespace,
+		},
+		Data: map[string][]byte{
+			"alertmanager.yaml": []byte(alertmanagerConfig),
```
doublek, line 177: I haven't reviewed the whole PR but this alertmanager configuration is supposed to be user configurable. This is resource should be treated as something that the operator doesn't overwrite if one is already provided.

hjiawei: OK. Probably I need to leave it unmanaged by the operator. Thanks for clarification this.

hjiawei: PR updated to have the Alertmanager configuration secret in the Operator namespace and then copied to the Prometheus namespace. Any changes in the Operator namespace will be reflected in the Prometheus namespace.

pkg/render/monitor.go
```diff
@@ -142,6 +171,56 @@ func (mc *monitorComponent) alertmanager() *monitoringv1.Alertmanager {
 	}
 }
 
+func (mc *monitorComponent) alertmanagerSecrets() []*corev1.Secret {
+	// This secret will be mounted as the Alertmanager configuration file.
+	// Source secret is in the tigera-operator namespace and the target secret is in the tigera-prometheus namespace.
+	var secrets []*corev1.Secret
+
+	if mc.alertmanagerConfigSecret != nil {
+		secrets = append(secrets, mc.alertmanagerConfigSecret)
```
Brian-McM, line 180: We should be creating the operator secret in render. If it doesn't exist, it needs to be created in the controller. Only items that are _solely_ managed by the operator should be created here, customers will be interacting directly with this secret. 

This has to do with how we set the owner reference on objects created in the render components.

@tmjd can you remember where we have an example of this? I've been looking for one but I can't find any, I'm not sure if that's a bug or not...

pkg/controller/monitor/monitor_controller.go
```diff
@@ -190,7 +191,12 @@ func (r *ReconcileMonitor) Reconcile(ctx context.Context, request reconcile.Requ
 	hdler := utils.NewComponentHandler(log, r.client, r.scheme, instance)
 
 	// render prometheus components
-	component := render.Monitor(install, pullSecrets)
+	alertmanagerConfigSecret, err := r.getAlertmanagerConfigSecret(ctx)
+	if err != nil {
+		r.setDegraded(reqLogger, err, "Error retrieving Alertmanager configuration secret")
```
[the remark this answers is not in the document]
hjiawei: This `setDegraded` function call will log and error message and set tigerastatus to degraded.

pkg/render/monitor.go
Brian-McM: I meant to say "shouldn't", I meant the complete opposite of what my comments suggested lol. What you're 2 points are excellent what I meant to get across.

PR #1504, opened 2021-09-06
Support control plane replicas in Installation

hjiawei:
## Description

Make the following Calico Enterprise components deployment replica
configurable via the new optional ControlPlaneReplicas field in
InstallationSpec:
* tigera-dex/tigera-dex
* tigera-elasticsearch/tigera-secure-es-gateway
* tigera-manager/tigera-manager (standalone cluster only)
* tigera-system/tigera-apiserver

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/api/v1/installation_types.go
+++ b/api/v1/installation_types.go
@@ -105,6 +105,11 @@ type InstallationSpec struct {
 	// +optional
 	ControlPlaneTolerations []v1.Toleration `json:"controlPlaneTolerations,omitempty"`
 
+	// ControlPlaneReplicas defines how many replicas of the control plane core components will be deployed.
+	// This field applies to all control plane components that support High Availability. Defaults to 2.
+	// +optional
+	ControlPlaneReplicas *int32 `json:"controlPlaneReplicas,omitempty"`
+
 	// NodeMetricsPort specifies which port calico/node serves prometheus metrics on. By default, metrics are not enabled.
 	// If specified, this overrides any FelixConfiguration resources which may exist. If omitted, then
 	// prometheus metrics may still be configured through FelixConfiguration.
--- a/pkg/controller/logstorage/logstorage_controller.go
+++ b/pkg/controller/logstorage/logstorage_controller.go
@@ -20,18 +20,6 @@ import (
 
 	esv1 "github.com/elastic/cloud-on-k8s/pkg/apis/elasticsearch/v1"
 	kbv1 "github.com/elastic/cloud-on-k8s/pkg/apis/kibana/v1"
-	operatorv1 "github.com/tigera/operator/api/v1"
-	logstoragecommon "github.com/tigera/operator/pkg/controller/logstorage/common"
-	"github.com/tigera/operator/pkg/controller/options"
-	"github.com/tigera/operator/pkg/controller/status"
-	"github.com/tigera/operator/pkg/controller/utils"
-	"github.com/tigera/operator/pkg/controller/utils/imageset"
-	"github.com/tigera/operator/pkg/dns"
-	"github.com/tigera/operator/pkg/render"
-	relasticsearch "github.com/tigera/operator/pkg/render/common/elasticsearch"
-	rmeta "github.com/tigera/operator/pkg/render/common/meta"
-	rsecret "github.com/tigera/operator/pkg/render/common/secret"
-	"github.com/tigera/operator/pkg/render/logstorage/esgateway"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
@@ -46,6 +34,19 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
+
+	operatorv1 "github.com/tigera/operator/api/v1"
+	logstoragecommon "github.com/tigera/operator/pkg/controller/logstorage/common"
+	"github.com/tigera/operator/pkg/controller/options"
+	"github.com/tigera/operator/pkg/controller/status"
+	"github.com/tigera/operator/pkg/controller/utils"
+	"github.com/tigera/operator/pkg/controller/utils/imageset"
+	"github.com/tigera/operator/pkg/dns"
+	"github.com/tigera/operator/pkg/render"
+	relasticsearch "github.com/tigera/operator/pkg/render/common/elasticsearch"
+	rmeta "github.com/tigera/operator/pkg/render/common/meta"
+	rsecret "github.com/tigera/operator/pkg/render/common/secret"
+	"github.com/tigera/operator/pkg/render/logstorage/esgateway"
 )
 
 var log = logf.Log.WithName("controller_logstorage")
@@ -210,6 +211,7 @@ func GetLogStorage(ctx context.Context, cli client.Client) (*operatorv1.LogStora
 	return instance, nil
 }
 
+// fillDefaults populates the default values onto an LogStorage object.
 func fillDefaults(opr *operatorv1.LogStorage) {
 	if opr.Spec.Retention == nil {
 		opr.Spec.Retention = &operatorv1.Retention{}
--- a/pkg/controller/manager/manager_controller.go
+++ b/pkg/controller/manager/manager_controller.go
@@ -19,8 +19,6 @@ import (
 	"fmt"
 	"time"
 
-	tigerakvc "github.com/tigera/operator/pkg/render/common/authentication/tigera/key_validator_config"
-
 	operatorv1 "github.com/tigera/operator/api/v1"
 	"github.com/tigera/operator/pkg/common"
 	"github.com/tigera/operator/pkg/controller/compliance"
@@ -30,6 +28,7 @@ import (
 	"github.com/tigera/operator/pkg/controller/utils/imageset"
 	"github.com/tigera/operator/pkg/dns"
 	"github.com/tigera/operator/pkg/render"
+	tigerakvc "github.com/tigera/operator/pkg/render/common/authentication/tigera/key_validator_config"
 	relasticsearch "github.com/tigera/operator/pkg/render/common/elasticsearch"
 	rmeta "github.com/tigera/operator/pkg/render/common/meta"
 
@@ -495,6 +494,14 @@ func (r *ReconcileManager) Reconcile(ctx context.Context, request reconcile.Requ
 	// Create a component handler to manage the rendered component.
 	handler := utils.NewComponentHandler(log, r.client, r.scheme, instance)
 
+	// Set replicas to 1 for management or managed clusters.
+	// TODO Remove after MCM tigera-manager HA deployment is supported.
+	var replicas *int32 = installation.ControlPlaneReplicas
+	if managementCluster != nil || managementClusterConnection != nil {
+		var mcmReplicas int32 = 1
+		replicas = &mcmReplicas
+	}
+
 	// Render the desired objects from the CRD and create or update them.
 	component, err := render.Manager(
 		keyValidatorConfig,
@@ -512,6 +519,7 @@ func (r *ReconcileManager) Reconcile(ctx context.Context, request reconcile.Requ
 		internalTrafficSecret,
 		r.clusterDomain,
 		elasticLicenseType,
+		replicas,
 	)
 	if err != nil {
 		log.Error(err, "Error rendering Manager")
--- a/pkg/render/apiserver.go
+++ b/pkg/render/apiserver.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/tigera/operator/pkg/render/common/podaffinity"
 	"github.com/tigera/operator/pkg/render/common/podsecuritycontext"
 	"github.com/tigera/operator/pkg/render/common/podsecuritypolicy"
 
@@ -781,7 +782,6 @@ func (c *apiServerComponent) apiServerDeployment() *appsv1.Deployment {
 		name = "calico-apiserver"
 	}
 
-	var replicas int32 = 1
 	hostNetwork := c.hostNetwork()
 	dnsPolicy := corev1.DNSClusterFirst
 	if hostNetwork {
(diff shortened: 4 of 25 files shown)


pkg/render/common/podaffinity/pod_anti_affinity.go
```diff
@@ -0,0 +1,24 @@
...
+
+import (
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func NewPodAntiAffinity(name, namespace string) *corev1.Affinity {
+	return &corev1.Affinity{
+		PodAntiAffinity: &corev1.PodAntiAffinity{
+			RequiredDuringSchedulingIgnoredDuringExecution: []corev1.PodAffinityTerm{
```
matthewdupre, line 11: I'd be tempted to slightly weaken this to preferred

api/v1/apiserver_types.go
```diff
@@ -23,6 +23,9 @@ import (
 
 // APIServerSpec defines the desired state of Tigera API server.
 type APIServerSpec struct {
+	// Replicas defines how many Tigera API server replicas in the cluster. Defaults to 1.
```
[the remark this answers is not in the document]
caseydavenport: For typha we have an autoscaler that adjusts the replicas as needed based on cluster size, without a config knob for the user to set.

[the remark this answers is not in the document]
caseydavenport: In other words, I don't think defaulting is the way to do this. If the user sets it, we probably need to respect the user config.

If unset, we could have some default scaling behavior though

pkg/render/apiserver.go
```diff
@@ -801,6 +805,11 @@ func (c *apiServerComponent) apiServerDeployment() *appsv1.Deployment {
 			rmeta.APIServerNamespace(c.installation.Variant)))
 	}
 
+	var replicas int32 = DefaultReplicas
```
caseydavenport, line 808: We generally do defaulting in the controller, not in the render package. I think this should be no different, unless you have an explicit reason for not wanting to follow the convention?

pkg/render/guardian.go
```diff
@@ -202,6 +200,8 @@ func (c *GuardianComponent) clusterRoleBinding() client.Object {
 }
 
 func (c *GuardianComponent) deployment() client.Object {
+	var replicas int32 = DefaultReplicas
```
caseydavenport, line 203: I'm not sure it makes sense to have a global DefaultReplicas variable - is there really a conceptual global default that applies to all deployments we might want?

Like I said above, this defaulting should probably be done in the controller code anyway.

api/v1/apiserver_types.go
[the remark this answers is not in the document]
hjiawei: > If unset, we could have some default scaling behavior though

I can probably implement the following for apiserver:
* if set, respect the user config.
* if unset, use 2 for 5+ nodes and 1 otherwise.

pkg/controller/manager/manager_controller.go
```diff
@@ -495,6 +505,14 @@ func (r *ReconcileManager) Reconcile(ctx context.Context, request reconcile.Requ
 	// Create a component handler to manage the rendered component.
 	handler := utils.NewComponentHandler(log, r.client, r.scheme, instance)
 
+	// Reset replicas to 1 in management or managed clusters
+	// TODO: Support MCM tigera-manager HA deployment
+	if managementCluster != nil || managementClusterConnection != nil {
+		if *instance.Spec.Replicas != 1 {
+			instance.Spec.Replicas = ptr.Int32ToPtr(1)
```
tmjd, line 512: We shouldn't override this in the case of a managementCluster but should be verifying that with a managementCluster the field is set to 1.
Especially since 1 is the default value, which would mean that a user would explicitly be setting the value to 2.

hjiawei: @tmjd Just want to make sure I understand your comments correctly. This check will be removed when MCM HA is supported. You are suggesting to check for replicas but not overriding the user settings? For management and manged clusters, if a user sets it to be not 1, shall we set a degraded state in the Manager component?

api/v1/apiserver_types.go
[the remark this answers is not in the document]
hjiawei: We decided to set the default replicas to 2 independent of cluster size.

pkg/controller/manager/manager_controller.go
tmjd: I think because we changed the API and aren't providing an individual knob for manager replicas, I think the behavior of forcing to 1 replica for MCM clusters is acceptable.

caseydavenport: Probably OK to default it, but we still need to respect the value the user provided if its set (or error)

tmjd: I was expecting to include a comment on the field in Installation that it doesn't apply to the manager when using multi-cluster management. If that is not ok, we'll add a change in the cloud build to ignore the setting for the manager so we can utilize HA for the other components.

api/v1/installation_types.go
```diff
@@ -105,6 +105,10 @@ type InstallationSpec struct {
 	// +optional
 	ControlPlaneTolerations []v1.Toleration `json:"controlPlaneTolerations,omitempty"`
 
+	// Replicas defines how many replicas of the control plane core components will be deployed. Defaults to 2.
+	// +optional
+	ControlPlaneReplicas *int32 `json:"controlPlaneReplicas,omitempty"`
```
tmjd, line 110: @caseydavenport do you think we should have this field be a string for possible future expansion to non-numeric values? Or is there no concern about switching the field type at a later date?
I think we've done that type of change before.

pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -32,11 +33,13 @@ import (
 	rmeta "github.com/tigera/operator/pkg/render/common/meta"
 	rsecret "github.com/tigera/operator/pkg/render/common/secret"
 	"github.com/tigera/operator/pkg/render/logstorage/esgateway"
+
```
tmjd, line 36: If you want to adjust these imports to ourt standard format, please have 3 sections:
1. golang packages (this group is already there)
2. non-tigera packages
3. tigera packages
With a newline between each grouping.

pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -468,6 +472,7 @@ func (r *ReconcileLogStorage) Reconcile(ctx context.Context, request reconcile.R
 			hdler,
 			reqLogger,
 			ctx,
+			install.ControlPlaneReplicas,
```
tmjd, line 475: Since ESGateway is already passed the `install`, there is no need to separately pass ControlPlaneReplicas.

pkg/render/kubecontrollers/kube-controllers.go
```diff
@@ -487,7 +487,7 @@ func (c *kubeControllersComponent) controllersDeployment() *appsv1.Deployment {
 			},
 		},
 		Spec: appsv1.DeploymentSpec{
-			Replicas: &kubeControllerReplicas,
+			Replicas: &replicas,
```
tmjd, line 490: Shouldn't we be adding podAntiAffinity here too?

api/v1/installation_types.go
caseydavenport: I don't think we need to worry about non-numeric values for this.

pkg/controller/manager/manager_controller.go
caseydavenport: not a fan of ignoring user intent in the general case. I don't think most people will want to set `replicas: 2` only to have it ignored, even if running a management cluster.

If we need, we can either have the cloud build do something special, or introduce a new replicas field that applies only to the manager component.

tmjd: What is the concern if we lock the manager UI to 1 in the case of MCM. We're not removing previous available functionality but allowing this feature to be used with compatible components until the manager can also support HA with MCM.
It feels kind of the same as saying we can't do Replicas because we can't run all our components with multiple replicas.
But I do understand that it is not as straight forward since it would be necessary to include that caveat.

pkg/render/kubecontrollers/kube-controllers.go
hjiawei: Replicas count is still 1 for kube-controllers so it isn't needed for now.

pkg/controller/manager/manager_controller.go
caseydavenport: I suppose so long as we're comfortable with the following things it's OK:
- Bug reports of "I specified X replicas but still only have 1"
- Changing the behavior of the `ControlPlaneReplicas` field to apply to a different set of components in the future, and any ensuing bug reports.

Agree that they are probably pretty minor, but a strict interpretation of the API "rules" would attempt to avoid those.

caseydavenport: I suppose if we say this field applies to all control plane components that support HA, then we're OK. That sort of needs to be the case anyway, because kube-controllers.

hjiawei: Update the description for `ControlPlaneReplicas` to the following:

```golang
// Replicas defines how many replicas of the control plane core components will be deployed.
// This field applies to all control plane components that support High Availability. Defaults to 2.
// +optional
```

pkg/controller/logstorage/esgateway.go
```diff
@@ -75,6 +75,7 @@ func (r *ReconcileLogStorage) createEsGateway(
 		EsInternalCertSecret:       esInternalCertSecret,
 		ClusterDomain:              r.clusterDomain,
 		EsAdminUserName:            esAdminUserName,
+		Replicas:                   install.ControlPlaneReplicas,
```
tmjd, line 78: I thought I had commented on this but don't see that anymore. We already pass the install in to esgateway.Config, so there is no need for an explicit Replicas field here.

pkg/render/apiserver.go
```diff
@@ -838,7 +838,8 @@ func (c *apiServerComponent) apiServerDeployment() *appsv1.Deployment {
 					Containers: []corev1.Container{
 						c.apiServerContainer(),
 					},
-					Volumes: c.apiServerVolumes(),
+					Volumes:  c.apiServerVolumes(),
+					Affinity: podaffinity.NewPodAntiAffinity(name, rmeta.APIServerNamespace(c.installation.Variant)),
```
tmjd, line 842: I think we should only add this affinity if replicas is > 1. That way we can ensure no Affinity will be set if replicas is specified as 1.

pkg/render/apiserver_test.go
```diff
@@ -54,9 +54,11 @@ var _ = Describe("API server rendering tests (Calico Enterprise)", func() {
 	var k8sServiceEp k8sapi.ServiceEndpoint
 
 	BeforeEach(func() {
+		var replicas int32 = 2
 		instance = &operatorv1.InstallationSpec{
-			Registry: "testregistry.com/",
```
tmjd, line 58: Please include a test specifying 1 replica and ensure that test case ensures no Affinity is specified.
Please also include checking affinity is set with more than 1 replica

pkg/render/dex.go
```diff
@@ -268,7 +267,8 @@ func (c *dexComponent) deployment() client.Object {
 							VolumeMounts: c.dexConfig.RequiredVolumeMounts(),
 						},
 					},
-					Volumes: c.dexConfig.RequiredVolumes(),
+					Volumes:  c.dexConfig.RequiredVolumes(),
+					Affinity: podaffinity.NewPodAntiAffinity(DexObjectName, DexNamespace),
```
tmjd, line 271: Same as previous comment, the affinity should only be specified when Replicas is > 1.

pkg/render/dex_test.go
```diff
@@ -36,6 +36,7 @@ var _ = Describe("dex rendering tests", func() {
 			dexSecret      *corev1.Secret
 			idpSecret      *corev1.Secret
 			pullSecrets    []*corev1.Secret
+			replicas       int32 = 2
```
tmjd, line 39: Please include tests for Affinity when replicas is 1 and 2.

pkg/render/logstorage/esgateway/esgateway.go
```diff
@@ -288,6 +291,7 @@ func (e esGateway) esGatewayDeployment() *appsv1.Deployment {
 					},
 				},
 			},
+			Affinity: podaffinity.NewPodAntiAffinity(DeploymentName, render.ElasticsearchName),
```
tmjd, line 294: Please only add Affinity if Replicas is > 1.

pkg/render/logstorage/esgateway/esgateway_test.go
```diff
@@ -46,6 +45,7 @@ type resourceTestObj struct {
 var _ = Describe("ES Gateway rendering tests", func() {
 	Context("ES Gateway deployment", func() {
 		var installation *operatorv1.InstallationSpec
+		var replicas int32 = 2
```
tmjd, line 48: Please include Affinity checks in the the tests.

pkg/render/manager.go
```diff
@@ -315,7 +318,8 @@ func (c *managerComponent) managerDeployment() *appsv1.Deployment {
 				relasticsearch.ContainerDecorate(c.managerEsProxyContainer(), c.esClusterConfig.ClusterName(), ElasticsearchManagerUserSecret, c.clusterDomain, c.SupportedOSType()),
 				c.managerProxyContainer(),
 			},
-			Volumes: c.managerVolumes(),
+			Volumes:  c.managerVolumes(),
+			Affinity: podaffinity.NewPodAntiAffinity("tigera-manager", ManagerNamespace),
```
tmjd, line 322: Please make the Affinity dependent on the replica count.

tmjd (commented):
A few more things.
Mainly I'd like to see more render tests and think we should only include Affinity when Replicas > 1.

tmjd (changes requested):
Sorry one more thing I thought of. We should validate that Replicas is >0, obviously negative is invalid but I think 0 should be invalid too. I don't think there is a valid use case for 0 replicas. So it should be invalid. Please add it to https://github.com/hjiawei/operator/blob/calient-service-ha/pkg/controller/installation/validation.go#L308

hjiawei:
Added ControlPlaneReplicas validation and UT.

tmjd (approved):
LGTM

PR #1505, opened 2021-09-07
changes for deploying l7 log collector daemon set using operator for calient 

penkeysuresh:
## Description

As of calico enterprise v3.10, we enable L7 logs manually by [applying manifests](https://docs.tigera.io/v3.10/visibility/elastic/l7/configure#configure-l7-logs). This UX is cumbersome and can be improved by using operator to deploy the l7 daemon set. This PR included the designed UX for the same. 

  * added new resource`ApplicationLayer`, used for deploying l7 log collector daemonset
  * added controller and renderer and test case for both
  * updated go to v1.16 and GO_BUILD_VER to v0.56
  * added envoy-config.yaml needed for log collection, embed in binary

issue on jira : https://tigera.atlassian.net/browse/BPF-1380
docs change: https://github.com/tigera/calico-private/pull/3701

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


pkg/render/applicationlayer_test.go
```diff
@@ -0,0 +1,150 @@
...
+	operatorv1 "github.com/tigera/operator/api/v1"
+	"github.com/tigera/operator/pkg/render"
+
+	rmeta "github.com/tigera/operator/pkg/render/common/meta"
+	rtest "github.com/tigera/operator/pkg/render/common/test"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+)
+
+var _ = FDescribe("Tigera Secure Application Layer rendering tests", func() {
```
penkeysuresh, line 31: note: have to remove this focus

api/v1/applicationlayer_types.go
```diff
@@ -0,0 +1,86 @@
...
+
+	// Interval in seconds for sending L7 log information for processing.
+	// adjust this to limit the frequency at which logs are sent from l7-collector
+	// +optional
+	// Default: 5 sec
+	LogIntervalSeconds *int64 `json:"logIntervalSeconds,omitempty"`
+
+	// Maximum number of unique L7 logs that are sent LogIntervalSeconds
+	// adjust this to limit the number of L7 logs sent per LogIntervalSeconds
+	// to felix for further processing, use negative number to ignore limits
```
mgleung, line 50: We may want to add a comment here saying what happens if we go over the limit

pkg/apis/crd.projectcalico.org/v1/felixconfig.go
```diff
@@ -349,6 +357,11 @@ type FelixConfigurationSpec struct {
 	// Set source-destination-check on AWS EC2 instances. Accepted value must be one of "DoNothing", "Enabled" or "Disabled".
 	// [Default: DoNothing]
 	AWSSrcDstCheck *AWSSrcDstCheckOption `json:"awsSrcDstCheck,omitempty" validate:"omitempty,oneof=DoNothing Enable Disable"`
+
+	// TPROXYMode sets whether traffic is directed through a transparent proxy for further processing
+	// or not and how is the proxying done.
```
mgleung, line 362: I think since the options are just "Enabled" and "Disabled" we can probably remove the portion explaining that this tells how the proxying is done since this just decides whether we set up the tproxy to go through envoy for select traffic.

mgleung:
Do we also need to add the relevant PSPs or SCCs for this? I think we might be able to get away with not doing the PSPs because those are being deprecated in some K8s v1.21 but I think OCP will need the SCCs still.

pkg/apis/crd.projectcalico.org/v1/felixconfig.go
penkeysuresh: fixed in https://github.com/tigera/operator/pull/1505/commits/c5ff8851d37c8e53234c9eb3276c45b555673fbd

penkeysuresh:
couple of things that are missing in this after recent comments on design doc
1. Adding a watcher for felix configuration.
2. Adding a hash of for config map to daemon set, so that any change in config map starts the daemon set rolling.

pkg/controller/applicationlayer/applicationlayer_controller.go
```diff
@@ -0,0 +1,249 @@
...
+	licenseAPIReady *utils.ReadyFlag
+}
+
+// Reconcile reads that state of the cluster for a ApplicationLayer object and makes changes
+// based on the state read and what is in the ApplicationLayer.Spec
+func (r *ReconcileApplicationLayer) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
+	reqLogger := log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
+	reqLogger.Info("Reconciling ApplicationLayer")
+
+	instance, err := GetApplicationLayer(ctx, r.client)
```
tmjd, line 134: We need to include a defaulting step below for this resource. You can see in the core or authentication controllers how we do that.

tmjd (changes requested):
I'm still reviewing the render but wanted to go ahead and submit this set of comments.

pkg/render/applicationlayer.go
```diff
@@ -0,0 +1,308 @@
+// Copyright (c) 2021 Tigera, Inc. All rights reserved.
```
Brian-McM, line 1: We're moving towards a different subpackage architecture, so this should be in an "applicationlayer" subpackage instead of at the root of the render folder.

Look at the logstorage, kubecontrollers, and intrusiondetection packages in the render folder for examples of what I mean. Also DM me if there's anything you want to talk about.

Brian-McM (changes requested):

pkg/render/applicationlayer.go
```diff
@@ -0,0 +1,308 @@
...
+	}
+}
+
+type L7LogCollectionSpec struct {
+	Enabled                bool
+	LogIntervalSeconds     *int64
+	LogRequestsPerInterval *int64
+}
+
+type ApplicationLayerComponent struct {
```
Brian-McM, line 67: I don't think this needs to be (or should be) exported.

pkg/render/applicationlayer.go
```diff
@@ -0,0 +1,308 @@
...
+)
+
+func ApplicationLayer(pullSecrets []*corev1.Secret, installation *operatorv1.InstallationSpec, osType rmeta.OSType,
+	applicationLayer *operatorv1.ApplicationLayer, l7Spec *L7LogCollectionSpec) Component {
+
+	return &ApplicationLayerComponent{
+		pullSecrets:      pullSecrets,
+		installation:     installation,
+		osType:           osType,
+		applicationLayer: applicationLayer,
```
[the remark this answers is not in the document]
Brian-McM: I may be wrong but it doesn't look like this field is used.

pkg/render/applicationlayer.go
```diff
@@ -0,0 +1,308 @@
...
+		l7Spec:           l7Spec,
+	}
+}
+
+type L7LogCollectionSpec struct {
```
Brian-McM, line 61: I'm not sure there's much benefit in having a struct for just these fields. I think these fields should either just be parameters to the ApplicationLayer function or create a general config struct that the ApplicationLayer function accepts (there's a few examples of this in the other component constructors).

statik/files/envoy-config.yaml
```diff
@@ -0,0 +1,180 @@
+static_resources:
```
Brian-McM, line 1: statik => static (folder name)

Brian-McM (changes requested):

pkg/render/applicationlayer.go
```diff
@@ -0,0 +1,308 @@
...
+func (c *ApplicationLayerComponent) SupportedOSType() rmeta.OSType {
+	return rmeta.OSTypeLinux
+}
+
+func (c *ApplicationLayerComponent) Objects() ([]client.Object, []client.Object) {
+	var objs []client.Object
+	// TODO: ensure that namespace exists before proceeding
+	// if l7spec is provided render the required objects
+	if c.l7Spec != nil && c.l7Spec.Enabled {
+		objs = append(objs, c.envoyL7ConfigMap())
```
tmjd, line 116: I also wonder if this will need a PodSecurityPolicy on Openshift like calico-node.
Have you tried on Openshift?

tmjd: Also looking at node, I think you might need admin role binding for Docker EE.
https://github.com/tigera/operator/blob/14b2e7ca82c64b94ebb9e420efd67980fee62c30/pkg/render/node.go#L568

pkg/render/applicationlayer.go
```diff
@@ -0,0 +1,308 @@
...
+			Containers:       c.containers(),
+			Volumes:          c.volumes(),
+		},
+	}
+
+	ds := &appsv1.DaemonSet{
+		TypeMeta: metav1.TypeMeta{Kind: "DaemonSet", APIVersion: "apps/v1"},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      L7LogCollectorDeamonsetName,
+			Namespace: CalicoSystemNamespace,
```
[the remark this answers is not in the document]
doublek: IIUC with GKE's default quotas + the recent fix this should be okay to be added to the calico-system namespace.

pkg/render/applicationlayer.go
```diff
@@ -0,0 +1,308 @@
...
+}
+
+func (c *ApplicationLayerComponent) volumes() []corev1.Volume {
+
+	var volumes []corev1.Volume
+
+	volumes = append(volumes, corev1.Volume{
+		Name: EnvoyLogsKey,
+		VolumeSource: corev1.VolumeSource{
+			EmptyDir: &corev1.EmptyDirVolumeSource{},
```
tmjd, line 255: What is this? adding a volume that is empty? what's the point?
If this is needed please include a comment why so it is clear why we need it.

pkg/render/applicationlayer.go
```diff
@@ -0,0 +1,308 @@
...
+
+func (c *ApplicationLayerComponent) Ready() bool {
+	return true
+}
+
+// daemonset creates a daemonset for the L7 log collector component.
+func (c *ApplicationLayerComponent) daemonset() *appsv1.DaemonSet {
+	maxUnavailable := intstr.FromInt(1)
+
+	annots := map[string]string{}
```
tmjd, line 131: We should have an annotation for the configmap so if it changes in the future the daemonset will be updated.

pkg/render/common/statik/statikfiles.go
```diff
@@ -0,0 +1,37 @@
...
+	"io/ioutil"
+
+	"github.com/rakyll/statik/fs"
+	_ "github.com/tigera/operator/statik"
+)
+
+// getStatikFile gets the static file embeded into the binary using statik library (https://github.com/rakyll/statik).
+// All static files are located in statik/files. statik/statik.go is the source file that is generated
+// from the static files. It can be recreated using the following command.
+// "statik -src=/home/suresh/go/src/github.com/tigera/operator/statik/files"
```
tmjd, line 14: I didn't understand, I thought you were adding something that would include a file at run time.
If this requires a separate step it should be containerized and ran as part of the build process.

tmjd (changes requested):

api/v1/applicationlayer_types.go
```diff
@@ -0,0 +1,86 @@
...
+package v1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// ApplicationLayerSpec defines the desired state of ApplicationLayer
+type ApplicationLayerSpec struct {
+	// specification for L7 log collection
+	L7LogCollection *L7LogCollectionSpec `json:"l7LogCollection,omitempty"`
```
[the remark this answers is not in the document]
penkeysuresh: removed L7 from all the references in 43f4a774e6dd86130f71056953049f8401b324a9

config/samples/operator_v1_applicationlayer.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: operator.tigera.io/v1
+kind: ApplicationLayer
+metadata:
+  name: applicationlayer-sample
```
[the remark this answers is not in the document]
penkeysuresh: fixed in 43f4a774e6dd86130f71056953049f8401b324a9

pkg/apis/crd.projectcalico.org/v1/felixconfig.go
```diff
@@ -349,6 +357,10 @@ type FelixConfigurationSpec struct {
 	// Set source-destination-check on AWS EC2 instances. Accepted value must be one of "DoNothing", "Enabled" or "Disabled".
 	// [Default: DoNothing]
 	AWSSrcDstCheck *AWSSrcDstCheckOption `json:"awsSrcDstCheck,omitempty" validate:"omitempty,oneof=DoNothing Enable Disable"`
+
+	// TPROXYMode sets whether traffic is directed through a transparent proxy for further processing or not
+	// [Default: Disabled]
+	TPROXYMode *TPROXYModeOption `json:"tproxyMode,omitempty"`
```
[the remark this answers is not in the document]
penkeysuresh: yes, the real felix config contains the same comment and name

pkg/controller/applicationlayer/applicationlayer_controller.go
```diff
@@ -0,0 +1,249 @@
...
+
+	if err = utils.AddNetworkWatch(c); err != nil {
+		return fmt.Errorf("applicationlayer-controller failed to watch Network resource: %w", err)
+	}
+
+	if err = imageset.AddImageSetWatch(c); err != nil {
+		return fmt.Errorf("applicationlayer-controller failed to watch ImageSet: %w", err)
+	}
+
+	if err = utils.AddAPIServerWatch(c); err != nil {
```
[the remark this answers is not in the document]
penkeysuresh: I took that as a boiler plate code from some other controller, removed it in 43f4a774e6dd86130f71056953049f8401b324a9

pkg/controller/applicationlayer/applicationlayer_controller.go
```diff
@@ -0,0 +1,249 @@
...
+			return reconcile.Result{}, err
+		}
+
+		if err := ch.CreateOrUpdateOrDelete(ctx, l7component, r.status); err != nil {
+			r.status.SetDegraded("Error creating / updating resource", err.Error())
+			return reconcile.Result{}, err
+		}
+
+	}
+
```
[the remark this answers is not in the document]
penkeysuresh: Thanks added in 43f4a774e6dd86130f71056953049f8401b324a9 , I went through others all seem to have it, [cluster connection controller](https://github.com/tigera/operator/blob/master/pkg/controller/clusterconnection/clusterconnection_controller.go#L218) isn't doing the status available check. Perhaps it doesn't need it ? or missing ?

pkg/controller/applicationlayer/applicationlayer_controller.go
```diff
@@ -0,0 +1,249 @@
...
+	l7Spec := instance.Spec.L7LogCollection
+
+	err = r.patchFelixTproxyMode(ctx, l7Spec)
+
+	if err != nil {
+		r.status.SetDegraded("Error patching felix configuration", err.Error())
+		return reconcile.Result{}, err
+	}
+
+	if r.enableL7LogsCollection(l7Spec) {
```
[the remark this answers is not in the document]
penkeysuresh: thanks added your observation as a todo for future devs in 43f4a774e6dd86130f71056953049f8401b324a9

pkg/controller/applicationlayer/applicationlayer_controller.go
```diff
@@ -0,0 +1,249 @@
...
+	}
+
+	pullSecrets, err := utils.GetNetworkingPullSecrets(installation, r.client)
+
+	if err != nil {
+		r.status.SetDegraded("Error retrieving pull secrets", err.Error())
+		return reconcile.Result{}, err
+	}
+
+	l7Spec := instance.Spec.L7LogCollection
```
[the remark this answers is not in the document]
penkeysuresh: added a validation, also added a note for future devs

pkg/controller/applicationlayer/applicationlayer_controller.go
```diff
@@ -0,0 +1,249 @@
...
+
+	if r.enableL7LogsCollection(l7Spec) {
+		enabled := crdv1.TPROXYModeOptionEnabled
+		fc.Spec.TPROXYMode = &enabled
+	} else {
+		disabled := crdv1.TPROXYModeOptionDisabled
+		fc.Spec.TPROXYMode = &disabled
+	}
+
+	//log.Info("Patch default FelixConfiguration with %s", *fc.Spec.TPROXYMode)
```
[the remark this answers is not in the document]
penkeysuresh: I added logging in https://github.com/tigera/operator/commit/43f4a774e6dd86130f71056953049f8401b324a9 will keep a check to modify it only when needed shortly.

pkg/controller/applicationlayer/applicationlayer_controller.go
```diff
@@ -0,0 +1,249 @@
...
+		r.status.SetDegraded("Unable to read FelixConfiguration", err.Error())
+		return err
+	}
+
+	patchFrom := client.MergeFrom(fc.DeepCopy())
+
+	if r.enableL7LogsCollection(l7Spec) {
+		enabled := crdv1.TPROXYModeOptionEnabled
+		fc.Spec.TPROXYMode = &enabled
+	} else {
```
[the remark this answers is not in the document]
penkeysuresh: I'm inclined to keep it as it was before, (not have any code that disables it explicitly)

pkg/render/applicationlayer.go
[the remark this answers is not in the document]
penkeysuresh: removed it in d17bee13c2976396b9f6a3e918c031a77eae8606

pkg/render/applicationlayer.go
penkeysuresh: refactored it in d17bee13c2976396b9f6a3e918c031a77eae8606

pkg/render/applicationlayer.go
penkeysuresh: removed the struct, passing it as parameters to the ApplicationLayer function

pkg/render/applicationlayer.go
penkeysuresh: added a comment explaining the reason d17bee1

pkg/controller/applicationlayer/applicationlayer_controller.go
```diff
@@ -0,0 +1,249 @@
...
+	pullSecrets, err := utils.GetNetworkingPullSecrets(installation, r.client)
+
+	if err != nil {
+		r.status.SetDegraded("Error retrieving pull secrets", err.Error())
+		return reconcile.Result{}, err
+	}
+
+	l7Spec := instance.Spec.L7LogCollection
+
+	err = r.patchFelixTproxyMode(ctx, l7Spec)
```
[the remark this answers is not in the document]
penkeysuresh: added a watch in 6ac228c14e95f47014b6ee761adb53bd5f8467fb testing the changes

pkg/render/applicationlayer/applicationlayer.go
```diff
@@ -0,0 +1,394 @@
...
+		pullSecrets:            pullSecrets,
+		installation:           installation,
+		osType:                 osType,
+		logsEnabled:            logsEnabled,
+		logIntervalSeconds:     logIntervalSeconds,
+		logRequestsPerInterval: logRequestsPerInterval,
+	}
+}
+
+type applicationLayerComponent struct {
```
Brian-McM, line 76: Nit but the package name is applicationLayer so you could just name this `component` and it's clear that it's an applicationLayer component.

penkeysuresh: fixed in 32a037380b877cb81e1d9e497672aeed296391a8

api/v1/applicationlayer_types.go
```diff
@@ -0,0 +1,86 @@
...
+
+package v1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// ApplicationLayerSpec defines the desired state of ApplicationLayer
+type ApplicationLayerSpec struct {
+	// specification for application layer (L7) log collection
```
Brian-McM, line 25: Nit please capitalise and add punctuation to you're comments.

pkg/controller/applicationlayer/applicationlayer_controller.go
```diff
@@ -0,0 +1,319 @@
...
+// limitations under the License.
+
+package applicationlayer
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/tigera/operator/pkg/common"
```
Brian-McM, line 22: Could you please keep all the operator imports together.

pkg/controller/applicationlayer/applicationlayer_controller.go
```diff
@@ -0,0 +1,319 @@
...
+	"github.com/tigera/operator/pkg/render/applicationlayer"
+
+	operatorv1 "github.com/tigera/operator/api/v1"
+	crdv1 "github.com/tigera/operator/pkg/apis/crd.projectcalico.org/v1"
+	rmeta "github.com/tigera/operator/pkg/render/common/meta"
+
+	"github.com/tigera/operator/pkg/controller/options"
+	"github.com/tigera/operator/pkg/controller/status"
+	"github.com/tigera/operator/pkg/controller/utils"
+	"github.com/tigera/operator/pkg/controller/utils/imageset"
```
Brian-McM, line 33: Could you please separate the operator imports from the k8s ones.

pkg/controller/applicationlayer/applicationlayer_controller.go
```diff
@@ -0,0 +1,319 @@
...
+
+// Add creates a new ApplicationLayer Controller and adds it to the Manager. The Manager will set fields on the Controller
+// and Start it when the Manager is Started.
+func Add(mgr manager.Manager, opts options.AddOptions) error {
+	if !opts.EnterpriseCRDExists {
+		// No need to start this controller.
+		return nil
+	}
+	var licenseAPIReady = &utils.ReadyFlag{}
+	// create the reconciler
```
Brian-McM, line 58: This is probably just personal preference (so you don't need to do anything) but I don't really think obvious comments like this are helpful. I think it's pretty clear calling `newReconciler` will "create the reconciler" lol.

pkg/controller/applicationlayer/applicationlayer_controller.go
```diff
@@ -0,0 +1,319 @@
...
+	err = c.Watch(&source.Kind{Type: &operatorv1.ApplicationLayer{}}, &handler.EnqueueRequestForObject{})
+	if err != nil {
+		return err
+	}
+
+	if err = imageset.AddImageSetWatch(c); err != nil {
+		return fmt.Errorf("applicationlayer-controller failed to watch ImageSet: %w", err)
+	}
+
+	// watch configmaps created for envoy
```
Brian-McM, line 106: Nit, but could you please be consistent with the capitalisation and punctuation in your comments. I think we always capitalise the sentences, and always add periods to the end.

pkg/controller/applicationlayer/applicationlayer_controller.go
```diff
@@ -0,0 +1,319 @@
...
+		clusterDomain:   opts.ClusterDomain,
+		licenseAPIReady: licenseAPIReady,
+	}
+	r.status.Run()
+	return r
+}
+
+// add adds watches for resources that are available at startup
+func add(mgr manager.Manager, c controller.Controller) error {
+	var err error
```
Brian-McM, line 94: I think you need to watch for the installation resource with `utils.AddNetworkWatch(c)`.

pkg/controller/applicationlayer/applicationlayer_controller.go
```diff
@@ -0,0 +1,319 @@
...
+		r.status.SetDegraded("Failed to write defaults", err.Error())
+		return reconcile.Result{}, err
+	}
+
+	variant, installation, err := utils.GetInstallation(ctx, r.client)
+
+	if err != nil {
+		if errors.IsNotFound(err) {
+			r.status.SetDegraded("Installation not found", err.Error())
+			return reconcile.Result{}, err
```
Brian-McM, line 178: I know we do this everywhere else for the installation controller, but I'm wondering if we should stop returning an error here. Normally if the resource isn't found we set degraded and just return, and just rely on watching the missing object to trigger the reconcile when the object is available.

@tmjd what do you think, should we stop returning errors on this?

pkg/controller/applicationlayer/applicationlayer_controller.go
```diff
@@ -0,0 +1,319 @@
...
+		}
+		r.status.SetDegraded("Error querying installation", err.Error())
```
Brian-McM, line 180: Could you please log the errors whenever you set the status to Degraded. This allows users to see where it failed in the logs.

This comment applies to all the missing logs below where you've set degraded.

pkg/controller/applicationlayer/applicationlayer_controller.go
```diff
@@ -0,0 +1,319 @@
...
+		r.status.SetDegraded("Error retrieving pull secrets", err.Error())
+		return reconcile.Result{}, err
+	}
+
+	lcSpec := applicationLayer.Spec.LogCollection
+
+	// if ApplicationLayer spec exists then LogCollection should be set
+	// TODO: when we will have multiple features in future this should change to at least one feature being set
+	if lcSpec == nil {
+		r.status.SetDegraded(fmt.Sprintf("Error retrieving LogCollection Spec"), "")
```
Brian-McM, line 196: Could you please be more specific with you error, this is a user facing message so we need to provide details on what is wrong, i.e. there is no Spec section in the ApplicationLayer resource and that is required.

pkg/controller/applicationlayer/applicationlayer_controller.go
```diff
@@ -0,0 +1,319 @@
...
+		return reconcile.Result{}, err
```
Brian-McM, line 197: err here has not been set to anything useful, and will likely just be nil. I would suggest not return an error here, just log (please add the error logline) and set degraded (as you have).

When an error is returned from Reconcile, Reconcile is immediately retried. This is not a temporary error, this is a misconfiguration of a resource that is being watched. Until the user fixes the resource, this will fail, and when the user fixes the resource, the controller is notified because it is watching said resource. This means that returning an error here doesn't help.

pkg/controller/applicationlayer/applicationlayer_controller.go
```diff
@@ -0,0 +1,319 @@
...
+
+	err = r.patchFelixTproxyMode(ctx, lcSpec)
+
+	if err != nil {
+		r.status.SetDegraded("Error patching felix configuration", err.Error())
+		return reconcile.Result{}, err
+	}
+
+	if r.enableL7LogsCollection(lcSpec) {
+
```
Brian-McM, line 208: Nit rogue space (you don't seem to space at the beginning of any of your other if statements).

pkg/controller/applicationlayer/applicationlayer_controller.go
```diff
@@ -0,0 +1,319 @@
...
+		return reconcile.Result{}, err
+	}
+
+	return reconcile.Result{}, nil
+}
+
+func validateApplicationLayer(al *operatorv1.ApplicationLayer) error {
+	lcSpec := al.Spec.LogCollection
+	if lcSpec != nil {
+		if *lcSpec.LogIntervalSeconds < 1 {
```
Brian-McM, line 250: @tmjd can't we get k8s to do this kind of validation by defining constraints in our CRDs, or am I mistaken?

pkg/controller/applicationlayer/applicationlayer_controller.go
```diff
@@ -0,0 +1,319 @@
...
+		}
+		if al.Spec.LogCollection.LogIntervalSeconds == nil {
+			al.Spec.LogCollection.LogIntervalSeconds = &defaultLogIntervalSeconds
+		}
+	}
+}
+
+// getApplicationLayer returns the default ApplicationLayer instance with defaults populated.
+func getApplicationLayer(ctx context.Context, cli client.Client) (*operatorv1.ApplicationLayer, error) {
+
```
Brian-McM, line 275: Nit, rogue space (none of you other functions have a space at the beginning).

Brian-McM (changes requested):

pkg/controller/applicationlayer/applicationlayer_controller.go
penkeysuresh: @Brian-McM We can, I just realised this, added it in this 971b9aa2427b8525d7bf163202d41274c57ff06f

pkg/controller/applicationlayer/applicationlayer_controller_test.go
```diff
@@ -0,0 +1,154 @@
...
+			proxy := ds.Spec.Template.Spec.Containers[0]
+			Expect(proxy).ToNot(BeNil())
+			Expect(proxy.Image).To(Equal(fmt.Sprintf("some.registry.org/%s:%s",
+				components.ComponentEnvoyProxy.Image, components.ComponentEnvoyProxy.Version)))
+
+			l7collector := ds.Spec.Template.Spec.Containers[1]
+			Expect(l7collector).ToNot(BeNil())
+			Expect(l7collector.Image).To(Equal(fmt.Sprintf("some.registry.org/%s:%s",
+				components.ComponentL7Collector.Image, components.ComponentL7Collector.Version)))
+		})
```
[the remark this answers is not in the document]
penkeysuresh: After thinking a bit more, If we explicitly set the `tproxyMode` to "Disabled" when the `LogCollection` is Disabled, user don't have to clear the L7 annotations they set on services. If we don't, then users are required to remove the annotations **and** set `LogCollection` to `Disabled` to properly turn off L7 proxy (if they don't remove the annotation, and set LogCollection to "Disabled" data plane will still try to proxy the service traffic to now non existent l7 proxy)  I think the former gives less chance for user to shoot themselves in the foot. 

This will get a bit messy when multiple features are using the `tproxyMode` though. But currently I'm explicitly turning the `tproxyMode` to "Disabled". Let me know if that makes sense or we can discuss more on slack.

pkg/render/applicationlayer/applicationlayer.go
```diff
@@ -0,0 +1,383 @@
...
+			},
+			Annotations: annots,
+		},
+		Spec: corev1.PodSpec{
+			HostIPC:            true,
+			HostNetwork:        true,
+			ServiceAccountName: APLName,
+			DNSPolicy:          corev1.DNSClusterFirstWithHostNet,
+			NodeSelector:       map[string]string{},
+			Tolerations:        c.tolerations(),
```
Brian-McM, line 169: I think you might be looking for this `rmeta.TolerateAll`.

Brian-McM: Great thanks for the comment.

pkg/controller/applicationlayer/applicationlayer_controller.go
```diff
@@ -0,0 +1,310 @@
...
+	return instance, nil
+}
+
+func (r *ReconcileApplicationLayer) isLogsCollectionEnabled(l7Spec *operatorv1.LogCollectionSpec) bool {
+	return l7Spec != nil && l7Spec.CollectLogs != nil && *l7Spec.CollectLogs == operatorv1.L7LogCollectionEnabled
+}
+
+// patchFelixTproxyMode takes all application layer specs as arguments and patches felix config.
+// If at least one of the specs requires TPROXYMode as "Enabled" it'll be patched as "Enabled" otherwise it is "Disabled".
+func (r *ReconcileApplicationLayer) patchFelixTproxyMode(ctx context.Context, l7Spec *operatorv1.LogCollectionSpec) error {
```
Brian-McM, line 279: I'm guessing this was discussed at a design level with Casey and Erik (so take my comments with a grain of salt), but I'm a little wary about patching FelixConfiguration. 

Is TPROXYMode something the user is allowed to have control of, or are we trying to take control away from them? If the user is allowed to have control over this, I would say we should set the status to degraded and fail if it's not set properly, and not subtly change this.

If we don't want the user to have any control over this setting then this is probably the right thing to do, although I can't really remember if we wanted non user configuration in this CR (I think we did so that the felix pod wouldn't restart on config changes, but there's a blurry line around this stuff).

pkg/render/applicationlayer/applicationlayer.go
```diff
@@ -0,0 +1,373 @@
...
+func (c *component) daemonset() *appsv1.DaemonSet {
+	maxUnavailable := intstr.FromInt(1)
+
+	annots := map[string]string{}
+
+	if c.envoyConfigMap != nil {
+		annots[EnvoyConfigMapName] = rmeta.AnnotationHash(c.envoyConfigMap)
+	}
+
+	podTemplate := &corev1.PodTemplateSpec{
```
Brian-McM, line 156: Nit (and there's nothing _really,_ wrong with this), but I don't think you need to create a reference for this object just to de-reference it.

pkg/render/applicationlayer/applicationlayer.go
```diff
@@ -0,0 +1,373 @@
...
+
+func (c *component) collectorVolMounts() []corev1.VolumeMount {
+	volumes := []corev1.VolumeMount{
+		{Name: EnvoyLogsVolumeName, MountPath: "/tmp/"},
+	}
+
+	return volumes
+}
+
+//go:embed envoy-config.yaml
```
Brian-McM, line 303: Woooww! This is great, I had know idea about this ability!

pkg/controller/applicationlayer/applicationlayer_controller.go
penkeysuresh: Users is not allowed to have control over `TPROXYMode`, so we are explicitly turning it on and off. User will still have control over turning the L7 logs on and off using the Application layer CRD (which in turn, turns this mode on and off along with creating/deleting the envoy daemon set).

pkg/controller/applicationlayer/applicationlayer_controller.go
```diff
@@ -0,0 +1,310 @@
...
+
+	return reconcile.Result{}, nil
+}
+
+// updateApplicationLayerWithDefaults populates the applicationlayer with defaults.
+func updateApplicationLayerWithDefaults(al *operatorv1.ApplicationLayer) {
+	defaultLogIntervalSeconds := int64(5)
+	defaultLogRequestsPerInterval := int64(-1)
+
+	if al.Spec.LogCollection != nil {
```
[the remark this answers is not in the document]
Brian-McM: The upside is that if somebody didn't add the Spec everything would just be defaulted and it wouldn't error out. Can this run purely on defaults?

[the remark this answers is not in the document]
Brian-McM: OK this makes sense then, thanks for explaining.

pkg/controller/applicationlayer/applicationlayer_controller.go
tmjd: Yeah I think that makes sense that we don't need to return an error.
@penkeysuresh if you're not aware when the return includes an error the request that triggered the reconcile will cause a re-queue of the request with an increasing backed off and the request is dropped eventually if it keeps erroring.
But like Brian said we should be able to depend on the watch on the Installation to trigger us.

So yeah lets not return err in this specific case.

pkg/controller/applicationlayer/applicationlayer_controller.go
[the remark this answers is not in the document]
tmjd: I'm not sure why we didn't include a State on the managementclusterconnection resource, and hence why it doesn't have a status check.

pkg/controller/applicationlayer/applicationlayer_controller.go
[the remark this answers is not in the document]
tmjd: I think you're suggesting we should not be disabling tProxyMode and I agree. But I believe the code currently would disable tProxyMode if `isLogsCollectionEnabled` returns false. So I think the code needs to be updated.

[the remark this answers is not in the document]
penkeysuresh: I actually gave out my reasoning on why explicitly disabling the tproxymode might be better here https://github.com/tigera/operator/pull/1505#discussion_r725327326, I made the change to disable it here following that rational, lmk if that makes sense.

tmjd (commented):
I've commented or mostly resolved on the previous comments. I still want to give it another look though but just wanted to submit what I could for now.

pkg/controller/applicationlayer/applicationlayer_controller.go
```diff
@@ -0,0 +1,310 @@
...
+		reqLogger.Error(err, "Error retrieving pull secrets")
+		r.status.SetDegraded("Error retrieving pull secrets", err.Error())
+		return reconcile.Result{}, err
+	}
+
+	lcSpec := applicationLayer.Spec.LogCollection
+
+	// If ApplicationLayer spec exists then LogCollection should be set.
+	// TODO: when we will have multiple features in future this should change to at least one feature being set
+	if lcSpec == nil {
```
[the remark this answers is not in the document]
penkeysuresh: fixed in cbafcfc481f74dee9957398282e0af0fa2051cdf

pkg/controller/applicationlayer/applicationlayer_controller.go
```diff
@@ -0,0 +1,343 @@
...
+		r.status.SetDegraded("Error reading ManagementClusterConnection", err.Error())
+		return reconcile.Result{}, err
+	}
+
+	if managementClusterConnection != nil && managementCluster != nil {
+		err = fmt.Errorf("having both a ManagementCluster and a ManagementClusterConnection is not supported")
+		reqLogger.Error(err, "")
+		r.status.SetDegraded(err.Error(), "")
+		return reconcile.Result{}, err
+	}
```
tmjd, line 211: I don't see these being used at all. Why do we need to get these?

pkg/controller/applicationlayer/applicationlayer_controller.go
```diff
@@ -0,0 +1,343 @@
...
+	err = r.patchFelixTproxyMode(ctx, applicationLayer)
+
+	if err != nil {
+		reqLogger.Error(err, "Error patching felix configuration")
+		r.status.SetDegraded("Error patching felix configuration", err.Error())
+		return reconcile.Result{}, err
+	}
+
+	lcSpec := applicationLayer.Spec.LogCollection
+	component := applicationlayer.ApplicationLayer(pullSecrets, installation, rmeta.OSTypeLinux,
```
tmjd, line 231: I think we should just be passing lcSpec directly into ApplicationLayer instead of passing 3 separate fields.

pkg/controller/applicationlayer/applicationlayer_controller.go
```diff
@@ -0,0 +1,343 @@
...
+	}
+
+	// Write the application layer back to the datastore, so the controllers depending on this can reconcile.
+	if err = r.client.Patch(ctx, applicationLayer, preDefaultPatchFrom); err != nil {
+		reqLogger.Error(err, "Failed to write defaults to applicationLayer")
+		r.status.SetDegraded("Failed to write defaults to applicationLayer", err.Error())
+		return reconcile.Result{}, err
+	}
+
+	variant, installation, err := utils.GetInstallation(ctx, r.client)
```
tmjd, line 179: We should be checking that variant is TigeraSecureEnterprise like https://github.com/tigera/operator/blob/master/pkg/controller/authentication/authentication_controller.go#L175

pkg/controller/applicationlayer/applicationlayer_controller.go
```diff
@@ -0,0 +1,343 @@
...
+	// If ApplicationLayer spec exists then LogCollection should be set.
+	// TODO: when we will have multiple features in future this should change to at least one feature being set
+	if lcSpec == nil {
+		return fmt.Errorf("missing required LogCollection spec in applicationLayer")
+	}
+
+	return nil
+}
+
+// getApplicationLayer returns the default ApplicationLayer instance with defaults populated.
```
tmjd, line 295: This comment is not accurate, we don't handle defaulting in this function.

pkg/render/applicationlayer/applicationlayer.go
```diff
@@ -0,0 +1,368 @@
...
+	return nil
+}
+
+func (c *component) SupportedOSType() rmeta.OSType {
+	return rmeta.OSTypeLinux
+}
+
+func (c *component) Objects() ([]client.Object, []client.Object) {
+	var objs []client.Object
+	// TODO: ensure that namespace exists before proceeding
```
tmjd, line 117: I don't think this TODO is necessary if we have the controller verifying that variant is set to TigeraSecureEnterprise in the controller.

pkg/render/applicationlayer/applicationlayer_test.go
```diff
@@ -0,0 +1,146 @@
...
+		}
+
+		collectorVolMounts := proxyContainer.VolumeMounts
+		expectedCollectorVolMounts := []corev1.VolumeMount{
+			{Name: applicationlayer.EnvoyLogsVolumeName, MountPath: "/tmp/"},
+		}
+		for _, expected := range expectedCollectorVolMounts {
+			Expect(collectorVolMounts).To(ContainElement(expected))
+		}
+	})
```
tmjd, line 144: I would like to see an additional test that sets the lcSpec to non-default values. You should only include checks for those specific values (no need to check everything that the above test does).

pkg/controller/applicationlayer/applicationlayer_controller.go
tmjd: @Brian-McM WDYT?
Pinging you after your comment here https://github.com/tigera/operator/pull/1505#discussion_r711209177

Brian-McM: I don't really have anything against passing the spec down, I just didn't think we needed another spec struct.

I also don't really have anything against pulling the fields out of the spec struct either, but maybe there should be a general config struct for all the parameters (instead of just a subset of them as there was before) in that case like we have for the some of the other components.

pkg/controller/applicationlayer/applicationlayer_controller.go
penkeysuresh: fixed in 6c8d0e4aaa28c9e011f5289a218584a313676c1c

pkg/controller/applicationlayer/applicationlayer_controller.go
penkeysuresh: after discussion on slack, we decided to pass the config as a separate struct 8ced1142ff48fde195b946f3466c202097b84902

pkg/controller/applicationlayer/applicationlayer_controller.go
penkeysuresh: fixed in 6c8d0e4

pkg/controller/applicationlayer/applicationlayer_controller_test.go
```diff
@@ -0,0 +1,152 @@
...
+			Expect(c.Create(ctx, &operatorv1.ApplicationLayer{
+				ObjectMeta: metav1.ObjectMeta{Name: "tigera-secure"},
+				Spec: operatorv1.ApplicationLayerSpec{
+					LogCollection: &operatorv1.LogCollectionSpec{
+						CollectLogs: &enabled,
+					},
+				},
+			})).NotTo(HaveOccurred())
+
+			_, err := r.Reconcile(ctx, reconcile.Request{})
```
tmjd, line 126: Please include a test to check that FelixConfiguration is being updated to enable tproxyMode.

pkg/controller/applicationlayer/applicationlayer_controller_test.go
```diff
@@ -0,0 +1,152 @@
...
+			Expect(proxy.Image).To(Equal(fmt.Sprintf("some.registry.org/%s:%s",
+				components.ComponentEnvoyProxy.Image, components.ComponentEnvoyProxy.Version)))
+
+			l7collector := ds.Spec.Template.Spec.Containers[1]
+			Expect(l7collector).ToNot(BeNil())
+			Expect(l7collector.Image).To(Equal(fmt.Sprintf("some.registry.org/%s:%s",
+				components.ComponentL7Collector.Image, components.ComponentL7Collector.Version)))
+		})
+
+	})
```
tmjd, line 151: Please include a Reconcile test that with no LogCollection resource the tproxyMode field will be disabled.

tmjd (commented):
One more test request. Otherwise I think we'll be good to go.

pkg/controller/applicationlayer/applicationlayer_controller_test.go
penkeysuresh: added tests an improved logging here 1a8e3f97e51178ba8e5c27e6c82fa6cc5d5b975d

PR #1506, opened 2021-09-08
Don't run so many Typha instances on small clusters

caseydavenport:
## Description

After some experience in the wild, we believe we don't need so many replicas on small clusters. This is largely because:
- during rolling update Typha will restart very quickly. 
- Running additional Typhas on very small clusters has additional resource impact 
- Running numTyphas == numNodes can also result in scheduling deadlocks during cluster scaling events, when the cluster autoscaler cannot find a suitable node to reschedule typha (since only one instance can run per node). 

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/installation/typha_autoscaler_test.go
+++ b/pkg/controller/installation/typha_autoscaler_test.go
@@ -160,12 +160,21 @@ var _ = Describe("Test typha autoscaler ", func() {
 		ta := newTyphaAutoscaler(c, nlw, tlw, statusManager, typhaAutoscalerPeriod(10*time.Millisecond))
 		ta.start(ctx)
 
+		// For clusters smaller than 3 nodes we only expect 1 replica.
+		verifyTyphaReplicas(c, 1)
+
+		// For three and four node clusters, we expect 2.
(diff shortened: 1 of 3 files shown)


pkg/controller/installation/typha_autoscaler_test.go
```diff
@@ -193,13 +202,15 @@ var _ = Describe("Test typha autoscaler ", func() {
 		// Create three nodes, one of which is not yet migrated
 		createNode(c, "node1", map[string]string{"kubernetes.io/os": "linux", "projectcalico.org/operator-node-migration": "migrated"})
 		createNode(c, "node2", map[string]string{"kubernetes.io/os": "linux", "projectcalico.org/operator-node-migration": "migrated"})
-		createNode(c, "node3", map[string]string{"kubernetes.io/os": "linux", "projectcalico.org/operator-node-migration": "pre-operator"})
+		createNode(c, "node3", map[string]string{"kubernetes.io/os": "linux", "projectcalico.org/operator-node-migration": "migrated"})
```
tmjd, line 205: Comment above is wrong now.

tmjd (approved):
One very minor change needed on a comment. Otherwise LGTM

PR #1507, opened 2021-09-08
Fix that control-plane tolerations were not applied to Typha pods

caseydavenport:
## Description

According to the API definition, `ControlPlaneTolerations` apply to all
pods deployed by the operator with the exception of daemonsets, which
means they should apply to Typha as well.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


tmjd (approved):
LGTM, I'm just hoping this doesn't bite us or someone when upgrading and picking up this change in behavior.
There was a test failure that looked related to Semaphore so I kicked CI again.

PR #1516, opened 2021-09-14
Create tigera-dpi namespace irrespective of DPI resource

Suraiya-Hameed:
## Description

Issue: Applying network policies fails during CaliEnt install as tigera-dpi namespace doesn't exist unit DeepPacketInspection resource is created.

This PR is to create a tigera-dpi namespace without depending on DeepPacketInspection resource.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/intrusiondetection/dpi/dpi.go
+++ b/pkg/render/intrusiondetection/dpi/dpi.go
@@ -84,26 +84,45 @@ func (d *dpiComponent) ResolveImages(is *operatorv1.ImageSet) error {
 }
 
 func (d *dpiComponent) Objects() (objsToCreate, objsToDelete []client.Object) {
-	var toDelete []client.Object
-	var toCreate []client.Object
+	var commonObjs []client.Object
+
+	nsObj := []client.Object{render.CreateNamespace(DeepPacketInspectionNamespace, d.cfg.Installation.KubernetesProvider)}
 
 	if d.cfg.HasNoDPIResource || d.cfg.HasNoLicense {
-		toDelete = append(toDelete, d.dpiNamespace())
-		return nil, toDelete
+		// create empty secrets and configMap when resource needs to be deleted.
+		commonObjs = append(commonObjs, &corev1.Secret{
+			TypeMeta:   metav1.TypeMeta{Kind: "Secret", APIVersion: "v1"},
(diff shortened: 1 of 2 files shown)


asincu (approved):

tmjd (changes requested):
I don't think this PR fixes what it is suppose to fix. I think to ensure that policies will be able to be created I think namespace always needs to be created.
I think there should be a test that with only IntrusiontDetection CR that the dpi namespace should be created. Right now as I read it, if there is no license or dpi resource then the namespace will not be created (and will be deleted in the no license case).
I think the render should always return the namespace as being created so the policy will always be able to be applied, even in the no license case.

Suraiya-Hameed:
@tmjd I've update the PR so dpi namespace is created even if there is no dpi resource. As discussed I haven't updated the condition where dpi namespace is deleted if there is no license as all other components are doing the same.

pkg/controller/intrusiondetection/intrusiondetection_controller.go
```diff
@@ -390,37 +390,35 @@ func (r *ReconcileIntrusionDetection) Reconcile(ctx context.Context, request rec
...
-			reqLogger.Error(err, fmt.Sprintf("Failed to retrieve %s secret", render.NodeTLSSecretName))
-			r.status.SetDegraded(fmt.Sprintf("Failed to retrieve %s secret", render.NodeTLSSecretName), err.Error())
-			return reconcile.Result{}, err
-		}
-		if nodeTLSSecret == nil {
-			reqLogger.Error(err, fmt.Sprintf("Waiting for %s secrets to be available", render.NodeTLSSecretName))
-			r.status.SetDegraded(fmt.Sprintf("Waiting for %s secrets to be available", render.NodeTLSSecretName), "")
-			return reconcile.Result{}, err
-		}
+	nodeTLSSecret, err = utils.GetSecret(ctx, r.client, render.NodeTLSSecretName, rmeta.OperatorNamespace())
```
tmjd, line 393: It shouldn't be necessary to get this and the following secrets and configmap if there is no DPI.
I see why you did that but it isn't necessary in the delete case for the render to have the full secrets and configmap (with data), when deleting it only needs the names and namespaces of the object to delete. WDYT about putting the controller code back to the way you had it and adjusting the render to create the empty secret and configmaps when it needs to delete the resources?

tmjd (approved):
LGTM

PR #1519, opened 2021-09-14
Remove es-gateway readiness probe

tmjd:
## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/logstorage/esgateway/esgateway.go
+++ b/pkg/render/logstorage/esgateway/esgateway.go
@@ -270,17 +270,7 @@ func (e esGateway) esGatewayDeployment() *appsv1.Deployment {
 					Image:        e.esGatewayImage,
 					Env:          envVars,
 					VolumeMounts: volumeMounts,
-					ReadinessProbe: &corev1.Probe{
-						Handler: corev1.Handler{
-							HTTPGet: &corev1.HTTPGetAction{
-								Path:   "/health",
-								Port:   intstr.FromInt(Port),
-								Scheme: corev1.URISchemeHTTPS,
-							},
-						},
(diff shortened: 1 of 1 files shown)


pkg/render/logstorage/esgateway/esgateway.go
```diff
@@ -270,17 +270,7 @@ func (e esGateway) esGatewayDeployment() *appsv1.Deployment {
...
-						Handler: corev1.Handler{
-							HTTPGet: &corev1.HTTPGetAction{
-								Path:   "/health",
-								Port:   intstr.FromInt(Port),
-								Scheme: corev1.URISchemeHTTPS,
-							},
-						},
-						InitialDelaySeconds: 10,
-						PeriodSeconds:       5,
-					},
```
stevegaossou, line 283: When we add this back in (after ES status and Kibana status are eliminated) we can add a value for `failureThreshold` so that it attempts multiple times before it deems es-gateway unhealthy. 

> failureThreshold: When a probe fails, Kubernetes will try failureThreshold times before giving up. Giving up in case of liveness probe means restarting the container. In case of readiness probe the Pod will be marked Unready. Defaults to 3. Minimum value is 1.

stevegaossou (approved):
LGTM. 

As discussed, we will remove the dependency on Elasticsearch status and Kibana status from es-gateway.

The only thing that you'll miss in the interim here is if kube API is not available. In that case, you won't know that es-gateway might misbehave (due to secrets being stale). But I think that scenario is rare, and removing the dependency of es-gateway health on Kibana and Elasticsearch APIs is more important.

ozdanborne (approved):

PR #1520, opened 2021-09-14
Make es admin username configurable for kube-controllers secret

vberezny:
## Description
The current assumption is that the `tigera-secure-es-elastic-user` secret contains an entry in the format `elastic: <password>` where `elastic` is the username of the admin level user. In the downstream Cloud Operator, specifically in the multi-tenant scenario, this username will not be `elastic` and will instead be `tigera-mgmt`. This change makes the username configurable for the kube-controllers Elasticsearch user secret, ensuring minimal conflicts downstream.

### DIFF
--- a/pkg/controller/logstorage/esgateway.go
+++ b/pkg/controller/logstorage/esgateway.go
@@ -47,7 +47,19 @@ func (r *ReconcileLogStorage) createEsGateway(
 		return reconcile.Result{}, false, nil
 	}
 
-	kubeControllersGatewaySecret, kubeControllersVerificationSecret, kubeControllersSecureUserSecret, err := common.CreateKubeControllersSecrets(ctx, esAdminUserSecret, r.client)
+	// This secret should only ever contain one key.
+	if len(esAdminUserSecret.Data) != 1 {
+		r.status.SetDegraded("Elasticsearch admin user secret contains too many entries", "")
+		return reconcile.Result{}, false, nil
+	}
+
+	var esAdminUserName string
+	for k := range esAdminUserSecret.Data {
+		esAdminUserName = k
+		break
+	}
+
+	kubeControllersGatewaySecret, kubeControllersVerificationSecret, kubeControllersSecureUserSecret, err := common.CreateKubeControllersSecrets(ctx, esAdminUserSecret, esAdminUserName, r.client)
 	if err != nil {
 		reqLogger.Error(err, err.Error())
 		r.status.SetDegraded("Failed to create kube-controllers secrets for Elasticsearch gateway", "")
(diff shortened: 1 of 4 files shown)


pkg/controller/logstorage/esgateway.go
```diff
@@ -47,7 +47,9 @@ func (r *ReconcileLogStorage) createEsGateway(
 		return reconcile.Result{}, false, nil
 	}
 
-	kubeControllersGatewaySecret, kubeControllersVerificationSecret, kubeControllersSecureUserSecret, err := common.CreateKubeControllersSecrets(ctx, esAdminUserSecret, r.client)
+	esAdminUserName := "elastic"
```
rene-dekker, line 50: Shouldn't the username come from the secret? Shouldn't there always be just one key in the secret, which is the username? (And the value being the password?)

vberezny: The key is `elastic` so the only way that I can think of to access the password is to do `esAdminUserSecret.Data["elastic"]`, however, for multi-tenancy, the key can't be `elastic` since that user is reserved for the cloud team and instead it will be `tigera-mgmt`. This is why I made this change so that I can instead go `esAdminUserSecret.Data[esAdminUserName]` and pass in whatever the key should be based on if the Operator is setting up a single or multi-tenant env.

vberezny: Actually I think I misunderstood what you're saying, I should just extract the key from the `esAdminUserSecret.Data` map and use that. I'll update the PR

rene-dekker: Yes, that is what I am thinking.

rene-dekker (approved):
lgtm

pkg/controller/logstorage/esgateway.go
```diff
@@ -47,7 +47,14 @@ func (r *ReconcileLogStorage) createEsGateway(
 		return reconcile.Result{}, false, nil
 	}
 
-	kubeControllersGatewaySecret, kubeControllersVerificationSecret, kubeControllersSecureUserSecret, err := common.CreateKubeControllersSecrets(ctx, esAdminUserSecret, r.client)
+	var esAdminUserName string
+	// This secret only ever contains one key.
```
tmjd, line 51: I think we should include a test to validate this assumption since we now depend on there only being a single key:value pair.

tmjd (changes requested):

pkg/controller/logstorage/esgateway.go
vberezny: I wasn't too sure how else to validate this assumption other than add a second key:value pair to a test es admin user secret and check to make sure that the correct (first) key:value pair is used as the username/password on the es-gateway deployment. Here is the change: https://github.com/tigera/operator/pull/1520/files#diff-569bac058c8112f0b2587062a0221cb55b557d6f071c32b4391354995c6571aeR452-R472

Let me know if that's what you had in mind @tmjd

tmjd: I was specifically thinking to include a check of `if len(esAdminUserSecret.Data) != 1` then Degrade here.

vberezny: I see, added the check

tmjd (approved):
LGTM

PR #1523, opened 2021-09-16
Add resource quota to components

doublek:
## Description

In GKE, resource quotas are set up in a way that pods with priority classes `system-node-critical` and `system-cluster-critical` are only schedulable if they are in the `kube-system` namespace. This PR adds resource quota for the `calico-system` and `tigera-fluentd` namespace, when deployed in GKE, which should unblock this.

A few things to note:
- I've only done this for GKE. I am not sure what impact this will have for other providers and wanted to minimize this.
- Rather than adding resource quotas from respective controllers and having resource quota as it's own component, it seemed to fit better being rendered in respective components and can be easily rendered in the right namespace.
- I didn't switch fluentd priority class to the earlier calico-priority mostly because I can't yet pick the right way to manage different priority classes. Should there be a single priority class for each component or should we pick a few priorities based on what each component does and assign things that way. 

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/node.go
+++ b/pkg/render/node.go
@@ -40,6 +40,7 @@ import (
 	"github.com/tigera/operator/pkg/render/common/configmap"
 	rmeta "github.com/tigera/operator/pkg/render/common/meta"
 	"github.com/tigera/operator/pkg/render/common/podsecuritypolicy"
+	"github.com/tigera/operator/pkg/render/common/resourcequota"
 	"github.com/tigera/operator/pkg/render/common/secret"
 )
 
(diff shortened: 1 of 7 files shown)


pkg/render/node.go
```diff
@@ -167,6 +168,14 @@ func (c *nodeComponent) Objects() ([]client.Object, []client.Object) {
 		c.nodeRoleBinding(),
 	}
 
+	if c.cfg.Installation.KubernetesProvider == operatorv1.ProviderGKE {
```
tmjd, line 171: I think this should go in the core_controller where it is coming up with the components that need to be created. It should go before https://github.com/tigera/operator/blob/master/pkg/controller/installation/core_controller.go#L1045.
You can probably use a "render.NewPassthrough"

doublek: Neat. I didn't know this existed. Will update the PR shortly. Thanks!

doublek:
@tmjd I've addressed your comment. PTAL.

tmjd (approved):
LGTM

PR #1532, opened 2021-09-23
Add support for Calico for Windows upgrades

lmm:
## Description

This PR adds a new controller to control the upgrade of Calico for Windows to the operator's supported enterprise version when the Linux nodes are upgraded to enterprise.

This PR works in conjunction with a new Windows node service that will help complete the upgrade.

**Tests are coming shortly** but the I think the PR is ready for some review as-is.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


pkg/common/windows.go
```diff
@@ -0,0 +1,67 @@
...
+
+const (
+	CalicoWindowsUpgradeResourceName = "calico-windows-upgrade"
+	CalicoWindowsUpgradeScript       = "calico-upgrade.ps1"
+	CalicoWindowsUpgradeVolumePath   = `c:\CalicoUpdate`
+	CalicoWindowsUpgradeScriptLabel  = "projectcalico.org/CalicoWindowsUpgradeScript"
+	CalicoWindowsVersionAnnotation   = "projectcalico.org/CalicoWindowsVersion"
+)
+
+func GetWindowsNodes(ctx context.Context, cli client.Client, filters ...func(node *corev1.Node) bool) ([]corev1.Node, error) {
```
tmjd, line 34: I only see this function being used in the windows controller, can it move there?

pkg/controller/status/status.go
```diff
@@ -778,3 +814,16 @@ func hasPendingCSRUsingCertV1beta1(ctx context.Context, cli client.Client, label
 	}
 	return false, nil
 }
+
+func hasPendingWindowsNodeUpgrade(ctx context.Context, cli client.Client, nodename, expectedVersion string) (bool, error) {
+	node := &corev1.Node{}
+	err := cli.Get(ctx, client.ObjectKey{Name: nodename}, node)
```
tmjd, line 820: I'm wondering if we should share the indexer for nodes (that I'm suggesting we add to the controller) with the status so it doesn't need to be querying for node information the controller will already have local access to.

tmjd (commented):
Some high level ideas of changes I think that should happen. I'll need to review this again for sure.

config/enterprise_versions.yml
```diff
@@ -91,6 +91,9 @@ components:
   deep-packet-inspection:
     image: tigera/deep-packet-inspection
     version: master
+  windows:
```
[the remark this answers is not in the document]
lmm: No, I fixed that now, we should have calico images too

pkg/controller/windows/windows_controller.go
```diff
@@ -0,0 +1,290 @@
...
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+
+		log.Info(fmt.Sprintf("Node %v has product=%v, version=%v", n.Name, currentProduct, currentVersion))
+		switch currentProduct {
+		case "Calico":
+			// If the node has Calico for Windows installed, we will upgrade to Enterprise version supported by
+			// this operator if the current Calico version is less-than/equal to the operator's supported Calico version.
+			// This prevents us from downgrading to an older Enterprise version.
```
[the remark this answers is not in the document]
lmm: I simplified this to just always install whatever version the operator supports, for the current installation variant.

pkg/controller/windows/windows_controller.go
```diff
@@ -0,0 +1,290 @@
...
+			return reconcile.Result{}, err
+		}
+
+		r.status.AddWindowsNodeUpgrade(n.Name, supportedUpgradeVersion)
+	}
+
+	// Query for the whole installation object, not just the spec. The
+	// Installation cr will own the resources of this component.
+	install := &operatorv1.Installation{}
+	if err := r.client.Get(ctx, utils.DefaultInstanceKey, install); err != nil {
```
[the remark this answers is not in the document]
lmm: Fixed

pkg/controller/windows/windows_controller.go
```diff
@@ -0,0 +1,290 @@
...
+		r.status.OnCRFound()
+
+		// Tell the status manager that we're ready to monitor the resources we've told it about and receive statuses.
+		r.status.ReadyToMonitor()
+
+	} else {
+		r.status.OnCRNotFound()
+	}
+
+	return reconcile.Result{RequeueAfter: 30 * time.Second}, nil
```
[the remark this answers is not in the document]
lmm: Fixed

pkg/controller/windows/windows_controller.go
```diff
@@ -0,0 +1,290 @@
...
+	clusterDomain string
+}
+
+func (r *reconcileWindows) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
+	reqLogger := log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
+	reqLogger.Info("Reconciling Windows")
+
+	// Get supported Windows nodes. Nodes are supported if they have a version
+	// annotation.
+	nodes, err := common.GetWindowsNodes(ctx, r.client, func(n *corev1.Node) bool {
```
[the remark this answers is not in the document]
lmm: Yep agreed. I don't think we need a dudup mechanism yet since we're filtering the node events we're going to watch, so that only the node events we care about will trigger a reconcile.

pkg/controller/windows/windows_controller.go
```diff
@@ -0,0 +1,290 @@
...
+
+	log.V(1).Info(fmt.Sprintf("Operator's supported Calico version=%v, Calico Enterprise version=%v", components.CalicoRelease, components.EnterpriseRelease))
+
+	// Get all Windows nodes that should get the upgrade triggered.
+	for _, n := range nodes {
+		log.V(1).Info(fmt.Sprintf("Checking node %v", n.Name))
+
+		// Skip the node if it's already running the supported Enterprise version.
+		if common.GetWindowsNodeVersion(&n) == supportedUpgradeVersion {
+			log.Info(fmt.Sprintf("Skipping upgrade of node %v, it already has the latest Enterprise version", n.Name))
```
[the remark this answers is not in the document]
lmm: Made most of them debug

pkg/common/windows.go
lmm: Removed this altogether as you mentioend earlier

pkg/controller/installation/core_controller.go
```diff
@@ -227,6 +235,12 @@ func add(mgr manager.Manager, r *ReconcileInstallation) error {
 		return fmt.Errorf("tigera-installation-controller failed to watch FelixConfiguration resource: %w", err)
 	}
 
+	// Watch for changes to Windows nodes.
+	err = utils.AddWindowsNodeWatch(c)
```
tmjd, line 239: I'm not a fan of this being another Node watch, I'm concerned with large clusters that might have lots of nodes. We already have the 1 watch for the typha Indexer.

tmjd: I think the only way to avoid this would be to have a trigger from the windows upgrade that can trigger the Reconcile.
This is where I think we could make use of the dedupe controller because then we could trigger a Reconcile by sending to the requestChan #1305

tmjd (commented):
I think we might have to make upgrading windows only functional for OS or Enterprise at 1st depending on the availablity of the windows-upgrade pod.

pkg/common/node.go
```diff
@@ -0,0 +1,132 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// GetWindowsNodes returns Windows nodes, optionally filtering the list of nodes
+// with the given filter functions.
+package common
```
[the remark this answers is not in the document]
lmm: Done: 515ccd278bc81e3d9f4a56164db8240cd89c4eb0

pkg/controller/installation/windows.go
```diff
@@ -0,0 +1,336 @@
...
+		return fmt.Errorf("Unable to remove label from node %v: %w", node.Name, err)
+	}
+
+	w.statusManager.AddWindowsNodeUpgrade(node.Name, expectedVersion)
+	return nil
+}
+
+func (w *calicoWindowsUpgrader) finishUpgrade(ctx context.Context, node *corev1.Node) error {
+	windowsLog.Info(fmt.Sprintf("Finishing upgrade on upgraded node %v", node.Name))
+	if err := common.RemoveNodeLabel(ctx, w.clientset, node.Name, common.CalicoWindowsUpgradeScriptLabel); err != nil {
```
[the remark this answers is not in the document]
lmm: Thanks, fixed e747996650398b908af65970aa40e839d49c118a

pkg/controller/installation/windows.go
```diff
@@ -0,0 +1,336 @@
...
+		if !ok {
+			return fmt.Errorf("Never expected index to have anything other than a Node object: %v", obj)
+		}
+
+		windowsLog.V(1).Info(fmt.Sprintf("Processing node %v", node.Name))
+		exists, version := common.GetWindowsNodeVersion(node)
+		// If the version annotation doesn't exist, the node does not support
+		// upgrades.
+		if !exists {
+			windowsLog.V(1).Info(fmt.Sprintf("Node %v doesn't have the version annotation, ignoring it", node.Name))
```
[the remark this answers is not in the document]
lmm: I think you're right, we should be degrading to let the user know something isn't right. Addressed with 0ec1e533ae38a24b4cc4bdff35700e0172a8dfe6

pkg/controller/installation/windows.go
```diff
@@ -0,0 +1,379 @@
...
+	expectedVersion := common.WindowsLatestVersionString(expectedProduct)
+	windowsLog.V(1).Info(fmt.Sprintf("Expected version: %v", expectedVersion))
+
+	// For now, do nothing if the upgrade version is not the upcoming
+	// Enterprise v3.11.0. This prevents a user from accidentally triggering an
+	// in-place upgrade from Calico v3.21.0 to Enterprise v3.10.x. That upgrade
+	// would fail since the calico windows upgrade is only supported in
+	// Enterprise v3.11.x and newer.
+	// TODO: remove this once Enterprise v3.11.0 is released.
+	if expectedProduct == operatorv1.TigeraSecureEnterprise && !strings.HasPrefix(expectedVersion, "Enterprise-v3.11.0") {
```
lmm, line 123: I still need to update this as discussed... we want to avoid doing an explicit version check here

pkg/controller/installation/windows.go
```diff
@@ -0,0 +1,336 @@
...
+	if err := removeTaint(ctx, w.clientset, node.Name, calicoWindowsUpgradingTaint); err != nil {
+		return fmt.Errorf("Unable to clear taint from node %v: %w", node.Name, err)
+	}
+
+	w.statusManager.RemoveWindowsNodeUpgrade(node.Name)
+	return nil
+}
+
+func (w *calicoWindowsUpgrader) processUpgrades(ctx context.Context, expectedVersion string) error {
+	for _, n := range w.nodesToUpgrade {
```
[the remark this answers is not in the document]
tmjd: Are you still planning on making some updates to how many upgrades we'll let run at a time? I forget where we landed with changing this.

pkg/controller/installation/core_controller.go
```diff
@@ -108,32 +108,43 @@ func newReconciler(mgr manager.Manager, opts options.AddOptions) (*ReconcileInst
...
 	nodeListWatch := cache.NewListWatchFromClient(cs.CoreV1().RESTClient(), "nodes", "", fields.Everything())
+
+	// Create a Typha autoscaler.
 	typhaListWatch := cache.NewListWatchFromClient(cs.AppsV1().RESTClient(), "deployments", "calico-system", fields.OneTermEqualSelector("metadata.name", "calico-typha"))
 	typhaScaler := newTyphaAutoscaler(cs, nodeListWatch, typhaListWatch, statusManager)
 
+	requestChan := make(chan utils.ReconcileRequest)
+
+	// Create a Calico Windows upgrader.
+	calicoWindowsUpgrader := newCalicoWindowsUpgrader(cs, mgr.GetClient(), nodeListWatch, statusManager, requestChan)
```
tmjd, line 126: I still think we should create the indexer outside of `newTyphaAutoscaler` and `newCalicoWindowsUpgrader`.
I think we could run a rather tight loop (like 1/sec) for windows and if any of the values 	w.nodesToUpgrade, w.nodesUpgrading, or w.nodesFinishedUpgrade changed then trigger a reconcile.
WDYT?

tmjd (dismissed):

pkg/controller/installation/core_controller.go
lmm: I started down this track with 13ab000ad193161caba41aee644be621681f5c80 ... maybe we can chat about this tomorrow a bit more.

pkg/controller/installation/windows.go
```diff
@@ -0,0 +1,648 @@
...
+	}
+
+	windowsLog.V(1).Info(fmt.Sprintf("new nodesToUpgrade=%v, nodesUpgrading=%v, nodesFinishedUpgrade=%v", len(currNodesToUpgrade), len(currNodesUpgrading), len(currNodesFinishedUpgrade)))
+
+	// Save copy of previous node upgrade status values.
+	prevNodesToUpgrade := make(map[string]*corev1.Node)
+	prevNodesUpgrading := make(map[string]*corev1.Node)
+	prevNodesFinishedUpgrade := make(map[string]*corev1.Node)
+
+	for k, v := range w.nodesToUpgrade {
```
tmjd, line 352: I think we should have locking to protect these shared variables.

pkg/controller/status/status.go
```diff
@@ -496,6 +540,14 @@ func (m *statusManager) syncState() {
 		}
 	}
 
+	for _, w := range m.windowsnodeupgrades {
```
tmjd, line 543: the syncState function this is in will be called every 5 seconds, if there are lots of nodes being upgraded all at once we'll be spamming the apiserver with Get'ing node. I guess this will only be a concern when upgrading from OS -> Ent.
Maybe we should not have the status manager do a separate Get of the Node and let the WindowsUpdate/Reconcile update `m.windowsnodeupgrades` and the status manager can use that state for status information. WDYT?

pkg/controller/installation/windows.go
lmm: Fixed in 5e8f813

pkg/controller/status/status.go
lmm: This is a good suggestion, addressed in c5fa45f3498b1adc2e8fcc363728fe2bcf755cf1

pkg/controller/installation/windows/calico_windows_upgrader.go
```diff
@@ -0,0 +1,405 @@
...
+		}
+
+		windowsLog.V(1).Info(fmt.Sprintf("Processing node %v", node.Name))
+		exists, variant, version := common.GetNodeVariantAndVersion(node)
+
+		// If the version annotation doesn't exist, something is wrong with the
+		// Calico Windows node or it needs to be upgraded manually to
+		// a version supported by the calicoWindowsUpgrader.
+		if !exists {
+			return fmt.Errorf("Node %v does not have the version annotation, it might be unhealthy or it might be running an unsupported Calico version.", node.Name), nil, nil, nil
```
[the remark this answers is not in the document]
lmm: That's a good point... the nodes will be reported ready by kubelet before the calico upgrade service on the node runs to set. And actually I think we are trying to work out a way to only install the windows upgrade service on the node if it's needed.

pkg/controller/installation/windows/calico_windows_upgrader.go
```diff
@@ -0,0 +1,405 @@
...
+		}
+
+		if variant != w.install.Variant || version != expectedVersion {
```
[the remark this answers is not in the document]
lmm: Fixed in 920b056d958ecf06e3244c8c718718fce4421369

pkg/controller/installation/windows/calico_windows_upgrader.go
```diff
@@ -0,0 +1,405 @@
...
+	numWindowsNodes := len(pending) + len(inprogress) + len(insync)
+	maxUnavailable, err := intstr.GetValueFromIntOrPercent(w.install.NodeUpdateStrategy.RollingUpdate.MaxUnavailable, numWindowsNodes, false)
+	if err != nil {
+		windowsLog.Error(err, "Invalid maxUnavailable value, falling back to default of 1")
+		maxUnavailable = defaultMaxUnavailable
+	}
+
+	for _, node := range pending {
+		// For upgrades from Calico -> Enterprise, we always upgrade regardless
+		// of maxUnavailable. For other upgrades, check that we have room
```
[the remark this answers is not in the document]
lmm: For OS -> EE we want to upgrade nodes as soon as possible to limit the time during which there may be EE policy that isn't enforced on the OS node. On some platforms, depending on the node upgrade strategy in place, there could be a large fraction of windows nodes suddenly being upgraded.

pkg/common/windows.go
```diff
@@ -0,0 +1,66 @@
...
+	CalicoWindowsUpgradeResourceName    = "calico-windows-upgrade"
+	CalicoWindowsUpgradeVolumePath      = `c:\CalicoUpgrade`
+	CalicoWindowsUpgradeLabel           = "projectcalico.org/windows-upgrade"
+	CalicoWindowsUpgradeLabelInProgress = "in-progress"
+	CalicoVersionAnnotation             = "projectcalico.org/version"
+	CalicoVariantAnnotation             = "projectcalico.org/variant"
+	CalicoWindowsUpgradeTaintKey        = "projectcalico.org/windows-upgrade"
+)
+
+func WindowsLatestVersionString(product operatorv1.ProductVariant) string {
```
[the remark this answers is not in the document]
lmm: Thanks, fixed in 7bda99f5f417087bc48c7116299dcc8a2ec30107

pkg/controller/installation/windows/calico_windows_upgrader.go
```diff
@@ -0,0 +1,405 @@
...
+func (w *calicoWindowsUpgrader) getExpectedVersion() string {
+	if w.install.Variant == operatorv1.TigeraSecureEnterprise {
+		return components.EnterpriseRelease
+	}
+	return components.CalicoRelease
+}
+
+func (w *calicoWindowsUpgrader) updateWindowsNodes() {
+	if w.install == nil {
+		windowsLog.V(1).Info("Skipping update since installation not set yet")
```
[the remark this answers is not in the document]
lmm: Removed by 875f1b9247bb11197501765d8c3896b0ecb57631 (though the `V(1)` will mean that the log is DEBUG)

pkg/controller/installation/windows/calico_windows_upgrader.go
```diff
@@ -0,0 +1,405 @@
...
+
+	w.statusManager.RemoveWindowsNodeUpgrade(node.Name)
+	return nil
+}
+
+// Start begins running the calicoWindowsUpgrader.
+func (w *calicoWindowsUpgrader) Start(ctx context.Context) {
+	go func() {
+		ticker := time.NewTicker(w.syncPeriod)
+		defer ticker.Stop()
```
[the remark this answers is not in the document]
lmm: Thanks this is much better 875f1b9247bb11197501765d8c3896b0ecb57631

pkg/controller/installation/windows/calico_windows_upgrader.go
```diff
@@ -0,0 +1,405 @@
...
+	return components.CalicoRelease
+}
+
+func (w *calicoWindowsUpgrader) updateWindowsNodes() {
+	if w.install == nil {
+		windowsLog.V(1).Info("Skipping update since installation not set yet")
+		return
+	}
+
+	err, pending, inprogress, insync := w.getNodeUpgradeStatus()
```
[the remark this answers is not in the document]
lmm: 👍 addressed in a7fd844b59c5b32652772ac7f6f632fa4bb1bc30

pkg/controller/installation/windows/calico_windows_upgrader.go
```diff
@@ -0,0 +1,405 @@
...
+}
+
+func (w *calicoWindowsUpgrader) startUpgrade(ctx context.Context, node *corev1.Node) error {
+	windowsLog.Info(fmt.Sprintf("Starting Calico Windows upgrade on node %v", node.Name))
+	if err := addTaint(ctx, w.clientset, node.Name, common.CalicoWindowsUpgradingTaint); err != nil {
+		return fmt.Errorf("Unable to add taint to node %v: %w", node.Name, err)
+	}
+
+	if err := nodeutils.AddNodeLabel(ctx, w.clientset, node.Name, common.CalicoWindowsUpgradeLabel, common.CalicoWindowsUpgradeLabelInProgress); err != nil {
+		return fmt.Errorf("Unable to add label to node %v: %w", node.Name, err)
```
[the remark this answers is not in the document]
lmm: Fixed in
4ab87fff5c1933d3c1a880e5b60177931d0d5a27
27150bd89eb9ac2c244e3f9d465d53dccf65ba3f

pkg/controller/migration/namespace_migration.go
```diff
@@ -480,6 +481,12 @@ func (m *CoreNamespaceMigration) ensureKubeSysNodeDaemonSetHasNodeSelectorAndIsR
 	})
 }
 
+type stringPatch struct {
```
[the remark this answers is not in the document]
lmm: Removed it

pkg/controller/status/status.go
```diff
@@ -275,6 +279,44 @@ func (m *statusManager) AddCertificateSigningRequests(name string, labels map[st
...
 
+type windowsNodeUpgrade struct {
+	nodeName        string
+	currentVersion  string
+	currentVariant  operator.ProductVariant
+	expectedVersion string
+	expectedVariant operator.ProductVariant
+}
+
+func (w *windowsNodeUpgrade) isPending(ctx context.Context, c client.Client) (bool, error) {
```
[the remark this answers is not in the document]
lmm: Removed

pkg/render/node.go
```diff
@@ -393,6 +393,14 @@ func (c *nodeComponent) nodeRole() *rbacv1.ClusterRole {
 				Resources: []string{"blockaffinities"},
 				Verbs:     []string{"watch"},
 			},
+			{
+				// calico-upgrade service needs to query the calico-node
```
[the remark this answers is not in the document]
lmm: I added the new RBAC to the calico-node role because the Calico for Windows installation uses the calico-node service account, and the new windows upgrade service uses it.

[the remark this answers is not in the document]
caseydavenport: So, unless I am missing some nuance, the windows upgrader daemonset should not be using the calico-node serviceaccount - it should have its own. The only reason might be if we use the serviceaccount credentials from the calico-for-windows upgrader for calico on the Windows nodes? I guess now that I think about it, that is probably the case.

Even so, we _probably_ should have our own serviceaccount and RBAC for the windows upgrader, even if it's 90% identical to calico/node. It's bad practice to provide unnecessary permissions to a service just because we didn't want to go through the effort of creating a separate service account.

pkg/render/windows.go
```diff
@@ -0,0 +1,178 @@
...
+		},
+	}
+
+	ds := &appsv1.DaemonSet{
+		TypeMeta: metav1.TypeMeta{Kind: "DaemonSet", APIVersion: "apps/v1"},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      common.CalicoWindowsUpgradeResourceName,
+			Namespace: "calico-system",
+		},
+		Spec: appsv1.DaemonSetSpec{
```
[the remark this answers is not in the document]
lmm: Fixed

pkg/controller/status/status.go
```diff
@@ -496,6 +546,11 @@ func (m *statusManager) syncState() {
 		}
 	}
 
+	for _, w := range m.windowsnodeupgrades {
+		if w.currentVersion != w.expectedVersion {
```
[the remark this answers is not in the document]
lmm: I refactored the status reporting in : 40fcafbf939d47496750477f38c4b71770761521

pkg/common/windows.go
```diff
@@ -0,0 +1,66 @@
...
+	operatorv1 "github.com/tigera/operator/api/v1"
+	"github.com/tigera/operator/pkg/components"
+	corev1 "k8s.io/api/core/v1"
+)
+
+var (
+	// This taint is applied to nodes upgrading Calico Windows.
+	CalicoWindowsUpgradingTaint = &corev1.Taint{
+		Key:    CalicoWindowsUpgradeTaintKey,
+		Effect: corev1.TaintEffectNoSchedule,
```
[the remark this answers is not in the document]
lmm: I think for now just preventing scheduling is enough. I chatted with Song about this and draining the node would be difficult to get right since the platform might also be actively evicting pods or setting taints on the node. We'll need to doc this as a limitation.

pkg/controller/utils/node/node.go
```diff
@@ -0,0 +1,140 @@
...
+		}
+
+		// no update needed
+		return true, nil
+	})
+}
+
+// CreateNodeSharedIndexInformer returns a Node shared index informer used
+// by the typhaAutoscaler and the calicoWindowsUpgrader.
+func CreateNodeSharedIndexInformer(nodeListWatch cache.ListerWatcher) cache.SharedIndexInformer {
```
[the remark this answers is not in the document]
lmm: Fixed

pkg/render/node.go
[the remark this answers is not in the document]
lmm: I agree we should have our own service account for Calico for Windows and we should fix this in the near future. After chatting with Song, we think we might not need this RBAC at all and can just verify the pod itself.

pkg/common/windows.go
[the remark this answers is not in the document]
caseydavenport: I don't really follow - do we need to evict pods in order to do this upgrade gracefully, or can it be done gracefully just by preventing scheduling?

What is the limitation you're referring to?

pkg/controller/installation/core_controller.go
```diff
@@ -109,33 +110,47 @@ func newReconciler(mgr manager.Manager, opts options.AddOptions) (*ReconcileInst
...
 	}
 
-	// Create a Typha autoscaler.
+	// Create the SharedIndexInformer used by the typhaAutoscaler and
+	// calicoWindowsUpgrader.
 	nodeListWatch := cache.NewListWatchFromClient(cs.CoreV1().RESTClient(), "nodes", "", fields.Everything())
+	nodeIndexInformer := cache.NewSharedIndexInformer(nodeListWatch, &corev1.Node{}, 0, cache.Indexers{})
+
+	go nodeIndexInformer.Run(opts.ShutdownContext.Done())
+	for nodeIndexInformer.HasSynced() {
```
caseydavenport, line 125: I think we should do this check within the typha autoscaler and windows upgrader, rather than block the entire reconcile here

caseydavenport: I don't think the rest of the core controller needs to depend on this syncing, right?

pkg/controller/installation/windows/calico_windows_upgrader.go
```diff
@@ -0,0 +1,414 @@
...
+		node, ok := obj.(*corev1.Node)
+		if !ok {
+			return nil, nil, nil, fmt.Errorf("Never expected index to have anything other than a Node object: %v", obj)
+		}
+
+		if node.Labels[corev1.LabelOSStable] != "windows" {
+			continue
+		}
+
+		windowsLog.V(1).Info(fmt.Sprintf("Processing node %v", node.Name))
```
caseydavenport, line 120: This log should probably be debug, otherwise it will get quite spammy.

pkg/controller/installation/windows/calico_windows_upgrader.go
```diff
@@ -0,0 +1,414 @@
...
+
+		// If the version annotation doesn't exist, something is wrong with the
+		// Calico Windows node or it needs to be upgraded manually to
+		// a version supported by the calicoWindowsUpgrader.
+		if !exists {
+			return nil, nil, nil, fmt.Errorf("Node %v does not have the version annotation, it might be unhealthy or it might be running an unsupported Calico version.", node.Name)
+		}
+
+		if node.Labels[common.CalicoWindowsUpgradeLabel] == common.CalicoWindowsUpgradeLabelInProgress {
+			windowsLog.V(1).Info(fmt.Sprintf("Node %v has the upgrade in-progress label", node.Name))
```
caseydavenport, line 131: Same with these logs - debug would be more appropriate. Info should only be used when we're taking a non-repetitive action so as to avoid spamming the log.

pkg/controller/installation/windows/calico_windows_upgrader.go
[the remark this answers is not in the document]
caseydavenport: I don't think I agree - we're going through a lot of effort to ensure this is a rolling update. If we just wanted to upgrade these as fast as possible, then we wouldn't be implementing this complex logic in the operator. The purpose of rolling update is to prevent rolling out a bad config, and also to ensure things like BGP graceful restart work properly  - if we ignore max unavailable, we don't get those.

Now, if you wanted to limit this _only_ to new nodes that come up after enterprise has been installed, that might be a different story. But certainly taking a cluster with 20 Windows nodes running Calico and turning those into 20 enterprise nodes should be done in a rolling fashion.

pkg/controller/installation/windows/calico_windows_upgrader.go
```diff
@@ -0,0 +1,414 @@
...
+			if err := w.startUpgrade(context.Background(), node); err != nil {
+				// Log the error and continue. We will retry when we update nodes again.
+				windowsLog.Info(fmt.Sprintf("Could not start upgrade on node %v: %v", node.Name, err))
+				continue
+			}
+			// Successfully started the upgrade. Moving the node
+			// from pending to in-progress.
+			inProgress[node.Name] = node
+			delete(pending, node.Name)
+			w.statusManager.AddWindowsNodeUpgrade(node.Name, true)
```
caseydavenport, line 219: I don't understand why we need this call - the only difference is that `inProgress` is now set to `true`, however, we've already started the upgrade from within `startUpgrade` called just a few lines above, and all we've done since then is change the contents of a local map. Can't we just get rid of this call?

pkg/controller/installation/windows/calico_windows_upgrader.go
```diff
@@ -0,0 +1,414 @@
...
+			if err := w.finishUpgrade(context.Background(), node); err != nil {
+				// Log the error and continue. We will retry when we update nodes again.
+				windowsLog.Info(fmt.Sprintf("Could not complete upgrade on node %v: %v", node.Name, err))
+				continue
+			}
+			// Successfully finished completing the upgrade. Moving the node
+			// from in-progress to in-sync.
+			inSync[node.Name] = node
+			delete(inProgress, node.Name)
+			w.statusManager.RemoveWindowsNodeUpgrade(node.Name)
```
caseydavenport, line 192: This is a duplicate call - we already call this directly from `finishUpgrade` so don't need to call it again here.

pkg/controller/installation/windows/calico_windows_upgrader.go
```diff
@@ -0,0 +1,414 @@
...
+		windowsLog.Error(err, "Failed to get Windows nodes upgrade status")
+		return
+	}
+
+	// Notify status manager of pending upgrades.
+	for _, node := range pending {
+		w.statusManager.AddWindowsNodeUpgrade(node.Name, false)
+	}
+
+	for _, node := range inProgress {
```
caseydavenport, line 181: Is it possible that a node will ever go from `pending` straight to `completed`? 

Only thing I can think of is if the node is manually upgraded through some other means, out of band of this controller. Seems pretty unlikely, but if it were to happen it would result in us calling `AddWindowsNodeUpgrade` and never calling `RemoveWindowsNodeUpgrade` within this loop (since the node skipped `inProgress`).

pkg/controller/migration/namespace_migration.go
```diff
@@ -615,107 +616,6 @@ func (m *CoreNamespaceMigration) getNumPodsDesiredAndReady(ctx context.Context,
 		nil
 }
 
-// addNodeLabels adds the specified labels to the named node. Perform
```
caseydavenport, line 618: I think we can put these back where they were since we're not longer using them. Let's minimize unnecessary changes in this set.

pkg/controller/status/status.go
```diff
@@ -275,6 +279,62 @@ func (m *statusManager) AddCertificateSigningRequests(name string, labels map[st
...
+	// If there are no more nodes upgrading or pending to be upgraded then clear
+	// the completed map.
+	if len(w.nodesPending) == 0 && len(w.nodesInProgress) == 0 {
+		w.nodesCompleted = make(map[string]struct{})
+	}
+}
+
+// AddWindowsNodeUpgrade tells the status manager to monitor the health of the given
+// Windows node upgrade.
+func (m *statusManager) AddWindowsNodeUpgrade(nodeName string, inProgress bool) {
```
caseydavenport, line 332: This API is confusing me a little bit. I think we could probably simplify it to a single call:

```go 
func SetWindowsUpgradeStatus(pending, inProgress, completed []string) {
}
```

We can do this because the calling code re-calculates these on every iteration anyway.

The main issues I see with the current approach is that it's unclear what `inProgress` means relative to `Add` and `Remove`, and that the status manager needs to attempt to reconstruct the caches that the controller is already building, which means they might get out of sync.

Sending all together like this is a bit easier to reason about. For example, it's not obvious that `removeNodeUpgrade` in the current API actually marks the node as completed. What happens, for example, if a node is deleted mid-upgrade? How will the status manager detect that and know what to do with its internal caches?

caseydavenport: The downside is that it's not fine-grained, so you need to call this at the end of every iteration of the reconcile loop rather than as you process each node. However, the reconcile loop never actually exits early (it just skips some nodes), so by the end of the function the `pending`, `inProgress`, and `completed` slices should be an accurate representation of the cluster.

pkg/controller/status/status.go
```diff
@@ -496,6 +564,11 @@ func (m *statusManager) syncState() {
 		}
 	}
 
+	if reason := m.windowsNodeUpgrades.progressingReason(); reason != "" {
+		progressing = append(progressing, reason)
+
```
caseydavenport, line 569: nit: extra newline here.

pkg/render/windows.go
```diff
@@ -0,0 +1,190 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	operatorv1 "github.com/tigera/operator/api/v1"
+	"github.com/tigera/operator/pkg/common"
+	"github.com/tigera/operator/pkg/components"
+	rmeta "github.com/tigera/operator/pkg/render/common/meta"
+)
+
+func Windows(
```
caseydavenport, line 30: Where is this function called? I don't see it in this PR.

pkg/render/windows.go
```diff
@@ -0,0 +1,190 @@
...
+	cr *operatorv1.InstallationSpec,
+	hasSupportedNodes bool,
```
caseydavenport, line 32: How are we going to calculate this variable? Is this just "Does the cluster have Windows nodes"? Or is it "Does the cluster have windows nodes that need to be upgraded"? Or something else?

lmm: The variable was used when we were still rendering the upgrade daemonset only if there were windows nodes to upgrade. I removed it since in an earlier review we had agreed that it probably wasn't worth the extra logic.

pkg/render/windows.go
lmm: I accidentally removed it while refactoring. Added back in: 0d1274149bfd5a03b7055f08ca455a383b1f1f79

pkg/controller/migration/namespace_migration.go
lmm: Fixed in 26191326aa3d8e0393d4f89f88c606ffee8f285a

pkg/controller/status/status.go
lmm: Refactored with 80324b4c4c535954326b59a6de69bf8b4cee03e1

I agree the API was a bit weird, I wanted to be consistent with the existing API of `AddXXX`s and `RemoveXXX`s, so the AddWindowsNodeUpgrade was tracking nodes pending and nodes in-progress. Doing the status reporting just once as you suggested makes some of the problems you noted elsewhere in this review go away (e.g. calling AddWindowsNodeUpgrade but not RemoveWindowsNodeUpgrade in some scenarios).

pkg/controller/installation/core_controller.go
lmm: Addressed with fd2289ef9ee93d5362ae06be843883f725fcdf91

pkg/controller/installation/windows/calico_windows_upgrader.go
lmm: I'll check again but I thought the default zap log level was info and so V(1) logs will be debug.

pkg/common/windows.go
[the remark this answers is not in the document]
lmm: > I don't really follow - do we need to evict pods in order to do this upgrade gracefully, or can it be done gracefully just by preventing scheduling?

Not from my testing so far, no I don't believe evicting the pods is needed. 

> What is the limitation you're referring to?

Sorry, I was jumping the gun here. I don't see any limitations yet but we're ramping up the testing.

pkg/controller/installation/windows/calico_windows_upgrader.go
caseydavenport: Oh yeah I forgot we're using zap in the operator.

pkg/controller/installation/windows/calico_windows_upgrader.go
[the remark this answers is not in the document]
caseydavenport: Discussed offline - going to leave this for now and fix in a future iteration. Definitely a bug, but one we're going to live with since the fix for the alternative is complex.

Makefile
```diff
@@ -651,7 +651,7 @@ fmt:
 vet:
 	$(CONTAINERIZED) $(CALICO_BUILD) \
 	sh -c '$(GIT_CONFIG_SSH) \
-	go vet ./...'
+	go vet -composites=false ./...'
```
caseydavenport, line 654: Why this change? We should just fix unkeyed composites in the code. Does this PR introduce any?

caseydavenport (commented):
Latest changes look good. I think @tmjd should give this one last look.

There are also one or two unresolved comments - not sure if you saw them. They probably aren't urgent, but worth you reviewing.

caseydavenport:
Also @lmm looks like tests are failing.

pkg/controller/installation/windows/calico_windows_upgrader.go
```diff
@@ -0,0 +1,438 @@
...
+func (w *calicoWindowsUpgrader) getExpectedVersion() string {
+	if w.install.Variant == operatorv1.TigeraSecureEnterprise {
+		return components.EnterpriseRelease
+	}
+	return components.CalicoRelease
+}
+
+func (w *calicoWindowsUpgrader) updateWindowsNodes() {
+	pending, inProgress, inSync, err := w.getNodeUpgradeStatus()
+	if err != nil {
```
tmjd, line 170: I think this error should be reflected in the status resource. I don't think it is ok to just log this error and continue on.

Makefile
lmm: Sorry I meant to revisit this, I added it to make things work. I had exported some structs used in our tests but I reverted the `go vet` change and refactored the test structs.

9325aadc84447b18cab320a30181f6a764cd2bd2

pkg/controller/installation/windows/calico_windows_upgrader.go
lmm: Fixed in 0cc494ff181599152ef9476ba3b661b0f591cfad

caseydavenport: Also, this error branch should be above - immediately below where the error is generated. Generally bad practice to put additional code between the error handling, since it can lead to unexpected bugs and makes the code harder to read.

pkg/controller/installation/windows/calico_windows_upgrader.go
```diff
@@ -0,0 +1,405 @@
...
+	// Get the total # of windows nodes we can have upgrading using the
+	// maxUnavailable value, if the node upgrade strategy was respected.
+	numWindowsNodes := len(pending) + len(inprogress) + len(insync)
+	maxUnavailable, err := intstr.GetValueFromIntOrPercent(w.install.NodeUpdateStrategy.RollingUpdate.MaxUnavailable, numWindowsNodes, false)
+	if err != nil {
+		windowsLog.Error(err, "Invalid maxUnavailable value, falling back to default of 1")
+		maxUnavailable = defaultMaxUnavailable
+	}
+
+	for _, node := range pending {
```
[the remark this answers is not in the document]
lmm: @caseydavenport Addressed this with 695c4dfb, wdyt? I think it is better to sort once at the end instead of keeping the slice sorted throughout

pkg/controller/installation/windows/calico_windows_upgrader.go
```diff
@@ -0,0 +1,456 @@
...
+}
+
+func (w *calicoWindowsUpgrader) updateWindowsNodes() {
+	pending, inProgress, inSync, err := w.getNodeUpgradeStatus()
+
+	// Make sorted slices of the node names from the node upgrade status maps.
+	// We'll loop through these slices to process node upgrades
+	// deterministically.
+	pendingNodeNames := sortedSliceFromMap(pending)
+	inProgressNodeNames := sortedSliceFromMap(inProgress)
```
caseydavenport, line 186: You need to move this logic to just before where we use it - the sorted slice isn't going to be accurate, because we modify the maps as part of the loop below.

For example, we move entries from in-progress to in-sync in the first loop.

caseydavenport: I'd probably even just recommend doing it like this:

```go
for _, nodeName := range sortedSliceFromMap(inProgress) {
}
```

lmm: fixed in 8dad23754491bd4fa5e4513d2b7d06becb7e2f0a

pkg/controller/installation/windows/calico_windows_upgrader.go
lmm: We discussed offline and have a plan to address the status handling properly in a follow PR.

PR #1535, opened 2021-09-24
Update versions for v1.22

Brian-McM:
## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/config/enterprise_versions.yml
+++ b/config/enterprise_versions.yml
@@ -2,95 +2,95 @@ title: v3.10.0
 components:
   cnx-manager:
     image: tigera/cnx-manager
-    version: release-calient-v3.10
+    version: v3.10.0
   voltron:
     image: tigera/voltron
-    version: release-calient-v3.10
+    version: v3.10.0
   cnx-apiserver:
     image: tigera/cnx-apiserver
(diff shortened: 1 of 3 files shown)


doublek (approved):
LGTM

stevehipwell:
@Brian-McM I'm interested as to why the operator v1.21 and v1.22 releases are both targeting Calico v3.20.0 and the operator v1.20 release is targeting the newer Calico v3.20.1?

Brian-McM:
@stevehipwell I think it was just a combination of the Calico patch coming out after the release (for v1.21) and me not updating to use the patch release after I'd created the release branch (for v1.22).

There's a patch release that was created for v1.21 after Calico 3.20.1 but it may have been targeting either an operator specific fix or a Calico Enterprise fix, in which case the version of Calico may not have been considered.

stevehipwell:
@Brian-McM is it worth patching v1.21 and v1.22 so they at least have the same version as the latest v1.20 release?

For completeness and to limit unwanted churn when upgrading the operator I'd suggest that this should be done. But I bet there are other considerations against this?

PR #1544, opened 2021-09-29
Add option for calico-node to run as non privileged and non root

mgleung:
## Description

Adds all of the configuration changes required to run calico-node as non-root and non-privileged.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/api/v1/installation_types.go
+++ b/api/v1/installation_types.go
@@ -141,6 +141,10 @@ type InstallationSpec struct {
 	// pods will be stuck during initialization.
 	// +optional
 	CertificateManagement *CertificateManagement `json:"certificateManagement,omitempty"`
+
+	// NonPrivileged configures Calico to be run in non privileged containers as non root users where possible.
+	// +optional
+	NonPrivileged *NonPrivilegedType `json:"nonPrivileged,omitempty"`
 }
 
 // TyphaAffinity allows configuration of node affinitiy characteristics for Typha pods.
@@ -215,6 +219,16 @@ type ProductVariant string
 var (
 	Calico                 ProductVariant = "Calico"
 	TigeraSecureEnterprise ProductVariant = "TigeraSecureEnterprise"
+)
+
+// NonPrivilegedType specifies whether Calico runs as permissioned or not
+//
+// One of: Enabled, Disabled
+type NonPrivilegedType string
+
+const (
+	NonPrivilegedEnabled  NonPrivilegedType = "Enabled"
+	NonPrivilegedDisabled NonPrivilegedType = "Disabled"
 )
 
 // ContainerIPForwardingType specifies whether the CNI config for container ip forwarding is enabled.
--- a/pkg/render/node.go
+++ b/pkg/render/node.go
@@ -641,6 +641,10 @@ func (c *nodeComponent) nodeDaemonset(cniCfgMap *corev1.ConfigMap) *appsv1.Daemo
 		initContainers = append(initContainers, c.bpffsInitContainer())
 	}
 
+	if c.runAsNonPrivileged() {
+		initContainers = append(initContainers, c.hostPathInitContainer())
+	}
+
 	var affinity *corev1.Affinity
 	if c.cfg.Installation.KubernetesProvider == operatorv1.ProviderAKS {
 		affinity = &corev1.Affinity{
@@ -734,8 +738,6 @@ func (c *nodeComponent) nodeVolumes() []corev1.Volume {
 
 	volumes := []corev1.Volume{
 		{Name: "lib-modules", VolumeSource: corev1.VolumeSource{HostPath: &corev1.HostPathVolumeSource{Path: "/lib/modules"}}},
-		{Name: "var-run-calico", VolumeSource: corev1.VolumeSource{HostPath: &corev1.HostPathVolumeSource{Path: "/var/run/calico"}}},
-		{Name: "var-lib-calico", VolumeSource: corev1.VolumeSource{HostPath: &corev1.HostPathVolumeSource{Path: "/var/lib/calico"}}},
 		{Name: "xtables-lock", VolumeSource: corev1.VolumeSource{HostPath: &corev1.HostPathVolumeSource{Path: "/run/xtables.lock", Type: &fileOrCreate}}},
 		{Name: "policysync", VolumeSource: corev1.VolumeSource{HostPath: &corev1.HostPathVolumeSource{Path: "/var/run/nodeagent", Type: &dirOrCreate}}},
 		{
@@ -752,6 +754,18 @@ func (c *nodeComponent) nodeVolumes() []corev1.Volume {
 			Name:         "felix-certs",
 			VolumeSource: certificateVolumeSource(c.cfg.Installation.CertificateManagement, NodeTLSSecretName),
 		},
+	}
+
+	if c.runAsNonPrivileged() {
+		volumes = append(volumes,
+			corev1.Volume{Name: "var-run", VolumeSource: corev1.VolumeSource{HostPath: &corev1.HostPathVolumeSource{Path: "/var/run"}}},
+			corev1.Volume{Name: "var-lib", VolumeSource: corev1.VolumeSource{HostPath: &corev1.HostPathVolumeSource{Path: "/var/lib"}}},
+		)
+	} else {
+		volumes = append(volumes,
+			corev1.Volume{Name: "var-run-calico", VolumeSource: corev1.VolumeSource{HostPath: &corev1.HostPathVolumeSource{Path: "/var/run/calico"}}},
+			corev1.Volume{Name: "var-lib-calico", VolumeSource: corev1.VolumeSource{HostPath: &corev1.HostPathVolumeSource{Path: "/var/lib/calico"}}},
+		)
 	}
 
 	if c.bpfDataplaneEnabled() {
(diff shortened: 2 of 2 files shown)


pkg/render/node.go
```diff
@@ -1433,6 +1472,42 @@ func (c *nodeComponent) nodePodSecurityPolicy() *policyv1beta1.PodSecurityPolicy
...
+		{
+			MountPath: "/var/lib",
+			Name:      "var-lib",
+			ReadOnly:  false,
+		},
+	}
+
+	return corev1.Container{
+		Name:         "hostpath-init",
+		Image:        "busybox:latest",
```
lmm, line 1494: I'm not sure but I think this image should go through `ResolveImages` so that the user can use their own image registry. But that might mean that we need to put this image under the calico "namespace" like `quay.io/calico/busybox`.

Also can we use a specific image tag so that we can control when the version is bumped?

api/v1/installation_types.go
```diff
@@ -141,6 +141,10 @@ type InstallationSpec struct {
 	// pods will be stuck during initialization.
 	// +optional
 	CertificateManagement *CertificateManagement `json:"certificateManagement,omitempty"`
+
+	// NonPrivileged configures Calico to be run in non privileged containers as non root users where possible.
+	// +optional
+	NonPrivileged *NonPrivilegedType `json:"nonPrivileged,omitempty"`
```
lmm, line 147: Does this need to be a pointer? If not, then it would simplify the code elsewhere (e.g. wouldn't need to check if this was nil).

pkg/render/node.go
```diff
@@ -939,11 +953,27 @@ func (c *nodeComponent) cniEnvvars() []corev1.EnvVar {
 // nodeContainer creates the main node container.
 func (c *nodeComponent) nodeContainer() corev1.Container {
 	lp, rp := c.nodeLivenessReadinessProbes()
+	sc := &corev1.SecurityContext{Privileged: ptr.BoolToPtr(true)}
+	if c.runAsNonPrivileged() {
+		uid := int64(1000)
+		guid := int64(0)
```
lmm, line 959: Could we add a comment about the group ID and why the root group is needed?

pkg/render/node.go
```diff
@@ -1433,6 +1472,42 @@ func (c *nodeComponent) nodePodSecurityPolicy() *policyv1beta1.PodSecurityPolicy
...
+		Command: []string{"sh", "-c", "mkdir -p /var/lib/calico/ && chown -R 1000 /var/lib/calico/ && mkdir -p /var/run/calico/ && chown -R 1000 /var/run/calico/"},
+	}
+}
+
+// runAsNonPrivileged checks to ensure that all of the proper installation values are set for running
+// Calico as non privileged.
+func (c *nodeComponent) runAsNonPrivileged() bool {
+	return (c.cfg.Installation.NonPrivileged != nil && *c.cfg.Installation.NonPrivileged == operatorv1.NonPrivilegedEnabled) &&
+		!(c.cfg.Installation.CalicoNetwork != nil && *c.cfg.Installation.CalicoNetwork.BGP == operatorv1.BGPEnabled) &&
+		c.cfg.Installation.Variant == operatorv1.Calico
```
lmm, line 1508: Can we please consider breaking this up into separate if statements? As a one liner it's hard to understand. (It might compile down to the same thing as this one-liner anyways.)

lmm (commented):
I took a quick look and added some comments. It might be worth getting a review from someone who was involved in the design.

pkg/render/node.go
lmm: I think we need tests to validate the rendering

pkg/render/node.go
mgleung: You're right I need to see what images we have available for this. I think we should have something we maintain for small tasks like this, but if not, I can bake something into calico-node and have that do these commands for me.

api/v1/installation_types.go
mgleung: I tried to keep this as a pointer so that a `nil` value is valid so that we can have the value omitted properly if it is empty.

pkg/render/node.go
mgleung: This should be resolved with https://github.com/projectcalico/node/pull/1238

pkg/render/node.go
```diff
@@ -1433,6 +1474,52 @@ func (c *nodeComponent) nodePodSecurityPolicy() *policyv1beta1.PodSecurityPolicy
...
+// Calico as non privileged.
+func (c *nodeComponent) runAsNonPrivileged() bool {
+	// Check that the NonPrivileged flag is set
+	nonPrivileged := c.cfg.Installation.NonPrivileged != nil && *c.cfg.Installation.NonPrivileged == operatorv1.NonPrivilegedEnabled
+
+	// Check that BGP is not enabled
+	noBGP := !bgpEnabled(c.cfg.Installation)
+
+	// Only allowed to run as non privileged for OS Calico
+	variantCalico := c.cfg.Installation.Variant != operatorv1.TigeraSecureEnterprise
```
tmjd, line 1518: This check should not be needed here. There should be validation done by the core_controller, the validation code should be added to https://github.com/tigera/operator/blob/master/pkg/controller/installation/validation.go#L30.
The result of specifying NonPrivileged and Enterprise should be a degraded status on the calico tigerastatus (which is where a validation failure that I linked to would show up).

pkg/render/node_test.go
```diff
@@ -752,6 +754,261 @@ var _ = Describe("Node rendering tests", func() {
...
+		i := 0
+		for _, expectedRes := range expectedResources {
+			rtest.ExpectResource(resources[i], expectedRes.name, expectedRes.ns, expectedRes.group, expectedRes.version, expectedRes.kind)
+			i++
+		}
+
+		cniCmResource := rtest.GetResource(resources, "cni-config", "calico-system", "", "v1", "ConfigMap")
+		Expect(cniCmResource).ToNot(BeNil())
+		cniCm := cniCmResource.(*corev1.ConfigMap)
+		Expect(cniCm.Data["config"]).To(MatchJSON(`{
```
tmjd, line 793: I don't see any specific config that is applicable for NonPrivileged/non root, so I don't think you should include cni config checking.

tmjd: In general I'd suggest removing most of the testing that isn't related to NonPrivilege/non-root, it makes future updates easier.

pkg/render/node.go
```diff
@@ -939,11 +953,29 @@ func (c *nodeComponent) cniEnvvars() []corev1.EnvVar {
 // nodeContainer creates the main node container.
 func (c *nodeComponent) nodeContainer() corev1.Container {
 	lp, rp := c.nodeLivenessReadinessProbes()
+	sc := &corev1.SecurityContext{Privileged: ptr.BoolToPtr(true)}
+	if c.runAsNonPrivileged() {
+		uid := int64(1000)
```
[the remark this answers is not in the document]
mgleung: I just randomly picked one. There's no rhyme or reason to it really.

pkg/controller/installation/validation.go
```diff
@@ -330,6 +330,19 @@ func validateCustomResource(instance *operatorv1.Installation) error {
...
+	// Verify that we are running in non privileged mode only with the appropriate feature set
+	if instance.Spec.NonPrivileged != nil && *instance.Spec.NonPrivileged == operatorv1.NonPrivilegedEnabled {
+		// BGP must be disabled
+		if instance.Spec.CalicoNetwork != nil && instance.Spec.CalicoNetwork.BGP != nil && *instance.Spec.CalicoNetwork.BGP == operatorv1.BGPEnabled {
+			return fmt.Errorf("Installation non privileged Calico is not supported when BGP is enabled")
+		}
+
+		// Only allowed to run as non privileged for OS Calico
+		if instance.Spec.Variant == operatorv1.TigeraSecureEnterprise {
+			return fmt.Errorf("Installation non privileged Calico is not supported for spec.Variant=%s", operatorv1.TigeraSecureEnterprise)
```
tmjd: I think this might need an `of`
`Installation of non privileged Calico is not supported for spec.Variant`

pkg/render/node.go
```diff
@@ -734,8 +738,6 @@ func (c *nodeComponent) nodeVolumes() []corev1.Volume {
 
 	volumes := []corev1.Volume{
 		{Name: "lib-modules", VolumeSource: corev1.VolumeSource{HostPath: &corev1.HostPathVolumeSource{Path: "/lib/modules"}}},
-		{Name: "var-run-calico", VolumeSource: corev1.VolumeSource{HostPath: &corev1.HostPathVolumeSource{Path: "/var/run/calico"}}},
-		{Name: "var-lib-calico", VolumeSource: corev1.VolumeSource{HostPath: &corev1.HostPathVolumeSource{Path: "/var/lib/calico"}}},
 		{Name: "xtables-lock", VolumeSource: corev1.VolumeSource{HostPath: &corev1.HostPathVolumeSource{Path: "/run/xtables.lock", Type: &fileOrCreate}}},
 		{Name: "policysync", VolumeSource: corev1.VolumeSource{HostPath: &corev1.HostPathVolumeSource{Path: "/var/run/nodeagent", Type: &dirOrCreate}}},
```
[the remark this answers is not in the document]
tmjd: So does that mean `/var/run/nodeagent` should be removed here and added below?

pkg/render/node.go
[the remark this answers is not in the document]
tmjd: This does mean that the 1000 uid would have access on the host to the calico-node files.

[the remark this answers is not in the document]
mgleung: I changed the UID so that it would not overlap with the UID range for human users.

pkg/controller/installation/validation.go
```diff
@@ -330,6 +330,19 @@ func validateCustomResource(instance *operatorv1.Installation) error {
 		}
 	}
 
+	// Verify that we are running in non-privileged mode only with the appropriate feature set
+	if instance.Spec.NonPrivileged != nil && *instance.Spec.NonPrivileged == operatorv1.NonPrivilegedEnabled {
+		// BGP must be disabled
+		if instance.Spec.CalicoNetwork != nil && instance.Spec.CalicoNetwork.BGP != nil && *instance.Spec.CalicoNetwork.BGP == operatorv1.BGPEnabled {
```
caseydavenport, line 336: Is this the only validation we need? What about say, eBPF mode?

api/v1/installation_types.go
mgleung: The `runAsNonPrivileged` function should treat any `nil` value as it being set to `Disabled` during the rendering. Not sure if we should explicitly set it instead of omitting the field when empty though.

pkg/controller/installation/validation.go
```diff
@@ -330,6 +330,21 @@ func validateCustomResource(instance *operatorv1.Installation) error {
 		}
 	}
 
+	// Verify that we are running in non-privileged mode only with the appropriate feature set
+	if instance.Spec.NonPrivileged != nil && *instance.Spec.NonPrivileged == operatorv1.NonPrivilegedEnabled {
+		// BPF must be disabled
+		if instance.Spec.CalicoNetwork != nil &&
```
tmjd, line 336: Could you include UTs for these validation cases? (see validation_tests.go)

tmjd (approved):
LGTM, though before you merge please squash the changes.

PR #1545, opened 2021-10-01
Get internal elasticsearch secret from proper namespace

Brian-McM:
The internal es secret is no longer in the operator namespace since other components don't use it (they use the gateway secret). The only secret that needs it now is the gateway.

This also fixes a bug where we're not checking if the kibana or elasticsearch internal secrets are nil, meaning they don't exist (because GetSecret returns a nil secret with no error if the secret doesn't exist).

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [x] Appropriate labels:

### DIFF
--- a/pkg/controller/logstorage/esgateway.go
+++ b/pkg/controller/logstorage/esgateway.go
@@ -35,15 +35,23 @@ func (r *ReconcileLogStorage) createEsGateway(
 
 	kibanaInternalCertSecret, err := utils.GetSecret(ctx, r.client, render.KibanaInternalCertSecret, common.OperatorNamespace())
 	if err != nil {
+		reqLogger.Error(err, "failed to get Kibana tls certificate secret")
+		r.status.SetDegraded("Failed to get Kibana tls certificate secret", err.Error())
+		return reconcile.Result{}, false, err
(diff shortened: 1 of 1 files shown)


pkg/controller/logstorage/esgateway.go
```diff
@@ -35,15 +35,23 @@ func (r *ReconcileLogStorage) createEsGateway(
...
 		reqLogger.Error(err, err.Error())
 		r.status.SetDegraded("Waiting for internal Kibana tls certificate secret to be available", "")
 		return reconcile.Result{}, false, nil
 	}
 
-	esInternalCertSecret, err := utils.GetSecret(ctx, r.client, relasticsearch.InternalCertSecret, common.OperatorNamespace())
+	esInternalCertSecret, err := utils.GetSecret(ctx, r.client, relasticsearch.InternalCertSecret, render.ElasticsearchNamespace)
 	if err != nil {
+		reqLogger.Error(err, "failed to get Kibana tls certificate secret")
+		r.status.SetDegraded("Failed to get Kibana tls certificate secret", err.Error())
```
doublek, line 50: Error message should be updated to state Elasticsearch.

doublek (approved):
LGTM

PR #1548, opened 2021-10-04
[release-v1.22] Auto pick #1545: Get internal elasticsearch secret from proper namespace

Brian-McM:
Cherry pick of #1545 on release-v1.22.

#1545: Get internal elasticsearch secret from proper namespace


doublek (approved):
LGTM

PR #1558, opened 2021-10-08
Add feature to manage CRDs

tmjd:
## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


main.go
```diff
@@ -92,8 +97,14 @@ func main() {
 		"Show version information")
 	flag.StringVar(&printImages, "print-images", "",
 		"Print the default images the operator could deploy and exit. Possible values: list")
+	flag.StringVar(&printCalicoCRDs, "print-calico-crds", "",
```
Brian-McM, line 100: Could you just drop requiring "all" if you want to print all, and just call this function with no parameters? I think doing that should result in the same effect as calling all since it will check if the file has a prefix of nothing, which is always true.

pkg/crds/crds.go
```diff
@@ -0,0 +1,192 @@
...
+		}
+
+		ret[entry.Name()] = yamlDelimRe.ReplaceAll(b, []byte("\n"))
+	}
+
+	return ret
+}
+
+func GetCRDs(variant opv1.ProductVariant) []*apiextenv1.CustomResourceDefinition {
+	lock.Lock()
```
Brian-McM, line 134: Why do you need a lock? What's the race you're trying to prevent?

pkg/crds/crds.go
```diff
@@ -0,0 +1,192 @@
...
+	defer lock.Unlock()
+
+	if variant == opv1.Calico {
+		if len(calicoCRDs) == 0 {
+			crdyamls := getCalicoCRDSource()
+			for _, yml := range crdyamls {
+
```
Brian-McM, line 141: Nit, Rogue empty line

pkg/controller/installation/core_controller.go
```diff
@@ -1692,3 +1721,28 @@ func cidrWithinCidr(cidr, pool string) bool {
...
+				// Don't skip the removal of the "ignore" annotation. We want to
+				// reconcile when that happens.
+				return true
+			}
+			// Otherwise, ignore updates to objects when metadata.Generation does not change.
+			return e.ObjectOld.GetGeneration() != e.ObjectNew.GetGeneration()
+		},
+		DeleteFunc: func(e event.DeleteEvent) bool { return true },
+	}
+	for _, x := range crds.GetCRDs(v) {
```
Brian-McM, line 1742: Maybe this doesn't matter, but I wonder if we should not return the full CRD definitions just the get the names. This leaves a potentially large, albeit it very temporary, memory footprint since we're storing the full CRD definitions in code. Having a subsequent `GetCRDNames` function that iterates through the CRDs and pulls out the names while only ever storing one CRD at a time might be a little less costly.

Brian-McM: You could even have a custom struct instead of the full `&apiextenv1.CustomResourceDefinition{}` resource that only has room for the meta data so that not even the full CRD is parsed (which will alleviate initial CPU usage).

pkg/crds/crds.go
```diff
@@ -0,0 +1,192 @@
...
+				err := yaml.Unmarshal(yml, crd)
+				if err != nil {
+					fmt.Println(fmt.Sprintf("%s.%s", crd.Spec.Names.Plural, crd.Spec.Group))
+					fmt.Println(err)
+					return calicoCRDs
+				}
+				crd.Name = fmt.Sprintf("%s.%s", crd.Spec.Names.Plural, crd.Spec.Group)
+				calicoCRDs = append(calicoCRDs, crd)
+			}
+			crdyamls = getOperatorCRDSource(variant)
```
Brian-McM, line 152: This seems to be a somewhat duplicate block of code that might be worth pulling out into a separate function.

pkg/crds/crds.go
```diff
@@ -0,0 +1,192 @@
...
+				crd.Name = fmt.Sprintf("%s.%s", crd.Spec.Names.Plural, crd.Spec.Group)
+				enterpriseCRDs = append(enterpriseCRDs, crd)
+			}
+		}
+		return enterpriseCRDs
+	}
+}
+
+// ToRuntimeObjects converts the given list of CRDs to a list of client.Objects
+func ToRuntimeObjects(crds ...*apiextenv1.CustomResourceDefinition) []client.Object {
```
Brian-McM, line 183: I think we already have a definition for this, but it might be in the render package (which I don't think you want to access here). You may want to consider pulling that definition out into a common package accessible by both.

pkg/controller/installation/core_controller.go
```diff
@@ -1460,6 +1474,21 @@ func (r *ReconcileInstallation) checkActive(log logr.Logger) (*corev1.ConfigMap,
 	}
 }
 
+func (r *ReconcileInstallation) updateCRDs(ctx context.Context, variant operator.ProductVariant, log logr.Logger) error {
+	if !r.manageCRDs {
```
Brian-McM, line 1478: I wonder if this check should be done by the caller, looking at the calling code it looks like we're updating the CRD's no matter what.

pkg/controller/utils/component.go
```diff
@@ -297,6 +302,11 @@ func mergeState(desired client.Object, current runtime.Object) client.Object {
 		dsa.Spec.ElasticsearchRef = csa.Spec.ElasticsearchRef
 		dsa.Status = csa.Status
 		return dsa
+	case *apiextenv1.CustomResourceDefinition:
+		c := current.(*apiextenv1.CustomResourceDefinition)
+		d := desired.(*apiextenv1.CustomResourceDefinition)
+		d.SetResourceVersion(c.GetResourceVersion())
```
Brian-McM, line 308: Is this correct? Would we want the resource version to be incremented, or fail if the resource current resource version is ahead of the desired (this protects against race conditions and reverting changes).

pkg/crds/crds.go
tmjd: I'm trying to prevent if there were multiple calls happening writing the global calicoCRDs or enterpriseCRDs.

Brian-McM: Is there a write here that I missed?

Brian-McM: This is only reading from the files and returning the results, or am I missing something? Two simultaneous calls to this would conflict.

pkg/crds/crds.go
tmjd: We don't have this for CustomResourceDefinitions, we have it for Secrets and/or ConfigMaps.
I don't think there is any reason that ToRuntimeObjects should go in the render package, we've just been putting it there because we've been using it for rendering. It seems good to keep it in a package that is named for what the ToRuntimeObjects operators on.

pkg/controller/installation/core_controller.go
tmjd: Do you think we need to know when looking through the Reconcile function if we actually do update the CRDs or not?
I thought it was reasonable to hide that decision, especially since nothing else in the Reconcile cares if we've updated the CRDs or not.

pkg/crds/crds.go
tmjd: Where the CRDs are being stored is global. So by write I meant appending to calicoCRDs or enterpriseCRDs.

Makefile
```diff
@@ -483,6 +493,53 @@ $(BINDIR)/gen-versions: $(shell find ./hack/gen-versions -type f)
...
+	@$(CONTAINERIZED) \
+	bash -c '$(GIT_CONFIG_SSH) \
+	cd .crds/$(dir); \
+	git init .;  \
+	git remote add -f --tags origin $(GITHUB_ACCESS_METHOD)$(project) 2>&1 | grep -v -e "new branch" -e "new tag"; \
+	git config --local core.sparseCheckout true; \
+	echo "config/crd" >> .git/info/sparse-checkout; \
+	git checkout -q tags/$(branch) &>/dev/null ||  git checkout -q origin/$(branch) --;' || echo "Failed to fetch $(dir) CRDs"
+	@cp .crds/$(dir)/config/crd/* pkg/crds/$(dir)/ && echo "Copied $(dir) CRDs"
+	git add pkg/crds/$(dir)
```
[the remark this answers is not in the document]
Brian-McM: We can change that, I definitely think the `dirty-check` should check if there's new files as seen be `git status`.

pkg/controller/utils/component.go
Brian-McM: > that meant the code above would not replace the ResourceVersion because there was already one set

Why do we want to replace the resource version though? If you were to edit a resource, the best way (I thought) would be to pull the resource with the resource version, edit it, then write it back. That way if something else did this simultaneously then this write would fail because the resource version would be different. I might be missing something though....

pkg/crds/crds.go
Brian-McM: OK, this is what I was missing, I didn't realise that `calicoCRDs` and `enterpriseCRDs` were global. The lock makes sense now to me now. Thanks for explaining.

pkg/controller/installation/core_controller.go
Brian-McM: Right, I didn't realise you were storing that globally. How much space does it take to keep those CRDs in memory?

I'm wondering if we should include that CRD watch in the installation controller vs having the installation controller write the CRDs once on startup then have a CRD controller watch for the individual CRD changes. That way if a CRD changes we could just have a map of CRD names to file (created at startup) and load the specific CRD that we detected was changed. This would give us improved CPU and memory, but it complicates the code base a little more.

pkg/controller/installation/core_controller.go
Brian-McM: Ya that's fair, having this here makes sense as well.

pkg/crds/crds.go
Brian-McM: Oh right.

pkg/controller/installation/core_controller.go
tmjd: Laurence is actually working on changes that add in a dedup of our Reconcile since we fully reconcile on each pass. So I think what you are suggesting would be counter productive to that change.

I think I'd be opposed to optimizing this more as I think my time spent optimizing with those globals was probably unwarranted since I didn't have any evidence that it was needed.

pkg/controller/utils/component.go
tmjd: We want our desired resource to have the "current" ResourceVersion so when we do the update if something else has edited the same resource our update will fail.
The whole point of this mergeState is to merge our 'desired' resource with the 'current' one that was retrieved, this function is doing what you are suggesting we need to do.
I'm not quite sure why we don't just overwrite the desired ResourceVersion with the current ResourceVersion. That should be something we can always do.

pkg/controller/clusterconnection/shim_test.go
```diff
@@ -18,14 +18,20 @@
...
+	"github.com/tigera/operator/pkg/controller/options"
 	"github.com/tigera/operator/pkg/controller/status"
 	"k8s.io/apimachinery/pkg/runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
 func NewReconcilerWithShims(cli client.Client, schema *runtime.Scheme, status status.StatusManager, provider operatorv1.Provider) reconcile.Reconciler {
+	opts := options.AddOptions{
+		ShutdownContext: context.TODO(),
```
Brian-McM, line 33: I think this should be something other than TODO

test/mainline_test.go
```diff
@@ -147,20 +148,21 @@ var _ = Describe("Mainline component function tests", func() {
 				return nil
 			}, 60*time.Second).Should(BeNil())
 
+			fmt.Printf("Status is %+v", instance.Status)
```
Brian-McM, line 151: Leftover debugging?

PR #1566, opened 2021-10-15
Use boringssl

caseydavenport:
## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -84,11 +84,18 @@ endif
 EXCLUDEARCH ?= s390x
 VALIDARCHES = $(filter-out $(EXCLUDEARCH),$(ARCHES))
 
+# We need CGO to leverage Boring SSL.  However, the cross-compile doesn't support CGO yet.
+ifeq ($(ARCH), $(filter $(ARCH),amd64))
+CGO_ENABLED=1
+else
+CGO_ENABLED=0
+endif
+
 ###############################################################################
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -125,8 +132,7 @@ CONTAINERIZED= mkdir -p .go-pkg-cache $(GOMOD_CACHE) && \
 		-e KUBECONFIG=/go/src/$(PACKAGE_NAME)/kubeconfig.yaml \
 		-w /go/src/$(PACKAGE_NAME) \
 		--net=host \
-		$(EXTRA_DOCKER_ARGS) \
-		$(CALICO_BUILD)
+		$(EXTRA_DOCKER_ARGS)
```
tmjd, line 135: Is there a reason you didn't just add the CGO_ENABLED to EXTRA_DOCKER_ARGS? I think that would have avoided a lot of these changes. I know it isn't necessary for most calls but would there have been a downside?

caseydavenport: enabling cgo for the other builds means the compiles binaries aren't static and rely on certain libraries being present in the environment when running the compiled code.

Maybe it would work, but doesn't seem worth the extra complexity for things like UTs, etc. Also, this aligns the makefile more closely with the rest of the projectcalico makefiles, which generally have a split between the docker run command and the image being run.

mgleung (approved):

PR #1574, opened 2021-10-19
Retry when Typha autoscaler reads zero nodes from indexer

hjiawei:
## Description

When the Typha autoscaler reads zero nodes from the indexer, it will
retry node indexer `List()` again until the schedulable nodes are
greater than zero.

We notice an issue in a cluster where kube apiserver performance is
degraded due to malfunctioning nodes/routers. The Operator will scale
Typha up/down which causes churns to the cluster.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/common/autoscale.go
+++ b/pkg/common/autoscale.go
@@ -21,7 +21,7 @@ func GetExpectedTyphaScale(nodes int) int {
 	typhas := (nodes / maxNodesPerTypha) + 1
 	// We add one more to ensure there is always 1 extra for high availability purposes.
 	typhas += 1
-	// If we don't have enough nodes to have 3 typhs then make sure there is one typha for each node.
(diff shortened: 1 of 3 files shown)


tmjd:
Interesting what you've done here. It isn't really what i was thinking, sorry I wasn't clear. I meant actually do a client call to get the nodes, instead of reading them from the indexer.
@caseydavenport WDYT?
My concern is that we can't actually scale to zero typhas, though of course if you're scaling to zero what is the point but I can imagine a few cases where that might be able to happen but then where is the operator running?

hjiawei:
Scaling Typha to 0 is found near the start of the Operator so there could be a race condition where node and typha informers are not synced. The fix for that is in https://github.com/tigera/operator/pull/1581. Thanks for looking into this @tmjd.

PR #1589, opened 2021-10-28
Ensure logstorage use patch for its controlling resource

jaderhs:
## Description

This PR is a bug fix that affects the LogStorage controller and renderer.
Patch LogStorage main resource to prevent overwriting fields the controller is unaware of
Move LogStorage finalizer logic outside of the renderer and into the logstorage controller package

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


pkg/controller/logstorage/logstorage_controller_test.go
```diff
@@ -1315,6 +1320,7 @@ func setUpLogStorageComponents(cli client.Client, ctx context.Context, storageCl
 		nil, nil, "cluster.local", nil, render.ElasticsearchLicenseTypeBasic)
 
 	createObj, _ := component.Objects()
+	createObj = append(createObj, ls)
```
tmjd, line 1323: I think instead of adding this here, you should add an `Expect(cli.Create(ctx, ls)).ShouldNot(HaveOccurred())` before the `render.LogStorage`.

pkg/render/logstorage_test.go
```diff
@@ -495,6 +479,7 @@ var _ = Describe("Elasticsearch rendering tests", func() {
 
 				createResources, _ := component.Objects()
 
+				createResources = append([]client.Object{ls}, createResources...)
```
tmjd, line 482: We don't need to include `ls` here, instead we should stop checking that it is created.

pkg/render/logstorage_test.go
```diff
@@ -528,6 +513,8 @@ var _ = Describe("Elasticsearch rendering tests", func() {
 
 				updatedResources, _ := updatedComponent.Objects()
 
+				updatedResources = append([]client.Object{ls}, updatedResources...)
```
tmjd, line 516: Same here
> We don't need to include `ls` here, instead we should stop checking that it is created.

pkg/controller/logstorage/logstorage.go
```diff
@@ -47,7 +47,7 @@ func (r *ReconcileLogStorage) createLogStorage(
 	hdler utils.ComponentHandler,
 	reqLogger logr.Logger,
 	ctx context.Context,
-) (reconcile.Result, bool, error) {
+) (reconcile.Result, bool, bool, error) {
```
tmjd, line 50: Could you add a function definition to this, especially to include comments on what the return values are?

pkg/render/logstorage_test.go
```diff
@@ -1551,12 +1528,7 @@ var deleteLogStorageTests = func(managementCluster *operatorv1.ManagementCluster
 			compareResources(deleteResources, []resourceTestObj{})
 		})
 		It("removes the finalizers from LogStorage if Elasticsearch and Kibana are both nil", func() {
-			expectedCreateResources := []resourceTestObj{
-				{"tigera-secure", "", &operatorv1.LogStorage{}, func(resource runtime.Object) {
-					ls := resource.(*operatorv1.LogStorage)
-					Expect(ls.Finalizers).ShouldNot(ContainElement(render.LogStorageFinalizer))
-				}},
-			}
+			expectedCreateResources := []resourceTestObj{}
```
tmjd, line 1531: This test is trying to verify something that this PR is removing so we should remove this test.

tmjd (changes requested):

pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -196,20 +198,11 @@ type ReconcileLogStorage struct {
 	clusterDomain string
 }
 
-func GetLogStorage(ctx context.Context, cli client.Client) (*operatorv1.LogStorage, error) {
-	instance := &operatorv1.LogStorage{}
-	err := cli.Get(ctx, utils.DefaultTSEEInstanceKey, instance)
-	if err != nil {
-		return nil, err
-	}
+func fillDefaultsAndValidateLogStorage(ctx context.Context, cli client.Client, instance *operatorv1.LogStorage) error {
```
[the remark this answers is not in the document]
jaderhs: Removed `fillDefaultsAndValidateLogStorage()` definition and changed in tests to:
```golang
                       fillDefaults(ls)
                       Expect(validateComponentResources(&ls.Spec)).To(BeNil())
```

pkg/controller/logstorage/logstorage_controller_test.go
jaderhs: Change applied:
```diff
+       Expect(cli.Create(ctx, ls)).ShouldNot(HaveOccurred())
        createObj, _ := component.Objects()
-       createObj = append(createObj, ls)
```

pkg/render/logstorage_test.go
jaderhs: Test removed.

pkg/controller/logstorage/logstorage.go
jaderhs: Definition added:
```diff
+// createLogStorage Is called by Reconcile() in the Logstorage controller to render its components
+//
+// It returns 4 values:
+// A reconcile.Result to be returned by Reconcile() if the 'proceed' bool is false or if there is an error,
+// a 'proceed' bool indicating if the Reconcile function should proceed with the reconcile process,
+// a 'finalizerCleanup' bool indicating if it's safe to remove the LogStorage finalizer (only on deletion and after
+// kibana and elasticsearch components were successfully removed), and lastly,
+// an error if there was any or nil otherwise
```

PR #1620, opened 2021-11-09
Fix cert-management for ECK and PCAP

rene-dekker:
I have noticed that Certificate Management was out of date ever since ECK was updated to a newer version. The way certificates are created and mounted has changed since then. This PR addresses that issues.

Also a bug was spotted with PCAP.

### DIFF
--- a/pkg/render/logstorage.go
+++ b/pkg/render/logstorage.go
@@ -152,12 +152,8 @@ const (
 	csrVolumeNameHTTP = "elastic-internal-http-certificates"
 	// Volume that is added by ECK and is overridden if certificate management is used.
 	csrVolumeNameTransport = "elastic-internal-transport-certificates"
-	// Volume that is added by us and contains the expected secret for ECK.
-	transportSecretVolumeName = "elastic-internal-transport-certificates-secret"
 	// Volume name that is added by ECK for the purpose of mounting certs.
 	caVolumeName = "elasticsearch-certs"
-	// Name of the secret that ECK mounts for the prepare-fs script.
-	transportSecretName = "tigera-secure-es-transport-certificates"
 )
 
 var log = logf.Log.WithName("render")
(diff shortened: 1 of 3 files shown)


pkg/render/logstorage.go
```diff
@@ -584,28 +580,12 @@ func (es elasticsearchComponent) podTemplate() corev1.PodTemplateSpec {
 				},
 			},
 			VolumeMounts: []corev1.VolumeMount{
-				// Even though it won't be used for TLS, we need to mount the secret volume for initialization.
```
rene-dekker, line 587: Comment to the reviewer. In earlier versions of ECK 3 volumes were mounted by ECK to the already present init container. When cert management is enabled we wanna add a fourth. ECK does something that looks like a strategic merge, but if we did not specify the existing 3, they would be left out and the container would crash.

In the new ECK version these 3 mounts are no longer present, so there is no need any longer to add them in our code.

pkg/render/logstorage.go
```diff
@@ -572,7 +568,7 @@ func (es elasticsearchComponent) podTemplate() corev1.PodTemplateSpec {
 			SecurityContext: &corev1.SecurityContext{
 				Privileged: ptr.BoolToPtr(false),
 			},
-			Args: []string{"bash", "-c", "/mnt/elastic-internal/scripts/prepare-fs.sh"},
+			Command: []string{"bash", "-c", "mkdir /mnt/elastic-internal/transport-certificates/ && touch /mnt/elastic-internal/transport-certificates/$HOSTNAME.tls.key && /mnt/elastic-internal/scripts/prepare-fs.sh"},
```
hjiawei, line 571: Not sure why we need to create the `transport-certificates` folder and the empty `$HOSTNAME.tls.key` file but the volume mount is removed below?

rene-dekker: The prepare-fs script waits until a file if present in this location. Normally, it would use that secret and place it in the location that the elasticsearch.yml file uses when booting elastic. However, we don't want an ECK secret, we want the CSR init container to create a certificate request and use that instead. This `touch` command lets the `prepare-fs` script finish.

hjiawei (approved):
LGTM

Brian-McM (approved):

PR #1621, opened 2021-11-09
Automated cherry pick of #1506: Don't run so many Typha instances on small clusters

caseydavenport:
Cherry pick of #1506 on release-v1.24.

#1506: Don't run so many Typha instances on small clusters

# Original PR Body below

## Description

After some experience in the wild, we believe we don't need so many replicas on small clusters. This is largely because:
- during rolling update Typha will restart very quickly. 
- Running additional Typhas on very small clusters has additional resource impact 
- Running numTyphas == numNodes can also result in scheduling deadlocks during cluster scaling events, when the cluster autoscaler cannot find a suitable node to reschedule typha (since only one instance can run per node). 

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/common/autoscale.go
+++ b/pkg/common/autoscale.go
@@ -14,17 +14,40 @@
 package common
 
 // GetTyphaScaleCount will return the number of Typhas needed for the number of nodes.
+//
+// Nodes       Replicas
+//   1-2              1
+//   3-4              2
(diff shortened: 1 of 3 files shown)


tmjd (approved):
LGTM

PR #1633, opened 2021-11-10
Add authn-proxy to prometheus as a side-car

rene-dekker:
As per this PR Prometheus will sit behind an authentication proxy:
- Traffic between voltron and prometheus is now done over TLS
- Authn proxy can read the authorization headers and authenticate a request
- The prometheus server listens on localhost, so it cannot be queried from outside of the pod.


pkg/render/monitor/monitor.go
```diff
@@ -138,6 +208,55 @@ func (mc *monitorComponent) Objects() ([]client.Object, []client.Object) {
 	return toCreate, toDelete
 }
 
+func (mc *monitorComponent) clusterRole() client.Object {
```
hjiawei, line 211: Is it possible to reduce ClusterRole/ClusterRoleBinding to namespaced Role/RoleBinding?

rene-dekker: All three of the permissions below are cluster scoped, so they cannot be managed on a per-ns basis. That being said, I think I might be able to remove the authenticationreviews, so I will explore that now.

pkg/render/monitor/monitor.go
```diff
@@ -183,16 +297,112 @@ func (mc *monitorComponent) alertmanagerService() *corev1.Service {
...
-			ServiceAccountName:     prometheusServiceAccountName,
+			Image:              &mc.prometheusImage,
+			ImagePullSecrets:   secret.GetReferenceList(mc.cfg.PullSecrets),
+			ServiceAccountName: prometheusServiceAccountName,
+			Volumes:            volumes,
+			InitContainers:     initContainers,
+			Containers: []corev1.Container{
+				{
+					Name:            "authn-proxy",
+					Image:           "gcr.io/tigera-dev/rd/tigera/prometheus-service:rene", //revert to mc.prometheusServiceImage
```
rene-dekker, line 374: todo: revert

pkg/controller/clusterconnection/clusterconnection_controller.go
```diff
@@ -95,6 +95,10 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 	if err = utils.AddSecretsWatch(c, render.PacketCaptureCertSecret, common.OperatorNamespace()); err != nil {
 		return fmt.Errorf("%s failed to watch Secret resource %s: %w", controllerName, render.PacketCaptureCertSecret, err)
 	}
+	// Watch for changes to the secrets associated with the PacketCapture APIs.
```
Brian-McM, line 98: This comment is incorrect.

pkg/render/monitor/monitor.go
```diff
@@ -45,35 +48,74 @@ const (
...
-	return &monitorComponent{
-		cfg: cfg,
+func Monitor(cfg *Config) (render.Component, error) {
+	var err error
+	var tlsSecrets []*corev1.Secret
+	var tlsHash string
+	if cfg.Installation.CertificateManagement == nil {
+		if cfg.TLSSecret == nil {
+			svcDNSNames := dns.GetServiceDNSNames(PrometheusHTTPAPIServiceName, common.TigeraPrometheusNamespace, cfg.ClusterDomain)
+			cfg.TLSSecret, err = secret.CreateTLSSecret(nil,
```
Brian-McM, line 77: Is the customer able to provide the TLS configuration?

pkg/controller/clusterconnection/clusterconnection_controller.go
```diff
@@ -194,14 +198,31 @@ func (r *ReconcileConnection) Reconcile(ctx context.Context, request reconcile.R
...
+		common.OperatorNamespace(),
+		render.PrometheusTLSSecretName,
+		"", // We don't need the key.
+		corev1.TLSCertKey,
+	)
+	if err != nil {
+		reqLogger.Error(err, fmt.Sprintf("failed to retrieve %s", render.PrometheusTLSSecretName))
+		r.status.SetDegraded(fmt.Sprintf("Failed to retrieve %s", render.PrometheusTLSSecretName), err.Error())
+		return reconcile.Result{}, err
+	} else if prometheusCertSecret == nil {
```
Brian-McM, line 211: This now forces the user to setup monitoring before the tunnel will be set up. I'm not sure if we should have that as a pre requisite of the tunnel setup, as I would think not having monitoring setup should be a valid configuration.

You might want to double check with PM on this.

pkg/controller/manager/manager_controller.go
```diff
@@ -489,6 +489,22 @@ func (r *ReconcileManager) Reconcile(ctx context.Context, request reconcile.Requ
...
+		common.OperatorNamespace(),
+		render.PrometheusTLSSecretName,
+		"", // We don't need the key.
+		corev1.TLSCertKey,
+	)
+	if err != nil {
+		reqLogger.Error(err, fmt.Sprintf("failed to retrieve %s", render.PrometheusTLSSecretName))
+		r.status.SetDegraded(fmt.Sprintf("Failed to retrieve %s", render.PrometheusTLSSecretName), err.Error())
+		return reconcile.Result{}, err
+	} else if prometheusCertSecret == nil {
```
Brian-McM, line 502: Do we state that you must have monitoring setup up for the manager to work? I didn't think we had this requirement, monitoring seems like it should be more of an addon, it might be worth checking with PM.

pkg/controller/monitor/monitor_controller.go
```diff
@@ -191,6 +199,45 @@ func (r *ReconcileMonitor) Reconcile(ctx context.Context, request reconcile.Requ
...
+			monitor.PrometheusTLSSecretName,
+			corev1.TLSPrivateKeyKey,
+			corev1.TLSCertKey,
+		)
+		if err != nil {
+			log.Error(err, "Invalid TLS Cert")
+			r.status.SetDegraded("Error validating TLS certificate", err.Error())
+			return reconcile.Result{}, err
+		}
+		r.status.RemoveCertificateSigningRequests(common.TigeraPrometheusNamespace)
```
Brian-McM, line 217: Is for the case that somebody switches from certificate management to no certificate manager?

pkg/controller/monitor/monitor_controller.go
```diff
@@ -191,6 +199,45 @@ func (r *ReconcileMonitor) Reconcile(ctx context.Context, request reconcile.Requ
...
+	} else {
+		// Monitor pending CSRs for the TigeraStatus
+		r.status.AddCertificateSigningRequests(common.TigeraPrometheusNamespace, map[string]string{"k8s-app": common.TigeraPrometheusNamespace})
+	}
+
+	// Fetch the Authentication spec. If present, we use to configure user authentication.
+	authenticationCR, err := utils.GetAuthentication(ctx, r.client)
```
Brian-McM, line 224: Just for my understanding, we're adding OIDC login capabilities to the prometheus UI, which is why we need this configuration, right?

Brian-McM:
Could you please update at least the description to detail what you're trying to accomplish? The end goal isn't adding a side cart to prometheus, it looks like you want to do two this:
1. Require TLS authentication for all requests to prometheus
2. Add login restrictions to the prometheus UI

Could you also please remove the ticket number from the title, this is open source so it doesn't make sense to have that here.

pkg/render/monitor/monitor.go
```diff
@@ -45,35 +48,74 @@ const (
...
 )
 
-func Monitor(cfg *Config) render.Component {
-	return &monitorComponent{
-		cfg: cfg,
+func Monitor(cfg *Config) (render.Component, error) {
+	var err error
+	var tlsSecrets []*corev1.Secret
+	var tlsHash string
+	if cfg.Installation.CertificateManagement == nil {
```
Brian-McM, line 74: I don't think this is supposed to be in the render package, I think this kind of verification is supposed to be done before render.

Are we doing this in other places, or is this new?

Brian-McM (changes requested):

pkg/controller/monitor/monitor_controller.go
rene-dekker: Yes, that's right. It is then no longer necessary to watch for pending csr's in this namespace.

pkg/controller/monitor/monitor_controller.go
rene-dekker: Yes, it is for reading the bearer tokens. Although, with the current implementation, it will mostly read tigera-manager impersonation tokens.

pkg/render/monitor/monitor.go
rene-dekker: I'm following the established pattern, which enables a user to provide the certificates. I have not yet added the doc for it. Let me add that to a ticket.

pkg/controller/clusterconnection/clusterconnection_controller.go
rene-dekker: I removed the dependency.

Brian-McM (approved):

PR #1636, opened 2021-11-10
removing validation for LogIntervalSeconds

penkeysuresh:
## Description

When using validations (minimum, maximum) helm v2 is expecting it to be a `float64`, so the current CRD fails. To fix this I am removing the validation. We are expected to keep this back on after we drop support for helm v2.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


doublek (approved):
LGTM

PR #1637, opened 2021-11-11
windows upgrader followup

lmm:
## Description

This PR addresses some issues with the calico windows upgrader:
- Fix node patching that can result in a panic
- Ensure the upgrader only runs on AKS
- Fix status handling when windows upgrader fails to get node upgrade status

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/installation/core_controller.go
+++ b/pkg/controller/installation/core_controller.go
@@ -1234,6 +1234,12 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 		reqLogger.V(1).Info("Unable to determine MTU - no explicit config, and /var/lib/calico is not mounted")
 	}
 
+	// Check whether the calicoWindowsUpgrader is degraded. If so, requeue
+	// a reconcile.
+	if r.calicoWindowsUpgrader.IsDegraded() {
+		return reconcile.Result{RequeueAfter: 30 * time.Second}, nil
+	}
+
 	// We have successfully reconciled the Calico installation.
 	if instance.Spec.KubernetesProvider == operator.ProviderOpenShift {
 		openshiftConfig := &configv1.Network{}
--- a/pkg/controller/installation/core_controller_test.go
+++ b/pkg/controller/installation/core_controller_test.go
@@ -1112,6 +1112,7 @@ var _ = Describe("Testing core-controller installation", func() {
 
 		Context("calicoWindowsUpgrader", func() {
 			BeforeEach(func() {
+				// calicoWindowsUpgrader only upgrades nodes on AKS.
 				cr.Spec.KubernetesProvider = operator.ProviderAKS
 			})
 
--- a/pkg/controller/installation/core_controller_test.go
+++ b/pkg/controller/installation/core_controller_test.go
@@ -704,7 +704,7 @@ var _ = Describe("Testing core-controller installation", func() {
 			mockStatus.On("AddCertificateSigningRequests", mock.Anything)
 			mockStatus.On("RemoveCertificateSigningRequests", mock.Anything)
 			mockStatus.On("ReadyToMonitor")
-			mockStatus.On("SetWindowsUpgradeStatus", mock.Anything, mock.Anything, mock.Anything)
+			mockStatus.On("SetWindowsUpgradeStatus", mock.Anything, mock.Anything, mock.Anything, nil)
 
 			// Create the indexer and informer shared by the typhaAutoscaler and
 			// calicoWindowsUpgrader.
@@ -1124,7 +1124,7 @@ var _ = Describe("Testing core-controller installation", func() {
 				// Create node with current Enterprise version.
 				n1 := test.CreateWindowsNode(cs, "windows1", cr.Spec.Variant, components.EnterpriseRelease)
 
-				mockStatus.On("SetWindowsUpgradeStatus", []string{}, []string{}, []string{"windows1"})
+				mockStatus.On("SetWindowsUpgradeStatus", []string{}, []string{}, []string{"windows1"}, nil)
 
 				// Node is up to date and should not have changed.
 				Consistently(func() error {
@@ -1150,7 +1150,7 @@ var _ = Describe("Testing core-controller installation", func() {
 				n1 := test.CreateWindowsNode(cs, "windows1", operator.Calico, "v3.21.999")
 				n2 := test.CreateWindowsNode(cs, "windows2", operator.TigeraSecureEnterprise, components.EnterpriseRelease)
 
-				mockStatus.On("SetWindowsUpgradeStatus", mock.Anything, mock.Anything, mock.Anything)
+				mockStatus.On("SetWindowsUpgradeStatus", mock.Anything, mock.Anything, mock.Anything, nil)
 
 				// Ensure that outdated nodes have the new label and taint.
 				Eventually(func() error {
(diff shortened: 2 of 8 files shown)


pkg/controller/installation/windows/calico_windows_upgrader.go
```diff
@@ -60,6 +61,8 @@ type calicoWindowsUpgrader struct {
 	syncPeriod        time.Duration
 	installChan       chan *operatorv1.InstallationSpec
 	install           *operatorv1.InstallationSpec
+	isDegraded        bool
```
tmjd, line 64: Don't these 2 need to be locked with a mutex?

pkg/controller/installation/core_controller.go
```diff
@@ -1234,6 +1234,12 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 		reqLogger.V(1).Info("Unable to determine MTU - no explicit config, and /var/lib/calico is not mounted")
 	}
 
+	// Check whether the calicoWindowsUpgrader is degraded. If so, requeue
+	// a reconcile.
+	if r.calicoWindowsUpgrader.IsDegraded() {
+		return reconcile.Result{RequeueAfter: 30 * time.Second}, nil
```
caseydavenport, line 1240: I'm having a bit of trouble remembering how exactly it's meant to work, but the typha autoscaler also sets degraded from within this loop... https://github.com/tigera/operator/blob/5ee263960b7b814083f77e60a2e1b31663a31ebe/pkg/controller/installation/core_controller.go#L806

@tmjd should we have all SetDegraded calls occurring within the main controller loop? The typha autoscaler seems to set it in two places - within the autoscaler itself, but also within the core controller after checking `IsDegraded` - one of those seems like a duplicate, unless I'm missing something... maybe it's because we can't guarantee a Reconcile() will trigger, so SetDegraded in this main loop won't necessarily be called? If that's the case, why do we need a SetDegraded in core controller?

caseydavenport: Also, is there a reason not to put this check right next to the typhaAutoscaler equivalent up at line 804? What's the reason for putting it down here?

pkg/controller/installation/core_controller_test.go
```diff
@@ -1111,6 +1111,10 @@ var _ = Describe("Testing core-controller installation", func() {
 		})
 
 		Context("calicoWindowsUpgrader", func() {
+			BeforeEach(func() {
+				cr.Spec.KubernetesProvider = operator.ProviderAKS
```
caseydavenport, line 1115: This needs a comment explaining why.

pkg/controller/installation/windows/calico_windows_upgrader.go
```diff
@@ -267,6 +286,10 @@ func (w *calicoWindowsUpgrader) Start(ctx context.Context) {
 			case install := <-w.installChan:
 				w.install = install
 			case <-ticker.C:
+				if w.install.KubernetesProvider != operatorv1.ProviderAKS {
```
caseydavenport, line 289: Rather than exit, I think we should do something like this:

```go
if w.install.KubernetesProvider == operatorv1.ProviderAKS {
  w.updateWindowsNodes()
}
```

this has a couple benefits:
- We don't exit the loop - thus, if the provider is manually set to something else and then back to AKS, we're recover (since Start() is only called once at controller startup). 
- We don't need the `w.isRunning` bool - we'll always handle config updates, we'll just not take any action unless we're on AKS.

lmm: @caseydavenport yes I agree. As it is now, if the user happens to change the installation's `kubernetesProvider` to anything other than AKS, that will effectively turn off the calicoWindowsUpgrader and the only way to get it running again is to restart the operator.

pkg/controller/installation/windows/calico_windows_upgrader.go
lmm: Thanks, yes `isDegraded` needs a mutex - it's being read from the core controller goroutine and set by the calicoWindowsUpgrader goroutine.

pkg/controller/installation/core_controller.go
lmm: I put the `r.calicoWindowsUpgrader.IsDegraded()` call later in the core `Reconcile()` to minimize the chance of this order of events:
- Partway through core `Reconcile()`, `r.calicoWindows.IsDegraded()` is called but returns false. core `Reconcile()` continues on.
- `updateWindowsNodes()` is triggered by the calicoWindowsUpgrader's internal ticker. The call results in an error and calicoWindowsUpgrader calls `SetDegraded`
- core `Reconcile()` reaches the end and calls `ClearDegraded`

When the calicoWindowsUpgrader's internal ticker ticks again, if `updateWindowsNodes()` returns an error again then it will `SetDegraded` so that's not so bad maybe, but the core `Reconcile()` won't be triggered and we might need to wait 5 minutes for core to reconcile again.

Perhaps this scenario isn't that bad. calico status will be degraded and resolving will involve updating the Windows nodes which would trigger a core reconcile.

tmjd: I don't know that it matters where the SetDegraded is called.
In the case of typhaAutoscaler, the core Reconcile triggers an autoscale attempt and if there is an error from that specific attempt then that is why SetDegraded is called in the core Reconcile (it probably could have also been called by the autoscaler loop that serviced the autoscale request). Otherwise if the autoscale attempt happens because of the autoscaler ticker then the autoscale loop does SetDegraded.

tmjd: I'm thinking this is possibly a better place for the typha autoscaler since the isDegraded for autoscale is followed up with triggering an attempt to autoscale. It is possible that the changes that were just applied in the Reconcile fixed some issue that was preventing autoscaling before.

tmjd: For Windows upgrading, I don't think that race condition is a concern assuming we have the windows upgrader reconciling frequently enough. Even if the upgrader had no errors until after the `r.calicoWindows.IsDegraded()` was called and the Reconcile called ClearDegraded, when the upgrader runs again it will SetDegraded. So assuming our period is low enough, it won't matter if the ClearDegraded removes the initial SetDegraded that happens.
(Am I remembering right and the windows upgrader is being ran on a lowish periodic timer?)

pkg/controller/installation/core_controller_test.go
lmm: 1cd3758

pkg/controller/installation/core_controller.go
lmm: > (Am I remembering right and the windows upgrader is being ran on a lowish periodic timer?)

Yes it's a 10s period.

pkg/controller/installation/windows/calico_windows_upgrader.go
lmm: 5cd4def

caseydavenport (approved):
@lmm as mentioned on Slack this LGTM but we should probably use SetDegraded and remove the error from `SetWindowsUpgradeStatus` so we don't have two ways of populating those fields. Feel free to merge after making that change and squashing, I don't need to look again.

Might want @tmjd to approve too.

pkg/controller/status/status.go
```diff
@@ -761,6 +773,9 @@ func (m *statusManager) degradedReason() string {
 	if m.explicitDegradedReason != "" {
 		reasons = append(reasons, m.explicitDegradedReason)
 	}
+	if m.windowsUpgradeDegradedReason != "" {
```
tmjd, line 776: do we also need to add a degradedMessage when windowsUpgradeDegradedReason is not empty?

tmjd (approved):

PR #1651, opened 2021-11-18
Ensure kibana secrets provided by the user do not have ownerreferences added

jaderhs:
## Description

This change is a bug fix to prevent an OwnerReference being added to an user provided kibana tls secret
This change introduce a new logstorage controller test to validate this bug fix

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/logstorage/logstorage_controller.go
+++ b/pkg/controller/logstorage/logstorage_controller.go
@@ -559,7 +559,7 @@ func (r *ReconcileLogStorage) getElasticsearchCertificateSecrets(ctx context.Con
 	}
 
 	// Ensure that cert is valid.
-	esKeyCert, err = utils.EnsureCertificateSecret(render.TigeraElasticsearchInternalCertSecret, esKeyCert, corev1.TLSPrivateKeyKey, corev1.TLSCertKey, rmeta.DefaultCertificateDuration, svcDNSNames...)
+	esKeyCert, _, err = utils.EnsureCertificateSecret(render.TigeraElasticsearchInternalCertSecret, esKeyCert, corev1.TLSPrivateKeyKey, corev1.TLSCertKey, rmeta.DefaultCertificateDuration, svcDNSNames...)
 	if err != nil {
 		return nil, nil, err
 	}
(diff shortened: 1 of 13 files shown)


pkg/controller/utils/certs.go
```diff
@@ -56,35 +56,37 @@ func GetSecret(ctx context.Context, client client.Client, name string, ns string
...
-// secret is created and returned. If the secret does have the
-// right DNS names then the secret is returned.
+// secret is created and returned alongside a true boolean. If the secret does have the
+// right DNS names then the secret is returned alongside a false boolean.
 // If the cert in the secret has invalid DNS names and the secret is operator
-// managed, then a new secret is created and returned. Otherwise,
+// managed, then a new secret is created and returned alongside a true boolean. Otherwise,
 // if the secret is user-supplied, an error is returned.
-func EnsureCertificateSecret(secretName string, secret *corev1.Secret, keyName string, certName string, certDuration time.Duration, svcDNSNames ...string) (*corev1.Secret, error) {
+func EnsureCertificateSecret(secretName string, secret *corev1.Secret, keyName string, certName string, certDuration time.Duration, svcDNSNames ...string) (*corev1.Secret, bool, error) {
```
tmjd, line 64: I like this change to return bool if the secret is created or not, but I think the comment could be improved.
I'd suggest something like
```
EnsureCertificateSecret ensures that the certificate in the
secret has the expected DNS names.
The first returned field (*corev1.Secret) is the validated or created Secret to use.
The second returned field (bool) is true if the Secret returned was generated and needs to be created, otherwise false is returned.
The third field (error) is nil if the Secret pass in is valid or was created successfully. If there was a problem creating the certificate or the Secret has invalid DNS names and the secret is not operator managed, an error is returned.
```
I think this is more readable.

pkg/controller/manager/manager_controller.go
```diff
@@ -296,7 +296,7 @@ func (r *ReconcileManager) Reconcile(ctx context.Context, request reconcile.Requ
 			svcDNSNames := dns.GetServiceDNSNames(render.ManagerServiceName, render.ManagerNamespace, r.clusterDomain)
 			svcDNSNames = append(svcDNSNames, "localhost")
 			certDur := 825 * 24 * time.Hour // 825days*24hours: Create cert with a max expiration that macOS 10.15 will accept
-			tlsSecret, err = utils.EnsureCertificateSecret(
+			tlsSecret, _, err = utils.EnsureCertificateSecret(
```
tmjd, line 299: I think now that we have this 2nd field we should make use of it for the manager secret also.

pkg/controller/utils/certs.go
jaderhs: Function description updated

pkg/render/logstorage.go
```diff
@@ -1202,6 +1224,8 @@ func (es elasticsearchComponent) kibanaCR() *kbv1.Kibana {
 					EmptyDir: &corev1.EmptyDirVolumeSource{}}})
 	}
 
+	kibanaSecrets := append(es.kibanaSecrets, es.cfg.KibanaCertSecret)
```
tmjd, line 1227: I don't think this is needed. A version of KibanaCertSecret is already in es.kibanaSecrets so there shouldn't be a need to add it again.

pkg/controller/manager/manager_controller.go
jaderhs: Updated to use the EnsureCertificateSecret 2nd return value to check if certificate is operator managed

pkg/controller/logstorage/logstorage_controller_test.go
```diff
@@ -1320,26 +1369,36 @@ func setUpLogStorageComponents(cli client.Client, ctx context.Context, storageCl
 		ElasticLicenseType: render.ElasticsearchLicenseTypeBasic,
 	}
 
+	kibanaSecrets := createKibanaSecrets()
```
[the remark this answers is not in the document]
jaderhs: No, I just replaced the previous secret creation with the new one. Removed those unused secrets

pkg/render/logstorage.go
jaderhs: Makes sense, the other version is a copy made alongside the component creation so it will always be up to date. Removed

pkg/controller/manager/manager_controller.go
```diff
@@ -275,41 +275,37 @@ func (r *ReconcileManager) Reconcile(ctx context.Context, request reconcile.Requ
...
 	// If the manager TLS secret exists, check whether it is managed by the
 	// operator.
-	certOperatorManaged := true
-	if tlsSecret != nil {
-		certOperatorManaged, err = utils.IsCertOperatorIssued(tlsSecret.Data[render.ManagerInternalSecretCertName])
+	var operatorManagedCertSecret bool
+	if installation.CertificateManagement == nil {
+		// If the secret does not exist, then create one.
+		// If the secret exists but is operator managed, then check that it has the
+		// expected DNS names and update it if necessary.
```
tmjd, line 282: These comments made more sense before this change. Please update them to better match what is being done here.
Maybe something like
```
We use EnsureCertificateSecret to ensure a secret exists, creating one if one is not passed in. It also ensures the secret passed has the proper DNS names if the secret is operator managed.
```
WDYT? I just think the current comments focus on what the code did so needs a little updating.

jaderhs: Makes sense since EnsureCertificateSecret is doing all that work now

pkg/controller/manager/manager_controller.go
```diff
@@ -275,41 +275,35 @@ func (r *ReconcileManager) Reconcile(ctx context.Context, request reconcile.Requ
...
+			render.ManagerTLSSecretName, tlsSecret, render.ManagerSecretKeyName, render.ManagerSecretCertName, certDur, svcDNSNames...,
+		)
+
+		if err != nil {
+			r.status.SetDegraded(fmt.Sprintf("Error ensuring manager TLS certificate %q exists and has valid DNS names", render.ManagerTLSSecretName), err.Error())
+			return reconcile.Result{}, err
+		}
+
+	} else if tlsSecret != nil {
+		operatorManagedCertSecret, err = utils.IsCertOperatorIssued(tlsSecret.Data[render.ManagerInternalSecretCertName])
```
tmjd, line 296: The code here does not need to set the higher level scoped `operatorManagedCertSecret`, we haven't generated the secret here so no need to create it or put ownership on it.

pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -599,62 +599,53 @@ func (r *ReconcileLogStorage) getElasticsearchCertificateSecrets(ctx context.Con
 	return esKeyCert, certSecret, err
 }
 
-func (r *ReconcileLogStorage) kibanaInternalSecrets(ctx context.Context, instl *operatorv1.InstallationSpec) ([]*corev1.Secret, error) {
+func (r *ReconcileLogStorage) kibanaInternalSecrets(ctx context.Context, instl *operatorv1.InstallationSpec) (*corev1.Secret, *corev1.Secret, bool, error) {
```
tmjd, line 602: I think you should change the order of the return values to `*corev1.Secret, bool, *corev1.Secret, error` since the bool is telling us about the 1st secret.

pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -599,62 +599,53 @@ func (r *ReconcileLogStorage) getElasticsearchCertificateSecrets(ctx context.Con
...
 
-	var secrets []*corev1.Secret
 	svcDNSNames := dns.GetServiceDNSNames(render.KibanaServiceName, render.KibanaNamespace, r.clusterDomain)
+	operatorManaged := false
```
tmjd, line 605: Because EnsureCertificateSecret is telling us if the secret was created or not we should rename it to reflect that instead of the fact if the secret is operatorManaged or not.

pkg/controller/logstorage/logstorage_controller.go
tmjd: Also please add a comment explaining the values returned from this function.

pkg/controller/logstorage/logstorage_controller_test.go
```diff
@@ -769,6 +769,56 @@ var _ = Describe("LogStorage controller", func() {
 					mockStatus.AssertExpectations(GinkgoT())
 				})
 
+				It("should not add OwnerReference to user supplied kibana TLS cert", func() {
```
tmjd, line 772: Could you add a test that the KibanaCertSecret is created with an OwnerReference when one is not provided? I don't see a test for that. We would have previously had that in the logstorage render.

tmjd (commented):
Looking good, I think we need a few more updates though.

pkg/controller/manager/manager_controller.go
tmjd: After chatting on a call, I think this comment should be ignored

pkg/controller/logstorage/logstorage_controller.go
tmjd: After our call, this suggestion should be ignored.

pkg/controller/logstorage/logstorage_controller_test.go
jaderhs: Created a new test for this case

pkg/controller/manager/manager_controller.go
```diff
@@ -275,41 +275,35 @@ func (r *ReconcileManager) Reconcile(ctx context.Context, request reconcile.Requ
...
-		certOperatorManaged, err = utils.IsCertOperatorIssued(tlsSecret.Data[render.ManagerInternalSecretCertName])
+	var operatorManagedCertSecret bool
+	if installation.CertificateManagement == nil {
+		// We use EnsureCertificateSecret to ensure a secret exists, creating one if one is not passed in.
+		// It also ensures the secret passed has the proper DNS names if the secret is operator managed.
+
+		svcDNSNames := dns.GetServiceDNSNames(render.ManagerServiceName, render.ManagerNamespace, r.clusterDomain)
+		svcDNSNames = append(svcDNSNames, "localhost")
+		certDur := 825 * 24 * time.Hour // 825days*24hours: Create cert with a max expiration that macOS 10.15 will accept
+		tlsSecret, operatorManagedCertSecret, err = utils.EnsureCertificateSecret(
```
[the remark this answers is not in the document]
jaderhs: I believe this does not apply anymore, correct?

pkg/controller/utils/certs.go
jaderhs: Function description updated

tmjd (approved):
LGTM

Issue #1510, opened 2021-09-09
Tigera Operator is causing a Massive amount of 404s on the apiserver trying to delete an inexistent resource 


djsly:
The Tigera operator running in AKS when selecting Calico causes a lot of 404s when trying to delete the following resource `/apis/operator.tigera.io/v1/tigerastatuses/apiserver`

we are getting over 17500 404s from the operator trying to delete that exact same resource over 24h, those shouldnt be present 

## Expected Behavior

0 404s in 24h

## Current Behavior

17k 404s in 24h

## Steps to Reproduce (for bugs)

1. Deploy AKS with calico using version 1.20 or above
2. look at the apiserver audit logs

tmjd:
I think https://github.com/tigera/operator/pull/1513 will address this. Though really I've moved this to doing gets instead of deletes. So if the Gets are being audited there would still be lots of logs.
Another possibility would only delete once when we know it needs to be deleted or at start, and just assume that nothing else will be creating the tigerastatus.

djsly:
in our case, we are look at 4.. and 5.. return code in general, for any verb. This helps us detect wrongly configured services or clients that might be doing things wrong.

This means that we would indeed still report it but with a different verb. 

> Another possibility would only delete once when we know it needs to be deleted or at start, and just assume that nothing else will be creating the tigerastatus.

This seems indeed like a cleaner solution, I haven't looked into the `tigerastatus` and it's lifecycle, but in the case of AKS, I"m they are not creating any. 

From looking at the Crd definition, https://docs.tigera.io/manifests/ocp/crds/01-crd-tigerastatus.yaml should we expect the status to be created after the first install and stay present to reflect the live status of the operator ?

djsly:
actually, from this comment in your code 
> // removeTigeraStatus returns true and removes the status displayed in TigeraStatus if corresponding CR not found
Could there be an issue where the CR should be present on the AKS cluster ?
I'm actually lost, since I cannot find the operator running in my cluster. 
and I only see a single CRD related to Tigera operator.
```
❯ k get crd | grep tigera
installations.operator.tigera.io                                   2021-01-19T01:37:17Z
```

No installations CR exist on the cluster.

djsly:
so I was tired and I was looking at a cluster that wasn't updated to 1.20 yet, so indeed I now have a tiger-operator namespace and the logs was showing 

it seems that the status manager isn't starting, it explains why it keeps trying to delete the status CR 

```
[...]

PR #1654, opened 2021-11-19
Don't call Delete if the CR has already been deleted

caseydavenport:
## Description

Fixes https://github.com/tigera/operator/issues/1510

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/status/status.go
+++ b/pkg/controller/status/status.go
@@ -106,10 +106,25 @@ type statusManager struct {
 	// if there are any, and report statuses based on the state of those resources.
 	readyToMonitor bool
 	hasSynced      bool
-	crExists       bool
+
+	// crExists tracks whether the status manager believes the CR to exist or not. It's used
+	// to determine whether we need to call Delete() on the object, without sending unnecessary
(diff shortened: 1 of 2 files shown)


caseydavenport:
@tmjd what do you think about this one?

pkg/controller/status/status.go
```diff
@@ -106,6 +106,7 @@ type statusManager struct {
 	// if there are any, and report statuses based on the state of those resources.
 	readyToMonitor bool
 	hasSynced      bool
+	crExists       bool
```
tmjd, line 109: Should we default this to true to ensure if the operator is restarted that an old status would get cleaned up if needed?

caseydavenport: Yeah that's probably a good idea.

tmjd: Maybe a better idea would be to query for the CR in the New call so we initialize to the real value.

tmjd (approved):
LGTM, don't forget to squash the commits.

PR #1656, opened 2021-11-22
Do not create ES generated secrets in the render

jaderhs:
## Description

This change is a bug fix to prevent an OwnerReference being added to the internal elasticsearch public certificate secret provided
This change also update two secrets constant variable names to clarify their usage.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


pkg/controller/logstorage/esgateway.go
```diff
@@ -44,7 +44,7 @@ func (r *ReconcileLogStorage) createEsGateway(
 		return reconcile.Result{}, false, nil
 	}
 
-	esInternalCertSecret, err := utils.GetSecret(ctx, r.client, relasticsearch.InternalCertSecret, render.ElasticsearchNamespace)
+	esInternalCertSecret, err := utils.GetSecret(ctx, r.client, relasticsearch.InternalPublicCertSecret, render.ElasticsearchNamespace)
```
tmjd, line 47: This name seems to contradict itself. Internal and Public are opposites in my mind.

pkg/controller/logstorage/logstorage.go
```diff
@@ -53,10 +53,11 @@ func (r *ReconcileLogStorage) createLogStorage(
 	pullSecrets []*corev1.Secret,
 	authentication *operatorv1.Authentication,
 	hdler utils.ComponentHandler,
+	hdlrPublicSecret utils.ComponentHandler,
 	reqLogger logr.Logger,
 	ctx context.Context,
 ) (reconcile.Result, bool, bool, error) {
-	var esInternalCertSecret, esCertSecret *corev1.Secret
+	var esInternalPublicCertSecret, esInternalCertSecret *corev1.Secret
```
tmjd, line 60: I don't think these name changes help, could you explain why you think these names are better?

tmjd (changes requested):
Could you please split this into 2 commits. It will be much easier reviewing name changes separately from the owner reference changes.

pkg/controller/logstorage/esgateway.go
jaderhs: I was thinkin of it as `internal` in the sense that it is used internally for components to communicate directly with elasticsearch and it should not be used/overrided by an user, and the `PublicCert` parts indicating this secret contains only the public certificate portion of a TLS certificate secret.

Maybe ESPublicCertSecret would make more sense, to indicate this is directly related to elasticsearch and not WSGateway, what do you think?
Another suggestion would be to add to the secrets that contains the key a `PrivateCert` variable name part.

pkg/controller/logstorage/logstorage.go
jaderhs: Their original names suggested to me that those were unrelated secrets (one is internal and the other is not), but it turns out their both `internal` in the sense that these should not to be overridden by the user.

My idea was to make these look like they are related, since one is the certificate portion of the other.

The `Public` is to make it clear this is the public certificate part of the secret (without the private key), though this secret is used by the components (maybe esgateway only) to communicate with elasticsearch. With the new naming I was trying to make it clearer in the sense of what they are, but it might be more appropriate to rename those to indicate their context.

What do you think?

pkg/controller/logstorage/esgateway.go
tmjd: I think I'd rather call ones that contain the key in addition to the cert to be `KeyCert` or `CertKey`.
I think Key is less ambiguous compared to Private, WDYT?

pkg/controller/logstorage/logstorage.go
tmjd: I think my suggestion from the other comment to use `KeyCert` for the secret with both and just `Cert` for the other would be better.

pkg/controller/logstorage/esgateway.go
tmjd: I'm going to retract my previous comment. I think it doesn't help because both the public and private parts are both called keys.
I do like your suggestion to add PrivateCert to the secrets that contain the private key.

I wonder if we should drop the "Internal" part of the name with this change and add "External" to secrets that might be externally provided. We should just assume secrets are Internal unless otherwise specified.
Maybe we shouldn't be trying to encode the Internal/External in the name though because after all these are just names, it is how we read and use them that determines if they are Internal/External.

pkg/render/logstorage.go
```diff
@@ -57,7 +57,7 @@ const (
 
 	ElasticsearchNamespace = "tigera-elasticsearch"
 
-	TigeraElasticsearchCertSecret         = "tigera-secure-elasticsearch-cert"
```
tmjd, line 60: I don't think we should rename this to include Gateway. It is true this secret is used by the ESGateway but it is created as the ES secret that will be externally visible.
If we want to change the name I'd say something like ExternalESCertSecretName. If we want another const for the ESGatewayCertSecret then I'd say we create a 2nd with the same value.

pkg/controller/utils/elasticsearch.go
```diff
@@ -310,7 +310,7 @@ func getClientCredentials(client client.Client, ctx context.Context) (string, st
 	}
 
 	esPublicCert := &corev1.Secret{}
-	if err := client.Get(ctx, types.NamespacedName{Name: relasticsearch.PublicCertSecret, Namespace: common.OperatorNamespace()}, esPublicCert); err != nil {
+	if err := client.Get(ctx, types.NamespacedName{Name: relasticsearch.ESGatewayPublicCertSecret, Namespace: common.OperatorNamespace()}, esPublicCert); err != nil {
```
tmjd, line 313: It seems incorrect to change this name since here this has nothing to do with the ESGateway.

pkg/controller/logstorage/esgateway.go
jaderhs: I agree on PrivateCert, that would make clear on what this variables are pointing to.

For `External`, it sounds to me that it **must** be provided instead of **could**, so I don't think we should add it (if we must identify those, something like `Overridable/Customizable` might be better, but it would make a necessary step to rename these every time there is a change if a secret may be customized or not, which does not seem ideal to me). 

Dropping `Internal` and letting the default be that secrets are internal makes sense to me.

With these changes, we would have these names as:
```
relasticsearch.InternalCertSecret -> relasticsearch.PrivateCertSecret
relasticsearch.InternalPublicCertSecret -> relasticsearch.PublicCertSecret
```

That makes sense to me in identifying the first one as the elasticsearch private secret and the second one is the public version of the same secret.

Any thoughts?

pkg/controller/utils/elasticsearch.go
jaderhs: Const name restored to `PublicCertSecret`

pkg/render/logstorage.go
jaderhs: Const name restored to `TigeraElasticsearchCertSecret`

pkg/controller/utils/elasticsearch.go
tmjd: Oh actually I guess it is related because this is the secret ESGateway presents.... ugh

pkg/controller/logstorage/logstorage.go
```diff
@@ -161,6 +162,17 @@ func (r *ReconcileLogStorage) createLogStorage(
 		return reconcile.Result{}, false, finalizerCleanup, err
 	}
 
+	// esInternalPublicCertSecret is a secret handled by ECK
```
tmjd, line 165: Is that the case with CertificateManagement also? I think the answer should be yes but I'm wondering if you confirmed it with someone that would know for sure.

pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -438,6 +438,9 @@ func (r *ReconcileLogStorage) Reconcile(ctx context.Context, request reconcile.R
 		hdler = utils.NewComponentHandler(reqLogger, r.client, r.scheme, managementClusterConnection)
 	}
 
+	//Handler for the ECK ES public secret
+	hdlrPublicSecret := utils.NewComponentHandler(reqLogger, r.client, r.scheme, nil)
```
tmjd, line 442: `hdlrPublicSecret` is what we're using it for but I think a better name would be `hdlrNoOwner`.

pkg/controller/logstorage/logstorage.go
```diff
@@ -53,10 +53,11 @@ func (r *ReconcileLogStorage) createLogStorage(
 	pullSecrets []*corev1.Secret,
 	authentication *operatorv1.Authentication,
 	hdler utils.ComponentHandler,
+	hdlrPublicSecret utils.ComponentHandler,
```
tmjd, line 56: Like my other comment I think we should call this `hndlerNoOwner`

pkg/controller/logstorage/esgateway.go
jaderhs: Restored to original names. Related naming modifications will be moved to another change.

pkg/controller/logstorage/logstorage.go
jaderhs: Restored to original names. Related naming modifications will be moved to another change.

pkg/controller/logstorage/logstorage_controller.go
jaderhs: Renamed to `hndlerNoOwner`.

pkg/controller/logstorage/logstorage.go
jaderhs: For CertificateManagement the secret I have checked with @rene-dekker that it should have an OwnerReference. Updated the component creation to make it so

pkg/controller/logstorage/logstorage.go
```diff
@@ -161,6 +162,21 @@ func (r *ReconcileLogStorage) createLogStorage(
 		return reconcile.Result{}, false, finalizerCleanup, err
 	}
 
+	// esInternalCertSecret is a secret handled by ECK
+	if esInternalCertSecret != nil {
+		esPublicSecretComponent := render.NewPassthrough([]client.Object{esInternalCertSecret})
```
rene-dekker, line 167: I suppose it is late now, but wouldn't variadic args be nicer for NewPassthrough?

```
func NewPassthrough(obects ...<object>) Component ()
```

pkg/controller/logstorage/logstorage.go
tmjd: @rene-dekker could you explain why we should put an OwnerReference on this? Does ECK operator get configured differently where it won't create the `tigera-secure-es-http-certs-public` in the `tigera-elasticsearch` namespace?

rene-dekker: If we do not create the secret, ECK will create it with the wrong contents. The contents should be installation.spec.certificatemanagement.cacert. When logstorage is removed, the secret should be garbage collected so the namespace can be terminated.

tmjd: A secret will be destroyed when a namespace is deleted so we don't really need to be worried about that.
But that is the first I've heard about ECK creating the wrong cert, we probably need to at a minimum remove the comment I added about it being an optimization on the non CertificateManagement case and replace it with a comment detailing that problem.

rene-dekker: If no certificate management is used and we do not create the public secret:

   - ECK takes the private secret and creates a public secret with just the tls.crt contents. (For the purpose of other components, like kb to mount)

If no certificate management is used no private secret is created:

   - ECK creates a private secret
   - ECK takes the private secret and creates a public secret with just the tls.crt contents. (For the purpose of other components, like kb to mount)
    
Whereas with cert management we want everything to mount installation.spec.certificatemanagement.cacert.

pkg/controller/logstorage/logstorage.go
```diff
@@ -161,6 +162,21 @@ func (r *ReconcileLogStorage) createLogStorage(
 		return reconcile.Result{}, false, finalizerCleanup, err
 	}
 
+	// esInternalCertSecret is a secret handled by ECK
+	if esInternalCertSecret != nil {
+		esPublicSecretComponent := render.NewPassthrough(esInternalCertSecret)
+		if install.CertificateManagement != nil {
```
tmjd, line 168: The behavior I understand we expect is that for CertificateManagement we need to create the secret because if we allow ECK operator to create it then it is incorrect. So it sounds like ECK operator won't overwrite the secret if it exists so it sounds like we should be owning that Secret.

Since the non-CertificateManagement case also creates this secret if it doesn't exist it sounds like a rare case that ECK will have actually created this cert. So it makes me think we should go ahead and create the Secret with the OwnerReference.
@jaderhs lets just put the OwnerReference on the secret in all cases. That is what happened before so lets keep that behavior for now.
@rene-dekker WDYT?

rene-dekker: Yes, that works.

tmjd (approved):
LGTM

PR #1662, opened 2021-11-23
Read ServiceAccount name from environment

hjiawei:
## Description

Read Tigera Operator ServiceAccount name in the following order:
1. Environment variable `OPERATOR_SERVICEACCOUNT`.
2. Pod file: `/var/run/secrets/kubernetes.io/serviceaccount/namespace`.
3. Use default name `tigera-operator`.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/common/operator_serviceaccount.go
+++ b/pkg/common/operator_serviceaccount.go
@@ -21,31 +21,32 @@ import (
 	"github.com/cloudflare/cfssl/log"
 )
 
-var namespace = ""
+var serviceAccount = ""
 
 func init() {
-	v, ok := os.LookupEnv("OPERATOR_NAMESPACE")
-	if ok {
-		namespace = v
-		return
-	}
-	body, err := ioutil.ReadFile("/var/run/secrets/kubernetes.io/serviceaccount/namespace")
-	if err != nil {
-		log.Errorf("Failed to read namespace file: %v", err)
-	} else {
-		namespace = string(body)
-		return
-	}
-
-	namespace = "tigera-operator"
+	serviceAccount = getServiceAccount()
 }
 
-// OperatorNamespace returns the namespace the operator is running in.
+// OperatorServiceAccount returns the ServiceAccount name the operator is running in.
 // The value returned is based on the following priority (these are evaluated at startup):
-//   If the OPERATOR_NAMESPACE environment variable is non-empty then that is return.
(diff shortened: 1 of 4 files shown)


pkg/render/monitor/monitor.go
```diff
@@ -57,6 +60,26 @@ const (
 	prometheusServiceAccountName = "prometheus"
 )
 
+var log = logf.Log.WithName("monitor_render")
+
+var operatorServiceAccountName = ""
+
+func init() {
```
hjiawei, line 67: @tmjd Manually tested the environment variable `OPERATOR_SERVICEACCOUNT` in the cluster and the ServiceAccount name will change accordingly. I tried to add another `init()` function in UT to set the environment variable but this `init()` runs first. I need some suggestions on how to test this.

pkg/common/operator_serviceaccount.go
```diff
@@ -0,0 +1,36 @@
+package common
```
tmjd, line 1: Please add the copyright and License to this file, could you also add it to operator_namespace.go since we missed that also.

pkg/common/operator_serviceaccount_test.go
```diff
@@ -0,0 +1,36 @@
...
+package common
+
+import (
+	"os"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+// workaround to set the environment variable before the init() function call in operator_serviceaccount.go.
```
tmjd, line 24: I didn't know this trick. Sorry I didn't understand your question previously was around testing the init. Probably another way would have been to move the code we call in the init into a function that returns the value and then for UT we could just run the function and then we wouldn't need to work around the init functionality.

tmjd (commented):
Looking good.

pkg/common/operator_serviceaccount.go
hjiawei: Copyright headers are added.

pkg/common/operator_serviceaccount_test.go
hjiawei: Removed workaround to set the env var. Test the private function instead as suggested.

tmjd (approved):
LGTM

PR #1678, opened 2021-12-03
Support kibana HA

hjiawei:
## Description

Copy InstallationSpec.ControlPlaneReplicas to Kibana CR.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/logstorage.go
+++ b/pkg/render/logstorage.go
@@ -44,6 +44,7 @@ import (
 	"github.com/tigera/operator/pkg/ptr"
 	relasticsearch "github.com/tigera/operator/pkg/render/common/elasticsearch"
 	rmeta "github.com/tigera/operator/pkg/render/common/meta"
+	"github.com/tigera/operator/pkg/render/common/podaffinity"
 	"github.com/tigera/operator/pkg/render/common/podsecuritypolicy"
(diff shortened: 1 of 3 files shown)


pkg/render/logstorage.go
```diff
@@ -1241,7 +1243,7 @@ func (es elasticsearchComponent) kibanaCR() *kbv1.Kibana {
 			Config: &cmnv1.Config{
 				Data: config,
 			},
-			Count: 1,
+			Count: *es.cfg.Installation.ControlPlaneReplicas,
```
caseydavenport, line 1246: You might get a nil pointer exception here - unlikely, since this field should be defaulted, but a safer thing to do would be to check if this field is nil and set `Count: 0` if it is.

hjiawei: Add a check for nil pointer and default the count to 1 as before. I think we need to have at least 1 kibana deployment.

caseydavenport: 1 is probably fine. I suggested 0 because it is more obviously an error scenario that can be asserted in tests, etc, but I don't really mind.

caseydavenport (approved):

Issue #1385, opened 2021-07-12
Allow objects to have custom labels


Symbianx:
Looks like it's impossible right now to create the `calico-system` namespace with a custom label.

## Expected Behavior

Tigera operator should merge the labels of an existing namespace (or any other resource) with the wanted labels.

## Current Behavior

Tigera operator ignores labels in the existing namespace, overriding them with the wanted ones.

## Possible Solution

Annotations are already merged in the `mergeState` util function, we could easily do the same for labels: https://github.com/tigera/operator/blob/756210bcace3e961f3ba67f8e795fd78617f8167/pkg/controller/utils/component.go#L214-L217

## Steps to Reproduce (for bugs)

1. Create the `calico-system` namespace
2. Label the namespace with `openpolicyagent.org/webhook: ignore`
3. Install tigera-operator and wait for it to start creating resources
4. The label `openpolicyagent.org/webhook` is gone.

## Context

We're trying to install the calico using the tigera operator. However, tigera is having issues creating deployments in the `calico-system` namespace because we have OPA running and configured to block deployment creations without the `app.kubernetes.io/name` label.
We were already expecting this to happen and created the `calico-system` namespace beforehand with the label `openpolicyagent.org/webhook: ignore` (namespaces labeled bypass the OPA admission webhook).

However, that label is removedby the tigera-operator.

## Your Environment

* AWS Configuration (policy only mode)
* OPA running

Symbianx:
If the possible solution is accepted by the maintainers, I can provide a PR for it.

tmjd:
I think I'd like to see the operator updated to add the `app.kubernetes.io/name` since that is the root of why you are looking to merge labels on the namespace.
I think that label is the direction that k8s is moving so it would be good to make that change.
@caseydavenport WDYT? Is there any reason we wouldn't want to start including that new label? 
I assume we'd keep the `k8s-app` we already have and add `app.kubernetes.io/name`.

Symbianx:
I think the operator having the standard labels definetely makes sense and should be there.
I also think we should still give the ability for people to label the `calico-system` namespace. 

In my case, I want to disable OPA for the calico namespace, it's good practice to disable OPA for core system namespace like `kube-system`. The calico namespace manages internal cluster networking which makes it very much a core namespace and could cause whole cluster failures if OPA starts denieing API requests for it. For instance, a misconfigured OPA policy could deny new `calico-node` pods to start up which would not be good.

caseydavenport:
[...]

PR #1679, opened 2021-12-05
Component labeler

tmjd:
## Description

I wanted to add the new label (`app.kubernetes.io/name`) talked about in #1385 but didn't want to manually add it everywhere so I took a look at the Deployments and DaemonSets and found that they were very consistent and took this approach.
WDYT?

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/utils/component.go
+++ b/pkg/controller/utils/component.go
@@ -116,6 +116,9 @@ func (c componentHandler) CreateOrUpdateOrDelete(ctx context.Context, component 
 		// Make sure any objects with images also have an image pull policy.
 		modifyPodSpec(obj, setImagePullPolicy)
 
+		// Make sure we have our standard selector and pod labels
+		setStandardSelectorAndLabels(obj)
+
 		// Keep track of some objects so we can report on their status.
 		switch obj.(type) {
 		case *apps.Deployment:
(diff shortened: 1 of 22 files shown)


pkg/render/apiserver.go
```diff
@@ -785,22 +785,21 @@ func (c *apiServerComponent) apiServerDeployment() *appsv1.Deployment {
...
 				"apiserver": "true",
-				"k8s-app":   name,
 			},
 		},
 		Spec: appsv1.DeploymentSpec{
 			Replicas: c.cfg.Installation.ControlPlaneReplicas,
 			Strategy: appsv1.DeploymentStrategy{
 				Type: appsv1.RecreateDeploymentStrategyType,
 			},
+			// Why do we use apiserver: true instead of k8s-app: name?
```
caseydavenport, line 795: I doubt anyone knows at this point.

pkg/controller/utils/component.go
```diff
@@ -361,6 +363,48 @@ func ensureOSSchedulingRestrictions(obj client.Object, osType rmeta.OSType) {
...
+			}
+		}
+		podTemplate = &d.Spec.Template
+	case *apps.DaemonSet:
+		d := obj.(*apps.DaemonSet)
+		name = d.ObjectMeta.Name
+		if d.Spec.Selector == nil {
+			d.Spec.Selector = &metav1.LabelSelector{
+				MatchLabels: map[string]string{
+					"k8s-app": name,
```
caseydavenport, line 392: Do we not want the app.kubernetes.io label here as well?

caseydavenport (commented):
My gut reaction is that maybe it's a little bit too magical, and we're still going to need a way for users to specify their own labels and annotations, which will need to be validated against the ones that the operator puts on by default.

But, I do like the general direction and a little bit of magic might be appropriate here (and some enforced consistency as well!) 

I think some comments explaining the pattern that we're enforcing would help a bit.

e.g., 

"For deployments, we ensure that they have labels X,Y,Z, that their pod selector is X, and that the pod template always has A,B,X."

pkg/render/apiserver.go
tmjd: :grinning: I figured that was probably the case.

caseydavenport:
@tmjd I was thinking a general code comment explaining what labels we apply to every Deployment / Daemonset, but maybe the code is self-explanatory. idk.

tmjd:
I'm not suggesting one is not needed, I'm just not sure where it would be useful. On the CreateOrUpdateOrDelete call?

caseydavenport:
I was imagining it would be at the top of `setCommonFields` (which maybe could be reduced in scope to "setStandardLabels" or something like that.

caseydavenport:
@tmjd should we bring this guy up to date?

tmjd:
I've updated this.
One notable change is I changed the setting of the labels to always be done by the setStandardSelectorAndLabels function, not only when the ObjectMeta.Labels was nil.

pkg/controller/utils/component.go
```diff
@@ -407,6 +410,51 @@ func ensureOSSchedulingRestrictions(obj client.Object, osType rmeta.OSType) {
...
+	var podTemplate *v1.PodTemplateSpec
+	var name string
+	switch obj.(type) {
+	case *apps.Deployment:
+		d := obj.(*apps.Deployment)
+		name = d.ObjectMeta.Name
+		if d.ObjectMeta.Labels == nil {
+			d.ObjectMeta.Labels = make(map[string]string)
+		}
+		d.ObjectMeta.Labels["k8s-app"] = name
```
caseydavenport, line 426: Guess we need to keep this around for now, but wondering if we should eventually migrate off it to the app.kubernetes.io label and remove this altogether.

pkg/render/apiserver.go
caseydavenport: I think we should remove this comment, though. Or at least update it to say:

```
// For legacy reasons we use apiserver: true here instead of the k8s-app: name label, so we need to set it explicitly rather than use the common labeling logic. 
```

caseydavenport (approved):
Couple minor comments, and a rebase, but otherwise lets get this in.

pkg/controller/utils/component.go
tmjd: We certainly need to keep it until we switch over the matchSelector and I'm not sure if we can just switch that over on upgrade. I don't see much harm in keeping it.

PR #1683, opened 2021-12-06
Update k8s dependencies to 0.21.7, update base build images, update c…

Brian-McM:
…ompiled go version (by updating the calico go build image)

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


tmjd (approved):
What is the plan for these changes are they going to be picked back to older versions? Because they seem fairly significant and should have significant testing before we release them.

tmjd (changes requested):
The tests failed on some tests that mention finalizers. I looked at the controller-runtime releases and one of the versions did mention changes with Finalizer so it seems like there are other changes needed here

pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -531,7 +531,10 @@ func (r *ReconcileLogStorage) Reconcile(ctx context.Context, request reconcile.R
 
 	r.status.ClearDegraded()
 
-	if ls != nil {
+	// Since we don't re poll for the object we need to make sure the object wouldn't have been deleted on the patch
+	// that may have removed the finalizers.
+	// TODO We may want to just return if we remove the finalizers from the LogStorage object.
+	if ls != nil && (ls.DeletionTimestamp == nil || len(ls.GetFinalizers()) > 0) {
```
tmjd, line 537: Is the `len(ls.GetFinalizers()) > 0` needed? It seems un-needed.

Brian-McM: It's needed because if the deletion timestamp is set but there are still finalizers the object isn't deleted so we may want to still set a status.

tmjd (approved):
LGTM

PR #1700, opened 2021-12-15
Improve git performance

rene-dekker:
Shallow clone the calico repo when fetching crds.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -525,9 +525,7 @@ define fetch_crds
     $(eval branch := $(2))
     $(eval dir := $(3))
 	@echo "Fetching $(dir) CRDs from $(project) branch $(branch)"
-	git -C .crds/$(dir) clone git@github.com:$(project).git ./
-	git -C .crds/$(dir) fetch --all origin 2>&1 | grep -v -e "new branch" -e "new tag"
-	git -C .crds/$(dir) checkout -q $(branch)
+	git -C .crds/$(dir) clone --depth 1 --branch $(branch) --single-branch git@github.com:$(project).git ./
 endef
 define copy_crds
     $(eval dir := $(1))


config/calico_crds.yml
```diff
@@ -0,0 +1,6 @@
+# The Makefile goals for read-calico-crds-version and read-calico-enterprise-crds-version read from here what branch
```
caseydavenport, line 1: This is probably good as a temporary fix but we will likely want the ability to pin to a release version once we have releases built from the new monorepo.

What do you think @tmjd ?

caseydavenport (approved):

caseydavenport (approved):

config/calico_crds.yml
tmjd: I don't think this is a good fix because it is adding a new file for versions when we already have files for calico and enterprise versions. @rene-dekker why do we need to add a new file for this?
If we need a quick fix I think the only thing that needs to change are the git commands used.

config/calico_crds.yml
```diff
@@ -0,0 +1,6 @@
...
+# on the calico repo to use for downloading the ./libcalico CRDs.
+calico:
```
tmjd, line 3: I think we can continue to use the current calico and enterprise versions files. I'd even suggest keeping the libcaclico-go name as I think changing the field will have other knock on effects to the build and release pipelines.

tmjd (changes requested):

config/calico_crds.yml
rene-dekker: I can revert these changes. It will work for master, but if you want to build release-v1.24, the libcalico-go version will point to v3.21.0, while that tag/branch does not exist in projectcalico/calico. So whoever builds that will probably want to tweak it into something like I propose to get un-stuck.

Want me to revert, so we can unblock master?

tmjd: I think in release-v1.24 it would need to be updated to point to the correct branch. Are you saying there is no v3.21.0 branch, I just checked and there is.

tmjd: Yes please revert adding this new file.

rene-dekker: I kind of misremembered. The branch indeed exists, but it has no monorepo stuff yet: https://github.com/projectcalico/calico/tree/v3.21.0

rene-dekker: I reverted the changed, other than the git commands for now.

tmjd (approved):
LGTM

PR #1704, opened 2021-12-20
Support for autodetection Calico node IP from k8s InternalIP

caseydavenport:
## Description

Adds config support for the featuer added in this PR: https://github.com/projectcalico/node/pull/1242

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/api/v1/installation_types.go
+++ b/api/v1/installation_types.go
@@ -370,7 +370,7 @@ type NodeAddressAutodetection struct {
 
 	// Kubernetes configures Calico to detect node addresses based on the Kubernetes API.
 	// +optional
-	// +kubebuilder:validation:Enum=InternalIP
+	// +kubebuilder:validation:Enum=NodeInternalIP
 	Kubernetes *KubernetesAutodetectionMethod `json:"kubernetes,omitempty"`
 
 	// Interface enables IP auto-detection based on interfaces that match the given regex.
--- a/api/v1/installation_types.go
+++ b/api/v1/installation_types.go
@@ -368,6 +368,11 @@ type NodeAddressAutodetection struct {
 	// +optional
 	FirstFound *bool `json:"firstFound,omitempty"`
 
+	// Kubernetes configures Calico to detect node addresses based on the Kubernetes API.
+	// +optional
+	// +kubebuilder:validation:Enum=InternalIP
+	Kubernetes *KubernetesAutodetectionMethod `json:"kubernetes,omitempty"`
+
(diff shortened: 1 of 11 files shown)


api/v1/installation_types.go
```diff
@@ -368,6 +368,11 @@ type NodeAddressAutodetection struct {
 	// +optional
 	FirstFound *bool `json:"firstFound,omitempty"`
 
+	// Kubernetes configures Calico to detect node addresses based on the Kubernetes API.
```
caseydavenport, line 371: @tmjd @Brian-McM what do you think about this API? Please give it a critical eye!

The reason behind not having it as a bool is that we might want to extend this to other options in the future. For example. any of these: https://kubernetes.io/docs/concepts/architecture/nodes/#addresses

Or our own modification - e.g., `NodeCalicoIP` - which looks for an address in the list with type `CalicoIP`

api/v1/installation_types.go
```diff
@@ -368,6 +368,11 @@ type NodeAddressAutodetection struct {
...
+	// +optional
+	// +kubebuilder:validation:Enum=InternalIP
```
tmjd, line 373: This validation does not match `NodeInternalIP`

api/v1/installation_types.go
tmjd: I'm always a fan of not using a bool in the API, I think that's a good choice.
`kubernetes` seems very broad/generic but I think the possible values (and the comments on them) should clarify any ambiguity.

pkg/controller/migration/convert/network.go
```diff
@@ -425,13 +426,21 @@ func handleAutoDetectionMethod(c *components, install *operatorv1.Installation)
...
+	// cidr=
 	if strings.HasPrefix(*method, AutodetectionMethodCIDR) {
 		ifStr := strings.TrimPrefix(*method, AutodetectionMethodCIDR)
 		cidrs := strings.Split(ifStr, ",")
 		install.Spec.CalicoNetwork.NodeAddressAutodetectionV4 = &operatorv1.NodeAddressAutodetection{CIDRS: cidrs}
 		return nil
 	}
 
+	// kubernetes-internal-ip
+	if *method == "" || *method == AutodetectionMethodNodeIP {
```
tmjd, line 438: Nice job adding migration code. WDYT about adding a test for it?

tmjd (commented):
Looks good, one change needed, maybe another test but it mainly looks good.

PR #1706, opened 2021-12-21
Update Elasticsearch and Kibana to v7.16.2

rene-dekker:


doublek (approved):
LGTM

rene-dekker:
Should not be merged before we have the Kibana and ES images.

PR #1710, opened 2021-12-22
Improving teardown

tmjd:
The CI tests have been flakey recently, these changes add some
finalizers to some Calico resources to ensure the CNI plugin can
function long enough to tear down kube-controllers.

There are associated test changes that adjust he FV tests to allow
operator to teardown and remove the finalizers so everything from
a test can be cleaned up correctly.

## Description

The changes add a finalizer to the Installation resource and to the Calico node ServiceAccount, ClusterRole, and ClusterRoleBinding. The finalizers allow the operator to know when the Installation resource is being torn down so then it can wait for kube-controllers to be removed before allowing the ServiceAccount, ClusterRole, and ClusterRoleBinding from being removed. This allows the CNI plugin to tear down the networking for kube-controllers which allows the pod to be removed.

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/installation/core_controller.go
+++ b/pkg/controller/installation/core_controller.go
@@ -27,12 +27,34 @@ import (
 	"strings"
 	"time"
 
-	"github.com/tigera/operator/pkg/render/kubecontrollers"
-
+	"github.com/elastic/cloud-on-k8s/pkg/utils/stringsutil"
+	"github.com/go-logr/logr"
+	configv1 "github.com/openshift/api/config/v1"
+	"github.com/openshift/library-go/pkg/crypto"
+	apps "k8s.io/api/apps/v1"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/cache"
 	apiregv1 "k8s.io/kube-aggregator/pkg/apis/apiregistration/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
 
 	operator "github.com/tigera/operator/api/v1"
 	"github.com/tigera/operator/pkg/active"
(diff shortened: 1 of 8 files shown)


pkg/render/node.go
```diff
@@ -161,56 +165,67 @@ func (c *nodeComponent) SupportedOSType() rmeta.OSType {
 }
 
 func (c *nodeComponent) Objects() ([]client.Object, []client.Object) {
```
Brian-McM, line 167: I may be confused by these changes because I don't understand how this works in k8s, but do you actually need to / should you explicitly delete anything here if we're terminating?

If the installation resource has been deleted, everything that doesn't have a finalizer will be automatically removed, and everything with a finalizer will be kept until the finalizers are removed.

I think you might introduce a weird race by returning objects to create as well. If kube controllers is deleted and this reconciles then the finalizers will be removed _and_ we might try to update / create these resources, which could lead to some funky behaviour.

I would think the best thing to do here is to return nothing if we're terminating...

pkg/render/kubecontrollers/kube-controllers.go
```diff
@@ -242,6 +244,11 @@ func (c *kubeControllersComponent) Objects() ([]client.Object, []client.Object)
 		objectsToDelete = append(objectsToDelete, c.prometheusService())
 	}
 
+	if c.cfg.Terminating {
```
Brian-McM, line 247: Do you need to explicitly delete this if the Installation resource has been terminated? It should be deleted on it's own if it doesn't have a finalizer.

pkg/render/node.go
tmjd: If we're terminating then we need to re-render (update) the resources that remained due to the finalizers on them previously so they can be cleaned up. Specifically because prior to "terminating": the SA, ClusterRoleBinding, and ClusterRole had a finalizer to prevent their deletion.

pkg/controller/status/status.go
```diff
@@ -698,8 +699,9 @@ func (m *statusManager) set(retry bool, conditions ...operator.TigeraStatusCondi
 
 	// Update the object in the API, creating it if necessary.
 	if isNotFound {
-		if err = m.client.Create(context.TODO(), &ts); err != nil {
-			log.WithValues("reason", err).Info("Failed to create tigera status")
+		if err = m.client.Create(context.TODO(),
```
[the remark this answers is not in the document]
tmjd: I'm not sure, I think this was maybe incorrectly added. I think I might have been doing this when trying something else out.

pkg/controller/installation/core_controller.go
```diff
@@ -748,6 +748,33 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
...
+			return reconcile.Result{}, err
+		}
+		found := false
+		for _, x := range crb.Finalizers {
+			if x == render.NodeFinalizer {
+				found = true
+			}
+		}
+		if !found {
+			// We don't remove the finalizer from the ClusterRoleBinding until we're terminating
```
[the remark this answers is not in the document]
tmjd: Instead of trying to comment on one part of the cleanup sequence here I added a big comment at the top of this file on the sequence.

pkg/render/node.go
Brian-McM: Oh right, I was (incorrectly) thinking the finalizer was removed separate to this.

pkg/render/kubecontrollers/kube-controllers.go
Brian-McM: > I was surprised by the behavior that these resources did automatically start getting cleaned up because the Installation resource still existed and it just had a DeletionTimestamp.

Maybe it's just perspective, but that didn't / doesn't surprise me, that's what I would have hoped would happen.

I feel like passing back these resources to be cleaned up is a slight change in the point of delete resource logic (at least what I thought the point was). The delete resource logic so far is used to clean up resources no longer needed because of a configuration change, but now it's being used to aid kubernetes cleanup, which we previously left alone. This will at the very least cause a division and possible confusion as to when we pass back these objects to be deleted (Logstorage controller uses finalizers but doesn't clean up the resource like this).

I won't argue to much against this, but in my opinion it would be best to just let kubernetes handle the cleanup naturally.

pkg/controller/installation/core_controller.go
```diff
@@ -748,6 +774,31 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
...
+		// (and ClusterRole and ServiceAccount) is removed.
+		crb := rbacv1.ClusterRoleBinding{}
+		crbKey := types.NamespacedName{Name: "calico-node"}
+		err := r.client.Get(ctx, crbKey, &crb)
+		if err != nil && !apierrors.IsNotFound(err) {
+			r.SetDegraded("Unable to get ClusterRoleBinding", err, reqLogger)
+			return reconcile.Result{}, err
+		}
+		found := false
+		for _, x := range crb.Finalizers {
```
Brian-McM, line 789: Thinking on this, it seems pretty unnecessary to check for the finalizer. If the finalizer is remove then the resource will be deleted.

Brian-McM (approved):

Brian-McM (approved):

PR #1711, opened 2021-12-23
Update elastic v7.16.2

rene-dekker:
Use the new Elastic images. If Authentication CR is provided, we set the base url.

The commits are already present in v1.22. #1706


doublek (approved):
LGTM

caseydavenport (approved):

PR #1729, opened 2022-01-13
Add component for the eck controller version.

rene-dekker:
We no longer use the elastic registry.
The controller-version for eck components needs to be set to the original eck versions, while the image versions need to be set to our own image tags.


pkg/components/references.go
```diff
@@ -43,8 +43,6 @@ func GetReference(c component, registry, imagePath, imagePrefix string, is *oper
 			ComponentWindows:
 
 			registry = CalicoRegistry
-		case ComponentElasticsearchOperator:
```
rene-dekker, line 46: @doublek Is this correct or should I revert?

config/enterprise_versions.yml
```diff
@@ -113,6 +113,9 @@ components:
   # The components below are third-party images that have been retagged under
   # quay.io/tigera so all enterprise images come from the same repository and org.
   elasticsearch-operator:
+    image: tigera/eck-operator
+    version: master
+  eck-elasticsearch-operator:
     image: tigera/eck-operator
```
doublek, line 119: I think in other ones above like `eck-kibana` we don't add the image. There is a default that is set elsewhere. So maybe we could do just that.

doublek (approved):
LGTM

bcreane:
@rene-dekker do you anticipate either upgrade or version skew issues resulting from this fix?

marvin-tigera:

rene-dekker:
> o you anticipate either upgrade or version skew issues resulting from this fix?

In terms of version skew it should be the same as what master was before this PR. ECK 1.8.0 is compatible with ES&KB 7.16.2. In terms of MCM version skew the recent upgrades to 7.16.2 does not change the api for the calls that fluentd and other components should make to ES/KB as far as we are currently aware.

doublek (approved):
LGTM

Issue #1653, opened 2021-11-19
The `controlPlaneNodeSelector` installation field doesn't effect Typha


stevehipwell:
## Expected Behavior

Based on the documentation for [controlPlaneNodeSelector](https://docs.projectcalico.org/reference/installation/api#operator.tigera.io/v1.InstallationSpec) it applies to all components which aren't DaemonSets. That means that it should apply to the _Typha_ deployment.

## Current Behavior

The `controlPlaneNodeSelector` doesn't apply to the _Typha_ deployment. ~~I suspect this might be because there is the `typhaAffinity` field, but affinity and node selectors can be used in parallel.~~

## Possible Solution

Use `controlPlaneNodeSelector` with the _Typha_ deployment.

## Steps to Reproduce (for bugs)

n/a

## Context

n/a

## Your Environment

n/a

tmjd:
Yeah we should state that Typha is an exception because it is a critical component and should be considered part of the dataPlane.

tmjd:
> @tmjd isn't that more reason to allow the nodes it runs on to be selected?

I don't think so, it isn't critical that individual Typhas are not destroyed/recreated. It is critical that a sufficient number exist, which is why there is a PodDisruptionBudget defined for it and we suggest only preferred(not required) typaAffinity. This will allow typha to always be present at the correct scale since it is not forced to any specific nodes.  So as long as any node scaling properly follows PodDisrutpionBudgets then Typha is fine.

If you still would like to prefer system nodes for typha you can set that with [spec.typhaAffinity](https://docs.projectcalico.org/reference/installation/api#operator.tigera.io/v1.InstallationSpec).

> If Typha isn't using controlPlaneNodeSelector then there either needs to be a typhaNodeSelector or dataPlaneNodeSelector to control this.

I do not believe this is necessary because typha's scaling and PodDisruptionBudget are set to ensure HA when nodes are removed or fail.
If you still believe you need to force typha to certain nodes checkout typhaAffinity which is of type [NodeAffinity](https://docs.projectcalico.org/reference/installation/api#operator.tigera.io/v1.NodeAffinity). But please take note of the warning on requiredDuringSchedulingIgnoredDuringExecution, which is

[...]

PR #1725, opened 2022-01-10
Do not run calico-node on AWS EKS Fargate nodes

niranjan94:
## Description

Resolves https://github.com/projectcalico/calico/issues/4731

[AWS Fargate with Amazon EKS](https://docs.aws.amazon.com/eks/latest/userguide/fargate.html) is an amazon service that allows users to run a kubernetes cluster without any standard nodes and rather use AWS Fargate (a serverless container service provided by AWS under ECS) to run the pods. These fargate nodes can only run a single pod and have many limitations as compared to normal nodes making it impossible for calico-node pods to run on them. 

The node is tainted to prevent pods like typha and kube-controllers from running there, but Calico-node tolerates all taints. And this results in many pending calico pods that can never run on those nodes.

This PR sets a node affinity to prevent calico pods from being scheduled on these fargate nodes. The PR has been implemented very similar to https://github.com/tigera/operator/pull/1048 which does the same but for AKS's virtual nodes.

```yaml
affinity:
  nodeAffinity:
    requiredDuringSchedulingIgnoredDuringExecution:
      nodeSelectorTerms:
      - matchExpressions:
        - key: eks.amazonaws.com/compute-type
          operator: NotIn
          values:
          - fargate
```

## For PR author

- [x] Tests for change.
- [ ] If changing pkg/apis/, run `make gen-files`
- [ ] If changing versions, run `make gen-versions`

## For PR reviewers

A note for code reviewers - all pull requests must have the following:

- [ ] Milestone set according to targeted release.
- [ ] Appropriate labels:
  - `kind/bug` if this is a bugfix.
  - `kind/enhancement` if this is a a new feature.
  - `enterprise` if this PR applies to Calico Enterprise only.

PR #1732, opened 2022-01-14
Add affinities and node selectors for daemonsets, control plane objects, and typha

mgleung:
## Description

Adds node selector and affinity logic to daemonset components in the operator. Also adds affinity for control plane objects and node selectors for Typha.

This should fix https://github.com/tigera/operator/issues/1653

This should be dependent on https://github.com/tigera/operator/pull/1725 also since the EKS fargate affinities haven't been merged yet.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/api/v1/installation_types.go
+++ b/api/v1/installation_types.go
@@ -95,6 +95,16 @@ type InstallationSpec struct {
 	// +optional
 	TyphaAffinity *TyphaAffinity `json:"typhaAffinity,omitempty"`
 
+	// TyphaNodeSelector is used to select the nodes on which to run Typha pods.
+	// +optional
+	TyphaNodeSelector map[string]string `json:"typhaNodeSelector,omitempty"`
+
+	// ControlPlaneAffinity allows configuration of node affinity characteristics for Calico
+	// components scheduled on control plane nodes. This is globally applied to all resources created
+	// by the operator excluding daemonsets.
+	// +optional
+	ControlPlaneAffinity *v1.Affinity `json:"controlPlaneAffinity,omitempty"`
+
 	// ControlPlaneNodeSelector is used to select control plane nodes on which to run Calico
 	// components. This is globally applied to all resources created by the operator excluding daemonsets.
 	// +optional
@@ -145,6 +155,15 @@ type InstallationSpec struct {
 	// NonPrivileged configures Calico to be run in non-privileged containers as non-root users where possible.
 	// +optional
 	NonPrivileged *NonPrivilegedType `json:"nonPrivileged,omitempty"`
+
+	// DaemonsetAffinity allows configuration of node affinity characteristics for all Calico daemonsets (i.e. calico-node).
+	// In EKS this defaults to avoiding Fargate nodes. In AKS this defaults to avoiding virtual nodes.
+	// +optional
+	DaemonSetAffinity *v1.Affinity `json:"daemonSetAffinity,omitempty"`
+
+	// DaemonsetNodeSelector is used to select nodes on which to run all Calico daemonsets (i.e. calico-node).
+	// +optional
+	DaemonSetNodeSelector map[string]string `json:"daemonSetNodeSelector,omitempty"`
 }
 
 // TyphaAffinity allows configuration of node affinitiy characteristics for Typha pods.
--- a/api/v1/logstorage_types.go
+++ b/api/v1/logstorage_types.go
@@ -42,6 +42,11 @@ type LogStorageSpec struct {
 	// +optional
 	StorageClassName string `json:"storageClassName,omitempty"`
 
+	// DataAffinity allows configuration of the affinity characteristics for Elasticsearch to provide more control over
+	// where Elasticsearch will run. The contents of DataAffinity will be added to the PodSpec of the Elasticsearch nodes.
+	// +optional
+	DataAffinity *corev1.Affinity `json:"dataAffinity,omitempty"`
+
 	// DataNodeSelector gives you more control over the node that Elasticsearch will run on. The contents of DataNodeSelector will
 	// be added to the PodSpec of the Elasticsearch nodes. For the pod to be eligible to run on a node, the node must have
 	// each of the indicated key-value pairs as labels as well as access to the specified StorageClassName.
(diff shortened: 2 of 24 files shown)


caseydavenport:
@mgleung @tmjd what do you guys think about resurrecting this and getting it in? Seems like a useful thing to have.

api/v1/logstorage_types.go
```diff
@@ -42,6 +42,11 @@ type LogStorageSpec struct {
 	// +optional
 	StorageClassName string `json:"storageClassName,omitempty"`
 
+	// DataAffinity allows configuration of the affinity characteristics for Elasticsearch to provide more control over
+	// where Elasticsearch will run. The contents of DataAffinity will be added to the PodSpec of the Elasticsearch nodes.
+	// +optional
+	DataAffinity *corev1.Affinity `json:"dataAffinity,omitempty"`
```
Brian-McM, line 48: I think you should look at the NodeSets definition. We already have affinity for elasticsearch it's just done a little differently. It also ensures that index replicas are spread across the nodes.

pkg/controller/installation/validation.go
```diff
@@ -361,6 +362,22 @@ func validateCustomResource(instance *operatorv1.Installation) error {
 		}
 	}
 
+	// Verify that daemonset affinity ensures that they are not run on virtual nodes (AKS) or fargate nodes (EKS).
+	if instance.Spec.DaemonSetAffinity != nil {
```
tmjd, line 366: previously when we've provided an option that would allow removing a defaulted configuration, we do not validate that the configuration retains the defaulted value.
In other words, I don't think we need the validation for either AKS or EKS.

tmjd: @caseydavenport WDYT?

pkg/controller/migration/convert/core.go
```diff
@@ -305,38 +305,36 @@ func removeExpectedAnnotations(existing, ignoreWithValue map[string]string, toBe
...
-							Operator: corev1.NodeSelectorOpNotIn,
-							Values:   []string{"virtual-kubelet"},
-						}},
-					}},
-				},
-			},
-		}) {
+		var err error
+		if install.Spec.KubernetesProvider == operatorv1.ProviderAKS {
+			if aksErr := checkAKSAffinity(c.node.Spec.Template.Spec.Affinity); aksErr != nil {
```
tmjd, line 310: Similar to my validation comment, but since we support configuration of affinity with this we can just take whatever they had set previously. It worked for them before so it should still work after migration.

tmjd (commented):
I wonder if since we say the controlPlane and Daemonset fields apply to all Deployments or Daemonsets if we should have the ComponentHandler add them and only set affinity/nodeSelectors on depoyments and daemonsets that need something different (like typha).
That would ensure no new components are added without those selectors and save having that code duplicated.

api/v1/installation_types.go
```diff
@@ -95,6 +95,16 @@ type InstallationSpec struct {
 	// +optional
 	TyphaAffinity *TyphaAffinity `json:"typhaAffinity,omitempty"`
 
+	// TyphaNodeSelector is used to select the nodes on which to run Typha pods.
+	// +optional
+	TyphaNodeSelector map[string]string `json:"typhaNodeSelector,omitempty"`
```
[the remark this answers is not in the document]
caseydavenport: "do not use" is probably too strong - there are legitimate reasons to use this. But a warning of the form "If used, you must ensure there is sufficient room for Typha to be scheduled or else your cluster may stop functioning" is probably appropriate.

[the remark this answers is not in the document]
caseydavenport: In theory we could build in a release valve - if the number of nodes selected by this is not large enough, we error and ignore the selector. 

I don't think it's quite worth doing, but if we end up getting a lot of issues where people break their clusters due to misconfiguring this field, we can look into it.

pkg/controller/installation/validation.go
caseydavenport: There's definitely no rule we need to validate that the config matches the previous defaults. But if we _know_ there are config options here that won't work on certain types of clusters, I think it's fair to validate against that.

That said, the virtual nodes can be avoided using other means than an affinity, so it is probably valid for this field to _not_ include those. This validation might be a tiny bit too strict.

caseydavenport: There's definitely no rule we need to validate that the config matches the previous defaults. But if we _know_ there are config options here that won't work on certain types of clusters, I think it's fair to validate against that.

That said, the virtual nodes can be avoided using other means than an affinity, so it is probably valid for this field to _not_ include those. This validation might be a tiny bit too strict.

pkg/controller/migration/convert/core_test.go
```diff
@@ -109,9 +109,9 @@ var _ = Describe("core handler", func() {
 
 	Context("nodeSelector", func() {
 		TestNodeSelectors := func(f func(map[string]string)) {
-			It("should error for unexpected nodeSelectors", func() {
+			It("should not error for unexpected nodeSelectors", func() {
```
caseydavenport, line 112: "unexpected" ?

aquam8:
Please add support for `tolerations` as part of this too - currently Node (calico-node) DS does not get custom tolerations when deployed with tigera-operator v3.22.2 AFAIK

Thank you

api/v1/logstorage_types.go
lmm: I see. So we should remove this since affinity for Elasticsearch is handled already.

api/v1/installation_types.go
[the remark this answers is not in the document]
tmjd: Here is the warning we have with the TyphaAffinity.RequiredDuringScheduling...
```
    // WARNING: Please note that if the affinity requirements specified by this field are not met at
    // scheduling time, the pod will NOT be scheduled onto the node.
    // There is no fallback to another affinity rules with this setting.
    // This may cause networking disruption or even catastrophic failure!
    // PreferredDuringSchedulingIgnoredDuringExecution should be used for affinity
    // unless there is a specific well understood reason to use RequiredDuringSchedulingIgnoredDuringExecution and
    // you can guarantee that the RequiredDuringSchedulingIgnoredDuringExecution will always have sufficient nodes to satisfy the requirement.  
```
I think we should have something similar here. I'm a bit opinionated here but if we don't want to include something that ominous for this one then I think we should reduce the warning on the other to the same level as what we use here.

mohamed-elmadeny (approved):

caseydavenport:
I think this PR was obsoleted by https://github.com/tigera/operator/pull/2063

PR #1739, opened 2022-01-19
Operator changes for WAF (Web Application Firewall).

bartolini:
## Description

The changes to the operator to support ModSecurity based Web Application Firewall.

## For PR author

- [ ] Tests still to follow
- [ ] Core ruleset configuration man still to follow

## For PR reviewers

- [ ] Appropriate labels:


api/v1/applicationlayer_types.go
```diff
@@ -22,15 +22,20 @@ import (
 
 // ApplicationLayerSpec defines the desired state of ApplicationLayer
 type ApplicationLayerSpec struct {
+	// Specification for WAF (Web Application Firewall).
```
caseydavenport, line 25: I know the other comment is bare-bones as well, but remember this is user-facing documentation so we really need more description for this.

"WebApplicationFirewall controls whether or not ModSecurity enforcement is enabled for the cluster. When enabled, Services may opt-in to having ingress traffic examed by ModSecurity"

Or something like that.

config/samples/operator_v1_applicationlayer.yaml
```diff
@@ -3,5 +3,7 @@ kind: ApplicationLayer
 metadata:
   name: tigera-secure
 spec:
+  wafSettings:
```
caseydavenport, line 6: This is out of date

pkg/controller/applicationlayer/applicationlayer_controller.go
```diff
@@ -105,8 +107,21 @@ func add(mgr manager.Manager, c controller.Controller) error {
 		return fmt.Errorf("applicationlayer-controller failed to watch Tigera network resource: %v", err)
 	}
 
-	// Watch configmaps created for envoy.
-	for _, configMapName := range []string{applicationlayer.EnvoyConfigMapName} {
+	// Watch for configmap changes in tigera-operator namespace; the cm contains ruleset for ModSecurity library:
+	err = utils.AddConfigMapWatch(c, applicationlayer.ModSecurityRulesetConfigMapName, common.OperatorNamespace())
```
caseydavenport, line 111: Looks like we're adding a watch for this twice? We should only do it once.

pkg/controller/applicationlayer/applicationlayer_controller.go
```diff
@@ -277,12 +308,71 @@ func updateApplicationLayerWithDefaults(al *operatorv1.ApplicationLayer) {
...
 func validateApplicationLayer(al *operatorv1.ApplicationLayer) error {
-	lcSpec := al.Spec.LogCollection
 
-	// If ApplicationLayer spec exists then LogCollection should be set.
-	// TODO: when we will have multiple features in future this should change to at least one feature being set
-	if lcSpec == nil {
-		return fmt.Errorf("missing required LogCollection spec in applicationLayer")
+	// If ApplicationLayer spec exists then one of its features should be set.
+	if al.Spec.LogCollection == nil && al.Spec.WebApplicationFirewall == nil {
+		return fmt.Errorf("ApplicationLayer is not configured")
```
caseydavenport, line 314: This is a user facing error, it needs to be more precise about what the problem is and how it can be resolved.

e.g., "At least one of webApplicationFirewall or locCollector must be specified on ApplicationLayer resource"

caseydavenport (changes requested):
@bartolini first pass done! Will take another look on Monday. Please reach out to me if you have any questions about the comments or if you want to talk through an approach - I know sometimes comments can be a little unclear.

api/v1/applicationlayer_types.go
bartolini: done.

pkg/controller/applicationlayer/applicationlayer_controller.go
bartolini: done.

config/samples/operator_v1_applicationlayer.yaml
bartolini: fixed.

api/v1/applicationlayer_types.go
```diff
@@ -22,15 +22,20 @@ import (
...
 }
 
 type LogCollectionStatusType string
+type WebApplicationStatusType string
 
 const (
-	L7LogCollectionDisabled LogCollectionStatusType = "Disabled"
-	L7LogCollectionEnabled  LogCollectionStatusType = "Enabled"
+	WafDisabled             WebApplicationStatusType = "Disabled"
+	WafEnabled              WebApplicationStatusType = "Enabled"
```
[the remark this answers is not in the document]
bartolini: done.

pkg/controller/applicationlayer/applicationlayer_controller.go
```diff
@@ -303,6 +393,11 @@ func (r *ReconcileApplicationLayer) isLogsCollectionEnabled(l7Spec *operatorv1.L
 	return l7Spec != nil && l7Spec.CollectLogs != nil && *l7Spec.CollectLogs == operatorv1.L7LogCollectionEnabled
 }
 
+func (r *ReconcileApplicationLayer) isWafEnabled(applicationLayerSpec *operatorv1.ApplicationLayerSpec) bool {
```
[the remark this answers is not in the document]
bartolini: done.

pkg/controller/applicationlayer/applicationlayer_controller.go
```diff
@@ -277,12 +308,71 @@ func updateApplicationLayerWithDefaults(al *operatorv1.ApplicationLayer) {
...
+		TypeMeta: metav1.TypeMeta{Kind: "ConfigMap", APIVersion: "v1"},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      applicationlayer.ModSecurityRulesetConfigMapName,
+			Namespace: common.OperatorNamespace(),
+			Labels:    map[string]string{},
+		},
+		Data: make(map[string]string),
+	}
+
+	for filename, dataBase64 := range applicationlayer.ModsecurityCoreRuleSet {
```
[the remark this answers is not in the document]
bartolini: I clicked resolved by accident. For now it's alright but it would be nice to have an alternative solution to ConfigMap. Have you any suggestions here?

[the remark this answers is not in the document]
caseydavenport: The only alternative solution I can think of is a way of splitting this across multiple configmaps. Let's see if @tmjd or @Brian-McM has thought about this at all. 

@bartolini what's the total size of the configmap once it's all put together? You you forsee that size changing substantially as users customize this? 

I suppose worst-case, we ship with a single configmap and extend to support user-provided extension configmaps in some way, but we haven't thought through how that might work yet.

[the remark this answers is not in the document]
bartolini: @caseydavenport ah, apologies. I was writing comments here and clicking `Resolve` hoping they will stay here.

The size is almost 900kb so below the limit but close.

From what I know - users are more likely to disable/remove/edit some of these files and add a handful of their own - from what I know that would be a typical use case.

This is a very good point you are making here as I didn't even think about the CM size limit here.

pkg/render/applicationlayer/applicationlayer.go
```diff
@@ -214,13 +245,44 @@ func (c *component) containers() []corev1.Container {
...
+				"--rules", "/etc/waf",
+			},
+			Env: []corev1.EnvVar{
+				{Name: "LOG_LEVEL", Value: "Info"},
+			},
+			VolumeMounts: []corev1.VolumeMount{
+				{Name: FelixSync, MountPath: "/var/run/felix"},
+				{Name: DikastesSyncVolumeName, MountPath: "/var/run/dikastes"},
+				{Name: CalicoLogsVolumeName, MountPath: "/var/log/calico"},
+				{Name: ModSecurityRulesetVolumeName, MountPath: "/etc/waf", ReadOnly: true},
```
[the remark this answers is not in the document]
bartolini: Dead right, I will change that.

pkg/controller/applicationlayer/applicationlayer_controller.go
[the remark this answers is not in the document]
tmjd: I did not consider the size of them and potentially hitting a limit. So no prior thoughts from me.

pkg/controller/applicationlayer/applicationlayer_controller.go
caseydavenport: yep that makes total sense, I was just reading too quickly and missed the nuance.

pkg/render/applicationlayer/applicationlayer.go
```diff
@@ -68,14 +78,19 @@ type Config struct {
 	Installation *operatorv1.InstallationSpec
 	OsType       rmeta.OSType
 
-	// Optional config.
+	// Optional config for WAF.
+	WafEnabled           bool
```
caseydavenport, line 82: Waf -> WAF

pkg/render/applicationlayer/applicationlayer.go
```diff
@@ -116,9 +135,21 @@ func (c *component) SupportedOSType() rmeta.OSType {
 func (c *component) Objects() ([]client.Object, []client.Object) {
 	var objs []client.Object
 	// If l7spec is provided render the required objects.
-	c.config.envoyConfigMap = c.envoyL7ConfigMap()
 	objs = append(objs, c.serviceAccount())
+
+	// If Web Application Firewall is enabled, we need WAF ruleset ConfigMap present.
+	if c.config.WafEnabled {
+		// the first ConfigMap exists within the calico-operator namespace and stays as provided in config
```
caseydavenport, line 142: calico-operator -> tigera-operator

Also, per my comment above, I think we need to take the tigera-operator version of this configmap out of this render logic and put it in the controller the same way we do here: https://github.com/tigera/operator/blob/ed7dc7288f55e3f8bbc5804f316afdb5e8e774cd/pkg/controller/installation/core_controller.go#L1133-L1134

pkg/render/applicationlayer/applicationlayer.go
```diff
@@ -130,11 +161,6 @@ func (c *component) Objects() ([]client.Object, []client.Object) {
 		objs = append(objs, c.securityContextConstraints())
 	}
 
-	// Delete all the objects if logs are not enabled.
```
caseydavenport, line 133: We still might have only one of logs / WAF enabled - if this is the case, we should delete any resources that aren't necessary for the operation of the other.

e.g., if logging is enabled but WAF is disabled, we probably should delete the modsec-ruleset from the namespace. 

This isn't critical though and we can address this in a follow-on - just create a backlog item for it.

pkg/render/applicationlayer/applicationlayer.go
```diff
@@ -214,13 +244,46 @@ func (c *component) containers() []corev1.Container {
...
+				"server",
+				"--dial", "/var/run/felix/nodeagent/socket",
+				"--listen", "/var/run/dikastes/dikastes.sock",
+				"--rules", "/etc/modsecurity-ruleset",
+			},
+			Env: []corev1.EnvVar{
+				{Name: "LOG_LEVEL", Value: "Info"},
+			},
+			VolumeMounts: []corev1.VolumeMount{
+				{Name: FelixSync, MountPath: "/var/run/felix"},
```
caseydavenport, line 274: I think in the design we mentioned a need to validate that this is configured properly on FelixConfiguration - this is another one for the backlog / next release. Let's not block this PR on it, but we should revisit how we ensure the system is set up properly.

As it is, a user can turn on WAF without enabling policy sync in Felix and I believe they will get a broken cluster.

pkg/render/applicationlayer/applicationlayer.go
```diff
@@ -214,13 +244,46 @@ func (c *component) containers() []corev1.Container {
...
+				{Name: DikastesSyncVolumeName, MountPath: "/var/run/dikastes"},
+				{Name: CalicoLogsVolumeName, MountPath: "/var/log/calico"},
+				{Name: ModSecurityRulesetVolumeName, MountPath: "/etc/modsecurity-ruleset", ReadOnly: true},
+			},
+			SecurityContext: &corev1.SecurityContext{
```
caseydavenport, line 279: Have we considered at all if there is a way around this requirement? We really shouldn't be running as root unless absolutely necessary. Talk with @mgleung about options here. This is another one to add to the backlog, because we likely won't be able to address in this release but we should fix it next release.

pkg/render/applicationlayer/envoy-config.yaml.template
```diff
@@ -1,3 +1,40 @@
+{{- define "access_log" -}}
```
caseydavenport, line 1: Another one for the backlog - we want do be doing conditional logic in golang rather than templates. This one seems borderline, but if this gets much more complex we'll definitely need to put this in golang so we can more easily unit test this.

caseydavenport (changes requested):
Mostly looking good. I think most of the comments can be addressed later, but I spotted on more issue with the way we're doing the default config map (sorry, this is my bad for not spotting earlier).

I also noticed there are no new tests with this PR - we absolutely need unit tests which verify the behavior of the render package and assert the correct volumes / mounts  / containers / env vars etc are emitted based on input.

We also need unit testing of the envoy config now that there is conditional logic especially.

If you want we can do that as a separate PR but we should not be recommending release without it.

pkg/render/applicationlayer/applicationlayer.go
```diff
@@ -39,14 +43,20 @@ import (
...
+	ApplicationLayerDaemonsetName    = "l7-log-collector"
+	L7CollectorContainerName         = "l7-collector"
+	ProxyContainerName               = "envoy-proxy"
+	EnvoyLogsVolumeName              = "envoy-logs"
+	EnvoyConfigMapName               = "envoy-config"
+	EnvoyConfigMapKey                = "envoy-config.yaml"
+	FelixSync                        = "felix-sync"
+	DikastesSyncVolumeName           = "dikastes-sync"
+	DikastesContainerName            = "dikastes"
+	ModSecurityRulesetVolumeName     = "owasp-ruleset"
```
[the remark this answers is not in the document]
bartolini: yup, you're right.

pkg/render/applicationlayer/applicationlayer.go
bartolini: Will do, thanks!

pkg/render/applicationlayer/envoy-config.yaml.template
bartolini: Noted - will create a ticket for further refactoring.

pkg/render/applicationlayer/applicationlayer.go
bartolini: Backlog item it is. Thanks!

pkg/controller/applicationlayer/applicationlayer_controller.go
```diff
@@ -216,14 +231,30 @@ func (r *ReconcileApplicationLayer) Reconcile(ctx context.Context, request recon
 		return reconcile.Result{}, err
 	}
 
+	var modSecurityRuleSet *corev1.ConfigMap
+	if r.isWAFEnabled(&applicationLayer.Spec) {
+		if modSecurityRuleSet, err = r.getModSecurityRuleSet(ctx); err != nil {
```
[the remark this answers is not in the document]
bartolini: I think this is sorted now. Will ping you in a few minutes.

pkg/controller/applicationlayer/applicationlayer_controller.go
```diff
@@ -277,12 +319,74 @@ func updateApplicationLayerWithDefaults(al *operatorv1.ApplicationLayer) {
...
+		},
+		ruleset,
+	); err == nil {
+		return ruleset, false, nil
+	} else if !apierrors.IsNotFound(err) {
+		return nil, false, err
+	}
+
+	if ruleset, err := getDefaultCoreRuleset(ctx); err == nil {
+		return ruleset, true, nil
```
caseydavenport, line 352: Another nit, but a more idiomatic way to do this in golang would be to put the error condition first.


```go
ruleset, err := getDefaultCoreRuleset(ctx)
if err != nil {
  return nil, false, err
}
return ruleset, true, nil
```

Having read lots of go code, seeing the branches the way they are makes me immediately think there's something atypical going on here when actually it's just a normal error branch.

pkg/controller/applicationlayer/applicationlayer_controller.go
```diff
@@ -277,12 +319,74 @@ func updateApplicationLayerWithDefaults(al *operatorv1.ApplicationLayer) {
...
+		} else {
+			return nil, err
+		}
+	}
+
+	return ruleset, nil
+}
+
+func validateModSecurityRuleSet(cm *corev1.ConfigMap) error {
+
```
caseydavenport, line 380: nit: extra newline

pkg/render/applicationlayer/applicationlayer.go
caseydavenport: @bartolini this one is still not resolved - Waf has the wrong casing

pkg/render/applicationlayer/applicationlayer.go
```diff
@@ -305,10 +411,36 @@ func (c *component) collectorVolMounts() []corev1.VolumeMount {
...
+		Data:       c.config.ModSecurityConfigMap.Data,
+		BinaryData: c.config.ModSecurityConfigMap.BinaryData,
+	}
+}
+
+//go:embed envoy-config.yaml.template
+var envoyConfigTemplate string
 
 func (c *component) envoyL7ConfigMap() *corev1.ConfigMap {
+
```
caseydavenport, line 431: nit extra newline

pkg/controller/applicationlayer/applicationlayer_controller.go
bartolini: sorted.

PR #1744, opened 2022-01-26
Fix that sync state could be blocked when the TigeraStatus CR doesn't exist

caseydavenport:
## Description

The problem here was that `removeTigeraStatus` used to gate execution of `updateStatus` if it returned `true`.

We had added code which short-circuited and returned true if `!m.crExists`, which was always hit at start of day. The `removeTigeraStatus` function itself was pretty convoluted and shouldn't have been the function deciding whether we should update the status or not anyway.

I believe the intention of the true / false return was that `updateStatus` should NOT be called if the status manager is explicitly, administratively disabled (`m.enabled != nil && !*m.enabled`). However, the short-circuit branch bypassed the enabled check and always indicated that the SM was disabled.

This PR simplifies `removeTigeraStatus` (all it does now is actually remove the tigera status) and moves the check for whether or not we should execute `updateStatus` outside of that function so that the logic is more clear. 

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/status/status.go
+++ b/pkg/controller/status/status.go
@@ -145,9 +145,13 @@ func (m *statusManager) updateStatus() {
 		return
 	}
 
-	if m.removeTigeraStatus() {
+	if m.enabled != nil && !*m.enabled {
+		// This status manager is explicitly disabled, because the controller has called OnCRNotFound.
+		// Remove any TigeraStatus object that had previously been created, and skip updating the status.
+		m.removeTigeraStatus()
 		return
 	}
+	// This status manager is enabled. Perform a sync.
 
 	// Unless we've been given an explicit degraded reason we are not ready to start reporting statuses until
 	// ReadyToMonitor has been called by the owner of the status manager. This means there's no point in syncing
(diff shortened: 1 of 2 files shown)


pkg/controller/status/status.go
```diff
@@ -581,27 +585,24 @@ func (m *statusManager) isInitialized() bool {
 	return m.enabled != nil
 }
 
-// removeTigeraStatus returns true and removes the status displayed in TigeraStatus if corresponding CR not found
-func (m *statusManager) removeTigeraStatus() bool {
+// removeTigeraStatus status displayed in TigeraStatus if corresponding controller's CR is not found
```
tmjd, line 588: This comment doesn't make sense to me.

tmjd (commented):
What was the case that was happening that this fixes things? Was this a case where no SetDegraded happened so the CR was never created and that blocked updateStatus from running fully?

pkg/controller/status/status.go
caseydavenport: well yeah, that's because the sentence doesn't make any sense, duh.

caseydavenport:
@tmjd updated the description with a better explanation of what the problem was.

tmjd (approved):
LGTM

lmm (approved):

PR #1784, opened 2022-02-16
Web Application Firewall updates

bartolini:
## Description
This PR covers the following changes to the Web Application Firewall configuration:
- default values for ApplicationLayers CRD are properly set.
- ModSecurity library core rule set is v3.3.2 (stable) instead of v.3.4.0 (dev).
- ModSecurity library is configured in self-contained mode instead of anomaly scoring.
- HTTP request body is made available to Dikastes, this is achieved by Envoy's config change.

## For PR reviewers
The PR might look big but vast majority of the changes here are to ModSecurity library core rule set.
Changes to the code of the operator are minimal.


pkg/controller/applicationlayer/applicationlayer_controller.go
```diff
@@ -304,16 +304,25 @@ func (r *ReconcileApplicationLayer) Reconcile(ctx context.Context, request recon
 
 // updateApplicationLayerWithDefaults populates the applicationlayer with defaults.
 func updateApplicationLayerWithDefaults(al *operatorv1.ApplicationLayer) {
-	defaultLogIntervalSeconds := int64(5)
-	defaultLogRequestsPerInterval := int64(-1)
+	var defaultLogIntervalSeconds int64 = 5
+	var defaultLogRequestsPerInterval int64 = -1
+	var defaultLogCollectionDisabled operatorv1.LogCollectionStatusType = "Disabled"
+	var defaultWebApplicationFirewallDisabled operatorv1.WAFStatusType = "Disabled"
```
caseydavenport, line 310: We should use the typed constants for these rather than the string "Disabled"

pkg/controller/applicationlayer/applicationlayer_controller.go
```diff
@@ -304,16 +304,25 @@ func (r *ReconcileApplicationLayer) Reconcile(ctx context.Context, request recon
...
-		if al.Spec.LogCollection.LogIntervalSeconds == nil {
-			al.Spec.LogCollection.LogIntervalSeconds = &defaultLogIntervalSeconds
-		}
+	if al.Spec.LogCollection == nil {
+		al.Spec.LogCollection = new(operatorv1.LogCollectionSpec)
+	}
+	if al.Spec.LogCollection.CollectLogs == nil {
+		al.Spec.LogCollection.CollectLogs = &defaultLogCollectionDisabled
+	}
+	if al.Spec.LogCollection.LogRequestsPerInterval == nil {
```
caseydavenport, line 318: It's a bit odd to default a feature to "off" and then default other fields related to it to be "on". I would say this is probably an invalid configuration - you're basically saying "Disable log collection, and also produce this many logs per interval".

For context, if we didn't have an `enabled` field already, I would think the correct idiomatic way to do this would be:

```yaml
# Enable log collection
spec:
  logCollection:
    logIntervalSeconds: 5
```

```yaml
# disable log collection
spec: 
  logCollection:
    logIntervalSeconds: 0
``` 

But we do already have an `enabled` field, and so we just need to make sure that the "enabled" field is consistent with the others.

So, this would be invalid:

```yaml
spec:
  logCollection:
    enabled: true
    logIntervalSeconds: 0
```

And so would this:

```yaml
spec:
  logCollection:
    enabled: false
    logInterval Seconds: 5
```

because it's two pieces of conflicting information given to us by the user.

pkg/controller/applicationlayer/applicationlayer_controller.go
```diff
@@ -304,16 +304,25 @@ func (r *ReconcileApplicationLayer) Reconcile(ctx context.Context, request recon
 
 // updateApplicationLayerWithDefaults populates the applicationlayer with defaults.
 func updateApplicationLayerWithDefaults(al *operatorv1.ApplicationLayer) {
-	defaultLogIntervalSeconds := int64(5)
-	defaultLogRequestsPerInterval := int64(-1)
+	var defaultLogIntervalSeconds int64 = 5
+	var defaultLogRequestsPerInterval int64 = -1
```
caseydavenport, line 308: I know it was this way before your PR, but why are we defaulting this to -1? What does a value of -1 mean for a rate field? Seems like it would have no semantic meaning to me - we should probably disallow negative numbers for this field and other time based fields, potentially even using `uint` here instead.

pkg/render/applicationlayer/applicationlayer.go
```diff
@@ -159,6 +159,11 @@ func (c *component) Objects() ([]client.Object, []client.Object) {
 		objs = append(objs, c.securityContextConstraints())
 	}
 
+	// Delete all objects if both log collection and WAF are disabled
+	if !c.config.LogsEnabled && !c.config.WAFEnabled {
```
caseydavenport, line 163: Why do we need this? Deletion should be handled by owner references, and I thought that we were validating against this being a possible configuration (i.e., it should not be possible to create an ApplicationLayer resource with both of these disabled, because that has no meaning whatsoever)

caseydavenport (changes requested):
A very quick skim from me with some questions

pkg/render/applicationlayer/applicationlayer.go
bartolini: We do - in case when e.g. either L7 or WAF was enabled and later on disabled the Envoy daemonset would stay present. We don't want that. This comes from manual testing.

pkg/controller/applicationlayer/applicationlayer_controller.go
bartolini: We can't pass a pointer to a typed constant hence this is in place.

pkg/controller/applicationlayer/applicationlayer_controller.go
bartolini: I wouldn't agree with this statement:
`"Disable log collection, and also produce this many logs per interval"`
It is a legitimate use case to have some configuration working and to disable it without loosing the previous values that were set.

This code is about L7 log collection - I couldn't find a doc that would be descriptive of what an expected behavior in this case should be. I'm sure the default values - that aren't new in this context here - \ have a meaning for that feature also.

pkg/render/applicationlayer/applicationlayer.go
caseydavenport: > was enabled and later on disabled 

But this is the problem - it should not be possible to create an Application layer resource with both `LogsEnabled: false` and `WAFEnabled: false` - that's a nonsensical API request, it's basically saying "Here's an explicit request to do nothing".

We should be enfocing users deleted the ApplicationLayer resource if they want to get rid of the envoy daemonset  - that's how every other one of our APIs works.

pkg/controller/applicationlayer/applicationlayer_controller.go
caseydavenport: > It is a legitimate use case to have some configuration working and to disable it without loosing the previous values that were set.

Not via a Kubernetes API it isn't. 

The way to meet this use-case in a Kubernetes world is like so:

```
kubectl get applicationlayer default > application-layer.backup
```

Then to revert any changes:
```
kubectl replace -f application-layer.backup
```

caseydavenport: ^ my point is that this is the model we have followed for all of our other APIs. There's no "create an APIServer but don't actually run the apiserver" option, etc., etc. We should be consistent across all of our APIs, and unless I'm mistaken there's nothing special about WAF that makes this an exception. 

It's an anti-pattern to have the "enabled" fields at all in my opinion, but following the above at least makes it consistent with how the structure would have worked if we hadn't introduced the "enabled" fields.

pkg/controller/applicationlayer/applicationlayer_controller.go
caseydavenport: sorry, what I meant was not to use a pointer to a constant, rather to do this:

```
disabled := operatorv1.Disabled
```

and then 

```
al.Spec.LogCollection.CollectLogs = &disabled
```

pkg/controller/applicationlayer/applicationlayer_controller.go
bartolini: I have seen that Suresh replied to this question in the operator-dev slack channel. Just to keep the record of this information here's a snippet from CRD:
```'Maximum number of unique L7 logs that are sent LogIntervalSeconds.
                      Adjust this to limit the number of L7 logs sent per LogIntervalSeconds
                      to felix for further processing, use negative number to ignore
                      limits. Default: -1'

File: pkg/controller/applicationlayer/applicationlayer_controller.go (hunk 1 above)
Comment (bartolini) [reply pkg/controller/applicationlayer/applicationlayer_controller.go]: done.

File: pkg/controller/applicationlayer/applicationlayer_controller.go (hunk 2 above)
Comment (bartolini) [reply pkg/controller/applicationlayer/applicationlayer_controller.go]: Done, as per our conversation.

File: pkg/render/applicationlayer/applicationlayer.go (hunk 4 above)
Comment (bartolini) [reply pkg/render/applicationlayer/applicationlayer.go]: Done as per our conversation. This won't be needed now.

File: pkg/controller/applicationlayer/applicationlayer_controller.go [hunk 5]
```diff
@@ -304,24 +304,36 @@ func (r *ReconcileApplicationLayer) Reconcile(ctx context.Context, request recon
...
+	var defaultLogCollectionDisabled operatorv1.LogCollectionStatusType = operatorv1.L7LogCollectionDisabled
+	var defaultWebApplicationFirewallDisabled operatorv1.WAFStatusType = operatorv1.WAFDisabled
 
-	if al.Spec.LogCollection != nil {
+	if al.Spec.LogCollection == nil {
+		al.Spec.LogCollection = new(operatorv1.LogCollectionSpec)
+		al.Spec.LogCollection.CollectLogs = &defaultLogCollectionDisabled
+	}
+
+	if *al.Spec.LogCollection.CollectLogs == operatorv1.L7LogCollectionEnabled {
```
caseydavenport, line 317: We're de-referencing a pointer that might be nil here. Imaging the case where a user provides this:

```go
LocCollection{
  LogRequestsPerInterval: 5
}
```

The struct is non-nil, so we won't enter the branch above, but the `CollectLogs` field is nil, so we'll panic here.

caseydavenport: I think the fix is to remove seeing the default in the branch above, and instead do something like this:


```go

// Determine if any fields were specified. If any fields were provided, that signals intent by the user 
// to enable log collection.
shouldEnableLogCollection := collectLogs != nil || logInterval != nil || logIntervalSeconds != nil
if al.Spec.LogCollection.CollectLogs == nil {
  if shouldEnableLogCollection {
    al.Spec.LogCollection.CollectLogs = Enabled
  } else {
      al.Spec.LogCollection.CollectLogs = Disabled
  }
}

if al.Spec.LogCollection.CollectLogs == Enabled {
  // default other log collection fields
}
```

pkg/controller/applicationlayer/applicationlayer_controller.go
```diff
@@ -304,24 +304,36 @@ func (r *ReconcileApplicationLayer) Reconcile(ctx context.Context, request recon
...
+		al.Spec.WebApplicationFirewall = &defaultWebApplicationFirewallDisabled
+	}
 }
 
 // validateApplicationLayer validates ApplicationLayer
 func validateApplicationLayer(al *operatorv1.ApplicationLayer) error {
 
 	// If ApplicationLayer spec exists then one of its features should be set.
-	if al.Spec.LogCollection == nil && al.Spec.WebApplicationFirewall == nil {
+	if *al.Spec.LogCollection.CollectLogs != operatorv1.L7LogCollectionEnabled &&
```
caseydavenport, line 335: We should still take care to check pointer dereferences here, as a safety precaution - we don't want to panic ever.

pkg/controller/applicationlayer/applicationlayer_controller.go
```diff
@@ -304,24 +304,36 @@ func (r *ReconcileApplicationLayer) Reconcile(ctx context.Context, request recon
...
+		*al.Spec.WebApplicationFirewall != operatorv1.WAFEnabled {
 		return fmt.Errorf("at least one of webApplicationFirewall or logCollector must be specified on ApplicationLayer resource")
```
caseydavenport, line 337: logCollector isn't the name of a field - this should be `logCollection.collectLogs` instead (missed this in the original review)

caseydavenport (changes requested):

caseydavenport:
>  .../crd.projectcalico.org_bgpconfigurations.yaml   | 31 ++++++++++++++++++++++

You also need to rebase or run `make gen-versions` to fix CI

pkg/controller/applicationlayer/applicationlayer_controller.go
bartolini: Changed now.

pkg/controller/applicationlayer/applicationlayer_controller.go
bartolini: Sure, I will add it here also.

pkg/controller/applicationlayer/applicationlayer_controller.go
bartolini: Yup, good point. I think this is resolved now, I didn't consider the structure as provided.

The default value for `CollectLogs` is `Disabled` though so I think in this example case L7 logs should remain disabled. I can change that if you want this feature indirectly enabled by other parameters - let me know.

caseydavenport: If the user provided intent to enable log collection (by setting one of the other log collection fields) but didn't explicitly set the collectLogs field and left it nil, we have two options:

- Default the CollectLogs field unconditionally to "Disabled", resulting in an invalid configuration and returning an error when we reach the validation step.
- Default the CollectLogs field conditionally in a way that is consistent with the other configuration that is provided (Disabled if no other fields set, Enabled if other fields are set). 

The latter is what I would recommend. If we want to go with the former, we should probably be marking the CollectLogs field as required in the API to make sure any request is rejected if it includes the LogCollection substruct without the CollectLogs field being set.

I can see a case for either approach, but the latter feels more "useful" to me and less frustrating to the user.

caseydavenport: That's a new behavioral change though so if you don't want to deal with it in this PR that's OK.

pkg/controller/applicationlayer/applicationlayer_controller.go
```diff
@@ -304,25 +304,47 @@ func (r *ReconcileApplicationLayer) Reconcile(ctx context.Context, request recon
...
+	// Just a precaution to make sure it is safe to dereference pointers below.
+	// This is already taken care of in updateApplicationLayerWithDefaults.
+	properlyConfigured := al.Spec.LogCollection != nil &&
+		al.Spec.LogCollection.CollectLogs != nil &&
+		al.Spec.WebApplicationFirewall != nil
+
 	// If ApplicationLayer spec exists then one of its features should be set.
-	if al.Spec.LogCollection == nil && al.Spec.WebApplicationFirewall == nil {
-		return fmt.Errorf("at least one of webApplicationFirewall or logCollector must be specified on ApplicationLayer resource")
+	if properlyConfigured &&
```
caseydavenport, line 344: I think this should be `if !properlyConfigured || . . .` - we want to return an error if the given application layer object is not properly configured.

caseydavenport (commented):
Just one remaining comment

pkg/controller/applicationlayer/applicationlayer_controller.go
bartolini: Yes, that is right.

caseydavenport (approved):

PR #1787, opened 2022-02-17
Add more logging to the check resource registration loop

robbrockbank:
## Description

Add some additional logging in the resource registration check in the create watcher loop.

Not sure if the lack of logs for the resource registration was deliberate or not - but given it's a log once per 30s it seems reasonable to include it - especially as we should only be watching the required resources if the corresponding CRs are present.  

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/utils/utils.go
+++ b/pkg/controller/utils/utils.go
@@ -429,6 +429,7 @@ func WaitToAddResourceWatch(controller controller.Controller, client kubernetes.
 	maxDuration := 30 * time.Second
 	duration := 1 * time.Second
 	ticker := time.NewTicker(duration)
+	log = ContextLoggerForResource(log, obj)
 	defer ticker.Stop()
 	for range ticker.C {
 		duration = duration * 2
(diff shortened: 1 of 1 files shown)


pkg/controller/utils/utils.go
```diff
@@ -430,6 +430,7 @@ func WaitToAddResourceWatch(controller controller.Controller, client kubernetes.
 	duration := 1 * time.Second
 	ticker := time.NewTicker(duration)
 	defer ticker.Stop()
+	log.Info("handlng watch of %s resource", obj.GetObjectKind().GroupVersionKind().Kind)
```
caseydavenport, line 433: Maybe "waiting for %s API to become available for watching" ?

pkg/controller/utils/utils.go
```diff
@@ -438,32 +439,33 @@ func WaitToAddResourceWatch(controller controller.Controller, client kubernetes.
...
-			if isResourceReady(client, obj.GetObjectKind().GroupVersionKind().Kind) {
-				err := controller.Watch(&source.Kind{Type: obj}, &handler.EnqueueRequestForObject{})
-				if err != nil {
-					log.Info("failed to watch %s resource: %v. Will retry to add watch", obj.GetObjectKind().GroupVersionKind().Kind, err)
-				} else {
-					flag.MarkAsReady()
-					return
-				}
+			if ok, err := isResourceReady(client, obj.GetObjectKind().GroupVersionKind().Kind); err != nil {
+				log.Info("failed to list registered groups and resources - will retry: %v", err)
```
caseydavenport, line 443: feel like the structure of this is a little bit off... technically this code doesn't know that the returned error means a failure to list registered groups, it could be anything (if for example the implementation of `isResourceReady` changes in the future).

It's pretty nitty because the functions are right next to each other and are unlikely to change, but feels like maybe this log should say "failed to check if resource is ready"

pkg/controller/utils/utils.go
caseydavenport: Would it be worth adding the `kind` to a context logger for this loop so it's present on subsequent logs as well?

pkg/controller/utils/utils.go
```diff
@@ -438,32 +439,33 @@ func WaitToAddResourceWatch(controller controller.Controller, client kubernetes.
...
+			} else if !ok {
+				log.Info("waiting for %s resource to be registered - will retry", obj.GetObjectKind().GroupVersionKind().Kind)
```
caseydavenport, line 445: if we add the Kind to a context logger we can remove the string formatting from this one and the one below.

pkg/controller/utils/utils.go
caseydavenport: This should probably be an `Error` level log as well

pkg/controller/utils/utils.go
```diff
@@ -438,32 +439,33 @@ func WaitToAddResourceWatch(controller controller.Controller, client kubernetes.
...
+			} else if err := controller.Watch(&source.Kind{Type: obj}, &handler.EnqueueRequestForObject{}); err != nil {
+				log.Info("failed to watch %s resource - will retry: %v", obj.GetObjectKind().GroupVersionKind().Kind, err)
+			} else {
+				flag.MarkAsReady()
```
caseydavenport, line 449: Do we want a success log? Maybe at a lower verbosity?

pkg/controller/utils/utils.go
caseydavenport: Eh, maybe it's fine at Info - we'll retry and it's not really breaking. Probably reasonable transient conditions exist where we'll hit this branch and don't want to alarm anyone.

caseydavenport:
@robbrockbank piiinggg

pkg/controller/utils/utils.go
robbrockbank: Sure - I think the same verbosity as the log on 434.... But perhaps they could both be downgraded to debug.

caseydavenport:
```
pkg/controller/utils/utils.go:433:15: log.WithField undefined (type logr.Logger has no field or method WithField)01:44
# github.com/tigera/operator/pkg/controller/utils [github.com/tigera/operator/pkg/controller/utils.test]01:44
pkg/controller/utils/utils.go:433:15: log.WithField undefined (type logr.Logger has no field or method WithField)
```

@robbrockbank whoopsie

caseydavenport (approved):

PR #1800, opened 2022-03-01
Remove env that is no longer used

rene-dekker:


pkg/render/manager.go
```diff
@@ -635,8 +635,6 @@ func (c *managerComponent) managerEsProxyContainer() corev1.Container {
 
 	env := []corev1.EnvVar{
 		{Name: "ELASTIC_LICENSE_TYPE", Value: string(c.cfg.ESLicenseType)},
-		// The ELASTIC_VERSION flag is used to set the kbn-version for a browser redirect.
-		{Name: "ELASTIC_VERSION", Value: components.ComponentEckKibana.Version},
```
tmjd, line 639: Did you check if we still need that ComponentEckKibana anymore? My guess would be probably not since the Kibana image is actually different from this version I believe.

Brian-McM: > My guess would be probably not since the Kibana image is actually different from this version I believe.

This isn't true, this should always have the correct Kibana version. This is set in the Kibana spec when the operator installs Kibana. The problem with cloud is that the operator doesn't install / upgrade Kibana so this is only not true for cloud.

Brian-McM (approved):

PR #1801, opened 2022-03-01
Set Kibana public domain based on Authentication domain

tmjd:
## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/logstorage.go
+++ b/pkg/render/logstorage.go
@@ -207,6 +207,7 @@ type ElasticsearchConfiguration struct {
 	KbService                   *corev1.Service
 	ClusterDomain               string
 	DexCfg                      DexRelyingPartyConfig
+	BaseURL                     string // BaseUrl is where the manager is reachable, for setting Kibana publicBaseUrl
 	ElasticLicenseType          ElasticsearchLicenseType
 }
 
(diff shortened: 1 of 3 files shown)


pkg/render/logstorage.go
```diff
@@ -207,6 +207,7 @@ type ElasticsearchConfiguration struct {
 	KbService                   *corev1.Service
 	ClusterDomain               string
 	DexCfg                      DexRelyingPartyConfig
+	ManagerDomain               string // Domain where the manager is reachable, for setting Kibana publicBaseUrl
```
rene-dekker, line 210: If this variable would be called KibanaPublicURL, you would need no comment to clarify it and `fmt.Sprintf("%s/%s", es.cfg.ManagerDomain, KibanaBasePath)`  could then be done in the controller, saving you an if-statement on line 1263. WDYT?

rene-dekker (approved):
lgtm

pkg/render/logstorage.go
tmjd: I like the idea but I think it is more appropriate to construct it in the render, that is the correct scope for doing that.

pkg/render/logstorage.go
```diff
@@ -1259,8 +1260,15 @@ func (es elasticsearchComponent) kibanaCR() *kbv1.Kibana {
 		"defaultRoute":    fmt.Sprintf(KibanaDefaultRoute, TimeFilter, url.PathEscape(FlowsDashboardName)),
 	}
 
-	if es.cfg.DexCfg != nil && es.cfg.DexCfg.BaseURL() != "" {
-		server["publicBaseUrl"] = fmt.Sprintf("%s/tigera-kibana", es.cfg.DexCfg.BaseURL())
+	baseUrl := es.cfg.ManagerDomain
+	// If we can't parse the ManagerDomain with url.Parse then assume it won't be valid for the Kibana publicBaseUrl
+	if u, err := url.Parse(baseUrl); err == nil {
```
Brian-McM, line 1265: It might be a good idea to log the error, might be hard to debug if we don't.

Brian-McM (approved):

PR #1803, opened 2022-03-02
bundle: update operator bundle tooling and docs

lmm:
## Description

This PR fixes the operator bundle tooling and docs that are used to publish operator releases on RedHat's operator catalog. These fixes are needed for the latest updates to the certification pipeline referenced in the updated releasing doc.

This PR also removes the long broken release image push to RedHat. That method no longer works, and the [API](https://catalog.redhat.com/api/containers/v1/ui/#/Certification%20projects%20requests/pyxis.rest.legacy.cert_project_scan_request.create_certification_scan_request) to trigger image scans is going to be removed soon. Image scans will need to be done manually for now.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -60,10 +60,13 @@ Before beginning, ensure that the docs at docs.projectcalico.org for the Calico 
 
 1. After the semaphore job in the releasing steps is complete, and images have been tagged and pushed, checkout the tag you released and create a new branch.
 
-1. Login to our operator project on connect.redhat.com and publish the operator image on the RH Catalog. This step needs to happen before we generate and submit the operator metadata bundle.
+1. Login to our operator project on connect.redhat.com and publish the operator image on the RH Catalog. This step needs to happen before we generate and submit the operator bundle.
 
-1. Create the operator bundle, using the tag version for VERSION and the version that the release replaces in PREV_VERSION. The versions are semver strings.
-   CHANNELS and DEFAULT_CHANNEL should be set to the release stream.
+1. Create the operator bundle using `make bundle` with the required variables `VERSION`, `PREV_VERSION`, `CHANNELS`, and `DEFAULT_CHANNEL`:
+   - **VERSION**: this release version. E.g. `1.13.1` (Note that these version strings are semver strings without the `v`.)
+   - **PREV_VERSION**: the latest published bundle version in this release stream. Navigate to the [certified operators production catalog](https://github.com/redhat-openshift-ecosystem/certified-operators/tree/main/operators/tigera-operator) and look for the most recent version from this release branch. E.g., if this release is `v1.24.11` but the most recently published v1.24.x bundle is `v1.24.9` then you would use `VERSION=v1.24.11` and `PREV_VERSION=v1.24.9`. If this release is the first in this release branch, then there is no prior version so set `PREV_VERSION=0.0.0`.
+   - **CHANNELS** and **DEFAULT_CHANNEL**: should be set to the release branch of this operator release. E.g., if the operator release tag is `v1.23.5`, then CHANNELS and DEFAULT_CHANNEL should be `release-v1.23`.
+
    For example:
 
    ```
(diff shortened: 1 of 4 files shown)


RELEASING.md
```diff
@@ -62,7 +62,7 @@ Before beginning, ensure that the docs at docs.projectcalico.org for the Calico
 
 1. Login to our operator project on connect.redhat.com and publish the operator image on the RH Catalog. This step needs to happen before we generate and submit the operator metadata bundle.
 
-1. Create the operator metadata bundle, using the tag version for VERSION and the version that the release replaces in PREV_VERSION. The versions are semver strings.
+1. Create the operator bundle, using the tag version for VERSION and the version that the release replaces in PREV_VERSION. The versions are semver strings.
```
tmjd, line 65: Could you include more notes on the PREV_VERSION?
Like if we already had bundles `1.13.0, 1.13.1, 1.14.0`: 
* If we were doing `1.13.2`, then PREV_VERSION would be `1.13.1`
* If we were doing `1.14.1`, then PREV_VERSION would be `1.14.0`
* If we were doing `1.15.0`, then PREV_VERSION would be `1.15.0`
* If we were doing `1.13.3` but `1.13.2` bundle had not been published, then would we use `1.13.1`?
I know these might all seem obvious.

Also why does the CHANNELS includes `release-`, is that because that is the branch we build the release from? That seems like an odd thing to include but maybe it is up to the packager what the CHANNELS are?

lmm: Thanks @tmjd , I've pushed up some edits to improve the instructions based on your questions: 676d64ee10420eacb46d9b4da4679530c0453f42

> Also why does the CHANNELS includes release-, is that because that is the branch we build the release from?

It's just the convention we've been using. The channel name was arbitrary, and is used to distinguish between the different channels we have for each operator release branch. (Maybe simply `1.13` without the `release-` prefix would have been better.)

When an OCP user creates a subscription to the tigera-operator they need to specify the channel. New versions in that channel that would replace their existing version are eligible for upgrades.

tmjd: How does one upgrade from (release-)1.13 to (release-)1.14? Seems like we should maybe we should just use `stable`.

lmm: They would update their tigera-operator subscription's channel to `release-v1.14`.

We're using multiple channels because really we have multiple "stable" streams of releases and the releases can come in any order.

RELEASING.md
```diff
@@ -60,10 +60,13 @@ Before beginning, ensure that the docs at docs.projectcalico.org for the Calico
...
+1. Login to our operator project on connect.redhat.com and publish the operator image on the RH Catalog. This step needs to happen before we generate and submit the operator bundle.
+
+1. Create the operator bundle using `make bundle` with the required variables `VERSION`, `PREV_VERSION`, `CHANNELS`, and `DEFAULT_CHANNEL`:
+   - **VERSION**: this release version. E.g. `1.13.1` (Note that these version strings are semver strings without the `v`.)
+   - **PREV_VERSION**: the latest published bundle version in this release stream. Navigate to the [certified operators production catalog](https://github.com/redhat-openshift-ecosystem/certified-operators/tree/main/operators/tigera-operator) and look for the most recent version from this release branch. E.g., if this release is `v1.24.11` but the most recently published v1.24.x bundle is `v1.24.9` then you would use `VERSION=v1.24.11` and `PREV_VERSION=v1.24.9`. If this release is the first in this release branch, then there is no prior version so set `PREV_VERSION=0.0.0`.
```
tmjd, line 67: Should `PREV_VERSION=0.0.0` be `PREV_VERSION=v0.0.0`? Or maybe it doesn't matter.

lmm: Thanks for catching that @tmjd , I fixed up the instructions

tmjd (approved):
LGTM, looks good to me. Please squash when merging.

PR #1804, opened 2022-03-02
Setting MaxUnavailable to 1 when calculation against # of available calico node pods resolves to 0

freecaykes:
## Description

Due to the potential rounding down of numNodesMaxUnavailable = (csMaxUnavailable %) * ksD+csD, where maxUnavailable is a percentage value, ,it may resolve to zero. Then we should default back maxUnavailable to 1 on the theory that surge might not work due to quota.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/installation/windows/calico_windows_upgrader.go
+++ b/pkg/controller/installation/windows/calico_windows_upgrader.go
@@ -40,10 +40,12 @@ import (
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 )
 
+const (
+	defaultMaxUnavailable int32 = 1
+)
+
 var (
 	windowsLog = logf.Log.WithName("windows_upgrader")
-
-	defaultMaxUnavailable = 1
 )
 
 type CalicoWindowsUpgrader interface {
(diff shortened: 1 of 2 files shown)


pkg/controller/migration/namespace_migration.go
```diff
@@ -571,23 +573,33 @@ func (m *CoreNamespaceMigration) getNodesToMigrate() []*v1.Node {
...
 		}
+
+		// num node desired schedule, num node ready, node max unavailable in calico-system
 		csD, csR, csMaxUnavailable, err := m.getNumPodsDesiredAndReady(ctx, common.CalicoNamespace, nodeDaemonSetName)
 		if err != nil {
 			return false, err
 		}
 
-		var maxUnavailable int32 = 1
+		var maxUnavailable int32 = int32(defaultMaxUnavailable)
```
freecaykes, line 590: can we set this to using int ? or should we set the windows one to using int32 ?

caseydavenport: Any API fields should not use `int` and should instead use either `int32` or `int64` per the API conventions, so eventually this needs to be converted to an in32 (since that's the type of the maxunavailable field in the API)

> All public integer fields MUST use the Go (u)int32 or Go (u)int64 types, not (u)int (which is ambiguous depending on target platform). Internal types may use (u)int.

https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md

freecaykes: changing the windows on to int32 https://github.com/tigera/operator/pull/1804/commits/6882da130e88f4cba96d1f1738e44ac45cb41414

tmjd (changes requested):
The K8s code that we're adjusting our calculation to match, does it log if maxUnavailable evaluates to zero (and maxSurge is zero)?
I can just imagine getting a report at some point about this being an error when it is just because they have too few nodes and hit the maxUnavailable 0 case. I think we should separate the cases to log error if err is returned but otherwise it is just an info level message, then I think the message can be more descriptive of the actual issue.

pkg/controller/installation/windows/calico_windows_upgrader.go
```diff
@@ -40,10 +40,12 @@ import (
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 )
 
+const (
+	defaultMaxUnavailable = 1
```
tmjd, line 44: can't this be defaulted to be an int32 with `defaultMaxUnavailable int32 = 1`, then the cast later isn't needed

freecaykes: https://github.com/tigera/operator/pull/1804/commits/f940907e37306e04655304bd9d771cc0b96bc1fb

tmjd (approved):
LGTM

marvin-tigera:

marvin-tigera:

PR #1822, opened 2022-03-07
Use certificatemanager for dex

rene-dekker:
Dex config currently wraps and renders secrets. With the recent introduction of the certificate manager, we can now refactor this code. Other changes:
- The dex tls secret was mounted for key validators, which is unnecessary. They do offline bearer token validation.
- Kibana no longer needs to support OIDC, we have moved away from this functionality a number of releases ago. (I believe EEv3.7?)
- This means that the relying party config is no longer used and is removed in this PR.

### DIFF
--- a/pkg/render/manager.go
+++ b/pkg/render/manager.go
@@ -265,9 +265,6 @@ func (c *managerComponent) managerDeployment() *appsv1.Deployment {
 
 // managerVolumes returns the volumes for the Tigera Secure manager component.
 func (c *managerComponent) managerVolumeMounts() []corev1.VolumeMount {
-	if c.cfg.KeyValidatorConfig != nil {
-		return c.cfg.KeyValidatorConfig.RequiredVolumeMounts()
-	}
 	return []corev1.VolumeMount{}
 }
 
@@ -290,9 +287,6 @@ func (c *managerComponent) managerVolumes() []corev1.Volume {
 			c.cfg.InternalTrafficSecret.Volume(),
 			c.cfg.TunnelSecret.Volume(),
 		)
-	}
-	if c.cfg.KeyValidatorConfig != nil {
-		v = append(v, c.cfg.KeyValidatorConfig.RequiredVolumes()...)
 	}
 
 	return v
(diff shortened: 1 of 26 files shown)


pkg/render/render_test.go
```diff
@@ -124,7 +124,7 @@ func allCalicoComponents(
 	return []render.Component{namespaces, secretsAndConfigMaps, render.Typha(typhaCfg), render.Node(nodeCfg), kubecontrollers.NewCalicoKubeControllers(kcCfg), render.Windows(winCfg), nodeCertComponent}, nil
 }
 
-var _ = FDescribe("Rendering tests", func() {
+var _ = Describe("Rendering tests", func() {
```
rene-dekker, line 127: !

rene-dekker: See also: https://github.com/tigera/operator/pull/1823

pkg/render/manager.go
```diff
@@ -418,6 +412,7 @@ func (c *managerComponent) managerProxyContainer() corev1.Container {
 		{Name: "VOLTRON_PACKET_CAPTURE_CA_BUNDLE_PATH", Value: c.cfg.TrustedCertBundle.MountPath()},
 		{Name: "VOLTRON_PROMETHEUS_CA_BUNDLE_PATH", Value: c.cfg.TrustedCertBundle.MountPath()},
 		{Name: "VOLTRON_COMPLIANCE_CA_BUNDLE_PATH", Value: c.cfg.TrustedCertBundle.MountPath()},
+		{Name: "VOLTRON_DEX_CA_BUNDLE_PATH", Value: c.cfg.TrustedCertBundle.MountPath()},
```
Brian-McM, line 415: Are you just adding this for completeness, or is this value changing from the default?

pkg/render/manager.go
```diff
@@ -264,9 +264,6 @@ func (c *managerComponent) managerDeployment() *appsv1.Deployment {
 
 // managerVolumes returns the volumes for the Tigera Secure manager component.
 func (c *managerComponent) managerVolumeMounts() []corev1.VolumeMount {
-	if c.cfg.KeyValidatorConfig != nil {
```
Brian-McM, line 267: I'm sure it's removed because it's unneeded, but could you explain why we can remove this please? It's not immediately intuitive to me.

Brian-McM (approved):

pkg/render/manager.go
rene-dekker: It is changing to the trusted bundle. Previously, it was using the default, which was "/etc/ssl/certs/tls-dex.crt".

pkg/render/manager.go
rene-dekker: The required volume that was previously mounted contained the certificate that Dex presents. It is no longer required, because in `pkg/controller/manager/manager_controller.go`, this certificate is added to the trusted bundle.

rene-dekker: I did some digging. And thanks to you I found out that we cannot remove this. It is needed for OIDCTypeTigera, which was implemented for an edge case with a customer (and maybe also used in cloud).
However, we can modify this. We do not need the Volume, because it is already there, but we do need the mount somewhere in the /etc/ssl/certs folder, which is alpines location for trusted certificates.
Thank you!

Brian-McM (approved):

PR #1832, opened 2022-03-11
Use  certificatemanagement for Elastic secrets and certificates

rene-dekker:
Changes:
- Unused files and functions have been deleted with regards to crypto or tls.
- Elastic certificates are now signed by the tigera-ca-private secret.
- Some volume (mounting) decoration was removed in favour of using the TrustedBundle methods. Since the bundle is always mounted in the same location we do not want to have any TrustedBundle logic in the elasticsearch decorators that may conflict.
- VolumeMount() now requires the OSType param, to be windows compatible.


pkg/render/kubecontrollers/kube-controllers.go
```diff
@@ -220,16 +212,7 @@ func (c *kubeControllersComponent) Objects() ([]client.Object, []client.Object)
 		c.controllersDeployment(),
 	}
 	objectsToDelete := []client.Object{}
-	if c.renderManagerInternalSecret {
```
Brian-McM, line 223: This is just left over cruft that's not needed anymore since we've moved to using the gateway, right?

Brian-McM (approved):

PR #1846, opened 2022-03-17
Respect felixconfig health port

caseydavenport:
## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/installation/core_controller.go
+++ b/pkg/controller/installation/core_controller.go
@@ -114,8 +114,10 @@ const (
 //   4. Once the calico-node ClusterRoleBinding finalizer is removed we have cleaned up everything
 //      necessary so we can remove the Installation finalizer and we're done.
 
-var log = logf.Log.WithName("controller_installation")
-var openshiftNetworkConfig = "cluster"
+var (
+	log                    = logf.Log.WithName("controller_installation")
+	openshiftNetworkConfig = "cluster"
+)
 
 // Add creates a new Installation Controller and adds it to the Manager. The Manager will set fields on the Controller
 // and Start it when the Manager is Started.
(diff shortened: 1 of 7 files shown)


pkg/controller/installation/core_controller.go
```diff
@@ -1177,6 +1180,22 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 			nodeTerminating = true
 		}
 	}
+
+	// Determine the felix health port to use. Prefer the configuration from FelixConfiguration,
+	// but default to 9099.
+	felixHealthPort := 9099
+	if felixConfiguration.Spec.HealthPort != nil {
```
tmjd, line 1187: I think you're still probably working/thinking about things for this PR but I would consider defaulting the felixConfiguration HealthPort and writing that back to the datastore.
Do you see any issue with doing that?

howels:
Are you able to incorporate a similar change for Typha Healthport also?

caseydavenport:
> Are you able to incorporate a similar change for Typha Healthport also?

There's no datastore config for this at the moment, so I can't do the exact same thing. However, I can probably make it so that Typha uses a port based on the felix port (e.g., typhaPort = felixPort + 1)

howels:
> > Are you able to incorporate a similar change for Typha Healthport also?
> 
> There's no datastore config for this at the moment, so I can't do the exact same thing. However, I can probably make it so that Typha uses a port based on the felix port (e.g., typhaPort = felixPort + 1)

If you are happy to do so, that is a step in the right direction and would help us to avoid clashes with existing applications.

caseydavenport:
Pushed a new commit that sets the typha port to be `felixPort -1`, which matches the old default behavior. 

@tmjd I think this is ready for a review.

tmjd (approved):
LGTM, a couple notes though.
So this is actually going to be a change for OpenShift clusters I think, Typha will now be changing from 9098 to 9198. Hopefully not a big deal but something to be aware of (please correct me if I missed something and that is not the case).

Also I'm not sure I like the `-1` for typha happening in the typha render, I think it would be more appropriate in the controller but not enough to block this PR, we can always change it later.

howels:
Thanks for doing this work. When is the next release planned now that this is included @caseydavenport ?

PR #1848, opened 2022-03-18
[master] Add missing rbac for kube controllers

rene-dekker:
For enterprise, some RBAC was missing for the dpi controller to work successfully.

### DIFF
--- a/pkg/render/kubecontrollers/kube-controllers.go
+++ b/pkg/render/kubecontrollers/kube-controllers.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2019 Tigera, Inc. All rights reserved.
+// Copyright (c) 2019,2022 Tigera, Inc. All rights reserved.
 
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


pkg/render/kubecontrollers/kube-controllers.go
```diff
@@ -342,19 +342,19 @@ func kubeControllersRoleEnterpriseCommonRules(cfg *KubeControllersConfiguration)
...
+			APIGroups: []string{"projectcalico.org", "crd.projectcalico.org"},
 			Resources: []string{"deeppacketinspections"},
 			Verbs:     []string{"get", "watch", "list"},
 		},
 		{
 			APIGroups: []string{"crd.projectcalico.org"},
-			Resources: []string{"deeppacketinspections"},
-			Verbs:     []string{"get"},
+			Resources: []string{"deeppacketinspections/status"},
+			Verbs:     []string{"update"},
```
Brian-McM, line 352: We don't need get?

rene-dekker: This diff view makes it look very confusing.

I added "crd.projectcalico.org" in line 345. So in actuality, the "crd.projectcalico.org" now has added watch and list.

And for packetcaptures, it has added get, list and watch:

image

Brian-McM: Ohhh right, gotcha. Get for the status is just get on the resource I'm not realising.

Brian-McM (approved):

Issue #1099, opened 2021-01-27
Add support for additional labels and annotations to be propagated to all pods


Chili-Man:
We make use of pod labels and annotations to [tag metrics](https://docs.datadoghq.com/agent/kubernetes/tag/?tab=containerizedagent#pod-labels-as-tags) and [configure checks](https://docs.datadoghq.com/agent/kubernetes/integrations/?tab=kubernetes) for pods with DataDog. However, due to the operator managing all of the underlying Kubernetes resources, there is in't a good way to automatically add the labels and configuration during the installation. We would like to see support for passing additional labels and annotations to each of the underlying components that get deployed by this operator, in particular for the `Node`, `Typha`, and `Kubecontrollers`.  

## Expected Behavior

The operator should allow passing user specified custom labels and annotations to the pods  

## Current Behavior

We are currently not able to propogate additional labels or annotations to the operator managed pods. 

## Possible Solution

The `Installation` CRD should have a field like:

```yaml
apiVersion: operator.tigera.io/v1
kind: Installation
metadata:
  name: default
  namespace: management
spec:
    componentLabels:
      - componentName: Node
        podLabels:
            foo: bar

   componentAnnotations:
      - componentName: Typha
        podAnnotations:
          cool: awesome
```

## Steps to Reproduce (for bugs)

N/A

## Context

We would like to add additional labels and annotations to the pods for monitoring and running checks through DataDog 

## Your Environment

* AWS EKS with Amazon Linux 2 nodes

tmjd:
Sorry we missed this request and took so long to respond. This seems like a reasonable request. I'm not sure of the behavior we would want when operator is already setting an annotation or label and the user provided the same one. I'm not sure if one should take precedence or if it should result in reporting degraded in tigerastatus.

Your proposal follows the pattern we started with `componentResources` but I'm not sure I like the pattern we started. I think I like the following better. I'd possibly even consider adding podResources and deprecating `componentResources` in the future.
```
componentMeta:
  - name: Typha
    podAnnotations: ....
    podLabels: ...
```

@caseydavenport WDYT?

I'm not sure of the best way to handle this, does anyone have any good examples of how others handled this? I think I've seen others allow the whole podTemplate be defined but I think we want to keep this operator more opinionated than that, so I don't think that is an option.

caseydavenport:
[...]

PR #1851, opened 2022-03-22
#1099 Add extra labels and annotations to pods

JannisM11:
## Description

- Enable custom annotations & labels for Typha, Node & KubeController
- some details on _why_ this PR should be merged
- It was tested with an installation
```
kind: Installation
metadata:
  name: default
spec:
  componentResources:
    - componentName: Typha
      labels:
        not-k8s-app: custom
      annotations:
        mycustomannotation: foo
```
- The componentResources[] have two new fields -> labels and annotations
- PR for #1099

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


pkg/controller/installation/validation.go
```diff
@@ -344,6 +344,10 @@ func validateCustomResource(instance *operatorv1.Installation) error {
 		if _, ok := validComponentNames[resource.ComponentName]; !ok {
 			return fmt.Errorf("Installation spec.ComponentResources.ComponentName %s is not supported", resource.ComponentName)
 		}
+
+		if _, found := resource.Labels["k8s-app"]; found {
```
caseydavenport, line 348: I think we'll also need to validate that the given label keys and values are valid Kubernetes label syntax.

caseydavenport: Might be doable via kubebuilder? https://book.kubebuilder.io/reference/markers/crd-validation.html

If not, then we should add the check here.

IoannisMatzaris: 1) kubebuilder does not seem to have a way to validate the map[string]string.

2) If I try to create a non conform label, it is already logged:
`{"level":"error","ts":1647979877.3085115,"logger":"controller_installation","msg":"Error creating / updating resource","Request.Namespace":"tigera-operator","Request.Name":"node-certs","error":"Deployment.apps \"calico-typha\" is invalid: [spec.template.labels: Invalid value: \"_batch.kubernetes.io/job-completion-index\": prefix part a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*'), spec.template.labels: Invalid value: \"3/\": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')]","stacktrace":"sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem\n\t/home/ioannis/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.9.7/pkg/internal/controller/controller.go:253\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2\n\t/home/ioannis/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.9.7/pkg/internal/controller/controller.go:214"}
`

Writing an extra check will not help, in order to throw an error on `kubectl apply -f`

caseydavenport: Yeah, the benefit of doing it here would be to prevent us attempting to write invalid resource to the API server and giving us an opportunity to provide a more targeted error message to the user via `TigeraStatus` - both relatively minor things but probably worth doing. 

We try to avoid sending requests to the apiserver that we know are going to fail

caseydavenport:
Also, make sure you sign the CLA to make the bot happy - we won't be able to merge otherwise.

pkg/controller/installation/validation.go
IoannisMatzaris: It sounds reasonable. I pushed the validation on the branch

pkg/controller/installation/validation.go
```diff
@@ -18,13 +18,19 @@ import (
...
 	"path"
+	"regexp"
 	"strings"
 
 	operatorv1 "github.com/tigera/operator/api/v1"
 	"github.com/tigera/operator/pkg/render"
 	appsv1 "k8s.io/api/apps/v1"
 )
 
+var (
```
caseydavenport, line 29: I think you actually want to just import the Kubernetes validation logic for these. For example:

- For label values: https://github.com/kubernetes/apimachinery/blob/master/pkg/util/validation/validation.go#L167
- For label+annotation keys: https://github.com/kubernetes/apimachinery/blob/master/pkg/util/validation/validation.go#L188

Annotation values are free-form and not structured so I don't think there's any validation we should perform on them.

IoannisMatzaris: For the label key, this seems to be the right one: https://github.com/kubernetes/apimachinery/blob/master/pkg/util/validation/validation.go#L43

gdziwoki:
Any way I can help with this feature? Docs?

lmm:
I think this PR can be closed since #2063 has added the support for extra labels/annotations.

PR #1858, opened 2022-03-24
Certificatemanager controller

rene-dekker:
Although not common to reproduce, I found a way where there is a conflict in the code with the new certificate code:
- The tigera-ca-private is created by core_controller, but only if everything is ok.
- Things are not always ok. For instance, when you have an invalid prometheus tls secret.
- If the tigera-ca-private secret is not yet existent a new one is created in memory.
These  conditions make that prometheus keeps reconciling and changing, because  the core_controller does not complete successfully.

To solve this a new controller was added that creates the tigera-ca-private secret.

---UPDATE---
I turned this into a draft. Some elements that are currently missing:
- Controllers should wait for this controller to be status=ready
- CreateCertificatemanager(...) should be broken down into two methods: one for creating the secret and one for using the existing secret.

### DIFF
--- a/pkg/controller/certificatemanager/certificatemanager_controller.go
+++ b/pkg/controller/certificatemanager/certificatemanager_controller.go
@@ -1,0 +1,134 @@
+// Copyright (c) 2022 Tigera, Inc. All rights reserved.
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package certificatemanager
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	operator "github.com/tigera/operator/api/v1"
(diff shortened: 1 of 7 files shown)


pkg/controller/certificatemanager/certificatemanager_controller.go
```diff
@@ -0,0 +1,134 @@
+// Copyright (c) 2022 Tigera, Inc. All rights reserved.
```
Brian-McM, line 1: If you're adding a controller to this package you'll need to move the "utility" functions somewhere else, like into util under a subpackage. If you don't you're you're breaking the contract that controllers don't access each other packages (and I think the other controllers access this package).

pkg/controller/certificatemanager/certificatemanager_controller_test.go
```diff
@@ -0,0 +1,96 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package certificatemanager
+
+import (
+	"context"
+
+	. "github.com/onsi/ginkgo"
```
Brian-McM, line 20: Nit, please organize the imports the same way we do for other packages.

Brian-McM:
> when you have an invalid prometheus tls secret.
Could you elabaorate on how an invalid prometheus tls secret stops the core controller from reconciling please?

Brian-McM (commented):
What, if anything, depends on the items created by this new controller? Do items created by the core controller depend on what's created by this new controller? 

It seems like this should probably be the first thing that's created, since this creates the root CA, does it not?

PR #1860, opened 2022-03-25
Support for migrating a BPF cluster to operator install

sridhartigera:
## Description

This PR adds support for migrating a BPF cluster from manifest install to operator based install.
Felix configuration is read and dataplane is set to BPF in the installation spec of the operator. 
## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/migration/convert/bpf.go
+++ b/pkg/controller/migration/convert/bpf.go
@@ -1,0 +1,42 @@
+// Copyright (c) 2022 Tigera, Inc. All rights reserved.
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package convert
+
+import (
+	"fmt"
+
+	operatorv1 "github.com/tigera/operator/api/v1"
+	crdv1 "github.com/tigera/operator/pkg/apis/crd.projectcalico.org/v1"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+// handleBPF is a migration handler which ensures BPF configuration is carried forward.
+func handleBPF(c *components, install *operatorv1.Installation) error {
+	felixConfiguration := &crdv1.FelixConfiguration{}
+	err := c.client.Get(ctx, types.NamespacedName{Name: "default"}, felixConfiguration)
+	if err != nil {
+		return fmt.Errorf("error reading felixconfiguration %w", err)
+	}
+	if felixConfiguration.Spec.BPFEnabled != nil && *felixConfiguration.Spec.BPFEnabled {
+		if install.Spec.CalicoNetwork == nil {
+			install.Spec.CalicoNetwork = &operatorv1.CalicoNetworkSpec{}
+		}
+
+		bpf := operatorv1.LinuxDataplaneBPF
+		install.Spec.CalicoNetwork.LinuxDataplane = &bpf
+		install.Spec.CalicoNetwork.HostPorts = nil
+	}
+	return nil
+}
--- a/pkg/controller/migration/convert/bpf.go
+++ b/pkg/controller/migration/convert/bpf.go
@@ -20,8 +20,48 @@ import (
 
 	operatorv1 "github.com/tigera/operator/api/v1"
 	crdv1 "github.com/tigera/operator/pkg/apis/crd.projectcalico.org/v1"
+	"github.com/tigera/operator/pkg/common"
+	"github.com/tigera/operator/pkg/render"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 )
+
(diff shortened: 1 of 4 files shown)


pkg/controller/migration/convert/bpf.go
```diff
@@ -0,0 +1,42 @@
...
+)
+
+// handleBPF is a migration handler which ensures BPF configuration is carried forward.
+func handleBPF(c *components, install *operatorv1.Installation) error {
+	felixConfiguration := &crdv1.FelixConfiguration{}
+	err := c.client.Get(ctx, types.NamespacedName{Name: "default"}, felixConfiguration)
+	if err != nil {
+		return fmt.Errorf("error reading felixconfiguration %w", err)
+	}
+	if felixConfiguration.Spec.BPFEnabled != nil && *felixConfiguration.Spec.BPFEnabled {
```
caseydavenport, line 32: Do we need to check the environment variables on the daemonset as well? 

I think users can enable BPF via an env var and we'll need to support that as well.

caseydavenport: Also, I think we'll need to handle extracting the API server endpoint to provision the config map for accessing the API server?

sridhartigera: Added changes to check for BPFENABLED env var in the bpf handler.

sridhartigera: I don't think we need to extract the API server endpoint as the BPF programs are attached to the interfaces during migration and the API server is reachable via the kubernetes service during migration.

pkg/controller/migration/convert/bpf.go
```diff
@@ -0,0 +1,90 @@
...
+func copyK8sServicesEPConfigMap(c *components) error {
+	// Extract end point host, port from configmap in kube-system namespace
+	cmName := render.K8sSvcEndpointConfigMapName
+	cm := &corev1.ConfigMap{}
+	cmNamespacedName := types.NamespacedName{
+		Name:      cmName,
+		Namespace: "kube-system",
+	}
+	if err := c.client.Get(ctx, cmNamespacedName, cm); err != nil {
+		return fmt.Errorf("Failed read to ConfigMap %q: %s", cmName, err)
```
caseydavenport, line 39: this is a super nit, but error strings generally start lowercase - e.g., `"failed to read" instead of "Failed to read" 

Makes them compose a bit better if error handling higher up the stack composes them.

pkg/controller/migration/namespace_migration.go
```diff
@@ -248,6 +250,11 @@ func (m *CoreNamespaceMigration) Run(ctx context.Context, log logr.Logger) error
 	if err := m.deleteKubeSystemTypha(ctx); err != nil {
 		return fmt.Errorf("failed to delete kube-system typha Deployment: %s", err.Error())
 	}
+	log.V(1).Info("kube-system typha deployment deleted")
+	if err := m.deleteKubeSystemServiceEndPointCm(ctx); err != nil {
+		return fmt.Errorf("failed to delete kube-system k8sServicesEndpoint ConfigMap: %s", err.Error())
+	}
+	log.V(1).Info("kube-system services endpoint configmap deleted")
```
caseydavenport, line 257: Should probably only log this if we actually delete the configmap (a large number of clusters won't have this configmap, and for those we don't want to log anything)

Can just move the log within `deleteKubeSystemServiceEdnpoint`

pkg/controller/migration/namespace_migration.go
```diff
@@ -357,6 +364,16 @@ func (m *CoreNamespaceMigration) CleanupMigration(ctx context.Context) error {
 	return nil
 }
 
+// deleteKubeSystemServiceEndPointCm deletes the kubernetes-services-endpoint configmap
+// in the kube-system namespace
+func (m *CoreNamespaceMigration) deleteKubeSystemServiceEndPointCm(ctx context.Context) error {
```
caseydavenport, line 369: "Cm" is short for configMap, so should be same-case given golang conventions (e.g., EndpointCM instead of EndpointCm)

Alternatively, just spell it out - it's already a long function name and ConfigMap is more clear than CM

pkg/controller/migration/convert/bpf_test.go
```diff
@@ -0,0 +1,155 @@
...
+	crdv1 "github.com/tigera/operator/pkg/apis/crd.projectcalico.org/v1"
+	"github.com/tigera/operator/pkg/common"
+	"github.com/tigera/operator/pkg/render"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+	kscheme "k8s.io/client-go/kubernetes/scheme"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+)
+
+var cmName = render.K8sSvcEndpointConfigMapName
```
caseydavenport, line 35: Multiple `var` lines at the top should be included within a multi-var stanza:

```go
var (
  cmName = ...
  cmData = ...
)
```

pkg/controller/migration/convert/bpf_test.go
```diff
@@ -0,0 +1,155 @@
...
+	)
+
+	BeforeEach(func() {
+		comps = emptyComponents()
+		i = &operatorv1.Installation{}
+		f = emptyFelixConfig()
+	})
+
+	It("converts bpfenabled felixconfig set to true", func() {
+		Expect(apis.AddToScheme(scheme)).ToNot(HaveOccurred())
```
caseydavenport, line 75: Should just do this in BeforeEach if we're doing it for every test. Looks like boilerplate. 

Although, we might even just be able to do it in `init`  to be honest - don't think it needs to be called more than once per process.

caseydavenport (approved):
Still missing the log capitalization change but that's minor and otherwise this LGTM.

PR #1877, opened 2022-03-31
Migrate IPv6 and dual-stack clusters

### LINKED ISSUE #1503: Operator migration: unsupported variables
## Expected Behavior

As long as nothing too outlandish is being done, the migration should detect options and migrate to use the calico operator

## Current Behavior

While migrating I hit a number of areas where it wouldn't let me go while still doing what seem to be very reasonable things:

- FELIX_IPV6SUPPORT=true is not supported
- IP6=autodetect is not supported
- IP_AUTODETECTION_METHOD=cidr=10.21.32.0/23 is not supported. To fix it, remove the IP_AUTODETECTION_METHOD env var or set it to 'first-found', 'can-reach=*', 'interface=*', or 'skip-
interface=*'

I should be able to get around it, but the last one is a particular issue because not all of my nodes have identical hardware, so autodetecting by interface name isn't a good option.

## Context

It doesn't look like it's possible for me to migrate without losing at least ipv6 support until after the migration; the IP_AUTODETECTION_METHOD issue makes migration iffy at best, since if I make a mistake the networking between my pods could get messed up; I use zerotier on my nodes to allow me to communicate with them from anywhere and sometimes calico tries to use that as the interface.

## Your Environment

Ubuntu 20.04, kubeadm cluster 1.22.0

lmm:
## Description

This PR adds migration support for IPv6 and dual-stack clusters and fixes #1503.

I think most of the logic is here. ~~Still need to address `CALICO_ROUTER_ID` - do we want to support custom values? We set this to `hash` currently.~~ Edit: this is now addressed and clarified in the PR.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/migration/convert/network_test.go
+++ b/pkg/controller/migration/convert/network_test.go
@@ -18,6 +18,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/intstr"
 	kscheme "k8s.io/client-go/kubernetes/scheme"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 )
 
@@ -103,12 +104,6 @@ var _ = Describe("Convert network tests", func() {
 			Expect(*cfg.Spec.CalicoNetwork.BGP).To(Equal(operatorv1.BGPEnabled))
 		})
 		It("should convert Calico v3.15 manifest", func() {
-			v4pool = crdv1.NewIPPool()
-			v4pool.Spec = crdv1.IPPoolSpec{
-				CIDR:        "192.168.4.0/24",
-				IPIPMode:    crdv1.IPIPModeAlways,
-				NATOutgoing: true,
-			}
 			c := fake.NewFakeClientWithScheme(scheme, append([]runtime.Object{v4pool, emptyFelixConfig()}, calicoDefaultConfig()...)...)
 			cfg, err := Convert(ctx, c)
 			Expect(err).NotTo(HaveOccurred())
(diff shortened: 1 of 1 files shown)


pkg/controller/migration/convert/network.go
```diff
@@ -93,15 +93,55 @@ func handleCalicoCNI(c *components, install *operatorv1.Installation) error {
...
+	}
+
+	ip6, err := c.node.getEnv(ctx, c.client, containerCalicoNode, "IP6")
+	if err != nil {
 		return err
 	}
+	// IP6 can be 'autodetect', 'none', or not defined.
+	if ip6 != nil {
+		if *ip6 == "none" {
+			if err := c.node.assertEnvIsSet(ctx, c.client, containerCalicoNode, "FELIX_IPV6SUPPORT", "false"); err != nil {
```
caseydavenport, line 123: Is it also valid to have `FELIX_IPv6SUPPORT` just not set at all?

pkg/controller/migration/convert/network.go
```diff
@@ -93,15 +93,55 @@ func handleCalicoCNI(c *components, install *operatorv1.Installation) error {
...
+				return err
+			}
+		} else if *ip6 == "autodetect" {
+			if err := handleIPv6AutoDetectionMethod(c, install); err != nil {
+				return err
+			}
+			if err := c.node.assertEnvIsSet(ctx, c.client, containerCalicoNode, "FELIX_IPV6SUPPORT", "true"); err != nil {
```
caseydavenport, line 130: Same here - what if this is set via FelixConfiguration instead of an env var? 

We can probably drop this assertion and not lose much - WDYT?

lmm: Yeah you're right, this should be changed.

Another option is to use `assertEnv` instead to check that IF `FELIX_IPV6SUPPORT` is defined then it has to be `true`, but maybe we have to assume the starting cluster was valid.

lmm: We chatted about this in Slack and agreed the best option might be to validate the env var if it is set.

caseydavenport (approved):

pkg/controller/migration/convert/ippools.go
```diff
@@ -77,13 +77,21 @@ func handleIPPools(c *components, install *operatorv1.Installation) error {
 				}
 			}
 		} else {
-			if v4pool != nil {
+			// If assign_ipv4="false" then remove any discovered IPv4 pools.
+			// Since we must have an IPv6 pool here, replace the operator
+			// pools with the IPv6 pool.
+			// This is needed because the operator renders IPAM config
+			// based on the presence of initial ippools in the installation CR.
+			ipv6Pool := render.GetIPv6Pool(install.Spec.CalicoNetwork.IPPools)
```
caseydavenport, line 85: Do we need to requery here or couldn't we use `v6pool` from above?

pkg/controller/migration/convert/network_test.go
```diff
@@ -527,63 +514,355 @@ var _ = Describe("Convert network tests", func() {
...
+					Entry("ipv4_pools", `"ipv4_pools": ["10.0.0.0/24"]`),
+					Entry("ipv6_pools", `"ipv6_pools": ["2001:db8::1/120"]`),
+					Entry("both pools", `"ipv4_pools": ["10.0.0.0/24"], "ipv6_pools": ["2001:db8::1/120"]`),
+				)
+			})
+		})
+
+	})
+
+	Context("with IPv6 pool", func() {
```
caseydavenport, line 536: Do we need a "With an IPv4 and IPv6 pool" context?

caseydavenport (approved):
LGTM - couple of questions

pkg/controller/migration/convert/ippools.go
lmm: Yeah that wasn't right, I fixed that.

pkg/controller/migration/convert/network_test.go
lmm: Yep that didn't make sense either. I removed those, added a v6 pool to the top and overhauled the tests.

caseydavenport (approved):

PR #1878, opened 2022-03-31
Add the internal-manager-tls pem to the trusted bundle for es-proxy

rene-dekker:
Es proxy needs to trust the internal-manager-tls certificate for mcm requests.


doublek (approved):
LGTM

Brian-McM (approved):

PR #1918, opened 2022-04-14
[v1.26] Allow downgrading from v1.26+ to v1.25.x

rene-dekker:
For Calico Cloud it is not an uncommon scenario that a v1.26+ cluster is turned into a v1.25.x cluster. Some standardization and refactoring are making it hard to downgrade, causing a bad experience. This PR addresses these compatibility issues.

### DIFF
--- a/pkg/controller/installation/core_controller.go
+++ b/pkg/controller/installation/core_controller.go
@@ -1416,7 +1416,12 @@ func (r *ReconcileInstallation) GetTyphaNodeTLSConfig() (*render.TyphaNodeTLS, e
 			_, okCN := node.Data[render.CommonName]
 			_, okUS := node.Data[render.URISAN]
 			if !(okCN || okUS) {
-				errMsgs = append(errMsgs, fmt.Sprintf("CertPair for Felix does not contain common-name or uri-san"))
+				if node.Data[corev1.TLSCertKey] != nil {
+					// This is a cluster downgraded from v1.26+, where the common name is assumed to be render.FelixCommonName.
+					node.Data[render.CommonName] = []byte(render.FelixCommonName)
+				} else {
+					errMsgs = append(errMsgs, fmt.Sprintf("CertPair for Felix does not contain common-name or uri-san"))
+				}
 			}
 		}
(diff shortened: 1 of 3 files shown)


pkg/controller/installation/core_controller.go
```diff
@@ -1436,11 +1441,28 @@ func (r *ReconcileInstallation) GetTyphaNodeTLSConfig() (*render.TyphaNodeTLS, e
...
 	}
 
+	if ca == nil && typha != nil && typha.Data[corev1.TLSCertKey] != nil && node != nil && node.Data[corev1.TLSCertKey] != nil {
+		// This is a cluster downgraded from v1.26+, where the configmap is no longer automatically created.
+		// We can create a compatible configmap by combining the PEM data of both certificates.
+		ca = &corev1.ConfigMap{
+			TypeMeta:   metav1.TypeMeta{Kind: "ConfigMap", APIVersion: "v1"},
+			ObjectMeta: metav1.ObjectMeta{Name: render.TyphaCAConfigMapName, Namespace: common.OperatorNamespace()},
+			Data: map[string]string{
+				render.TyphaCABundleName: fmt.Sprintf("# certificate: node-certs\n%s\n\n# certificate: node-certs\n%s", string(typha.Data[corev1.TLSCertKey]), string(node.Data[corev1.TLSCertKey])),
```
tmjd, line 1461: This is what I meant earlier. Both say node-certs, I think the 1st was intended to say typha-certs.
I didn't realize a bundle was literally the contents of multiple certs in the same file.

rene-dekker: I'll rectify the typo.

If you follow the docs you would likely only put the CA PEM in here, but placing the two individual certificate PEMs in here works as well as an alternative.

tmjd (approved):
You've mentioned this works for you and it looks good AFAICT. I think my concern I mentioned in slack is unrelated to this specific change. It also looks like you've guarded the changes enough that this should not affect anything unexpectedly.
tldr; this LGTM :grinning:

Issue #1055, opened 2021-01-06
The CR "installations" deployed by tigera-operator is missing "conditions" as part of the status


sirajyasin:
The CR **installations** deployed by tigera-operator is missing "conditions" as part of the status. This creates issue verifying the resources applied. As an example the [cli-utils](https://github.com/kubernetes-sigs/cli-utils) status poller (through applier) fails to fetch status of this resource due to [condition](https://github.com/kubernetes-sigs/cli-utils/tree/master/pkg/kstatus#conditions) not satisfied. 
I can understand that the status of calico installation can be tracked through tigerastatus CR, but then is it possible to have "conditions" added for installations for consistency.

```
kubectl get installations -o json | jq -r ".items[0].status"
{
  "mtu": 1440,
  "variant": "Calico"
}
```
## Expected Behavior
"conditions" should be available as part of the status

## Current Behavior

Status has just mtu and variant and missing "conditions"

## Possible Solution

add "conditions" to status in addition to mtu and variant

## Steps to Reproduce (for bugs)

1. Deploy tigera-operator
2. kubectl get installations -o json | jq -r ".items[0].status"

## Context
This creates issue verifying the resources applied.

## Your Environment

* Operating System and version:
* Link to your project (optional):

caseydavenport:
Seems like we already have the information available to implement the "Ready" condition in use for the TigeraStatus, so we can mirror that into conditions in the respective CRs. Anyone interested in taking a stab at this?

tmjd:
I believe this has been addressed with #1954 and is available in the latest Calico release.

PR #1930, opened 2022-04-28
Fixing Installation CR conditions

vara2504:
## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/api/v1/installation_types.go
+++ b/api/v1/installation_types.go
@@ -605,6 +605,7 @@ type InstallationStatus struct {
 	// +optional
 	Computed *InstallationSpec `json:"computed,omitempty"`
 
+	// Conditions represents the latest observed set of conditions for this component. A component may be one or more of
 	// Available, Progressing, or Degraded.
 	Conditions []TigeraStatusCondition `json:"conditions,omitempty"`
 }
--- a/api/v1/installation_types.go
+++ b/api/v1/installation_types.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2020 Tigera, Inc. All rights reserved.
+// Copyright (c) 2020-2022 Tigera, Inc. All rights reserved.
 /*
 
 Licensed under the Apache License, Version 2.0 (the "License");
@@ -604,6 +604,9 @@ type InstallationStatus struct {
 	// Computed is the final installation including overlaid resources.
 	// +optional
 	Computed *InstallationSpec `json:"computed,omitempty"`
+
+	// Available, Progressing, or Degraded.
+	Conditions []TigeraStatusCondition `json:"conditions,omitempty"`
 }
 
 // +kubebuilder:object:root=true
--- a/pkg/controller/installation/core_controller.go
+++ b/pkg/controller/installation/core_controller.go
@@ -750,19 +750,19 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 		// update Installation resource with existing install if it exists.
 		nc, err := convert.NeedsConversion(ctx, r.client)
 		if err != nil {
-			r.SetDegraded("Error checking for existing installation", err, instance, ctx, reqLogger)
+			r.SetDegraded(ctx, instance, "Error checking for existing installation", err, reqLogger)
 			return reconcile.Result{}, err
 		}
 		if nc {
 			install, err := convert.Convert(ctx, r.client)
 			if err != nil {
 				if errors.As(err, &convert.ErrIncompatibleCluster{}) {
-					r.SetDegraded("Existing Calico installation can not be managed by Tigera Operator as it is configured in a way that Operator does not currently support. Please update your existing Calico install config", err, instance, ctx, reqLogger)
+					r.SetDegraded(ctx, install, "Existing Calico installation can not be managed by Tigera Operator as it is configured in a way that Operator does not currently support. Please update your existing Calico install config", err, reqLogger)
 					// We should always requeue a convert problem. Don't return error
 					// to make sure we never back off retrying.
 					return reconcile.Result{RequeueAfter: 15 * time.Second}, nil
 				}
-				r.SetDegraded("Error converting existing installation", err, instance, ctx, reqLogger)
+				r.SetDegraded(ctx, instance, "Error converting existing installation", err, reqLogger)
 				return reconcile.Result{}, err
 			}
 			instance.Spec = utils.OverrideInstallationSpec(install.Spec, instance.Spec)
@@ -771,7 +771,7 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 
 	// update Installation with defaults
 	if err := updateInstallationWithDefaults(ctx, r.client, instance, r.autoDetectedProvider); err != nil {
-		r.SetDegraded("Error querying installation", err, instance, ctx, reqLogger)
+		r.SetDegraded(ctx, instance, "Error querying installation", err, reqLogger)
 		return reconcile.Result{}, err
 	}
 
@@ -779,7 +779,7 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 
 	// Validate the configuration.
 	if err := validateCustomResource(instance); err != nil {
-		r.SetDegraded("Invalid Installation provided", err, instance, ctx, reqLogger)
+		r.SetDegraded(ctx, instance, "Invalid Installation provided", err, reqLogger)
 		return reconcile.Result{}, err
 	}
 
@@ -791,7 +791,7 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 		crbKey := types.NamespacedName{Name: "calico-node"}
 		err := r.client.Get(ctx, crbKey, &crb)
 		if err != nil && !apierrors.IsNotFound(err) {
-			r.SetDegraded("Unable to get ClusterRoleBinding", err, instance, ctx, reqLogger)
+			r.SetDegraded(ctx, instance, "Unable to get ClusterRoleBinding", err, reqLogger)
 			return reconcile.Result{}, err
 		}
 		found := false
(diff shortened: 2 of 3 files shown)


pkg/controller/status/status.go
```diff
@@ -691,7 +695,10 @@ func (m *statusManager) set(retry bool, conditions ...operator.TigeraStatusCondi
 			ts.Status.Conditions = append(ts.Status.Conditions, condition)
 		}
 	}
+	// Save the ts in the status manager
+	m.conditions = ts.Status.Conditions
 
+	// If nothing has changed, we don't need to update in the API.
```
rene-dekker, line 701: This line is duplicate.

pkg/controller/status/status.go
```diff
@@ -111,6 +112,9 @@ type statusManager struct {
 	// to determine whether we need to call Delete() on the object, without sending unnecessary
 	// get/delete calls to the API server.
 	crExists bool
+
+	// Available, Progressing, or Degraded.
```
rene-dekker, line 116: I think it is better to use the original comment we have in tigerastatus_types.

pkg/controller/installation/core_controller.go
```diff
@@ -1569,7 +1592,7 @@ func (r *ReconcileInstallation) updateCRDs(ctx context.Context, variant operator
 	// Installation CR will not remove the CRDs.
 	handler := utils.NewComponentHandler(log, r.client, r.scheme, nil)
 	if err := handler.CreateOrUpdateOrDelete(ctx, crdComponent, nil); err != nil {
-		r.SetDegraded("Error creating / updating CRD resource", err, log)
+		r.SetDegraded("Error creating / updating CRD resource", err, instance, ctx, log)
```
rene-dekker, line 1595: In golang, I think it is a convention to always pass ctx as the first argument

pkg/controller/installation/core_controller.go
```diff
@@ -917,86 +919,82 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
...
 		if err != nil {
-			r.status.SetDegraded(fmt.Sprintf("Error ensuring internal manager TLS certificate %q exists and has valid DNS names", render.ManagerInternalTLSSecretName), err.Error())
+			r.SetDegraded(fmt.Sprintf("Error ensuring internal manager TLS certificate %q exists and has valid DNS names", render.ManagerInternalTLSSecretName), err, instance, ctx, reqLogger)
 			return reconcile.Result{}, err
 		}
 	}
 
 	typhaNodeTLS, err := GetOrCreateTyphaNodeTLSConfig(r.client, certificateManager)
 	if err != nil {
 		log.Error(err, "Error with Typha/Felix secrets")
```
rene-dekker, line 965: All the log.Error logs outside of the setDegraded method are now redundant; they can all be removed.

api/v1/installation_types.go
```diff
@@ -604,6 +604,10 @@ type InstallationStatus struct {
 	// Computed is the final installation including overlaid resources.
 	// +optional
 	Computed *InstallationSpec `json:"computed,omitempty"`
+
+	// Conditions represents the latest observed set of conditions for this component. A component may be one or more of
+	// Available, Progressing, or Degraded.
+	Conditions []TigeraStatusCondition `json:"conditions,omitempty"`
```
rene-dekker, line 610: I think the type of the object should be []Condition using the metav1 import: "k8s.io/apimachinery/pkg/apis/meta/v1"

rene-dekker: I'm sorry. That is not compatible. Please disregard.

pkg/controller/installation/core_controller.go
```diff
@@ -1356,6 +1349,10 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 		instance.Status.ImageSet = imageSet.Name
 	}
 	instance.Status.Computed = &instance.Spec
+
+	// Update the installation status conditions from the tigerastatus.
+	r.updateInstallationStatus(instance, r.status.Conditions())
```
caseydavenport, line 1354: Unless I'm missing something, I think there are one or two issues with this approach as it's currently formulated. Namely:

- this line is only hit after `r.Status.IsAvailable()` is true. So, we'll only update the status for explicit Degraded calls, and once Available, but never while progressing (which is the time between those states).
- The StatusManager updates the TigeraStatus object asynchronously from this loop - It's possible that the status manager will mark the TigeraStatus as degraded at some point in the future, and this controller won't recognize that.

The good thing about the way you have done this, though, is that it doesn't require any additional queries on the Installation API, and is all synchronous with the main controller loop - meaning we won't have conflicts between two controllers attempting to update the same object. 

We might want to sacrifice that in order to make sure the status is updated consistently with and at the same time as the TigeraStatus API.

I'm not sure yet the best way to do that. One option might be to adjust the StatusManager to take a new argument - a function which updates conditions. It can call that function as part of its syncloop.

e.g.,


```go
// Define a new type to use as the argument.
type StatusUpdater func([]TigeraStatusCondition) error
```

```go
// Add the new argument to the StatusManager.
func New(client client.Client, component string, kubernetesVersion *common.VersionInfo, updateFunc StatusUpdater) StatusManager {
```

And then you can replace this line: https://github.com/tigera/operator/blob/6262d7798f5ca2e6a09e85b262eaa1cb91f59fdc/pkg/controller/status/status.go#L700

With `err := sm.updateFunc(ts.Status.Conditions)`

In order for this to work, we'd also need to move the rest of the status updating logic into that function as well - otherwise we'd have two goroutines trying to update the status at the same time, which would result in conflict errors being returned from the API server. It also has the downside of not having access to all of the fields and logic that may have occurred in the Reconcile() loop (e.g., imageSet).

I suppose another alternative would be to have the controller watch the TigeraStatus resource, so that any time the TigeraStatus resource changes we can query it, and update the Installation CR status based on what we find. 

Before you go down this path would want to hear from @tmjd to see if he has any other ideas or thoughts on these approaches.

caseydavenport (changes requested):
A few comments - I think we might need to adjust the architecture of this change slightly. Let's chat with @tmjd to get his thoughts as well.

pkg/controller/installation/core_controller.go
rene-dekker: "The StatusManager updates the TigeraStatus object asynchronously from this loop - It's possible that the status manager will mark the TigeraStatus as degraded at some point in the future, and this controller won't recognize that."

I was under the impression that the status manager does not set a `Degraded` condition by itself and that the controller is the authority on this condition.
At the same time, the status manager is the authority on the `Progressing` condition. 
Lastly, `Available` is the result of `!(Degraded | Progressing)`. 
Vara's approach is mostly correct, but once the status moves to `Progressing=True`, a lagging effect appears in the Installation.Status, because this is not noticed until the next reconciliation. At that time `Progressing` may be false again and therefore, `Progressing=true` may not be noticed at all.
I am writing this because I am personally wondering how big of a problem it is when missing `Progressing=true` is if it is short-lived. 

Rather than having two controllers update the Installation.Status, could the status manager send a message to a channel that the core_controller configures the status manager with to update the status?

tmjd: I think the TigeraStatus watching strategy seems like a possibly nice solution as far as keeping resources only updated from one routine (and not needing to fetch the resource before updating). We could possibly even have the Reconcile check that it is a tigerastatus update and skip the rest of the reconcile and only update the resource status.

I'm not understanding how the StatusUpdater could work, since as you mentioned it would need other field info from the Reconcile.

What about if we had a channel where the controller (Reconcile) would send the resource to the StatusManager and the StatusManager was responsible for making the call to update the Status?

@rene-dekker the Progressing portion, especially for calico, may be a significant time on a sizeable cluster.
I'm kind of suggesting something similar to your suggestion but the other way.
I'm not sure I understand what you're fully suggesting because I don't know what would be able to listen on the channel for changes from the Status manager, it would still be at the mercy of the Reconcile being triggered.

tmjd:
I'm assuming this PR is meant to address #1055. From the links in that issue, looking at [the conditions](https://github.com/kubernetes-sigs/cli-utils/tree/master/pkg/kstatus#conditions) I don't think this PR addresses it. We do not have a Ready condition and I think we should also start having an `observedGeneration` field in the status. I think we should also have a Reconciling condition.

pkg/controller/installation/core_controller.go
caseydavenport: > I was under the impression that the status manager does not set a Degraded condition by itself

@rene-dekker the status manager can mark a CR as degraded independently of the controller, if for example there are pods that are failing in some way: https://github.com/tigera/operator/blob/6262d7798f5ca2e6a09e85b262eaa1cb91f59fdc/pkg/controller/status/status.go#L467-L468

> I'm not understanding how the StatusUpdater could work, since as you mentioned it would need other field info from the Reconcile.

It would be awkward for some fields - we'd need to cache them on the Installation and then use a callback to a bound function on the reconciler that would set the fields appropriately. Probably not my favorite approach.

> What about if we had a channel where the controller (Reconcile) would send the resource to the StatusManager and the StatusManager was responsible for making the call to update the Status?

The problem is that the controller / reconciler doesn't know all the times that it should send the resource to the StatusManager - the StatusManager independently updates conditions even if no reconcile has occurred, and so the trigger needs to come from the other direction. 

> I think the TigeraStatus watching strategy seems like a possibly nice solution as far as keeping resources only updated from one routine (and not needing to fetch the resource before updating). We could possibly even have the Reconcile check that it is a tigerastatus update and skip the rest of the reconcile and only update the resource status.

Yep exactly - that's why I'm leaning towards this approach right now. Seems to have a solid separation of concerns.

rene-dekker: Sounds good. So to sum up there are some take-aways:
- You can add a watch for TigeraStatus with name "calico" in the func `add(...)`  core_controllers. This will make the controller reconcile whenever the TigeraStatus is modified in the datastore.
- Then in the start of the `Reconcile(...)` func, I think you can use the request parameter to figure out if the TigeraStatus was updated. I'm not sure if there might be a better way, since` reconcile.Request` only has the namespace and name to find out what was modified.

tmjd: > > What about if we had a channel where the controller (Reconcile) would send the resource to the StatusManager and the StatusManager was responsible for making the call to update the Status?

> The problem is that the controller / reconciler doesn't know all the times that it should send the resource to the StatusManager - the StatusManager independently updates conditions even if no reconcile has occurred, and so the trigger needs to come from the other direction.

I was imagining the StatusManager would hold the last sent resource (Installation CR) and use the last sent one during any updates until it received a new one. We might need to send it multiple times during a Reconcile, one at the start and one at the finish. If we include `observedGeneration` then that would include the information the StatusManager would need to determine the proper state of the "Reconciling" condition.

PR #1947, opened 2022-05-10
Update CRDs to the latest

hjiawei:
## Description

`make gen-files` and `make gen-versions`.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/crds/enterprise/crd.projectcalico.org_alertexceptions.yaml
+++ b/pkg/crds/enterprise/crd.projectcalico.org_alertexceptions.yaml
@@ -37,25 +37,30 @@ spec:
               description:
                 description: The description is displayed by the UI.
                 type: string
-              period:
-                description: Period controls how long an alert exception will be active.
-                  It is optional and omitting Period will make the alert exception
-                  active forever.
+              endTime:
+                description: EndTime defines the end time at which this alert exception
+                  will expire. If omitted the alert exception filtering will continue
+                  indefinitely.
+                format: date-time
                 type: string
               selector:
                 description: Selector defines a query string for alert events to be
                   excluded from UI search results.
                 type: string
+              startTime:
+                description: StartTime defines the start time from which this alert
+                  exception will take effect. If the value is in the past, matched
+                  alerts will be filtered immediately. If the value is changed to
+                  a future time, alert exceptions will restart at that time.
+                format: date-time
+                type: string
             required:
             - description
             - selector
+            - startTime
             type: object
           status:
             description: AlertExceptionStatus contains the status of an alert exception.
-            properties:
-              lastExecuted:
-                format: date-time
-                type: string
             type: object
         type: object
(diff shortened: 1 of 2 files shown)


rene-dekker (approved):
From what I understand these AlertExceptions have never been a documented feature in a release version. For this reason removing a field from the CRD should not cause issues for any users.

LGTM

rene-dekker (commented):
From what I understand these AlertExceptions have never been a documented feature in a release version. For this reason removing a field from the CRD should not cause issues for any users.

pkg/crds/operator/operator.tigera.io_amazoncloudintegrations.yaml
```diff
@@ -76,8 +76,6 @@ spec:
                 items:
                   type: string
                 type: array
-            required:
```
caseydavenport, line 79: Was this done manually? I don't see a corresponding CRD go code update

pkg/crds/enterprise/crd.projectcalico.org_alertexceptions.yaml
```diff
@@ -37,25 +37,30 @@ spec:
...
+                description: StartTime defines the start time from which this alert
+                  exception will take effect. If the value is in the past, matched
+                  alerts will be filtered immediately. If the value is changed to
+                  a future time, alert exceptions will restart at that time.
+                format: date-time
+                type: string
             required:
             - description
             - selector
+            - startTime
```
caseydavenport, line 60: Introducing a new required field is a bit scary - old resources that used to apply cleanly won't apply any more. Is this really what we want?

caseydavenport:
> From what I understand these AlertExceptions have never been a documented feature in a release version. For this reason removing a field from the CRD should not cause issues for any users.

@rene-dekker unfortunately documentation isn't the deciding factor in whether or not a feature exists, as much as we might like it to be. The API and CRDs are self-documenting, meaning if it's in a release then the feature exists.

pkg/crds/operator/operator.tigera.io_amazoncloudintegrations.yaml
hjiawei: No. Not manual.

pkg/crds/enterprise/crd.projectcalico.org_alertexceptions.yaml
hjiawei: We haven't officially released this yet (though in the release already) so there should be no old resources in the cluster.

pkg/crds/operator/operator.tigera.io_amazoncloudintegrations.yaml
caseydavenport: Ah, I see. sounds good.

PR #1954, opened 2022-05-11
Issue1055 Status Condition for Installation's CR from Tigerastatus conditions

vara2504:
## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/api/v1/installation_types.go
+++ b/api/v1/installation_types.go
@@ -605,6 +605,11 @@ type InstallationStatus struct {
 	// Computed is the final installation including overlaid resources.
 	// +optional
 	Computed *InstallationSpec `json:"computed,omitempty"`
+
+	// Conditions represents the latest observed set of conditions for the component. A component may be one or more of
+	// Ready, Progressing, Degraded or other customer types.
+	// +optional
+	Conditions []metav1.Condition `json:"conditions,omitempty"`
 }
 
 // +kubebuilder:object:root=true
--- a/api/v1/tigerastatus_types.go
+++ b/api/v1/tigerastatus_types.go
@@ -1,5 +1,5 @@
 /*
-
+Copyright (c) 2020-2022 Tigera, Inc. All rights reserved.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -74,6 +74,9 @@ const (
 
 	// Degraded means the component is not operating as desired and user action is required.
 	ComponentDegraded StatusConditionType = "Degraded"
+
+	// Ready indicates that the component is healthy and ready.it is identical to Available and used in Status conditions for CRs.
+	ComponentReady StatusConditionType = "Ready"
 )
 
 // TigeraStatusCondition represents a condition attached to a particular component.
@@ -93,6 +96,12 @@ type TigeraStatusCondition struct {
 
 	// Optionally, a detailed message providing additional context.
 	Message string `json:"message,omitempty"`
+
+	// observedGeneration represents the generation that the condition was set based upon.
+	// For instance, if generation is currently 12, but the .status.conditions[x].observedGeneration is 9, the condition is out of date
+	// with respect to the current state of the instance.
+	// +optional
+	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
 }
 
 // +kubebuilder:object:root=true
@@ -103,6 +112,29 @@ type TigeraStatusList struct {
 	metav1.ListMeta `json:"metadata,omitempty"`
 	Items           []TigeraStatus `json:"items"`
 }
+
+// TigeraStatusReason represents the reason for a particular condition. A reason may be one of the below.
+type TigeraStatusReason string
+
+const (
+	AllObjectsAvailable       TigeraStatusReason = "AllObjectsAvailable"
+	ResourceNotReady          TigeraStatusReason = "ResourceNotReady"
+	PodFailure                TigeraStatusReason = "PodFailure"
+	CertificateError          TigeraStatusReason = "CertificateError"
+	InvalidConfigurationError TigeraStatusReason = "InvalidConfigurationError"
+	ResourceCreateError       TigeraStatusReason = "ResourceCreateError"
+	ResourceMigrationError    TigeraStatusReason = "ResourceMigrationError"
+	ResourceNotFound          TigeraStatusReason = "ResourceNotFound"
+	ResourcePatchError        TigeraStatusReason = "ResourcePatchError"
+	ResourceReadError         TigeraStatusReason = "ResourceReadError"
+	ResourceScalingError      TigeraStatusReason = "ResourceScalingError"
+	ResourceUpdateError       TigeraStatusReason = "ResourceUpdateError"
+	ResourceValidationError   TigeraStatusReason = "ResourceValidationError"
+	MigrationError            TigeraStatusReason = "MigrationError"
+	InternalServerError       TigeraStatusReason = "InternalServerError"
+	NotApplicable             TigeraStatusReason = "NotApplicable"
+	Unknown                   TigeraStatusReason = "Unknown"
+)
 
 func init() {
 	SchemeBuilder.Register(&TigeraStatus{}, &TigeraStatusList{})
--- a/pkg/controller/installation/core_controller.go
+++ b/pkg/controller/installation/core_controller.go
@@ -89,6 +89,8 @@ const (
 	CalicoFinalizer         = "tigera.io/operator-cleanup"
 )
 
+const InstallationName string = "calico"
+
 //// Node and Installation finalizer
 // There is a problem with tearing down the calico resources where removing the calico-node ClusterRoleBinding
 // will block the kube-controller pod from teminating because the CNI plugin no longer has permissions.
@@ -194,6 +196,14 @@ func add(mgr manager.Manager, r *ReconcileInstallation) error {
 	err = c.Watch(&source.Kind{Type: &operator.Installation{}}, &handler.EnqueueRequestForObject{})
 	if err != nil {
 		return fmt.Errorf("tigera-installation-controller failed to watch primary resource: %w", err)
+	}
+
+	// Watch for changes to TigeraStatus.
+	err = c.Watch(&source.Kind{Type: &operator.TigeraStatus{ObjectMeta: metav1.ObjectMeta{Name: "calico"}}}, &handler.EnqueueRequestForObject{}, predicate.NewPredicateFuncs(func(object client.Object) bool {
+		return object.GetName() == InstallationName
+	}))
+	if err != nil {
+		return fmt.Errorf("tigera-installation-controller failed to watch calico Tigerastatus: %w", err)
 	}
 
 	if r.autoDetectedProvider == operator.ProviderOpenShift {
@@ -749,30 +759,47 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 		reqLogger.Error(err, "An error occurred when querying the Installation resource")
 		return reconcile.Result{}, err
 	}
-	status := instance.Status
+
 	terminating := (instance.DeletionTimestamp != nil)
 	preDefaultPatchFrom := client.MergeFrom(instance.DeepCopy())
 
 	// Mark CR found so we can report converter problems via tigerastatus
 	r.status.OnCRFound()
+	//Set the meta info in the tigerastatus like observedGenerations
+	defer r.status.SetMetaData(&instance.ObjectMeta)
 
+	// Changes for updating installation status conditions
+	if request.Name == InstallationName && request.Namespace == "" {
+		ts := &operator.TigeraStatus{}
+		err := r.client.Get(ctx, types.NamespacedName{Name: InstallationName}, ts)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+		instance.Status.Conditions = status.UpdateStatusCondition(instance.Status.Conditions, ts.Status.Conditions)
+		if err := r.client.Status().Update(ctx, instance); err != nil {
+			log.WithValues("reason", err).Info("Failed to create Installation status conditions.")
+			return reconcile.Result{}, err
+		}
+	}
+
+	status := instance.Status
 	if !r.migrationChecked {
 		// update Installation resource with existing install if it exists.
 		nc, err := convert.NeedsConversion(ctx, r.client)
 		if err != nil {
-			r.SetDegraded("Error checking for existing installation", err, reqLogger)
+			r.SetDegraded(operator.ResourceValidationError, "Error checking for existing installation", err, reqLogger)
 			return reconcile.Result{}, err
 		}
 		if nc {
 			install, err := convert.Convert(ctx, r.client)
 			if err != nil {
 				if errors.As(err, &convert.ErrIncompatibleCluster{}) {
-					r.SetDegraded("Existing Calico installation can not be managed by Tigera Operator as it is configured in a way that Operator does not currently support. Please update your existing Calico install config", err, reqLogger)
+					r.SetDegraded(operator.MigrationError, "Existing Calico installation can not be managed by Tigera Operator as it is configured in a way that Operator does not currently support. Please update your existing Calico install config", err, reqLogger)
 					// We should always requeue a convert problem. Don't return error
 					// to make sure we never back off retrying.
 					return reconcile.Result{RequeueAfter: 15 * time.Second}, nil
 				}
-				r.SetDegraded("Error converting existing installation", err, reqLogger)
+				r.SetDegraded(operator.MigrationError, "Error converting existing installation", err, reqLogger)
 				return reconcile.Result{}, err
 			}
 			instance.Spec = utils.OverrideInstallationSpec(install.Spec, instance.Spec)
@@ -781,7 +808,7 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 
 	// update Installation with defaults
 	if err := updateInstallationWithDefaults(ctx, r.client, instance, r.autoDetectedProvider); err != nil {
-		r.SetDegraded("Error querying installation", err, reqLogger)
+		r.SetDegraded(operator.ResourceReadError, "Error querying installation", err, reqLogger)
 		return reconcile.Result{}, err
 	}
 
@@ -789,7 +816,7 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 
 	// Validate the configuration.
 	if err := validateCustomResource(instance); err != nil {
-		r.SetDegraded("Invalid Installation provided", err, reqLogger)
+		r.SetDegraded(operator.InvalidConfigurationError, "Invalid Installation provided", err, reqLogger)
 		return reconcile.Result{}, err
 	}
 
@@ -801,7 +828,7 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 		crbKey := types.NamespacedName{Name: "calico-node"}
 		err := r.client.Get(ctx, crbKey, &crb)
 		if err != nil && !apierrors.IsNotFound(err) {
-			r.SetDegraded("Unable to get ClusterRoleBinding", err, reqLogger)
+			r.SetDegraded(operator.ResourceNotFound, "Unable to get ClusterRoleBinding", err, reqLogger)
 			return reconcile.Result{}, err
 		}
 		found := false
@@ -823,7 +850,7 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 	// Note that we only write the 'base' installation back. We don't want to write the changes from 'overlay', as those should only
 	// be stored in the 'overlay' resource.
 	if err := r.client.Patch(ctx, instance, preDefaultPatchFrom); err != nil {
-		r.SetDegraded("Failed to write defaults", err, reqLogger)
+		r.SetDegraded(operator.ResourceUpdateError, "Failed to write defaults", err, reqLogger)
 		return reconcile.Result{}, err
 	}
 
@@ -841,7 +868,7 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 
 		// Validate the configuration.
 		if err := validateCustomResource(instance); err != nil {
-			r.SetDegraded("Invalid computed config", err, reqLogger)
+			r.SetDegraded(operator.InvalidConfigurationError, "Invalid computed config", err, reqLogger)
 			return reconcile.Result{}, err
 		}
 	}
@@ -864,7 +891,7 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 	if reflect.DeepEqual(status, operator.InstallationStatus{}) {
 		instance.Status = operator.InstallationStatus{}
 		if err := r.client.Status().Update(ctx, instance); err != nil {
-			r.SetDegraded("Failed to write default status", err, reqLogger)
+			r.SetDegraded(operator.ResourceUpdateError, "Failed to write default status", err, reqLogger)
 			return reconcile.Result{}, err
 		}
 	}
@@ -873,7 +900,7 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 	// the run the reset the degraded status and requeue the request.
 	if r.typhaAutoscaler.isDegraded() {
 		if err := r.typhaAutoscaler.triggerRun(); err != nil {
-			r.SetDegraded("Failed to scale typha", err, reqLogger)
+			r.SetDegraded(operator.ResourceScalingError, "Failed to scale typha", err, reqLogger)
 			return reconcile.Result{RequeueAfter: 30 * time.Second}, nil
 		}
 	}
@@ -888,9 +915,9 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 			log.Info("Rebooting to enable TigeraSecure controllers")
 			os.Exit(0)
 		} else if err != nil {
-			r.SetDegraded("Error discovering Tigera Secure availability", err, reqLogger)
+			r.SetDegraded(operator.InternalServerError, "Error discovering Tigera Secure availability", err, reqLogger)
 		} else {
-			r.SetDegraded("Cannot deploy Tigera Secure", fmt.Errorf("Missing Tigera Secure custom resource definitions"), reqLogger)
+			r.SetDegraded(operator.InternalServerError, "Cannot deploy Tigera Secure", fmt.Errorf("Missing Tigera Secure custom resource definitions"), reqLogger)
 		}
 
 		// Queue a retry. We don't want to watch the APIServer API since it might not exist and would cause
@@ -904,7 +931,7 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 	if !r.amazonCRDExists {
 		amazonCRDRequired, err := utils.RequiresAmazonController(r.config)
 		if err != nil {
-			r.SetDegraded("Error discovering AmazonCloudIntegration CRD", err, reqLogger)
+			r.SetDegraded(operator.ResourceNotFound, "Error discovering AmazonCloudIntegration CRD", err, reqLogger)
 			reqLogger.Info("Scheduling a retry in 30 seconds")
 			return reconcile.Result{RequeueAfter: 30 * time.Second}, nil
 		}
@@ -917,7 +944,7 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 	// Query for pull secrets in operator namespace
 	pullSecrets, err := utils.GetNetworkingPullSecrets(&instance.Spec, r.client)
 	if err != nil {
-		r.SetDegraded("Error retrieving pull secrets", err, reqLogger)
+		r.SetDegraded(operator.ResourceReadError, "Error retrieving pull secrets", err, reqLogger)
 		return reconcile.Result{}, err
 	}
 
@@ -928,8 +955,7 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 		logCollector, err = utils.GetLogCollector(ctx, r.client)
 		if logCollector != nil {
 			if err != nil {
-				log.Error(err, "Error reading LogCollector")
-				r.status.SetDegraded("Error reading LogCollector", err.Error())
+				r.SetDegraded(operator.ResourceReadError, "Error reading LogCollector", err, reqLogger)
 				return reconcile.Result{}, err
 			}
 		}
@@ -937,31 +963,27 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 		managementCluster, err = utils.GetManagementCluster(ctx, r.client)
 		if managementCluster != nil {
 			if err != nil {
-				log.Error(err, "Error reading ManagementCluster")
-				r.status.SetDegraded("Error reading ManagementCluster", err.Error())
+				r.SetDegraded(operator.ResourceReadError, "Error reading ManagementCluster", err, reqLogger)
 				return reconcile.Result{}, err
 			}
 		}
 
 		managementClusterConnection, err = utils.GetManagementClusterConnection(ctx, r.client)
 		if err != nil {
-			log.Error(err, "Error reading ManagementClusterConnection")
-			r.status.SetDegraded("Error reading ManagementClusterConnection", err.Error())
+			r.SetDegraded(operator.ResourceReadError, "Error reading ManagementClusterConnection", err, reqLogger)
 			return reconcile.Result{}, err
 		}
 
 		if managementClusterConnection != nil && managementCluster != nil {
 			err = fmt.Errorf("having both a managementCluster and a managementClusterConnection is not supported")
-			log.Error(err, "")
-			r.status.SetDegraded(err.Error(), "")
+			r.SetDegraded(operator.ResourceValidationError, "", err, reqLogger)
 			return reconcile.Result{}, err
 		}
 	}
 
 	certificateManager, err := certificatemanager.Create(r.client, &instance.Spec, r.clusterDomain)
 	if err != nil {
-		log.Error(err, "unable to create the Tigera CA")
-		r.status.SetDegraded("Unable to create the Tigera CA", err.Error())
+		r.SetDegraded(operator.ResourceCreateError, "Unable to create the Tigera CA", err, reqLogger)
 		return reconcile.Result{}, err
 	}
 	var managerInternalTLSSecret certificatemanagement.KeyPairInterface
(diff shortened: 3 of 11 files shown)


api/v1/installation_types.go
```diff
@@ -660,3 +664,14 @@ type CertificateManagement struct {
...
 }
+
+// ConditionTypes describes the Installation CR's status condition type .
+//
+// One of: Ready, Degraded, Progressing
+type ConditionType string
+
+const (
+	ConditionTypeReady       ConditionType = "Ready"
+	ConditionTypeDegarded    ConditionType = "Degraded"
```
rene-dekker, line 675: There is a typo in ConditionTypeDegarded

api/v1/installation_types.go
```diff
@@ -604,6 +604,10 @@ type InstallationStatus struct {
 	// Computed is the final installation including overlaid resources.
 	// +optional
 	Computed *InstallationSpec `json:"computed,omitempty"`
+
+	// Conditions represents the latest observed set of conditions for the component. A component may be one or more of
+	// Ready, Progressing, Degraded or other customer types
```
rene-dekker, line 609: Please add the line `// +optional`

api/v1/installation_types.go
```diff
@@ -660,3 +664,13 @@ type CertificateManagement struct {
 	// +optional
 	SignatureAlgorithm string `json:"signatureAlgorithm,omitempty"`
 }
+
+// ConditionTypes describes Installation CR's status condition type.
+// One of: Ready, Degraded, Progressing.
+type ConditionType string
+
+const (
+	ConditionTypeReady       ConditionType = "Ready"
```
rene-dekker, line 673: Please add comments for each. You can copy them from tigerastatus_types.go.

Please note that each comment should start with the variable name. (This is currently not the case in tigerastatus_types.go and should be rectified).
Example:
```
// ComponentAvailable indicates that the component is healthy.
	ComponentAvailable StatusConditionType = "Available"
```

api/v1/tigerastatus_types.go
```diff
@@ -104,6 +104,27 @@ type TigeraStatusList struct {
 	Items           []TigeraStatus `json:"items"`
 }
 
+// TigeraStatusReason represents the reason for a particular condition. A reason may be one of the below.
+type TigeraStatusReason string
+
+const (
+	AllObjectsAvailable       TigeraStatusReason = "AllObjectsAvailable"
```
rene-dekker, line 111: nit: Only the first line needs to declare the type. If the first is done, the rest will automatically be assumed:

Example:
```
	ResourceNotReady  TigeraStatusReason = "ResourceNotReady"
	PodFailure = "PodFailure"
```

pkg/controller/installation/core_controller.go
```diff
@@ -194,6 +194,12 @@ func add(mgr manager.Manager, r *ReconcileInstallation) error {
 		return fmt.Errorf("tigera-installation-controller failed to watch primary resource: %w", err)
 	}
 
+	// Watch for changes to TigeraStatus.
+	err = c.Watch(&source.Kind{Type: &operator.TigeraStatus{ObjectMeta: metav1.ObjectMeta{Name: "calico"}}}, &handler.EnqueueRequestForObject{})
```
rene-dekker, line 198: Please extract a constant for "calico" and use it throughout this file.

pkg/controller/installation/core_controller.go
```diff
@@ -742,26 +748,41 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 	terminating := (instance.DeletionTimestamp != nil)
 	preDefaultPatchFrom := client.MergeFrom(instance.DeepCopy())
 
+	// Changes for updating installation status conditions
+	if request.Name == "calico" && request.Namespace == "" {
+		ts := &operator.TigeraStatus{ObjectMeta: metav1.ObjectMeta{Name: "calico"}}
+		err := r.client.Get(context.TODO(), types.NamespacedName{Name: "calico"}, ts)
```
rene-dekker, line 754: You can use `ctx` in stead of `context.TODO()`

pkg/controller/installation/core_controller.go
```diff
@@ -742,26 +748,41 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 	terminating := (instance.DeletionTimestamp != nil)
 	preDefaultPatchFrom := client.MergeFrom(instance.DeepCopy())
 
+	// Changes for updating installation status conditions
+	if request.Name == "calico" && request.Namespace == "" {
+		ts := &operator.TigeraStatus{ObjectMeta: metav1.ObjectMeta{Name: "calico"}}
```
rene-dekker, line 753: You can simplify this to `ts :=  &operator.TigeraStatus{ObjectMeta: metav1.ObjectMeta{Name: "calico"}}`

pkg/controller/installation/core_controller.go
```diff
@@ -742,26 +748,41 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
...
+
 	// Mark CR found so we can report converter problems via tigerastatus
 	r.status.OnCRFound()
 
 	if !r.migrationChecked {
 		// update Installation resource with existing install if it exists.
 		nc, err := convert.NeedsConversion(ctx, r.client)
 		if err != nil {
-			r.SetDegraded("Error checking for existing installation", err, reqLogger)
+			r.SetDegraded(operator.ResourceUpdateError, "Error checking for existing installation", err, reqLogger)
```
rene-dekker, line 773: NeedsConversion can throw an error for different reasons: NotFound, parsing, env missing. Therefore you should modify the func such that it also returns the Reason.

pkg/controller/installation/core_controller.go
```diff
@@ -742,26 +748,41 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
...
 			if err != nil {
 				if errors.As(err, &convert.ErrIncompatibleCluster{}) {
-					r.SetDegraded("Existing Calico installation can not be managed by Tigera Operator as it is configured in a way that Operator does not currently support. Please update your existing Calico install config", err, reqLogger)
+					r.SetDegraded(operator.ResourceValidationError, "Existing Calico installation can not be managed by Tigera Operator as it is configured in a way that Operator does not currently support. Please update your existing Calico install config", err, reqLogger)
 					// We should always requeue a convert problem. Don't return error
 					// to make sure we never back off retrying.
 					return reconcile.Result{RequeueAfter: 15 * time.Second}, nil
 				}
-				r.SetDegraded("Error converting existing installation", err, reqLogger)
+				r.SetDegraded(operator.ResourceUpdateError, "Error converting existing installation", err, reqLogger)
```
rene-dekker, line 785: This method does not update a resource, so I do not think we should use ResourceUpdateError here.

api/v1/tigerastatus_types.go
```diff
@@ -104,6 +104,27 @@ type TigeraStatusList struct {
...
+	ResourceNotReady          TigeraStatusReason = "ResourceNotReady"
+	PodFailure                TigeraStatusReason = "PodFailure"
+	CertificateError          TigeraStatusReason = "CertificateError"
+	InvalidConfigurationError TigeraStatusReason = "InvalidConfigurationError"
```
rene-dekker, line 115: I was wondering if we could consider merging InvalidConfigurationError and ResourceValidationError or if there are situations where this is not possible.

api/v1/tigerastatus_types.go
```diff
@@ -104,6 +104,27 @@ type TigeraStatusList struct {
...
+	ResourceCreateError       TigeraStatusReason = "ResourceCreateError"
+	ResourceMigrationError    TigeraStatusReason = "ResourceMigrationError"
+	ResourceNotFound          TigeraStatusReason = "ResourceNotFound"
+	ResourcePatchError        TigeraStatusReason = "ResourcePatchError"
+	ResourceReadError         TigeraStatusReason = "ResourceReadError"
+	ResourceScalingError      TigeraStatusReason = "ResourceScalingError"
```
rene-dekker, line 121: I think in essence scaling is just modifying an integer in a resource and update it. Therefore, I think we can remove this and use ResourcePatchError/ResourceUpdateError

pkg/controller/installation/core_controller.go
```diff
@@ -812,7 +833,7 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 	// Note that we only write the 'base' installation back. We don't want to write the changes from 'overlay', as those should only
 	// be stored in the 'overlay' resource.
 	if err := r.client.Patch(ctx, instance, preDefaultPatchFrom); err != nil {
-		r.SetDegraded("Failed to write defaults", err, reqLogger)
+		r.SetDegraded(operator.ResourceUpdateError, "Failed to write defaults", err, reqLogger)
```
rene-dekker, line 836: Should be ResourceUpdateError

pkg/controller/installation/core_controller.go
```diff
@@ -877,9 +898,9 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 			log.Info("Rebooting to enable TigeraSecure controllers")
 			os.Exit(0)
 		} else if err != nil {
-			r.SetDegraded("Error discovering Tigera Secure availability", err, reqLogger)
+			r.SetDegraded(operator.ResourceNotFound, "Error discovering Tigera Secure availability", err, reqLogger)
```
rene-dekker, line 901: This should not be a ResourceNotFound error.

I am not entirely sure what it should be. Maybe we can have an InternalServerError or something like that to catch unexpected errors, similar to HTTP 500?

rene-dekker: There is another similar "Error discovering Tigera Secure availability" as well below.

pkg/controller/installation/core_controller.go
```diff
@@ -1392,9 +1404,14 @@ func calicoDirectoryExists() bool {
 	return true
 }
 
-func (r *ReconcileInstallation) SetDegraded(reason string, err error, log logr.Logger) {
-	log.Error(err, reason)
-	r.status.SetDegraded(reason, err.Error())
+//func (r *ReconcileInstallation) SetDegraded(reason string, err error, log logr.Logger) {
```
rene-dekker, line 1407: You can remove the commented block.

pkg/controller/installation/core_controller.go
```diff
@@ -1392,9 +1404,14 @@ func calicoDirectoryExists() bool {
...
+//	log.Error(err, reason)
+//	r.status.SetDegraded(reason, err.Error())
+//}
+
+func (r *ReconcileInstallation) SetDegraded(reason operator.TigeraStatusReason, message string, err error, log logr.Logger) {
+	log.Error(err, string(reason)+message)
```
rene-dekker, line 1413: log.WithValues("reason", reason, "message", message).Error(err, reason)

pkg/controller/installation/core_controller.go
```diff
@@ -1392,9 +1404,14 @@ func calicoDirectoryExists() bool {
...
+	r.status.SetDegraded(string(reason), message+err.Error())
```
rene-dekker, line 1414: I imagine this does not look pretty. You probably want to use `fmt.Sprintf(...)` to format this nicely.

pkg/controller/status/status.go
```diff
@@ -402,8 +402,8 @@ func (m *statusManager) ClearDegraded() {
 	m.lock.Lock()
 	defer m.lock.Unlock()
 	m.degraded = false
-	m.explicitDegradedReason = ""
-	m.explicitDegradedMsg = ""
+	m.explicitDegradedReason = string(operator.NotApplicable)
```
rene-dekker, line 405: I think this file can have its own constants, and it is probably better not to rely on any variables from the "operator" import.

api/v1/installation_types.go
vara2504: Done

pkg/controller/installation/core_controller.go
vara2504: Done

pkg/controller/installation/core_controller.go
vara2504: Done

pkg/controller/installation/core_controller.go
vara2504: Done

pkg/controller/installation/core_controller.go
vara2504: Done

pkg/controller/installation/core_controller.go
vara2504: Done

pkg/controller/installation/core_controller.go
vara2504: Done

pkg/controller/status/status.go
vara2504: Done

pkg/controller/status/status.go
```diff
@@ -751,7 +759,7 @@ func (m *statusManager) clearDegraded() {
 	defer m.lock.Unlock()
 
 	conditions := []operator.TigeraStatusCondition{
-		{Type: operator.ComponentDegraded, Status: operator.ConditionFalse},
+		{Type: operator.ComponentDegraded, Status: operator.ConditionFalse, Reason: NotAvailable, Message: "Not Available"},
```
rene-dekker, line 762: clearDegraded() is called to reset the condition to the most basic state, which means that it is not degraded.
The reason should reflect that most basic state. NotAvailable means almost the opposite of that. I suppose we could add a parameter to the `SetDegraded(...)` method for the message and the reason and make sure that there is always a degraded message and reason whenever this method is called.
@tmjd  What do you think?

pkg/controller/status/status.go
```diff
@@ -39,6 +39,14 @@ import (
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 )
 
+const (
```
rene-dekker, line 42: I may have given conflicting comments now. Sorry if I did. But I think these could be merged with the tigerastatustypes.go.

pkg/controller/installation/core_controller.go
```diff
@@ -1783,3 +1797,57 @@ func removeInstallationFinalizer(i *operator.Installation) {
...
+			Type:               ctype,
+			Status:             status,
+			LastTransitionTime: condition.LastTransitionTime,
+			ObservedGeneration: instance.GetGeneration(),
+		}
+
+		if len(condition.Reason) > 0 {
+			ic.Reason = condition.Reason
+		} else {
+			ic.Reason = string(operator.NotAvailable)
```
rene-dekker, line 1829: I think we should use `Unknown` if we do not know the reason. NotAvailable might imply that the resource or deployment or whatever is not available.

pkg/controller/installation/core_controller.go
```diff
@@ -1783,3 +1797,57 @@ func removeInstallationFinalizer(i *operator.Installation) {
...
+		}
+		if len(condition.Message) > 0 {
+			ic.Message = condition.Message
+		} else {
+			ic.Message = "Not Available"
```
rene-dekker, line 1834: Not Available might imply that the resource or deployment or whatever is not available.

If I am not mistaken message is not a mandatory field. We could keep the message empty or otherwise we could use "Unknown"

pkg/controller/installation/core_controller_test.go
```diff
@@ -1274,4 +1274,375 @@ var _ = Describe("Testing core-controller installation", func() {
 			})
 		})
 	})
+	Context("Installation CR's Status conditions tests", func() {
```
rene-dekker, line 1277: You can move this describe inside of the context called "Reconcile tests". Then you can delete the BeforeEach and AfterEach from this one. That way we do not need all this code duplication.

pkg/controller/installation/core_controller.go
vara2504: Done

pkg/controller/installation/core_controller.go
vara2504: Message is a required filed - https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#Condition

vara2504:
Highlights of the changes in this PR:	
- Have used [metav1.condition](https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#Condition) for Installation CR status condition as per the [API convention](https://github.com/kubernetes-sigs/cli-utils/blob/master/pkg/kstatus/README.md#conditions).
- Tigerastatus condition type Available is mapped as "Ready" in installtion's condition
- Reasons will be one of the newly introduced TigeraStatusReason Enum.
- Modified SetDegraded func definition to accept TigeraStatusReason and appended the existing free form reasons in the code to message.
- Clearing Available,Progressing or degraded status will set the reason as operator.unknown and message as Unknow. Since both are required field in metav1.condition
- ObservedGeneration will be available in new Installation condition (edited) 
[
Detail Design Document ](https://docs.google.com/document/d/1wrQDvQIFGCElmrZ0XxOFuFLUXY-SoSeG/edit#)
https://docs.google.com/document/d/1wrQDvQIFGCElmrZ0XxOFuFLUXY-SoSeG/edit#

api/v1/installation_types.go
```diff
@@ -604,6 +604,11 @@ type InstallationStatus struct {
 	// Computed is the final installation including overlaid resources.
 	// +optional
 	Computed *InstallationSpec `json:"computed,omitempty"`
+
+	// Conditions represents the latest observed set of conditions for the component. A component may be one or more of
+	// Ready, Progressing, Degraded or other customer types
```
rene-dekker, line 609: nit: Please end with a "."

api/v1/installation_types.go
```diff
@@ -660,3 +665,18 @@ type CertificateManagement struct {
...
+
+// ConditionTypes describes Installation CR's status condition type.
+// One of: Ready, Degraded, Progressing.
+type ConditionType string
+
+const (
+	// ConditionTypeReady indicates that the component is healthy.
+	ConditionTypeReady ConditionType = "Ready"
+
+	// ConditionTypeDegraded indicated that the component is not operating as desired and user action is required..
```
rene-dekker, line 677: nit: There are two periods at the end of the sentence.

api/v1/installation_types.go
```diff
@@ -660,3 +665,18 @@ type CertificateManagement struct {
...
+	ConditionTypeDegraded ConditionType = "Degraded"
+
+	// ConditionTypeProgressing indicates that the component is in the process of being installed or upgraded
```
rene-dekker, line 680: Please end with a sentence.

All comments in this directory will be used in auto-generated docs on our website. They will also be part of the CRDs that users install. For example, see here: 
- https://docs.tigera.io/reference/installation/api
- https://projectcalico.docs.tigera.io/reference/installation/api

pkg/controller/installation/core_controller.go
```diff
@@ -738,30 +746,46 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 		reqLogger.Error(err, "An error occurred when querying the Installation resource")
 		return reconcile.Result{}, err
 	}
-	status := instance.Status
+
 	terminating := (instance.DeletionTimestamp != nil)
 	preDefaultPatchFrom := client.MergeFrom(instance.DeepCopy())
 
+	// Changes for updating installation status conditions
```
rene-dekker, line 753: I think that `r.status.OnCRFound()` should be before this block. That order makes more sense to me.

pkg/controller/status/status.go
```diff
@@ -162,13 +162,13 @@ func (m *statusManager) updateStatus() {
 		// We've collected knowledge about the current state of the objects we're monitoring.
 		// Now, use that to update the TigeraStatus object for this manager.
 		if m.IsAvailable() {
-			m.setAvailable("All objects available", "")
+			m.setAvailable(string(operator.AllObjectsAvailable), "All objects available")
```
rene-dekker, line 165: I would prefer if the setAvailable method only accepted TigeraStatusReasons:
```
func (m *statusManager) setAvailable(reason operator.TigeraStatusReason, msg string) {
```
Similar for the other methods. That way, you can never enter the wrong string in the reason field.

rene-dekker: And same for the StatusManager interface.

pkg/controller/status/status.go
```diff
@@ -879,3 +879,58 @@ func hasPendingCSRUsingCertV1beta1(ctx context.Context, cli client.Client, label
 	}
 	return false, nil
 }
+
+func UpdateStatusCondition(statuscondition []metav1.Condition, conditions []operator.TigeraStatusCondition, generation int64) []metav1.Condition {
```
rene-dekker, line 883: Please add a comment to explain how to use this function.

pkg/controller/status/status.go
```diff
@@ -402,8 +402,8 @@ func (m *statusManager) ClearDegraded() {
 	m.lock.Lock()
 	defer m.lock.Unlock()
 	m.degraded = false
-	m.explicitDegradedReason = ""
-	m.explicitDegradedMsg = ""
+	m.explicitDegradedReason = operator.Unknown
+	m.explicitDegradedMsg = "Unknown"
```
rene-dekker, line 406: If I am correct, the message is allowed to be empty. If this is correct, I think we can change all the messages for "Unknown" to "".

pkg/controller/status/status.go
```diff
@@ -879,3 +879,58 @@ func hasPendingCSRUsingCertV1beta1(ctx context.Context, cli client.Client, label
...
+			LastTransitionTime: condition.LastTransitionTime,
+			ObservedGeneration: generation,
+		}
+
+		if len(condition.Reason) > 0 {
+			ic.Reason = condition.Reason
+		} else {
+			ic.Reason = operator.Unknown
+		}
+		if len(condition.Message) > 0 {
```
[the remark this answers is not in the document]
rene-dekker: Even better, you can add the message directly in this block:
```
ic := metav1.Condition{
			Type:               ctype,
			Status:             status,
			LastTransitionTime: condition.LastTransitionTime,
			ObservedGeneration: generation,
		}

```

api/v1/installation_types.go
vara2504: Done

api/v1/installation_types.go
vara2504: Done

api/v1/installation_types.go
vara2504: Done

pkg/controller/status/status.go
vara2504: yes.. it can be empty. Updated to ""

pkg/controller/status/status.go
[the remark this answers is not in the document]
vara2504: Done

pkg/controller/status/status.go
vara2504: Sure Rene. i will make this changes once i finish this implementation for all the CR's
since
- Signature of SetAvailable and setProgressing can be changed as it is harded but for SetDegarded, the reason is passed across from all the CR controllers. 
- Few Unit test cases in others CRs are based on the current free form reason string which i am fixing along with the respective CR changes

api/v1/tigerastatus_types.go
vara2504: Marking as done

pkg/controller/status/status.go
rene-dekker: SGTM!

pkg/controller/status/status.go
vara2504: Done

rene-dekker (approved):
lgtm

pkg/controller/installation/core_controller.go
```diff
@@ -746,30 +754,46 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
...
 	r.status.OnCRFound()
 
+	// Changes for updating installation status conditions
+	if request.Name == InstallationName && request.Namespace == "" {
+		ts := &operator.TigeraStatus{}
+		err := r.client.Get(ctx, types.NamespacedName{Name: InstallationName}, ts)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+		instance.Status.Conditions = status.UpdateStatusCondition(instance.Status.Conditions, ts.Status.Conditions, instance.GetGeneration())
```
tmjd, line 771: It seems wrong to use the just read generation of the instance. There is nothing ensuring that the generation we just read matches with the generation when the Status was updated. As it is written the generation could very possibly be newer than generation that has been applied and TigeraStatus represents.
Has this been hashed out and the generation not being connected to the TigeraStatus update is ok?

I think it makes sense to pass the generation from the controller to `r.status` anytime we're doing a `r.status.SetDegraded` or when we call `r.status.ReadyToMonitor()`, then the update to the TigeraStatus should include a generation and then when the controller gets the TigeraStatus we use the generation embedded there.

tmjd (changes requested):
Sorry it took me so long to review this. Let's chat if you have any concerns with any of my comments.

pkg/controller/installation/core_controller.go
```diff
@@ -746,30 +754,46 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
...
 			if err != nil {
 				if errors.As(err, &convert.ErrIncompatibleCluster{}) {
-					r.SetDegraded("Existing Calico installation can not be managed by Tigera Operator as it is configured in a way that Operator does not currently support. Please update your existing Calico install config", err, reqLogger)
+					r.SetDegraded(operator.ResourceValidationError, "Existing Calico installation can not be managed by Tigera Operator as it is configured in a way that Operator does not currently support. Please update your existing Calico install config", err, reqLogger)
 					// We should always requeue a convert problem. Don't return error
 					// to make sure we never back off retrying.
 					return reconcile.Result{RequeueAfter: 15 * time.Second}, nil
 				}
-				r.SetDegraded("Error converting existing installation", err, reqLogger)
+				r.SetDegraded(operator.ResourceValidationError, "Error converting existing installation", err, reqLogger)
```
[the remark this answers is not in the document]
vara2504: Done

pkg/controller/installation/core_controller.go
```diff
@@ -925,86 +949,81 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
...
+			r.SetDegraded(operator.ResourceValidationError, "Error in BGP layout ConfigMap", err, reqLogger)
 			return reconcile.Result{}, err
 		}
 	}
 
 	err = utils.GetK8sServiceEndPoint(r.client)
 	if err != nil {
 		log.Error(err, "Error reading services endpoint configmap")
-		r.SetDegraded("Error reading services endpoint configmap", err, reqLogger)
+		r.SetDegraded(operator.ResourceReadError, "Error reading services endpoint configmap", err, reqLogger)
```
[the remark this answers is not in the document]
vara2504: Done

pkg/controller/installation/core_controller.go
```diff
@@ -1013,7 +1032,7 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 		openShiftOnAws, err = isOpenshiftOnAws(instance, ctx, r.client)
 		if err != nil {
 			log.Error(err, "Error checking if OpenShift is on AWS")
-			r.SetDegraded("Error checking if OpenShift is on AWS", err, reqLogger)
+			r.SetDegraded(operator.ResourceValidationError, "Error checking if OpenShift is on AWS", err, reqLogger)
```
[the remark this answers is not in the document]
vara2504: Done

pkg/controller/installation/core_controller.go
```diff
@@ -1035,8 +1053,7 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 		if apierrors.IsNotFound(err) {
 			aci = nil
 		} else if err != nil {
-			log.Error(err, "Error reading AmazonCloudIntegration")
-			r.status.SetDegraded("Error reading AmazonCloudIntegration", err.Error())
+			r.SetDegraded(operator.ResourceValidationError, "Error reading AmazonCloudIntegration", err, reqLogger)
```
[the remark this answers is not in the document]
vara2504: Done

pkg/controller/installation/core_controller.go
```diff
@@ -1024,8 +1043,7 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 	// to mimic a 'normal' rolling update.
 	needNsMigration, err := r.namespaceMigration.NeedsCoreNamespaceMigration(ctx)
 	if err != nil {
-		log.Error(err, "Error checking if namespace migration is needed")
-		r.status.SetDegraded("Error checking if namespace migration is needed", err.Error())
+		r.SetDegraded(operator.ResourceValidationError, "Error checking if namespace migration is needed", err, reqLogger)
```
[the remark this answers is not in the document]
vara2504: Done

pkg/controller/installation/core_controller.go
```diff
@@ -1045,7 +1062,7 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 	felixConfiguration := &crdv1.FelixConfiguration{}
 	err = r.client.Get(ctx, types.NamespacedName{Name: "default"}, felixConfiguration)
 	if err != nil && !apierrors.IsNotFound(err) {
-		r.SetDegraded("Unable to read FelixConfiguration", err, reqLogger)
+		r.SetDegraded(operator.ResourceNotFound, "Unable to read FelixConfiguration", err, reqLogger)
```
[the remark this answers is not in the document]
vara2504: Done

pkg/controller/installation/core_controller.go
```diff
@@ -1068,23 +1085,21 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
...
-			r.SetDegraded("invalid metrics port", err, reqLogger)
+			r.SetDegraded(operator.InvalidConfigurationError, "invalid metrics port", err, reqLogger)
 			return reconcile.Result{}, err
 		}
 
 		nodePrometheusTLS, err = certificateManager.GetOrCreateKeyPair(r.client, render.NodePrometheusTLSServerSecret, common.OperatorNamespace(), dns.GetServiceDNSNames(render.CalicoNodeMetricsService, common.CalicoNamespace, r.clusterDomain))
 		if err != nil {
-			log.Error(err, "Error creating TLS certificate")
-			r.status.SetDegraded("Error creating TLS certificate", err.Error())
+			r.SetDegraded(operator.CertificateError, "Error creating TLS certificate", err, reqLogger)
```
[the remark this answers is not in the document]
vara2504: Done

pkg/controller/installation/core_controller.go
```diff
@@ -1096,7 +1111,7 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 	kubeControllersConfig := &crdv1.KubeControllersConfiguration{}
 	err = r.client.Get(ctx, types.NamespacedName{Name: "default"}, kubeControllersConfig)
 	if err != nil && !apierrors.IsNotFound(err) {
-		r.SetDegraded("Unable to read KubeControllersConfiguration", err, reqLogger)
+		r.SetDegraded(operator.ResourceNotFound, "Unable to read KubeControllersConfiguration", err, reqLogger)
```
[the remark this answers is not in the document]
vara2504: Done

pkg/controller/installation/core_controller.go
```diff
@@ -1194,7 +1209,7 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 	bgpConfiguration := &crdv1.BGPConfiguration{}
 	err = r.client.Get(ctx, types.NamespacedName{Name: "default"}, bgpConfiguration)
 	if err != nil && !apierrors.IsNotFound(err) {
-		r.SetDegraded("Unable to read BGPConfiguration", err, reqLogger)
+		r.SetDegraded(operator.ResourceNotFound, "Unable to read BGPConfiguration", err, reqLogger)
```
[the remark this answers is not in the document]
vara2504: Done

pkg/controller/installation/core_controller.go
```diff
@@ -744,30 +752,46 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
...
-	status := instance.Status
+
 	terminating := (instance.DeletionTimestamp != nil)
 	preDefaultPatchFrom := client.MergeFrom(instance.DeepCopy())
 
 	// Mark CR found so we can report converter problems via tigerastatus
 	r.status.OnCRFound()
 
+	// Changes for updating installation status conditions
+	if request.Name == InstallationName && request.Namespace == "" {
```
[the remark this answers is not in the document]
rene-dekker: If there are non-namespaced resources with the name "calico" this could introduce a bug. The only one I can think of would be if a user created a managed cluster called "calico". Would that be a deal breaker? Would it weigh up to the benefits of ending the reconciliation loop pre-emptively?

Since the reconcile.Request is created by one of our dependency, I do not think we can get richer information about what event triggered the reconcile loop.

That said, if we want to roll out these status conditions to other controllers, we must check per controller if the CR name clashes with any other resources. As far as I am aware Vara is going through this exercise  or is still in the process of going through this exercise.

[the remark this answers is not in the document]
rene-dekker: Maybe we can use another system to end the reconcile() function pre-emptively by relying on the `TigeraStatus.metadata.resourceVersion`. If we keep track either in memory or in the installation status if the TigeraStatus is new, in addition to checking the request name and request namespace, we can conclude that we can end the reconcile. If not, we continue the reconcile loop.

I considered proposing `TigeraStatus.metadata.generation`, but this value is always set to "1". It never increments, because we only update the `TigeraStatus.status`, not TigeraStatus itself.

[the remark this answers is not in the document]
rene-dekker: > I really don't like the possibility of a managed cluster named "calico" being a bug.

I agree.

> Would it be a problem if we did a full reconcile anytime TigeraStatus was updated?

Works for me.

pkg/controller/installation/core_controller.go
vara2504: Done.. invoking SetMetaData in defer

pkg/controller/installation/core_controller.go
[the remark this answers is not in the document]
vara2504: Done..

pkg/controller/installation/core_controller.go
```diff
@@ -196,6 +198,14 @@ func add(mgr manager.Manager, r *ReconcileInstallation) error {
 		return fmt.Errorf("tigera-installation-controller failed to watch primary resource: %w", err)
 	}
 
+	// Watch for changes to TigeraStatus.
+	err = c.Watch(&source.Kind{Type: &operator.TigeraStatus{ObjectMeta: metav1.ObjectMeta{Name: "calico"}}}, &handler.EnqueueRequestForObject{}, predicate.NewPredicateFuncs(func(object client.Object) bool {
```
rene-dekker, line 202: For the subsequent controllers it would be nice to extract a func for this where you pass in the name. For this PR I'm ok keeping it like this.

rene-dekker (approved):
lgtm

pkg/controller/installation/core_controller.go
vara2504: Sure.. Will incorporate as part of the next PR while implementing the changes for all CRs

tmjd:
Fixes #1055

PR #1957, opened 2022-05-12
Add mount-cgroup init container to node ds

mazdakn:
## Description
This PR covers operator changes for fixing GH issue: https://github.com/projectcalico/calico/issues/5852
The main part of fix is here: https://github.com/projectcalico/calico/pull/6078

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/node.go
+++ b/pkg/render/node.go
@@ -837,7 +837,7 @@ func (c *nodeComponent) nodeVolumes() []corev1.Volume {
 			// Volume for the bpffs itself, used by the main node container.
 			corev1.Volume{Name: "bpffs", VolumeSource: corev1.VolumeSource{HostPath: &corev1.HostPathVolumeSource{Path: "/sys/fs/bpf", Type: &dirMustExist}}},
 			// Volume used by mount-cgroupv2 init container to access root cgroup name space of node.
-			corev1.Volume{Name: "proc", VolumeSource: corev1.VolumeSource{HostPath: &corev1.HostPathVolumeSource{Path: "/proc"}}},
+			corev1.Volume{Name: "init-proc", VolumeSource: corev1.VolumeSource{HostPath: &corev1.HostPathVolumeSource{Path: "/proc/1"}}},
 		)
 	}
 
@@ -981,8 +981,9 @@ func (c *nodeComponent) bpffsInitContainer() corev1.Container {
 			MountPropagation: &bidirectional,
 		},
 		{
-			MountPath: "/node-proc",
-			Name:      "proc",
+			MountPath: "/initproc",
+			Name:      "init-proc",
+			ReadOnly:  true,
 		},
 	}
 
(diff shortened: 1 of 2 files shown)


fasaxc (approved):
LGTM but I think you need sign-off from  @tigera/tigera-operator-maintainers

pkg/render/node.go
```diff
@@ -836,6 +836,8 @@ func (c *nodeComponent) nodeVolumes() []corev1.Volume {
 			corev1.Volume{Name: "sys-fs", VolumeSource: corev1.VolumeSource{HostPath: &corev1.HostPathVolumeSource{Path: "/sys/fs", Type: &dirOrCreate}}},
 			// Volume for the bpffs itself, used by the main node container.
 			corev1.Volume{Name: "bpffs", VolumeSource: corev1.VolumeSource{HostPath: &corev1.HostPathVolumeSource{Path: "/sys/fs/bpf", Type: &dirMustExist}}},
+			// Volume used by mount-cgroupv2 init container to access root cgroup name space of node.
+			corev1.Volume{Name: "proc", VolumeSource: corev1.VolumeSource{HostPath: &corev1.HostPathVolumeSource{Path: "/proc"}}},
```
tmjd, line 840: It looks like the code only uses `/node-proc/1/ns`, could this volume be reduced in scope to `/proc/1/ns`?

pkg/render/node.go
```diff
@@ -978,6 +980,10 @@ func (c *nodeComponent) bpffsInitContainer() corev1.Container {
 			// so that it outlives the init container.
 			MountPropagation: &bidirectional,
 		},
+		{
+			MountPath: "/node-proc",
```
tmjd, line 984: I'm guessing the mount command we do with nsenter needs write access to the volume. Even though I think the answer is no, could this VolumeMount be done ReadOnly?

pkg/render/node.go
fasaxc: Worth a shot!  `/proc/1/ns` is still "very privileged" but might as well reduce our scope as much as possible. I don't think PID 1 can go away or the system panics so it may be the only process where that is workable.

pkg/render/node.go
mazdakn: I'll try it.

pkg/render/node.go
mazdakn: Fixed it.

mazdakn: @tmjd now we need to also access `/proc/1/mountinfo`, so I limited the mount to `/proc/1`. How does it sound?

pkg/render/node.go
mazdakn: Setting it to `ReadOnly` works fine. Done!

tmjd (approved):
LGTM

PR #1969, opened 2022-05-20
Put create and update into function so it can be retried

tmjd:
## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/utils/component.go
+++ b/pkg/controller/utils/component.go
@@ -88,12 +88,7 @@ func (c componentHandler) CreateOrUpdateOrDelete(ctx context.Context, component 
 	objsToCreate, objsToDelete := component.Objects()
 	osType := component.SupportedOSType()
 
-	for _, obj := range objsToCreate {
-		om, ok := obj.(metav1.ObjectMetaAccessor)
-		if !ok {
-			return fmt.Errorf("Object is not ObjectMetaAccessor")
-		}
-
+	createOrUpdateObject = func(obj client.Object) error {
 		// Add owner ref for controller owned resources,
 		switch obj.(type) {
 		case *v3.UISettings:
@@ -116,18 +111,6 @@ func (c componentHandler) CreateOrUpdateOrDelete(ctx context.Context, component 
 		// Make sure any objects with images also have an image pull policy.
 		modifyPodSpec(obj, setImagePullPolicy)
 
-		// Keep track of some objects so we can report on their status.
-		switch obj.(type) {
-		case *apps.Deployment:
-			deployments = append(deployments, key)
-		case *apps.DaemonSet:
-			daemonSets = append(daemonSets, key)
-		case *apps.StatefulSet:
-			statefulsets = append(statefulsets, key)
-		case *batchv1beta.CronJob:
-			cronJobs = append(cronJobs, key)
-		}
-
 		cur, ok := obj.DeepCopyObject().(client.Object)
 		if !ok {
 			logCtx.V(2).Info("Failed converting object", "obj", obj)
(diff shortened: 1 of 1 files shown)


caseydavenport:
@tmjd what's the motivation for this one? Could use a PR description

pkg/controller/utils/component.go
```diff
@@ -88,19 +87,14 @@ func (c componentHandler) CreateOrUpdateOrDelete(ctx context.Context, component
 	objsToCreate, objsToDelete := component.Objects()
 	osType := component.SupportedOSType()
 
-	for _, obj := range objsToCreate {
-		om, ok := obj.(metav1.ObjectMetaAccessor)
-		if !ok {
-			return fmt.Errorf("Object is not ObjectMetaAccessor")
-		}
-
+	createOrUpdateObject := func(obj client.Object, metaObj metav1.Object) error {
```
Brian-McM, line 90: Maybe this is just personal preference, but I don't really like have anonymous functions used like this. This makes the function more difficult to read because:

1. Variable scoping gets a little more confusing to read
2. You have a massive chunk of code that you basically need to skip over when reading the function because it's not called in line
3. Adds a lot of code that could just as easily be taken out to make the rest of the logic easier to follow (this is possibly the same point as 2)

pkg/controller/utils/component_test.go
```diff
@@ -1174,3 +1294,91 @@ func (c *fakeComponent) Objects() ([]client.Object, []client.Object) {
...
+	v, ok := mc.Info[mc.Index].Return.(error)
+	if !ok {
+		panic(fmt.Sprintf("mockClient Info didn't have right type for entry %d for %s %v", mc.Index, funcName, key))
+	}
+
+	return v
+}
+
+func (mc *mockClient) List(ctx context.Context, list client.ObjectList, opts ...client.ListOption) error {
+	panic("List not implemented in mockClient")
```
rene-dekker, line 1332: The changes in this file have unreachable return statements and unused function params.

Brian-McM: These are likely here to meet the interface requirements without having to fully implement the interface, i.e. he only needs `Get` and `Update` for his tests so `Get` and `Update` are the only ones implemented.

rene-dekker: You need either panic, or return, not both.

Brian-McM: That's true but the function parameters are required to meet the interface definition.

rene-dekker: The types of the params are, not the names.

rene-dekker (approved):
lgtm

pkg/controller/utils/component.go
```diff
@@ -66,14 +66,19 @@ type componentHandler struct {
 	log    logr.Logger
 }
 
-func (c componentHandler) createOrUpdateObject(ctx context.Context, obj client.Object, metaObj metav1.Object, osType rmeta.OSType) error {
+func (c componentHandler) createOrUpdateObject(ctx context.Context, obj client.Object, osType rmeta.OSType) error {
+	om, ok := obj.(metav1.ObjectMetaAccessor)
+	if !ok {
+		return fmt.Errorf("Object is not ObjectMetaAccessor")
```
rene-dekker, line 72: Do we perhaps need to log/learn more about this object? If this can happen it looks indicative of a bug. Won't this error keep repeating on any reconcile?

tmjd: Yeah it would repeat. I copied this from before I just moved it. I don't imagine this is actually needed. It might be something to make the linter happy, to check that casting was successful.

PR #1970, opened 2022-05-21
allow-tigera reconciliation

pasanw:
## Description
This PR brings reconciliation of the allow-tigera tier into the operator, in line with the design spec discussed internally.

### Reviewer notes
* Please review commit-by-commit. In retrospect, I would have executed on this with multiple PRs in mind instead of one. I've tried to split the work by commit to introduce the approach gradually, but I'm happy to adjust/split the PR in whatever way makes the review/release process easiest.
* The PR aims to mirror the existing policies as closely as possible. In some cases, there are slight changes that make policies more uniform across components. In order to make these changes easily reviewable, I've created a [dummy PR here](https://github.com/pasanw/operator/pull/1).
* There will be some self-review comments to highlight some areas of the code that I'd like to discuss with you

### Implementation notes
* Policies, wherever possible, are rendered in the component they ensure access to.
* Policies are reconciled using the V3 API. This means that controllers must wait for the V3 API to be available, in addition to waiting for the allow-tigera Tier to be created.
* A Tiers controller is established, which creates the allow-tigera Tier resource, and houses core and apiserver policies to decouple their policies from the creation of the API server.
* Render packages export EntityRules that standardize how components are referenced for egress and ingress rules.
* Tests for render packages verify that policies correctly adapt for cluster and provider type. Expected policies are stored as JSON files.
* In addition to new render and controller tests, manual validation of rendered policy was performed for a standalone K8S cluster

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/main.go
+++ b/main.go
@@ -23,6 +23,10 @@ import (
 	goruntime "runtime"
 	"strings"
 
+	"github.com/tigera/operator/pkg/render/common/networkpolicy"
+	"k8s.io/apimachinery/pkg/labels"
+	"sigs.k8s.io/controller-runtime/pkg/cache"
+
 	"github.com/cloudflare/cfssl/log"
 	"github.com/ghodss/yaml"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -177,6 +181,12 @@ func main() {
 		os.Exit(1)
 	}
 
+	policySelector, err := labels.Parse(fmt.Sprintf("projectcalico.org/tier == %s", networkpolicy.TigeraComponentTierName))
+	if err != nil {
+		log.Error(err, "")
+		os.Exit(1)
+	}
+
 	// Because we only run this as a job that is set up by the operator, it should not be
 	// launched except by an operator that is the active operator. So we do not need to
 	// check that we're the active operator before running the AWS SG setup.
@@ -211,6 +221,18 @@ func main() {
 		ClientDisableCacheFor: []client.Object{
 			&v3.LicenseKey{},
 		},
+		// NetworkPolicy is served through the Tigera API Server, which currently restricts List and Watch
+		// operations on NetworkPolicy to a single tier only, specified via label or field selector. If no
+		// selector is specified, List and Watch return policies from the 'default' tier. The manager cache
+		// must therefore apply a selector to specify the tier that the operator currently reconciles policy
+		// within so that it can receive the expected resources for List and Watch. If the operator needs to
+		// reconcile policy within multiple tiers, the API Server should be updated to serve policy from all
+		// tiers that the user is authorized for.
+		NewCache: cache.BuilderWithOptions(cache.Options{
+			SelectorsByObject: cache.SelectorsByObject{
+				&v3.NetworkPolicy{}: {Label: policySelector},
+			},
+		}),
 	})
 	if err != nil {
 		setupLog.Error(err, "unable to start manager")
--- a/pkg/controller/authentication/authentication_controller.go
+++ b/pkg/controller/authentication/authentication_controller.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2020-2021 Tigera, Inc. All rights reserved.
+// Copyright (c) 2020-2022 Tigera, Inc. All rights reserved.
 
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -18,6 +18,12 @@ import (
 	"context"
 	"fmt"
 	"time"
+
+	"k8s.io/client-go/kubernetes"
+
+	"github.com/tigera/operator/pkg/render/common/networkpolicy"
+
+	v3 "github.com/tigera/api/pkg/apis/projectcalico/v3"
 
 	"github.com/go-ldap/ldap"
 
@@ -61,30 +67,46 @@ func Add(mgr manager.Manager, opts options.AddOptions) error {
 		// No need to start this controller.
 		return nil
 	}
-	return add(mgr, newReconciler(mgr, opts))
+
+	// Create the reconciler
+	reconciler := newReconciler(mgr, opts)
+
+	// Create a new controller
+	c, err := controller.New(controllerName, mgr, controller.Options{Reconciler: reconcile.Reconciler(reconciler)})
+	if err != nil {
+		return fmt.Errorf("failed to create %s: %w", controllerName, err)
+	}
+
+	k8sClient, err := kubernetes.NewForConfig(mgr.GetConfig())
+	if err != nil {
+		return fmt.Errorf("%s failed to establish a connection to k8s: %w", controllerName, err)
+	}
+
+	go utils.WaitToAddNetworkPolicyWatches(c, k8sClient, log, reconciler.policyWatchesReady, []types.NamespacedName{
+		{Name: render.DexPolicyName, Namespace: render.DexNamespace},
+		{Name: networkpolicy.TigeraComponentDefaultDenyPolicyName, Namespace: render.DexNamespace},
+	})
+
+	return add(mgr, c)
 }
 
 // newReconciler returns a new reconcile.Reconciler
 func newReconciler(mgr manager.Manager, opts options.AddOptions) *ReconcileAuthentication {
 	r := &ReconcileAuthentication{
-		client:        mgr.GetClient(),
-		scheme:        mgr.GetScheme(),
-		provider:      opts.DetectedProvider,
-		status:        status.New(mgr.GetClient(), "authentication", opts.KubernetesVersion),
-		clusterDomain: opts.ClusterDomain,
+		client:             mgr.GetClient(),
+		scheme:             mgr.GetScheme(),
+		provider:           opts.DetectedProvider,
+		status:             status.New(mgr.GetClient(), "authentication", opts.KubernetesVersion),
+		clusterDomain:      opts.ClusterDomain,
+		policyWatchesReady: &utils.ReadyFlag{},
 	}
 	r.status.Run(opts.ShutdownContext)
 	return r
 }
 
-// add adds a new Controller to mgr with r as the reconcile.Reconciler
-func add(mgr manager.Manager, r *ReconcileAuthentication) error {
-	c, err := controller.New(controllerName, mgr, controller.Options{Reconciler: r})
-	if err != nil {
-		return fmt.Errorf("failed to create %s: %w", controllerName, err)
-	}
-
-	err = c.Watch(&source.Kind{Type: &oprv1.Authentication{}}, &handler.EnqueueRequestForObject{})
+// add adds watches for resources that are available at startup
+func add(mgr manager.Manager, c controller.Controller) error {
+	err := c.Watch(&source.Kind{Type: &oprv1.Authentication{}}, &handler.EnqueueRequestForObject{})
 	if err != nil {
 		return fmt.Errorf("%s failed to watch resource: %w", controllerName, err)
 	}
@@ -121,11 +143,12 @@ var _ reconcile.Reconciler = &ReconcileAuthentication{}
 
 // ReconcileAuthentication reconciles an Authentication object
 type ReconcileAuthentication struct {
-	client        client.Client
-	scheme        *runtime.Scheme
-	provider      oprv1.Provider
-	status        status.StatusManager
-	clusterDomain string
+	client             client.Client
+	scheme             *runtime.Scheme
+	provider           oprv1.Provider
+	status             status.StatusManager
+	clusterDomain      string
+	policyWatchesReady *utils.ReadyFlag
 }
 
 // Reconciles the cluster state with the Authentication object that is found in the cluster.
@@ -183,6 +206,17 @@ func (r *ReconcileAuthentication) Reconcile(ctx context.Context, request reconci
 		return reconcile.Result{}, nil
 	}
 
+	// Ensure the API Server is ready, before rendering any objects that utilize the V3 API.
+	if !utils.IsAPIServerReady(r.client, reqLogger) {
+		r.status.SetDegraded("Waiting for Tigera API server to be ready", "")
+		return reconcile.Result{}, nil
+	}
+
+	if !r.policyWatchesReady.IsReady() {
+		r.status.SetDegraded("Waiting for NetworkPolicy watches to be established", "")
+		return reconcile.Result{RequeueAfter: 10 * time.Second}, nil
+	}
+
 	// Make sure the tigera-dex namespace exists, before rendering any objects there.
 	if err := r.client.Get(ctx, client.ObjectKey{Name: render.DexObjectName}, &corev1.Namespace{}); err != nil {
 		if errors.IsNotFound(err) {
@@ -192,6 +226,19 @@ func (r *ReconcileAuthentication) Reconcile(ctx context.Context, request reconci
 		} else {
 			log.Error(err, "Error querying tigera-dex namespace")
 			r.status.SetDegraded("Error querying tigera-dex namespace", err.Error())
+			return reconcile.Result{}, err
+		}
+	}
+
+	// Ensure the allow-tigera tier exists, before rendering any network policies within it.
+	if err := r.client.Get(ctx, client.ObjectKey{Name: networkpolicy.TigeraComponentTierName}, &v3.Tier{}); err != nil {
+		if errors.IsNotFound(err) {
+			log.Error(err, "Waiting for Tigera component policy tier to be created")
+			r.status.SetDegraded("Waiting for Tigera component policy tier to be created", err.Error())
+			return reconcile.Result{RequeueAfter: 10 * time.Second}, nil
+		} else {
+			log.Error(err, "Error querying Tigera component policy tier")
+			r.status.SetDegraded("Error querying Tigera component policy tier", err.Error())
 			return reconcile.Result{}, err
 		}
 	}
--- a/pkg/controller/clusterconnection/clusterconnection_controller.go
+++ b/pkg/controller/clusterconnection/clusterconnection_controller.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2020-2021 Tigera, Inc. All rights reserved.
+// Copyright (c) 2020-2022 Tigera, Inc. All rights reserved.
 
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -18,6 +18,12 @@ import (
 	"context"
 	"fmt"
 	"time"
+
+	"k8s.io/client-go/kubernetes"
+
+	"github.com/tigera/operator/pkg/render/common/networkpolicy"
+
+	v3 "github.com/tigera/api/pkg/apis/projectcalico/v3"
 
 	operatorv1 "github.com/tigera/operator/api/v1"
 	"github.com/tigera/operator/pkg/common"
@@ -53,33 +59,49 @@ func Add(mgr manager.Manager, opts options.AddOptions) error {
 		// No need to start this controller.
 		return nil
 	}
+
 	statusManager := status.New(mgr.GetClient(), "management-cluster-connection", opts.KubernetesVersion)
-	return add(mgr, newReconciler(mgr.GetClient(), mgr.GetScheme(), statusManager, opts.DetectedProvider, opts))
+	policyWatchesReady := &utils.ReadyFlag{}
+
+	// Create the reconciler
+	reconciler := newReconciler(mgr.GetClient(), mgr.GetScheme(), statusManager, policyWatchesReady, opts.DetectedProvider, opts)
+
+	// Create a new controller
+	controller, err := controller.New(controllerName, mgr, controller.Options{Reconciler: reconciler})
+	if err != nil {
+		return fmt.Errorf("failed to create %s: %w", controllerName, err)
+	}
+
+	k8sClient, err := kubernetes.NewForConfig(mgr.GetConfig())
+	if err != nil {
+		return fmt.Errorf("%s failed to establish a connection to k8s: %w", controllerName, err)
+	}
+	go utils.WaitToAddNetworkPolicyWatches(controller, k8sClient, log, policyWatchesReady, []types.NamespacedName{
+		{Name: render.GuardianEgressPolicyName, Namespace: render.GuardianNamespace},
+		{Name: networkpolicy.TigeraComponentDefaultDenyPolicyName, Namespace: render.GuardianNamespace},
+	})
+
+	return add(mgr, controller)
 }
 
 // newReconciler returns a new reconcile.Reconciler
-func newReconciler(cli client.Client, schema *runtime.Scheme, statusMgr status.StatusManager, p operatorv1.Provider, opts options.AddOptions) reconcile.Reconciler {
+func newReconciler(cli client.Client, schema *runtime.Scheme, statusMgr status.StatusManager, policyWatchesReady *utils.ReadyFlag, p operatorv1.Provider, opts options.AddOptions) *ReconcileConnection {
 	c := &ReconcileConnection{
-		Client:        cli,
-		Scheme:        schema,
-		Provider:      p,
-		status:        statusMgr,
-		clusterDomain: opts.ClusterDomain,
+		Client:             cli,
+		Scheme:             schema,
+		Provider:           p,
+		status:             statusMgr,
+		clusterDomain:      opts.ClusterDomain,
+		policyWatchesReady: policyWatchesReady,
 	}
 	c.status.Run(opts.ShutdownContext)
 	return c
 }
 
 // add adds a new controller to mgr with r as the reconcile.Reconciler
-func add(mgr manager.Manager, r reconcile.Reconciler) error {
-	// Create a new controller
-	c, err := controller.New(controllerName, mgr, controller.Options{Reconciler: r})
-	if err != nil {
-		return fmt.Errorf("failed to create %s: %w", controllerName, err)
-	}
-
+func add(mgr manager.Manager, c controller.Controller) error {
 	// Watch for changes to primary resource ManagementCluster
-	err = c.Watch(&source.Kind{Type: &operatorv1.ManagementCluster{}}, &handler.EnqueueRequestForObject{})
+	err := c.Watch(&source.Kind{Type: &operatorv1.ManagementCluster{}}, &handler.EnqueueRequestForObject{})
 	if err != nil {
 		return fmt.Errorf("%s failed to watch primary resource: %w", controllerName, err)
 	}
@@ -124,11 +146,12 @@ var _ reconcile.Reconciler = &ReconcileConnection{}
 
 // ReconcileConnection reconciles a ManagementClusterConnection object
 type ReconcileConnection struct {
-	Client        client.Client
-	Scheme        *runtime.Scheme
-	Provider      operatorv1.Provider
-	status        status.StatusManager
-	clusterDomain string
+	Client             client.Client
+	Scheme             *runtime.Scheme
+	Provider           operatorv1.Provider
+	status             status.StatusManager
+	clusterDomain      string
+	policyWatchesReady *utils.ReadyFlag
 }
 
 // Reconcile reads that state of the cluster for a ManagementClusterConnection object and makes changes based on the
@@ -143,6 +166,30 @@ func (r *ReconcileConnection) Reconcile(ctx context.Context, request reconcile.R
 	variant, instl, err := utils.GetInstallation(ctx, r.Client)
 	if err != nil {
 		return result, err
+	}
+
+	// Ensure the API Server is ready, before rendering any objects that utilize the V3 API.
+	if !utils.IsAPIServerReady(r.Client, reqLogger) {
+		r.status.SetDegraded("Waiting for Tigera API server to be ready", "")
+		return reconcile.Result{}, nil
+	}
+
+	if !r.policyWatchesReady.IsReady() {
+		r.status.SetDegraded("Waiting for NetworkPolicy watches to be established", "")
+		return reconcile.Result{RequeueAfter: 10 * time.Second}, nil
+	}
+
+	// Ensure the allow-tigera tier exists, before rendering any network policies within it.
+	if err := r.Client.Get(ctx, client.ObjectKey{Name: networkpolicy.TigeraComponentTierName}, &v3.Tier{}); err != nil {
+		if k8serrors.IsNotFound(err) {
+			log.Error(err, "Waiting for Tigera component policy tier to be created")
+			r.status.SetDegraded("Waiting for Tigera component policy tier to be created", err.Error())
+			return reconcile.Result{RequeueAfter: 10 * time.Second}, nil
+		} else {
+			log.Error(err, "Error querying Tigera component policy tier")
+			r.status.SetDegraded("Error querying Tigera component policy tier", err.Error())
+			return reconcile.Result{}, err
+		}
 	}
 
 	managementCluster, err := utils.GetManagementCluster(ctx, r.Client)
--- a/pkg/controller/monitor/monitor_controller.go
+++ b/pkg/controller/monitor/monitor_controller.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2021 Tigera, Inc. All rights reserved.
+// Copyright (c) 2021-2022 Tigera, Inc. All rights reserved.
 
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -20,6 +20,10 @@ import (
 	"fmt"
 	"reflect"
 	"time"
+
+	"github.com/tigera/operator/pkg/render/common/networkpolicy"
+
+	v3 "github.com/tigera/api/pkg/apis/projectcalico/v3"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
@@ -60,10 +64,11 @@ func Add(mgr manager.Manager, opts options.AddOptions) error {
 		return nil
 	}
 
-	var prometheusReady = &utils.ReadyFlag{}
+	prometheusReady := &utils.ReadyFlag{}
+	policyWatchesReady := &utils.ReadyFlag{}
 
 	// Create the reconciler
-	reconciler := newReconciler(mgr, opts, prometheusReady)
+	reconciler := newReconciler(mgr, opts, prometheusReady, policyWatchesReady)
 
 	// Create a new controller
 	controller, err := controller.New("monitor-controller", mgr, controller.Options{Reconciler: reconciler})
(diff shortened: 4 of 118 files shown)


pkg/render/tiers/tiers.go
```diff
@@ -0,0 +1,333 @@
...
+	render.GuardianNamespace,
+	render.ComplianceNamespace,
+	render.DexNamespace,
+	render.ElasticsearchNamespace,
+	render.LogCollectorNamespace,
+	render.IntrusionDetectionNamespace,
+	render.KibanaNamespace,
+	render.ManagerNamespace,
+	common.TigeraPrometheusNamespace,
+	"tigera-skraper",
```
pasanw, line 49: How should we handle this cloud-specific namespace? I adapted this policy from the cloud install as requested, but wasn't sure we should handle this dangling namespace

pkg/render/tiers/tiers.go
```diff
@@ -0,0 +1,333 @@
...
+func (t tiersComponent) Objects() ([]client.Object, []client.Object) {
+	objsToCreate := []client.Object{
+		t.allowTigeraTier(),
+		t.allowTigeraClusterDNSPolicy(),
+		t.allowTigeraAPIServerPolicy(),
+		t.allowTigeraKubeControllersPolicy(),
+		t.allowTigeraPacketCapturePolicy(),
+	}
+
+	// Delete equivalent policies under different namespaced names previously managed outside the operator.
```
pasanw, line 84: Not sure the best approach here, this probably needs to be changed.

There are three policies (kube-dns, kube-dns-egress, guardian-calico-cloud-egress) currently created during the cloud install process, and I've now brought them into the operator under different names.

We should determine how we'd like these existing policies to be handled on upgrade. The options I see are:

1. The operator is responsible for deleting them, and explicitly names them here
2. The updated cloud install script is responsible for deleting them
3. Do nothing in either place

pkg/render/tiers/tiers.go
```diff
@@ -0,0 +1,333 @@
...
+// Reconciled here rather than the core controller since the creation of v3 NetworkPolicy depends on the API Server, and
+// the core controller is a prerequisite of the API server controller.
+func (t *tiersComponent) allowTigeraKubeControllersPolicy() *v3.NetworkPolicy {
+	egressRules := []v3.Rule{}
+	egressRules = networkpolicy.AppendDNSEgressRules(egressRules, t.cfg.Openshift)
+	egressRules = append(egressRules, []v3.Rule{
+		{
+			Action:   v3.Allow,
+			Protocol: &networkpolicy.TCPProtocol,
+			Destination: v3.EntityRule{
```
pasanw, line 228: This doesn't use the 'shared' API Server entity rule since it doesn't specify a namespace selector like all other policies do.

Is there a reason why this is the case, or can I align this with the others?

pkg/render/common/networkpolicy/networkpolicy.go
```diff
@@ -0,0 +1,201 @@
...
+		Selector:          fmt.Sprintf("k8s-app == '%s'", deploymentName),
+		Ports:             Ports(port),
+	}
+}
+
+// CreateSourceEntityRule creates a conventional entity rule that matches ingress traffic based on namespace and deployment name.
+func CreateSourceEntityRule(namespace string, deploymentName string) v3.EntityRule {
+	return v3.EntityRule{
+		Selector:          fmt.Sprintf("k8s-app == '%s'", deploymentName),
+		NamespaceSelector: fmt.Sprintf("name == '%s'", namespace),
```
pasanw, line 86: All of the existing ingress rules seem to use use `name` instead of `projectcalico.org/name`. Should this be changed?

pasanw (commented):
Leaving some comments for discussion

pkg/render/tiers/tiers.go
pasanw: In general, I am a bit confused by the original CC policy, but that could be due to a lack of product/NetworkPolicy knowledge

Why would a managed cluster select for management namespaces? My instinct is that the list of namespaces here should adapt to the cluster type, and that the skraper namespace shouldn't be here at all

pkg/render/tiers/tiers.go
```diff
@@ -0,0 +1,333 @@
...
+					Source: v3.EntityRule{
+						NamespaceSelector: "all()",
+						Selector:          DNSIngressNamespaceSelector,
+					},
+				},
+				{
+					Action: v3.Pass,
+				},
+			},
+			Egress: []v3.Rule{
```
pasanw, line 150: The original CC policies had the egress split out separately so that users could restrict it if required. Since we are moving to a model where users do not edit the allow-tigera tier, I've combined them into one policy. The user can create policy in adjacent tiers to override

pkg/controller/tiers/tiers_controller.go
```diff
@@ -0,0 +1,158 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+// The Tiers controller reconciles system tiers and network policies using the v3 API. v3 resources require the API server to be available.
+// System tiers are reconciled here instead of the installation controller since the API server is not available until after installation completes.
+// Similarly, policies for installation and API server controllers are reconciled here to ensure the API server is available when they are reconciled.
```
pasanw: During further testing, I realized there are more dependency ordering considerations when it comes to managed clusters, since the license key needs to be pushed from the management cluster.

I've added a commit that addresses this. You can check [this comment ](https://github.com/tigera/operator/pull/1970/commits/e070915949aea39d8a358da6ea69b4c53a104461#diff-43340c178ffd2e84ca3fc589d654e19ed933e66852fca37c609c063cc207ea35R46)for an explanation/preview of the change

caseydavenport (commented):
Still working my way through this but thought I'd put what I have so far. Looking good!

Haven't reviewed the policies themselves / tests in detail yet.

pkg/render/common/networkpolicy/networkpolicy.go
pasanw: Is there a reason for our choice of "name" vs "projectcalico/name" to differ depending on whether the rule is ingress or egress? The pattern of the existing manifest appears to be "name" for ingress, and "projectcalico.org/name" for egress (hence the two different EntityRule-generating functions)

main.go
```diff
@@ -211,6 +221,18 @@ func main() {
...
 			&v3.LicenseKey{},
 		},
+		// NetworkPolicy is served through the Tigera API Server, which currently restricts List and Watch
+		// operations on NetworkPolicy to a single tier only, specified via label or field selector. If no
+		// selector is specified, List and Watch return policies from the 'default' tier. The manager cache
+		// must therefore apply a selector to specify the tier that the operator currently reconciles policy
+		// within so that it can receive the expected resources for List and Watch. If the operator needs to
+		// reconcile policy within multiple tiers, the API Server should be updated to serve policy from all
+		// tiers that the user is authorized for.
+		NewCache: cache.BuilderWithOptions(cache.Options{
```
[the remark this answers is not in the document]
pasanw: I believe we have to specify it on the cache for this kind, else disable the cache for this kind

Within ComponentHandler.CreateOrUpdateOrDelete, Client.Get is called to determine whether we need to update the object in question. If the cache is not disabled for the GVK, then the Client returned by the controller runtime uses its cache  to retrieve the object, which is where we run into the issue: the cache is filled using an inital List and updated via Watch. It's those List and Watch requests that we need to specify the label selector on for NetworkPolicy, else the cache is only going to have NetworkPolicy from the default tier.

From what I could see, this appeared to be the cleanest approach to apply the label selector to that List and Watch used for the cache, but I'm open to suggestions! I might be missing something.

pkg/controller/tiers/tiers_controller.go
```diff
@@ -0,0 +1,158 @@
...
+	k8sClient, err := kubernetes.NewForConfig(mgr.GetConfig())
+	if err != nil {
+		log.Error(err, "Failed to establish a connection to k8s")
+		return err
+	}
+
+	go utils.WaitToAddTierWatch(networkpolicy.TigeraComponentTierName, c, k8sClient, log, tierWatchReady)
+
+	policyNames := []types.NamespacedName{
+		{Name: tiers.APIServerPolicyName, Namespace: rmeta.APIServerNamespace(operatorv1.TigeraSecureEnterprise)},
```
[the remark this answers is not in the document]
pasanw: I didn't consider that option but I like it, I think it strikes a better balance than the current implementation. If I understand correctly, it would look something like this:
- Impacted* controllers will check if the tier is available**, and pass this result to their render component(s) as a boolean flag
- The render components will only render policy (i.e. return policy when Objects() is called) if the flag is set to true

Now evaluating how this plays out during an installation:
1. Impacted controllers will not reconcile any NetworkPolicy, and eventually become Available
2. The Tiers controller will create the allow-tigera tier once the APIServer and LicenseKey are available
3. Impacted controllers will need to be 'kicked' via watch to reconcile again, and at this time, their network policies will be included in the objects to reconcile since the tier is available.

My main uncertainty is how the impacted controllers are kicked? They can't establish watches on the Tier, since that would require the APIServer/LicenseKey to be available. What would you suggest?

\* Impacted controller being a controller that should reconcile NetworkPolicy, but blocks the creation of Tier by being a requirement for the APIServer or LicenseKey to be available
\*\* We would need to treat failure to query the tier availability as 'no'

pkg/controller/tiers/tiers_controller.go
```diff
@@ -0,0 +1,158 @@
...
+		{Name: tiers.KubeControllerPolicyName, Namespace: common.CalicoNamespace},
+		{Name: tiers.PacketCapturePolicyName, Namespace: render.PacketCaptureNamespace},
+	}
+	if opts.DetectedProvider == operatorv1.ProviderOpenShift {
```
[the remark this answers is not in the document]
pasanw: Not that I'm aware of -  I'm guessing you prefer the simplicity of watching both?

pkg/render/apiserver.go
```diff
@@ -679,7 +679,7 @@ func (c *apiServerComponent) apiServerService() *corev1.Service {
 					Name:       "apiserver",
 					Port:       443,
 					Protocol:   corev1.ProtocolTCP,
-					TargetPort: intstr.FromInt(apiServerPort),
+					TargetPort: intstr.FromInt(ApiServerPort),
```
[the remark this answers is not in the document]
pasanw: It gets referenced in the Tiers render component since policy for API server is currently there

I did some after-the-fact splitting of commits to make review easier, so if you are reviewing commit-by-commit perhaps you encountered an imperfection of that split, apologies if so

pkg/render/common/networkpolicy/networkpolicy.go
```diff
@@ -0,0 +1,201 @@
...
+	}
+}
+
+// AppendServiceSelectorDNSEgressRules is equivalent to AppendDNSEgressRules, utilizing service selector instead of label selector and ports.
+func AppendServiceSelectorDNSEgressRules(egressRules []v3.Rule, openShift bool) []v3.Rule {
```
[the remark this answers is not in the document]
pasanw: Was more so being cautious and leaving any potential migrations to service selectors to the review process as we discussed. Shall I mark this as a case that we can safely migrate?

pkg/render/common/networkpolicy/networkpolicy.go
```diff
@@ -0,0 +1,201 @@
...
+		Name:      "kubernetes",
+	},
+}
+
+// The entity rules below are extracted from render subpackages to prevent cyclic dependencies.
+var EsGatewayEntityRule = CreateEntityRule("tigera-elasticsearch", "tigera-secure-es-gateway", 5554)
+var EsGatewaySourceEntityRule = CreateSourceEntityRule("tigera-elasticsearch", "tigera-secure-es-gateway")
+var EsGatewayServiceSelectorEntityRule = CreateServiceSelectorEntityRule("tigera-elasticsearch", "tigera-secure-es-gateway-http")
+
+const PrometheusSelector = "(app == 'prometheus' && prometheus == 'calico-node-prometheus') || (app.kubernetes.io/name == 'prometheus' && prometheus == 'calico-node-prometheus')"
```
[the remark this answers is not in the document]
pasanw: Just to clarify, are you referring to all selectors in this file, or all policy selectors in general?

pkg/controller/tiers/tiers_controller.go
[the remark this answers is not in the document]
caseydavenport: I think what you have is probably fine... technically someone might change the provider via the API and this logic wouldn't catch that - that's what I was originally thinking, but changing provider from "not openshift" to "openshift" is probably going to break everything anyway...

pkg/controller/tiers/tiers_controller.go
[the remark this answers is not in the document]
caseydavenport: > My main uncertainty is how the impacted controllers are kicked? They can't establish watches on the Tier, since that would require the APIServer/LicenseKey to be available. What would you suggest?

Couldn't they use a deferred watch in the same way as other controllers? The only diference would be that instead of returning if the watch isn't ready, the controllers would just not install the policy if the watch wasn't ready.

pkg/render/common/networkpolicy/networkpolicy.go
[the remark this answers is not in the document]
caseydavenport: I was referring specifically to the ones in this file, but really all would be better! Selectors aren't always obvious so a human readable description is useful, especially for the more obtuse ones like this one here.

pkg/render/tiers/tiers.go
caseydavenport: My preference would be (2) since it's less cruft in the operator accounting for resources the operator never owned itself. but not sure how feasible that is. CC @tmjd

pkg/render/common/networkpolicy/networkpolicy.go
caseydavenport: I don't think that ingress vs egress makes a difference. This is probably just us being inconsistent. 

Being more consistent would be good - in general, it's better to use "real" labels than the fake labels we do behind the scenes because the real labels are viewable in the API.

pkg/render/intrusiondetection/dpi/dpi.go
```diff
@@ -331,3 +339,45 @@ func (d *dpiComponent) dpiAnnotations() map[string]string {
 	annotations[d.cfg.TyphaNodeTLS.NodeSecret.HashAnnotationKey()] = d.cfg.TyphaNodeTLS.NodeSecret.HashAnnotationValue()
 	return annotations
 }
+
+// This policy uses service selectors.
```
caseydavenport, line 343: Would like to see more detailed comments than this, especially at the function level. The fact that it uses service selectors should be obvious from reading the code, what's maybe less obvious is the purpose of this policy.

e.g., 

```
// dpiAllowTigeraPolicy creates a network policy which selects component X and allows the following traffic:
//
// ingress:
// - ...
//
// egress:
// - ...
//
// This policy is installed in Calico Enterprise management and managed clusters
```

Or something like that ^ 

This level of detail is probably something we should be writing in for each of these policies. Otherwise they start to just become "magic" and become difficult to maintain or modify over time.

pkg/render/fluentd.go
```diff
@@ -1046,3 +1060,60 @@ func (c *fluentdComponent) eksLogForwarderClusterRole() *rbacv1.ClusterRole {
...
+
+func (c *fluentdComponent) allowTigeraPolicy() *v3.NetworkPolicy {
+	egressRules := []v3.Rule{}
+	if c.cfg.ManagedCluster {
+		egressRules = append(egressRules, v3.Rule{
+			Action:   v3.Deny,
+			Protocol: &networkpolicy.TCPProtocol,
+			Source:   v3.EntityRule{},
+			Destination: v3.EntityRule{
+				NamespaceSelector: fmt.Sprintf("projectcalico.org/name == '%s'", GuardianNamespace),
```
caseydavenport, line 1072: We should add to your list of cleanups to standardize on how we select namespaces. 

We can do a follow-on PR that implements those cleanups to avoid putting too much onto this PR - probably better to do something more close to a 1:1 translation and then a follow-on for cleanups.

pkg/render/dex.go
```diff
@@ -343,3 +351,87 @@ func (c *dexComponent) configMap() *corev1.ConfigMap {
...
+					Source:      networkpolicy.PrometheusSourceEntityRule,
+					Destination: dexIngressPortDestination,
+				},
+			},
+			Egress: egressRules,
+		},
+	}
+}
+
+func allowTigeraDefaultDeny() *v3.NetworkPolicy {
```
caseydavenport, line 435: Same here - not much value in a wrapper around a single function call?

pkg/render/dex.go
```diff
@@ -43,8 +46,11 @@ const (
 	DexPort          = 5556
 	DexTLSSecretName = "tigera-dex-tls"
 	DexClientId      = "tigera-manager"
+	DexPolicyName    = networkpolicy.TigeraComponentPolicyPrefix + "allow-tigera-dex"
 )
 
+var DexEntityRule = networkpolicy.CreateEntityRule(DexNamespace, DexObjectName, DexPort)
```
caseydavenport, line 52: Do we need to declare this globally? Seems like its only used in one place?

pasanw: It's referenced by multiple render components - should be referenced by the tiers component in the first commit (and by more components in subsequent commits)

pkg/controller/tiers/tiers_controller.go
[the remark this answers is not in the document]
pasanw: Of course🤦. I think the implementation will be much better off this way, so I'd like to have this refactor as part of the current PR. 

Let me know if/when you are done with your initial review. I'll perform this refactor along with the other requested changes

pkg/render/fluentd.go
pasanw: Sounds good. That's exactly what I had in mind: target 1:1 translation to validate correctness of implementation and inform the design, then handle improvements to existing policy.

pkg/render/intrusiondetection/dpi/dpi.go
pasanw: Agreed that this specific comment is not very useful.

After some thought I'm inclined to agree with your suggestion of having explicit comments to serve as documentation. I hesitate because I would have liked the code to self-document, and don't want comments to drift from implementation. But I personally can't see a way to improve the nature of how the code documents itself, and I agree that the way it documents itself is not sufficient due to how decentralized the self-documentation is

main.go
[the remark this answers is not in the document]
pasanw: @Brian-McM when you and I last spoke, I was leaning towards that as well. However, as I was chatting with Matt on timelines for delivering this feature, we came to the conclusion that it was more pragmatic to make the small change in the operator now for the current requirement rather than rework tier RBAC in the API server for a future requirement (multiple tiers in the operator). I updated the design doc at that time to make sure it was included in the design review

I'm curious what both you and @caseydavenport think is best? I tried my best to summarize the context above and in the code comment, but also happy to hop on a call if its easier that way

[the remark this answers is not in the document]
Brian-McM: Maybe this is outside of the scope of what we should be commenting in a PR (and outside of what I should say as I'm not really in this design anymore and many decisions have been made without me) but it didn't seem like that API change was going to take an extraordinarily large amount of time. I had thought that more less exactly what needed to be done was figured out with you, Rob, and I.

Personally I think that if it's not going to be done right now for this feature, it should be done shortly after (as I think this is the correct thing to do). Given this, I'd say do whatever is quickest and easiest to implement (and undo) here, as it will just be reverted later when the proper fix is in for the API server.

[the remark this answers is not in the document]
caseydavenport: @pasanw thanks for that detailed write up. I think I see the issue now, and I'm leaning towards agreeing with you that specifying the selector on the cache is the most pragmatic solution to this for now. 

The downside is that we won't be able to do any Calico policy watches / management outside of the allow-tigera tier, which I believe will be overly limiting in the near future, so a proper fix is likely in order shortly after this.

I don't know what that fix would be - it might be server-side, or perhaps we can re-jig the way the controller infrastructure works somehow to solve this client-side. So basically, I also agree with @Brian-McM that we should schedule the "right" fix for this very soon after this.

pkg/render/tiers/tiers.go
pasanw: @caseydavenport Looking at this again, I'm wondering if we should rethink inheriting the egress rule into the operator... seems a bit too broad of scope for the operator/allow-tigera. WDYT?

pkg/controller/utils/utils.go
```diff
@@ -218,6 +218,41 @@ func IsFeatureActive(license v3.LicenseKey, featureName string) bool {
...
+	}
+
+	// If the policies to be reconciled require any license features, validate that the license is available.
+	if len(licenseFeatureNames) > 0 {
+		if _, err := FetchLicenseKey(ctx, cli); err != nil {
+			return false
+		}
+	}
+
+	return true
```
pasanw, line 253: As an aside, this function is where I was pondering adding a check to validate the API server availability, like so:

```
apiService := &apiregv1.APIService{}
cli.Get(ctx, client.ObjectKey{Name: "v3.projectcalico.org"}, apiService)
// ... return false if apiService.Status.Conditions does not have Available = ConditionTrue
```

If this _was_ comprehensive enough to validate API server availability here, then apiserver/core could fully trust this guard function (like monitor/guardian currently do since they do not impact apiserver). That could in theory remove the need for rendering policy last in those controllers

pasanw (commented):
@caseydavenport I've addressed the comments and performed some cleanups, I believe we are ready for re-review. I've re-validated against an MCM setup and all seems good.

I would recommend [first reviewing this range of commits](https://github.com/tigera/operator/pull/1970/files/e070915949aea39d8a358da6ea69b4c53a104461..fdeecd2b748badfc2722d414ab79ae83b3672a14), which addresses the comments made, including the main rework around core/apiserver/monitor/guardian controllers. I would advise against reviewing commit-by-commit in this range as some fixes/tweaks are split across commits.

Afterwards, the [last few commits](https://github.com/tigera/operator/pull/1970/files/fdeecd2b748badfc2722d414ab79ae83b3672a14..9a5f15f862317e3016278195710cca075a841a19) bring the implementation in-line with master elsewhere in calico/calient, namely policy and CRD changes.

Here is the current list of follow-up PRs:
* Add detailed comments on policies and label selectors
* Cleanup and normalize policies (incl. standardizing namespace selectors, service selectors)
* Schedule/investigate: "right" fix for reconciling policies across multiple tiers

pkg/controller/monitor/monitor_controller.go
```diff
@@ -265,6 +294,64 @@ func (r *ReconcileMonitor) Reconcile(ctx context.Context, request reconcile.Requ
...
+
+			if !r.policyWatchesReady.IsReady() {
+				r.status.SetDegraded("Waiting for NetworkPolicy watches to be established", "")
+				return reconcile.Result{RequeueAfter: 10 * time.Second}, nil
+			}
+		} else {
+			includeV3NetworkPolicy = false
+		}
+	} else {
+		// Wait for and render NetworkPolicy as usual.
```
caseydavenport, line 331: I think we could restructure this branching to avoid needing to write out this section twice. Couldn't we do something like this?


```go
// No matter what, we wait for watches to be established.
if ! r.tierWatchReady() {
  return
}

if !policyWatchReady() {
  return
}

// Determine if we should include policy.
includePolicy = true
if monitorReconcilesV3NetworkPolicy && !utils.IsReconcilable {
  includePolicy = false
}

if includePolicy {
  // Check that the allow-tigera tier exists. 
}
```

pkg/controller/utils/utils.go
caseydavenport: I don't think it's really worth trying to verify the API server is available. We'll find that out when we try to write the resources to the API and they either fail or succeed! 

We should just make sure that policies are sent to the API server _after_ the API server deployment / serviceaccount / etc.

pkg/controller/monitor/monitor_controller.go
```diff
@@ -265,6 +294,64 @@ func (r *ReconcileMonitor) Reconcile(ctx context.Context, request reconcile.Requ
...
+	if err != nil {
+		log.Error(err, "Error reading ManagementClusterConnection")
+		r.status.SetDegraded("Error reading ManagementClusterConnection", err.Error())
+		return reconcile.Result{}, err
+	}
+
+	// In managed clusters with certificate management disabled, successful reconciliation of non-NetworkPolicy resources
+	// in the monitor controller ensures that NetworkPolicy is reconcilable (by enabling* the creation of containing Tier
+	// and License). Therefore, to prevent a chicken-and-egg scenario, we only reconcile non-NetworkPolicy resources once
+	// we can confirm that all requirements to reconcile NetworkPolicy have been met.
```
caseydavenport, line 307: >  we only reconcile non-NetworkPolicy resources once we can confirm that all requirements to reconcile NetworkPolicy have been met.

Do you mean we only reconcile NetworkPolicy once all the requirements have been met? We should reconcile non-NP resources unconditionally.

If so, isn't it enough to simply put the NetworkPolicy component at the end, conditional on whether or not the v3 NetworkPolicy API and allow-tigera tier is present in the cluster?

pkg/controller/monitor/monitor_controller.go
```diff
@@ -265,6 +294,64 @@ func (r *ReconcileMonitor) Reconcile(ctx context.Context, request reconcile.Requ
...
+			return reconcile.Result{RequeueAfter: 10 * time.Second}, nil
+		}
+
+		if !r.policyWatchesReady.IsReady() {
+			r.status.SetDegraded("Waiting for NetworkPolicy watches to be established", "")
+			return reconcile.Result{RequeueAfter: 10 * time.Second}, nil
+		}
+
+		// Ensure the allow-tigera tier exists, before rendering any network policies within it.
+		if err := r.client.Get(ctx, client.ObjectKey{Name: networkpolicy.TigeraComponentTierName}, &v3.Tier{}); err != nil {
```
caseydavenport, line 343: I find it a little confusing that in some places, we query the Tier directly to determine if it exists, and in some cases we essentially do the same by calling `utils.IsV3NetworkPolicyReconcilable(ctx, r.client, networkpolicy.TigeraComponentTierName)`

The obvious difference seems to be that the former has more detailed error handling, whereas the latter has no error handling (errors querying the tier are effectively ignored, and assumed to be 404 errors). I suspect we do actually need to handle errors when calling `IsV3NetworkPolicyReconcilable` though.

pkg/controller/tiers/tiers_controller.go
```diff
@@ -0,0 +1,165 @@
...
+		r.status.SetDegraded("Waiting for Tier watch to be established", "")
+		return reconcile.Result{RequeueAfter: 10 * time.Second}, nil
+	}
+
+	if !r.policyWatchesReady.IsReady() {
+		r.status.SetDegraded("Waiting for NetworkPolicy watches to be established", "")
+		return reconcile.Result{RequeueAfter: 10 * time.Second}, nil
+	}
+
+	// Ensure that a license is present so this controller can establish/manage tiers and domain-based policy.
```
caseydavenport, line 139: Normally we use "domain-based policy" to refer to policies that use domain names in their rules, so this would be more clear if it said "tiers and policy to ensure cluster DNS is not broken"

pkg/controller/utils/utils.go
```diff
@@ -217,6 +218,41 @@ func IsFeatureActive(license v3.LicenseKey, featureName string) bool {
...
+func IsV3NetworkPolicyReconcilable(ctx context.Context, cli client.Client, tierName string, licenseFeatureNames ...string) bool {
+	// Validate tier presence by querying for the tier. Note that this does not validate Tigera API server availability
+	// as resources are cached.
+	if err := cli.Get(ctx, client.ObjectKey{Name: tierName}, &v3.Tier{}); err != nil {
+		return false
+	}
+
+	// If the policies to be reconciled require any license features, validate that the license is available.
+	if len(licenseFeatureNames) > 0 {
+		if _, err := FetchLicenseKey(ctx, cli); err != nil {
```
caseydavenport, line 248: I think this check is redundant - we can't ever create the allow-tigera tier if there is no license present, so the check above will fail in all of the cases where this check would fail.

I suppose it handles the case where a license is created, a tier is created, and a license is deleted? But that will show up with our catch-all error handling anyway, since the API server will return an error to us and we'll emit that error.

It seems like you're lining this up to check for specific license features in a future iteration though, so perhaps a call to https://github.com/tigera/operator/blob/ed1a85cb571e8f5cce4b599022ad378f5fb51fb5/pkg/controller/utils/utils.go#L209-L218 would make this function more true to its description?

pkg/controller/monitor/monitor_controller.go
pasanw: As discussed, tier and policy watches need not block controllers from reconciling if they are not ready. We'll rely on the tier being available as a strong enough signal that we can proceed.

pkg/render/tiers/tiers.go
pasanw: Ah, sorry I think this was a misunderstanding of policy application on my part. I didn't consider the fact that if we restrict the allowed egress in this tier, that (even with the Pass) all other egress will be denied unless explicitly allowed in a later tier. 

Do let me know if I understand that correctly.

Will resolve in current PR (by ensuring the current rule encapsulates all known NS and removing the skraper namespace)

pkg/controller/authentication/authentication_controller.go
```diff
@@ -183,6 +206,17 @@ func (r *ReconcileAuthentication) Reconcile(ctx context.Context, request reconci
 		return reconcile.Result{}, nil
 	}
 
+	// Ensure the API Server is ready, before rendering any objects that utilize the V3 API.
+	if !utils.IsAPIServerReady(r.client, reqLogger) {
```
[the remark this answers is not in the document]
pasanw: I agree, I think we can safely remove this.

Flagged for resolution in the immediate follow-up PR (alongside the removal/restructure of other waits and conditionals).

pkg/controller/authentication/authentication_controller.go
```diff
@@ -196,6 +230,18 @@ func (r *ReconcileAuthentication) Reconcile(ctx context.Context, request reconci
 		}
 	}
 
+	// Ensure the allow-tigera tier exists, before rendering any network policies within it.
+	if err := r.client.Get(ctx, client.ObjectKey{Name: networkpolicy.TigeraComponentTierName}, &v3.Tier{}); err != nil {
+		if errors.IsNotFound(err) {
+			r.status.SetDegraded("Waiting for Tigera component policy tier to be created", err.Error())
```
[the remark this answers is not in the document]
pasanw: Will resolve in current PR.

pkg/controller/clusterconnection/clusterconnection_controller.go
```diff
@@ -212,14 +253,69 @@ func (r *ReconcileConnection) Reconcile(ctx context.Context, request reconcile.R
...
+	//
+	// * In managed clusters, the License can only be pushed once Guardian has been deployed, and (as always) the Tier
+	//   can only be created once the License is available.
+	includeNetworkPolicy := true
+	var egressAccessControlFeatureRequired bool
+	var networkPolicyIsReconcilable bool
+	if clusterAddrHasDomain, err := managementClusterAddrHasDomain(managementClusterConnection); err == nil && clusterAddrHasDomain {
+		egressAccessControlFeatureRequired = true
+		networkPolicyIsReconcilable = utils.IsV3NetworkPolicyReconcilable(ctx, r.Client, networkpolicy.TigeraComponentTierName, common.EgressAccessControlFeature)
+	} else {
```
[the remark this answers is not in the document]
pasanw: Will resolve in current PR.

pkg/controller/installation/core_controller.go
```diff
@@ -951,6 +970,7 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 	var managementCluster *operator.ManagementCluster
 	var managementClusterConnection *operator.ManagementClusterConnection
 	var logCollector *operator.LogCollector
+	includeV3NetworkPolicy := false
```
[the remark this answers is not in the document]
pasanw: Will resolve in current PR.

pkg/controller/installation/core_controller.go
```diff
@@ -979,6 +999,27 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
...
+		// is reconcilable (by enabling the availability of the API server and the creation of containing Tier). Therefore,
+		// to prevent a chicken-and-egg scenario, we only reconcile non-NetworkPolicy resources once we can confirm that
+		// all requirements to reconcile NetworkPolicy have been met.
+		//
+		// utils.IsV3NetworkPolicyReconcilable does not verify API server availability, so we take extra precaution
+		// when rendering components below.
+		if utils.IsV3NetworkPolicyReconcilable(ctx, r.client, networkpolicy.TigeraComponentTierName) {
+			includeV3NetworkPolicy = true
+
+			if !r.tierWatchReady.IsReady() {
```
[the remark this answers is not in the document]
pasanw: As discussed, I'll update all controllers such that tier and policy watches do not block reconcile from proceeding - we'll rely on the tier being available as a strong enough signal that we can proceed. 

Flagged for resolution in the immediate follow-up PR.

pkg/controller/utils/utils.go
```diff
@@ -217,6 +218,41 @@ func IsFeatureActive(license v3.LicenseKey, featureName string) bool {
...
+// Controllers can use this method to verify that the state required to reconcile v3 NetworkPolicy has been met.
+//
+// Limitation: the availability of the Tigera API server is not verified in this method to avoid excess queries to
+// the K8S API server. Therefore, components that render both NetworkPolicy resources and resources that impact
+// API server availability should render NetworkPolicy resources after their API-server-impacting resources.
+func IsV3NetworkPolicyReconcilable(ctx context.Context, cli client.Client, tierName string, licenseFeatureNames ...string) bool {
+	// Validate tier presence by querying for the tier. Note that this does not validate Tigera API server availability
+	// as resources are cached.
+	if err := cli.Get(ctx, client.ObjectKey{Name: tierName}, &v3.Tier{}); err != nil {
+		return false
```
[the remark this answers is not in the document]
pasanw: I believe we want this function to return false in either case - I assume you want the API access case to be logged?

Will resolve in current PR (by adding a log).

pkg/controller/monitor/monitor_controller.go
pasanw: Yep, I did mean NetworkPolicy, thanks for catching that. As discussed, we'll move towards an approach where controllers that impact NetworkPolicy reconciliation reconcile NetworkPolicy last, and check whether or not they can reconcile based on tier availability (without the use of a helper function).

Regarding the comment typo: Will resolve in current PR.

Regarding the controller update: Flagged for resolution in the immediate follow-up PR.

pkg/controller/tiers/tiers_controller.go
pasanw: Nice catch, originally this comment was in reference to the Guardian policy which had the possibility of using a domain name in its egress rules. It's now out of place given the refactor.

Will address in current PR.

pkg/controller/utils/utils.go
```diff
@@ -467,3 +524,37 @@ func isResourceReady(client kubernetes.Interface, resourceKind string) (bool, er
 	}
 	return false, nil
 }
+
+// Creates a predicate for CRUD operations that matches the object's namespace, and name if provided.
+// If neither name nor namespace is provided, all objects will be matched.
+func createNamespacePredicate(objMeta metav1.Object) predicate.Predicate {
```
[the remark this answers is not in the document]
pasanw: Will address in current PR.

pkg/render/packet_capture_api.go
```diff
@@ -69,6 +77,10 @@ func PacketCaptureAPI(cfg *PacketCaptureApiConfiguration) Component {
 	}
 }
 
+func PacketCaptureAPIPolicy(cfg *PacketCaptureApiConfiguration) Component {
```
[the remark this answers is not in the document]
pasanw: As discussed, this pattern is intentional since some controllers have multiple components, therefore placing policy at the end of a components Objects() list is not a guarantee that policy is rendered last in the controller.

Will keep as-is, and apply the same pattern when clusterconnection and monitor controllers are adjusted in the immediate follow-up PR.

pkg/controller/apiserver/apiserver_controller.go
```diff
@@ -278,6 +305,27 @@ func (r *ReconcileAPIServer) Reconcile(ctx context.Context, request reconcile.Re
...
 		}
+
+		// Successful reconciliation of non-NetworkPolicy resources in the apiserver controller ensures that NetworkPolicy
+		// is reconcilable (by creating the API server and enabling the creation of containing Tier). Therefore, to prevent
+		// a chicken-and-egg scenario, we only reconcile non-NetworkPolicy resources once we can confirm that all
+		// requirements to reconcile NetworkPolicy have been met.
+		//
+		// utils.IsV3NetworkPolicyReconcilable does not verify API server availability, so we take extra precaution
+		// when rendering components below.
+		if utils.IsV3NetworkPolicyReconcilable(ctx, r.client, networkpolicy.TigeraComponentTierName) {
```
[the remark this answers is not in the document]
pasanw: As discussed, I'll update all controllers such that tier and policy watches do not block reconcile from proceeding - we'll rely on the tier being available as a strong enough signal that we can proceed. We'll also move towards an approach where controllers that impact NetworkPolicy reconciliation reconcile NetworkPolicy last, and check whether or not they can reconcile NetworkPolicy based on tier availability (without the use of a helper function).

Flagged for resolution in the immediate follow-up PR.

pkg/controller/utils/utils.go
pasanw: As discussed, we'll move towards an approach where controllers that impact NetworkPolicy reconciliation reconcile NetworkPolicy last, and check whether or not they can reconcile NetworkPolicy based on tier availability (without the use of a helper function).

Flagged for resolution in the immediate follow-up PR.

pkg/controller/utils/utils.go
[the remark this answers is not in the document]
caseydavenport: I think we want more than a log - it should be hooked into the status manager for both cases (404 and other errors)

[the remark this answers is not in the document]
caseydavenport: We might need to pass a status manager object to this function to properly report errors with the correct context.

pkg/render/tiers/tiers.go
caseydavenport: Ah, yeah I think I just missed the "Pass" statement. 

I'd probably still leave it as-is and keep to our stance of translating 1:1 for the first pass, and then making changes in a subsequent iteratino.

PR #2008, opened 2022-06-10
Tolerate IP6_AUTODETECTION_METHOD none when IP6 is none

tmjd:
Even though IP6_AUTODETECTION_METHOD = none is not valid, kops sets
it when IP6 is set to none also.
https://github.com/kubernetes/kops/blob/v1.23.2/upup/models/cloudup/resources/addons/networking.projectcalico.org/k8s-1.16.yaml.template#L4393

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/migration/convert/network.go
+++ b/pkg/controller/migration/convert/network.go
@@ -138,6 +138,11 @@ func handleCalicoCNI(c *components, install *operatorv1.Installation) error {
 			if err := c.node.assertEnv(ctx, c.client, containerCalicoNode, "FELIX_IPV6SUPPORT", "false"); err != nil {
 				return err
 			}
+			// If IP6=none then if IP6_AUTODETECTION_METHOD is set it must be none. This is not a valid value
+			// for the env var but Kops sets it.
+			if err := c.node.assertEnv(ctx, c.client, containerCalicoNode, "IP6_AUTODETECTION_METHOD", "none"); err != nil {
+				return err
+			}
 		} else if *ip6 == "autodetect" {
 			if err := handleIPv6AutoDetectionMethod(c, install); err != nil {
 				return err
--- a/pkg/controller/migration/convert/network_test.go
+++ b/pkg/controller/migration/convert/network_test.go
@@ -195,7 +195,35 @@ var _ = Describe("Convert network tests", func() {
 					Value: "false",
 				},
 			)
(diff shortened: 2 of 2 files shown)


pkg/controller/migration/convert/network.go
```diff
@@ -138,6 +138,11 @@ func handleCalicoCNI(c *components, install *operatorv1.Installation) error {
 			if err := c.node.assertEnv(ctx, c.client, containerCalicoNode, "FELIX_IPV6SUPPORT", "false"); err != nil {
 				return err
 			}
+			// If IP6=none then if IP6_AUTODETECTION_METHOD is set it must be none. This is not a valid value
+			// for the env var but Kops sets it.
+			if err := c.node.assertEnv(ctx, c.client, containerCalicoNode, "IP6_AUTODETECTION_METHOD", "none"); err != nil {
```
rene-dekker, line 143: Maybe it could be something like:
```
	env, err := getEnv(ctx, c.client, c.node.Spec.Template.Spec, containerCalicoNode, ComponentCalicoNode, "IP6_AUTODETECTION_METHOD")
	if err != nil {
		return err
	} else if env != nil && *env == "none" {
		c.node.ignoreEnv(containerCalicoNode, "IP6_AUTODETECTION_METHOD")
	}
	if err := c.node.assertEnv(ctx, c.client, containerCalicoNode, "IP6_AUTODETECTION_METHOD", "none"); err != nil {
		return err
	}
```

That way the intent of the comment matches better with the code. We want to give Kops users a pass with their (mis)configuration without asserting that the value is correct.

rene-dekker:
Do we want the fix to be picked to other release branches? I'm asking because we are preparing a possible EE patch release containing operator v1.26.x and v1.25.x

pkg/controller/migration/convert/network.go
tmjd: I believe what you have written is what assertEnv does. I don't think I'd want to use a different pattern here since we use assertEnv in many places for exactly this purpose.

tmjd: Do you think the check here is different functionally than the other places it is used?

rene-dekker (commented):
lgtm

rene-dekker (approved):

pkg/controller/migration/convert/network.go
rene-dekker: My snippet missed checking for empty values, but I realize now that does not really apply here. If the env does not exist then nil is returned from getEnv. I approved it.

PR #2057, opened 2022-06-29
Refactor semaphore jobs

Brian-McM:
## Description
This commit:
- Separates the `cd` / `maybe-build-release` targets into a pipeline
  that can be promoted
- Remove `vet` as a dependency of build so we don't vet for all arches
  when creating the images (takes a very long time for non amd64 builds)
- Splits UT / FV / static tests to run asynchronously to reduce build
  time
- Build / download go dependencies for arm64 and ppc64le asynchronously
  while running the CI and cache them so the push images step doesn't
  have to spend 15-20m gettings the dependencies / building the
  binaries.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -1,8 +1,18 @@
 version: v1.0
-name: Initial Pipeline
+name: Operator CI
+
+execution_time_limit:
+  hours: 4
+
+auto_cancel:
+  running:
+    when: "branch != 'master'"
+  queued:
+    when: "branch != 'master'"
+
 agent:
   machine:
-    type: e1-standard-8
+    type: e1-standard-2
     os_image: ubuntu1804
 global_job_config:
   secrets:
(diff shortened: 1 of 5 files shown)


.semaphore/semaphore.yml
```diff
@@ -11,52 +21,116 @@ global_job_config:
...
+
+promotions:
+  - name: Push Images
+    pipeline_file: push_images.yml
+    auto_promote:
+      when: "branch =~ 'master|release-.*'"
+  - name: Clean Up
+    pipeline_file: clean_up.yml
+    # Don't auto promote if this is master or a release branch so the cache is available for the Push Images pipeline.
+#    auto_promote:
```
tmjd, line 132: Should this auto_promote when something fails? If that happens then don't we need to rerun anyway so there won't be any use of the cache and ideally the failure will mean we need to make an update which would mean the current cache won't be used either.

tmjd: Probably should promote when it fails or not on master or release branch.

.semaphore/semaphore.yml
```diff
@@ -11,52 +21,116 @@ global_job_config:
...
+      jobs:
+        - name: Static / File Generation Checks, Vet
+          execution_time_limit:
+            minutes: 15
           commands:
-            - cache store go-pkg-cache .go-pkg-cache
-            - 'cache store go-mod-cache ${HOME}/go/pkg/mod/cache'
+            - make format-check validate-gen-versions fmt
+            - make dirty-check
+            - make vet
```
tmjd, line 96: I think most of these were accomplished from the `ci` target before, but I don't see a call to `test-crds`, I think that should be added here too.

Makefile
```diff
@@ -388,7 +393,7 @@ validate-gen-versions:
 	make dirty-check
 
 ## Deploys images to registry
-cd:
+cd: image-all
```
tmjd, line 396: If anyone ever has to do cd manually it might be useful to have the `cd` target not invoking the `image-all` target. WDYT?

.semaphore/semaphore.yml
Brian-McM: Right, I'll add that, thanks

Makefile
Brian-McM: cd doesn't make sense without image-all though, it really is a requirement of calling cd.

.semaphore/semaphore.yml
Brian-McM: This was just temporarily commented out, it will auto promote for non master / release branches. The push images pipeline is responsible for promoting to clean on master / release branches (if the pipeline doesn't fail, there's a comment there about that).

Makefile
tmjd: That means though that running cd will always build the images, it won't be possible to separate those steps. Ideally there would be a file indicator or way to check that the images exist rather than having to invoke `image-all` on any invocation of `cd`.
But we can always change this in the future, if we split building and pushing of images into separate steps.

.semaphore/semaphore.yml
Brian-McM: Because we'd likely want to either 1. rerun the job or 2. log into the box and try cutting the images manually. If we promote on failure then the go package cache would be clear, making either of those options take 20 minutes instead of 2.

So I see it as quite important to not clear the cache unless it passes.

Makefile
Brian-McM: Running CD should always build the images if they're not there, it is a dependency of CD. If image-all has already been called then it should be quick (rebuilding docker images when nothing has changed is pretty immediate). It could be skipped entirely using some other method (possibly the file descriptor method but I'm not a huge fan of that), but again, rebuilding a docker image where nothing has changed is extremely fast (docker got smart about this).

.semaphore/semaphore.yml
tmjd: Those seem like edge cases where more generally I think clearing the cache is the right thing to do. I guess if we find not clearing the cache is problematic we can change it then.

Issue #1990, opened 2022-06-01
Upcoming operator component configuration changes


lmm:
Hey everyone!

We are planning on adding new fields to the operator API to allow more configuration of the resources managed by the operator. We would appreciate your feedback on these proposed changes.

## Why?

The goal of these proposed changes is to meet some of the use cases highlighted in these GitHub issues:
- [The controlPlaneNodeSelector installation field doesn't effect Typha](https://github.com/tigera/operator/issues/1653)
- [Add support for additional labels and annotations to be propagated to all pods](https://github.com/tigera/operator/issues/1099)
- [Allow objects to have custom labels](https://github.com/tigera/operator/issues/1385)
- [Allow resource settings to be specified for API Server](https://github.com/tigera/operator/issues/1757)
- [Hard-coded tolerations allow non-daemonset resources to be scheduled on a cordoned node](https://github.com/projectcalico/calico/issues/6136)
- [Topology spread or antiAffinity rules for calico controllers and typha](https://github.com/projectcalico/calico/issues/5577)
- [calico/node tolerations](https://github.com/tigera/operator/issues/2020)

## The proposed changes

We will add new fields to allow configuring a component resource (such as calico node daemonset, calico kube-controllers deployment, etc). These fields will include (depending on the component):
- annotations and labels
- node affinity and node selectors
- tolerations
- minReadySeconds
- container [resources](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#resource-requests-and-limits-of-pod-and-container)

We will start by adding configuration to the core Calico components:
- calico-node (calico-node daemonset)
- calico-typha (calico-typha deployment)
- calico-kube-controllers (calico-kube-controllers deployment)
- calico-apiserver (calico-apiserver deployment)

The new fields will be backed by types that closely resemble the upstream Kubernetes resources (daemonset, deployment) that are being configured. The core Calico components `calico-node`, `calico-typha`, and `calico-kube-controllers` are managed and configured in the Installation CRD, so the new fields for these components will be added to the InstallationSpec.
The new field to configure the `calico-apiserver` component will be added to the APIServerSpec. (In general, the new fields will be added to the component CRD that manages/configures those resources.) 

Here is an example Installation resource with these new fields (note: we have not ironed out the granular details of the API changes such as field names, etc.):

```
apiVersion: operator.tigera.io/v1
kind: Installation
metadata:
  name: default
spec:
  calicoNodeTemplate:
    metadata:
      labels:
        daemonset-label: value
[...]

PR #2063, opened 2022-07-05
Add support for component resource overrides

lmm:
## Description

This PR addresses #1990 by adding support for the configuration of the core Calico component resources. These are:
- calico-node daemonset
- calico-typha deployment
- calico-kube-controllers deployment
- calico-apiserver deployment
- calico-windows-upgrade daemonset

Each of those resources will have a configuration field added to the InstallationSpec. The configurable fields are:
- top-level labels and annotations
- minReadySeconds
- Pod labels and annotations
- initContainer/container resources
- affinity
- nodeSelector
- tolerations

**Tasks**
- [x] Add new field for calico-node daemonset
- [x] Update validation of installation spec before render
- [ ] Handle migration of new field
- [x] Add other fields

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


api/v1/calico_node_types.go
```diff
@@ -0,0 +1,108 @@
...
+	// available. Defaults to 0 (pod will be considered available as soon as it
+	// is ready).
+	// +optional
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:validation:Maximum=2147483647
+	MinReadySeconds *int32 `json:"minReadySeconds,omitempty"`
+	// Template describes the calico-node DaemonSet pod that will be created.
+	// +optional
+	Template *CalicoNodeDaemonSetPodTemplateSpec `json:"template,omitempty"`
+}
```
lmm, line 108: With these nested struct pointers, I'm thinking it might be nice to have accessors to reach into the nested struct. E.g.:

```
// Container is an arbitrary container.
type Container struct {
	// Name is identifies the container by name.
	Name string `json:"name"`
	// Resources allows customization of limits and requests for compute resources such as cpu and memory.
	Resources *v1.ResourceRequirements `json:"resources"`
}

func (c *CalicoNodeDaemonSet) GetPodTemplateMetadata() *Metadata {
	if c.Spec != nil {
		if c.Spec.Template != nil {
			return c.Spec.Template.Metadata
		}
	}
	return nil
}

func (c *CalicoNodeDaemonSet) GetInitContainers() []Container {
	if c.Spec != nil {
		if c.Spec.Template != nil {
			if c.Spec.Template.Spec != nil {
				if c.Spec.Template.Spec.InitContainers != nil {
					cs := make([]Container, len(c.Spec.Template.Spec.InitContainers))
					for i, v := range c.Spec.Template.Spec.InitContainers {
						c := Container{Name: v.Name, Resources: v.Resources}
						cs[i] = c
					}
					return cs
				}
			}
		}
	}
	return nil
}
```

pkg/render/common/components/components.go
```diff
@@ -0,0 +1,176 @@
...
+	if tolerations != nil {
+		ds.Spec.Template.Spec.Tolerations = tolerations
+	}
+
+	return ds
+}
+
+// mergeMaps adds all key/value pairs from the provided map to the current map
+// if those keys do not already exist in the current map.
+func mergeMaps(current map[string]string, provided map[string]string) {
```
caseydavenport, line 144: I believe we have a function in another package that does this already - maybe should make that one public.

caseydavenport: https://github.com/tigera/operator/blob/7de6f2820556f766aed93eded7ef1f2895f980f7/pkg/controller/utils/component.go#L527-L537

pkg/render/common/components/components.go
```diff
@@ -0,0 +1,176 @@
...
+func mergeContainers(current []corev1.Container, provided []container) {
+	for _, pc := range provided {
+		// Skip if no resources defined.
+		if pc.Resources == nil {
+			continue
+		}
+
+		var index int
+		var found bool
+		for i, cc := range current {
```
caseydavenport, line 165: This might be a bit simpler if `provided` was instead a `map[string]container` - then, you could adjust the logic to be something like this:

```go
for i, c : = range current {
  if override, ok := provided[c.Name]; ok {
    current[i].Resources = override.Resources
  }
}
```

pkg/render/node.go
```diff
@@ -215,7 +216,10 @@ func (c *nodeComponent) Objects() ([]client.Object, []client.Object) {
 		objs = append(objs, c.nodePodSecurityPolicy())
 	}
 
-	objs = append(objs, c.nodeDaemonset(cniConfig))
+	// Note: this will override the values from nodeResources() (if it was defined for calico-node).
+	overrides := c.cfg.Installation.CalicoNodeDaemonSet
+	ds := rcomp.ApplyDaemonSetOverrides(c.nodeDaemonset(cniConfig), overrides)
```
caseydavenport, line 221: I think it would be nicer to put this within the `c.nodeDaemonset` function call, so the complete generation of the daemonset is handled within that function.

pkg/render/common/components/components.go
```diff
@@ -0,0 +1,176 @@
...
+	corev1 "k8s.io/api/core/v1"
+)
+
+type container struct {
+	Name      string                       `json:"name"`
+	Resources *corev1.ResourceRequirements `json:"resources"`
+}
+
+// ApplyDaemonSetOverrides applies the overrides to the given DaemonSet.
+func ApplyDaemonSetOverrides(ds *appsv1.DaemonSet, overrides interface{}) *appsv1.DaemonSet {
```
caseydavenport, line 29: I'm not a huge fan of allowing an `interface{}`as an argument here. I suspect you're doing this because we'll need to handle multiple `XDaemonSet` API structs that have the same fields but different types, but it makes the code a bit messy IMO.

What do you think about an interface-based approach? Something like this maybe: https://go.dev/play/p/QuizDYKfdmr

It would have a few benefits:
- Makes the code easier to follow, and easier to understand what is a valid argument for this function.
- Ensures that each XDaemonSet has a consistent set of fields, requires that they implement the same interface.

caseydavenport (changes requested):
First round of comments / questions - let me know what you think!

pkg/render/common/components/components.go
lmm: Yeah I think I was beginning to come around to your suggestion with my [comment](https://github.com/tigera/operator/pull/2063#discussion_r915282729) above

api/v1/calico_node_types.go
```diff
@@ -0,0 +1,108 @@
...
+*/
+
+package v1
+
+import (
+	v1 "k8s.io/api/core/v1"
+)
+
+// Metadata contains the standard Kubernetes labels and annotations fields.
+type Metadata struct {
```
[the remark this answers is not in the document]
lmm: e52a1765d83f5942e90fe2b201085ad27b0de52c

api/v1/calico_node_types.go
```diff
@@ -0,0 +1,108 @@
...
+	// Labels is a map of string keys and values that may match replicaset and
+	// service selectors. Each of these key/value pairs are added to the
+	// object's labels provided the key does not already exist in the object's labels.
+	// +optional
+	Labels map[string]string `json:"labels,omitempty"`
```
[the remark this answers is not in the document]
lmm: 2cfbf61e751d18d332440b0db0b003a62249a0ca

api/v1/calico_node_types.go
```diff
@@ -0,0 +1,108 @@
...
+	Resources *v1.ResourceRequirements `json:"resources"`
+}
+
+// CalicoNodeInitContainer is a calico-node DaemonSet init container.
+type CalicoNodeInitContainer struct {
+	// Name is an enum which identifies the calico-node DaemonSet init container by name.
+	// +kubebuilder:validation:Enum=install-cni;hostpath-init;flexvol-driver;mount-bpffs;node-certs-key-cert-provisioner;calico-node-prometheus-server-tls-key-cert-provisioner
+	Name string `json:"name"`
+	// Resources allows customization of limits and requests for compute resources such as cpu and memory.
+	Resources *v1.ResourceRequirements `json:"resources"`
```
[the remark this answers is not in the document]
lmm: b5847dd2ff8bc6fc51b21ad66256b1fea1e3deb9

api/v1/calico_node_types.go
```diff
@@ -0,0 +1,108 @@
...
+	Annotations map[string]string `json:"annotations,omitempty"`
+}
+
+// CalicoNodeContainer is a calico-node DaemonSet container.
+type CalicoNodeContainer struct {
+	// Name is an enum which identifies the calico-node DaemonSet container by name.
+	// +kubebuilder:validation:Enum=calico-node
+	Name string `json:"name"`
+	// Resources allows customization of limits and requests for compute resources such as cpu and memory.
+	Resources *v1.ResourceRequirements `json:"resources"`
```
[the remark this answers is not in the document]
lmm: b5847dd2ff8bc6fc51b21ad66256b1fea1e3deb9
and
b1d04180f29e40ed70eab4efa0ea478cb45f8e60

api/v1/calico_node_types.go
```diff
@@ -0,0 +1,108 @@
...
+
+// CalicoNodeDaemonSetPodSpec is the calico-node DaemonSet's PodSpec.
+type CalicoNodeDaemonSetPodSpec struct {
+	// InitContainers is a list of calico-node init containers.
+	// +optional
+	InitContainers []CalicoNodeInitContainer `json:"initContainers,omitempty"`
+	// Containers is a list of calico-node containers.
+	// +optional
+	Containers []CalicoNodeContainer `json:"containers,omitempty"`
+	// Affinity is a group of affinity scheduling rules for the calico-node pods.
```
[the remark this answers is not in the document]
lmm: b1d04180f29e40ed70eab4efa0ea478cb45f8e60

api/v1/installation_types.go
```diff
@@ -145,6 +148,14 @@ type InstallationSpec struct {
 	// NonPrivileged configures Calico to be run in non-privileged containers as non-root users where possible.
 	// +optional
 	NonPrivileged *NonPrivilegedType `json:"nonPrivileged,omitempty"`
+
+	// CalicoNodeDaemonset configures the calico-node DaemonSet. If used in
+	// conjunction with the deprecated ComponentResources, then these overrides take precedence.
+	// WARNING: Please note that this configuration overrides the operator's
+	// default calico-node DaemonSet configuration. The operator cannot validate
+	// that the provided configuration will work in your cluster. An invalid
+	// configuration may may cause networking disruption or even catastrophic failure!
```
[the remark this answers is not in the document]
lmm: fixed

api/v1/installation_types.go
```diff
@@ -145,6 +148,14 @@ type InstallationSpec struct {
 	// NonPrivileged configures Calico to be run in non-privileged containers as non-root users where possible.
 	// +optional
 	NonPrivileged *NonPrivilegedType `json:"nonPrivileged,omitempty"`
+
+	// CalicoNodeDaemonset configures the calico-node DaemonSet. If used in
+	// conjunction with the deprecated ComponentResources, then these overrides take precedence.
+	// WARNING: Please note that this configuration overrides the operator's
+	// default calico-node DaemonSet configuration. The operator cannot validate
```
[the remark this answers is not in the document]
lmm: I agree with you that there are a multitude of ways to break an operator-managed cluster via the API.

That said, I added this warning because I recalled that when we all chatted about the design that @tmjd wanted some warning in the API docs (I may be wrong). A plus is that this warning _might_ nudge users to think twice about their config changes.

api/v1/installation_types.go
```diff
@@ -131,6 +132,8 @@ type InstallationSpec struct {
 	// +optional
 	NodeUpdateStrategy appsv1.DaemonSetUpdateStrategy `json:"nodeUpdateStrategy,omitempty"`
 
+	// Deprecated. Please use CalicoNodeDaemonSet,
+	// TyphaDeployment, and KubeControllersDeployment.
```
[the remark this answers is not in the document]
lmm: fixed!

api/v1/installation_types.go
```diff
@@ -189,6 +200,7 @@ const (
 	ComponentNameKubeControllers ComponentName = "KubeControllers"
 )
 
+// Deprecated. Please use component resource config fields in Installation.Spec instead.
```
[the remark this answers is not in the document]
lmm: I think this struct we can deprecate. The enterprise components define their own ComponentResources type.

pkg/render/node.go
lmm: Fixed

pkg/render/common/components/components.go
lmm: What do you think of: a929912d44939dce425bf670df170fbe6145bdcf ?

I wanted the Container type to be an interface too but that leads to an import cycle.

pkg/render/common/components/components.go
```diff
@@ -0,0 +1,176 @@
...
+		for i, cc := range current {
+			if cc.Name == pc.Name {
+				index = i
+				found = true
+				break
+			}
+		}
+		if found {
+			current[index].Resources = *pc.Resources
+		}
```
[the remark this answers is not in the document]
lmm: Because this is called from the node render, there isn't a good way to bubble up errors. The render.Component interface: https://github.com/tigera/operator/blob/master/pkg/render/component.go#L32

We'll have CRD validation and installation validation from the core controller at least.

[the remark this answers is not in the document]
caseydavenport: > should only happen in the case of a code bug, but would be good to fail loudly rather than silently.

Sorry, what I meant by this is that it should never happen assuming there are no bugs in the calling code - we don't need to bubble an error up, but we should at least log an error, e.g., `log.Error("BUG: hit unexpected line")`

api/v1/common_types.go
```diff
@@ -0,0 +1,47 @@
...
+package v1
+
+import (
+	v1 "k8s.io/api/core/v1"
+)
+
+// Container is an arbitrary Kubernetes container.
+// Note: this should not be used directly as a field type within a component resource type.
+// Component resources should define their own container types with validation.
+type Container struct {
```
caseydavenport, line 26: Mentioned in another thread - what do you think about getting rid of this intermediate struct and instead using a v1.Container? 

I think it might make things a bit simpler. Given this struct isn't actually part of our API, I think it's OK and will remove the need to define a new structure.

pkg/render/common/components/components.go
```diff
@@ -0,0 +1,121 @@
...
+	}
+
+	var metadata *operatorv1.Metadata
+	var minReadySeconds *int32
+	var podTemplateMetadata *operatorv1.Metadata
+
+	var affinity *corev1.Affinity
+	var tolerations []corev1.Toleration
+	var nodeSelector map[string]string
+	var initContainers []operatorv1.Container
```
caseydavenport, line 41: This code will be a lot easier to follow if we move the declaration of the variables closer to where we use them. 

Also, I don't think we need to have a separate variable declaration / population any more. Could probably even do something like this:

```go
if ics := overrides.GetInitContainers(); ics != nil {
  // Handle init container overrides.
}
```

pkg/render/node.go
```diff
@@ -215,7 +216,8 @@ func (c *nodeComponent) Objects() ([]client.Object, []client.Object) {
 		objs = append(objs, c.nodePodSecurityPolicy())
 	}
 
-	objs = append(objs, c.nodeDaemonset(cniConfig))
```
caseydavenport, line 218: Let's just revert this to the original line so that it's consistent with the rest of the object rendering. 

```go
objs = append(objs, c.nodeDaemonset(cniConfig))
```

api/v1/installation_types.go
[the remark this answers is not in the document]
tmjd: I'm not sure that we need a warning at this level. There is a lot that is configurable that we expect is normal and fairly safe to change.
I think we should probably be more targeted with our warnings, like I feel like we should have warnings on NodeSelector, Tolerations, and affinity.

api/v1/common_types.go
lmm: a96a441d9cc159e1c9f145ec053523222fdb42a7

pkg/render/common/components/components.go
lmm: Fixed

pkg/render/common/components/components.go
lmm: Done in a96a441d9cc159e1c9f145ec053523222fdb42a7

api/v1/installation_types.go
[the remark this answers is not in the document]
lmm: Addressed in 7e095012ebfd566145685e3fb5f32c619efaeb2b

pkg/common/validation/overrides.go
```diff
@@ -0,0 +1,96 @@
...
+type ValidateContainer func(container corev1.Container) error
+
+// NoContainersDefined is a container validation function that is used when no container is expected.
+var NoContainersDefined ValidateContainer = func(container corev1.Container) error {
+	return fmt.Errorf("container %q is invalid. No containers are expected", container.Name)
+}
+
+// ValidateReplicatedPodResourceOverrides validates the given replicated pod resource overrides.
+// validateContainerFn and validateInitContainerFn are used to validate the container overrides.
+func ValidateReplicatedPodResourceOverrides(overrides components.ReplicatedPodResourceOverrides, validateContainerFn ValidateContainer, validateInitContainerFn ValidateContainer) error {
```
caseydavenport, line 38: Maybe a nice cleanup for a future PR would be to instead define an interface type for the validation funcs here:

```go

type OverrideValidator interface {
  ValidateContainer func() err
  ValidateInitContainer func() err
}
```

It makes it fairly easy to extend this function in the future as overrides grows. This is very minor and would recommend against doing that in this PR, but maybe someday if we need to add a third argument to this function an interface would be a clean way to do it.

pkg/render/common/components/components.go
```diff
@@ -0,0 +1,159 @@
...
+	if tolerations := overrides.GetTolerations(); tolerations != nil {
+		r.podTemplateSpec.Spec.Tolerations = tolerations
+	}
+
+	return r
+}
+
+// ApplyDaemonSetOverrides applies the overrides to the given DaemonSet.
+// Note: overrides must not be nil pointer.
+func ApplyDaemonSetOverrides(ds *appsv1.DaemonSet, overrides components.ReplicatedPodResourceOverrides) *appsv1.DaemonSet {
```
caseydavenport, line 92: This function takes a pointer and returns a pointer to the same object - might be simpler to just always rely on the side effect? Or to make a copy at the top so we're not modifying the original object.

caseydavenport: Looks like you don't actually use the return value, so simply removing the return would do it.

lmm: Fixed cf665ca31fcb7ccfd197bfee40136a07b90fd6b2

pkg/common/k8svalidation/helper/helper.go
```diff
@@ -0,0 +1,245 @@
...
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/conversion"
+	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/selection"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/apimachinery/pkg/util/validation"
+
+	// "k8s.io/kubernetes/pkg/apis/core"
```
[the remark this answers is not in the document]
lmm: Done

api/v1/apiserver_types.go
```diff
@@ -61,3 +67,202 @@ type APIServerList struct {
 func init() {
 	SchemeBuilder.Register(&APIServer{}, &APIServerList{})
 }
+
+// APIServerDeploymentContainer is an API server deployment container.
+type APIServerDeploymentContainer struct {
+	// Name is an enum which identifies the API server Deployment container by name.
+	// +kubebuilder:validation:Enum=calico-apiserver;tigera-apiserver;tigera-queryserver
```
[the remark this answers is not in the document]
lmm: We chatted about this over slack. It's probably okay to make this container name change.

I've made the change in 5c6f577a7bba814487ca67dc2843cc8844a7407e

pkg/common/validation/calico-node/validation.go
```diff
@@ -0,0 +1,43 @@
...
+		return fmt.Errorf("Installation spec.CalicoNodeDaemonSet.Spec.Template.Spec.Containers[%q] is not a supported container", container.Name)
+	}
+
+	errs := k8svalidation.ValidateResourceRequirements(&container.Resources, field.NewPath("spec", "template", "spec", "containers"))
+	return errs.ToAggregate()
+}
+
+// ValidateCalicoNodeDaemonSetInitContainer validates the given container is a valid calico-node DaemonSet init container.
+func ValidateCalicoNodeDaemonSetInitContainer(container corev1.Container) error {
+	if _, ok := render.CalicoNodeDaemonSetInitContainerNames[container.Name]; !ok {
```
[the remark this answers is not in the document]
lmm: That 🐛 is now squashed

pkg/common/validation/apiserver/validation.go
```diff
@@ -0,0 +1,45 @@
...
+	corev1 "k8s.io/api/core/v1"
+
+	"github.com/tigera/operator/pkg/common/k8svalidation"
+	"github.com/tigera/operator/pkg/render"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+)
+
+// ValidateAPIServerDeploymentContainer validates the given container is a valid API server Deployment container.
+func ValidateAPIServerDeploymentContainer(container corev1.Container) error {
+	if _, ok := render.APIServerDeploymentContainerNames[container.Name]; !ok {
```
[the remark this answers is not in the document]
lmm: Yeah this other layer of validation doesn't make sense. I'll remove it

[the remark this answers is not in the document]
lmm: [549a3e0](https://github.com/tigera/operator/pull/2063/commits/549a3e06e688a6815056d780cfed74a917f95eb8)

Makefile
```diff
@@ -309,9 +309,13 @@ cluster-create: $(BINDIR)/kubectl $(BINDIR)/kind
 
 ## Deploy CRDs needed for UTs.  CRDs needed by ECK that we don't use are not deployed.
 ## kubectl create is used for prometheus as a workaround for https://github.com/prometheus-community/helm-charts/issues/1500
+## kubectl create is used for operator CRDS since the Installation API is large enough now that we hit the following error:
+##
+##   The CustomResourceDefinition "installations.operator.tigera.io" is invalid: metadata.annotations: Too long: must have at most 262144 bytes
```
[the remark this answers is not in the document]
lmm: Based on your suggestion, I've tested out a helm install with the CRDs generated from this PR and it installed without error.

pkg/render/apiserver.go
```diff
@@ -55,6 +57,32 @@ const (
 	QueryServerPort        = 8080
 	QueryserverNamespace   = "tigera-system"
 	QueryserverServiceName = "tigera-api"
+
+	TigeraAPIServerResourceName = "tigera-apiserver"
```
caseydavenport, line 61: Is this ever used?

pkg/render/node.go
```diff
@@ -66,6 +67,24 @@ var (
 	nodeBGPReporterPort int32 = 9900
 
 	NodeTLSSecretName = "node-certs"
+
+	// calico-node DaemonSet container names used for validation.
+	// These should match CalicoNodeContainer.Name validation on the CalicoNodeDaemonSet type.
+	CalicoNodeDaemonSetContainerNames = map[string]struct{}{
```
caseydavenport, line 73: Doesn't look like this map is ever used (nor the one below)

pkg/render/typha.go
```diff
@@ -42,12 +44,21 @@ const (
...
 
 var (
 	TyphaTLSSecretName   = "typha-certs"
 	TyphaCAConfigMapName = "typha-ca"
 	TyphaCABundleName    = "caBundle"
+
+	// typha Deployment init container names used for validation.
+	// These should match TyphaDeploymentInitContainer.Name validation on the TyphaDeployment type.
+	// This name is generated in certificatemanagement.CreateCSRInitContainer
+	TyphaInitContainerName = fmt.Sprintf("%s-%s", TyphaTLSSecretName, certificatemanagement.CSRInitContainerName)
```
caseydavenport, line 61: Unused also, I think.

pkg/render/apiserver.go
lmm: Removed 71f99a5

pkg/render/node.go
lmm: Removed 71f99a5

pkg/render/typha.go
lmm: Removed with 71f99a5

caseydavenport (approved):
Migration logic to come in part II! 

Thanks @lmm!

Makefile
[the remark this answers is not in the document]
sswaminathan-godaddy: you could use a server-side apply

PR #2076, opened 2022-07-08
Fix PodSecurityStandard for intrusion detection

rene-dekker:
Fixes:
```
$ k describe replicaset.apps/anomaly-detection-api-5fc4f84f99
  Warning  FailedCreate  62s   replicaset-controller  Error creating: pods "anomaly-detection-api-5fc4f84f99-6hrgg" is forbidden: violates PodSecurity "restricted:latest": allowPrivilegeEscalation != false (container "anomaly-detection-api" must set securityContext.allowPrivilegeEscalation=false), unrestricted capabilities (container "anomaly-detection-api" must set securityContext.capabilities.drop=["ALL"]), runAsNonRoot != true (pod or container "anomaly-detection-api" must set securityContext.runAsNonRoot=true), seccompProfile (pod or container "anomaly-detection-api" must set securityContext.seccompProfile.type to "RuntimeDefault" or "Localhost")

```
that affects both deployments in the tigera-intrusion-detection namespace

### DIFF
--- a/pkg/render/intrusion_detection.go
+++ b/pkg/render/intrusion_detection.go
@@ -146,7 +146,7 @@ func (c *intrusionDetectionComponent) SupportedOSType() rmeta.OSType {
 func (c *intrusionDetectionComponent) Objects() ([]client.Object, []client.Object) {
 	// Configure pod security standard. If syslog forwarding is enabled, we
 	// need hostpath volumes which require a privileged PSS.
-	pss := PSSRestricted
+	pss := PSSBaseline
 	if c.syslogForwardingIsEnabled() {
 		pss = PSSPrivileged
 	}


pkg/render/intrusion_detection.go
```diff
@@ -146,7 +146,7 @@ func (c *intrusionDetectionComponent) SupportedOSType() rmeta.OSType {
 func (c *intrusionDetectionComponent) Objects() ([]client.Object, []client.Object) {
 	// Configure pod security standard. If syslog forwarding is enabled, we
 	// need hostpath volumes which require a privileged PSS.
-	pss := PSSRestricted
+	pss := PSSBaseline
```
hjiawei, line 149: > container "anomaly-detection-api" must set securityContext.allowPrivilegeEscalation=false

Probably better do so in ad-api container `SecurityContext` instead of setting this to baseline.

PR #2082, opened 2022-07-15
Add CSI driver support

Josh-Tigera:
With FlexVolumes being deprecated we need to add support for CSI to the operator.

Manually tested on a local cluster to verify that CSI daemonset and CSIDriver object is correctly added when setting the new field to "Enabled" and removed when removing it or setting it to "Disabled". Also added new UTs to validate API values and test rendering of newly created component.

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


api/v1/installation_types.go
```diff
@@ -126,6 +126,11 @@ type InstallationSpec struct {
 	// +optional
 	FlexVolumePath string `json:"flexVolumePath,omitempty"`
 
+	// VolumePlugin optionally specifies enablement of Calico CSI plugin. If not specified,
+	// VolumePlugin will be disabled by default. Valid values are "Enabled" or "Disabled".
+	// +optional
+	VolumePlugin string `json:"volumePlugin,omitempty"`
```
tmjd, line 132: I think we should have a type for this since it is two static strings possible. Then the CRD validation can handle ensuring it is a valid value.

hack/gen-versions/components.go
```diff
@@ -31,6 +31,8 @@ var defaultImages = map[string]string{
 	"calicoctl":                  "calico/ctl",
 	"flannel":                    "coreos/flannel",
 	"flexvol":                    "calico/pod2daemon-flexvol",
+	"calico/csi":                 "calico/csi",
+	"csi-node-driver-registrar":  "csi-node-driver-registrar",
```
tmjd, line 35: I'm thinking the value should be `sig-storage/csi-node-driver-registrar` and then remove `sig-storage` from the registry?

pkg/controller/installation/validation.go
```diff
@@ -315,6 +315,11 @@ func validateCustomResource(instance *operatorv1.Installation) error {
 			instance.Spec.FlexVolumePath)
 	}
 
+	// Verify that the volumePlugin value is valid - either "Enabled" or "Disabled"
+	if instance.Spec.VolumePlugin != "Enabled" && instance.Spec.VolumePlugin != "Disabled" && instance.Spec.VolumePlugin != "" {
```
tmjd, line 319: I'm hoping we can move this validation into the CRD definition by defining a type for VolumePlugin with only Enabled and Disabled as the two valid values.

pkg/render/csi.go
```diff
@@ -0,0 +1,334 @@
...
+			Privileged: ptr.BoolToPtr(true),
+		},
+		Env: []corev1.EnvVar{
+			corev1.EnvVar{
+				Name:  "ADDRESS",
+				Value: "/csi/csi.sock",
+			},
+			corev1.EnvVar{
+				Name:  "DRIVER_REG_SOCK_PATH",
+				Value: "/var/lib/kubelet/plugins/csi.tigera.io/csi.sock",
```
tmjd, line 167: I'm wondering if we should be exposing this or any of the other paths that are below. Should we change VolumePlugin to be a struct that way we can expose any of these paths in the future if we need to without adding other fields at the same level?

hack/gen-versions/components.go
Josh-Tigera: That makes sense to be more consistent with the others.

pkg/controller/installation/validation.go
Josh-Tigera: Will do.

pkg/render/csi.go
Josh-Tigera: Do you mean rather than VolumePlugin simply being "Enabled" or "Disabled" at the API level the user can specify a series of kv pairs that would then be populated here?

tmjd: No we shouldn't simply expose kv pairs but I was thinking we would create a field for each path that the user would need to specify and those fields would be in an embedded in a struct so we might have something like:
```
spec:
  volumePlugin:
    sockPath: ...
    kubeletPath: ...
    registrationDir: ...
```
I think we'd also need a field to set Enable/Disable but I'm not sure what the name of the field should be.

Josh-Tigera: Ok I see what you mean, let me come up with a draft of this.

pkg/render/csi.go
```diff
@@ -0,0 +1,338 @@
...
+				MountPropagation: &mountPropagation,
+			},
+		},
+	}
+
+	// Construct "csi-node-driver-registrar" container
+	registrarContainer := corev1.Container{
+		Name:            CSIRegistrarContainerName,
+		Image:           c.csiRegistrarImage,
+		ImagePullPolicy: corev1.PullAlways,
```
mgleung, line 153: We might want to change this to pull if not available.

pkg/render/csi.go
```diff
@@ -0,0 +1,338 @@
...
+	}
+	return tolerations
+}
+
+func (c *csiComponent) csiContainers() []corev1.Container {
+	mountPropagation := corev1.MountPropagationBidirectional
+	csiContainer := corev1.Container{
+		Name:            CSIContainerName,
+		Image:           c.csiImage,
+		ImagePullPolicy: corev1.PullAlways,
```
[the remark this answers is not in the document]
Josh-Tigera: I just replicated the manifest from https://projectcalico.docs.tigera.io/manifests/csi-driver.yaml verbatim. If we want to deviate from that though it's no problem to me.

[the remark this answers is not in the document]
mgleung: shoot, that means that I hadn't cleaned up the manifest there. We should probably change both of them. Generally, `Always` is good for debugging but probably not what a user wants.

mgleung (approved):
LGTM

pkg/render/csi.go
```diff
@@ -0,0 +1,338 @@
...
+	return corev1.PodTemplateSpec{
+		ObjectMeta: templateMeta,
+		Spec:       templateSpec,
+	}
+}
+
+// csiDaemonset creates the daemonset necessary to enable the CSI driver
+func (c *csiComponent) csiDaemonset() *appsv1.DaemonSet {
+	dsLabels := map[string]string{
+		"app.kubernetes.io/name": CSIDaemonSetName,
```
tmjd, line 271: We don't need to specify this because the CreateOrUpdateOrDelete function will take care adding this, see [this for where it happens](https://github.com/tigera/operator/blob/93ea11c32e182aea91a49aeae56529895af8968e/pkg/controller/utils/component.go#L487)

pkg/render/csi.go
```diff
@@ -0,0 +1,338 @@
...
+		Namespace: CSIDaemonSetNamespace,
+		Labels:    dsLabels,
+	}
+
+	typeMeta := metav1.TypeMeta{
+		Kind:       "DaemonSet",
+		APIVersion: "apps/v1",
+	}
+
+	selector := &metav1.LabelSelector{
```
tmjd, line 284: Same as my other comment, [this](https://github.com/tigera/operator/blob/93ea11c32e182aea91a49aeae56529895af8968e/pkg/controller/utils/component.go#L487) will add the selector automatically.

tmjd (commented):
Looking good, just a couple things that aren't needed due to some backend magic that handles them for us.

pkg/render/csi.go
Josh-Tigera: I wasn't aware of this piece of infra, thanks for the pointer.

tmjd (approved):
LGTM

rene-dekker:
I believe this approach is not going to work for windows users of EE. We need a separate set of images for Windows nodes and Linux nodes.

I can help and create a repackaged image for the registrar for windows builds, since I am already making a build for amd64 with CGO on.

As for CSI @mgleung will probably work within his team to see how to best address the rest.

lmm:
> I believe this approach is not going to work for windows users of EE. We need a separate set of images for Windows nodes and Linux nodes.

The CSI components in the PR will work for Linux only due to `SupportedOSType()`. But you're right, this PR does not consider Windows support.

The current state of Windows support in the operator is a bit adhoc. We've added components as needed, e.g. the `calico-windows-upgrader` and `fluentd-windows`. There are plans to add Calico for Windows daemonset support (with HostProcess containers) to the operator. When we do that work, I think that will be a good time to revisit what supporting Windows nodes and Windows components in the operator will look like.

Based on that, I think we could consider leaving Windows CSI support out of the operator for now until we have an idea of how all this Windows stuff is supposed to hang together. WDYT?

PR #2087, opened 2022-07-19
 Add IPv6 Wireguard and IPv6 VXLAN custom MTU support

coutinhop:
## Description

Set IPv6 VXLAN and Wireguard MTU if IPv6 is going to be enabled.

Enable IPv6 in render/node_test.go in order to correctly test IPv6 VXLAN and Wireguard MTUs.

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/node.go
+++ b/pkg/render/node.go
@@ -1426,6 +1426,14 @@ func (c *nodeComponent) nodeEnvVars() []corev1.EnvVar {
 		if v4Method == "" && bgpEnabled(c.cfg.Installation) {
 			nodeEnv = append(nodeEnv, corev1.EnvVar{Name: "CALICO_ROUTER_ID", Value: "hash"})
 		}
+
+		// Set IPv6 VXLAN and Wireguard MTU
+		if mtu != nil {
+			vxlanMtuV6 := strconv.Itoa(int(*mtu))
+			wireguardMtuV6 := strconv.Itoa(int(*mtu))
+			nodeEnv = append(nodeEnv, corev1.EnvVar{Name: "FELIX_VXLANMTUV6", Value: vxlanMtuV6})
+			nodeEnv = append(nodeEnv, corev1.EnvVar{Name: "FELIX_WIREGUARDMTUV6", Value: wireguardMtuV6})
+		}
 	} else {
 		// IPv6 Auto-detection is disabled.
 		nodeEnv = append(nodeEnv, corev1.EnvVar{Name: "IP6", Value: "none"})
--- a/pkg/render/node_test.go
+++ b/pkg/render/node_test.go
@@ -16,6 +16,7 @@ package render_test
 
 import (
 	"fmt"
+	"net"
 	"strings"
 
 	. "github.com/onsi/ginkgo"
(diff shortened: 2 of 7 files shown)


pkg/crds/calico/crd.projectcalico.org_felixconfigurations.yaml
```diff
@@ -306,7 +306,6 @@ spec:
                   are auto-detected.
                 type: string
               floatingIPs:
-                default: Disabled
```
coutinhop, line 309: I'm not entirely sure where this change was picked up from, it doesn't seem at all related to the wireguard changes

lmm: That was from master it looks like: https://github.com/tigera/operator/commit/82459ad7f232e5bc1b195e06564d0e3c82ae3ade#diff-b6824b8776988360d61c4a8e3c38552c52037ef281cac3cfdff0ed4882e70a27L308-L311

pkg/render/node.go
```diff
@@ -1349,9 +1349,13 @@ func (c *nodeComponent) nodeEnvVars() []corev1.EnvVar {
 	mtu := getMTU(c.cfg.Installation)
 	if mtu != nil {
 		vxlanMtu := strconv.Itoa(int(*mtu))
+		vxlanMtuV6 := strconv.Itoa(int(*mtu))
 		wireguardMtu := strconv.Itoa(int(*mtu))
+		wireguardMtuV6 := strconv.Itoa(int(*mtu))
 		nodeEnv = append(nodeEnv, corev1.EnvVar{Name: "FELIX_VXLANMTU", Value: vxlanMtu})
+		nodeEnv = append(nodeEnv, corev1.EnvVar{Name: "FELIX_VXLANMTUV6", Value: vxlanMtuV6})
 		nodeEnv = append(nodeEnv, corev1.EnvVar{Name: "FELIX_WIREGUARDMTU", Value: wireguardMtu})
+		nodeEnv = append(nodeEnv, corev1.EnvVar{Name: "FELIX_WIREGUARDMTUV6", Value: wireguardMtuV6})
```
lmm, line 1358: Do we want to set the *V6 env vars for IPv4 mode as well? In other places of this node rendering file, we check if IPv6 is enabled before setting the V6 env vars.

I don't know enough about wg and V6 but I'm wondering if there could be possible side effects in Felix where it detects `FELIX_WIREGUARDMTUV6` is set and does something related to V6 wireguard.

pkg/crds/calico/crd.projectcalico.org_felixconfigurations.yaml
coutinhop: Got it, thanks @lmm!

pkg/render/node.go
coutinhop: Good point, I'll only add those when `FELIX_IPV6SUPPORT` is `true`. Thanks!

pkg/render/node_test.go
```diff
@@ -76,8 +76,9 @@ var _ = Describe("Node rendering tests", func() {
 			},
 			CalicoNetwork: &operatorv1.CalicoNetworkSpec{
 				BGP:                        &bgpEnabled,
-				IPPools:                    []operatorv1.IPPool{{CIDR: "192.168.1.0/16"}},
```
tmjd, line 79: I think we should keep the general default tests to be IPv4 only. I think we should add some additional tests that include an IPv6 pool and include mtu setting in those tests also, those new tests should only test for the specific differences we expect (in other words to include checking every Env var and setting that we do in the default tests).

caseydavenport: Yeah, I think having separate v4, v6, and dual-stack tests would be the right move

caseydavenport (commented):
LGTM, but think we should separate the tests out a bit - maybe could do it with a loop to avoid code duplication, or maybe could just copy some tests :grin:

pkg/render/node_test.go
coutinhop: Agreed, will do that! Thanks!

coutinhop: Just force-pushed a commit extending the whole node_test.go to run for IPv4, IPv6 and dual-stack. Please let me know your thoughts. Thanks!

caseydavenport (approved):

PR #2090, opened 2022-07-20
allow-tigera reconciliation: follow-ups

pasanw:
## Description
Addresses the 'immediate follow-up' items from #1970.
The bulk of the changes come from removing Reconcile waits for policy and tier watches, this is captured in it's own commit.

Items addressed:
* Removed use of reconcilability util function in favour of direct Tier Get()
* Ensured that controllers which impact policy reconcilability render policy last
* Removed waits for policy and tier watches
* Removed API server waits added in allow-tigera PR
 
## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/apiserver/apiserver_controller.go
+++ b/pkg/controller/apiserver/apiserver_controller.go
@@ -19,6 +19,9 @@ import (
 	"fmt"
 	"time"
 
+	v3 "github.com/tigera/api/pkg/apis/projectcalico/v3"
+	"k8s.io/apimachinery/pkg/api/meta"
+
 	"github.com/tigera/operator/pkg/render/common/networkpolicy"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes"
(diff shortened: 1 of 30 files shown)


caseydavenport (approved):
A couple of minor comments but overall LGTM - thanks for this!

pkg/controller/clusterconnection/clusterconnection_controller.go
```diff
@@ -254,88 +244,85 @@ func (r *ReconcileConnection) Reconcile(ctx context.Context, request reconcile.R
...
-					return reconcile.Result{RequeueAfter: 10 * time.Second}, nil
-				}
-				r.status.SetDegraded("Error querying license", err.Error())
-				return reconcile.Result{RequeueAfter: 10 * time.Second}, nil
-			}
-
-			if !utils.IsFeatureActive(license, common.EgressAccessControlFeature) {
-				r.status.SetDegraded("Feature is not active", "License does not support feature: egress-access-control")
+	if egressAccessControlFeatureRequired {
+		license, err := utils.FetchLicenseKey(ctx, r.Client)
```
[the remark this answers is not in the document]
pasanw: The license check was not present prior to the allow-tigera work and was added for the feature check, so I was thinking it would be best to limit the scope of this check. We also need to be careful with the timing of a license check, as this controller is initially a dependency for the license being deployed

I've reorganized to make these considerations a bit more clear, but the logic should be equivalent

pkg/controller/installation/core_controller.go
```diff
@@ -1000,25 +999,18 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
...
-			if !r.tierWatchReady.IsReady() {
-				r.status.SetDegraded("Waiting for Tier watch to be established", "")
-				return reconcile.Result{RequeueAfter: 10 * time.Second}, nil
-			}
-
-			if !r.policyWatchesReady.IsReady() {
-				r.status.SetDegraded("Waiting for NetworkPolicy watches to be established", "")
-				return reconcile.Result{RequeueAfter: 10 * time.Second}, nil
+		// Ensure the allow-tigera tier exists, before rendering any network policies within it.
+		if err := r.client.Get(ctx, client.ObjectKey{Name: networkpolicy.TigeraComponentTierName}, &v3.Tier{}); err != nil {
```
caseydavenport, line 1003: Do we need / want a tier watch guarding this Get() call?

pkg/controller/apiserver/apiserver_controller.go
```diff
@@ -306,25 +305,18 @@ func (r *ReconcileAPIServer) Reconcile(ctx context.Context, request reconcile.Re
...
-
-			if !r.tierWatchReady.IsReady() {
-				r.status.SetDegraded("Waiting for Tier watch to be established", "")
-				return reconcile.Result{RequeueAfter: 10 * time.Second}, nil
-			}
-
-			if !r.policyWatchesReady.IsReady() {
-				r.status.SetDegraded("Waiting for NetworkPolicy watches to be established", "")
-				return reconcile.Result{RequeueAfter: 10 * time.Second}, nil
+		// Ensure the allow-tigera tier exists, before rendering any network policies within it.
```
caseydavenport, line 308: Same here - do we want to guard this Get() call with a tier watch check?

pkg/controller/installation/core_controller.go
caseydavenport: Basically just conditionally execute this whole block based on whether the tier watch is established? 


```go
includeNetworkPolicy = false
if tierWatch.Ready() {
  if err := r.client.Get(tier); err != nil {
    // Handle error
  } else {
    // Tier exists.
    includePolicy = true
  }
}

caseydavenport (approved):

PR #2092, opened 2022-07-21
Add AssignIpv[46] in CalicoNetworkSpec

sknat:
## Description

This patchs adds two additional parameters to the CalicoNetworkSpec
allowing to force the value of assign_ipv4 / assign_ipv6 in the CNI
configuration. This is useful when the ippools are not created at
installation time but later on during provisionning.

If omitted, the previous behaviour will be preserved (deriving the
values from the presence of v4/v6 pools in the default ippools)

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/api/v1/installation_types.go
+++ b/api/v1/installation_types.go
@@ -326,6 +326,16 @@ type CalicoNetworkSpec struct {
 	// +optional
 	IPPools []IPPool `json:"ipPools,omitempty"`
 
+	// AssignIpv4 configure whether ipv4 pools are enabled or disabled
+	// If omitted, its value will be derived from the presence of ipv4 ippools in IPPools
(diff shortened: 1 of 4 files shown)


tmjd:
There was some conversation about this in [CalicoUsers slack](https://calicousers.slack.com/archives/G01HDE9CDPU/p1658422169973409).
I'd say the summary is that this isn't the direction we would want to go. Instead of adding fields to the API, it would be more desirable to auto-detect the need for assign IPv[46] from the existing IPPools.

Thank you @sknat for submitting this to trigger some discussion about this issue.
I'm going to close this PR, we can re-open it in the future if we want to move this direction.

PR #2094, opened 2022-07-21
Add option for Persistent Storage for AD API

freecaykes:
feat(ids): set api for AD storageclass and render PVs PVCs

feat(ids_controller): add watcher for storageclasses in ids + config ad deployment to set storageclass volume

feat(ids): validation for AD storageclass name

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/api/v1/intrusiondetection_types.go
+++ b/api/v1/intrusiondetection_types.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2020 Tigera, Inc. All rights reserved.
+// Copyright (c) 2020, 2022 Tigera, Inc. All rights reserved.
 /*
 
 Licensed under the Apache License, Version 2.0 (the "License");
@@ -27,7 +27,39 @@ type IntrusionDetectionSpec struct {
 	// Only DeepPacketInspection is supported for this spec.
 	// +optional
 	ComponentResources []IntrusionDetectionComponentResource `json:"componentResources,omitempty"`
+
+	// AnomalyDetectionSpec provides configuration for running AnomalyDetection Component within
+	// IntrusionDetection
+	AnomalyDetectionSpec *AnomalyDetectionSpec `json:"anomalyDetectionSpec,omitempty"`
 }
+
+type AnomalyDetectionSpec struct {
+	// StorageType Sets the type of storage to use for storing Anomaly Detection Models. By default it will use the ephemeral
+	// emptyDir on the node AD will be deployed to. This field sets precedent for all AD Storage related fields and will only
+	// take into effect on Management and Standalone clusters.
+	// default: Ephemeral
+	StorageType StorageType `json:"storageType,omitempty"`
+
+	// StorageClassName will populate the PersistentVolumeClaim.StorageClassName that is used to provision disks for the
+	// Anomaly Detection API pod for model storage. The StorageClassName should only be modified when no LogStorage is currently
+	// active. We recommend choosing a storage class dedicated to Tigera LogStorage only. Otherwise, model retention
+	// cannot be guaranteed during upgrades. See https://docs.tigera.io/maintenance/upgrading for up-to-date instructions.
+	// Default: tigera-anomaly-detection-storage
+	// +optional
+	StorageClassName string `json:"storageClassName,omitempty"`
+}
+
+// StorageType is sets the type of storage to be used for the specified component.
+type StorageType string
+
+const (
+	// Ephermeal storage type sets the ephermeal emptyDir() to be used by the component. Data created in this storage type will
+	// follow the Pod's lifetime and get created and deleted along with the Pod.
+	EphermealStorageType StorageType = "Ephermeal"
+	// Persistent storage type sets the will attempt to query for the StorageClass specified by StorageClassName created for
+	// the component. It will utilize the provided StorageClass for the componenet's Volume.
+	PersistentStorageType StorageType = "Persistent"
+)
 
 // IntrusionDetectionStatus defines the observed state of Tigera intrusion detection capabilities.
 type IntrusionDetectionStatus struct {
--- a/pkg/controller/intrusiondetection/intrusiondetection_controller.go
+++ b/pkg/controller/intrusiondetection/intrusiondetection_controller.go
@@ -41,6 +41,7 @@ import (
 
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
+	storagev1 "k8s.io/api/storage/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -75,7 +76,7 @@ func Add(mgr manager.Manager, opts options.AddOptions) error {
 	// Create a new controller
 	controller, err := controller.New("intrusiondetection-controller", mgr, controller.Options{Reconciler: reconcile.Reconciler(reconciler)})
 	if err != nil {
-		return fmt.Errorf("Failed to create intrusiondetection-controller: %v", err)
+		return fmt.Errorf("failed to create intrusiondetection-controller: %v", err)
 	}
 
 	k8sClient, err := kubernetes.NewForConfig(mgr.GetConfig())
@@ -149,6 +150,11 @@ func add(mgr manager.Manager, c controller.Controller) error {
 
 	if err = imageset.AddImageSetWatch(c); err != nil {
 		return fmt.Errorf("intrusiondetection-controller failed to watch ImageSet: %w", err)
+	}
+
+	// Watch for changes in storage classes to queue changes if new storage classes may be made available for AD API.
+	if err = c.Watch(&source.Kind{Type: &storagev1.StorageClass{}}, &handler.EnqueueRequestForObject{}); err != nil {
+		return fmt.Errorf("intrusiondetection-controller failed to watch StorageClass resource: %w", err)
 	}
 
 	if err = utils.AddAPIServerWatch(c); err != nil {
@@ -308,6 +314,29 @@ func (r *ReconcileIntrusionDetection) Reconcile(ctx context.Context, request rec
 		return reconcile.Result{}, err
 	}
 
+	// Query for StorageClass for AD API if provided
+	if instance.Spec.AnomalyDetectionSpec.StorageType == operatorv1.PersistentStorageType {
+		// validate to degrade early if the storage class name is not valid
+		if err = utils.ValidateResourceNameIsQualified(instance.Spec.AnomalyDetectionSpec.StorageClassName); err != nil {
+			errMessage := "invalid AD Storage Class name provided"
+			log.Error(err, errMessage)
+			r.status.SetDegraded("Error retrieving pull secrets", err.Error())
+			return reconcile.Result{}, err
+		}
+
+		if err = r.client.Get(ctx, client.ObjectKey{Name: instance.Spec.AnomalyDetectionSpec.StorageClassName}, &storagev1.StorageClass{}); err != nil {
+			if errors.IsNotFound(err) {
+				log.Error(err, "Anomaly Detection Storage Class not found")
+				r.status.SetDegraded("Failed to get storage class for anomaly detection", err.Error())
+				return reconcile.Result{}, err
+			} else {
+				reqLogger.Error(err, "Failed to query for Anomaly Detection Storage Class")
+				r.status.SetDegraded("Failed to get storage class for anomaly detection", err.Error())
+				return reconcile.Result{}, err
+			}
+		}
+	}
+
 	// Query for the LogCollector instance. We need this to determine whether or not
 	// to forward IDS event logs. Since this is optional, we don't need to degrade or
 	// change status if LogCollector is not found.
@@ -420,6 +449,7 @@ func (r *ReconcileIntrusionDetection) Reconcile(ctx context.Context, request rec
 	// Render the desired objects from the CRD and create or update them.
 	hasNoLicense := !utils.IsFeatureActive(license, common.ThreatDefenseFeature)
 	intrusionDetectionCfg := &render.IntrusionDetectionConfiguration{
+		IntrusionDetction:     instance,
 		LogCollector:          lc,
 		ESSecrets:             esSecrets,
 		Installation:          network,
@@ -529,7 +559,8 @@ func (r *ReconcileIntrusionDetection) Reconcile(ctx context.Context, request rec
 	return reconcile.Result{}, nil
 }
 
-// setDefaultsOnIntrusionDetection updates the IntrusionDetection resource with defaults if ComponentResources is not populated.
+// setDefaultsOnIntrusionDetection updates the IntrusionDetection resource with defaults if
+// ComponentResources and AnomalyDetectionSpec are not populated.
 func (r *ReconcileIntrusionDetection) setDefaultsOnIntrusionDetection(ctx context.Context, ids *operatorv1.IntrusionDetection) error {
 	if ids.Spec.ComponentResources == nil {
 		ids.Spec.ComponentResources = []operatorv1.IntrusionDetectionComponentResource{
@@ -546,6 +577,16 @@ func (r *ReconcileIntrusionDetection) setDefaultsOnIntrusionDetection(ctx contex
 					},
 				},
 			},
+		}
+
+		if ids.Spec.AnomalyDetectionSpec == nil {
+			ids.Spec.AnomalyDetectionSpec = &operatorv1.AnomalyDetectionSpec{
+				StorageType: operatorv1.EphermealStorageType,
+			}
+		} else {
+			if ids.Spec.AnomalyDetectionSpec.StorageType == operatorv1.PersistentStorageType && len(ids.Spec.AnomalyDetectionSpec.StorageClassName) == 0 {
+				ids.Spec.AnomalyDetectionSpec.StorageClassName = render.DefaultADStorageClassName
+			}
 		}
 
 		if err := r.client.Update(ctx, ids); err != nil {
--- a/pkg/controller/utils/utils_test.go
+++ b/pkg/controller/utils/utils_test.go
@@ -183,6 +183,25 @@ var _ = Describe("Utils APIServer type tests", func() {
 	)
 })
 
+var _ = Describe("ValidateResourceNameIsQualified", func() {
+
+	It("returns nil for a compliant kubernetes name.", func() {
+		qualifiedName := "proper-resource-name"
+
+		err := ValidateResourceNameIsQualified(qualifiedName)
+
+		Expect(err).To(BeNil())
+	})
+
+	It("returns nil for an invalid resource name", func() {
+		invalidName := "inproper_resource_name"
+
+		err := ValidateResourceNameIsQualified(invalidName)
+
+		Expect(err).ToNot(BeNil())
+	})
+})
+
 type fakeClient struct {
 	discovery discovery.DiscoveryInterface
 	kubernetes.Interface
--- a/pkg/render/intrusion_detection.go
+++ b/pkg/render/intrusion_detection.go
@@ -28,6 +28,7 @@ import (
 
 	policyv1beta1 "k8s.io/api/policy/v1beta1"
 	rbacv1 "k8s.io/api/rbac/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -60,14 +61,17 @@ const (
 	IntrusionDetectionControllerPolicyName = networkpolicy.TigeraComponentPolicyPrefix + IntrusionDetectionControllerName
 	IntrusionDetectionInstallerPolicyName  = networkpolicy.TigeraComponentPolicyPrefix + "intrusion-detection-elastic"
 
-	ADJobPodTemplateBaseName      = "tigera.io.detectors"
-	adDetectorPrefixName          = "tigera.io.detector."
-	adDetectorName                = "anomaly-detectors"
-	ADDetectorPolicyName          = networkpolicy.TigeraComponentPolicyPrefix + adDetectorName
-	adDetectionJobsDefaultPeriod  = 15 * time.Minute
-	ADResourceGroup               = "detectors.tigera.io"
-	ADDetectorsModelResourceName  = "models"
-	ADLogTypeMetaDataResourceName = "metadata"
+	DefaultADStorageClassName              = "tigera-anomaly-detection"
+	DefaultAnomalyDetectionPVRequestSizeGi = 10
+	adAPIVolumePath                        = "/storage"
+	ADJobPodTemplateBaseName               = "tigera.io.detectors"
+	adDetectorPrefixName                   = "tigera.io.detector."
+	adDetectorName                         = "anomaly-detectors"
+	ADDetectorPolicyName                   = networkpolicy.TigeraComponentPolicyPrefix + adDetectorName
+	adDetectionJobsDefaultPeriod           = 15 * time.Minute
+	ADResourceGroup                        = "detectors.tigera.io"
+	ADDetectorsModelResourceName           = "models"
+	ADLogTypeMetaDataResourceName          = "metadata"
 
 	ADAPIObjectName          = "anomaly-detection-api"
 	ADAPIObjectPortName      = "anomaly-detection-api-https"
@@ -97,6 +101,7 @@ func IntrusionDetection(cfg *IntrusionDetectionConfiguration) Component {
 
 // IntrusionDetectionConfiguration contains all the config information needed to render the component.
 type IntrusionDetectionConfiguration struct {
+	IntrusionDetction     *operatorv1.IntrusionDetection
 	LogCollector          *operatorv1.LogCollector
 	ESSecrets             []*corev1.Secret
 	Installation          *operatorv1.InstallationSpec
@@ -201,6 +206,14 @@ func (c *intrusionDetectionComponent) Objects() ([]client.Object, []client.Objec
 			c.adAPIAccessClusterRole(),
 			c.adAPIAccessRoleBinding(),
 		)
+
+		if c.cfg.IntrusionDetction.Spec.AnomalyDetectionSpec.StorageType == operatorv1.PersistentStorageType {
+			objs = append(objs,
+				c.adPersistentVolume(),
+				c.adPersistentVolumeClaim(),
+			)
+		}
+
 		objs = append(objs,
 			c.adAPIService(),
 			c.adAPIDeployment(),
@@ -1279,7 +1292,7 @@ func (c *intrusionDetectionComponent) adAPIService() *corev1.Service {
 		},
 		Spec: corev1.ServiceSpec{
 			Selector: map[string]string{
-				"k8s-app": ADAPIObjectName,
+				AppLabelName: ADAPIObjectName,
 			},
 			Ports: []corev1.ServicePort{
 				{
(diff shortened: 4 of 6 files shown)


api/v1/intrusiondetection_types.go
```diff
@@ -27,8 +27,40 @@ type IntrusionDetectionSpec struct {
...
+	StorageClassName string `json:"storageClassName,omitempty"`
 }
 
+// StorageType is sets the type of storage to be used for the specified component.
+type StorageType string
+
+const (
+	// Ephermeal storage type sets the ephermeal emptyDir() to be used by the component. Data created in this storage type will
+	// follow the Pod's lifetime and get created and deleted along with the Pod.
+	EphermealStorageType StorageType = "Ephermeal"
```
rene-dekker, line 58: Typo: ephemeral

api/v1/intrusiondetection_types.go
```diff
@@ -27,8 +27,40 @@ type IntrusionDetectionSpec struct {
...
 	ComponentResources []IntrusionDetectionComponentResource `json:"componentResources,omitempty"`
+
+	// AnomalyDetectionSpec provides configuration for running AnomalyDetection Component within
+	// IntrusionDetection
+	AnomalyDetectionSpec *AnomalyDetectionSpec `json:"anomalyDetectionSpec,omitempty"`
+}
+
+type AnomalyDetectionSpec struct {
+	// StorageType Sets the type of storage to use for storing Anomaly Detection Models. By default it will use the ephemeral
+	// emptyDir on the node AD will be deployed to. This field sets precedent for all AD Storage related fields and will only
```
rene-dekker, line 38: > Anomaly Detection Models
models (lowercase)

> This field sets precedent for all AD Storage related fields 
I think we better refrain from saying this, since other Operaror overrides are making its way into the operator.

api/v1/intrusiondetection_types.go
```diff
@@ -27,8 +27,40 @@ type IntrusionDetectionSpec struct {
...
+	// take into effect on Management and Standalone clusters.
+	// default: Ephemeral
+	StorageType StorageType `json:"storageType,omitempty"`
+
+	// StorageClassName will populate the PersistentVolumeClaim.StorageClassName that is used to provision disks for the
+	// Anomaly Detection API pod for model storage. The StorageClassName should only be modified when no LogStorage is currently
```
rene-dekker, line 44: > LogStorage

I think you erroneously copied this from another spec. There are multiple mentions.

api/v1/intrusiondetection_types.go
```diff
@@ -27,8 +27,40 @@ type IntrusionDetectionSpec struct {
...
 }
 
+// StorageType is sets the type of storage to be used for the specified component.
+type StorageType string
+
+const (
+	// Ephermeal storage type sets the ephermeal emptyDir() to be used by the component. Data created in this storage type will
+	// follow the Pod's lifetime and get created and deleted along with the Pod.
+	EphermealStorageType StorageType = "Ephermeal"
+	// Persistent storage type sets the will attempt to query for the StorageClass specified by StorageClassName created for
```
rene-dekker, line 59: I think there is something wrong with this sentence, I don't understand it.

> componenet's

typo

api/v1/intrusiondetection_types.go
rene-dekker: I think you still need  `+kubebuilder:validation:Enum` to ensure that the CRD validates the valid enum options

pkg/render/intrusion_detection.go
```diff
@@ -201,9 +207,22 @@ func (c *intrusionDetectionComponent) Objects() ([]client.Object, []client.Objec
 			c.adAPIAccessClusterRole(),
 			c.adAPIAccessRoleBinding(),
 		)
+
+		configureADStorage := c.cfg.IntrusionDetction != nil &&
+			c.cfg.IntrusionDetction.Spec.AnomalyDetectionSpec != nil &&
+			c.cfg.IntrusionDetction.Spec.AnomalyDetectionSpec.StorageType == operatorv1.PersistentStorageType
+
+		if configureADStorage {
+			// ignore all fields if it's set to using default epehrmeal storage
```
rene-dekker, line 216: > epehrmeal

typo

pkg/render/intrusion_detection.go
```diff
@@ -60,14 +61,18 @@ const (
...
-	ADJobPodTemplateBaseName      = "tigera.io.detectors"
-	adDetectorPrefixName          = "tigera.io.detector."
-	adDetectorName                = "anomaly-detectors"
-	ADDetectorPolicyName          = networkpolicy.TigeraComponentPolicyPrefix + adDetectorName
-	adDetectionJobsDefaultPeriod  = 15 * time.Minute
-	ADResourceGroup               = "detectors.tigera.io"
-	ADDetectorsModelResourceName  = "models"
-	ADLogTypeMetaDataResourceName = "metadata"
+	DefaultADStorageClassName              = "tigera-anomaly-detection"
+	DefaultAnomalyDetectionPVRequestSizeGi = 10
```
rene-dekker, line 65: Maybe you can simply change it to "10Gi", because all 4 times it is used, you string format to get that value.

api/v1/intrusiondetection_types.go
```diff
@@ -27,8 +27,42 @@ type IntrusionDetectionSpec struct {
 	// Only DeepPacketInspection is supported for this spec.
 	// +optional
 	ComponentResources []IntrusionDetectionComponentResource `json:"componentResources,omitempty"`
+
+	// AnomalyDetectionSpec provides configuration for running AnomalyDetectionSpec Component within
+	// IntrusionDetection
+	// +optional
+	AnomalyDetectionSpec AnomalyDetectionSpec `json:"anomalyDetection,omitempty"`
```
rene-dekker, line 34: `AnomalyDetection AnomalyDetectionSpec `json:"anomalyDetection,omitempty"``

api/v1/intrusiondetection_types.go
```diff
@@ -27,8 +27,42 @@ type IntrusionDetectionSpec struct {
...
+}
+
+type AnomalyDetectionSpec struct {
+	// StorageType Sets the type of storage to use for storing Anomaly Detection Models. By default it will use the ephemeral
```
rene-dekker, line 38: Sets -> sets

> AD

I think it is best to fully write out abbreviations in our specs, because the CRD comments are part of the user experience.

api/v1/intrusiondetection_types.go
```diff
@@ -27,8 +27,42 @@ type IntrusionDetectionSpec struct {
...
+	// default: Ephemeral
+	// +optional
+	// +kubebuilder:validation:Enum=Ephemeral;Persistent
+	StorageType StorageType `json:"storageType,omitempty"`
+
+	// StorageClassName will populate the PersistentVolumeClaim.StorageClassName that is used to provision disks for the
+	// Anomaly Detection API pod for model storage. The StorageClassName should only be modified when no StorageClass is currently
+	// active. We recommend choosing a storage class dedicated to AnomalyDetection only. Otherwise, model retention
+	// cannot be guaranteed during upgrades. See https://docs.tigera.io/maintenance/upgrading for up-to-date instructions.
+	// Default: tigera-anomaly-detection-storage
```
rene-dekker, line 49: I believe we renamed the default to "tigera-anomaly-detection"

api/v1/intrusiondetection_types.go
```diff
@@ -27,8 +27,42 @@ type IntrusionDetectionSpec struct {
...
+	// +optional
+	StorageClassName string `json:"storageClassName,omitempty"`
 }
 
+// StorageType is sets the type of storage to be used for the specified component.
```
rene-dekker, line 54: typo:

> is sets

api/v1/intrusiondetection_types.go
```diff
@@ -27,8 +27,42 @@ type IntrusionDetectionSpec struct {
...
+type StorageType string
+
+const (
+	// ephemeral storage type sets the ephemeral emptyDir() to be used by the component. Data created in this storage type will
+	// follow the Pod's lifetime and get created and deleted along with the Pod.
+	EphemeralStorageType StorageType = "Ephemeral"
+	// Persistent storage type sets the will attempt to query for the StorageClass specified by StorageClassName created for
```
rene-dekker, line 61: PersistentStorageType mounts a PersistentVolume of the provided StorageClassName to Anomaly Detection pods in order to store data.

pkg/controller/intrusiondetection/intrusiondetection_controller.go
```diff
@@ -310,6 +316,29 @@ func (r *ReconcileIntrusionDetection) Reconcile(ctx context.Context, request rec
 		return reconcile.Result{}, err
 	}
 
+	// Query for StorageClass for AD API if provided
```
rene-dekker, line 319: I'm just wondering if we want to perhaps throw an error if the user configured Ephemeral storage while filling in the storageclassname.

pkg/controller/intrusiondetection/intrusiondetection_controller.go
```diff
@@ -550,10 +581,21 @@ func (r *ReconcileIntrusionDetection) setDefaultsOnIntrusionDetection(ctx contex
 			},
 		}
 
-		if err := r.client.Update(ctx, ids); err != nil {
-			return err
+	}
+
+	if ids.Spec.AnomalyDetectionSpec == (operatorv1.AnomalyDetectionSpec{}) {
```
rene-dekker, line 586: I think this can be rewritten to:
```
	if len(ids.Spec.AnomalyDetectionSpec.StorageType) == 0 {
		ids.Spec.AnomalyDetectionSpec.StorageType = operatorv1.EphemeralStorageType
	}
```

pkg/controller/intrusiondetection/intrusiondetection_controller.go
```diff
@@ -531,7 +561,8 @@ func (r *ReconcileIntrusionDetection) Reconcile(ctx context.Context, request rec
 	return reconcile.Result{}, nil
 }
 
-// setDefaultsOnIntrusionDetection updates the IntrusionDetection resource with defaults if ComponentResources is not populated.
+// setDefaultsOnIntrusionDetection updates the IntrusionDetection resource with defaults if
+// ComponentResources and AnomalyDetectionSpec are not populated.
 func (r *ReconcileIntrusionDetection) setDefaultsOnIntrusionDetection(ctx context.Context, ids *operatorv1.IntrusionDetection) error {
```
rene-dekker, line 566: nit: I believe similar functions in other controllers are named "fillDefaults(...)"  and the updating is then done from the reconcile() func.
For example: `logstorage_controller.go`

pkg/controller/utils/utils_test.go
```diff
@@ -183,6 +183,25 @@ var _ = Describe("Utils APIServer type tests", func() {
...
+	It("returns nil for a compliant kubernetes name.", func() {
+		qualifiedName := "proper-resource-name"
+
+		err := ValidateResourceNameIsQualified(qualifiedName)
+
+		Expect(err).To(BeNil())
+	})
+
+	It("returns nil for an invalid resource name", func() {
+		invalidName := "inproper_resource_name"
```
rene-dekker, line 197: > inproper

improper* is the correct spelling. Unless the pun was intended ;-)

pkg/render/intrusion_detection.go
```diff
@@ -97,6 +103,7 @@ func IntrusionDetection(cfg *IntrusionDetectionConfiguration) Component {
 
 // IntrusionDetectionConfiguration contains all the config information needed to render the component.
 type IntrusionDetectionConfiguration struct {
+	IntrusionDetction     *operatorv1.IntrusionDetection
```
rene-dekker, line 106: There is a typo in intrusiondetection.

Moreover, it is not possible for this value to be nil, because the controller would not get to this point. Perhaps change it back from a pointer to the value, because there are nil checks throughout this file that are unnecessary.

pkg/render/intrusion_detection.go
```diff
@@ -1297,9 +1320,48 @@ func (c *intrusionDetectionComponent) adAPIService() *corev1.Service {
 	}
 }
 
-func (c *intrusionDetectionComponent) adAPIDeployment() *appsv1.Deployment {
-	adAPIStorageVolumePath := "/storage"
-	adAPIStorageVolumeName := "volume-storage"
+func (c *intrusionDetectionComponent) adPersistentVolumeClaim(configureADStorage bool) *corev1.PersistentVolumeClaim {
```
rene-dekker, line 1323: Does it really matter what the value of the boolean is here? Outside of this func you delete the spec if the value is false, and otherwise you create it.

pkg/render/intrusion_detection.go
```diff
@@ -1297,9 +1320,48 @@ func (c *intrusionDetectionComponent) adAPIService() *corev1.Service {
...
+		},
+		Spec: pvSpec,
+	}
+
+	return &adPVC
+}
+
+func (c *intrusionDetectionComponent) adAPIDeployment(configureADStorage bool) *appsv1.Deployment {
+
+	adModelVolumeSource := corev1.VolumeSource{
```
rene-dekker, line 1352: This code could be more efficient:
```
var adModelVolumeSource  corev1.VolumeSource
if { 
...
} else {
...
}
```

pkg/controller/intrusiondetection/intrusiondetection_controller.go
freecaykes: was relying on the `StorageType` field to fully dictate whether to query for a StorageClass or not, which ignore other Storage related fields

Pros of **not** throwing error:
- does not block components from being deployed
Cons:
- users might  be unaware they have not set the StorageType properly (though it will be mentioned in the docs StorageType needs to be set to Persistent to query for the StorageClass) and is using Ephemeral storage


Pros of  throwing error:
- AnomalyDetection needs to be fully configured, gives users more assurance that the StorageClass is being used
Cons:
- blocks the other components from deploying

@rene-dekker let me know what you  think

rene-dekker: It would be a shame if the user is under the assumption that data is persistent, while it is not. That is why I thought an error would be helpful.
I am fine either way. Perhaps others have an opinion on this as well.

pkg/render/intrusion_detection.go
freecaykes: it matters only during a creation (ie. if it is an object that needs to be added in the `Objects()` method), since it needs to select whether to use `emptyDir()` or  set the volume to refer to the StorageClass

for deletion it would always get the value of false

pkg/render/intrusion_detection.go
```diff
@@ -201,9 +210,22 @@ func (c *intrusionDetectionComponent) Objects() ([]client.Object, []client.Objec
 			c.adAPIAccessClusterRole(),
 			c.adAPIAccessRoleBinding(),
 		)
+
+		configureADStorage := c.cfg.IntrusionDetection.Spec.AnomalyDetection.StorageType == operatorv1.PersistentStorageType
+
+		if configureADStorage {
+			// ignore all fields if it's set to using default ephermeal storage
```
rene-dekker, line 217: This comment seems out of place (and contains a typo)

rene-dekker (approved):
lgtm

pkg/render/intrusion_detection.go
freecaykes: https://github.com/tigera/operator/pull/2094/commits/66c82b2c34b018e5472ab22a1ba6ad27b96bcc62

pkg/render/intrusion_detection.go
freecaykes: ended up removing this, ^ invalid comment now, was thinking of the deployment

pkg/controller/intrusiondetection/intrusiondetection_controller.go
caseydavenport: > I'm just wondering if we want to perhaps throw an error if the user configured Ephemeral storage while filling in the storageclassname.

Yes, absolutely. We should validate against configuration that doesn't make it clear what the user's intent was.

Generally we should:

- Default any fields that aren't set.
- Check that the resulting object is valid and "makes sense" 

In this case, setting it to "Ephemeral" and also specifying a storage class is not self-consistent, so we should reject it and refuse to take any action until the user clarifies their intent.

pkg/controller/intrusiondetection/intrusiondetection_controller.go
```diff
@@ -550,9 +594,20 @@ func (r *ReconcileIntrusionDetection) setDefaultsOnIntrusionDetection(ctx contex
 			},
 		}
 
-		if err := r.client.Update(ctx, ids); err != nil {
-			return err
+	}
+
+	if len(ids.Spec.AnomalyDetection.StorageType) == 0 {
```
caseydavenport, line 599: This can result in an invalid configuration. e.g., if the user gives:

```
storageType: ""
storageClassName: my-storage-class
```

We'll default it to:

```
storageType: Ephemeral
storageClassName: my-storage-class
```

Which is not a valid config. I think we need to default this field conditionally based on whether a storage class name is given as well.

api/v1/intrusiondetection_types.go
```diff
@@ -27,8 +27,41 @@ type IntrusionDetectionSpec struct {
 	// Only DeepPacketInspection is supported for this spec.
 	// +optional
 	ComponentResources []IntrusionDetectionComponentResource `json:"componentResources,omitempty"`
+
+	// AnomalyDetection provides configuration for running AnomalyDetection Component within
+	// IntrusionDetection
+	// +optional
+	AnomalyDetection AnomalyDetectionSpec `json:"anomalyDetection,omitempty"`
```
caseydavenport, line 34: If this is an optional field, it needs to be a pointer.

caseydavenport: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required

> Optional fields have the following properties:
> - They have the +optional comment tag in Go.
> - They are a pointer type in the Go definition (e.g. AwesomeFlag *SomeFlag) or have a built-in nil value (e.g. maps and slices).
> - The API server should allow POSTing and PUTing a resource with this field unset.

pkg/render/intrusion_detection.go
```diff
@@ -97,15 +103,20 @@ func IntrusionDetection(cfg *IntrusionDetectionConfiguration) Component {
...
+	Installation       *operatorv1.InstallationSpec
+	ESClusterConfig    *relasticsearch.ClusterConfig
+	PullSecrets        []*corev1.Secret
+	Openshift          bool
+	ClusterDomain      string
+	ESLicenseType      ElasticsearchLicenseType
+	ManagedCluster     bool
+
+	// PVC fields Spec fields are immutable, set to true when and existing AD PVC
+	// is not found as to avoind
```
caseydavenport, line 118: typo: avoind

Also looks like the comment isn't 100% complete :grin: 

Also, we could probably avoid the need to have this flag altogether by using the `mergeState` logic to handle this.

Since this is common logic that, IIUC, would apply to ALL PVCs that we create, we could have `mergeState` check if the object already exists, and skip it if so.

https://github.com/tigera/operator/blob/a1a58b7d6a29786602fe6bd6c881ec4d52ee6826/pkg/controller/utils/component.go#L265-L266

Or, we could even do something sneaky like what we do for Jobs and delete / recreate the PVC, but that might not work because PVCs won't actually be removed until the pods using them are torn down as well. 

^ I don't think we need to block this PR on that, but might be something to look into as a follow on.

api/v1/intrusiondetection_types.go
caseydavenport: (Although.. I can find clear examples in the Kubernetes API iteself which ignore this rule... so maybe we're OK :sweat_smile: )

api/v1/intrusiondetection_types.go
```diff
@@ -27,8 +27,41 @@ type IntrusionDetectionSpec struct {
...
+}
+
+type AnomalyDetectionSpec struct {
+	// StorageType sets the type of storage to use for storing Anomaly Detection Models. By default it will use the ephemeral
+	// emptyDir on the node Anomaly Detection will be deployed to.
+	// default: Ephemeral
+	// +optional
+	// +kubebuilder:validation:Enum=Ephemeral;Persistent
```
caseydavenport, line 42: I believe this validation will mean that an empty string will not be accepted here, which is probably why pointers are preferred (since it won't try to validate a pointer against it?) 

Worth double-checking the behavior of this when StorageType = ""

freecaykes: just checked this, was able to provide an empty `anomalydetection:` in the IntrustionDetection CR and it updated to Ephemeral by our `fillDefaults(...)` function

pkg/controller/intrusiondetection/intrusiondetection_controller.go
freecaykes: think it would default to just `storageType: Ephemeral` since we replace the whole AnomalyDetection struct below

caseydavenport: That's right, but that in itself is a bug - our defaulting logic should never remove a field that was set by the user.

pkg/controller/intrusiondetection/intrusiondetection_controller.go
freecaykes: added validation https://github.com/tigera/operator/pull/2094/commits/0242a6e687adf0bac93f81213982ec3928c81dd8

pkg/controller/intrusiondetection/intrusiondetection_controller.go
freecaykes: understood, https://github.com/tigera/operator/pull/2094/commits/0242a6e687adf0bac93f81213982ec3928c81dd8 - ended up validating before default setting of the fields for ADSpec in case we see an ADSpec provided with that scenario 

and then check only to set default in the case if the whole ADSpec struct is empty

pkg/render/intrusion_detection.go
freecaykes: added a TODO for this https://github.com/tigera/operator/pull/2094/commits/0242a6e687adf0bac93f81213982ec3928c81dd8

pkg/controller/intrusiondetection/intrusiondetection_controller.go
caseydavenport: That's somewhat atypical for the operator codebase, and makes it harder to understand this defaulting logic because it now needs to be read in the context of some prior validation. We want our validation and defaulting functions to make sense without requiring an understanding of each other.

I think we should go with my original suggestion and default this field conditionally:

```go
if len(storageType) == 0 {
  if (storageClass) == 0 {
    // Set to ephemeral.
  } else {
    // User specified a storage class, set to persistent
  }
}
```

Alternatively, we make this field required on the API and reject it unless the user specifies it. But that's a breaking change so probably not a valid option.

pkg/render/intrusion_detection.go
caseydavenport: Would rather not leave TODOs in the code. If we want to track something, we should raise it as a github issue so that it doesn't get lost.

freecaykes: removed https://github.com/tigera/operator/pull/2094/commits/fe81c4cb19794ca1b97dd65394e9276d1e3a623d will create a github issue

pkg/controller/intrusiondetection/intrusiondetection_controller.go
freecaykes: https://github.com/tigera/operator/pull/2094/commits/fe81c4cb19794ca1b97dd65394e9276d1e3a623d

api/v1/intrusiondetection_types.go
```diff
@@ -27,8 +27,41 @@ type IntrusionDetectionSpec struct {
...
+	// Anomaly Detection API pod for model storage. The StorageClassName should only be modified when no StorageClass is currently
+	// active. We recommend choosing a storage class dedicated to AnomalyDetection only. Otherwise, model retention
+	// cannot be guaranteed during upgrades. See https://docs.tigera.io/maintenance/upgrading for up-to-date instructions.
+	// Default: tigera-anomaly-detection
+	// +optional
+	StorageClassName string `json:"storageClassName,omitempty"`
 }
 
+// StorageType sets the type of storage to be used for the specified component.
+type StorageType string
```
tmjd, line 55: I think this should include a comment 
`// One of: Ephemeral, Persistent`

api/v1/intrusiondetection_types.go
tmjd: Is it possible to have a StorageClassName and still be Ephemeral?

freecaykes: https://github.com/tigera/operator/pull/2094/files#diff-ea4aad85809ff62082b52040a934ebbf3953095d563e545bec7daa4d295f2da9R609-R618 our `filldefaults` overrides the type as perssitent if a storageclassname is provided

pkg/render/intrusion_detection.go
```diff
@@ -97,15 +103,20 @@ func IntrusionDetection(cfg *IntrusionDetectionConfiguration) Component {
...
+	ESSecrets          []*corev1.Secret
+	Installation       *operatorv1.InstallationSpec
+	ESClusterConfig    *relasticsearch.ClusterConfig
+	PullSecrets        []*corev1.Secret
+	Openshift          bool
+	ClusterDomain      string
+	ESLicenseType      ElasticsearchLicenseType
+	ManagedCluster     bool
+
+	// PVC fields Spec fields are immutable, set to true when and existing AD PVC
```
tmjd, line 117: typo
`and` should be `an`

pkg/render/intrusion_detection.go
```diff
@@ -201,9 +214,22 @@ func (c *intrusionDetectionComponent) Objects() ([]client.Object, []client.Objec
...
 			c.adAPIAccessRoleBinding(),
 		)
+
+		shouldConfigureADStorage := c.cfg.IntrusionDetection.Spec.AnomalyDetection.StorageType == operatorv1.PersistentStorageType
+
+		if !shouldConfigureADStorage {
+			objsToDelete = append(objsToDelete,
+				c.adPersistentVolumeClaim(),
+			)
+		} else if shouldConfigureADStorage && c.cfg.ShouldRenderADPVC { //
```
tmjd, line 224: trailing `//` should be removed

pkg/controller/intrusiondetection/intrusiondetection_controller.go
```diff
@@ -310,6 +334,40 @@ func (r *ReconcileIntrusionDetection) Reconcile(ctx context.Context, request rec
...
+				r.SetDegraded(operatorv1.ResourceNotFound, errMessage, err, reqLogger)
+				return reconcile.Result{}, err
+			} else {
+				errMessage := "failed to query for Anomaly Detection Storage Class"
+				r.SetDegraded(operatorv1.InternalServerError, errMessage, err, reqLogger)
+				return reconcile.Result{}, err
+			}
+		}
+
+		err = r.client.Get(ctx, client.ObjectKey{Name: render.ADPersistentVolumeClaimName, Namespace: render.IntrusionDetectionNamespace}, &corev1.PersistentVolumeClaim{})
```
tmjd, line 359: Your comment in the render says that PVCs are immutable. So if we find one I think we should be checking that the storageClassName matches what is in the PVC and if not the PVC should be removed so a new one can be created. I'm not sure if that also means the deployment (or whatever resource) that uses the PVC might need to be stopped first too.

pkg/controller/intrusiondetection/intrusiondetection_controller.go
```diff
@@ -549,11 +603,47 @@ func (r *ReconcileIntrusionDetection) setDefaultsOnIntrusionDetection(ctx contex
...
 		}
+	}
 
-		if err := r.client.Update(ctx, ids); err != nil {
-			return err
+	if !isManagedCluster { // does not set defaults for managed clustes the AD Fields specs will simply be ignored
+		if len(ids.Spec.AnomalyDetection.StorageClassName) > 0 {
+			// User specified a storage class, set to persistent
+			ids.Spec.AnomalyDetection = operatorv1.AnomalyDetectionSpec{
+				StorageType:      operatorv1.PersistentStorageType,
```
tmjd, line 612: We shouldn't overwrite this if it is already set. That of course can lead to an invalid config with a storageClassName and Ephemeral but that should be caught by validation and reported to the user through a SetDegraded and blocking changes until that is fixed.

pkg/controller/intrusiondetection/intrusiondetection_controller.go
```diff
@@ -310,6 +334,40 @@ func (r *ReconcileIntrusionDetection) Reconcile(ctx context.Context, request rec
...
+		}
+
+		if err = r.client.Get(ctx, client.ObjectKey{Name: instance.Spec.AnomalyDetection.StorageClassName}, &storagev1.StorageClass{}); err != nil {
+			if errors.IsNotFound(err) {
+				errMessage := "failed to get storage class for anomaly detection"
+				r.SetDegraded(operatorv1.ResourceNotFound, errMessage, err, reqLogger)
+				return reconcile.Result{}, err
+			} else {
+				errMessage := "failed to query for Anomaly Detection Storage Class"
+				r.SetDegraded(operatorv1.InternalServerError, errMessage, err, reqLogger)
```
tmjd, line 354: I think instead of `InternalServerError` here we should be using `ResourceReadError`

pkg/controller/intrusiondetection/intrusiondetection_controller.go
```diff
@@ -310,6 +334,40 @@ func (r *ReconcileIntrusionDetection) Reconcile(ctx context.Context, request rec
...
+			}
+		}
+
+		err = r.client.Get(ctx, client.ObjectKey{Name: render.ADPersistentVolumeClaimName, Namespace: render.IntrusionDetectionNamespace}, &corev1.PersistentVolumeClaim{})
+		if err != nil {
+			if errors.IsNotFound(err) {
+				shouldRenderADPVC = true
+			} else {
+				errMessage := "failed to query persistentvolumeclaim for anomaly detection"
+				r.SetDegraded(operatorv1.InternalServerError, errMessage, err, reqLogger)
```
tmjd, line 365: I think instead of `InternalServerError` here we should be using `ResourceReadError`

pkg/controller/intrusiondetection/intrusiondetection_controller.go
```diff
@@ -247,12 +254,29 @@ func (r *ReconcileIntrusionDetection) Reconcile(ctx context.Context, request rec
 	r.status.OnCRFound()
 	reqLogger.V(2).Info("Loaded config", "config", instance)
 
-	if err := r.setDefaultsOnIntrusionDetection(ctx, instance); err != nil {
+	managementClusterConnection, err := utils.GetManagementClusterConnection(ctx, r.client)
+	if err != nil {
+		log.Error(err, "Failed to read ManagementClusterConnection")
+		r.status.SetDegraded("Failed to read ManagementClusterConnection", err.Error())
```
tmjd, line 260: This needs to use one of the `TigeraStatusReason` values as the first argument.

tmjd: It should also use the `r.SetDegraded` instead of directly using the `r.status.SetDegraded`

pkg/controller/intrusiondetection/intrusiondetection_controller.go
```diff
@@ -531,8 +596,9 @@ func (r *ReconcileIntrusionDetection) Reconcile(ctx context.Context, request rec
 	return reconcile.Result{}, nil
 }
 
-// setDefaultsOnIntrusionDetection updates the IntrusionDetection resource with defaults if ComponentResources is not populated.
-func (r *ReconcileIntrusionDetection) setDefaultsOnIntrusionDetection(ctx context.Context, ids *operatorv1.IntrusionDetection) error {
+// fillDefaults updates the IntrusionDetection resource with defaults if
+// ComponentResources and AnomalyDetectionSpec are not populated.
+func (r *ReconcileIntrusionDetection) fillDefaults(ctx context.Context, ids *operatorv1.IntrusionDetection, isManagedCluster bool) error {
```
tmjd, line 601: isManagedCluster is no longer used in this function so you should remove it.

pkg/controller/intrusiondetection/intrusiondetection_controller_test.go
```diff
@@ -418,6 +430,177 @@ var _ = Describe("IntrusionDetection controller tests", func() {
...
+					},
+				},
+			})).ShouldNot(HaveOccurred())
+
+			_, err := r.Reconcile(ctx, reconcile.Request{})
+			Expect(err).Should(HaveOccurred())
+			Expect(mockStatus.AssertNumberOfCalls(nil, "SetDegraded", 1)).To(BeTrue())
+		})
+
+		It("should degrade if ADSpec,StorageClassName is provided when type is Ephemeral", func() {
```
tmjd, line 509: This title isn't accurate anymore. Is there anything special this test is covering that isn't covered by the other tests?

pkg/controller/intrusiondetection/intrusiondetection_controller.go
tmjd: I still have a concern that there may be some setups that will block us from being able to clean up the PVC with a deployment still referencing it, but that is probably an edge case that we might not need to worry about.

pkg/controller/intrusiondetection/intrusiondetection_controller.go
freecaykes: removed https://github.com/tigera/operator/pull/2094/commits/e393f5a8741e3e1206575f5c92123a92ebc837f9

pkg/controller/intrusiondetection/intrusiondetection_controller_test.go
freecaykes: nope nothing new, just forgot to remove https://github.com/tigera/operator/pull/2094/commits/e393f5a8741e3e1206575f5c92123a92ebc837f9

tmjd (approved):
LGTM

marvin-tigera:

marvin-tigera:

PR #2096, opened 2022-07-25
WIP: handle new operator component fields in migration

lmm:
## Description

This PR is a follow-up to #2063 and adds migration support for the new component resource fields that the operator now supports.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


pkg/controller/migration/convert/core.go
```diff
@@ -211,210 +467,372 @@ func checkNodeHostPathVolume(spec corev1.PodSpec, name, path string) error {
...
+	install.Spec.ComponentResources = nil
+}
+
+// migrateContainerResources adds the resources for the specified component container if none was previously set. If the Installation
+// already had a resource for the component container then they are compared and if they are different then an error is returned.
+// If the component container resource is added to the Installation or the existing one matches then nil is returned.
+func migrateContainerResources(install *operatorv1.Installation, componentName, containerName string, resources *corev1.ResourceRequirements) error {
+	// If resources already exist for this component container, verify that they equal the container's existing resources.
+	if existingResources, found := getContainerResourceOverride(install, componentName, containerName); found {
+		if !reflect.DeepEqual(existingResources, resources) {
```
lmm, line 617: I need to address this. This is wrong since slice element order is taken into account in reflect.DeepEqual.

pkg/controller/migration/convert/core.go
```diff
@@ -198,6 +238,222 @@ func handleCore(c *components, install *operatorv1.Installation) error {
...
+			if installTolerations == nil {
+				helpers.EnsureCalicoNodePodSpecNotNil(install)
+				install.Spec.CalicoNodeDaemonSet.Spec.Template.Spec.Tolerations = compTolerations
+			} else {
+				if !areEqualTolerations(compTolerations, installTolerations) {
+					return ErrIncompatibleTolerations(ComponentCalicoNode)
+				}
+			}
+		}
+	}
```
tmjd, line 379: Feels like you could create a function that takes in the `installTolerations, compTolerations, defaultTolerations` and returns the tolerations (that should be on the install), err, then you could use that one function for all node, typha, and kube controllers

pkg/controller/migration/convert/core.go
```diff
@@ -198,6 +238,222 @@ func handleCore(c *components, install *operatorv1.Installation) error {
...
+			if minReadySeconds != nil {
+				return ErrIncompatibleMinReadySeconds(ComponentKubeControllers)
+			}
+		}
+	}
+
+	return nil
+}
+
+func getTolerations(override comp.ReplicatedPodResourceOverrides) []corev1.Toleration {
```
tmjd, line 314: This function does not seem very useful to me. In each of the places it is used couldn't the function `GetTolerations()` be called on what is passed into this function?

tmjd (commented):
Just submitting what I've reviewed so far since I'm done for the day.

pkg/controller/migration/convert/core.go
lmm: Addressed in cd45e1a5c8eb0920c69d23f04b64f20fb3bde90a

Probably more refactoring I could do in core.go, I'll take a closer look tomorrow.

pkg/controller/migration/convert/core.go
lmm: Removed

pkg/controller/migration/convert/core.go
```diff
@@ -211,210 +443,372 @@ func checkNodeHostPathVolume(spec corev1.PodSpec, name, path string) error {
...
 		}
+	} else {
+		// Affinity is not set on the component, check if it's set on the installation.
+		if install.Spec.CalicoNodeDaemonSet != nil && install.Spec.CalicoNodeDaemonSet.GetAffinity() != nil {
+			return ErrIncompatibleAffinity(ComponentCalicoNode)
+		}
 	}
+
 	nodeSel := removeOSNodeSelectors(c.node.Spec.Template.Spec.NodeSelector)
 	delete(nodeSel, "projectcalico.org/operator-node-migration")
```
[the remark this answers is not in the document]
tmjd: I think we also need to call `removeOSNodeSelectors` before we do that assignment above also.

pkg/controller/migration/convert/core.go
```diff
@@ -211,210 +443,372 @@ func checkNodeHostPathVolume(spec corev1.PodSpec, name, path string) error {
...
-					},
-				}
+			// Ensure that the affinity on the installation uses the new field and clears the deprecated TyphaAffinity.
+			helpers.EnsureTyphaPodSpecNotNil(install)
+			install.Spec.TyphaDeployment.Spec.Template.Spec.Affinity = c.typha.Spec.Template.Spec.Affinity
+			install.Spec.TyphaAffinity = nil
+		} else {
+			// Affinity is not set on the component, check if it's set on the installation.
+			if installAffinity != nil {
+				return ErrIncompatibleAffinity(ComponentTypha)
```
tmjd, line 760: I think the user adding affinity to the install would be fine, why should we error?

pkg/controller/migration/convert/core.go
```diff
@@ -211,210 +443,372 @@ func checkNodeHostPathVolume(spec corev1.PodSpec, name, path string) error {
...
-			return ErrIncompatibleCluster{
-				err:       "node affinity not supported for calico-node daemonset",
-				component: ComponentCalicoNode,
-				fix:       "remove the affinity",
+		if install.Spec.CalicoNodeDaemonSet == nil || install.Spec.CalicoNodeDaemonSet.GetAffinity() == nil {
+			// Affinity set on the component but not the installation so migrate it over.
+			helpers.EnsureCalicoNodePodSpecNotNil(install)
+			install.Spec.CalicoNodeDaemonSet.Spec.Template.Spec.Affinity = c.node.Spec.Template.Spec.Affinity
+		} else {
+			// Affinity is set on the component and the installation, verify that they match.
```
tmjd, line 715: Do we do this in other places of the migration? (Check that an installation setting matches what is being migrated.)
I'm thinking that if the user has specified an affinity that we should just use it. They've defined the setting so who knows if they want it to match what they previously had or not.

pkg/controller/apiserver/apiserver_controller.go
```diff
@@ -16,10 +16,13 @@ package apiserver
```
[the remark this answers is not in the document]
lmm: It does but the manifests and instructions to do so are under the Operations section: https://projectcalico.docs.tigera.io/maintenance/install-apiserver

pkg/controller/migration/convert/core.go
```diff
@@ -198,6 +238,198 @@ func handleCore(c *components, install *operatorv1.Installation) error {
 	return nil
 }
 
+func getMinReadySeconds(override comp.ReplicatedPodResourceOverrides) *int32 {
```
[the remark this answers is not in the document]
lmm: It's located here: https://github.com/tigera/operator/blob/c2f16b5ba5bb20a31bb3828ced9f4c4bf4a0e69c/pkg/components/overrides.go#L24

pkg/controller/migration/convert/core.go
```diff
@@ -198,6 +238,198 @@ func handleCore(c *components, install *operatorv1.Installation) error {
...
+			helpers.EnsureCalicoNodeSpecNotNil(install)
+			install.Spec.CalicoNodeDaemonSet.Spec.MinReadySeconds = &c.node.Spec.MinReadySeconds
+		} else {
+			if *minReadySeconds != c.node.Spec.MinReadySeconds {
+				return ErrIncompatibleMinReadySeconds(ComponentCalicoNode)
+			}
+		}
+	} else {
+		// minReadySeconds is not set on the component, check if it's set on the installation.
+		if minReadySeconds != nil {
```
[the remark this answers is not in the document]
lmm: I thought I saw the same logic elsewhere in the migration package where if the value was defined in the Installation CR, that we'd verify that it was set to the autodetected value on the component. I might be wrong though.

PR #2101, opened 2022-07-27
Tweak Typha tolerations to be less permissive

caseydavenport:
## Description

As discussed in https://github.com/projectcalico/calico/issues/6136

Now that users can override these, we have the opportunity to make our default values a bit more "correct" and friendly. 

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/common/meta/meta.go
+++ b/pkg/render/common/meta/meta.go
@@ -20,6 +20,7 @@ import (
 	"time"
 
 	operatorv1 "github.com/tigera/operator/api/v1"
+	"github.com/tigera/operator/pkg/ptr"
 
 	corev1 "k8s.io/api/core/v1"
 )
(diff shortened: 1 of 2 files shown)


pkg/render/common/meta/meta.go
```diff
@@ -67,6 +68,26 @@ var (
...
+			// in the future for an extended period of time.
+			Effect:            corev1.TaintEffectNoSchedule,
+			Key:               "node.kubernetes.io/not-ready",
+			TolerationSeconds: ptr.Int64ToPtr(300),
+		},
+		{
+			// Tolerate a node becoming unreachable by the apiserver for
+			// a short while.
+			Effect:            corev1.TaintEffectNoExecute,
+			Key:               "node.kubernetes.io/unreachable",
```
tmjd, line 87: Is there documentation about the taints that a node will get during bootstrap?

caseydavenport: Not specifically in the context of bootstrap, but it is discussed a bit here: https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/#taint-based-evictions

While I do think it is probably a better experience for most users to be able to configure eviction properly, there is a risk that changing these will have unintended consequences for other users.

tmjd: What is the reasoning that bootstrap needs to tolerate "unreachable".
It doesn't seem like we would want to tolerate unreachable during bootstrap, because if a node was unreachable then we wouldn't want typha to be scheduled to that node anyway.

PR #2102, opened 2022-07-28
Fixing the Reason validation

vara2504:
## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/status/status.go
+++ b/pkg/controller/status/status.go
@@ -182,7 +182,7 @@ func (m *statusManager) updateStatus() {
 		}
 
 		if m.IsDegraded() {
-			m.setDegraded(m.degradedReason(), m.degradedMessage())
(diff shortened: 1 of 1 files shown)


pkg/controller/status/status.go
```diff
@@ -182,7 +182,7 @@ func (m *statusManager) updateStatus() {
 		}
 
 		if m.IsDegraded() {
-			m.setDegraded(m.degradedReason(), m.degradedMessage())
+			m.setDegraded(string(operator.PodFailure), fmt.Sprintf("%s - %s", m.degradedReason(), m.degradedMessage()))
```
tmjd, line 185: I think understand why you did this but I don't think putting `PodFailure` here will be correct most of the time.

PR #2103, opened 2022-07-28
Fix broken reasons, due to semicolon

rene-dekker:
- Avoid having unknown reason combined with a valid reason.
- Avoid having the semicolon in the reason

Reasons may never be empty strings, because that does not pass the validation for Conditions.

### DIFF
--- a/pkg/controller/status/status.go
+++ b/pkg/controller/status/status.go
@@ -845,7 +845,27 @@ func (m *statusManager) degradedReason() string {
 	if len(m.failing) != 0 {
 		reasons = append(reasons, "Some pods are failing")
 	}
-	return strings.Join(reasons, "; ")
+
+	return formatReasons(reasons)
+}
+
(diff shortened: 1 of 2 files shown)


pkg/controller/status/status.go
```diff
@@ -845,7 +845,27 @@ func (m *statusManager) degradedReason() string {
 	if len(m.failing) != 0 {
 		reasons = append(reasons, "Some pods are failing")
 	}
-	return strings.Join(reasons, "; ")
+
+	return formatReasons(reasons)
```
tmjd, line 849: I think instead of combining these we should just return the "highest" priority of the reasons and I would say the order they appear in this function is the right priority:
1. explicitDegradedReason
2. windowsUpgradeDegradedMsg
3. pods failing
I think we'll still have the full set of problems in the degradedMessage, so I don't think we'll be losing much info that way.

PR #2104, opened 2022-07-28
Sanitize reason messages until we have them all updated

tmjd:
## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/status/status.go
+++ b/pkg/controller/status/status.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"fmt"
 	"reflect"
+	"regexp"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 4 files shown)


pkg/controller/status/status.go
```diff
@@ -79,6 +80,16 @@ type StatusManager interface {
 	SetMetaData(meta *metav1.ObjectMeta)
 }
 
+// This regex matches any characters that are not allowed in a Status Reason
+var reasonInvalidCharacters = regexp.MustCompile(`[^A-Za-z0-9_,:]`)
+
+// This regex matches any characters not allows at the start of a Status Reason
```
rene-dekker, line 86: nit:
"allows" -> "allowed" . Happens 2x.
Also, it's best to end the comments with a period.

pkg/controller/status/status.go
```diff
@@ -79,6 +80,16 @@ type StatusManager interface {
...
+var reasonInvalidStartCharacters = regexp.MustCompile(`^[^A-Za-z]*`)
```
rene-dekker, line 87: Why is this plural again for the start character and why does it have the asterisk?

rene-dekker: Never mind.

tmjd: I had to think about that a bit myself when I looked at this. I'll add more to the comments to try to clarify what it is used for.

rene-dekker (approved):
lgtm

tmjd:
I received verbal approval from @rene-dekker that these changes still look good.

PR #2106, opened 2022-07-29
Add Fips mode installation flag

rene-dekker:
Some key features are disabled when fips mode is enabled:
- The python based pods (intrusion detection installer, anomaly detection)
- Kibana
- Application layer


api/v1/installation_types.go
```diff
@@ -162,8 +162,21 @@ type InstallationSpec struct {
 
 	// CalicoWindowsUpgradeDaemonSet configures the calico-windows-upgrade DaemonSet.
 	CalicoWindowsUpgradeDaemonSet *CalicoWindowsUpgradeDaemonSet `json:"calicoWindowsUpgradeDaemonSet,omitempty"`
+
+	// FIPSMode uses images and features only that are using FIPS 140-2 validated cryptographic modules and standards.
+	// Default: Disabled
+	// +kubebuilder:validation:Enum=Enabled;Disabled
+	// +optional
+	FIPSMode FIPSMode `json:"fipsMode,omitempty"`
```
caseydavenport, line 170: I think this should be a pointer, right?

pkg/controller/logstorage/esgateway.go
```diff
@@ -51,15 +52,18 @@ func (r *ReconcileLogStorage) createEsGateway(
...
-		reqLogger.Error(err, "failed to get Kibana tls certificate secret")
-		r.status.SetDegraded("Failed to get Kibana tls certificate secret", err.Error())
-		return reconcile.Result{}, false, err
-	} else if kibanaCertificate == nil {
-		reqLogger.Info("Waiting for internal Kibana tls certificate secret to be available")
-		r.status.SetDegraded("Waiting for internal Kibana tls certificate secret to be available", "")
-		return reconcile.Result{}, false, nil
+	var kibanaCertificate certificatemanagement.CertificateInterface
+	if install.FIPSMode != operatorv1.FIPSModeEnabled {
+		kibanaCertificate, err := certificateManager.GetCertificate(r.client, render.TigeraKibanaCertSecret, common.OperatorNamespace())
```
caseydavenport, line 57: This is because kibana is not FIPS compliant I'm guessing?

pkg/render/node.go
```diff
@@ -1454,6 +1455,7 @@ func (c *nodeComponent) nodeEnvVars() []corev1.EnvVar {
 				corev1.EnvVar{Name: "FELIX_PROMETHEUSREPORTERCERTFILE", Value: c.cfg.PrometheusServerTLS.VolumeMountCertificateFilePath()},
 				corev1.EnvVar{Name: "FELIX_PROMETHEUSREPORTERKEYFILE", Value: c.cfg.PrometheusServerTLS.VolumeMountKeyFilePath()},
 				corev1.EnvVar{Name: "FELIX_PROMETHEUSREPORTERCAFILE", Value: c.cfg.TLS.TrustedBundle.MountPath()},
+				corev1.EnvVar{Name: "FELIX_PROMETHEUSREPORTERFIPSMODEENABLED", Value: fmt.Sprintf("%v", c.cfg.Installation.FIPSMode == operatorv1.FIPSModeEnabled)},
```
caseydavenport, line 1458: I thought we said we're bundling all of the FELIX_XFIPSMODEENABLED into a single FIPS_MODE_ENABLED var? 

>  fips mode is a cluster wide setting

There's not really a use case for turning FIPS mode on per-connection to Felix, it's either on for all of them or off for all of them.

api/v1/installation_types.go
rene-dekker: I think you are right. Let me change that.

I just did a quick evaluation on the string aliases that we have for installation_types. We are not very consistent currently.

We don't use pointers for:
- ProductVariant
- Provider
- ComponentName
- EncapsulationType
- NATOutgoingType
- CNIPluginType
- IPAMPluginType

But we do for:
- NonPrivilegedType
- ContainerIPForwardingType
- HostPortsType
- MultiInterfaceMode
- BGPOption
- LinuxDataplaneOption
- KubernetesAutodetectionMethod

rene-dekker: The resulting CRD is identical.

pkg/controller/logstorage/esgateway.go
```diff
@@ -51,15 +52,18 @@ func (r *ReconcileLogStorage) createEsGateway(
...
-	if err != nil {
-		reqLogger.Error(err, "failed to get Kibana tls certificate secret")
-		r.status.SetDegraded("Failed to get Kibana tls certificate secret", err.Error())
-		return reconcile.Result{}, false, err
-	} else if kibanaCertificate == nil {
-		reqLogger.Info("Waiting for internal Kibana tls certificate secret to be available")
-		r.status.SetDegraded("Waiting for internal Kibana tls certificate secret to be available", "")
-		return reconcile.Result{}, false, nil
+	var kibanaCertificate certificatemanagement.CertificateInterface
+	if !operatorv1.IsFIPSModeEnabled(install.FIPSMode) {
```
Brian-McM, line 56: Out of curiosity, why don't we care about the certificate in FIPS mode?

pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -488,6 +489,11 @@ func (r *ReconcileLogStorage) Reconcile(ctx context.Context, request reconcile.R
 			r.status.SetDegraded("Failed to get curator credentials", err.Error())
 			return reconcile.Result{}, err
 		}
+		applyTrial, err = r.applyElasticTrialSecret(ctx, install)
```
Brian-McM, line 492: What does this have to do with FIPS mode?

pkg/controller/logstorage/esgateway.go
Brian-McM: Oh I guess we just don't install Kibana if FIPS mode is enabled, is that correct?

pkg/render/intrusion_detection.go
```diff
@@ -193,7 +193,7 @@ func (c *intrusionDetectionComponent) Objects() ([]client.Object, []client.Objec
 	objs = append(objs, c.globalAlertTemplates()...)
 
 	// AD Related deployment only for management/standalone cluster
-	if !c.cfg.ManagedCluster {
+	if !c.cfg.ManagedCluster && !operatorv1.IsFIPSModeEnabled(c.cfg.Installation.FIPSMode) {
```
Brian-McM, line 196: Could you add a comment saying why we don't want these objects in non FIPS mode?

pkg/render/intrusion_detection.go
```diff
@@ -217,7 +217,7 @@ func (c *intrusionDetectionComponent) Objects() ([]client.Object, []client.Objec
 		objs = append(objs, c.adDetectorPodTemplates()...)
 	}
 
-	if !c.cfg.ManagedCluster {
+	if !c.cfg.ManagedCluster && !operatorv1.IsFIPSModeEnabled(c.cfg.Installation.FIPSMode) {
```
Brian-McM, line 220: I'm a little curious about this. Why can't IDS use / configure elasticsearch in none fips mode? I thought IDS was pretty reliant on ES, do we basically disable all IDS features in FIPS mode?

Brian-McM (commented):
Just had a few questions about what's disabled in FIPs mode.

pkg/controller/logstorage/esgateway.go
rene-dekker: That's right.

pkg/controller/logstorage/logstorage_controller.go
rene-dekker: We will leverage a FIPS build of elasticsearch, which requires a license.

Brian-McM (approved):

pkg/controller/logstorage/logstorage_controller.go
rene-dekker: I added a comment to the method itself to detail why.

Issue #2006, opened 2022-06-10
node clusterrole does have permissions to read secrets.


georgyo:
## Expected Behavior

When configuring a BGPPeer with has a password, I would expect calico-node to be able to read the secretref.

## Current Behavior

Calico-node does not have permission to list or get secrets, and as such it logs spew the following lines:
```
W0610 00:16:46.957663      69 reflector.go:324] pkg/mod/k8s.io/client-go@v0.23.3/tools/cache/reflector.go:167: failed to list *v1.Secret: secrets "bgp-password" is forbidden: User "system:serviceaccount:calico-system:calico-node" cannot list resource "secrets" in API group "" in the namespace "calico-system"
E0610 00:16:46.957705      69 reflector.go:138] pkg/mod/k8s.io/client-go@v0.23.3/tools/cache/reflector.go:167: Failed to watch *v1.Secret: failed to list *v1.Secret: secrets "bgp-password" is forbidden: User "system:serviceaccount:calico-system:calico-node" cannot list resource "secrets" in API group "" in the namespace "calico-system"
```

## Possible Solution

Add the ability to clusterrole calcio-node to be able to list and read secrets.

https://github.com/tigera/operator/blob/fc4f3d8f2c6c44ccf6b6b93623972dd2b45db91e/pkg/render/node.go#L282

## Steps to Reproduce (for bugs)

1. Create a BGP Peer like the following: 
``` 
---
apiVersion: projectcalico.org/v3
kind: BGPPeer
metadata:
  name: my-global-peer4
spec:
  peerIP: 169.254.169.254
  asNumber: 64515
  sourceAddress: UseNodeIP
  password:
    secretKeyRef:
      key: password
      name: vultr-bgp-password
```
2. Watch the logs show it fail to be able to read the secret.

## Your Environment

* Operating System and version:
* Link to your project (optional):

ruslanloman:
Hi @georgyo , we faced with same issue. I've added appropriate PR to fix it. please vote - https://github.com/tigera/operator/pull/2107

PR #2107, opened 2022-08-01
[Issue #2006] Add {get,list,watch}/secrets permissions to calico-node rbac.

ruslanloman:
Hi,
We faced with issue that calico-node couldn't read bgp password from secrets since there is no appropriate permissions for this role [1]. This PR adds {get, list, watch}/secrets permissions to calico-node rbac. Also i have created PR to calico helm chart too. [2]

Thanks!
[1] https://projectcalico.docs.tigera.io/reference/resources/bgppeer#bgppassword
[2] https://github.com/projectcalico/calico/pull/6475

### DIFF
--- a/pkg/render/node.go
+++ b/pkg/render/node.go
@@ -316,6 +316,12 @@ func (c *nodeComponent) nodeRole() *rbacv1.ClusterRole {
 				Verbs:     []string{"watch", "list", "get"},
 			},
 			{
+				// This permission is needed to get bgp password for BGPPeer resource.
+				APIGroups: []string{""},
+				Resources: []string{"secrets"},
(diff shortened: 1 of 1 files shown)


caseydavenport:
@ruslanloman this is intentional - we don't want calico/node to have secrets access unless its really needed. 

What you should do is create your own `Role` and `RoleBinding` to provide access to the specific secret you need. I'm not sure why this is missing from the Calico OSS docs, but this is what should be there: https://docs.tigera.io/v3.14/security/comms/secure-bgp#ensure-rbac-permissions

ruslanloman:
ok, thanks for your reply.

PR #2111, opened 2022-08-04
Support Compliance CR omission

pasanw:
## Description
Allow for the Compliance CR to be omitted by
- Only enforcing the Manager dependency on Compliance when both Compliance CR and license feature are present (previously only license feature)
- Setting the new feature-flag environment variable on the Manager container to disable Compliance in the Manager UI when the Compliance CR is not present.

I've validated that CE install (standalone) progresses with Compliance CR omitted, and that Compliance and Manager both transition to available after creating the Compliance CR.

This mitigates CI-1021 and pairs with EV-2392.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/manager/manager_controller.go
+++ b/pkg/controller/manager/manager_controller.go
@@ -496,7 +496,7 @@ func (r *ReconcileManager) Reconcile(ctx context.Context, request reconcile.Requ
 		ESLicenseType:           elasticLicenseType,
 		Replicas:                replicas,
 		Compliance:              complianceCR,
-		ComplianceFeatureActive: complianceLicenseFeatureActive,
+		ComplianceLicenseActive: complianceLicenseFeatureActive,
 		UsePSP:                  r.usePSP,
 	}
 
--- a/pkg/controller/manager/manager_controller.go
+++ b/pkg/controller/manager/manager_controller.go
@@ -313,14 +313,14 @@ func (r *ReconcileManager) Reconcile(ctx context.Context, request reconcile.Requ
 	complianceLicenseFeatureActive := utils.IsFeatureActive(license, common.ComplianceFeature)
 	complianceCR, err := compliance.GetCompliance(ctx, r.client)
 	if err != nil && !errors.IsNotFound(err) {
-		r.status.SetDegraded("Error querying compliance", err.Error())
+		r.status.SetDegraded(string(operatorv1.ResourceReadError), fmt.Sprintf("Error querying compliance: %s", err.Error()))
 		return reconcile.Result{}, err
 	}
 
 	if complianceLicenseFeatureActive && complianceCR != nil {
 		// Check that compliance is running.
 		if complianceCR.Status.State != operatorv1.TigeraStatusReady {
-			r.status.SetDegraded("Compliance is not ready", fmt.Sprintf("compliance status: %s", complianceCR.Status.State))
+			r.status.SetDegraded(string(operatorv1.ResourceNotReady), "Compliance is not ready")
 			return reconcile.Result{}, nil
 		}
 		trustedSecretNames = append(trustedSecretNames, render.ComplianceServerCertSecret)
@@ -481,23 +481,23 @@ func (r *ReconcileManager) Reconcile(ctx context.Context, request reconcile.Requ
 	}
 
 	managerCfg := &render.ManagerConfiguration{
-		KeyValidatorConfig:             keyValidatorConfig,
-		ESSecrets:                      esSecrets,
-		TrustedCertBundle:              trustedBundle,
-		ESClusterConfig:                esClusterConfig,
-		TLSKeyPair:                     tlsSecret,
-		PullSecrets:                    pullSecrets,
-		Openshift:                      r.provider == operatorv1.ProviderOpenShift,
-		Installation:                   installation,
-		ManagementCluster:              managementCluster,
-		TunnelSecret:                   tunnelSecret,
-		InternalTrafficSecret:          internalTrafficSecret,
-		ClusterDomain:                  r.clusterDomain,
-		ESLicenseType:                  elasticLicenseType,
-		Replicas:                       replicas,
-		Compliance:                     complianceCR,
-		ComplianceLicenseFeatureActive: complianceLicenseFeatureActive,
-		UsePSP:                         r.usePSP,
+		KeyValidatorConfig:      keyValidatorConfig,
+		ESSecrets:               esSecrets,
+		TrustedCertBundle:       trustedBundle,
+		ESClusterConfig:         esClusterConfig,
+		TLSKeyPair:              tlsSecret,
+		PullSecrets:             pullSecrets,
+		Openshift:               r.provider == operatorv1.ProviderOpenShift,
+		Installation:            installation,
+		ManagementCluster:       managementCluster,
+		TunnelSecret:            tunnelSecret,
+		InternalTrafficSecret:   internalTrafficSecret,
+		ClusterDomain:           r.clusterDomain,
+		ESLicenseType:           elasticLicenseType,
+		Replicas:                replicas,
+		Compliance:              complianceCR,
+		ComplianceFeatureActive: complianceLicenseFeatureActive,
+		UsePSP:                  r.usePSP,
 	}
 
 	// Render the desired objects from the CRD and create or update them.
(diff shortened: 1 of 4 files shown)


rene-dekker (approved):
lgtm

pkg/render/manager.go
```diff
@@ -110,21 +110,22 @@ func Manager(cfg *ManagerConfiguration) (Component, error) {
...
+	Openshift                      bool
+	Installation                   *operatorv1.InstallationSpec
+	ManagementCluster              *operatorv1.ManagementCluster
+	TunnelSecret                   certificatemanagement.KeyPairInterface
+	InternalTrafficSecret          certificatemanagement.KeyPairInterface
+	ClusterDomain                  string
+	ESLicenseType                  ElasticsearchLicenseType
+	Replicas                       *int32
+	Compliance                     *operatorv1.Compliance
+	ComplianceLicenseFeatureActive bool
```
tmjd, line 128: I don't think `Feature` adds any information to this variable name with `License` in the name now so please remove it since it is a bit long.

pkg/controller/manager/manager_controller.go
```diff
@@ -309,20 +309,18 @@ func (r *ReconcileManager) Reconcile(ctx context.Context, request reconcile.Requ
 	}
 
 	trustedSecretNames := []string{render.PacketCaptureCertSecret, render.PrometheusTLSSecretName, relasticsearch.PublicCertSecret}
-	installCompliance := utils.IsFeatureActive(license, common.ComplianceFeature)
-	if installCompliance {
+
+	complianceLicenseFeatureActive := utils.IsFeatureActive(license, common.ComplianceFeature)
+	complianceCR, err := compliance.GetCompliance(ctx, r.client)
+	if err != nil && !errors.IsNotFound(err) {
+		r.status.SetDegraded("Error querying compliance", err.Error())
```
tmjd, line 316: Since we're touching this SetDegraded call we should update them to use the appropriate TigeraStatusReason like can be found [here](https://github.com/tigera/operator/blob/master/api/v1/tigerastatus_types.go#L120).

pkg/controller/manager/manager_controller.go
```diff
@@ -309,20 +309,18 @@ func (r *ReconcileManager) Reconcile(ctx context.Context, request reconcile.Requ
...
-				r.status.SetDegraded("Compliance not found", err.Error())
-				return reconcile.Result{}, err
-			}
-			r.status.SetDegraded("Error querying compliance", err.Error())
-			return reconcile.Result{}, err
-		}
-		if compliance.Status.State != operatorv1.TigeraStatusReady {
-			r.status.SetDegraded("Compliance is not ready", fmt.Sprintf("compliance status: %s", compliance.Status.State))
+		if complianceCR.Status.State != operatorv1.TigeraStatusReady {
+			r.status.SetDegraded("Compliance is not ready", fmt.Sprintf("compliance status: %s", complianceCR.Status.State))
```
tmjd, line 323: Same about this SetDegraded call, we should use a TigeraStatusReason.

pkg/controller/manager/manager_controller.go
```diff
@@ -483,22 +481,23 @@ func (r *ReconcileManager) Reconcile(ctx context.Context, request reconcile.Requ
...
+		PullSecrets:                    pullSecrets,
+		Openshift:                      r.provider == operatorv1.ProviderOpenShift,
+		Installation:                   installation,
+		ManagementCluster:              managementCluster,
+		TunnelSecret:                   tunnelSecret,
+		InternalTrafficSecret:          internalTrafficSecret,
+		ClusterDomain:                  r.clusterDomain,
+		ESLicenseType:                  elasticLicenseType,
+		Replicas:                       replicas,
+		Compliance:                     complianceCR,
```
tmjd, line 498: Why pass both Compliance and ComplianceLicenseFeatureActive? It seems like it would be fine to have one variable that represented if compliance is Enabled.

pkg/render/manager.go
```diff
@@ -353,6 +354,7 @@ func (c *managerComponent) managerEnvVars() []corev1.EnvVar {
 		{Name: "ENABLE_MULTI_CLUSTER_MANAGEMENT", Value: strconv.FormatBool(c.cfg.ManagementCluster != nil)},
 		// Kibana does not have a FIPS compatible mode, therefore we disable the button in the UI.
 		{Name: "ENABLE_KIBANA", Value: strconv.FormatBool(!operatorv1.IsFIPSModeEnabled(c.cfg.Installation.FIPSMode))},
+		{Name: "ENABLE_COMPLIANCE_REPORTS", Value: strconv.FormatBool(c.cfg.Compliance != nil)},
```
tmjd, line 357: Why doesn't this one also depend on the License for Compliance?

tmjd (changes requested):

pkg/render/manager.go
pasanw: My original intent was to differentiate between license feature and product feature, however I was on the fence. I can dock `License`

pkg/controller/manager/manager_controller.go
pasanw: Will update both SetDegraded calls - thanks for the suggestion and link.

pkg/controller/manager/manager_controller.go
pasanw: I opted to pass the compliance CR presence and license feature status as two separate inputs since the Manager UI has a 'license insufficient' rendering state, where it renders the nav item but informs the user that they are not licensed to use a feature. This is separate from its 'feature flagged off' state, which omits the nav item.

Assuming your suggestion is one boolean input which is true when the CR is present and license feature is active: if we were to set the compliance feature flag on the Manager to false as a result of the license feature not being active, there would be no differentiation in the UI between the CR not being present and the license being insufficient. Hence why I opted to pass two separate inputs

WDYT? Let me know if I'm missing something.

pkg/render/manager.go
pasanw: See my [response](https://github.com/tigera/operator/pull/2111#discussion_r953118710) to your earlier comment, let me know if it answers your question.

As an aside, I left a code comment in the manager test explaining this. If we keep it this way, perhaps it would be worth  moving that here?

pkg/render/manager.go
tmjd: I was trying to suggest we remove `Feature` from the name if we're including `License`.

pkg/render/manager.go
```diff
@@ -431,7 +433,7 @@ func (c *managerComponent) managerProxyContainer() corev1.Container {
 		{Name: "VOLTRON_ENABLE_MULTI_CLUSTER_MANAGEMENT", Value: strconv.FormatBool(c.cfg.ManagementCluster != nil)},
 		{Name: "VOLTRON_TUNNEL_PORT", Value: defaultTunnelVoltronPort},
 		{Name: "VOLTRON_DEFAULT_FORWARD_SERVER", Value: "tigera-secure-es-gateway-http.tigera-elasticsearch.svc:9200"},
-		{Name: "VOLTRON_ENABLE_COMPLIANCE", Value: fmt.Sprintf("%v", c.cfg.ComplianceFeatureActive)},
+		{Name: "VOLTRON_ENABLE_COMPLIANCE", Value: strconv.FormatBool(c.cfg.Compliance != nil && c.cfg.ComplianceFeatureActive)},
```
tmjd, line 436: We are letting manager handle the state of the feature (Compliance) based on the license, but why does the operator need to handle the state of the feature for voltron? Seems like we should do the same as the manager UI and count on the component understanding the license and if the feature is enabled or not.

What is enabled/disabled with this env var?

pkg/render/manager.go
pasanw: :facepalm:

pkg/render/manager.go
pasanw: This env var is used [here](https://github.com/tigera/calico-private/blob/e2b4d67f589f0889baacd956e7c6b789d23d26d8/voltron/cmd/voltron/main.go#L189) - it registers compliance as a target for the default proxy (i.e. proxy within management cluster IIUC).

It appears to be the pattern for Voltron to be passed feature enable decisions via env var rather than make those decision itself

tmjd (approved):
LGTM

rene-dekker (commented):
lgtm

PR #2115, opened 2022-08-10
FIPS changes for Elastic.

rene-dekker:
- Adds init container for FIPS
- Moved resource requirements code up, so vm arguments can be parsed more easily
- Added image ending with -fips for elasticsearch, similar to how cni is done


pkg/render/logstorage.go
```diff
@@ -138,6 +138,39 @@ const (
 )
 
 const (
+	// ElasticsearchKeystoreSecret Currently only used when FIPS mode is enabled, we need to initialize the keystore with a password.
+	ElasticsearchKeystoreSecret         = "tigera-secure-elasticsearch-keystore"
+	ElasticsearchKeystoreEnvName        = "KEYSTORE_PASSWORD"
+	ElasticsearchKeystoreHashAnnotation = "hash.operator.tigera.io/keystore-password"
+
+	keystoreInitContainerName = "elastic-internal-init-jvm-keystore"
```
hjiawei, line 146: call it `elastic-internal-init-keystore` (remove `jvm`)?

pkg/render/logstorage.go
```diff
@@ -138,6 +138,39 @@ const (
...
+	keystoreInitScript        = `#!/usr/bin/env bash
```
hjiawei, line 147: Move this script template into a template file and use [go embed](https://pkg.go.dev/embed)?

pkg/render/logstorage.go
```diff
@@ -498,6 +540,59 @@ func (es elasticsearchComponent) podTemplate() corev1.PodTemplateSpec {
...
+		// Set to 50% of the default memory, such that resources can be divided over ES and Lucene.
+		javaOptsEnv = corev1.EnvVar{
+			Name:  "ES_JAVA_OPTS",
+			Value: fmt.Sprintf("-Xms2G -Xmx2G"),
+		}
+	}
+	var env []corev1.EnvVar
+	if operatorv1.IsFIPSModeEnabled(es.cfg.Installation.FIPSMode) {
+		javaOpts := fmt.Sprintf("--module-path /usr/share/bc-fips/ "+
+			"-Djavax.net.ssl.trustStore=/usr/share/elasticsearch/config/cacerts.bcfks"+
```
hjiawei, line 575: nit. space after `cacerts.bcfks` and also update that in your UT.

pkg/render/logstorage.go
```diff
@@ -138,6 +138,39 @@ const (
...
+	-providerpath /usr/share/bc-fips/bc-fips.jar \
+	-srckeystore /usr/share/elasticsearch/jdk/lib/security/cacerts \
+	-srcstorepass changeit \
+	-srcstoretype jks
+
+echo "Keystore initialization successful."
+
+echo "Initializing Elasticsearch keystore..."
+
+bin/elasticsearch-keystore create -p <<EOF
```
hjiawei, line 167: nit. we can use the absolute path `/usr/share/elasticsearch/bin/elasticsearch-keystore` in case the `WORKDIR` is changed in the future.

pkg/render/logstorage.go
```diff
@@ -570,11 +632,38 @@ func (es elasticsearchComponent) podTemplate() corev1.PodTemplateSpec {
...
+							Key:                  ElasticsearchKeystoreEnvName,
+						},
+					},
+				},
+				{
+					Name:  "ES_JAVA_OPTS",
+					Value: "--module-path /usr/share/bc-fips/",
+				},
+			},
+			Command: []string{"/usr/bin/env", "bash", "-c", fmt.Sprintf(KeystoreInitScript, ElasticsearchKeystoreEnvName, ElasticsearchKeystoreEnvName, ElasticsearchKeystoreEnvName)},
```
Brian-McM, line 661: I can't say I'm a fan of this, I don't think we should be generating these containers on the fly and I don't think the code should live in the operator (the operator is basically injecting code into the container).

If the contents of `KeystoreInitScript` are needed to run the container, then I think it should be a part of the container.

pkg/render/embed/initialize_keystore.sh
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env bash
```
Brian-McM, line 1: I can't say I'm a fan of this living in the operator, I don't believe we should be injecting bash commands into any container from the operator (this is more than just configuring a container to run).

If this is needed to run, then either create a separate init container to run, or add it to the elasticsearch container and put it behind a flag or env variable.

Brian-McM (changes requested):

Brian-McM (approved):

hjiawei (approved):

PR #2146, opened 2022-09-01
Update GO_BUILD_VERSION

rene-dekker:


Brian-McM (approved):

pkg/controller/utils/component_test.go
```diff
@@ -1357,8 +1357,6 @@ func (mc *mockClient) Get(ctx context.Context, key client.ObjectKey, obj client.
 		panic(fmt.Sprintf("mockClient current (%d) call is for %v, not %s", mc.Index, mc.Info[mc.Index].Method, funcName))
 	}
 	if mc.Info[mc.Index].Return == nil {
-		//nolint ignore SA4005 ignore ineffassign
-		obj = mc.Info[mc.Index].Obj
```
tmjd, line 1361: Why remove this? It seems like it was decided it was needed before. Couldn't that assignment have an effect since client.Object is an interface?
I guess it hasn't caused any tests to fail.

rene-dekker: obj is not used after being assigned

I cannot think of any scenario where this line would make a difference. Since the CI passed, I figured there should no longer be a need for it.

tmjd: But if the obj was a pointer then the contents of the pointer was modified, the caller could be effected since this function changed the value.

tmjd: But I'll admit golang passing of pointers vs values isn't as clear to me as it is with C++, so this could be something that hasn't clicked for me.

PR #2160, opened 2022-09-07
Update tigera/api pin

rene-dekker:


pasanw (approved):
LGTM, approved to unblock hash-release in case @tigera/tigera-operator-maintainers do not have time

PR #2167, opened 2022-09-13
Update pins

rene-dekker:
Update tigera/api


tmjd (approved):
LGTM, make sure you squash before (or as part of) merging, that will also ensure picking this is easy.

PR #2201, opened 2022-09-28
chore(AutoDiscoverProvider): implements conflicts discovery

alcidesmig:
## Description

I recently went through a scenario where using an RKE2 cluster it was necessary to create the `GKEBackendConfig` to maintain compatibility during the migration of some services. Because of this, the `networking.gke.io/v1` group was created in the cluster and this caused the tigera/operator to erroneously infer the `kubernetesProvider` field and lead us to a failure scenario.
Considering this scenario and other similar ones, this MR implements the detection of conflicts during the inference of the Kubernetes Provider, generating an error if more than one provider is detected.

## For PR author

- [X] Tests for change.
- [X] If changing pkg/apis/, run `make gen-files`
- [X] If changing versions, run `make gen-versions`

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/utils/discovery.go
+++ b/pkg/controller/utils/discovery.go
@@ -124,10 +124,15 @@ func AutoDiscoverProvider(ctx context.Context, clientset kubernetes.Interface) (
 		detectedProviders = append(detectedProviders, operatorv1.ProviderRKE2)
 	}
 
+	// More than one provider was detected. Can't be sure which one is correct.
 	if len(detectedProviders) > 1 {
 		return operatorv1.ProviderNone, fmt.Errorf(
-			"Failed to assert provider caused by detection of more than one provider. Detected providers: %s",
+			"Failed to assert provider caused by detection of more than one. Detected providers: %s",
 			detectedProviders)
+	}
+	// Only one provider was detected.
+	if len(detectedProviders) == 1 {
+		return detectedProviders[0], nil
 	}
 
 	// Couldn't detect any specific platform.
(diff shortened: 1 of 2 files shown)


caseydavenport:
Well, I don't know how this one slipped through the cracks but it seems reasonable to me. Very sorry about that!

tmjd:
@alcidesmig did you test this in a cluster with things that would have triggered multiple providers?
I think if we merge this then the operator will fail to startup in cases with multiple providers detected. We call the AutoDiscoverProvider in main and if it returns an error then the operator exits.

I think this change is ultimately desired because it would ensure the operator won't try to setup a cluster with an incorrectly detected provider but it needs some additional changes. Those additional changes would be:
1. Not exit [main when AutoDiscoverProvider](https://github.com/tigera/operator/blob/97b1dc7a3093a2b946dabc4d1d57d3d12af7b1f9/cmd/main.go#L362) detects multiple providers
2. Not have the [core controller report degraded](https://github.com/tigera/operator/blob/97b1dc7a3093a2b946dabc4d1d57d3d12af7b1f9/pkg/controller/installation/core_controller.go#L824) because the [auto detected and configured provider do not match](https://github.com/tigera/operator/blob/97b1dc7a3093a2b946dabc4d1d57d3d12af7b1f9/pkg/controller/installation/core_controller.go#L736).

Because I don't think we have heard of issues with incorrect providers being detected, it seems like we should possible keep the validation that if there are multiple auto-detected providers that any provider specified in the Installation CR matches one of the auto-detected providers. I think this would also mean that we need to pass the providers. I think there is probably more impact to this since we most (probably all) of our controllers are passed the detected provider.

PR #2204, opened 2022-10-04
Remove ADJ api pod when fips is enabled

hjiawei:
## Description

This changeset fixes an issue when user flips FIPS mode from enabled
to disabled and then to enabled, the ADJ api pod doesn't get deleted.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/intrusion_detection.go
+++ b/pkg/render/intrusion_detection.go
@@ -207,53 +207,60 @@ func (c *intrusionDetectionComponent) Objects() ([]client.Object, []client.Objec
 	// AD Related deployment only for management/standalone cluster
 	// When FIPS mode is enabled, we currently disable our python based images.
 	if !c.cfg.ManagedCluster {
-		if !operatorv1.IsFIPSModeEnabled(c.cfg.Installation.FIPSMode) {
-			// Service + Deployment + RBAC for AD API
-			objs = append(objs,
-				c.adAPIAllowTigeraPolicy(),
-				c.adAPIServiceAccount(),
-				c.adAPIAccessClusterRole(),
-				c.adAPIAccessRoleBinding(),
-			)
+		var adObjs []client.Object
 
-			shouldConfigureADStorage := len(c.cfg.IntrusionDetection.Spec.AnomalyDetection.StorageClassName) > 0
+		// Service + Deployment + RBAC for AD API
+		adObjs = append(adObjs,
+			c.adAPIAllowTigeraPolicy(),
+			c.adAPIServiceAccount(),
+			c.adAPIAccessClusterRole(),
+			c.adAPIAccessRoleBinding(),
+		)
+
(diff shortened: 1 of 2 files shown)


hjiawei:
Will pick to v1.28 release branch after this is reviewed/merged.

freecaykes (approved):
LGTM

pkg/render/intrusion_detection.go
```diff
@@ -206,41 +206,46 @@ func (c *intrusionDetectionComponent) Objects() ([]client.Object, []client.Objec
...
+				c.adDetectorAllowTigeraPolicy(),
+				c.adDetectorServiceAccount(),
+				c.adDetectorSecret(),
+				c.adDetectorAccessRole(),
+				c.adDetectorRoleBinding(),
+			)
+			objs = append(objs, c.adDetectorPodTemplates()...)
+		} else {
+			objsToDelete = append(objsToDelete, c.adAPIService())
+			objsToDelete = append(objsToDelete, c.adAPIDeployment(shouldConfigureADStorage))
```
freecaykes, line 247: actually maybe it's better to force `shouldConfigureADStorage` to `false`  since if true it will try to grab the Storageclassname value from IDS CR

hjiawei: This is addressed in b0e2ea1e0bdd0cedc01ce30496f314e1f941b2cb.

pkg/render/intrusion_detection.go
```diff
@@ -206,41 +206,48 @@ func (c *intrusionDetectionComponent) Objects() ([]client.Object, []client.Objec
...
+				c.adDetectorSecret(),
+				c.adDetectorAccessRole(),
+				c.adDetectorRoleBinding(),
 			)
+			objs = append(objs, c.adDetectorPodTemplates()...)
+		} else {
+			objsToDelete = append(objsToDelete, c.adAPIService())
+			// Always set shouldConfigureADStorage flag to false so that it won't grab the StorageClassName
+			// from IDS custom resource. Name and namespace identify the resource to be deleted.
+			objsToDelete = append(objsToDelete, c.adAPIDeployment(false))
```
Brian-McM, line 249: Shouldn't you delete everything else that may have been configured, like the bindings / secrets and pod templates? Is there any reason to keep them?

Brian-McM: The persistent volume claim should be added here too, if fips was disabled, then enabled, it wouldn't be cleaned up but you have logic above the shows cleaning it up when it's not needed should be done.

Brian-McM (approved):
So I accidentally approved and merged your other PR for the pick of this but I have comments on deleting resources.

pkg/render/intrusion_detection.go
tmjd: I thought I had commented something similar that we should delete everything that may have been configured. (I'm guessing I lost it during one of my now regular Chrome crashes/kills.)

Brian-McM (approved):

PR #2225, opened 2022-10-17
fix(render/ids): add temp dir for new method of training model creation

freecaykes:
## Description

Add an `emptyDir` for some detectors as a temp folder to write data to for the process of model creation.  Only a temporary directory is required and would be deleted once the detector pod is done running and the model has been committed to the model storage api

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


freecaykes:
not needed - fixed AD pod side

PR #2232, opened 2022-10-21
Prepare v1.27.15

pasanw:


Brian-McM (approved):

config/common_versions.yml
```diff
@@ -1,5 +1,5 @@
-title: v1.1.3
+title: v1.1.4
```
rene-dekker, line 1: Is this also motivated by CVE fixes?

rene-dekker (approved):
lgtm

config/common_versions.yml
pasanw: Yes, we updated key-cert-provisioner to address CVEs

PR #2241, opened 2022-10-26
Support k8s 1.25 for CE 3.15

pasanw:
## Description
Updates tigera/operator to support k8s 1.25 for CE 3.15.

This PR is the primary PR amongst 3 which enable support for k8s 1.25:
* tigera operator: update tigera operator to support k8s 1.25 and update eck-operator to 2.5.0 (which supports k8s 1.25)
* calico private: update the ECK CRDs and versions config for ECK 2.5.0, regenerate manifests
* eck operator build: update to ECK 2.5.0.  
 
### Update tigera operator to support k8s 1.25
* policy/v1beta1 deprecation: fixed regression with PSP exclusion logic
* batch/v1beta1 deprecation: migrated to batch/v1 (supported from k8s 1.21)

### Update to utilize ECK 2.5.0
The following changes were made:
* Updated cloud-on-k8s to 2.5.0 pin
* Updated RBAC in accordance with ECK 2.5.0 RBAC

The following breaking changes occur from ECK 1.8.0 to 2.5.0
* For ES >= 8.0.0, eck-operator defaults to automatically determining whether to set a security context that enables volume access for the ES pod via fsGroup. The automatic determination is based on provider: if OCP is detected, the context is not set - otherwise the context is set. This behaviour seems fine to me, and also does not impact us at the moment since we are on ES 7. I've left a comment for future consideration.
* Init containers will inherit the resource requirements of their main container (e.g. ES). We add multiple init containers to ES that do not require the same amount of resources as ES, and do not specify their own requirements. Given the nature of the pod lifecycle (init containers spawn and exit one-by-one, then the main container spawns), this behaviour is likely fine.

## Testing
Tested install of Standalone CE on a k8s 1.25 and k8s 1.24 with updated eck-operator, manifests.


pkg/apis/addtoscheme_operator_v1.go
```diff
@@ -15,8 +15,8 @@
 package apis
 
 import (
-	esv1 "github.com/elastic/cloud-on-k8s/pkg/apis/elasticsearch/v1"
-	kbv1 "github.com/elastic/cloud-on-k8s/pkg/apis/kibana/v1"
+	esv1 "github.com/elastic/cloud-on-k8s/v2/pkg/apis/elasticsearch/v1"
```
matthewdupre, line 18: I can live with it, but it's a bit weird to have the v2 api imported as esv1

matthewdupre (approved):

pkg/apis/addtoscheme_operator_v1.go
caseydavenport: `/v2/.../v1/` 

What a crazy world we live in.

matthewdupre: Yeah, @pasanw explained to me that I didn't read it carefully enough and the CRDs are still v1...

PR #2243, opened 2022-10-27
prevent operator lockup when unused apiservices are down

ozdanborne:
## Description

This resolves a bug where any APIService in an unhealthy state will cause several TigeraStatuses to enter a `Degraded=true Progressing=false` state with the error `Waiting for LicenseKeyAPI to be ready`. Note this only impacts a few of the Enterprise tigerastatuses, i.e. `compliance`, `intrusion-detection`, `log-collector`, and `manager`

You can reproduce the issue by getting an APIService into `Available=False` . I achieved this by installing the 'metrics-server' on a kind cluster using helm. 
```
$ k get deployment,pod
NAME                             READY   UP-TO-DATE   AVAILABLE   AGE
deployment.apps/metrics-server   0/1     1            0           4h2m

NAME                                  READY   STATUS    RESTARTS   AGE
pod/metrics-server-7d76b744cd-fpw6m   0/1     Running   0          4h2m

$ k get apiservice v1beta1.metrics.k8s.io
NAME                     SERVICE                  AVAILABLE                  AGE
v1beta1.metrics.k8s.io   default/metrics-server   False (MissingEndpoints)   4h3m
```

With that in place, tigerastatuses will enter a degraded state.

You can also trigger the same API failure the Operator is hitting by running `kubectl api-resources`, and alongside the working API's, you'll see an error that says:
```
error: unable to retrieve the complete list of server APIs: metrics.k8s.io/v1beta1: the server is currently unable to handle the request
```

The reason for this failure is that Operator uses the Discovery API to call `ServerPreferredResources` which first gets all API Groups, and then connects to each one individually to get the list of APIs it is hosting. Since the metrics API is down, it can't respond, and our call will fail. A special error which details which ones failed if any of them are unavailable: https://github.com/kubernetes/client-go/blob/v0.25.3/discovery/discovery_client.go#L334

One solution would be to parse that error type and ignore the ones we don't care about, but this PR instead switches to a different discovery API which specifically requests only the API's that we need.

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/utils/utils.go
+++ b/pkg/controller/utils/utils.go
@@ -460,7 +460,7 @@ func WaitToAddResourceWatch(controller controller.Controller, c kubernetes.Inter
 		for obj := range resourcesToWatch {
 			objLog := resourcesToWatch[obj].logger
 			predicateFn := resourcesToWatch[obj].predicate
-			if ok, err := isResourceReady(c, obj.GetObjectKind().GroupVersionKind().Kind); err != nil {
+			if ok, err := isCalicoResourceReady(c, obj.GetObjectKind().GroupVersionKind().Kind); err != nil {
 				objLog.WithValues("Error", err).Info("Failed to check if resource is ready - will retry")
 			} else if !ok {
 				objLog.Info("Waiting for resource to be ready - will retry")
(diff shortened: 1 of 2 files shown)


ozdanborne:
Ignore the `ci/semaphore/push` failed job. It's because I accidentally also pushed to `origin` instead of my fork and CI is now including that status on that check.

tmjd (approved):
LGTM

PR #2283, opened 2022-11-15
Update versions for CE v3.15

pasanw:
Update prometheus version to reflect version tracked by prometheus-docker

Pairs with changes in Calico CE versions (see referenced PR)


rene-dekker (approved):
lgtm

PR #2303, opened 2022-11-22
Add support for Typha graceful shutdown

fasaxc:
## Description

- Change default Typha terminationGracePeriodSeconds and deployment strategy to allow for graceful shutdown.
- Allow override of both of the above.

These changes expose the Typha change here: https://github.com/projectcalico/calico/pull/6973

## For PR author

- [x] Tests for change.

## For PR reviewers

- [x] Appropriate labels:

### DIFF
--- a/api/v1/typha_deployment_types.go
+++ b/api/v1/typha_deployment_types.go
@@ -17,6 +17,7 @@ limitations under the License.
 package v1
 
 import (
+	appsv1 "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
 )
 
@@ -48,7 +49,7 @@ type TyphaDeploymentInitContainer struct {
 	Resources *v1.ResourceRequirements `json:"resources,omitempty"`
 }
 
-// TyphaDeploymentDeploymentPodSpec is the typha Deployment's PodSpec.
+// TyphaDeploymentPodSpec is the typha Deployment's PodSpec.
 type TyphaDeploymentPodSpec struct {
 	// InitContainers is a list of typha init containers.
 	// If specified, this overrides the specified typha Deployment init containers.
@@ -76,6 +77,17 @@ type TyphaDeploymentPodSpec struct {
 	// If omitted, the calico-typha Deployment will use its default value for nodeSelector.
 	// WARNING: Please note that this field will modify the default calico-typha Deployment nodeSelector.
 	NodeSelector map[string]string `json:"nodeSelector,omitempty"`
+
+	// Optional duration in seconds the pod needs to terminate gracefully. May be decreased in delete request.
+	// Value must be non-negative integer. The value zero indicates stop immediately via
+	// the kill signal (no opportunity to shut down).
+	// If this value is nil, the default grace period will be used instead.
+	// The grace period is the duration in seconds after the processes running in the pod are sent
+	// a termination signal and the time when the processes are forcibly halted with a kill signal.
+	// Set this value longer than the expected cleanup time for your process.
+	// Defaults to 30 seconds.
+	// +optional
+	TerminationGracePeriodSeconds *int64 `json:"terminationGracePeriodSeconds,omitempty" protobuf:"varint,4,opt,name=terminationGracePeriodSeconds"`
 
 	// TopologySpreadConstraints describes how a group of pods ought to spread across topology
 	// domains. Scheduler will schedule pods in a way which abides by the constraints.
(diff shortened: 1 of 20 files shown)


pkg/render/common/components/components_test.go
```diff
@@ -555,8 +556,522 @@ var _ = Describe("Common components render tests", func() {
...
+					RollingUpdate: &appsv1.RollingUpdateDeployment{
+						MaxUnavailable: intOrStrPtr("0"),
+						MaxSurge:       intOrStrPtr("100%"),
+					},
+				}))
+			}),
+	)
 })
 
+func intOrStrPtr(v string) *intstr.IntOrString {
```
caseydavenport, line 1066: We have a package in the operator for these sorts of functions already, so might be good to put this one there and use the existing int64 function: https://github.com/tigera/operator/blob/master/pkg/ptr/conversion.go

pkg/render/typha.go
```diff
@@ -346,10 +349,23 @@ func (c *typhaComponent) typhaRole() *rbacv1.ClusterRole {
...
-	var terminationGracePeriod int64 = 0
+	// We set a fairly long grace period by default. Typha sheds load during the grace period rather than
+	// disconnecting all clients at once.
+	var terminationGracePeriod int64 = defaultTyphaTerminationGracePeriod
 	var revisionHistoryLimit int32 = 2
+	// Allowing 1 unavailable Typha by default ensures that we make progress in a cluster with constrained scheduling.
 	maxUnavailable := intstr.FromInt(1)
-	maxSurge := intstr.FromString("25%")
+	// Allowing 100% surge allows a complete replacement fleet of Typha instances to start during an upgrade. When
+	// combined with Typha's graceful shutdown, we get nice emergenet behavior:
```
caseydavenport, line 359: typo: emergenet

pkg/render/typha_test.go
```diff
@@ -646,3 +687,8 @@ var _ = Describe("Typha rendering tests", func() {
 		})
 	})
 })
+
+func intOrStrPtr(v string) *intstr.IntOrString {
```
caseydavenport, line 691: This guy too

pkg/render/common/components/components.go
```diff
@@ -118,10 +125,11 @@ func ApplyDeploymentOverrides(d *appsv1.Deployment, overrides components.Replica
...
 	r := &replicatedPodResource{
-		labels:          d.Labels,
-		annotations:     d.Annotations,
-		minReadySeconds: &d.Spec.MinReadySeconds,
-		podTemplateSpec: &d.Spec.Template,
+		labels:             d.Labels,
+		annotations:        d.Annotations,
+		minReadySeconds:    &d.Spec.MinReadySeconds,
+		podTemplateSpec:    &d.Spec.Template,
+		deploymentStrategy: &d.Spec.Strategy,
```
caseydavenport, line 132: I think we should add MinReadySeconds in `ApplyDaemonSetOverrides` as well: https://github.com/tigera/operator/pull/2303/files#diff-39f0bed05108326edaac1c8f66574d43d5ed843ff91ff6670e813223dac9b3a1R106

DeploymentStrategy doesn't apply there, but if someone wants to extend this to allow configuration of a DaemonSet minReadySeconds, would be nice if it was just a change to the corresponding ReplicatedPodResourceOverrides rather than also needing to make a change here.

Even though we're not using them now, if they get added to a DaemonSet in the future it will "just work" - we've already implemented `Get

caseydavenport (approved):
A few minor comments but otherwise LGTM

pkg/render/common/components/components.go
fasaxc: I think it's already there (was before my change).  Maybe just the formatting changed here to make it look like I added it?

api/v1/typha_deployment_types.go
```diff
@@ -128,6 +140,11 @@ type TyphaDeploymentSpec struct {
 	// Template describes the typha Deployment pod that will be created.
 	// +optional
 	Template *TyphaDeploymentPodTemplateSpec `json:"template,omitempty"`
+
+	// The deployment strategy to use to replace existing pods with new ones.
+	// +optional
+	// +patchStrategy=retainKeys
+	Strategy *appsv1.DeploymentStrategy `json:"strategy,omitempty" patchStrategy:"retainKeys" protobuf:"bytes,4,opt,name=strategy"`
```
tmjd, line 147: I think this means we're allowing Recreate as a strategy, which I don't see any use case for. Possibly allowing a case where typhas can't come up successfully and resulting in outage because all typhas were killed before creating new ones.
I think our intention here was to allow setting MaxUnavailable and MaxSurge on the Rolling updates, so I think we should limit to the RollingUpdate type.

PR #2308, opened 2022-11-23
add gen-files and fix

rene-dekker:
Semaphore now runs gen-files and fix before checking for dirty on static checks job.

Makefile::ci now also includes these goals.


tmjd (approved):
LGTM

PR #2312, opened 2022-11-25
Add functionality to include system root certs into the tigera-ca-bundle

rene-dekker:
There is a private doc motivating this change, please reach out to me.

In short, we want the ability to include system root ca's into the tigera-ca-bundle cm. This PR introduces a new function that - in addition to tigera-ca-bundle.crt - also includes ca-bundle.crt, which is copied from the operator's UBI base image.

This way it is possible to mount everything in /etc/pki/tls/certs/ without overriding and removing the system certs.

### DIFF
--- a/pkg/controller/certificatemanager/certificatemanager.go
+++ b/pkg/controller/certificatemanager/certificatemanager.go
@@ -70,6 +70,8 @@ type CertificateManager interface {
 	GetCertificate(cli client.Client, secretName, secretNamespace string) (certificatemanagement.CertificateInterface, error)
 	// CreateTrustedBundle creates a TrustedBundle, which provides standardized methods for mounting a bundle of certificates to trust.
 	CreateTrustedBundle(certificates ...certificatemanagement.CertificateInterface) certificatemanagement.TrustedBundle
+	// CreateTrustedBundleWithSystemRootCertificates creates a TrustedBundle, which provides standardized methods for mounting a bundle of certificates to trust.
+	CreateTrustedBundleWithSystemRootCertificates(certificates ...certificatemanagement.CertificateInterface) (certificatemanagement.TrustedBundle, error)
 	// AddToStatusManager lets the status manager monitor pending CSRs if the certificate management is enabled.
 	AddToStatusManager(manager status.StatusManager, namespace string)
 	// KeyPair Returns the CA KeyPairInterface, so it can be rendered in the operator namespace.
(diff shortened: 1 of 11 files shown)


pkg/controller/intrusiondetection/intrusiondetection_controller.go
```diff
@@ -460,7 +460,11 @@ func (r *ReconcileIntrusionDetection) Reconcile(ctx context.Context, request rec
 	}
 
 	var esLicenseType render.ElasticsearchLicenseType
-	trustedBundle := certificateManager.CreateTrustedBundle(esgwCertificate)
+	trustedBundle, err := certificateManager.CreateTrustedBundle(false, esgwCertificate)
```
hjiawei, line 463: IDS controller may need the system certs for GTF to talk to external hosts.

hjiawei:
How does this work with BYO certs?

pkg/controller/certificatemanager/certificatemanager.go
```diff
@@ -335,6 +336,43 @@ func HasExpectedDNSNames(secretName, secretNamespace string, cert *x509.Certific
...
+		if err != nil {
+			return nil, err
+		}
+		certificates = append(certificates, systemCertificates...)
+	}
+	return certificatemanagement.CreateTrustedBundle(append([]certificatemanagement.CertificateInterface{cm.keyPair}, certificates...)...), nil
+}
+
+// certFiles is copied from the x509 package.
+// Possible certificate files; stop after finding one.
```
hjiawei, line 351: Maybe include a link to the golang source here in case we need to update this list later?

rene-dekker:
BYO certs are still automatically added to the bundle, just like before.

hjiawei (approved):

pkg/tls/certificatemanagement/certificatebundle.go
```diff
@@ -136,3 +178,32 @@ func (c *certificate) GetName() string {
...
+// The code of this function is loosely based on x509's loadSystemRoots() func:
+// https://go.dev/src/crypto/x509/root_unix.go
+func getSystemCertificates() ([]byte, error) {
+	for _, filename := range certFiles {
+		data, err := os.ReadFile(filename)
+		if err == nil {
+			return data, nil
+		}
+		if err != nil && !os.IsNotExist(err) {
+			return nil, fmt.Errorf(fmt.Sprintf("error occurred when loading system root certificate with name %s", filename), err)
```
Brian-McM, line 205: Nit but you don't need the fmt.Sprintf, Errorf handles formatting.

Brian-McM (approved):

PR #2335, opened 2022-12-13
[RS-647] Adds access to applicationlayer resources

mikestephen:
## Description

This PR is part of the work needed to implement [PMREQ-242](https://tigera.atlassian.net/browse/PMREQ-242). It adds access to the `tigera-network-admin` clusterrole to be able to perform CRUD operations on the `tigera-secure` instance of the `applicationlayer` resource.

I have tested this change on a calient cluster, and was able to perform the REST calls the UI will need to perform in order to enable/disable WAF.

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


matthewdupre (commented):
One small concern here: in Calico Cloud this may mean that we'll need to review any future field additions to this resource (ApplicationLayer) for whether they're appropriate for end user use.

For example: if we add a field that significantly impacts CPU/memory usage or limits, this might allow users to impact their service availability.

I did look at the current resource spec, and it doesn't seem harmful right now, but I do have some concern that this could create future problems that we couldn't reasonably rely on catching in review?  WDYT?  I might also have misunderstood this PR and it doesn't apply to management clusters - this wasn't a thorough review!

mikestephen:
> One small concern here: in Calico Cloud this may mean that we'll need to review any future field additions to this resource (ApplicationLayer) for whether they're appropriate for end user use.
> 
> For example: if we add a field that significantly impacts CPU/memory usage or limits, this might allow users to impact their service availability.
> 
> I did look at the current resource spec, and it doesn't seem harmful right now, but I do have some concern that this could create future problems that we couldn't reasonably rely on catching in review? WDYT? I might also have misunderstood this PR and it doesn't apply to management clusters - this wasn't a thorough review!

This would apply to management clusters I think, but I'm not really understanding the issue.
The user can already alter this resource using kubectl I think, so this just exposes it through the UI.
Are you thinking that exposing it through the UI could be harmful?

pkg/render/apiserver.go
```diff
@@ -1607,6 +1607,13 @@ func (c *apiServerComponent) tigeraNetworkAdminClusterRole() *rbacv1.ClusterRole
 			Verbs:         []string{"*"},
 			ResourceNames: []string{"cluster-settings", "user-settings"},
 		},
+		// Allow the user to read and write the applicationlayers resource to enable/disable WAF.
```
tmjd, line 1610: I'm not sure but I think this should probably be a different role. And I'm not sure that the operator is the right place to be creating or giving access to it.
I don't think the same users that have tigera-network-admin should have access to configuring the operator and what it deploys.
It is a bit of a circular dependency to have the operator granting permissions to control the operator.

mikestephen: As discussed with @tigeraphil and @malavikabala and ourselves, PM are ok for these permissions are ok to go into `tigera-network-admin` for now, possibly revisited in a future release.

caseydavenport:
> One small concern here: in Calico Cloud this may mean that we'll need to review any future field additions to this resource (ApplicationLayer) for whether they're appropriate for end user use.

It's part of the API, it needs to be suitable for end-user use. 

Whether or not it's something configurable by any user of the UI, or only a special subset of users, is another question controlled by RBAC. Same goes for whether or not it's present in the management cluster or hte managed cluster (this would be a managed cluster resource)

tmjd (approved):
LGTM

PR #2336, opened 2022-12-14
Egress Gateway - operator changes

sridhartigera:
## Description
This PR adds the following
1. New API to configure Egress Gateways
2. Added a new egress gateway controller.
3. Updated installation resource to update version of calico installed.
4. Sequence EGW upgrade - EGW should be upgraded only after calico is upgraded.
5. UTs for new controller, render package.

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


pkg/render/egressgateway/egressgateway.go
```diff
@@ -0,0 +1,244 @@
...
+func (c *component) egwVolume() *corev1.Volume {
+	return &corev1.Volume{
+		Name:         "policysync",
+		VolumeSource: corev1.VolumeSource{CSI: &corev1.CSIVolumeSource{Driver: "csi.tigera.io"}},
+	}
+}
+
+func (c *component) egwVolumeMounts() []corev1.VolumeMount {
+	return []corev1.VolumeMount{
+		corev1.VolumeMount{Name: "policysync", MountPath: "/var/run"},
```
fasaxc, line 214: Did we always mount to `/var/run` ?  I'd have expected something like `/var/run/calico`  Perhaps now would be a good time to change that (and update the egressd arg to match).

fasaxc (commented):
This looks the right sort of shape to me and the rendering section looks how I was expecting :+1: Think you still need an operator expert to review the detail (and you already flagged the need to deal with RKE) but LGTM so far.

pkg/controller/egressgateway/egressgateway_controller.go
```diff
@@ -0,0 +1,609 @@
...
+	// yellow. Though yellow is reconciled and rendered successfully, Tigerastatus should still be degraded as Red, Blue are
+	// degraded. Now lets assume Blue gets updated and gets rendered properly. In this case, Tigerastatus should still be
+	// degraded as Red is unhealthy.
+	egws, err := getEgressGateways(ctx, r.client, reconcile.Request{NamespacedName: types.NamespacedName{Namespace: ""}})
+	if err != nil {
+		// Leave the status as it is and return
+		return reconcile.Result{}, nil
+	}
+
+	// Get the cumulative status of all Egress Gateway resources.
```
rene-dekker, line 308: Would it be possible to only do this once in total and do it outside this func in `Reconcile()`?

pkg/controller/egressgateway/egressgateway_suit_test.go
```diff
@@ -0,0 +1,29 @@
+// Copyright (c) 2022 Tigera, Inc. All rights reserved.
```
rene-dekker, line 1: suite_test * file name typo

pkg/controller/installation/core_controller.go
```diff
@@ -1446,6 +1447,7 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 	} else {
 		instance.Status.ImageSet = imageSet.Name
 	}
+	instance.Status.Version = releaseToBeInstalled
```
rene-dekker, line 1450: See my comment on getCalicoVersion, I think this may not be necessary.

pkg/crds/operator/operator.tigera.io_installations.yaml
```diff
@@ -12708,6 +12708,9 @@ spec:
                 - Calico
                 - TigeraSecureEnterprise
                 type: string
+              version:
```
rene-dekker, line 12711: May not be necessary, see my comment on getCalicoVersion.

pkg/render/egressgateway/egressgateway_suit_test.go
```diff
@@ -0,0 +1,29 @@
+// Copyright (c) 2022 Tigera, Inc. All rights reserved.
```
rene-dekker, line 1: typo in file name

pkg/render/egressgateway/egressgateway.go
```diff
@@ -0,0 +1,404 @@
...
+	corev1 "k8s.io/api/core/v1"
+	policyv1beta1 "k8s.io/api/policy/v1beta1"
+	rbacv1 "k8s.io/api/rbac/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	EGWPortName               = "health"
```
rene-dekker, line 39: Does not need to be exported.

pkg/render/egressgateway/egressgateway.go
```diff
@@ -0,0 +1,404 @@
...
+
+type component struct {
+	config *Config
+}
+
+// Config contains all the config information EgressGateway needs to render component.
+type Config struct {
+	PullSecrets  []*corev1.Secret
+	Installation *operatorv1.InstallationSpec
+	OsType       rmeta.OSType
```
rene-dekker, line 61: OS should be capitalized.

pkg/render/egressgateway/egressgateway.go
```diff
@@ -0,0 +1,404 @@
...
+	return nil
+}
+
+func (c *component) SupportedOSType() rmeta.OSType {
+	return rmeta.OSTypeLinux
+}
+
+func (c *component) Objects() ([]client.Object, []client.Object) {
+	objects := []client.Object{}
+	if c.config.Provider == operatorv1.ProviderRKE2 {
```
rene-dekker, line 93: See how this is done for example in compliance_controller.go and compliance.go. UsePSP should be passed as a boolean and is determined by the reconciler options.

For openshift, we might need security constraints.

pkg/render/egressgateway/egressgateway.go
```diff
@@ -0,0 +1,404 @@
...
+	d := appsv1.Deployment{
+		TypeMeta: metav1.TypeMeta{Kind: "Deployment", APIVersion: "apps/v1"},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      c.config.EgressGW.Name,
+			Namespace: c.config.EgressGW.Namespace,
+			Labels:    c.config.EgressGW.Spec.Template.Metadata.Labels,
+		},
+		Spec: appsv1.DeploymentSpec{
+			Replicas: c.config.EgressGW.Spec.Replicas,
+			Selector: &metav1.LabelSelector{MatchLabels: map[string]string{"k8s-app": c.config.EgressGW.Name}},
```
rene-dekker, line 117: Does this still work if the user has specified `EgressGatewayMetadata.Labels` ?

pkg/render/egressgateway/egressgateway.go
```diff
@@ -0,0 +1,404 @@
...
+}
+
+func (c *component) egwInitContainer() *corev1.Container {
+	initContainerPrivileges := true
+	return &corev1.Container{
+		Name:            "egress-gateway-init",
+		Image:           c.config.egwImage,
+		ImagePullPolicy: corev1.PullIfNotPresent,
+		Command:         []string{"/init-gateway.sh"},
+		SecurityContext: &corev1.SecurityContext{Privileged: &initContainerPrivileges},
```
rene-dekker, line 167: We have a func called BoolToPtr() in the ptr package.

pkg/render/egressgateway/egressgateway.go
```diff
@@ -0,0 +1,404 @@
...
+		InitialDelaySeconds: 3,
+		TimeoutSeconds:      1,
+		SuccessThreshold:    1,
+		PeriodSeconds:       3,
+	}
+}
+
+func (c *component) egwSecurityContext() *corev1.SecurityContext {
+	return &corev1.SecurityContext{
+		Capabilities: &corev1.Capabilities{
```
rene-dekker, line 204: Would it be possible to specify the other fields as well?

Especially: RunAsUser, RunAsGroup, RunAsNonRoot, Privileged

For other EE components we are normalizing all of the users and groups to 10001:10001. It would be nice  if EGW would do the same.

pkg/render/egressgateway/egressgateway.go
```diff
@@ -0,0 +1,404 @@
...
+		{
+			APIGroups:     []string{"policy"},
+			Resources:     []string{"podsecuritypolicies"},
+			ResourceNames: []string{"tigera-egress-gateway"},
+			Verbs:         []string{"use"},
+		},
+	}
+	return &rbacv1.ClusterRole{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "tigera-egress-gateway",
```
rene-dekker, line 303: Would be nice to extract a const and share it with the roleref in the binding.

pkg/render/egressgateway/egressgateway.go
```diff
@@ -0,0 +1,404 @@
...
+		},
+		RoleRef: rbacv1.RoleRef{
+			APIGroup: "rbac.authorization.k8s.io",
+			Kind:     "ClusterRole",
+			Name:     "tigera-egress-gateway",
+		},
+		Subjects: []rbacv1.Subject{
+			{
+				Kind:      "ServiceAccount",
+				Name:      "tigera-egress-gateway",
```
rene-dekker, line 322: Can there be more than 1 egw in the same namespace and is the name always this value?
Otherwise, this should be EgressGW.name

pkg/render/egressgateway/egressgateway.go
```diff
@@ -0,0 +1,404 @@
...
+	return ""
+}
+
+func (c *component) getHealthTimeoutDs() string {
+	egw := c.config.EgressGW
+	return fmt.Sprintf("%ds", *egw.Spec.EgressGatewayFailureDetection.HealthTimeoutDataStoreSeconds)
+}
+
+func concatString(arr []string) string {
+	ret := "["
```
rene-dekker, line 395: golang has a builder object:
```
	sb := strings.Builder{}
	sb.WriteString( ... )
```

rene-dekker (changes requested):

pkg/controller/egressgateway/egressgateway_controller.go
```diff
@@ -0,0 +1,609 @@
...
+	// Update the status of this CR.
+	egw.Status.State = operatorv1.TigeraStatusReady
+	setAvailable(r.client, ctx, egw, string(operatorv1.AllObjectsAvailable), "All objects available")
+
+	// After the resource has been created/updated, Tigerastatus needs to be set by taking the cumulative status of the
+	// available EGW resources. Lets say we create 2 EGW resources Red, Blue. Both are degraded. Now lets create 3rd resource
+	// yellow. Though yellow is reconciled and rendered successfully, Tigerastatus should still be degraded as Red, Blue are
+	// degraded. Now lets assume Blue gets updated and gets rendered properly. In this case, Tigerastatus should still be
+	// degraded as Red is unhealthy.
+	egws, err := getEgressGateways(ctx, r.client, reconcile.Request{NamespacedName: types.NamespacedName{Namespace: ""}})
```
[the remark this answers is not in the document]
sridhartigera: If the reconcile request is to a single EGW resource, we wont fetch all the EGW resources. Hence we need to fetch all the EGWs again.

pkg/render/egressgateway/egressgateway.go
sridhartigera: Yeah. It works fine even if we specify labels.

pkg/controller/installation/core_controller.go
sridhartigera: As discussed, we still need this.

pkg/crds/operator/operator.tigera.io_installations.yaml
sridhartigera: As discussed we need this field to sequence component upgrades.

pkg/controller/egressgateway/egressgateway_controller.go
```diff
@@ -0,0 +1,609 @@
...
+			return reconcile.Result{}, nil
+		}
+		reqLogger.Error(err, "Error querying installation")
+		r.status.SetDegraded(operatorv1.ResourceReadError, "Error querying installation", err, reqLogger)
+		setDegraded(r.client, ctx, egw, reconcileErr,
+			fmt.Sprintf("Error querying installation Name = %s, Namespace = %s, err = %s", egw.Name, egw.Namespace, err.Error()))
+		return reconcile.Result{}, err
+	}
+
+	if variant != operatorv1.TigeraSecureEnterprise {
```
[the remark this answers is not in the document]
sridhartigera: I moved the check to ```Reconcile()``` so that I do it only for a single EGW. w.r.t the message, this is what I see in all the other controllers.

pkg/controller/egressgateway/egressgateway_controller.go
```diff
@@ -0,0 +1,609 @@
...
+			// Owned objects are automatically garbage collected. For additional cleanup logic use finalizers.
+			reqLogger.Info("EgressGateway object not found")
+			// Since the EGW resource is not found, remove the deployment.
+			r.status.RemoveDeployments(types.NamespacedName{Name: request.Name, Namespace: request.Namespace})
+			// Get the cumulative Egress Gateway status. Lets say we have 2 EGW resources red and blue.
+			// Red has already degraded. When the user deletes Red, Tigerastatus should go back to available
+			// as Blue is healthy. Hence get the cumulative EGW status. If all the EGWs are ready, clear the degraded
+			// Tigerastatus. If at least one of the EGWs is unhealthy, get the degraded msg from the conditions and
+			// update the Tigerastatus.
+			egws, err := getEgressGateways(ctx, r.client, reconcile.Request{NamespacedName: types.NamespacedName{Namespace: ""}})
```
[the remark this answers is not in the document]
sridhartigera: Once we delete a resource, we need to get all the existing EGWs and find out the cumulative status. Once a resource is deleted, controller needs to get the status of all other existing EGW resource and if all resources are ```Ready```, clear the degraded status. If one or more EGW resources are degraded, find the first degraded EGW resource and get the degraded reason and update Tigerastatus condition.

api/v1/egressgateway_types.go
```diff
@@ -0,0 +1,274 @@
...
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:validation:Maximum=2147483647
+	// +required
+	Replicas *int32 `json:"replicas"`
+
+	// IPPools defines the IP Pools that the Egress Gateway pods should be using.
+	// Either name or CIDR must be specified.
+	// IPPools must match an existing IPPools.
+	// +required
+	IPPools []EgressGatewayIPPool `json:"ipPools"`
```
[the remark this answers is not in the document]
sridhartigera: This is a required field. Guess the wording in design doc is confusing.

pkg/render/egressgateway/egressgateway.go
sridhartigera: Have addressed UsePSP and have done it similar to compliance. Need to check about openshift.

api/v1/egressgateway_types.go
```diff
@@ -0,0 +1,274 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// EgressGatewaySpec defines the desired state of EgressGateway
+type EgressGatewaySpec struct {
+	// Replicas defines how many instances of the Egress Gateway pod will run.
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:validation:Maximum=2147483647
+	// +required
+	Replicas *int32 `json:"replicas"`
```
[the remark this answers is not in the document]
sridhartigera: Added it as a pointer to make it consistent with other fields. We did not consider the option of making this an optional field.

api/v1/egressgateway_types.go
```diff
@@ -0,0 +1,274 @@
...
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Maximum=2147483647
+	// +optional
+	HealthTimeoutDataStoreSeconds *int32 `json:"healthTimeoutDataStoreSeconds,omitempty"`
+
+	// ICMPProbes define outgoing ICMP probes that Egress Gateway will use to
+	// verify its upstream connection. Egress Gateway will report not ready if all
+	// fail. Timeout must be greater than interval.
+	// +optional
+	ICMPProbes *ICMPProbes `json:"icmpProbe,omitempty"`
```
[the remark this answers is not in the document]
sridhartigera: I think it is because we specify multiple IPs, URLs. I had it as a singular and changed it as per a comment in the design.

api/v1/egressgateway_types.go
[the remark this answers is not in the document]
rene-dekker: If it is required, then I suppose you also want to add validation that the len is 1 or more.

api/v1/egressgateway_types.go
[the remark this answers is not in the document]
rene-dekker: [egw-operator](https://github.com/sridhartigera/operator/tree/egw-operator)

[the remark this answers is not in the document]
rene-dekker: If this is required, then I suppose you want to also add validation that the length is at least 1.

api/v1/egressgateway_types.go
```diff
@@ -0,0 +1,277 @@
...
+
+	// HTTPProbes define outgoing HTTP probes that Egress Gateway will use to
+	// verify its upsteam connection. Egress Gateway will report not ready if all
+	// fail. Timeout must be greater than interval.
+	// +optional
+	HTTPProbes *HTTPProbes `json:"httpProbes,omitempty"`
+}
+
+// ICMPProbes defines the ICMP probe configuration for Egress Gateway.
+type ICMPProbes struct {
```
rene-dekker, line 167: ICMPProbe, should be singular

pkg/controller/egressgateway/egressgateway_controller.go
```diff
@@ -0,0 +1,646 @@
...
+// limitations under the License.
+
+package egressgateway
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	operatorv1 "github.com/tigera/operator/api/v1"
```
rene-dekker, line 22: imports need to be sorted

pkg/controller/egressgateway/egressgateway_controller.go
```diff
@@ -0,0 +1,646 @@
...
+// based on the state read and what is in the EgressGateway.Spec.
+func (r *ReconcileEgressGateway) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
+	reqLogger := log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
+	reqLogger.Info("Reconciling EgressGateway")
+
+	// Wait for right version of the calico to be installed. When upgrading EGWs, calico-node needs to be upgraded
+	// first, before proceeding with the EGW upgrade. Hence wait for the right version of calico to be installed.
+	installStatus, err := utils.GetInstallationStatus(ctx, r.client)
+	if err != nil || installStatus.CalicoVersion != components.EnterpriseRelease {
+		return reconcile.Result{RequeueAfter: 30 * time.Second}, nil
```
rene-dekker, line 147: I'd prefer:
```
if != nil {
// lset degraded
// return error
} else if installStatus.CalicoVersion != components.EnterpriseRelease {
// set degraded with informative log message saying you are waiting for the version to match
// return error
}

```

rene-dekker (changes requested):

pkg/controller/egressgateway/egressgateway_controller.go
```diff
@@ -0,0 +1,646 @@
...
+	if err != nil {
+		return []operatorv1.EgressGateway{}, err
+	}
+	return []operatorv1.EgressGateway{*instance}, err
+
+}
+
+// fillDefaults sets the default values of the EGW resource.
+func fillDefaults(egw *operatorv1.EgressGateway, installation *operatorv1.InstallationSpec) {
+	defaultLogSeverity := operatorv1.LogLevelInfo
```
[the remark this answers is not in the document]
sridhartigera: These can't be constants as these need to passed as pointers.

pkg/controller/egressgateway/egressgateway_controller.go
```diff
@@ -0,0 +1,666 @@
...
+	}
+	r.status.Run(opts.ShutdownContext)
+	return r
+}
+
+// add adds watches for resources that are available at startup.
+// Watching namespaced resources must be avoided as the controller
+// can't differentiate if the request namespaced resource is an
+// Egress Gateway resource or not.
+func add(mgr manager.Manager, c controller.Controller) error {
```
rene-dekker, line 102: You can wildcard manager, as it is unused.

```
func add(_ manager.Manager, c controller.Controller) error {}
```

pkg/controller/egressgateway/egressgateway_controller.go
```diff
@@ -0,0 +1,666 @@
...
+			// Owned objects are automatically garbage collected. For additional cleanup logic use finalizers.
+			reqLogger.Info("EgressGateway object not found")
+			// Since the EGW resource is not found, remove the deployment.
+			r.status.RemoveDeployments(types.NamespacedName{Name: request.Name, Namespace: request.Namespace})
+			// Get the cumulative Egress Gateway status. Let's say we have 2 EGW resources red and blue.
+			// Red has already degraded. When the user deletes Red, TigeraStatus should go back to available
+			// as Blue is healthy. If all the EGWs are ready, clear the degraded TigeraStatus.
+			// If at least one of the EGWs is unhealthy, get the degraded msg from the conditions and
+			// update the TigeraStatus.
+			status, egw := getCumulativeEgressGatewayStatus(egws)
```
rene-dekker, line 193: status collides with package name "github.com/tigera/operator/pkg/controller/status"

pkg/controller/egressgateway/egressgateway_controller.go
```diff
@@ -0,0 +1,666 @@
...
+	crdv1 "github.com/tigera/operator/pkg/apis/crd.projectcalico.org/v1"
+	"github.com/tigera/operator/pkg/components"
+
+	"github.com/tigera/operator/pkg/controller/options"
+	"github.com/tigera/operator/pkg/controller/status"
+	"github.com/tigera/operator/pkg/controller/utils"
+	"github.com/tigera/operator/pkg/controller/utils/imageset"
+	rmeta "github.com/tigera/operator/pkg/render/common/meta"
+	"github.com/tigera/operator/pkg/render/egressgateway"
+	errors "k8s.io/apimachinery/pkg/api/errors"
```
rene-dekker, line 34: redundant package name

pkg/controller/egressgateway/egressgateway_controller.go
```diff
@@ -0,0 +1,666 @@
...
+					return fmt.Errorf("AWS subnet ID must be set when NativeIP is enabled")
+				}
+				return nil
+			}
+		}
+	}
+	return fmt.Errorf("IPPool matching CIDR = %s not present", ipPool.CIDR)
+}
+
+func getCumulativeEgressGatewayStatus(egws []operatorv1.EgressGateway) (bool, *operatorv1.EgressGateway) {
```
rene-dekker, line 611: The func name does not align with what it is doing. It is returning an EgressGateway, not a status. 

And the boolean seems redundant as it is always equal to (ewg !=nil)

Should it be named getUnreadyEgressGateway() or something like that?

pkg/controller/egressgateway/egressgateway_controller.go
```diff
@@ -0,0 +1,666 @@
...
+		for _, egw := range egwsToReconcile {
+			r.setDegraded(ctx, &egw, "Error reading felix configuration", err)
+		}
+		return reconcile.Result{}, err
+	}
+
+	// Reconcile all the EGWs
+	var errMsgs []string
+	for _, egw := range egwsToReconcile {
+		_, err := r.reconcileEgressGateway(ctx, &egw, reqLogger, variant, fc, pullSecrets, installation, status, degradedEGW)
```
rene-dekker, line 275: Could also be `_, err = `

rene-dekker (approved):
lgtm

pkg/controller/egressgateway/egressgateway_controller.go
```diff
@@ -0,0 +1,666 @@
...
+		reqLogger.Error(err, fmt.Sprintf("Waiting for network to be %s", operatorv1.TigeraSecureEnterprise))
+		r.status.SetDegraded(operatorv1.ResourceNotReady, fmt.Sprintf("Waiting for network to be %s", operatorv1.TigeraSecureEnterprise), nil, reqLogger)
+		for _, egw := range egwsToReconcile {
+			r.setDegraded(ctx, &egw, fmt.Sprintf("Waiting for network to be %s", operatorv1.TigeraSecureEnterprise), nil)
+		}
+		return reconcile.Result{}, nil
+	}
+
+	if installStatus.CalicoVersion != components.EnterpriseRelease {
+		return reconcile.Result{RequeueAfter: 30 * time.Second}, nil
```
tmjd, line 248: Seems like we should provide a reason why we're not doing anything. I'm not sure if Degraded is appropriate and I realize I don't think we have a way to set explicit Progressing and a reason.

Do we need to requeue here because status updates don't trigger a reconcile?

tmjd: It should be clear if the operator is stuck waiting for this condition, and shouldn't be silent as far as TigeraStatus.

pkg/controller/utils/utils.go
```diff
@@ -352,14 +352,27 @@ func GetAuthentication(ctx context.Context, cli client.Client) (*operatorv1.Auth
...
 	if err := client.Get(ctx, DefaultInstanceKey, instance); err != nil {
-		return nil, err
+		return nil, nil, err
 	}
-	return &instance.Status, nil
+
+	spec := instance.Spec
+
+	// update Installation with 'overlay'
+	overlay := operatorv1.Installation{}
```
tmjd, line 366: I don't like the duplication of this code with GetInstallation. The comment doesn't mention that it accounts for overlays.
Is the rationale that it is a good trade off to duplicate this code to avoid a client.Get for the same resource?

sridhartigera: Yes. The rationale was to avoid another client.Get. @rene-dekker

rene-dekker: It was my suggestion to return both the status and the installation object. I think it is costly to hold up the PR over this though, so if you think it is a bad idea, I am ok to either:
- make two calls: GetInstallation + a call for the status.
- have the GetInstallation reuse this func to reduce the overlay code duplication.
- add an extra returned object to getInstallation(). (Maybe that would be best done in a follow up PR)

api/v1/egressgateway_types.go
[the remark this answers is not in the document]
caseydavenport: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required

This discusses optional vs. required - if it's required, it should _not_ be a pointer.

However, I would question this being a required field - why not make it optional, and default to a single replica?

api/v1/egressgateway_types.go
```diff
@@ -0,0 +1,277 @@
...
+
+	// IPPools defines the IP Pools that the Egress Gateway pods should be using.
+	// Either name or CIDR must be specified.
+	// IPPools must match existing IPPools.
+	// +required
+	IPPools []EgressGatewayIPPool `json:"ipPools"`
+
+	// LogSeverity defines the logging level of the Egress Gateway.
+	// Default: Info
+	// +kubebuilder:validation:Enum=Trace;Debug;Info;Warn;Error;Fatal
```
caseydavenport, line 40: We should put a kubebuilder defaulting tag on this.

api/v1/egressgateway_types.go
```diff
@@ -0,0 +1,277 @@
...
+
+const (
+	NativeIPEnabled  NativeIP = "Enabled"
+	NativeIPDisabled NativeIP = "Disabled"
+)
+
+type LogLevel string
+
+const (
+	LogLevelTrace LogLevel = "Trace"
```
caseydavenport, line 132: Do we actually have a meaningful distinction between Trace and Debug? Most of our components do not, so I think I'd recommend dropping "Trace" and just using "Debug" to mean the most verbose log level.

api/v1/egressgateway_types.go
```diff
@@ -0,0 +1,277 @@
...
+)
+
+// EgressGatewayFailureDetection defines the fields the needed for determining Egress Gateway
+// readiness.
+type EgressGatewayFailureDetection struct {
+
+	// HealthTimeoutDataStoreSeconds defines how long Egress Gateway can fail to connect
+	// to the datastore before reporting not ready.
+	// This value must be greater than 0.
+	// Default: 90
```
caseydavenport, line 147: We should have a `kubebuilder:default` tag here.

api/v1/egressgateway_types.go
```diff
@@ -0,0 +1,277 @@
...
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Maximum=2147483647
+	// +optional
+	HealthTimeoutDataStoreSeconds *int32 `json:"healthTimeoutDataStoreSeconds,omitempty"`
+
+	// ICMPProbe define outgoing ICMP probes that Egress Gateway will use to
+	// verify its upstream connection. Egress Gateway will report not ready if all
+	// fail. Timeout must be greater than interval.
+	// +optional
+	ICMPProbe *ICMPProbe `json:"icmpProbe,omitempty"`
```
caseydavenport, line 157: Don't we need one of ICMPProbe / HTTPProbe to be defined? Is it valid to have both of them nil? I would have thought that's an invalid config, and if so we should call it out in the comments.

api/v1/egressgateway_types.go
```diff
@@ -0,0 +1,274 @@
...
+	// fail. Timeout must be greater than interval.
+	// +optional
+	HTTPProbes *HTTPProbes `json:"httpProbe,omitempty"`
+}
+
+// ICMPProbes defines the ICMP probe configuration for Egress Gateway.
+type ICMPProbes struct {
+	// IPs define the list of ICMP probe IPs. Egress Gateway will probe each IP
+	// periodically. If all probes faile, Egress Gateway will report non-ready.
+	// +optional
```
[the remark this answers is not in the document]
caseydavenport: > If the user adds ICMPProbes, but leaves the IPs empty, normally you'd wanna flag it as a misconfiguration.

Shouldn't this be required? How can we do an ICMP probe without an IPs to probe?

pkg/controller/utils/utils.go
tmjd: I think it is a great idea to have GetInstallation call this function. I can totally be on board with that. Plus a comment update that this function handles the the overlay also.

api/v1/egressgateway_types.go
```diff
@@ -0,0 +1,277 @@
...
+	// +kubebuilder:validation:Maximum=2147483647
+	// +optional
+	TimeoutSeconds *int32 `json:"timeoutSeconds"`
+}
+
+// HTTPProbe defines the HTTP probe configuration for Egress Gateway.
+type HTTPProbe struct {
+	// URLs define the list of HTTP probe URLs. Egress Gateway will probe each URL
+	// periodically.If all probes fail, Egress Gateway will report non-ready.
+	// +optional
```
caseydavenport, line 192: Same here - shouldn't this be required?

api/v1/egressgateway_types.go
```diff
@@ -0,0 +1,277 @@
...
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []EgressGateway `json:"items"`
+}
+
+func (c *EgressGateway) GetLogSeverity() string {
+	return string(*c.Spec.LogSeverity)
+}
+
+func (c *EgressGateway) GetTerminationGracePeriod() *int64 {
```
caseydavenport, line 259: I think you want to do nil checking on these, right? 

See typha as an example: https://github.com/tigera/operator/blob/74f866faa72aa7be30583b207c7a05fc26540809/api/v1/typha_deployment_types.go#L269-L275

pkg/render/egressgateway/egressgateway.go
```diff
@@ -0,0 +1,438 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Annotations: c.egwBuildAnnotations(),
+			Labels:      c.config.EgressGW.Spec.Template.Metadata.Labels,
+		},
+		Spec: corev1.PodSpec{
+			ImagePullSecrets:              ps,
+			Affinity:                      c.config.EgressGW.GetAffinity(),
+			TopologySpreadConstraints:     c.config.EgressGW.GetTopoConstraints(),
+			NodeSelector:                  c.config.EgressGW.GetNodeSelector(),
+			TerminationGracePeriodSeconds: c.config.EgressGW.GetTerminationGracePeriod(),
```
caseydavenport, line 144: You don't need to put these here.

Instead, you should call this function so that the overrides machinery handles this for you: https://github.com/tigera/operator/blob/74f866faa72aa7be30583b207c7a05fc26540809/pkg/render/typha.go#L429-L431

This will make EGW automatically pick up any new fields that get added in the future. 

Please take a look at how the Typha render logic does this today and model EGW after it.

pkg/controller/egressgateway/egressgateway_controller.go
caseydavenport: Agreed - degraded definitely isn't appropriate, but we should at a minimum log out a reason here explaining that we're waiting for the roll-out to occur, and probably should set a TigeraStatus message to explain that we're waiting for calico-node to finish upgrading.

pkg/controller/egressgateway/egressgateway_controller.go
```diff
@@ -0,0 +1,666 @@
...
+
+func (r *ReconcileEgressGateway) setDegraded(ctx context.Context, egw *operatorv1.EgressGateway, msg string, err error) {
+	if err != nil {
+		setDegraded(r.client, ctx, egw, reconcileErr, fmt.Sprintf("%s Name = %s, Namespace = %s, err = %s", msg, egw.Name, egw.Namespace, err.Error()))
+	} else {
+		setDegraded(r.client, ctx, egw, reconcileErr, fmt.Sprintf("%s Name = %s, Namespace = %s", msg, egw.Name, egw.Namespace))
+	}
+}
+
+func (r *ReconcileEgressGateway) reconcileEgressGateway(ctx context.Context, egw *operatorv1.EgressGateway, reqLogger logr.Logger,
```
caseydavenport, line 295: This function has a massive signature. Any idea on how we can reduce the number of required arguments?

caseydavenport: This function shouldn't be returning a reconcile.Result, as it's not the main reconcile function.

Really, it should return just an error, and then the calling code should handle it IMO

caseydavenport: In fact, the reconcile.Result appears to be unused where this function is called anyway.

pkg/controller/egressgateway/egressgateway_controller.go
```diff
@@ -0,0 +1,666 @@
...
+	// Update the status of this CR.
+	egw.Status.State = operatorv1.TigeraStatusReady
+	setAvailable(r.client, ctx, egw, string(operatorv1.AllObjectsAvailable), "All objects available")
+
+	// After the resource is created/updated, TigeraStatus needs to be set by taking the cumulative status of the
+	// available EGW resources. Lets say we create 2 EGW resources Red, Blue. Both are degraded. Now lets create 3rd resource
+	// yellow. Though yellow is reconciled and rendered successfully, TigeraStatus should still be degraded as Red, Blue are
+	// degraded. Now lets assume Blue gets updated and gets rendered properly. In this case, TigeraStatus should still be
+	// degraded as Red is unhealthy. "status" represents, the cumulative status of EGWs.
+	if unreadyEGW != nil {
```
caseydavenport, line 375: I don't think this should be done inside of this function - we call this for every EGW that exists in the cluster, and there's no point in checking the same single unreadyEGW for all of them.

We should handle unreadyEGWs separately.

I'm not sure why we need to handle unready EGWs in the controller at all - the status manager is meant to handle that itself by monitoring deployment statuses - is this logic even needed?

pkg/controller/egressgateway/egressgateway_controller.go
```diff
@@ -0,0 +1,666 @@
...
+	if !r.status.IsAvailable() {
+		// Schedule a kick to check again in the near future, hopefully by then things will be available.
+		return reconcile.Result{RequeueAfter: 30 * time.Second}, nil
+	}
+	return reconcile.Result{}, nil
+}
+
+// getRequestedEgressGateway returns the namespaced EgressGateway instance.
+func getRequestedEgressGateway(egws []operatorv1.EgressGateway, request reconcile.Request) (*operatorv1.EgressGateway, int) {
+	for index, egw := range egws {
```
caseydavenport, line 392: Why do we iterate over the list rather than just querying it using the client? 

The client uses a cache, so it won't generate a new request unless it needs to. We should probably refactor this to use the client unless there's a reason I'm not seeing.

pkg/controller/egressgateway/egressgateway_controller.go
```diff
@@ -0,0 +1,666 @@
...
+		egw.Status.Conditions = append(egw.Status.Conditions, condition)
+	}
+	if updateStatus {
+		if err := cli.Status().Update(ctx, egw); err != nil {
+			log.WithValues("Name", egw.Name, "Namespace", egw.Namespace, "error", err).Info("Error updating status")
+		}
+	}
+}
+
+func getDegradedMsg(egw *operatorv1.EgressGateway) string {
```
caseydavenport, line 659: We shouldn't need this function, IIUC. If we do the status handling properly in the status manager

pkg/controller/egressgateway/egressgateway_controller.go
```diff
@@ -0,0 +1,666 @@
...
+
+func setProgressing(cli client.Client, ctx context.Context, egw *operatorv1.EgressGateway, reason, msg string) {
+	updateEgwStatusConditions(cli, ctx, egw, operatorv1.ComponentProgressing, metav1.ConditionTrue, reason, msg, false)
+}
+
+func setAvailable(cli client.Client, ctx context.Context, egw *operatorv1.EgressGateway, reason, msg string) {
+	updateEgwStatusConditions(cli, ctx, egw, operatorv1.ComponentAvailable, metav1.ConditionTrue, reason, msg, true)
+}
+
+func updateEgwStatusConditions(cli client.Client, ctx context.Context, egw *operatorv1.EgressGateway, ctype operatorv1.StatusConditionType, status metav1.ConditionStatus, reason, msg string, updateStatus bool) {
```
caseydavenport, line 632: Egw -> "EGW"

Golang conventions use same-case acronyms.

pkg/controller/egressgateway/egressgateway_controller.go
```diff
@@ -0,0 +1,666 @@
...
+					fmt.Errorf("%s", getDegradedMsg(unreadyEGW)), reqLogger)
+				return reconcile.Result{}, nil
+			}
+			r.status.ClearDegraded()
+			return reconcile.Result{}, nil
+		}
+		// If the EGW resource is present, reconcile only that resource.
+		// Remove this from the list of EGWs before computing status.
+		egwsToReconcile = []operatorv1.EgressGateway{*requestedEGW}
+		egws = append(egws[:idx], egws[idx+1:]...)
```
caseydavenport, line 206: > // Remove this from the list of EGWs before computing status.

IIUC, I think this is only done because we are trying to handle unready EGW status reporting in this controller. However, once we remove that logic and use the status manager instead, we won't need this line.

pkg/controller/egressgateway/egressgateway_controller.go
```diff
@@ -0,0 +1,666 @@
...
+	installation *operatorv1.InstallationSpec, unreadyEGW *operatorv1.EgressGateway) (reconcile.Result, error) {
+
+	preDefaultPatchFrom := client.MergeFrom(egw.DeepCopy())
+	// update the EGW resource with default values.
+	fillDefaults(egw, installation)
+	// Validate the EGW resource.
+	err := validateEgressGateway(ctx, r.client, egw)
+	if err != nil {
+		reqLogger.Error(err, fmt.Sprintf("Error validating Egress Gateway Name = %s, Namespace = %s", egw.Name, egw.Namespace))
+		r.status.SetDegraded(operatorv1.ResourceValidationError,
```
caseydavenport, line 306: Can we simplify this setDegraded logic at all?

We have two functions in this file named `setDegraded`, and another one on the status manager, and it's not clear what each of them do since they don't have any comments, which also makes it hard to tell if they are being called in the correct places.

Maybe clarifying the names and adding comments would resolve this, but also wonder if any of them can be combined? 

I'll keep looking closer to see if I can come up with anything.

pkg/controller/egressgateway/egressgateway_controller.go
```diff
@@ -0,0 +1,666 @@
...
+	}
+	return nil
+}
+
+func setDegraded(cli client.Client, ctx context.Context, egw *operatorv1.EgressGateway, reason, msg string) {
+	updateEgwStatusConditions(cli, ctx, egw, operatorv1.ComponentDegraded, metav1.ConditionTrue, reason, msg, true)
+}
+
+func setProgressing(cli client.Client, ctx context.Context, egw *operatorv1.EgressGateway, reason, msg string) {
+	updateEgwStatusConditions(cli, ctx, egw, operatorv1.ComponentProgressing, metav1.ConditionTrue, reason, msg, false)
```
caseydavenport, line 625: Why is updateStatus set to false for progressing?

pkg/controller/utils/utils.go
caseydavenport: client.Get() should be pretty darn cheap - it uses the cache first before hitting the API server so I would expect most cases there's not even an API call

api/v1/egressgateway_types.go
sridhartigera: I read that  default value will be used  ```TerminationGracePeriodSeconds``` is ```nil```. Hence I thought I need not do a nil check.

pkg/render/egressgateway/egressgateway.go
```diff
@@ -0,0 +1,438 @@
...
+	path := c.config.Installation.ImagePath
+	prefix := c.config.Installation.ImagePrefix
+
+	if c.config.OSType != c.SupportedOSType() {
+		return fmt.Errorf("Egress Gateway is supported only on %s", c.SupportedOSType())
+	}
+
+	var err error
+	c.config.egwImage, err = components.GetReference(components.ComponentEgressGateway, reg, path, prefix, is)
+	if err != nil {
```
caseydavenport, line 84: A nit, but this is logically the same as `return err`

pkg/render/egressgateway/egressgateway.go
```diff
@@ -0,0 +1,438 @@
...
+
+// Config contains all the config information EgressGateway needs to render component.
+type Config struct {
+	PullSecrets  []*corev1.Secret
+	Installation *operatorv1.InstallationSpec
+	OSType       rmeta.OSType
+	EgressGW     *operatorv1.EgressGateway
+
+	egwImage          string
+	EgressGWVxlanVNI  int
```
caseydavenport, line 65: nit: VXLAN instead of Vxlan

caseydavenport: Or even just "VXLANVNI" and "VXLANPort" since we don't need to re-write "EgressGW" - the context is already clear.

pkg/render/egressgateway/egressgateway.go
```diff
@@ -0,0 +1,438 @@
...
+	return true
+}
+
+func (c *component) egwDeployment() *appsv1.Deployment {
+	d := appsv1.Deployment{
+		TypeMeta: metav1.TypeMeta{Kind: "Deployment", APIVersion: "apps/v1"},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      c.config.EgressGW.Name,
+			Namespace: c.config.EgressGW.Namespace,
+			Labels:    c.config.EgressGW.Spec.Template.Metadata.Labels,
```
caseydavenport, line 118: Labels are handled by the override mechanism as well, so shouldn't be done here IIUC.

pkg/render/egressgateway/egressgateway.go
```diff
@@ -0,0 +1,438 @@
...
+	psp.Spec.Volumes = []policyv1beta1.FSType{
+		policyv1beta1.CSI,
+		policyv1beta1.Projected,
+	}
+	return psp
+}
+
+func (c *component) egwClusterRole() *rbacv1.ClusterRole {
+	namespacedName := fmt.Sprintf("%s-%s", c.config.EgressGW.Namespace, c.config.EgressGW.Name)
+	rules := []rbacv1.PolicyRule{
```
caseydavenport, line 296: Should only include these rules if `usePSP` is enabled

pkg/render/egressgateway/egressgateway.go
```diff
@@ -0,0 +1,438 @@
...
+		SELinuxContext:           ocsv1.SELinuxContextStrategyOptions{Type: ocsv1.SELinuxStrategyMustRunAs},
+		SupplementalGroups:       ocsv1.SupplementalGroupsStrategyOptions{Type: ocsv1.SupplementalGroupsStrategyRunAsAny},
+		Users: []string{
+			fmt.Sprintf("system:serviceaccount:%s:%s", c.config.EgressGW.Namespace, c.config.EgressGW.Name),
+		},
+		Volumes: []ocsv1.FSType{"*"},
+	}
+}
+
+func concatString(arr []string) string {
```
caseydavenport, line 425: I think this function is essentially reinventing `json.Marshal` for a slice, no?

api/v1/egressgateway_types.go
caseydavenport: It's not a matter of default values, it's to avoid a nil pointer dereference in case any of those intermediate structures is nil.

pkg/render/egressgateway/egressgateway.go
sridhartigera: ```egwClusterRole``` is called only when ```usePSP``` is Enabled.

caseydavenport: Then we need to also delete them when usePSP is Disabled, otherwise these configs will stay around when a user upgrades their cluster to a version of k8s that doesn't have PSP support.

pkg/controller/egressgateway/egressgateway_controller.go
```diff
@@ -0,0 +1,666 @@
...
+	// To start with all EGW resources must be reconciled.
+	egwsToReconcile := egws
+	if request.Namespace != "" {
+		requestedEGW, idx := getRequestedEgressGateway(egws, request)
+		if requestedEGW == nil {
+			// Request object not found, could have been deleted after reconcile request.
+			// Owned objects are automatically garbage collected. For additional cleanup logic use finalizers.
+			reqLogger.Info("EgressGateway object not found")
+			// Since the EGW resource is not found, remove the deployment.
+			r.status.RemoveDeployments(types.NamespacedName{Name: request.Name, Namespace: request.Namespace})
```
caseydavenport, line 187: I think this type of controller might warrant the addition of a `SetDeployements` function that allows us to add / remove all at once, rather than try to add / remove the right deployments in the controller. I think it will simplify the logic and make it less likely we accidentally forget to add or remove a deployment in a certain branch. 

We can include this in a separate PR, though.

pkg/controller/egressgateway/egressgateway_controller.go
```diff
@@ -0,0 +1,666 @@
...
+	// degraded. Now lets assume Blue gets updated and gets rendered properly. In this case, TigeraStatus should still be
+	// degraded as Red is unhealthy. "status" represents, the cumulative status of EGWs.
+	if unreadyEGW != nil {
+		r.status.SetDegraded(operatorv1.ResourceCreateError,
+			fmt.Sprintf("Error reconciling Egress Gateway resource. Name=%s Namespace=%s", unreadyEGW.Name, unreadyEGW.Namespace),
+			fmt.Errorf("%s", getDegradedMsg(unreadyEGW)), reqLogger)
+		return reconcile.Result{}, nil
+	}
+
+	r.status.ClearDegraded()
```
caseydavenport, line 382: I think that the handling of unreadyEGW and clearDegraded should be moved outside of this function.

This function is meant for handling an individual instance of an EGW CR, but the unreadyEGW and ClearDegraded logic are global across _all_ EGWs, so shouldn't be done as part of this loop.

api/v1/egressgateway_types.go
sridhartigera: EGW pod has some logs which are trace and some which are debug. Can I just leave ``Trace``` as it is.

api/v1/egressgateway_types.go
sridhartigera: It is valid to have both of them nil.

api/v1/egressgateway_types.go
caseydavenport: Gotcha

api/v1/egressgateway_types.go
caseydavenport: What does it mean when both of them are nil? What is the reason a user would do that?

sridhartigera: ICMP probes and HTTP probes are used to verify if the EGW can communicate with an external server. This is used to do a health check if EGW pods can reach outside the cluster. I think users can choose not to configure any of the probes and still have EGWs working fine. @fasaxc

caseydavenport: Yes, exactly. If users want to disable failure detection altogether, then they must leave `EgressGatewayFailureDetection` as nil the main spec.

If the user provides an `EgressGatewayFailureDetection`, but doesn't configure any failure detection modes (ICMP or HTTP) then that is an invalid configuration that doesn't make sense and we cannot action, and so we need to reject it.

For example, the following yaml is invalid and should be rejected: 

```yaml
spec:
  egressGatewayFailureDetection: 
    healthTimeoutDataStoreSeconds: 30
    icmpProbe: nil
    httpProbe: nil
```

caseydavenport: This is how disabling failure detection should be done:

```yaml
spec:
  egressGatewayFailureDetection: nil
```

pkg/controller/utils/utils.go
sridhartigera: I have a used a separate function that just returns the status.

pkg/controller/egressgateway/egressgateway_controller.go
sridhartigera: Since all the status management is done here, I need the index of the requested resource. Hence I did not use the client.

pkg/controller/egressgateway/egressgateway_controller.go
sridhartigera: For now, I have just logged a message.

pkg/controller/egressgateway/egressgateway_controller.go
sridhartigera: Updating Name and Namespace of the EGW resource in its own condition is not needed. Hence removed r.setDegraded. Added comments as well.

pkg/render/egressgateway/egressgateway.go
caseydavenport: I think you missed this comment - Labels and other metadata are handled here: https://github.com/tigera/operator/blob/ddbaa22cd1b351564e952b851bc1b6b5d0aeae0d/pkg/render/common/components/components.go#L41-L50

Is there a reason that isn't sufficient for EGW?

pkg/render/egressgateway/egressgateway.go
```diff
@@ -0,0 +1,470 @@
...
+}
+
+func (c *component) deploymentPodTemplate() *corev1.PodTemplateSpec {
+	var ps []corev1.LocalObjectReference
+	for _, x := range c.config.PullSecrets {
+		ps = append(ps, corev1.LocalObjectReference{Name: x.Name})
+	}
+	return &corev1.PodTemplateSpec{
+		ObjectMeta: metav1.ObjectMeta{
+			Annotations: c.egwBuildAnnotations(),
```
caseydavenport, line 144: Do we not need to provide labels for the pods?

pkg/render/egressgateway/egressgateway.go
```diff
@@ -0,0 +1,470 @@
...
+			fmt.Sprintf("system:serviceaccount:%s:%s", c.config.EgressGW.Namespace, c.config.EgressGW.Name),
+		},
+		Volumes: []ocsv1.FSType{"*"},
+	}
+}
+
+func concatString(arr []string) string {
+	str, err := json.Marshal(arr)
+	if err != nil {
+		return ""
```
caseydavenport, line 467: we need to at a minimum log an error if we hit this branch.

pkg/controller/egressgateway/egressgateway_controller.go
caseydavenport: @sridhartigera nit, but calico -> "Calico" since it's a proper noun.

Also, it would be good to include the version of Calico that we are waiting for in the log message.

pkg/render/egressgateway/egressgateway.go
```diff
@@ -0,0 +1,470 @@
...
+	rbacv1 "k8s.io/api/rbac/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	egwPortName               = "health"
+	DefaultEGWVxlanPort int   = 4790
```
caseydavenport, line 42: another nit, but the casing is wrong on this for golang as well

should be `DefaultEGWVXLANPort` and `DefaultEGWVXLANVNI`. Although again, could just be `DefaultVXLANPort` and `DefaultVXLANVNI` for simplicity (since we're already within the egressgateway package)

pkg/render/egressgateway/egressgateway.go
sridhartigera: Should be handled by ```GetPodTemplateMetadata```

pkg/render/egressgateway/egressgateway.go
sridhartigera: For now, ```GetMetadata``` returns nil for EGW. To be fixed soon.

caseydavenport (approved):
LGTM - don't forget to squash your commits please

PR #2352, opened 2022-12-20
Create node trustedbundle with root certificates.

sridhartigera:
## Description

When using EGWs in AWS, felix uses Ec2 APIs to read/set instances, IPs. These API calls started failing as these need to be signed with root certificate. 

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


rene-dekker:
The main question I have is do we know under what conditions we need the root cert bundle and under what conditions we don't? 
 
If we do, we could make this PR perhaps more specific.

sridhartigera:
> The main question I have is do we know under what conditions we need the root cert bundle and under what conditions we don't?
> 
> If we do, we could make this PR perhaps more specific.

Felix talks outside the cluster when we have EGW enabled on AWS, EKS. To be more specific when ```AWSSecondaryIPSupport``` is enabled, felix talks outside using Ec2 APIs.

rene-dekker (approved):
lgtm

PR #2353, opened 2022-12-20
Fix that operator could not remove a ClusterIP.

fasaxc:
## Description

A recent fix removed ClusterIPs from our metrics services to make them headless.  That change cannot be applied via update, the service must be recreated.

- Add special case to `createOrUpdateObject` to handle services, modelled on similar logic for secrets/jobs. 
- Change `mergeState` to operate on a copy, otherwise it updates the resource version on the desired object, which makes `Create()` fail.

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/utils/component.go
+++ b/pkg/controller/utils/component.go
@@ -137,7 +137,9 @@ func (c componentHandler) createOrUpdateObject(ctx context.Context, obj client.O
 				return err
 			}
 
-			if err := c.client.Create(ctx, obj); err != nil {
+			// Do the Create() with the merged object so that we preserve external labels/annotations.
+			resetMetadataForCreate(mobj)
+			if err := c.client.Create(ctx, mobj); err != nil {
 				return err
 			}
 			return nil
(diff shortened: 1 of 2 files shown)


pkg/controller/utils/component.go
```diff
@@ -155,17 +156,29 @@ func (c componentHandler) createOrUpdateObject(ctx context.Context, obj client.O
 				if err := c.client.Create(ctx, obj); err != nil {
```
tmjd, line 156: I couldn't add a comment on the line above but the Secret case called obj.SetResourceVersion("") so the Create would work.

tmjd (commented):
This points out that previously we would keep any labels or annotations on Secrets even if they changed types. The change to operate on a copy in mergeState I think would result in the secret losing the labels and annotations.
Perhaps we should be using mobj even when we Create then Delete and we can remove the resource version with a SetResourceVersion("").

tmjd (approved):
LGTM

PR #2369, opened 2023-01-05
Add CalicoVersion to the installation status

sridhartigera:
## Description
Added a new field CalicoVersion to the installation status and updated GetInstallation to return both installation spec and status. This can be used in upgrades, when we want a component to wait for calico-node to be upgraded, before proceeding.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/api/v1/installation_types.go
+++ b/api/v1/installation_types.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2022 Tigera, Inc. All rights reserved.
+// Copyright (c) 2023 Tigera, Inc. All rights reserved.
 /*
 
 Licensed under the Apache License, Version 2.0 (the "License");
@@ -651,6 +651,9 @@ type InstallationStatus struct {
 	// Computed is the final installation including overlaid resources.
 	// +optional
 	Computed *InstallationSpec `json:"computed,omitempty"`
+
+	// CalicoVersion shows the current running version of calico.
+	CalicoVersion string `json:"calicoVersion,omitempty"`
 
 	// Conditions represents the latest observed set of conditions for the component. A component may be one or more of
 	// Ready, Progressing, Degraded or other customer types.
--- a/pkg/controller/installation/core_controller.go
+++ b/pkg/controller/installation/core_controller.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2019-2022 Tigera, Inc. All rights reserved.
+// Copyright (c) 2019-2023 Tigera, Inc. All rights reserved.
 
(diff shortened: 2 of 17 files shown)


pkg/controller/installation/core_controller.go
```diff
@@ -1446,6 +1448,7 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 	} else {
 		instance.Status.ImageSet = imageSet.Name
 	}
+	instance.Status.CalicoVersion = calicoVersion
```
rene-dekker, line 1451: I think we should have some unit test with an "existing installation + status" that then reconciles and verifies that the version has been updated so simulate an upgrade behaviour.

api/v1/installation_types.go
```diff
@@ -652,6 +652,9 @@ type InstallationStatus struct {
 	// +optional
 	Computed *InstallationSpec `json:"computed,omitempty"`
 
+	// CalicoVersion shows the current running version of calico.
+	CalicoVersion string `json:"calicoVersion,omitempty"`
```
rene-dekker, line 656: This field was added so at the end of the core_controllers' reconcile loop the version is set. This way, you know if calico-system has rolled out successfully and that you can go ahead with deploying other things, in this case EGW. This is better than simply using the Ready state, since the EGW controller might win the "race" when an operator pod is just starting after an upgrade and that Ready state might be tied to the old version being rolled out.

For this reason it is also important that the contents of this field reflect the OSS version in case of Variant OSS and Enterprise in case of Calico Enterprise. With the current implementation if we take branch 1.27 as an example the version will be either:
- v3.14.3 for EE
- v3.23.5 for OSS
@tmjd I was wondering if this is confusing since there are v3.x.x version overlaps between OSS and EE. I was wondering if we should prefix or postfix it with an indicator or if this is ok, since the variant is also part of the status. What is your opinion on this?

I would say that if we keep the implementation the way it is right now, we should at least improve the comment of this field.

tmjd: I don't think it is confusing enough to duplicate the Variant info by including a prefix or postfix on the CalicoVersion. I do think it would be good to make it clear in the comments that both the Variant and CalicoVersion are needed to know the exact version deployed.

pkg/controller/installation/core_controller.go
tmjd, line 1451: I'd like to see this line right after the `instance.Status.Variant` assignment and a comment that says the two fields should be updated at the same time so there isn't a chance of an inconsistent status state.

PR #2377, opened 2023-01-05
Fix race condition where trusted bundle is rendered twice with differ…

rene-dekker:
The trusted bundle is rendered twice in the elasticsearch namespace. Since all the required certificates for esmetrics should already be rendered before the esgw is rendered, we can simply refrain from re-rendering it.

### DIFF
--- a/pkg/controller/logstorage/esgateway.go
+++ b/pkg/controller/logstorage/esgateway.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2022 Tigera, Inc. All rights reserved.
+// Copyright (c) 2023 Tigera, Inc. All rights reserved.
 
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -21,6 +21,7 @@ import (
 	"github.com/tigera/operator/pkg/controller/certificatemanager"
 	"github.com/tigera/operator/pkg/dns"
 	rcertificatemanagement "github.com/tigera/operator/pkg/render/certificatemanagement"
+	"github.com/tigera/operator/pkg/render/monitor"
 	"github.com/tigera/operator/pkg/tls/certificatemanagement"
 	corev1 "k8s.io/api/core/v1"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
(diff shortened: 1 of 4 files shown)


pkg/controller/logstorage/esgateway.go
```diff
@@ -43,41 +44,51 @@ func (r *ReconcileLogStorage) createEsGateway(
...
+		return reconcile.Result{}, nil, false, err
 	} else if esInternalCertificate == nil {
 		reqLogger.Info("Waiting for internal Elasticsearch tls certificate secret to be available")
 		r.status.SetDegraded(operatorv1.ResourceNotReady, "Waiting for internal Elasticsearch tls certificate secret to be available", nil, reqLogger)
-		return reconcile.Result{}, false, nil
+		return reconcile.Result{}, nil, false, nil
 	}
-	trustedBundle := certificateManager.CreateTrustedBundle(esInternalCertificate, kibanaCertificate)
+
+	prometheusCertificate, err := certificateManager.GetCertificate(r.client, monitor.PrometheusClientTLSSecretName, common.OperatorNamespace())
```
Brian-McM, line 77: I don't think the prometheus cert is for es gateway, so I think this is a little misleading adding it here. I understand though why you've done it as other solutions would take longer.

I think since we're going towards having a single tigera ca config map, we'd need to develop a pattern where components can add what they need to the CA but don't actually render the secret for it (in case some other component needs to add more to the CA). This would probably involve a refactor of how we render all the elasticsearch components since they're essentially separate controllers.

Brian-McM: If we're leaving this as is (which I think we are given the time constraints) could you please add a note here saying that this certificate is not for es-metrics and why it needs to be rendered here. Maybe also add a note in es metrics in case somebody is looking in es-metrics for the prometheus cert addition.

Brian-McM (approved):

pkg/controller/logstorage/esgateway.go
```diff
@@ -129,7 +131,7 @@ func (r *ReconcileLogStorage) createEsGateway(
 		TrustedBundle: trustedBundle,
 	})
 
-	for _, comp := range []render.Component{esGatewayComponent, certificateComponent} {
+	for _, comp := range []render.Component{certificateComponent, esGatewayComponent} {
```
rene-dekker, line 134: Changed the order so the secrets and cm are rendered before the deployments etc. The namespace is already rendered by logstorage.go

marvin-tigera:

marvin-tigera:

PR #2382, opened 2023-01-06
Run job until it eventually completes

rene-dekker:
If for some reason ES it down for a while, the job may not succeed within 6 tries (which used to be the old backoffLimit).


marvin-tigera:

marvin-tigera:

PR #2387, opened 2023-01-11
Enable mtls for kube controller metrics endpoint

vara2504:
## Description

https://tigera.atlassian.net/browse/EV-2287

Added mtls change to secure kube controller metrics endpoints.

Operator changes -
PR - https://github.com/tigera/operator/pull/2387

When variant = TigeraSecureEnterprise ,
Created new tls keypair for kubecontrollers. Mounted the Prometheus keypair to the kubecontroller trustedbundle.
Added the kube controllers certificate to the bundle in monitor_controller (prometheus)
passed the cert and key to kubecontroller pods via env variables.
Made changes in prometheus pods to watch for kubecontroller metrics. Added new service monitors.
updated network policy egress in allow-tigera.prometheus to allow connecting with kube controllers metric endpoint
Kubecontroller changes:
PR - https://github.com/tigera/calico-private/pull/5576

prometheus server in kube controllers will accept traffic over https when cert and key values are passed and only accept certs from the trusted bundle and the common name in the prometheus cert.
else if cert and key not available it accept traffic over http.
Tested scenarios:
sudo -E ~/bin/kubefwd svc -n calico-system

Default port :9094
Kubecontroller port can be configure to diff port via
kubectl patch kubecontrollersconfiguration default --type=merge --patch '{"spec":{"prometheusMetricsPort": 9095}}'

setup:
https://docs.tigera.io/v3.14/maintenance/monitor/prometheus/byo-prometheus#verify-byo-prometheus
## For PR author

- [ ] Tests for change.

## For PR reviewers

Must be merged with kube controller PR
Kubecontroller changes:
PR - https://github.com/tigera/calico-private/pull/5576

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/installation/core_controller.go
+++ b/pkg/controller/installation/core_controller.go
@@ -1159,18 +1159,34 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 		calicoVersion = components.EnterpriseRelease
 	}
 
-	// Query the KubeControllersConfiguration object. We'll use this to help configure kube-controllers.
-	kubeControllersConfig := &crdv1.KubeControllersConfiguration{}
-	err = r.client.Get(ctx, types.NamespacedName{Name: "default"}, kubeControllersConfig)
-	if err != nil && !apierrors.IsNotFound(err) {
+	kubeControllersMetricsPort, err := utils.GetKubeControllerMetricsPort(ctx, r.client)
+	if err != nil {
 		r.status.SetDegraded(operator.ResourceReadError, "Unable to read KubeControllersConfiguration", err, reqLogger)
 		return reconcile.Result{}, err
 	}
 
-	// Determine the port to use for kube-controllers metrics.
-	kubeControllersMetricsPort := 0
-	if kubeControllersConfig.Spec.PrometheusMetricsPort != nil {
-		kubeControllersMetricsPort = *kubeControllersConfig.Spec.PrometheusMetricsPort
+	// Secure calico kube controller metrics.
+	var kubeControllerTLS certificatemanagement.KeyPairInterface
+	if instance.Spec.Variant == operator.TigeraSecureEnterprise {
+		// Create or Get TLS certificates for kube controller.
+		kubeControllerTLS, err = certificateManager.GetOrCreateKeyPair(
+			r.client,
+			kubecontrollers.KubeControllerPrometheusTLSSecret,
+			common.OperatorNamespace(),
+			dns.GetServiceDNSNames(kubecontrollers.KubeControllerMetrics, common.CalicoNamespace, r.clusterDomain))
+		if err != nil {
+			r.status.SetDegraded(operator.ResourceReadError, "Error finding or creating TLS certificate kube controllers metric", err, reqLogger)
+			return reconcile.Result{}, err
+		}
+
+		// Add prometheus client certificate to Trusted bundle.
+		kubecontrollerprometheusTLS, err := certificateManager.GetCertificate(r.client, monitor.PrometheusClientTLSSecretName, common.OperatorNamespace())
+		if err != nil {
+			r.status.SetDegraded(operator.ResourceReadError, "Failed to get certificate for kube controllers", err, reqLogger)
+			return reconcile.Result{}, err
+		} else if kubecontrollerprometheusTLS != nil {
+			typhaNodeTLS.TrustedBundle.AddCertificates(kubeControllerTLS, kubecontrollerprometheusTLS)
+		}
 	}
 
 	nodeAppArmorProfile := ""
@@ -1226,7 +1242,7 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 	components = append(components,
 		rcertificatemanagement.CertificateManagement(&rcertificatemanagement.Config{
 			Namespace:       common.CalicoNamespace,
-			ServiceAccounts: []string{render.CalicoNodeObjectName, render.TyphaServiceAccountName},
+			ServiceAccounts: []string{render.CalicoNodeObjectName, render.TyphaServiceAccountName, kubecontrollers.KubeControllerServiceAccount},
 			KeyPairOptions: []rcertificatemanagement.KeyPairOption{
 				// this controller is responsible for rendering the tigera-ca-private secret.
 				rcertificatemanagement.NewKeyPairOption(certificateManager.KeyPair(), true, false),
@@ -1234,6 +1250,7 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 				rcertificatemanagement.NewKeyPairOption(nodePrometheusTLS, true, true),
 				rcertificatemanagement.NewKeyPairOption(managerInternalTLSSecret, true, true),
 				rcertificatemanagement.NewKeyPairOption(typhaNodeTLS.TyphaSecret, true, true),
+				rcertificatemanagement.NewKeyPairOption(kubeControllerTLS, true, true),
 			},
 			TrustedBundle: typhaNodeTLS.TrustedBundle,
 		}))
@@ -1320,6 +1337,8 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 		ManagerInternalSecret:       managerInternalTLSSecret,
 		Terminating:                 terminating,
 		UsePSP:                      r.usePSP,
+		MetricsServerTLS:            kubeControllerTLS,
+		TrustedBundle:               typhaNodeTLS.TrustedBundle,
 	}
 	components = append(components, kubecontrollers.NewCalicoKubeControllers(&kubeControllersCfg))
 
--- a/pkg/controller/monitor/monitor_controller.go
+++ b/pkg/controller/monitor/monitor_controller.go
@@ -49,6 +49,7 @@ import (
 	rcertificatemanagement "github.com/tigera/operator/pkg/render/certificatemanagement"
 	"github.com/tigera/operator/pkg/render/common/networkpolicy"
 	rsecret "github.com/tigera/operator/pkg/render/common/secret"
+	"github.com/tigera/operator/pkg/render/kubecontrollers"
 	"github.com/tigera/operator/pkg/render/logstorage/esmetrics"
 	"github.com/tigera/operator/pkg/render/monitor"
 	"github.com/tigera/operator/pkg/tls/certificatemanagement"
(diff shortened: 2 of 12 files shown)


pkg/controller/installation/core_controller.go
```diff
@@ -1309,9 +1336,22 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 		ManagerInternalSecret:       managerInternalTLSSecret,
 		Terminating:                 terminating,
 		UsePSP:                      r.usePSP,
+		MetricsServerTLS:            kubeControllerTLS,
+		TrustedBundle:               kubeControllerTrustedBundle,
 	}
 	components = append(components, kubecontrollers.NewCalicoKubeControllers(&kubeControllersCfg))
 
+	components = append(components,
```
rene-dekker, line 1344: The trusted bundles are a per-namespace thing as it currently is implemented. With this addition two configmaps are competing for one spot. This block has to be removed. The certificates added in this PR, should be in the `typhaNodeTLS.TrustedBundle`

pkg/controller/installation/core_controller.go
```diff
@@ -1162,6 +1162,33 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
...
+			return reconcile.Result{}, err
+		}
+
+		// Add prometheus client certificate to Trusted bundle.
+		kubecontrollerprometheusTLS, err := certificateManager.GetCertificate(r.client, monitor.PrometheusClientTLSSecretName, common.OperatorNamespace())
+		if err != nil {
+			r.status.SetDegraded(operator.ResourceReadError, "Failed to get certificate for kube controllers", err, reqLogger)
+			return reconcile.Result{}, err
+		} else if kubecontrollerprometheusTLS == nil {
+			r.status.SetDegraded(operator.ResourceNotFound, "Prometheus secrets are not available yet, waiting until they become available", err, reqLogger)
```
rene-dekker, line 1186: nit: err is nil here, perhaps better to explicitly put nil in this line.

pkg/controller/installation/core_controller.go
```diff
@@ -1162,6 +1162,33 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 		kubeControllersMetricsPort = *kubeControllersConfig.Spec.PrometheusMetricsPort
 	}
 
+	// Secure calico kube controller metrics.
+	var kubeControllerTLS certificatemanagement.KeyPairInterface
+	kubeControllerTrustedBundle := certificateManager.CreateTrustedBundle()
```
rene-dekker, line 1167: Reuse typhaNodeTLS.TrustedBundle. See my comment below.

pkg/render/kubecontrollers/kube-controllers.go
```diff
@@ -63,6 +64,7 @@ const (
 	ElasticsearchKubeControllersUserName               = "tigera-ee-kube-controllers"
 	ElasticsearchKubeControllersSecureUserSecret       = "tigera-ee-kube-controllers-elasticsearch-access-gateway"
 	ElasticsearchKubeControllersVerificationUserSecret = "tigera-ee-kube-controllers-gateway-verification-credentials"
+	KubeControllerPrometheusTLSSecret                  = "tigera-kube-controller-prometheus-tls"
```
rene-dekker, line 67: I'd call it `calico-kube-controllers-metrics-tls`, I suspect sooner or later this might be supported for both Variants. Either way, the calico-kube-controllers deployment will also support it for Enterprise as well.

pkg/controller/monitor/monitor_controller.go
```diff
@@ -332,6 +336,20 @@ func (r *ReconcileMonitor) Reconcile(ctx context.Context, request reconcile.Requ
 		return reconcile.Result{}, err
 	}
 
+	// Query the KubeControllersConfiguration object. We'll use this to help configure kube-controllers metric port.
+	kubeControllersConfig := &crdv1.KubeControllersConfiguration{}
+	err = r.client.Get(ctx, types.NamespacedName{Name: "default"}, kubeControllersConfig)
+	if err != nil && !apierrors.IsNotFound(err) {
```
rene-dekker, line 342: Could you extract this method into the utils package, so both core_controller and monitor_controller can use it?

pkg/render/kubecontrollers/kube-controllers.go
```diff
@@ -614,6 +633,9 @@ func (c *kubeControllersComponent) kubeControllersVolumes() []corev1.Volume {
 	if c.cfg.TrustedBundle != nil {
 		volumes = append(volumes, c.cfg.TrustedBundle.Volume())
 	}
+	if c.cfg.MetricsServerTLS != nil {
+		volumes = append(volumes, c.cfg.MetricsServerTLS.Volume())
```
rene-dekker, line 637: It needs to also be added to the `annotations()` func.

pkg/render/kubecontrollers/kube-controllers.go
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2019,2022 Tigera, Inc. All rights reserved.
+// Copyright (c) 2019,2023 Tigera, Inc. All rights reserved.
```
rene-dekker, line 1: Do we need a network policy ingress rule for prometheus in here?

pkg/render/monitor/monitor.go
```diff
@@ -992,3 +1004,28 @@ func allowTigeraPrometheusOperatorPolicy(cfg *Config) *v3.NetworkPolicy {
...
+			Name:      KubeControllerMetrics,
+			Namespace: common.TigeraPrometheusNamespace,
+			Labels:    map[string]string{"team": "network-operators"},
+		},
+		Spec: monitoringv1.ServiceMonitorSpec{
+			Selector:          metav1.LabelSelector{MatchLabels: map[string]string{"k8s-app": "calico-kube-controllers"}},
+			NamespaceSelector: monitoringv1.NamespaceSelector{MatchNames: []string{"calico-system"}},
+			Endpoints: []monitoringv1.Endpoint{
+				{
+					HonorLabels:   true,
```
rene-dekker, line 1021: What does this do exactly? I don't know that I understand it from the description in prometheus:
```
// HonorLabels chooses the metric's labels on collisions with target labels.
```

pkg/controller/installation/core_controller.go
vara2504: Done

pkg/controller/installation/core_controller.go
vara2504: Done

pkg/controller/installation/core_controller.go
vara2504: Done

pkg/controller/monitor/monitor_controller.go
vara2504: Done

pkg/render/kubecontrollers/kube-controllers.go
vara2504: Done

pkg/render/kubecontrollers/kube-controllers.go
vara2504: Done

pkg/render/monitor/monitor.go
vara2504: ServiceMonitorSpec has TargetLabels 
```// TargetLabels transfers labels from the Kubernetes `Service` onto the created metrics.
// All labels set in `selector.matchLabels` are automatically transferred.```
 As discussed, followed the same behaviour  in other service monitors and used HonorLabels as true.

File: pkg/render/kubecontrollers/kube-controllers.go (hunk 7 above)
Comment (vara2504) [reply pkg/render/kubecontrollers/kube-controllers.go:1]: Kube-controllers does not have default deny hence prometheus pod is able to fetch kubecontroller metrics once we added  egress rule in prometheus ns alone.  Please suggest, do  we need to add explicit ingress rule in kube-controllers?

Comment (rene-dekker) [reply pkg/render/kubecontrollers/kube-controllers.go:1]: @pasanw @caseydavenport would you know more about that?

File: pkg/controller/installation/core_controller.go [hunk 9]
```diff
@@ -1148,18 +1148,36 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
...
+			return reconcile.Result{}, err
+		}
+
+		// Add prometheus client certificate to Trusted bundle.
+		kubecontrollerprometheusTLS, err := certificateManager.GetCertificate(r.client, monitor.PrometheusClientTLSSecretName, common.OperatorNamespace())
+		if err != nil {
+			r.status.SetDegraded(operator.ResourceReadError, "Failed to get certificate for kube controllers", err, reqLogger)
+			return reconcile.Result{}, err
+		} else if kubecontrollerprometheusTLS == nil {
+			r.status.SetDegraded(operator.ResourceNotFound, "Prometheus secrets are not available yet, waiting until they become available", nil, reqLogger)
```
rene-dekker, line 1177: I think if the secret does not exist, we can simply skip adding it to the bundle and continue with the rest of the code. I don't think we want to risk breaking the core controller's progress (which creates calico networking) .

pkg/controller/monitor/monitor_controller.go
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2021-2022 Tigera, Inc. All rights reserved.
+// Copyright (c) 2023 Tigera, Inc. All rights reserved.
```
rene-dekker, line 1: nit: should this be 2021-2023? I'm not so sure what we do with these.

vara2504: Done

pkg/controller/installation/core_controller.go
vara2504: Done

pkg/render/kubecontrollers/kube-controllers.go
pasanw: Vara's analysis is correct in that, given a default configuration, there is no policy selecting ingress traffic to kube-controllers, so an explicit allow is not required.

However, it's likely best practice for us to ensure that any required ingress flows are explicitly allowed so that our product continues to function as expected regardless of the customers policy configuration. As I understand it, that's the main objective of allow-tigera.

If we were to explicitly allow the prometheus ingress, this would cause policy selection of ingress traffic to kube-controllers and therefore begin denying any ingress not explicitly allowed. This would need to be considered/accommodated

pasanw: ^ Is that fair assessment @caseydavenport?

pkg/controller/installation/core_controller.go
```diff
@@ -1226,14 +1242,15 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 	components = append(components,
 		rcertificatemanagement.CertificateManagement(&rcertificatemanagement.Config{
 			Namespace:       common.CalicoNamespace,
-			ServiceAccounts: []string{render.CalicoNodeObjectName, render.TyphaServiceAccountName},
+			ServiceAccounts: []string{render.CalicoNodeObjectName, render.TyphaServiceAccountName, kubecontrollers.KubeControllerMetrics},
```
rene-dekker, line 1245: These should be service accounts. Please remove that value and add these two: [kubecontrollers.KubeControllerServiceAccount, kubecontrollers.EsKubeControllerServiceAccount]

rene-dekker (approved):
lgtm

PR #2393, opened 2023-01-12
Separate calico-node and calico-cni-plugin service accounts

MichalFupso:
## Description
Split of calico-node to calico-node and calico-cni-plugin service accounts

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


pkg/render/node.go
```diff
@@ -283,6 +304,38 @@ func (c *nodeComponent) nodeRoleBinding() *rbacv1.ClusterRoleBinding {
 	return crb
 }
 
+// nodeRoleBinding creates a clusterrolebinding giving the node service account the required permissions to operate.
```
caseydavenport, line 307: The comment looks like a copy/paste leftover.

pkg/render/node.go
```diff
@@ -251,6 +255,23 @@ func (c *nodeComponent) nodeServiceAccount() *corev1.ServiceAccount {
 	}
 }
 
+// nodeServiceAccount creates the node's service account.
```
caseydavenport, line 258: The comment looks like a copy/paste leftover.

pkg/render/node.go
```diff
@@ -283,6 +304,38 @@ func (c *nodeComponent) nodeRoleBinding() *rbacv1.ClusterRoleBinding {
...
+func (c *nodeComponent) CNIPluginRoleBinding() *rbacv1.ClusterRoleBinding {
+	finalizer := []string{}
+	if !c.cfg.Terminating {
+		finalizer = []string{NodeFinalizer}
+	}
+	crb := &rbacv1.ClusterRoleBinding{
+		TypeMeta: metav1.TypeMeta{Kind: "ClusterRoleBinding", APIVersion: "rbac.authorization.k8s.io/v1"},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:       CalicoCNIPluginObjectName,
+			Labels:     map[string]string{},
```
caseydavenport, line 317: You can leave this as `nil` if there are no labels we want to put on it.

pkg/render/node.go
```diff
@@ -283,6 +304,38 @@ func (c *nodeComponent) nodeRoleBinding() *rbacv1.ClusterRoleBinding {
...
+		Subjects: []rbacv1.Subject{
+			{
+				Kind:      "ServiceAccount",
+				Name:      CalicoCNIPluginObjectName,
+				Namespace: common.CalicoNamespace,
+			},
+		},
+	}
+	if c.cfg.MigrateNamespaces {
+		migration.AddBindingForKubeSystemNode(crb)
```
caseydavenport, line 334: I don't think this is doing the right thing - this is binding this role to the `calico-node` serviceaccount in the kube-system namespace.

What we want to do is bind it to the `calico-cni-plugin` service account in the kube-system namespace. LMK if it's not clear why - the migration logic isn't necessarily obvious.

pkg/render/node.go
```diff
@@ -283,6 +304,38 @@ func (c *nodeComponent) nodeRoleBinding() *rbacv1.ClusterRoleBinding {
 	return crb
 }
 
+// nodeRoleBinding creates a clusterrolebinding giving the node service account the required permissions to operate.
+func (c *nodeComponent) CNIPluginRoleBinding() *rbacv1.ClusterRoleBinding {
```
caseydavenport, line 308: nit, but please call this `cniPluginClusterRoleBinding` - roles and clusterroles, and thus rolebindings and clusterrolebindings are different resources.

caseydavenport: I understand that some of the other functions in this file aren't correct either... e.g., `nodeRole` should be `nodeClusterRole`, but that's just another pre-existing naming bug :sweat:

pkg/render/node.go
```diff
@@ -344,7 +397,7 @@ func (c *nodeComponent) nodeRole() *rbacv1.ClusterRole {
 				// Used for creating service account tokens to be used by the CNI plugin.
 				APIGroups:     []string{""},
 				Resources:     []string{"serviceaccounts/token"},
-				ResourceNames: []string{"calico-node"},
+				ResourceNames: []string{"calico-node", "calico-cni-plugin"},
```
caseydavenport, line 400: We want to remove the ability for calico-node to create serviceaccount tokens for itself, so I think you should remove `calico-node` from this list

pkg/render/node.go
```diff
@@ -434,13 +487,6 @@ func (c *nodeComponent) nodeRole() *rbacv1.ClusterRole {
 				Resources: []string{"blockaffinities"},
 				Verbs:     []string{"watch"},
 			},
-			{
```
caseydavenport, line 437: huh, I guess we just had a duplicate entry?

pkg/render/node.go
```diff
@@ -490,6 +536,99 @@ func (c *nodeComponent) nodeRole() *rbacv1.ClusterRole {
 	return role
 }
 
+// nodeRole creates the clusterrole containing policy rules that allow the node daemonset to operate normally.
+func (c *nodeComponent) CNIPluginRole() *rbacv1.ClusterRole {
```
caseydavenport, line 540: Comment for this function also looks like copy/paste


But this should also be called `cniPluginClusterRole`

pkg/render/node.go
caseydavenport: Also, golang uses the casing of the function to determine if it should be public / private.

I don't think this function needs to be public, so it should be lowercase - `cniPluginClusterRoleBinding`

pkg/render/node.go
caseydavenport: This should also be a private function

pkg/render/node.go
```diff
@@ -490,6 +536,99 @@ func (c *nodeComponent) nodeRole() *rbacv1.ClusterRole {
...
+					"stagedkubernetesnetworkpolicies",
+					"stagednetworkpolicies",
+					"networksets",
+				},
+				Verbs: []string{"get", "list", "watch"},
+			},
+		},
+	}
+
+	if c.cfg.Installation.KubernetesProvider != operatorv1.ProviderOpenShift {
```
caseydavenport, line 620: I don't think we need this for the CNI plugin

pkg/render/node.go
```diff
@@ -490,6 +536,99 @@ func (c *nodeComponent) nodeRole() *rbacv1.ClusterRole {
...
+					"hostendpoints",
+					"ipamblocks",
+					"ippools",
+					"ipreservations",
+					"networkpolicies",
+					"stagedkubernetesnetworkpolicies",
+					"stagednetworkpolicies",
+					"networksets",
+				},
+				Verbs: []string{"get", "list", "watch"},
```
caseydavenport, line 615: The CNI plugin doesn't perform any watches, so we can remove that permission. 

It's probably moot though - I also don't think it needs most of these resources, so we can probably just delete this entire block. The only ones it should need are the IPAM resources which are already covered above.

pkg/render/node.go
```diff
@@ -490,6 +536,99 @@ func (c *nodeComponent) nodeRole() *rbacv1.ClusterRole {
...
+				Verbs: []string{"get", "list", "create", "update", "delete"},
+			},
+			{
+				// But, we only need to be able to query for IPAM config.
+				APIGroups: []string{"crd.projectcalico.org"},
+				Resources: []string{"ipamconfigs"},
+				Verbs:     []string{"get"},
+			},
+			{
+				// Calico creates some configuration on startup.
```
caseydavenport, line 585: I don't think the CNI plugin needs these

caseydavenport (changes requested):

pkg/render/node.go
```diff
@@ -253,6 +257,23 @@ func (c *nodeComponent) nodeServiceAccount() *corev1.ServiceAccount {
 	}
 }
 
+// cniPluginServiceAccount creates the node's service account.
```
caseydavenport, line 260: "the Calico CNI plugin's service account"

pkg/render/node.go
```diff
@@ -285,6 +306,36 @@ func (c *nodeComponent) nodeRoleBinding() *rbacv1.ClusterRoleBinding {
 	return crb
 }
 
+// cniPluginRoleBinding creates a rolebinding giving the node service account the required permissions to operate.
```
caseydavenport, line 309: Looks like a few of these comments have copy/paste errors - "the node service account"

pkg/render/node.go
caseydavenport: This comment still needs addressing @MichalFupso - need to remove "calico-node" from ResouceNames.

pkg/render/node.go
```diff
@@ -492,6 +536,51 @@ func (c *nodeComponent) nodeRole() *rbacv1.ClusterRole {
 	return role
 }
 
+// cniPluginRole creates the clusterrole containing policy rules that allow the node daemonset to operate normally.
```
caseydavenport, line 539: comment copy/paste

caseydavenport (changes requested):
A few minor changes.

I suspect there are also tests that will need to be written / modified.

pkg/render/node.go
```diff
@@ -492,6 +536,51 @@ func (c *nodeComponent) nodeRole() *rbacv1.ClusterRole {
...
 }
 
+// cniPluginRole creates the role containing policy rules that allow the Calico CNI plugin to operate normally.
+func (c *nodeComponent) cniPluginRole() *rbacv1.Role {
+	finalizer := []string{}
+	if !c.cfg.Terminating {
+		finalizer = []string{NodeFinalizer}
+	}
+	role := &rbacv1.Role{
+		TypeMeta: metav1.TypeMeta{Kind: "Role", APIVersion: "rbac.authorization.k8s.io/v1"},
```
caseydavenport, line 546: I think this needs to be a cluster role? The CNI plugin operates on resources across multiple namespaces.

The binding can still be a RoleBinding, though.

caseydavenport: In your other PR, this is a ClusterRole: https://github.com/projectcalico/calico/pull/7106/files#diff-a13e3dc70d98c53fce0f3a76fa93787186a8d68d182bc2ae231f45afd5d6b365R182

PR #2398, opened 2023-01-13
Add externalNetwork to the EGW operator CRD

sridhartigera:
## Description

We support EGW with multi VRF in 3.16. This PR has the changes to add externalNetworks to the EGW CRD.

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


api/v1/egressgateway_types.go
```diff
@@ -65,6 +65,12 @@ type EgressGatewaySpec struct {
 	// +required
 	IPPools []EgressGatewayIPPool `json:"ipPools"`
 
+	// ExternalNetworks defines the external network names this Egress Gateway is
+	// associated with.
+	// ExternalNetworks must match existing external networks.
+	// +optional
+	ExternalNetworks []string `json:"externalNetworks"`
```
song-jiang, line 72: Should be omitempty?

api/v1/egressgateway_types.go
```diff
@@ -65,6 +65,12 @@ type EgressGatewaySpec struct {
 	// +required
 	IPPools []EgressGatewayIPPool `json:"ipPools"`
 
+	// ExternalNetworks defines the external network names this Egress Gateway is
+	// associated with.
+	// ExternalNetworks must match existing external networks.
+	// +optional
+	ExternalNetworks []string `json:"externalNetworks,omitempty"`
```
caseydavenport, line 72: omitempty was a good catch - we should put that on all of the optional fields. Sorry for missing that in the original review.

pkg/controller/egressgateway/egressgateway_controller.go
```diff
@@ -526,6 +534,17 @@ func fillDefaults(egw *operatorv1.EgressGateway, installation *operatorv1.Instal
 	}
 }
 
+// validateExternalNetwork validates if the specified external network exists.
+func validateExternalNetwork(ctx context.Context, cli client.Client, externalNetwork string) error {
+	instance := &crdv1.ExternalNetwork{}
+	key := types.NamespacedName{Name: externalNetwork}
+	err := cli.Get(ctx, key, instance)
```
caseydavenport, line 541: @song-jiang Do we need to assert that the external network exists? Is there an ordering requirement here?

If it's possible to create them in either order, we might want to allow that - it would make installation slightly easier.

api/v1/egressgateway_types.go
sridhartigera: updated omitempty

pkg/controller/egressgateway/egressgateway_controller.go
song-jiang: @caseydavenport I think no customer should create EGW before external networks. If we don't enforce it, a customer may create EGW with misspelled network names ["valid-name", "wrong-name"], EGW will still be created for "valid-name", but for "worng-name" it is not doing the right thing. This could lead to security issues.

song-jiang (approved):
LGTM

Issue #2224, opened 2022-10-17
"csi-node-driver" daemonset is using default service account in calico version - v.3.24.x


malavika-nbs:
"csi-node-driver" daemonset is using default service account in calico version - v.3.24.x

## Expected Behavior

csi-node-driver daemonset should use a custom service account.

## Current Behavior

csi-node-driver daemonset created from tigera operator calico version - v3.24.x is using the default service account.

## Possible Solution

a. A custom service account should be created by the operator and used by the csi-node-driver 
                                                                                               (or)
b. Provide an option to pass the custom service account via values.yaml just like we have for image and registry variables currently.

## Context

Our gatekeeper policies are not letting the pods to get created as its using the default service account instead of having a custom one. Though we can amend our gatekeeper policies, It's not ideal to have a default service account being used.

caseydavenport:
Yeah, we should do this for best-practice purposes.

I should note that in this case:

- The default serviceaccount has no permissions bound to it, so anyone with that serviceaccount gets no API access.
- The CSI driver itself doesn't make any API calls, so once we split off to another serviceaccount that SA will also not get any permissions bound to it.

caseydavenport:
This is no longer the case

PR #2399, opened 2023-01-17
CSI - AutomountServiceAccountToken false when using default service account 

njgibbon:
### Background
Hello, 

First time that I've tried to contribute to an operator but it's one of the things I want to learn about this year. Happy to make any changes if I've missed anything conventional 'cus there could be all sorts i've overlooked. Thank you for your time :slightly_smiling_face: :peace_symbol:.

### Overview
Currently when not using OpenShift and using PSPs then for the CSI DaemonSet a new Service Account is created with RBAC which is used. When this is not the case then no Service Account is explicitly set for the DS pods. This means that the namespace default service account is set. In this circumstance the CIS K8s benchmark fails for "5.1.6 Ensure that Service Account Tokens are only mounted where necessary (Manual)" (at time of writing).

This little change should make it such that in the case the default SA is used for CSI then the token wouldn't be mounted and the benchmark check passes.

### Realsies
I understand that default service account tokens (given RBAC hasn't been added) can not do anything. And so there is realistically very little risk in mounting a token that couldn't access anything if used. However CIS is influential and this rule is / will be baked into all sorts of scanning and policy software. So, given there is little or no downside, this type of change is more to make life easier for teams to operate calico in environments that have all of these reviews, scans and policy things constantly. Mostly big ol' enterprises aha. So the value is there even if you aren't buying the security aspect of the change. 

### Future
This change would make the benchmark technically pass for CSI which it doesn't now. But also in the spirit of being pragmatic it would be nice if I / we could go on to always use a unique Service Account for CSI even in the case where it's not needed. And in this scenario we just don't create the RBAC elements.

The logic here being that a short hand policy implementation that is widely used in the industry to cover the benchmark is that the default token should never be used and pods which have no SA config fail these types of scan. Even though a default token that is not mounted is technically sufficient to cover the benchmark. Again, this just ends up reducing headaches for users in these types of environments.

## Tags
Improvement, Security, UX (Ops Experience).

## Issues
* https://github.com/tigera/operator/issues/2224

## For PR author
- [x] Tests for change.
- [x] If changing pkg/apis/, run `make gen-files` - n/a.
- [x] If changing versions, run `make gen-versions` - n/a.

## For PR reviewers
- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/csi.go
+++ b/pkg/render/csi.go
@@ -268,6 +268,10 @@ func (c *csiComponent) csiTemplate() corev1.PodTemplateSpec {
 		AutomountServiceAccountToken: ptr.BoolToPtr(false),
 	}
 
+	// SA is only used as part of associating PSP.
+	// If PSP is not on then SA is not created.
+	// Default SA will be used and token will not be mounted
+	// as it is not needed.
 	if !c.cfg.Openshift && c.cfg.UsePSP {
 		templateSpec.ServiceAccountName = CSIDaemonSetName
 		templateSpec.AutomountServiceAccountToken = ptr.BoolToPtr(true)
--- a/pkg/render/csi.go
+++ b/pkg/render/csi.go
@@ -261,14 +261,16 @@ func (c *csiComponent) csiTemplate() corev1.PodTemplateSpec {
 		Labels: templateLabels,
 	}
 	templateSpec := corev1.PodSpec{
-		Tolerations:      c.csiTolerations(),
-		Containers:       c.csiContainers(),
-		ImagePullSecrets: c.cfg.Installation.ImagePullSecrets,
-		Volumes:          c.csiVolumes(),
+		Tolerations:                  c.csiTolerations(),
+		Containers:                   c.csiContainers(),
+		ImagePullSecrets:             c.cfg.Installation.ImagePullSecrets,
+		Volumes:                      c.csiVolumes(),
+		AutomountServiceAccountToken: ptr.BoolToPtr(false),
 	}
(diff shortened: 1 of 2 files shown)


pkg/render/csi.go
```diff
@@ -261,14 +261,16 @@ func (c *csiComponent) csiTemplate() corev1.PodTemplateSpec {
...
+		Tolerations:                  c.csiTolerations(),
+		Containers:                   c.csiContainers(),
+		ImagePullSecrets:             c.cfg.Installation.ImagePullSecrets,
+		Volumes:                      c.csiVolumes(),
+		AutomountServiceAccountToken: ptr.BoolToPtr(false),
 	}
 
 	if !c.cfg.Openshift && c.cfg.UsePSP {
 		templateSpec.ServiceAccountName = CSIDaemonSetName
+		templateSpec.AutomountServiceAccountToken = ptr.BoolToPtr(true)
```
caseydavenport, line 273: This change LGTM, but I think it could do with a code comment explaining that right now the only permission bound to the serviceaccount is the capability to use podsecuritypolicies, so we only need / want to mount in the serviceaccount token if PSP is enabled on the cluster.

njgibbon:
Hiya @caseydavenport  - what's with the failed check? Is there something I need to do?

caseydavenport:
@njgibbon I'm really sorry, somehow this one slipped through the cracks on my end.

I believe the failed check is just a static file generation issue - rebasing on master and fixing the conflicts should fix the CI.

PR #2406, opened 2023-01-20
Fix EGW upgrade and crd comments for doc

sridhartigera:
## Description

When upgrading from 3.15 to 3.16, users are required to create an EGW CR similar to their current deployment. Once this is created, the operator overwrites the existing deployment with the new image. 

When creating the EGW deployment, the operator was adding additional labels on top of the specified ones. This results in a change in the selector of the existing deployment, which is immutable. As a result, upgrade fails. 

Fix - Not adding the default label if labels are specified. 

CRD changes - <name> was used in the comments of a field. This gets interpreted as html tags when building API reference docs. Changing this to [name]. 

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


rene-dekker (approved):
lgtm

PR #2427, opened 2023-01-25
Create, delete cluster scoped resources from EGW controller

sridhartigera:
## Description

When creating cluster scoped resources from EGW controller, controller-runtime throws an error that cluster scoped resources cannot have namespace scoped owners. Hence to avoid this

1. When PSP is enabled, create Role, Rolebindings instead of clusterrole, clusterrolebinding
2. When PSP is enabled, create the pod security policy from the controller. When the CR is deleted, delete the pod security policy.
3. When running in openShift, create or delete the securitycontext from the controller. 
## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/egressgateway/egressgateway_controller.go
+++ b/pkg/controller/egressgateway/egressgateway_controller.go
@@ -327,7 +327,7 @@ func (r *ReconcileEgressGateway) Reconcile(ctx context.Context, request reconcil
 	// Reconcile all the EGWs
 	var errMsgs []string
 	for _, egw := range egwsToReconcile {
-		err = r.reconcileEgressGateway(ctx, &egw, reqLogger, variant, fc, pullSecrets, installation, getEGWNamespacedNames(egws))
+		err = r.reconcileEgressGateway(ctx, &egw, reqLogger, variant, fc, pullSecrets, installation, getEGWNamespaceAndNames(egws))
 		if err != nil {
 			reqLogger.Error(err, "Error reconciling egress gateway")
 			errMsgs = append(errMsgs, err.Error())
@@ -354,7 +354,7 @@ func (r *ReconcileEgressGateway) Reconcile(ctx context.Context, request reconcil
 
 func (r *ReconcileEgressGateway) reconcileEgressGateway(ctx context.Context, egw *operatorv1.EgressGateway, reqLogger logr.Logger,
 	variant operatorv1.ProductVariant, fc *crdv1.FelixConfiguration, pullSecrets []*v1.Secret,
-	installation *operatorv1.InstallationSpec, namespacedNames []string) error {
+	installation *operatorv1.InstallationSpec, namespaceAndNames []string) error {
 
 	preDefaultPatchFrom := client.MergeFrom(egw.DeepCopy())
 	// update the EGW resource with default values.
@@ -391,15 +391,15 @@ func (r *ReconcileEgressGateway) reconcileEgressGateway(ctx context.Context, egw
 
 	openshift := r.provider == operatorv1.ProviderOpenShift
 	config := &egressgateway.Config{
-		PullSecrets:     pullSecrets,
-		Installation:    installation,
-		OSType:          rmeta.OSTypeLinux,
-		EgressGW:        egw,
-		VXLANPort:       egwVXLANPort,
-		VXLANVNI:        egwVXLANVNI,
-		UsePSP:          r.usePSP,
-		OpenShift:       openshift,
-		NamespacedNames: namespacedNames,
+		PullSecrets:       pullSecrets,
+		Installation:      installation,
+		OSType:            rmeta.OSTypeLinux,
+		EgressGW:          egw,
+		VXLANPort:         egwVXLANPort,
+		VXLANVNI:          egwVXLANVNI,
+		UsePSP:            r.usePSP,
+		OpenShift:         openshift,
+		NamespaceAndNames: namespaceAndNames,
 	}
 
 	component := egressgateway.EgressGateway(config)
@@ -700,7 +700,7 @@ func getDegradedMsg(egw *operatorv1.EgressGateway) string {
 	return ""
 }
 
-func getEGWNamespacedNames(egws []operatorv1.EgressGateway) []string {
+func getEGWNamespaceAndNames(egws []operatorv1.EgressGateway) []string {
 	namespacedName := []string{}
 	for _, egw := range egws {
 		namespacedName = append(namespacedName, fmt.Sprintf("%s:%s", egw.Namespace, egw.Name))
--- a/pkg/controller/egressgateway/egressgateway_controller.go
+++ b/pkg/controller/egressgateway/egressgateway_controller.go
@@ -177,6 +177,7 @@ func (r *ReconcileEgressGateway) Reconcile(ctx context.Context, request reconcil
 		err := ch.CreateOrUpdateOrDelete(ctx, render.NewPassthrough(true, objects...), r.status)
 		if err != nil {
 			reqLogger.Error(err, "error deleting cluster scoped resources")
+			return reconcile.Result{}, nil
 		}
 		r.status.OnCRNotFound()
 		return reconcile.Result{}, nil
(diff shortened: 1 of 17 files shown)


pkg/controller/egressgateway/egressgateway_controller.go
```diff
@@ -158,7 +160,24 @@ func (r *ReconcileEgressGateway) Reconcile(ctx context.Context, request reconcil
...
+		}
+		// If provider is openshift, remove the SCC
+		if r.provider == operatorv1.ProviderOpenShift {
+			scc := egressgateway.SecurityContextConstraints()
+			objects = append(objects, scc)
+		}
+
+		err := ch.CreateOrUpdateOrDelete(ctx, render.NewPassthrough(true, objects...), r.status)
+		if err != nil {
+			reqLogger.Error(err, "error deleting cluster scoped resources")
```
caseydavenport, line 179: I think we should return an error if we encounter an error here - as it stands, we'll return `nil`

pkg/controller/egressgateway/egressgateway_controller.go
```diff
@@ -186,6 +205,28 @@ func (r *ReconcileEgressGateway) Reconcile(ctx context.Context, request reconcil
 			reqLogger.Info("EgressGateway object not found")
 			// Since the EGW resource is not found, remove the deployment.
 			r.status.RemoveDeployments(types.NamespacedName{Name: request.Name, Namespace: request.Namespace})
+
+			// In case of openShift, we are using a single SCC.
```
caseydavenport, line 209: nit: "In the case of OpenShift" (add "the", and also capitalized since it's a name)

pkg/controller/egressgateway/egressgateway_controller.go
```diff
@@ -186,6 +205,28 @@ func (r *ReconcileEgressGateway) Reconcile(ctx context.Context, request reconcil
...
+			// Whenever a EGW resource is deleted, remove the corresponding user from the SCC
+			// and update the resource.
+			if r.provider == operatorv1.ProviderOpenShift {
+				scc, err := getOpenShiftSCC(ctx, r.client)
+				if err != nil {
+					reqLogger.Error(err, "Error querying SecurityContextConstraints")
```
caseydavenport, line 215: we need to return an error here. If we do, we can also get rid of the `else` branch I think

pkg/controller/egressgateway/egressgateway_controller.go
```diff
@@ -186,6 +205,28 @@ func (r *ReconcileEgressGateway) Reconcile(ctx context.Context, request reconcil
 			reqLogger.Info("EgressGateway object not found")
 			// Since the EGW resource is not found, remove the deployment.
 			r.status.RemoveDeployments(types.NamespacedName{Name: request.Name, Namespace: request.Namespace})
+
+			// In case of openShift, we are using a single SCC.
+			// Whenever a EGW resource is deleted, remove the corresponding user from the SCC
+			// and update the resource.
+			if r.provider == operatorv1.ProviderOpenShift {
```
caseydavenport, line 212: We really should generate the list of names we need to create and pass that to the `render` package as part of the Config struct - that would better follow our established paradigms. 

e.g.,

```go
	config := &egressgateway.Config{
		PullSecrets:  pullSecrets,
		Installation: installation,
		OSType:       rmeta.OSTypeLinux,
		EgressGW:     egw,
		VXLANPort:    egwVXLANPort,
		VXLANVNI:     egwVXLANVNI,
		Openshift:    openshift,
		UsePSP:       r.usePSP,
                EgressGWNames: getEGWNames(egws),
	}

func getEGWNames(egws []v1.EgressGateway) []string { ... }
```

The render logic would then build the SCC, sort the given names (so that the output of the render package is deterministic), and include them in the produced SCC as serviceaccounts.

pkg/controller/egressgateway/egressgateway_controller.go
```diff
@@ -256,7 +297,7 @@ func (r *ReconcileEgressGateway) Reconcile(ctx context.Context, request reconcil
 	}
 
 	if installStatus.CalicoVersion != components.EnterpriseRelease {
-		reqLogger.Info("Waiting for expected version of Calico to be installed, expectedVersion = %s", components.EnterpriseRelease)
+		reqLogger.Info(fmt.Sprintf("Waiting for expected version of Calico to be installed, expectedVersion = %s", components.EnterpriseRelease))
```
caseydavenport, line 300: You should use `WithValues` here instead.

e.g.,

```go
reqLogger.WithValues("version", components.EnterpriseRelease).Info("Waiting for expected version of Calico to be installed")
```

pkg/controller/egressgateway/egressgateway_controller.go
sridhartigera: Does that mean, I don't need to create this as a Passthrough object? I think, passing egwNames to the SecurityContextConstraints directly will be beneficial. The reason being, when a EGW resource gets deleted, SCC must be updated and having egwNames outside the config can be helpful right? WDYT?

pkg/render/egressgateway/egressgateway.go
```diff
@@ -448,11 +452,19 @@ func (c *component) getHealthTimeoutDs() string {
...
+func (c *component) getSecurityContextConstraints() *ocsv1.SecurityContextConstraints {
+	scc := SecurityContextConstraints()
+	sort.Strings(c.config.NamespacedNames)
+	for _, egwNames := range c.config.NamespacedNames {
+		scc.Users = append(scc.Users, fmt.Sprintf("system:serviceaccount:%s", egwNames))
+	}
+	return scc
+}
+
+func SecurityContextConstraints() *ocsv1.SecurityContextConstraints {
```
caseydavenport, line 464: I feel like maybe I didn't explain myself very clearly in the last comment. Here is what I was imagining this code would look like:

```go
func (c *component) securityContextConstraints() *ocsv1.SecurityContextConstraints {
        // Create a slice of all the EGW service accounts on the cluster that we need to apply this SCC to.
        serviceAccountNames := []string
        for _, egw := range c.Config.NamespacedNames {
            serviceAccountNames = append(serviceAccountNames, fmt.Sprintf("system:serviceaccount:%s:%s", egw.Namespace, egw.Name)
        }
        sort.Strings(serviceAccountNames)

	return &ocsv1.SecurityContextConstraints{
		TypeMeta:                 metav1.TypeMeta{Kind: "SecurityContextConstraints", APIVersion: "security.openshift.io/v1"},
		ObjectMeta:               metav1.ObjectMeta{Name: namespacedName},
		ObjectMeta:               metav1.ObjectMeta{Name: OpenShiftSCCName},
		AllowHostDirVolumePlugin: true,
		AllowHostIPC:             false,
		AllowHostNetwork:     false,
		AllowHostPID:             false,
		AllowHostPorts:           false,
		AllowPrivilegeEscalation: ptr.BoolToPtr(true),
		AllowPrivilegedContainer: true,
		FSGroup:                  ocsv1.FSGroupStrategyOptions{Type: ocsv1.FSGroupStrategyRunAsAny},
		RunAsUser:                ocsv1.RunAsUserStrategyOptions{Type: ocsv1.RunAsUserStrategyRunAsAny},
		ReadOnlyRootFilesystem:   false,
		SELinuxContext:           ocsv1.SELinuxContextStrategyOptions{Type: ocsv1.SELinuxStrategyMustRunAs},
		SupplementalGroups:       ocsv1.SupplementalGroupsStrategyOptions{Type: ocsv1.SupplementalGroupsStrategyRunAsAny},
		Users: serviceAccountNames, 
		Volumes: []ocsv1.FSType{"*"},
		Users:                    []string{},
		Volumes:                  []ocsv1.FSType{"*"},
	}
}
```

The intention is to remove the logic from the controller that is inserting / removing entries from the list, and instead do this logic declaratively - given the same input, we always get the same output. It doesn't rely on us following a series of steps.

pkg/render/egressgateway/egressgateway.go
```diff
@@ -39,10 +40,12 @@ import (
...
 const (
-	egwPortName             = "health"
-	DefaultVXLANPort  int   = 4790
-	DefaultVXLANVNI   int   = 4097
-	DefaultHealthPort int32 = 8080
+	egwPortName                 = "health"
+	DefaultVXLANPort      int   = 4790
+	DefaultVXLANVNI       int   = 4097
+	DefaultHealthPort     int32 = 8080
+	PodSecurityPolicyName       = "tigera-egressgateway"
```
caseydavenport, line 47: I don't think these need to be exported?

pkg/render/egressgateway/egressgateway.go
caseydavenport: > The reason being, when a EGW resource gets deleted, SCC must be updated and having egwNames outside the config can be helpful right? WDYT?

Ah, right. If one is deleted, we won't call through to `Render` will we?

This is unfortunately proving more tricky than I thought!

caseydavenport: Perhaps the right thing is for the `SecurityContextConstraint` to be a completely separate render component than the rest of the EGW resources. 

That way, we can render the SCC every time the Reconcile function is called.

sridhartigera: Yeah. Even if one is deleted, we wont call through to ```Render```. I will move PSP and SCC to a new component which takes care of global resources in a separate PR.

pkg/controller/egressgateway/egressgateway_controller.go
```diff
@@ -643,3 +699,11 @@ func getDegradedMsg(egw *operatorv1.EgressGateway) string {
 	}
 	return ""
 }
+
+func getEGWNamespacedNames(egws []operatorv1.EgressGateway) []string {
```
caseydavenport, line 703: This is a little confusing, since a `NamespacedName` is a common type used in the operator and this function doesn't return that type...

caseydavenport (approved):
One comment, but otherwise LGTM per our discussion on Slack, we'll do a follow-on to clean up a few things.

PR #2431, opened 2023-01-28
Pass user and password to connect to ES

asincu:
## Description

Pass user and password when connecting to ES, alongside certificates. Enabled access between Linseed and Elastic, Linseed and Manager (es-proxy)

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/logstorage/linseed.go
+++ b/pkg/controller/logstorage/linseed.go
@@ -16,6 +16,7 @@ package logstorage
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/tigera/operator/pkg/render/logstorage/linseed"
 	"github.com/tigera/operator/pkg/render/monitor"
@@ -76,7 +77,7 @@ func (r *ReconcileLogStorage) createLinseed(
 
 	// This secret should only ever contain one key.
 	if len(esAdminUserSecret.Data) != 1 {
-		r.status.SetDegraded(operatorv1.ResourceValidationError, "Elasticsearch admin user secret contains too many entries", nil, reqLogger)
+		r.status.SetDegraded(operatorv1.ResourceValidationError, fmt.Sprintf("secret should have exactly one entry. found %d", len(esAdminUserSecret.Data)), nil, reqLogger)
 		return reconcile.Result{}, false, nil
 	}
 
@@ -92,7 +93,7 @@ func (r *ReconcileLogStorage) createLinseed(
 		TrustedBundle:   trustedBundle,
 		ClusterDomain:   r.clusterDomain,
 		KeyPair:         linseedKeyPair,
-		EsAdminUserName: esAdminUserName,
+		ESAdminUserName: esAdminUserName,
 	}
 
 	linseedComponent := linseed.Linseed(cfg)
--- a/pkg/controller/logstorage/linseed.go
+++ b/pkg/controller/logstorage/linseed.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2022 Tigera, Inc. All rights reserved.
+// Copyright (c) 2022,2023 Tigera, Inc. All rights reserved.
 
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -38,6 +38,7 @@ func (r *ReconcileLogStorage) createLinseed(
 	install *operatorv1.InstallationSpec,
 	variant operatorv1.ProductVariant,
 	pullSecrets []*corev1.Secret,
+	esAdminUserSecret *corev1.Secret,
 	hdler utils.ComponentHandler,
 	reqLogger logr.Logger,
 	ctx context.Context,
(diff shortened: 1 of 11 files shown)


pkg/render/logstorage/linseed/linseed.go
```diff
@@ -86,6 +86,9 @@ type Config struct {
 
 	// ClusterDomain to use when building service URLs.
 	ClusterDomain string
+
+	// EsAdminUserName is the admin user used to connect to Elastic
+	EsAdminUserName string
```
caseydavenport, line 91: nit: ESAdminUserName

pkg/render/logstorage/linseed/linseed_test.go
```diff
@@ -86,9 +86,10 @@ var _ = Describe("Linseed rendering tests", func() {
 				PullSecrets: []*corev1.Secret{
 					{ObjectMeta: metav1.ObjectMeta{Name: "tigera-pull-secret"}},
 				},
-				KeyPair:       kp,
-				TrustedBundle: bundle,
-				ClusterDomain: clusterDomain,
+				KeyPair:         kp,
+				TrustedBundle:   bundle,
+				ClusterDomain:   clusterDomain,
+				EsAdminUserName: "elastic",
```
caseydavenport, line 92: I think long term we're going to want separate user names for each instance of Linseed we install, but this is OK for now

pkg/render/logstorage/linseed/linseed_test.go
```diff
@@ -376,7 +377,11 @@ func expectedContainers() []corev1.Container {
 				PeriodSeconds:       5,
 			},
 			Env: []corev1.EnvVar{
-				{Name: "LINSEED_LOG_LEVEL", Value: "INFO"},
+				{
```
caseydavenport, line 380: We don't need to explicitly nil out ValueFrom, since that's the default value

pkg/render/logstorage/linseed/linseed_test.go
```diff
@@ -395,6 +400,37 @@ func expectedContainers() []corev1.Container {
...
+					Name:      "LINSEED_ELASTIC_USERNAME",
+					Value:     "elastic",
+					ValueFrom: nil,
+				},
+				{
+					Name:  "LINSEED_ELASTIC_PASSWORD",
+					Value: "",
+					ValueFrom: &corev1.EnvVarSource{
+						FieldRef:         nil,
```
caseydavenport, line 412: All of these can just be omitted - no need to explicitly set them to their default value, just makes the code more verbose.

pkg/controller/logstorage/linseed.go
```diff
@@ -73,12 +74,25 @@ func (r *ReconcileLogStorage) createLinseed(
 	}
 	trustedBundle := certificateManager.CreateTrustedBundle(esInternalCertificate, prometheusCertificate)
 
+	// This secret should only ever contain one key.
+	if len(esAdminUserSecret.Data) != 1 {
+		r.status.SetDegraded(operatorv1.ResourceValidationError, "Elasticsearch admin user secret contains too many entries", nil, reqLogger)
```
caseydavenport, line 79: In theory, data could be of length 0, in which case the error should say "not enough entries"

Maybe just adjust the log to say "secret should have exactly one entry. found %d"

pkg/render/logstorage/linseed/linseed.go
```diff
@@ -183,6 +186,17 @@ func (e linseed) linseedDeployment() *appsv1.Deployment {
 
 		// Configuration for connection to Elasticsearch.
 		{Name: "LINSEED_ELASTIC_ENDPOINT", Value: ElasticsearchHTTPSEndpoint},
+		{Name: "LINSEED_ELASTIC_USERNAME", Value: e.cfg.EsAdminUserName},
```
caseydavenport, line 189: I believe we have some common logic for handling ES connection vars. 

e.g., https://github.com/tigera/operator/blob/ddbaa22cd1b351564e952b851bc1b6b5d0aeae0d/pkg/render/compliance.go#L391

Full thing here: https://github.com/tigera/operator/blob/ddbaa22cd1b351564e952b851bc1b6b5d0aeae0d/pkg/render/common/elasticsearch/decorator.go

asincu: Sure. I think we are not using the prefix of the component to set those variables, so I might need to make some changes in Linseed for this.

asincu: @caseydavenport I see that this is not setting some environments as I would expect them.
 - ELASTIC_HOST points to es-gateway 
 - ELASTIC_PASSWORD is using the same secret name as the one given as parameter. For the moment, we want user `elastic` and password from `tigera-secure-es-elastic-user`

caseydavenport (approved):

PR #2450, opened 2023-02-03
Omit updates for NetworkPolicy/Tier when Spec is identical

pasanw:
## Description
Omits updates for NetworkPolicy and Tier when Spec identical. The defaulting and conversion code paths were validated to ensure that we can safely perform a DeepEquals check on the spec.

This only addresses a subset of unnecessary PUTs, a broader solution will be explored.

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/utils/component.go
+++ b/pkg/controller/utils/component.go
@@ -449,6 +449,20 @@ func mergeState(desired client.Object, current runtime.Object) client.Object {
 		// APIServer.
 		dui.SetOwnerReferences(cui.GetOwnerReferences())
 		return dui
+	case *v3.NetworkPolicy:
+		cnp := current.(*v3.NetworkPolicy)
+		dnp := desired.(*v3.NetworkPolicy)
+		if reflect.DeepEqual(cnp.Spec, dnp.Spec) {
+			return nil
+		}
+		return dnp
+	case *v3.Tier:
+		ct := current.(*v3.Tier)
+		dt := desired.(*v3.Tier)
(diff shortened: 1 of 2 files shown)


pkg/controller/utils/component.go
```diff
@@ -449,6 +449,20 @@ func mergeState(desired client.Object, current runtime.Object) client.Object {
...
+		dnp := desired.(*v3.NetworkPolicy)
+		if reflect.DeepEqual(cnp.Spec, dnp.Spec) {
+			return nil
+		}
+		return dnp
+	case *v3.Tier:
+		ct := current.(*v3.Tier)
+		dt := desired.(*v3.Tier)
+		if reflect.DeepEqual(ct.Spec, dt.Spec) {
+			return nil
```
pasanw, line 463: Other resources in this file return the current object when they don't want a change. I assume we should change those to nil?

Didn't want to bring that in discussion/testing in scope here though.

pkg/controller/utils/component_test.go
```diff
@@ -1388,6 +1387,65 @@ var _ = Describe("Mocked client Component handler tests", func() {
 
 		Expect(mc.Index).To(Equal(4))
 	})
+
+	It("NetworkPolicy updates are omitted if there is no change", func() {
```
rene-dekker, line 1391: PR lgtm, but it would be nice to have a similar test case for a scenario where there is a change.

rene-dekker (approved):
lgtm

PR #2498, opened 2023-02-23
mTLS between linseed and es-proxy

caseydavenport:
## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


pkg/controller/logstorage/linseed.go
```diff
@@ -45,15 +45,15 @@ func (r *ReconcileLogStorage) createLinseed(
 	ctx context.Context,
 	certificateManager certificatemanager.CertificateManager,
 ) (reconcile.Result, bool, error) {
-	svcDNSNames := dns.GetServiceDNSNames(render.ElasticsearchServiceName, render.ElasticsearchNamespace, r.clusterDomain)
-	svcDNSNames = append(svcDNSNames, dns.GetServiceDNSNames(linseed.ServiceName, render.ElasticsearchNamespace, r.clusterDomain)...)
-
+	// Create a KeyPair for Linseed to present to clients.
+	svcDNSNames := dns.GetServiceDNSNames(linseed.ServiceName, render.ElasticsearchNamespace, r.clusterDomain)
 	linseedKeyPair, err := certificateManager.GetOrCreateKeyPair(r.client, render.TigeraLinseedSecret, common.OperatorNamespace(), svcDNSNames)
```
rene-dekker, line 50: If I am not mistaken the linseedKeyPair is not actually rendered anywhere in this branch.
Like so (in esgateway.go for example):
```
	certificateComponent := rcertificatemanagement.CertificateManagement(&rcertificatemanagement.Config{
		Namespace:       render.ElasticsearchNamespace,
		ServiceAccounts: []string{esgateway.ServiceAccountName},
		KeyPairOptions: []rcertificatemanagement.KeyPairOption{
			rcertificatemanagement.NewKeyPairOption(gatewayKeyPair, true, true),
		},
		TrustedBundle: trustedBundle,
	})
```

pkg/controller/manager/manager_controller.go
```diff
@@ -327,15 +328,16 @@ func (r *ReconcileManager) Reconcile(ctx context.Context, request reconcile.Requ
...
+	// Determine if compliance is enabled.
 	complianceLicenseFeatureActive := utils.IsFeatureActive(license, common.ComplianceFeature)
 	complianceCR, err := compliance.GetCompliance(ctx, r.client)
 	if err != nil && !errors.IsNotFound(err) {
 		r.status.SetDegraded(operatorv1.ResourceReadError, "Error querying compliance: ", err, reqLogger)
 		return reconcile.Result{}, err
 	}
 
+	// Build a trusted bundle containing all of the certificates of components that communicate with the manager pod.
+	trustedSecretNames := []string{render.PacketCaptureCertSecret, monitor.PrometheusTLSSecretName, relasticsearch.PublicCertSecret, render.ProjectCalicoApiServerTLSSecretName(installation.Variant)}
```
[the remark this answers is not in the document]
caseydavenport: I am not sure it needs it, because CreateTrustedBundle already includes the Calico root certificate which can be used to verify any certs signed by this operator. 

I'm not 100% sure how all of this logic works though, so maybe we should have a chat offline?

pkg/controller/logstorage/linseed.go
caseydavenport: Doesn't this call to `GetOrCreateKeyPair` create the keypair if it doesn't exist?

pkg/render/manager.go
```diff
@@ -109,17 +111,28 @@ func Manager(cfg *ManagerConfiguration) (Component, error) {
...
-	InternalTrafficSecret   certificatemanagement.KeyPairInterface
+	KeyValidatorConfig authentication.KeyValidatorConfig
+	ESSecrets          []*corev1.Secret
+	ESClusterConfig    *relasticsearch.ClusterConfig
+	PullSecrets        []*corev1.Secret
+	Openshift          bool
+	Installation       *operatorv1.InstallationSpec
+	ManagementCluster  *operatorv1.ManagementCluster
+
+	// If provided, the KeyPair to used for external connections to Voltron.
```
asincu, line 122: I am not sure this comment is correct. I think need it for external connections for users that access the UI ? My memory could play tricks on me though ...

pkg/render/manager.go
```diff
@@ -475,13 +488,23 @@ func (c *managerComponent) volumeMountsForProxyManager() []corev1.VolumeMount {
 
 // managerEsProxyContainer returns the ES proxy container
 func (c *managerComponent) managerEsProxyContainer() corev1.Container {
+	var keyPath, certPath string
+	if c.cfg.TLSKeyPair != nil {
+		keyPath, certPath = c.cfg.TLSKeyPair.VolumeMountKeyFilePath(), c.cfg.TLSKeyPair.VolumeMountCertificateFilePath()
```
asincu, line 493: Shouldn't we use the trusted bundle instead of this certificate ? Isn't this created for tigera-manager service ?

pkg/controller/manager/manager_controller.go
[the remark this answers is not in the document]
rene-dekker: You are right, but you are creating a gap in functionality though. The certificate management stuff is designed in a way that each tls secret can be replaced by a user with their own by dropping it in the operator namespace. When a controller runs it reads the key pair from the operator namespace and copies it over to the target namespace. You are missing this part. Also the certificate management is taking care of the installation.spec.certificateManagement feature.

pkg/controller/logstorage/linseed.go
rene-dekker: It fetches the existing copy from the operator namespace, which never exists in your case and because it does not, it creates a new instance. It does not store it in k8s. You still need this code snippet I made above to render it in the operator namespace and the target namespace where the pod picks it up.

caseydavenport: It looks like we have that code here right now: https://github.com/tigera/operator/blob/0146a0b1b2d3c9995812d617b00cd6b12dd8330a/pkg/controller/logstorage/linseed.go#L107-L114

pkg/render/manager.go
caseydavenport: This is one of the confusing things about how this works (and that we will be fixing in the near future).

Voltron is both the tunnel manager, and ALSO the reverse-proxy for the Manager UI - all requests to the UI come in through Voltron (which is run within the manager pod).

So Voltron is technically the correct word right now - Voltron terminates the TLS connection and proxies it using HTTP (not encrypted) to localhost within the pod, where the manager nginx container ultimately handles the request.

pkg/render/manager.go
caseydavenport: The trusted bundle is used to verify the authenticity of Linseed's certificate, but we need to mount our own certificate to present to Linseed in order to prove our own identity.

pkg/render/manager.go
caseydavenport: The comment is slightly wrong as of this PR since it is also used for es-proxy -> Linseed

pkg/controller/apiserver/apiserver_controller.go
```diff
@@ -262,8 +262,8 @@ func (r *ReconcileAPIServer) Reconcile(ctx context.Context, request reconcile.Re
 	}
 
 	// We need separate certificates for OSS vs Enterprise.
-	secretName := render.ProjectCalicoApiServerTLSSecretName(network.Variant)
-	tlsSecret, err := certificateManager.GetOrCreateKeyPair(r.client, secretName, common.OperatorNamespace(), dns.GetServiceDNSNames(render.ProjectCalicoApiServerServiceName(network.Variant), rmeta.APIServerNamespace(network.Variant), r.clusterDomain))
+	secretName := render.ProjectCalicoAPIServerTLSSecretName(network.Variant)
```
caseydavenport, line 265: Sigh, I couldn't help myself - Api instead of API was killing me.

pkg/controller/manager/manager_controller.go
```diff
@@ -327,15 +328,24 @@ func (r *ReconcileManager) Reconcile(ctx context.Context, request reconcile.Requ
...
 	}
 
+	// Build a trusted bundle containing all of the certificates of components that communicate with the manager pod.
+	// This bundle contains the root CA used to sign all operator-generated certificates, as well as the explicitly named
+	// certificates, in case the user has provided their own cert in lieu of the default certificate.
+	trustedSecretNames := []string{
+		render.PacketCaptureCertSecret,
+		monitor.PrometheusTLSSecretName,
+		relasticsearch.PublicCertSecret,
+		render.ProjectCalicoAPIServerTLSSecretName(installation.Variant),
```
caseydavenport, line 346: I'm not sure that the API server TLS secret is actually needed here, but it was pre-existing so leaving it for now.

Nothing should be talking directly to the Calico API server, AFAIK

pkg/controller/logstorage/linseed.go
```diff
@@ -45,15 +45,17 @@ func (r *ReconcileLogStorage) createLinseed(
 	ctx context.Context,
 	certificateManager certificatemanager.CertificateManager,
 ) (reconcile.Result, bool, error) {
-	svcDNSNames := dns.GetServiceDNSNames(render.ElasticsearchServiceName, render.ElasticsearchNamespace, r.clusterDomain)
-	svcDNSNames = append(svcDNSNames, dns.GetServiceDNSNames(linseed.ServiceName, render.ElasticsearchNamespace, r.clusterDomain)...)
-
+	// Create a server KeyPair for Linseed to present to clients.
+	// This fetches the existing key pair from the tigera-operator namespace if it exists, or generates a new one in-memory otherwise.
+	// It will be provisoined into the cluster in the render stage later on.
```
asincu, line 50: Typo: `provisoined`

pkg/controller/logstorage/linseed.go
rene-dekker: That is a problem. The trustedbundle should only be rendered once, otherwise two controllers are competing and the last one wins.

I think it should probably be rendered from the logstorage_controller and not from esgateway.go and here.

asincu (approved):

pkg/controller/manager/manager_controller.go
rene-dekker: The apiserver deploy also has a queryserver. Both mount the same TLS secret. Maybe that is why?

pkg/render/manager.go
```diff
@@ -475,13 +488,23 @@ func (c *managerComponent) volumeMountsForProxyManager() []corev1.VolumeMount {
 
 // managerEsProxyContainer returns the ES proxy container
 func (c *managerComponent) managerEsProxyContainer() corev1.Container {
+	var keyPath, certPath string
+	if c.cfg.TLSKeyPair != nil {
```
rene-dekker, line 492: nit: This can never be nil.

pkg/controller/manager/manager_controller.go
caseydavenport: Hm, yeah maybe that's it.

pkg/render/manager.go
caseydavenport: True, unless we write a bug into the code somewhere I guess.

caseydavenport: I copied this check from other places that we access it. I suggest we keep the check but add a comment so nobody is confused.

pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -349,6 +352,128 @@ func setLogStorageFinalizer(ls *operatorv1.LogStorage) {
...
+		c.gateway,
+	)
+}
+
+func (c *keyPairCollection) component(bundle certificatemanagement.TrustedBundle) render.Component {
+	// Create a render.Component to provision or update key pairs and the trusted bundle.
+	return rcertificatemanagement.CertificateManagement(&rcertificatemanagement.Config{
+		Namespace: render.ElasticsearchNamespace,
+		ServiceAccounts: []string{
+			// TODO: If any of these SA need a CSR binding, all of them get it. Is this OK?
```
rene-dekker, line 382: Yes, this is right.

pkg/render/fluentd_test.go
```diff
@@ -230,7 +230,7 @@ var _ = Describe("Tigera Secure Fluentd rendering tests", func() {
 		}
 
 		cfg.OSType = rmeta.OSTypeWindows
-		cfg.MetricsServerTLS = nil
+		cfg.FluentdKeyPair = nil
```
asincu, line 233: I have noticed I have the following variables set locally:

```
		{Name: "LINSEED_ENABLED", Value: "true"},
		{Name: "LINSEED_ENDPOINT", Value: "https://tigera-linseed.tigera-elasticsearch.svc"},
		{Name: "LINSEED_CA_PATH", Value: c.cfg.TrustedBundle.MountPath()},

```

caseydavenport: What about them?

pkg/controller/logstorage/logstorage.go
```diff
@@ -190,35 +173,21 @@ func (r *ReconcileLogStorage) createLogStorage(
...
 		r.status.SetDegraded(operatorv1.ResourceUpdateError, "Error with images from ImageSet", err, reqLogger)
 		return reconcile.Result{}, false, finalizerCleanup, err
 	}
 
-	components = append(components, component,
-		rcertificatemanagement.CertificateManagement(&rcertificatemanagement.Config{
-			Namespace:       render.ElasticsearchNamespace,
-			ServiceAccounts: []string{render.ElasticsearchName},
-			KeyPairOptions: []rcertificatemanagement.KeyPairOption{
-				// We do not want to delete the secret from the tigera-elasticsearch namespace when CertificateManagement is
```
rene-dekker, line 205: Could you move this comment to logstorage_controller:404?

pkg/controller/logstorage/logstorage.go
```diff
@@ -190,35 +173,21 @@ func (r *ReconcileLogStorage) createLogStorage(
...
-	)
+	components = append(components, logStorageComponent)
 	if !operatorv1.IsFIPSModeEnabled(install.FIPSMode) {
+		// Render the key pair and trusted bundle into the Kibana namespace, which should be created by the log storage component above.
 		components = append(components,
 			rcertificatemanagement.CertificateManagement(&rcertificatemanagement.Config{
 				Namespace:       render.KibanaNamespace,
 				ServiceAccounts: []string{render.KibanaName},
 				KeyPairOptions: []rcertificatemanagement.KeyPairOption{
-					// We do not want to delete the secret from the tigera-elasticsearch when CertificateManagement is
```
rene-dekker, line 219: Why do you want to delete this comment? It is different from most other cases for a reason.

pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -349,6 +352,171 @@ func setLogStorageFinalizer(ls *operatorv1.LogStorage) {
...
+		return nil, err
+	} else if esProxyCert == nil {
+		// ES proxy is deployed after the first execution of this controller, so it is normal for the
+		// secret to not exist. We'll add it to the trusted bundle when it becomes avaialble though.
+		log.Info("es-proxy cert not available yet, will add it to the bundle when it becomes available")
+	} else {
+		collection.esProxy = esProxyCert
+	}
+
+	// Get certificate for fluntd, which Linseed needs to trust in a standalone or management cluster.
```
rene-dekker, line 449: nit: flunt

pkg/controller/logstorage/esgateway.go
```diff
@@ -44,53 +40,15 @@ func (r *ReconcileLogStorage) createEsGateway(
...
-		r.status.SetDegraded(operatorv1.ResourceReadError, "Failed to get certificate", err, reqLogger)
-		return reconcile.Result{}, nil, false, err
-	} else if prometheusCertificate == nil {
-		reqLogger.Info("Prometheus secrets are not available yet, waiting until they become available")
-		r.status.SetDegraded(operatorv1.ResourceNotReady, "Prometheus secrets are not available yet, waiting until they become available", nil, reqLogger)
-		return reconcile.Result{}, nil, false, nil
-	}
-	trustedBundle := certificateManager.CreateTrustedBundle(esInternalCertificate, kibanaCertificate, prometheusCertificate, gatewayKeyPair)
-
+	esInternalCertificate certificatemanagement.KeyPairInterface,
```
rene-dekker, line 43:
```
	certificateManager certificatemanager.CertificateManager,
	esInternalCertificate certificatemanagement.KeyPairInterface,
	kibanaCertificate certificatemanagement.KeyPairInterface,
```
are never used.

pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -657,7 +857,7 @@ func (r *ReconcileLogStorage) Reconcile(ctx context.Context, request reconcile.R
 			return result, err
 		}
 
-		result, proceed, err = r.createEsMetrics(
+		result, proceed, err = r.createESMetrics(
```
rene-dekker, line 860: clusterdomain is never used anymore.

pkg/controller/logstorage/linseed.go
```diff
@@ -44,37 +41,9 @@ func (r *ReconcileLogStorage) createLinseed(
 	reqLogger logr.Logger,
 	ctx context.Context,
 	certificateManager certificatemanager.CertificateManager,
```
rene-dekker, line 43: certificatemanager is never used

pkg/controller/logstorage/logstorage.go
```diff
@@ -76,11 +75,12 @@ func (r *ReconcileLogStorage) createLogStorage(
 	certificateManager certificatemanager.CertificateManager,
```
rene-dekker, line 75: certificatemanager is no longer used.

pkg/controller/logstorage/logstorage.go
caseydavenport: Yeah, I deleted these I think because they didn't really make sense to me. I'll add it back.

pkg/controller/logstorage/logstorage_controller.go
caseydavenport: it is - it's passed to the render code and used there

PR #2499, opened 2023-02-23
Copy pull secrets to the EGW namespace

sridhartigera:
## Description
The changes cover the steps to copy
1. Pull secrets to the EGW namespace. If the pull secrets are already present, the ownerReferences in the secret is updated to include the new resource.
2. Patch ```policySyncPathPrefix``` in the felix config.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/common/common.go
+++ b/pkg/common/common.go
@@ -33,4 +33,6 @@ const (
 	TiersFeature = "tiers"
 	// EgressAccessControl enables creation/update of NetworkPolicy with Domains
 	EgressAccessControlFeature = "egress-access-control"
+	// MultipleOwnersLabel used to indicate multiple owner references.
+	MultipleOwnersLabel = "operator.tigera.io/multipleOwners"
 )
--- a/pkg/common/common.go
+++ b/pkg/common/common.go
@@ -34,5 +34,8 @@ const (
 	// EgressAccessControl enables creation/update of NetworkPolicy with Domains
 	EgressAccessControlFeature = "egress-access-control"
 	// MultipleOwnersLabel used to indicate multiple owner references.
+	// If the render code places this label on an object, the object mergeState machinery will merge owner
+	// references with any that already exist on the object rather than replace the owner references. Further
+	// the controller in the owner reference will not be set.
 	MultipleOwnersLabel = "operator.tigera.io/multipleOwners"
 )
--- a/pkg/controller/egressgateway/egressgateway_controller.go
+++ b/pkg/controller/egressgateway/egressgateway_controller.go
@@ -324,13 +324,6 @@ func (r *ReconcileEgressGateway) Reconcile(ctx context.Context, request reconcil
 	// Reconcile all the EGWs
 	var errMsgs []string
 	for _, egw := range egwsToReconcile {
-		// Check if there are pull secrets in the EGW namespace.
-		// If present, use the existing pull secrets, as owner references need to
-		// be updated.
-		nsSecrets, err := r.getNetworkingPullSecrets(pullSecrets, egw.Namespace)
-		if err == nil {
-			pullSecrets = nsSecrets
-		}
 		err = r.reconcileEgressGateway(ctx, &egw, reqLogger, variant, fc, pullSecrets, installation, namespaceAndNames)
 		if err != nil {
 			reqLogger.Error(err, "Error reconciling egress gateway")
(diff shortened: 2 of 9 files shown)


pkg/controller/egressgateway/egressgateway_controller.go
```diff
@@ -423,6 +430,40 @@ func (r *ReconcileEgressGateway) reconcileEgressGateway(ctx context.Context, egw
...
+	}
+
+	return secrets, nil
+}
+
+func (r *ReconcileEgressGateway) patchFelixConfig(ctx context.Context) (*crdv1.FelixConfiguration, error) {
+	// Fetch any existing default FelixConfiguration object.
+	fc := &crdv1.FelixConfiguration{}
+	err := r.client.Get(ctx, types.NamespacedName{Name: "default"}, fc)
+
```
caseydavenport, line 451: nit, but conventionally in golang error branches should go right below the call that generates them, with no newline.

```go
	err := r.client.Get(ctx, types.NamespacedName{Name: "default"}, fc)
		r.status.SetDegraded(operatorv1.ResourceReadError, "Unable to read FelixConfiguration", err, log)
		return nil, err
	}
```

pkg/controller/egressgateway/egressgateway_controller.go
```diff
@@ -423,6 +430,40 @@ func (r *ReconcileEgressGateway) reconcileEgressGateway(ctx context.Context, egw
...
+	if err != nil && !errors.IsNotFound(err) {
+		r.status.SetDegraded(operatorv1.ResourceReadError, "Unable to read FelixConfiguration", err, log)
+		return nil, err
+	}
+
+	patchFrom := client.MergeFrom(fc.DeepCopy())
+	if fc.Spec.PolicySyncPathPrefix == "" {
+		fc.Spec.PolicySyncPathPrefix = "/var/run/nodeagent"
+	}
+	if err := r.client.Patch(ctx, fc, patchFrom); err != nil {
```
caseydavenport, line 461: We should only send the patch if we needed to change the felix configuration, if the value was already correct, we should skip the `Patch` call.

pkg/controller/egressgateway/egressgateway_controller.go
```diff
@@ -423,6 +430,40 @@ func (r *ReconcileEgressGateway) reconcileEgressGateway(ctx context.Context, egw
...
 
+func (r *ReconcileEgressGateway) getNetworkingPullSecrets(pullSecrets []*v1.Secret, ns string) ([]*v1.Secret, error) {
+	secrets := []*v1.Secret{}
+	for _, ps := range pullSecrets {
+		s := &v1.Secret{}
+		err := r.client.Get(context.Background(), client.ObjectKey{Name: ps.Name, Namespace: ns}, s)
+		if err != nil {
+			return nil, err
+		}
+		secrets = append(secrets, s)
```
caseydavenport, line 441: I don't think the implementation here is quite right. We need to account for the case where the user might modify the secrets in the tigera-operator namespace and update the secrets in this namespace accordingly. Using the current approach, we'll throw away any changed secret and keep using the old version of it, assuming it already exists in the target namespace.

I think the best option would be to update the `mergeState` logic within the component handler utility so that it knows how to handle multiple owner references. This might involve adding some context to indicate whether we want to merge owner references, or overwrite them.

https://github.com/tigera/operator/blob/c73b79ad5dbee7f410b57a84506603d7cc68b2df/pkg/controller/utils/component.go#L301-L302

The main benefit of that is that we're not introducing yet-another Get() call, and it also puts the logic in a place that is common for all controllers, should we need to do something similar in the future in another controller.

Another alternative would be to modify this code to be more like the following, but I think this is less ideal:

```go
existing := &v1.Secret{}
err := r.client.Get(context.Background(), client.ObjectKey{Name: ps.Name, Namespace: ns}, existing)
if IsNotExist(err) {
    // Secret doesn't exist, create it.
    continue
} else {
    // General error.
    return nil, err
}

// Update the pull secret with existing owner references. We need to maintain these across
// updates, since a secret may have multiple owners.
ps.OwnerReferences = existing.OwnerReferences
```

pkg/controller/egressgateway/egressgateway_controller.go
```diff
@@ -423,6 +430,40 @@ func (r *ReconcileEgressGateway) reconcileEgressGateway(ctx context.Context, egw
 	return nil
 }
 
+func (r *ReconcileEgressGateway) getNetworkingPullSecrets(pullSecrets []*v1.Secret, ns string) ([]*v1.Secret, error) {
```
caseydavenport, line 433: We should pass a context into this function and use it, rather than using `context.Background()`

caseydavenport (changes requested):

pkg/render/egressgateway/egressgateway.go
```diff
@@ -287,6 +290,18 @@ func (c *component) egwInitEnvVars() []corev1.EnvVar {
 	}
 }
 
+func (c *component) egwPullSecrets() []*corev1.Secret {
+	var secrets []*corev1.Secret
+	for _, secret := range c.config.PullSecrets {
+		x := secret.DeepCopy()
+		x.ObjectMeta = metav1.ObjectMeta{Name: secret.Name, Namespace: c.config.EgressGW.Namespace}
+		x.ObjectMeta.Labels = common.MapExistsOrInitialize(x.ObjectMeta.Labels)
+		x.ObjectMeta.Labels[common.MultipleOwnersLabel] = "true"
```
caseydavenport, line 299: We should have a comment here explaining why we put the multiple owners label on.

Something like this:

> Each pull secret is shared across all of the EGW deployments in this namespace. As such, we mark it as having multiple owners so that we maintain multiple owner references when creating the secret so that it will only be GC'd when all of its owners have been deleted.

pkg/common/common.go
```diff
@@ -33,4 +33,6 @@ const (
 	TiersFeature = "tiers"
 	// EgressAccessControl enables creation/update of NetworkPolicy with Domains
 	EgressAccessControlFeature = "egress-access-control"
+	// MultipleOwnersLabel used to indicate multiple owner references.
```
caseydavenport, line 36: I think this needs more detail as well.

> // MultipleOwnersLabel used to indicate multiple owner references. If the render code places this label on an object, the object mergeState machinery will merge owner references with any that already exist on the object rather than replace the owner references.

caseydavenport (approved):
LGTM, but please enhance those couple of comments that I called out so that the behavior isn't too cryptic for anyone looking at this in the future.

PR #2503, opened 2023-02-24
Expose logging config for calico cni

ti-afra:
## Description
This change is exposing a new logging property in the installation object. With this change users are able to provide customized logging configuration for calico cni plugin via tigera-operator.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


api/v1/installation_types.go
```diff
@@ -178,6 +179,35 @@ type InstallationSpec struct {
...
+	CNILogging *CNILogging `json:"cniLogging,omitempty"`
+}
+
+type CNILogging struct {
+	// Default: Info
+	// +kubebuilder:validation:Enum=Error;Warning;Debug;Info
+	// +optional
+	LogSeverity *LogLevel `json:"logSeverity,omitempty"`
+
+	// Default: 100Mi
```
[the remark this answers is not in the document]
ti-afra: In order for the default values set by kubebuilder, I need to remove the code in core_controller (fillDefaults). O.w., the values set in fillDefaults will take place.
Looks like environment is set differently for some tests and default values are not set via kubebuilder there. Thus, the side effect of not setting default values in fillDefaults is that some tests are failing and we need to check for nil values while we are setting configmap (node.go). This means that for some of tests, cni-config will not have the logging properties. This is a different behaviour from production, where defaults are always set via kubebuilder and therefore the logging values will always be present in cni-config (whether user provides it or not).

What should we do in this case?
I am more towards using fillDefaults over kubebuilder since this is how default values for most of the properties in the installation object is defined already and tests are build around this setup.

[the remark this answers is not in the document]
ti-afra: note: when we implement defaulting both in fillDefaults and kubebuilder, the ones in fillDefault are what we see and values set in kubebuilder:default tag are being ignored (I am not sure what is happening behind the scene - this is just observations from testing various scenarios)

pkg/controller/installation/core_controller.go
```diff
@@ -692,6 +693,38 @@ func fillDefaults(instance *operator.Installation) error {
 			mm := operator.MultiInterfaceModeNone
 			instance.Spec.CalicoNetwork.MultiInterfaceMode = &mm
 		}
+
+		// setting default values for calico-cni logging configuration when not provided by the user
+		if instance.Spec.Logging == nil {
```
rene-dekker, line 698: Should we only be setting the logging struct if we are using the calico cni?

api/v1/installation_types.go
```diff
@@ -178,6 +179,35 @@ type InstallationSpec struct {
...
+
+	// Logging Configuration for Components
+	// +optional
+	Logging *Logging `json:"logging,omitempty"`
+}
+
+type Logging struct {
+	// Customized logging specification for calico-cni plugin
+	// +optional
+	CNILogging *CNILogging `json:"cniLogging,omitempty"`
```
caseydavenport, line 191: I would personally call this field `CNI` - it's already within the `Logging` struct so we don't need to say logging again.

```yaml
logging:
  cni:
    logSeverity: Info
```

You can (and should) still call the struct `CNILogging`, I'm just suggesting renaming the field name (and json tag)

pkg/controller/installation/validation.go
```diff
@@ -54,6 +56,26 @@ func validateCustomResource(instance *operatorv1.Installation) error {
 				),
 			)
 		}
+
+		// validate logging config for calico-cni
+		if instance.Spec.Logging != nil && instance.Spec.Logging.CNILogging != nil {
```
caseydavenport, line 61: I think we have one last semantic problem with this API we need to at least conceptualize a solution to (if not implement), which is how to let users turn off file logging altogether.

Right now, it's impossible for a user to set these field to `nil` (because our defaulting will fill them in) and it's impossible for a user to set these fields to zero (because our validation will error). 

So basically there's no way to turn off the logging altogether. 

It's a tricky situation - usually, we would say that specifying `spec.Logging.CNILoggin: nil` means "disable this feature", however we need to be backwards compatible and doing this would mean that old clusters would lose file logging on upgrade, so our defaulting needs to fill this in.

We might need a `fileLogging: Enabled` field in this struct that defaults to `Enabled` but can be set to `Disabled`.

That said, there's no way to disable file loggingt _today_ prior to your changes, so this isn't a regression and I think maybe we shouldn't block this PR on this. Just wanted to call it out so you're aware.

pkg/render/node.go
```diff
@@ -492,14 +493,7 @@ func (c *nodeComponent) nodeRole() *rbacv1.ClusterRole {
...
 
-// nodeCNIConfigMap returns a config map containing the CNI network config to be installed on each node.
-// Returns nil if no configmap is needed.
-func (c *nodeComponent) nodeCNIConfigMap() *corev1.ConfigMap {
-	if c.cfg.Installation.CNI.Type != operatorv1.PluginCalico {
-		// If calico cni is not being used, then no cni configmap is needed.
-		return nil
-	}
-
+func (c *nodeComponent) createCalicoPlugin() map[string]interface{} {
```
caseydavenport, line 496: maybe `createCalicoPluginConfig`

pkg/render/node.go
```diff
@@ -520,62 +514,112 @@ func (c *nodeComponent) nodeCNIConfigMap() *corev1.ConfigMap {
...
 	}
 
-	// Build the CNI configuration json.
+	return portmapPlugin
+}
+
+// nodeCNIConfigMap returns a config map containing the CNI network config to be installed on each node.
+// Returns nil if no configmap is needed.
+func (c *nodeComponent) nodeCNIConfigMap() *corev1.ConfigMap {
+
```
caseydavenport, line 599: very much a nit, but this newline can be removed

pkg/render/node.go
```diff
@@ -520,62 +514,112 @@ func (c *nodeComponent) nodeCNIConfigMap() *corev1.ConfigMap {
...
+	if c.cfg.Installation.CNI.Type != operatorv1.PluginCalico {
+		// If calico cni is not being used, then no cni configmap is needed.
+		return nil
+	}
+
+	var plugins = make([]interface{}, 0)
+
```
caseydavenport, line 606: same here - very nitty, but good to keep the `var plugins` definition in the same block as the population below, so can remove this newline.

pkg/render/node.go
```diff
@@ -604,24 +648,38 @@ func (c *nodeComponent) getCalicoIPAM() string {
...
 	v6 := GetIPv6Pool(cns.IPPools) != nil
 	v4 := GetIPv4Pool(cns.IPPools) != nil
 
 	if v4 && v6 {
 		// Dual-stack
-		return `{ "type": "host-local", "ranges": [[{"subnet": "usePodCidr"}],[{"subnet": "usePodCidrIPv6"}]]}`
+		return map[string]interface{}{
+			"type": "host-local",
+			"ranges": [][]map[string]string{
+				[]map[string]string{{"subnet": "usePodCidr"}},
```
caseydavenport, line 667: These inner type definitions are actually redundant and not needed in go - you can do this instead:

```go
return map[string]interface{}{
		"type":   "host-local",
		"ranges": [][]map[string]string{{{"subnet": "usePodCidr"}}, {{"subnet": "usePodCidrIPv6"}}},
	}
```

Go knows how to infer the type of the list elements from the main `[][]map[string]string`

pkg/controller/installation/validation.go
ti-afra: That's a good idea - I looked at how cni-plugin is setting up the logging and looks like this is a feature that needs to be implemented on cni-plugin as well. Having that in mind, as you said, it makes more sense for this change to be done separately. Should I create a ticket to track this task?

I am thinking we need to define the fileLoggingEnabled flag on both cni-plugin and operator api. on way, is to not overwrite the values of other fields and just ignore them when fileLoggingEnabled flag is ON. the flag could be a shortcut to not setting / changing any loggingconfig values on operator side and not setting up a fileLogger on cni-plugin side.

What do you think?

caseydavenport: I'd raise a backlog item for it, yeah.

file logging enabled in the CNI plugin appears to be configured by the `log_file_path` option, according to my read of this code: https://github.com/projectcalico/calico/blob/1ffe1b75b4c2b24bdd308b9150ede44ed3b555a4/cni-plugin/internal/pkg/utils/utils.go#L739-L770

api/v1/installation_types.go
ti-afra: Done.

pkg/render/node.go
ti-afra: Done.

pkg/render/node.go
ti-afra: Done.

pkg/render/node.go
ti-afra: Done.

pkg/render/node.go
ti-afra: Done.

pkg/controller/installation/validation.go
ti-afra: ticket created here: https://tigera.atlassian.net/browse/EV-3421?atlOrigin=eyJpIjoiOWYwZWI3NmUyZTZiNDMyYWEyMjg1MjA0MjBjZmE2OWMiLCJwIjoiaiJ9

pkg/render/node_test.go
```diff
@@ -993,8 +1014,11 @@ var _ = Describe("Node rendering tests", func() {
       "datastore_type": "kubernetes",
       "mtu": 0,
       "nodename_file_optional": false,
-      "log_level": "Info",
+      "log_level": "Debug",
       "log_file_path": "/var/log/calico/cni/cni.log",
+      "log_file_max_size": 1,
+	  "log_file_max_age": 5,
```
caseydavenport, line 1020: weird indentation on this one

caseydavenport (approved):

pkg/render/node_test.go
ti-afra: ah! looks like this is a tab not a space! I'll fix that.

Issue #2444, opened 2023-02-02
Make etccalico mount path configurable


laibe:
## Current Behavior

Currently the path is hardcoded to `/etc/calico` here: https://github.com/tigera/operator/blob/673ecc4c4d9774dcab03942ea93fb212a26e5cb4/pkg/render/csi.go#L140 

This is a problem for Talos OS that allows for writes only under `/var`.

## Possible Solution

make path configurable, e.g. so we can set it to `/var/calico` 

## Context

I would like to deploy a talos cluster with calico

## Your Environment

Talos 1.3

epleterte:
Hi,
I'd like to see the same! I am also on Talos.

I don't really need the mount path to be configurable, I need the hostPath to be configurable.
I think it's these lines right here (currently starting on line 216):


		{
			Name: "etccalico",
			VolumeSource: corev1.VolumeSource{
				HostPath: &corev1.HostPathVolumeSource{
					Path: filepath.Clean("/etc/calico"),
				},
			},

My current test/workaround has been this:

Edit csi-node-driver DaemonSet and set this annotation:

`unsupported.operator.tigera.io/ignore: "true"`

...to make the operator leave the DS alone and not reset our next change, and then to change hostPath /etc/calico in csi-node-driver DS to something else:

      - hostPath:
          path: /var/lib/calico/etc
          type: ""
        name: etccalico

This works, but I have to leave the DS 'tainted' to keep it running.

uhthomas:
@caseydavenport Apologies for the ping, but I believe this issue blocks anyone using an OS which does not persist `/etc/` from using Tigera. Unfortunately I will have to use Cilium in the meantime.

tmjd:
@laibe do you think you need the MountPath configurable, I think it seems more reasonable that you would need the hostPath configurable like @epleterte suggested?

@epleterte using the unsupported annotation will be an issue when you upgrade as that component won't be upgraded with the annotation.

@uhthomas are you suggesting there are OS's that will discard `/etc` while they are running? I'm asking because I would expect a node that was rebooted to have the CSI driver pod redeployed so no need for `/etc` to be persisted across reboots. (When I hear about something not persisting it is usually in reference to across reboots but maybe you have a unique case.) Also incase you are unaware, the CSI driver I believe is only needed when using the Istio integration, so if you are not using that then you can disable it by setting the `KubeletVolumePluginPath` to `None`.

[...]

PR #2518, opened 2023-03-03
Allow configuration of Calico host mount path

uhthomas:
## Description

Calico persists data to `/etc/calico` on the host but unfortunately some operating systems, like [Talos](https://www.talos.dev/), mount `/etc/` as read-only. This change allows configuration of the host mount path.

Fixes https://github.com/tigera/operator/issues/2444

I am aware there are no tests. I wanted to get a draft PR open to move the issue along. I'd like to understand if:
- This is the right way to make the change.
- If the field name is inline with expectations.
- If the field name is in the correct place.
- What the best way to test this change would be.

Thanks!

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/api/v1/installation_types.go
+++ b/api/v1/installation_types.go
@@ -123,6 +123,11 @@ type InstallationSpec struct {
 	// +optional
 	TyphaMetricsPort *int32 `json:"typhaMetricsPort,omitempty"`
 
+	// DataVolumePath optionally specifies a custom path for Calico to persist data. If not specified, the path will
+	// default to /etc/calico.
+	// +optional
+	DataVolumePath string `json:"dataVolumePath"`
+
 	// FlexVolumePath optionally specifies a custom path for FlexVolume. If not specified, FlexVolume will be
 	// enabled by default. If set to 'None', FlexVolume will be disabled. The default is based on the
 	// kubernetesProvider.
--- a/pkg/controller/utils/merge.go
+++ b/pkg/controller/utils/merge.go
@@ -108,6 +108,11 @@ func OverrideInstallationSpec(cfg, override operatorv1.InstallationSpec) operato
 		inst.TyphaMetricsPort = override.TyphaMetricsPort
 	}
 
+	switch compareFields(inst.DataVolumePath, override.DataVolumePath) {
+	case BOnlySet, Different:
+		inst.DataVolumePath = override.DataVolumePath
+	}
+
 	switch compareFields(inst.FlexVolumePath, override.FlexVolumePath) {
 	case BOnlySet, Different:
 		inst.FlexVolumePath = override.FlexVolumePath
--- a/pkg/crds/operator/operator.tigera.io_installations.yaml
+++ b/pkg/crds/operator/operator.tigera.io_installations.yaml
@@ -5586,6 +5586,11 @@ spec:
                         type: object
                     type: object
                 type: object
+              dataVolumePath:
+                description: DataVolumePath optionally specifies a custom path for
+                  Calico to persist data. If not specified, the path will default
+                  to /etc/calico.
+                type: string
               fipsMode:
                 description: 'FIPSMode uses images and features only that are using
                   FIPS 140-2 validated cryptographic modules and standards. Default:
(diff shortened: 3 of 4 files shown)


mgleung:
Hey @uhthomas , I may not be able to answer all of your questions above about the naming conventions and optimal setting location, but I should be able to help with some testing.

I think the first thing you'll want to do is run `make test` locally to run the UTs. After that, I think the easiest way to see if this works with regards to all of our components is to build your version of the operator `make image` and then deploy your created operator image into a test cluster (using the steps at https://docs.tigera.io/calico/3.25/getting-started/kubernetes/quickstart and modifying the operator image to point to yours would probably be easiest). You should see everything come up without any issue but it might be a good idea to also check that there's no strange error logs in the calico-node instances that boot. Let me know if you need any help with any help validating any of these steps.

epleterte:
@uhthomas I've had a similar patch lying around for a couple of weeks, but wanted to test it first. Had some trouble building initially and haven't had time to pick it up, but just gave building another go with success. Did you get around to do any testing yet?

ErikLundJensen:
Any progress?

agmimidi:
Thanks for the contribution Thomas. 
Any updates on when this will be part of an official release?

uhthomas:
Awesome, thanks for testing this change. Is there anything else which needs to be done for this to be merged?

tmjd:
I think you'll need to rebase your changes on master to pick up updates in #2613 to address one issue in CI.
Also you need to update `pkg/controller/utils/merge.go` to account for the new field. You should be able to see the failure with `go test ./pkg/controller/utils`.

uhthomas:
Thanks @tmjd, I updated `pkg/controller/utils/merge.go`.

As an aside: `make gen-files` does not work on macOS.

```sh
for x in $(find config/crd/bases/*); do sed -i -e '/creationTimestamp: null/d' -e '/^---/d' -e '/^\s*$/d' $x; done
sed: -e: No such file or directory
sed: -e: No such file or directory
sed: -e: No such file or directory
sed: -e: No such file or directory
sed: -e: No such file or directory
sed: -e: No such file or directory
sed: -e: No such file or directory
sed: -e: No such file or directory
sed: -e: No such file or directory
sed: -e: No such file or directory
sed: -e: No such file or directory
sed: -e: No such file or directory
sed: -e: No such file or directory
sed: -e: No such file or directory
sed: -e: No such file or directory
sed: -e: No such file or directory
sed: -e: No such file or directory
make: *** [manifests] Error 1
```

uhthomas:
@tmjd Looks like CI is happy, is this okay to merge?

tmjd:
Sorry no response for a while. I've been trying to chat with people and figure out if this is what we need/want. Initially I was going to ask you to change the field name to be more specific but from looking at the PR I was compelled to ask a few more questions about what the directory is used for. The answer was that it is only for a configuration file that no one probably uses and isn't really the way we should be configuring features with an operator install.
This configuration file idea is a copy of how we handled FlexVol configuration, FlexVol consisted of a binary that was on a host system and so it needed a file based configuration. With CSI, which uses this host mount path, the code is containerized so we can provide configuration to the pod directly instead of needing to use a configuration file.
Hopefully that provides the context for what we should do with this PR.

So with all that info, instead of exposing a new configuration field to set a host path that no one is probably using, we should remove the volume mount so it is not present.

I'm open to hear if this is problematic for anyone that has been a part of this conversation (or anyone not yet part of it too).

uhthomas:
Thanks for following up @tmjd. This sounds like a good thing to do, though I do see there are [169 references](https://github.com/search?q=org%3Atigera+%2Fetc%2Fcalico&type=code) for "/etc/calico" across the tigera organisation. How much work would it be to remove the need for that path altogether?

tmjd:
I don't think we need to worry about any of those other references. None of them are used as a host path volume with an operator install. Many of them are still the location where some files go but it is only inside of a pod (no volume mount) so it doesn't really matter what that path is.

I know calicoctl references that path for configuration but it is a binary that is expected to be a user utility, I'm not sure if that path is configurable (I think it is) but that is outside of the scope of the operator. You could check on calicoctl and make that path configurable if it is not yet.

tmjd:
I'm closing this PR since #2654 was merged instead of this one.

PR #2533, opened 2023-03-08
[1.28] Add missing egress flow to Guardian access policy

pasanw:
- Adds missing egress flow to Guardian access policy.
- The source code of Guardian was consulted to ensure there are no other missing egress flows.
- Fix was validated in a 3.15 MCM cluster by getting cluster into a state where a default-deny in a tier after allow-tigera was causing denies. Engaging the operator with this fix resolves denies.


rene-dekker (approved):
lgtm

PR #2541, opened 2023-03-10
update go-build to v0.82 (golang 1.19.7)

rene-dekker:
- Now using go-build 0.82, which updates golang to v1.19.7
- Now using official controller-gen binary to create CRDs as our own one did not work.


Makefile
```diff
@@ -655,9 +654,9 @@ deploy: manifests kustomize
 # Generate manifests e.g. CRD
 # Can also generate RBAC and webhooks but that is not enabled currently
 manifests:
-	$(DOCKER_RUN) sh -c "\
-		controller-gen $(CRD_OPTIONS) paths="./api/..." output:crd:artifacts:config=config/crd/bases"
-		for x in $$(find config/crd/bases/*); do sed -i -e '/creationTimestamp: null/d' -e '/^---/d' -e '/^\s*$$/d' $$x; done
+	$(DOCKER_RUN) sh -c 'go install sigs.k8s.io/controller-tools/cmd/controller-gen@$(CONTROLLER_GEN_VERSION) && \
```
caseydavenport, line 657: I think we should at least add a comment explaining why we're not using the controller-gen packages in the container.

pkg/common/autoscale.go
```diff
@@ -16,16 +16,22 @@ package common
 // GetTyphaScaleCount will return the number of Typhas needed for the number of nodes.
 //
 // Nodes       Replicas
-//   1-2              1
-//   3-4              2
-//  <200              3
-//  >400              4
-//  >600              5
-//  >800              6
+//
```
caseydavenport, line 19: The spacing looks wrong here now - doesn't line up with the column headers.

pkg/crds/operator/operator.tigera.io_amazoncloudintegrations.yaml
```diff
@@ -187,13 +185,7 @@ spec:
               AmazonCloudIntegration
             type: object
         type: object
-    served: true
+    served: false
```
caseydavenport, line 188: Is this right? Isn't this the served version of the API?

caseydavenport: I am unsure about this change, and not sure why the other CRDs all have this as `true`

rene-dekker: So I figured out what is going on. There is a v1 and a v1beta1 in this manifest. The v1 is served and the v1beta1 is not.

PR #2569, opened 2023-03-22
Prepare for 3.15.2

danudey:


rene-dekker (approved):
lgtm, once you run `make format-check validate-gen-versions fmt test-crds gen-files fix` and check in the changes.

rene-dekker (changes requested):
This is the wrong branch

PR #2572, opened 2023-03-24
Add shards and replicas to Linseed

asincu:
## Description

- Add shards and replicas in Linseed just like we do for FluentD

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/logstorage/linseed/linseed.go
+++ b/pkg/render/logstorage/linseed/linseed.go
@@ -16,7 +16,10 @@ package linseed
 
 import (
 	"fmt"
+	"strconv"
 	"strings"
+
+	relasticsearch "github.com/tigera/operator/pkg/render/common/elasticsearch"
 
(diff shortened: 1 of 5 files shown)


pkg/render/logstorage/linseed/linseed.go
```diff
@@ -308,6 +253,60 @@ func (l *linseed) linseedDeployment() *appsv1.Deployment {
 	}
 }
 
+func (l *linseed) volumeMounts() []corev1.VolumeMount {
```
caseydavenport, line 256: I'd rather we didn't split these out into separate functions - these are only used once, so there's no need to make a common function for it. The code is easier to read if all of this lives within `linseedDeployment`.

Same goes for `envVars` and `container`

asincu: Sounds good to me!

caseydavenport (approved):

PR #2576, opened 2023-03-28
Fix periodic reconcile logic

pasanw:
## Description
The current periodic reconcile implementation requeues each successful reconcile every 5 minutes. This causes each event from the initial set of events (from source.Kind watches) to be replayed every 5 minutes.

This PR
* Triggers a reconcile every 5 minutes by injecting an event via source.Channel.
* Re-enables the Info level logging of Reconcile invocations of the core controller. This helped masked the issue previously.

## For PR author
- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/installation/core_controller.go
+++ b/pkg/controller/installation/core_controller.go
@@ -91,6 +91,7 @@ const (
 	// This is separate from the calico/node prometheus metrics port, which is user configurable.
 	defaultNodeReporterPort = 9081
 	CalicoFinalizer         = "tigera.io/operator-cleanup"
+	reconcilePeriod         = 5 * time.Minute
 )
 
 const InstallationName string = "calico"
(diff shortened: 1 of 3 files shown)


rene-dekker (approved):
lgtm

tmjd (approved):
Nice. LGTM

pkg/controller/installation/core_controller.go
```diff
@@ -793,7 +801,7 @@ func mergeProvider(cr *operator.Installation, provider operator.Provider) error
 // Result.Requeue is true, otherwise upon completion it will remove the work from the queue.
 func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
 	reqLogger := log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
-	reqLogger.V(1).Info("Reconciling Installation.operator.tigera.io")
+	reqLogger.Info("Reconciling Installation.operator.tigera.io")
```
caseydavenport, line 804: You're sure this isn't too spammy now? :grin: 

In general, we should be aiming for no repetitive log output. In an ideal world, the INFO level logs would only show when something meaningful has happened.

pasanw: :grin: 

From my tests, the log behaviour is much more sane now. At steady state, I only see the periodic reconcile invocations. I figured I would re-enable this since I believe it's the only controller that doesn't log when it reconciles, which can lead to misleading log interpretations.

PR #2582, opened 2023-04-05
Add patch permissions to tigera-network-admin for services

LorcanMcVeigh:
## Description
Fix the change i made a few weeks ago, that change doesn't work.
There was an error in the testing I did on it which is why im only catching it now. 

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/crds/calico/crd.projectcalico.org_felixconfigurations.yaml
+++ b/pkg/crds/calico/crd.projectcalico.org_felixconfigurations.yaml
@@ -89,7 +89,7 @@ spec:
                 description: 'BPFEnforceRPF enforce strict RPF on all host interfaces
                   with BPF programs regardless of what is the per-interfaces or global
                   setting. Possible values are Disabled, Strict or Loose. [Default:
-                  Strict]'
+                  Loose]'
                 type: string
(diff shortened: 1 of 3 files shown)


tmjd (approved):
LGTM

electricjesus (approved):

caseydavenport:
@LorcanMcVeigh could you share what changed here? Why was the previous PR not good enough?

LorcanMcVeigh:
> @LorcanMcVeigh could you share what changed here? Why was the previous PR not good enough?


yeah no problem. @caseydavenport 

I was building on the work i did a couple of weeks ago on the change to the operator where I add the `services/status` resource to the `tigera-network-admin` clusterrole.
That change doesn't work. `tigera-network-admin` needs to have patch permissions on services resource not services/status . I made a mistake when i was testing that change. I forgot to remove the patch permission from  services in the clusterrole which is why it passed. Because the annotation resides under the `metadate` field it doesn't make sense to have permissions to the `status` field. 
However, this change is only for v3.17 . There is work planned for v3.18 which should allow us to revert this change.

PR #2584, opened 2023-04-12
Support Egress Gateway Policy resource

mazdakn:
## Description
Related to adding destination based routing feature to EGWs: https://github.com/tigera/calico-private/pull/5836

Changes:
- Add the new Egress Gateway Policy CRD.
- Adding Egress Gateway Policy RBAC to Typha.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


pkg/render/typha.go
```diff
@@ -256,6 +256,7 @@ func (c *typhaComponent) typhaRole() *rbacv1.ClusterRole {
 					"ipreservations",
 					"networkpolicies",
 					"networksets",
+					"egressgatewaypolicies",
```
caseydavenport, line 259: This doesn't belong here - it needs to go below in the Calico Enterprise only section.

caseydavenport (changes requested):
One comment

pkg/render/typha.go
mazdakn: I cannot find EE only section in this file. At the same time, it seems `ExternalNetwork` resource which is also EE only is listed here.

caseydavenport: This is where it should be: https://github.com/tigera/operator/blob/cfa369163949ec3281c5240fc0bea55d45e3bee4/pkg/render/typha.go#L314-L341

ExternalNetworks is a bug and also shouldn't be here - could you move that as well?

mazdakn: sure, will move it.

mazdakn: Fixed.

caseydavenport (approved):

PR #2611, opened 2023-04-28
Update versions for v3.26.0

danudey:
Update versions files for operator v1.30:

* Calico v3.26.0
* Calient v3.16.1 (but we don't want people to use this operator for enterprise)


tmjd (approved):
LGTM, please make sure you squash when (or before) you merge this.

tmjd:
I've now kicked the CI on this a 2nd time and it seemed to fail with the same failures as before. So lets see if this 3rd run passes.

PR #2612, opened 2023-05-01
Add policy to support nodelocal dnscache

ti-afra:
## Description

What has been added:
- Adding a global policy to enable traffic from tigera / calico to nodelocal-dnscache.
- Adding watch for the new global policy to only be added when nodelocal-dnscache is deployed to the cluster and remove it when it's not deployed anymore.

How:
 global policy called node-local-dns is added to the allow-tigera tier when nodelocal-dnscache is deployed to the cluster. This global policy needs to be added otherwise the connectivity between calico components will be broken after nodelocal-dnscache deployment. Thus, it is added to the "allow-tigera" tier.
 The allow-tigera.node-local-dns global policy is allowing egress traffic from all tigera components to the kube-dns ip address (where nodelocal-dnscache) is also broadcasting as it's own ip to serve all traffic heading to "kube-dns".

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


pkg/render/tiers/tiers.go
```diff
@@ -40,9 +43,13 @@ var DNSIngressNamespaceSelector = createDNSIngressNamespaceSelector(
 	render.IntrusionDetectionNamespace,
 	render.KibanaNamespace,
 	render.ManagerNamespace,
+	render.ECKOperatorNamespace,
+	render.PacketCaptureNamespace,
+	render.PolicyRecommendationNamespace,
 	common.TigeraPrometheusNamespace,
-	"tigera-skraper",
+	dpi.DeepPacketInspectionNamespace,
```
matthewdupre, line 50: Was this correctly allowed before somewhere else, or was this not being here a bug you're fixing along the way?

pkg/render/tiers/tiers.go
```diff
@@ -121,7 +138,7 @@ func (t tiersComponent) allowTigeraClusterDNSPolicy() *v3.NetworkPolicy {
 					Action: v3.Allow,
 					Source: v3.EntityRule{
 						NamespaceSelector: "all()",
-						Selector:          DNSIngressNamespaceSelector,
+						Selector:          TigeraNamespaceSelector,
```
matthewdupre, line 141: I think it's strictly not needed, but I think it's OK for this policy to still exist when running with nodelocaldnscache

pkg/render/tiers/tiers.go
```diff
@@ -138,7 +155,34 @@ func (t tiersComponent) allowTigeraClusterDNSPolicy() *v3.NetworkPolicy {
...
+		},
+		Spec: v3.GlobalNetworkPolicySpec{
+			Order:    &networkpolicy.AfterHighPrecendenceOrder,
+			Tier:     networkpolicy.TigeraComponentTierName,
+			Selector: nodeLocalDNSPolicySelector,
+			Egress: []v3.Rule{
+				{
+					Action: v3.Allow,
+					Destination: v3.EntityRule{
+						Nets:  []string{t.cfg.KubeDNSCIDR},
```
matthewdupre, line 174: This is a very unusual thing to include in a policy (a service IP): I'd like to see a comment saying why it works.

Something like:

"nodeLocalDNSCache creates and listens on the kube-dns ClusterIP on each node, so we can use that IP address directly in the policy where a normal service IP wouldn't match"

matthewdupre (commented):
I'm not an expert in the intended structure of the operator so I haven't reviewed from that perspective.

pkg/controller/tiers/tiers_controller.go
```diff
@@ -98,18 +117,11 @@ func add(mgr manager.Manager, c controller.Controller) error {
 		return fmt.Errorf("tiers-controller failed to watch APIServer resource: %v", err)
 	}
 
-	return nil
-}
-
-var _ reconcile.Reconciler = &ReconcileTiers{}
+	if err := utils.AddNodeLocalDNSWatch(c); err != nil {
+		return fmt.Errorf("tiers-controller failes to watch node-local-dns daemonset: #{err}")
```
pasanw, line 121: FYI: typo (failes -> failed) and `err` is not provided to the Errorf

pasanw, line 121: FYI: typo (failes -> failed) and `err` is not provided to the Errorf

pkg/controller/tiers/tiers_controller.go
```diff
@@ -136,7 +148,37 @@ func (r *ReconcileTiers) Reconcile(ctx context.Context, request reconcile.Reques
...
+	tiersConfig := tiers.Config{
+		Openshift:    r.provider == operatorv1.ProviderOpenShift,
+		NodeLocalDNS: false,
+		KubeDNSCIDR:  "0.0.0.0/32",
+	}
+
+	if r.provider != operatorv1.ProviderOpenShift {
+		nodeLocalDNSExists, _ := utils.IsNodeLocalDNSAvailable(ctx, r.client)
+		if err != nil {
+			r.status.SetDegraded(operatorv1.ResourceNotFound, "Node Local DNS not enabled", err, reqLogger)
```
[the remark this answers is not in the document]
pasanw: Two notes:
* We need to differentiate between when we fail to determine node local DNS cache presence (I would expect this to be when `err != nil`) and when we determine that node local DNS cache is not present (I would expect this to be when the function returns `false, nil`)
* We should return from the reconcile loop when we fail to determine node local DNS cache presence

pkg/controller/tiers/tiers_controller.go
```diff
@@ -136,7 +148,37 @@ func (r *ReconcileTiers) Reconcile(ctx context.Context, request reconcile.Reques
...
+		} else if nodeLocalDNSExists {
+			// discover kube-dns ip address
+			k8sClient, _ := kubernetes.NewForConfig(r.config)
+			kubeDNSService, _ := k8sClient.CoreV1().Services("kube-system").Get(ctx, "kube-dns", metav1.GetOptions{})
```
pasanw, line 164: Is it possible to query Services using the client provided by the controller runtime (i.e. the Manager client)? 

Motivation being: this Get will not be cached, and will hit the API server every reconcile. The manager client maintains caches for resources

pkg/controller/tiers/tiers_controller.go
```diff
@@ -136,7 +148,37 @@ func (r *ReconcileTiers) Reconcile(ctx context.Context, request reconcile.Reques
...
+			if err != nil {
+				r.status.SetDegraded(operatorv1.ResourceNotFound, "kube-dns service not found", err, reqLogger)
```
pasanw, line 166: See the comment I left in tiers_controller.go, the same two notes apply here

pkg/controller/utils/utils.go
```diff
@@ -615,3 +615,27 @@ func GetElasticsearch(ctx context.Context, c client.Client) (*esv1.Elasticsearch
...
+func AddNodeLocalDNSWatch(c controller.Controller) error {
+	ds := &v1.DaemonSet{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: "kube-system",
+			Labels: map[string]string{
+				"k8s-app": "node-local-dns",
+			},
+		},
+	}
+	return c.Watch(&source.Kind{Type: ds}, &handler.EnqueueRequestForObject{}, createPredicateForObject(ds))
```
pasanw, line 640: I think the only thing that `source.Kind` cares about is the underlying object kind, I don't think it references the namespace or labels, i.e. I think the first argument should be `&v1.Daemonset{}`

Also, when passing this to `createPredicateForObject`, the predicate is constructed based on name and namespace. I believe since name is omitted, this will match all Daemonsets in kube-system

pkg/controller/utils/utils.go
```diff
@@ -615,3 +615,27 @@ func GetElasticsearch(ctx context.Context, c client.Client) (*esv1.Elasticsearch
 	}
 	return &es, nil
 }
+
+func IsNodeLocalDNSAvailable(ctx context.Context, cli client.Client) (bool, error) {
+	ds := &v1.DaemonSet{}
+
+	err := cli.Get(ctx, client.ObjectKey{Namespace: "kube-system", Name: "node-local-dns"}, ds)
+	if err != nil {
+		return false, err
```
pasanw, line 624: We should check the error type here and determine if it is 'not found'. If so, we can return `false, nil` - else return `false, err`

main.go
```diff
@@ -230,7 +230,8 @@ func main() {
 		// tiers that the user is authorized for.
 		NewCache: cache.BuilderWithOptions(cache.Options{
 			SelectorsByObject: cache.SelectorsByObject{
-				&v3.NetworkPolicy{}: {Label: policySelector},
+				&v3.NetworkPolicy{}:       {Label: policySelector},
+				&v3.GlobalNetworkPolicy{}: {Label: policySelector},
```
pasanw, line 234: Nice catch!

pkg/controller/tiers/tiers_controller.go
```diff
@@ -136,7 +148,37 @@ func (r *ReconcileTiers) Reconcile(ctx context.Context, request reconcile.Reques
...
+			}
+			kubeDNSIpAddress := kubeDNSService.Spec.ClusterIP
+
+			var builder strings.Builder
+			builder.WriteString(kubeDNSIpAddress)
+			builder.WriteString("/32")
```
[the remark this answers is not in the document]
ti-afra: we should be using 128 for ipv6, is that right?

pkg/render/tiers/tiers.go
ti-afra: sure - here we are just refactoring the selector to be used by both policies.
maybe later on when we have more confident in cluster running with nodelocaldnscache we can refactor a bit more here.

pkg/controller/tiers/tiers_controller.go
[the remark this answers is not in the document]
ti-afra: should we add "/128" for ipv6?

pkg/render/tiers/tiers.go
ti-afra: sure - we just renamed the var to be used in both policy. Maybe later on when we have more confident in everything working with nodelocal dnscache, we can also delete this policy as part of watching for nodelocal dnscache in the cluster?

pkg/render/tiers/tiers.go
ti-afra: I dont think we have deleted anything - just adding and also re-ordered. "tigera-skraper" moved to the end of the list.

matthewdupre: What I mean is - we've added dpi.DeepPacketInspectionNamespace to this policy.  That means one of a few things that I can think of:
1. The DPI namespace was previously not allowed to make DNS requests and that was fine because it never did (possibly even it wasn't used).
2. The DPI namespace was previously not allowed to make DNS requests by this policy, but there was some other policy somewhere else that allowed it.
3. The DPI namespace was previously not allowed to make DNS requests, and we never noticed that we'd broken it.

Basically, what I want to understand is:
- Why did you choose to add this line, and how did you discover that you needed to?
- What does that mean for DPI previously?  Was it broken?

ti-afra: probably needs more investigation on dpi to be able to answer these questions.
based on the conversation we had, since dpi namespace is created by the operator, we include in the list to avoid any possible problems because of the denied access to dns / dns-cache.

pkg/controller/tiers/tiers_controller.go
```diff
@@ -136,7 +145,48 @@ func (r *ReconcileTiers) Reconcile(ctx context.Context, request reconcile.Reques
...
+		KubeDNSCIDR:  "0.0.0.0/32",
+	}
+
+	if r.provider != operatorv1.ProviderOpenShift {
+		nodeLocalDNSExists, err := utils.IsNodeLocalDNSAvailable(ctx, r.client)
+		if err != nil {
+			r.status.SetDegraded(operatorv1.ResourceReadError, "Error querying node-local-dns pods", err, reqLogger)
+			return reconcile.Result{RequeueAfter: 10 * time.Second}, nil
+		} else if nodeLocalDNSExists {
+			// discover kube-dns ip address - node-local-dns is not supported on OpenShift which is the only platform without
```
caseydavenport, line 160: I would tweak this to say "Discover the kube-dns Service cluster IP address"

pkg/controller/tiers/tiers_controller.go
```diff
@@ -136,7 +145,48 @@ func (r *ReconcileTiers) Reconcile(ctx context.Context, request reconcile.Reques
...
+			err = r.client.Get(ctx, types.NamespacedName{Name: "kube-dns", Namespace: "kube-system"}, kubeDNSService)
+			if err != nil {
+				if errors.IsNotFound(err) {
+					r.status.SetDegraded(operatorv1.ResourceNotFound, "kube-dns service not found", err, reqLogger)
+				} else {
+					r.status.SetDegraded(operatorv1.ResourceReadError, "Error querying kube-dns service", err, reqLogger)
+				}
+				return reconcile.Result{RequeueAfter: 10 * time.Second}, nil
+			}
+			kubeDNSIpAddress := kubeDNSService.Spec.ClusterIP
```
caseydavenport, line 172: Looks like you're handling IPv6 already, which is great. However, we also need to handle dual-stack services here as well. Instead of looking at `Spec.ClusterIP`, we will need to look at `Spec.ClusterIPs` which can contain up to two IP addresses.

caseydavenport: ref: https://kubernetes.io/docs/concepts/services-networking/dual-stack/

pkg/controller/utils/utils.go
```diff
@@ -25,8 +25,8 @@ import (
 	esv1 "github.com/elastic/cloud-on-k8s/v2/pkg/apis/elasticsearch/v1"
 
 	"github.com/go-logr/logr"
+	v1 "k8s.io/api/apps/v1"
```
caseydavenport, line 28: let's call this "appsv1" to be consistent.

pkg/controller/tiers/tiers_controller_test.go
```diff
@@ -20,13 +20,13 @@ import (
...
 	. "github.com/onsi/gomega"
 	"github.com/stretchr/testify/mock"
-
 	v3 "github.com/tigera/api/pkg/apis/projectcalico/v3"
 	operatorv1 "github.com/tigera/operator/api/v1"
 	"github.com/tigera/operator/pkg/apis"
 	"github.com/tigera/operator/pkg/common"
 	"github.com/tigera/operator/pkg/controller/status"
 	"github.com/tigera/operator/pkg/controller/utils"
+	v1 "k8s.io/api/apps/v1"
```
caseydavenport, line 29: same here - "appsv1"

pkg/render/tiers/tiers.go
caseydavenport: Sounds like we should omit dpi.DeepPacketInspectionNamespace for now since it previously didn't have any rules to allow DNS, so either:

- It didn't need DNS, and still doesn't.
- It does need DNS, but we weren't allowing it. This is a bug and we should fix in a separate PR.

pkg/render/tiers/tiers.go
```diff
@@ -52,7 +59,9 @@ func Tiers(cfg *Config) render.Component {
 }
 
 type Config struct {
-	Openshift bool
+	Openshift    bool
+	NodeLocalDNS bool
```
caseydavenport, line 63: To simplify this a little bit, I don't think we need both `NodeLocalDNS` as well as `KubeDNSCIDR`, because right now they are bound to each other. You'll never have NodeLocalDNS set to `false` and `KubeDNSCIDR` set to an IP, and vice-versa. 

We should be able to do something like this:

```go
type Config struct {
  Openshift bool
  DNSEgressIP string
}
```

and then 

```go

if DNSEgressIP != "" {
  // Add the policy for DNS egress
} else {
  // Remove the policy for DNS egress
}
```

This just simplifies the logic a tiny bit and removes the ability to specify "invalid" combinations.

caseydavenport (changes requested):

pkg/controller/tiers/tiers_controller_test.go
caseydavenport: @ti-afra I think you missed renaming this one :)

pkg/render/tiers/tiers.go
```diff
@@ -69,7 +74,15 @@ func (t tiersComponent) Objects() ([]client.Object, []client.Object) {
 		t.allowTigeraClusterDNSPolicy(),
 	}
 
-	return objsToCreate, nil
+	objsToDelete := []client.Object{}
+
+	if t.cfg.DNSEgressCIDRs != nil && len(t.cfg.DNSEgressCIDRs) > 0 {
```
caseydavenport, line 79: In Golang, for length checks you don't actually need to do a `nil` check first - `len` will return 0 if the slice is nil, so you can simplify this if statement.

https://go.dev/play/p/6bAfsebsrLB

pkg/render/tiers/tiers_test.go
```diff
@@ -30,10 +30,15 @@ var _ = Describe("Tiers rendering tests", func() {
...
 	clusterDNSPolicyForOCP := testutils.GetExpectedPolicyFromFile("../testutils/expected_policies/dns_ocp.json")
+	nodeLocalDNSPolicy := testutils.GetExpectedGlobalPolicyFromFile("../testutils/expected_policies/node_local_dns.json")
 
 	BeforeEach(func() {
 		// Establish default config for test cases to override.
-		cfg = &tiers.Config{Openshift: false}
+		cfg = &tiers.Config{
+			Openshift:    false,
+			NodeLocalDNS: false,
+			KubeDNSCIDR:  "10.96.0.10/32",
```
caseydavenport, line 40: I think these tests will need an update to match the new `tiers.Config` struct.

I'd also suggest adding a dual-stack, as well as an IPv6 single-stack test case here as well.

pkg/controller/tiers/tiers_controller_test.go
ti-afra: oops!

pkg/controller/tiers/tiers_controller.go
```diff
@@ -147,3 +163,49 @@ func (r *ReconcileTiers) Reconcile(ctx context.Context, request reconcile.Reques
 
 	return reconcile.Result{}, nil
 }
+
+func (r *ReconcileTiers) prepareTiersConfig(ctx context.Context, reqLogger logr.Logger) (*tiers.Config /*requeue*/, bool) {
```
caseydavenport, line 167: nit, but it's not very common to inline comments like this in the return values. Instead, I'd recommend using a function-level comment that explains it, like this:

```
// prepareTiersConfig returns a tiers.Config based on cluster state, and a boolean indicating whether or not the controller should requeue the reconcile.
```

It's also possible to use named return values, but I generally advise against those since it's not an especially common pattern. https://go.dev/tour/basics/7

caseydavenport: In this case though, it seems like we always mark as degraded and schedule a requeue in the same branch. You could just return an error instead of a boolean, which would be a more normal way to propagate errors up to the caller.

The caller would then set degraded on the error, and return it, triggering a requeue anyway.

pkg/render/tiers/tiers_test.go
```diff
@@ -30,10 +30,40 @@ var _ = Describe("Tiers rendering tests", func() {
 
 	clusterDNSPolicy := testutils.GetExpectedPolicyFromFile("../testutils/expected_policies/dns.json")
 	clusterDNSPolicyForOCP := testutils.GetExpectedPolicyFromFile("../testutils/expected_policies/dns_ocp.json")
+	nodeLocalDNSPolicy_ipv4 := testutils.GetExpectedGlobalPolicyFromFile("../testutils/expected_policies/node_local_dns_ipv4.json")
```
caseydavenport, line 33: nit, but `_` isn't commonly used in variable names in Golang - these should be `nodeLocalDNSPolicyIPv4`, etc.

pkg/controller/tiers/tiers_controller.go
ti-afra: I guess in this case if we return err, we might also need to return the msg we are passing to SetDegraded as well or we have to decide to not care about explaining the details of what happened. is that right? or there is other ways to inject the msg in the error obj?

pkg/render/testutils/policy.go
```diff
@@ -35,6 +35,12 @@ type AllowTigeraScenario struct {
 	Openshift      bool
 }
 
+type IPMode string
+
+const IPV4 IPMode = "IPV4"
```
caseydavenport, line 40: another nit, these should be grouped together like so:

```go
const (
  IPV4 IPMode = "IPV4"
  IPV6 IPMode = "IPV6"
)
```
)

pkg/render/tiers/tiers.go
```diff
@@ -138,7 +156,56 @@ func (t tiersComponent) allowTigeraClusterDNSPolicy() *v3.NetworkPolicy {
...
+		IPV6Rule := egressRuleTemplate
+		appendIPVersionCIDRs(&IPV6Rule.Destination.Nets, t.cfg.DNSEgressCIDRs.IPV6)
+
+		nodeLocalDNSPolicy.Spec.Egress = append(nodeLocalDNSPolicy.Spec.Egress, IPV6Rule)
+	}
+
+	return nodeLocalDNSPolicy
+}
+
+func appendIPVersionCIDRs(net *[]string, cidrArray []string) {
```
caseydavenport, line 204: I don't think you need this function - you should just be able to do this inline, right?

```go
IPV4Rule.Destination.Nets = append(IPV4.Destination.Nets, t.cfg.DNSEgressCIDRs.IPv4)
```

caseydavenport (approved):
LGTM!

PR #2615, opened 2023-05-01
Check more indicators for EKS discovery

tmjd:
## Description

We have found in testing that the provider is not being detected as EKS in setups that were detected as EKS before. Upon looking at our detection/discovery code and looking at some EKS clusters, there was and still isn't a great means of detecting an EKS cluster. This PR expands the detection a bit to include a couple more indicators.

## For PR author

- [X] Tests for change.
  - I've also manually tested the discovery in a couple EKS clusters

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/utils/discovery.go
+++ b/pkg/controller/utils/discovery.go
@@ -167,15 +167,40 @@ func isDockerEE(ctx context.Context, c kubernetes.Interface) (bool, error) {
 // EKS doesn't have any provider-specific API groups, so we need to use a different approach than
 // we use for other platforms in autodetectFromGroup.
 func isEKS(ctx context.Context, c kubernetes.Interface) (bool, error) {
-	cm, err := c.CoreV1().ConfigMaps("kube-system").Get(ctx, "eks-certificates-controller", metav1.GetOptions{})
-	if err != nil {
-		if kerrors.IsNotFound(err) {
-			return false, nil
-		}
+	// This looks for a configmap that is used in EKS clusters for enabling access to EKS clusters
+	// https://docs.aws.amazon.com/eks/latest/userguide/add-user-role.html
+	_, err := c.CoreV1().ConfigMaps("kube-system").Get(ctx, "aws-auth", metav1.GetOptions{})
+	if err == nil {
+		return true, nil
+	} else if !kerrors.IsNotFound(err) {
 		return false, err
 	}
 
-	return (cm != nil), nil
+	// This is a config map that that use to be present in EKS cluster but now seems to be deprecated.
+	// We'll keep this detection so we ensure we detect EKS in older clusters that have this.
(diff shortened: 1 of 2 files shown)


pkg/controller/utils/discovery.go
```diff
@@ -167,15 +167,34 @@ func isDockerEE(ctx context.Context, c kubernetes.Interface) (bool, error) {
...
-	cm, err := c.CoreV1().ConfigMaps("kube-system").Get(ctx, "eks-certificates-controller", metav1.GetOptions{})
-	if err != nil {
-		if kerrors.IsNotFound(err) {
-			return false, nil
-		}
+	cm, err := c.CoreV1().ConfigMaps("kube-system").Get(ctx, "aws-auth", metav1.GetOptions{})
+	if err != nil && !kerrors.IsNotFound(err) {
 		return false, err
 	}
+	if cm != nil {
```
ozdanborne, line 174: I see you're just copying this logic from the existing code so that's fine but it looks odd to me because presence of this configmap is indicated by err being IsNotFound. I don't think cm being nil is behavior that's meant to be relied on for anything.

Fine to merge as is or take a crack at refactoring, either is ok to me

ozdanborne (approved):
1 optional suggestion but lgtm

ozdanborne (approved):

pkg/controller/utils/discovery.go
```diff
@@ -167,15 +167,34 @@ func isDockerEE(ctx context.Context, c kubernetes.Interface) (bool, error) {
 // EKS doesn't have any provider-specific API groups, so we need to use a different approach than
 // we use for other platforms in autodetectFromGroup.
 func isEKS(ctx context.Context, c kubernetes.Interface) (bool, error) {
-	cm, err := c.CoreV1().ConfigMaps("kube-system").Get(ctx, "eks-certificates-controller", metav1.GetOptions{})
-	if err != nil {
-		if kerrors.IsNotFound(err) {
-			return false, nil
+	_, err := c.CoreV1().ConfigMaps("kube-system").Get(ctx, "aws-auth", metav1.GetOptions{})
```
caseydavenport, line 170: Would be nice for future readers to add some comments to this code explaining what we're checking. I'm not sure what "aws-auth" is, so it's hard to review whether or not this is an appropriate way to determine if we're running on EKS (e.g., as opposed to other AWS based Kubernetes providers).

caseydavenport (approved):
One comment - basically, I think we should document what this code is trying to do with comments a bit better. Maybe with a link to details on what "aws-auth", "eks-certificates-controller" are.

tmjd:
I've put some comments in. Unfortunately eks-certificates-controller and the kube-dns label are not something I found any specific documentation about them.

PR #2630, opened 2023-05-08
add calico node cluster role

Tamas-Biro1:
## Description
It seems the `calico-node` ClusterRole with `serviceaccounts/token` is missing from `nodeRole` function causing the following error when migrating from unmanaged manifest based calico install to Tigera Operator: `2023-05-06 10:31:46.151 [FATAL][1] cni-installer/<nil> <nil>: Unable to create token for CNI kubeconfig error=serviceaccounts "calico-node" is forbidden: User "system:serviceaccount:kube-system:calico-node" cannot create resource "serviceaccounts/token" in API group "" in the namespace "kube-system"`

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/node.go
+++ b/pkg/render/node.go
@@ -393,10 +393,10 @@ func (c *nodeComponent) nodeRole() *rbacv1.ClusterRole {
 				Verbs:     []string{"patch"},
 			},
 			{
-				// Used for creating service account tokens to be used by the CNI plugin.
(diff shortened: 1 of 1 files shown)


caseydavenport (changes requested):
This is actually as intended - we don't want calico-node to have permissions to propagate its own token.

PR #2639, opened 2023-05-15
Add dynamic layers to tigera infra layer in SG EV-3506

vara2504:
## Description

Issue: https://tigera.atlassian.net/browse/EV-3506

Changes:

Service graph Layer should display any all the namespace created by the product.

Added a changes to dynamically populate namespace that applicable for  cluster type
Standalone cluster/Management cluster:
- Additional layers : tigera-manager, tigera-dex, tigera-amazon-cloud-integration based on CR.

Managed cluster: 
 - Additional layers :  tigera-guardian, tigera-amazon-cloud-integration based on CR.

Testing:

Management cluster: with dex enabled
image

Managed cluster:
image

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/clusterconnection/clusterconnection_controller.go
+++ b/pkg/controller/clusterconnection/clusterconnection_controller.go
@@ -331,16 +331,28 @@ func (r *ReconcileConnection) Reconcile(ctx context.Context, request reconcile.R
 		}
 	}
 
+	// Populate a list of namespaces to be displayed in the service graph Tigera infrastructure layer.
+	sgLayerTigeraNameSpaces := render.DefaultSGLayerTigeraNamespaces()
+	sgLayerTigeraNameSpaces[render.GuardianNamespace] = true
+
+	amz, err := utils.GetAmazonCloudIntegration(ctx, r.Client)
+	if err != nil && !k8serrors.IsNotFound(err) {
+		r.status.SetDegraded(operatorv1.ResourceReadError, "Failed to get the GetAmazonCloudIntegration configuration", err, reqLogger)
+	} else if amz != nil {
+		sgLayerTigeraNameSpaces[render.AmazonCloudIntegrationNamespace] = true
+	}
+
 	ch := utils.NewComponentHandler(log, r.Client, r.Scheme, managementClusterConnection)
 	guardianCfg := &render.GuardianConfiguration{
-		URL:               managementClusterConnection.Spec.ManagementClusterAddr,
-		TunnelCAType:      managementClusterConnection.Spec.TLS.CA,
-		PullSecrets:       pullSecrets,
-		Openshift:         r.Provider == operatorv1.ProviderOpenShift,
-		Installation:      instl,
-		TunnelSecret:      tunnelSecret,
-		TrustedCertBundle: trustedCertBundle,
-		UsePSP:            r.usePSP,
+		URL:                     managementClusterConnection.Spec.ManagementClusterAddr,
+		TunnelCAType:            managementClusterConnection.Spec.TLS.CA,
+		PullSecrets:             pullSecrets,
+		Openshift:               r.Provider == operatorv1.ProviderOpenShift,
+		Installation:            instl,
+		TunnelSecret:            tunnelSecret,
+		TrustedCertBundle:       trustedCertBundle,
+		UsePSP:                  r.usePSP,
+		SGLayerTigeraNameSpaces: sgLayerTigeraNameSpaces,
 	}
 
 	components := []render.Component{render.Guardian(guardianCfg)}
--- a/pkg/controller/clusterconnection/clusterconnection_controller.go
+++ b/pkg/controller/clusterconnection/clusterconnection_controller.go
@@ -338,6 +338,7 @@ func (r *ReconcileConnection) Reconcile(ctx context.Context, request reconcile.R
 	amz, err := utils.GetAmazonCloudIntegration(ctx, r.Client)
 	if err != nil && !k8serrors.IsNotFound(err) {
 		r.status.SetDegraded(operatorv1.ResourceReadError, "Failed to get the GetAmazonCloudIntegration configuration", err, reqLogger)
+		return reconcile.Result{}, nil
 	} else if amz != nil {
 		sgLayerTigeraNameSpaces[render.AmazonCloudIntegrationNamespace] = true
 	}
--- a/pkg/controller/manager/manager_controller.go
+++ b/pkg/controller/manager/manager_controller.go
@@ -529,6 +529,7 @@ func (r *ReconcileManager) Reconcile(ctx context.Context, request reconcile.Requ
 	amz, err := utils.GetAmazonCloudIntegration(ctx, r.client)
 	if err != nil && !errors.IsNotFound(err) {
 		r.status.SetDegraded(operatorv1.ResourceReadError, "Failed to get the GetAmazonCloudIntegration configuration", err, reqLogger)
+		return reconcile.Result{}, nil
 	} else if amz != nil {
 		sgLayerTigeraNameSpaces[render.AmazonCloudIntegrationNamespace] = true
 	}
--- a/pkg/controller/manager/manager_controller.go
+++ b/pkg/controller/manager/manager_controller.go
@@ -359,6 +359,10 @@ func (r *ReconcileManager) Reconcile(ctx context.Context, request reconcile.Requ
 		trustedSecretNames = append(trustedSecretNames, render.ComplianceServerCertSecret)
 	}
 
+	// Populate a list of namespaces to be displayed in the service graph Tigera infrastructure layer.
+	sgLayerTigeraNameSpaces := render.DefaultSGLayerTigeraNamespaces()
+	sgLayerTigeraNameSpaces[render.ManagerNamespace] = true
+
 	// Fetch the Authentication spec. If present, we use to configure user authentication.
 	authenticationCR, err := utils.GetAuthentication(ctx, r.client)
 	if err != nil && !errors.IsNotFound(err) {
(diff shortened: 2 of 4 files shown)


pkg/controller/manager/manager_controller.go
```diff
@@ -521,6 +523,13 @@ func (r *ReconcileManager) Reconcile(ctx context.Context, request reconcile.Requ
 		replicas = &mcmReplicas
 	}
 
+	amz, err := utils.GetAmazonCloudIntegration(ctx, r.client)
+	if err != nil {
+		log.Error(err, "Failed to  fetch GetAmazonCloudIntegration info")
+	} else if amz != nil {
+		additionalNameSpaces = append(additionalNameSpaces, render.DexNamespace)
```
rene-dekker, line 530: I think this should be "tigera-amazon-cloud-integration"

pkg/controller/manager/manager_controller.go
```diff
@@ -359,6 +359,7 @@ func (r *ReconcileManager) Reconcile(ctx context.Context, request reconcile.Requ
 		trustedSecretNames = append(trustedSecretNames, render.ComplianceServerCertSecret)
 	}
 
+	var additionalNameSpaces []string
```
rene-dekker, line 362: nit: Maybe additional uiLayerNamespaces would be more indicative of how this variable is used.

pkg/controller/clusterconnection/clusterconnection_controller.go
```diff
@@ -331,16 +331,25 @@ func (r *ReconcileConnection) Reconcile(ctx context.Context, request reconcile.R
 		}
 	}
 
+	optionaUILayerNamespaces := []string{render.GuardianNamespace}
```
caseydavenport, line 334: I think this needs a comment explaining why these are optional and what a UI layer is at all.

pkg/render/manager.go
```diff
@@ -942,25 +944,27 @@ func managerUserSpecificSettingsGroup() *v3.UISettingsGroup {
 // all of the tigera namespaces.
 //
 // Calico Enterprise only
-func managerClusterWideTigeraLayer() *v3.UISettings {
+func managerClusterWideTigeraLayer(namespace []string) *v3.UISettings {
```
caseydavenport, line 947: nit: "namespace" is singular, but it's a slice of multiple elements.

This should be called something like `additionalNamespaces` or something like that. 

Alternatively, remove `namespaces` from this function altogether and callers just need to pass the full list of namespaces they need.

pkg/render/manager.go
```diff
@@ -942,25 +944,27 @@ func managerUserSpecificSettingsGroup() *v3.UISettingsGroup {
...
-		"tigera-manager",
 		"tigera-operator",
 		"tigera-packetcapture",
 		"tigera-policy-recommendation",
 		"tigera-prometheus",
 		"tigera-system",
 		"calico-system",
 	}
+
+	for _, ns := range namespace {
```
caseydavenport, line 964: What's the behavior if there are duplicates in this slice?

pkg/controller/manager/manager_controller.go
vara2504: Done

pkg/controller/clusterconnection/clusterconnection_controller.go
vara2504: Added a comment and updated the variable name as sgLayerTigeraNameSpaces

pkg/render/manager.go
vara2504: Done. Passing the full list from the calling controllers.

pkg/render/manager.go
vara2504: Replaced the slice into map[string]bool to avoid duplicates

pkg/controller/clusterconnection/clusterconnection_controller.go
```diff
@@ -331,16 +331,29 @@ func (r *ReconcileConnection) Reconcile(ctx context.Context, request reconcile.R
 		}
 	}
 
+	// Populate a list of namespaces to be displayed in the service graph Tigera infrastructure layer.
+	sgLayerTigeraNameSpaces := render.DefaultSGLayerTigeraNamespaces()
+	if !sgLayerTigeraNameSpaces[render.GuardianNamespace] {
```
rene-dekker, line 336: I think there is no need for the if statement and we can simply set it to true or am I missing something?

pkg/controller/clusterconnection/clusterconnection_controller.go
```diff
@@ -331,16 +331,29 @@ func (r *ReconcileConnection) Reconcile(ctx context.Context, request reconcile.R
...
+		sgLayerTigeraNameSpaces[render.GuardianNamespace] = true
+	}
+	amz, err := utils.GetAmazonCloudIntegration(ctx, r.Client)
+	if err != nil {
+		log.Error(err, "Failed to  fetch GetAmazonCloudIntegration info")
```
rene-dekker, line 341: nit: 2 spaces "to  fetch"

pkg/controller/clusterconnection/clusterconnection_controller.go
```diff
@@ -331,16 +331,29 @@ func (r *ReconcileConnection) Reconcile(ctx context.Context, request reconcile.R
 		}
 	}
 
+	// Populate a list of namespaces to be displayed in the service graph Tigera infrastructure layer.
+	sgLayerTigeraNameSpaces := render.DefaultSGLayerTigeraNamespaces()
+	if !sgLayerTigeraNameSpaces[render.GuardianNamespace] {
+		sgLayerTigeraNameSpaces[render.GuardianNamespace] = true
+	}
+	amz, err := utils.GetAmazonCloudIntegration(ctx, r.Client)
+	if err != nil {
```
rene-dekker, line 340: If there is no AmazonCloudIntegration in the cluster, then err will not be nil. I think you want to filter out the NotFound error. If NotFound then we do nothing. If it is another error, we need to SetDegraded()

pkg/controller/clusterconnection/clusterconnection_controller.go
rene-dekker: I think this applies to all of these if-statements in the PR.

pkg/render/manager.go
```diff
@@ -1011,3 +995,21 @@ func managerClusterWideDefaultView() *v3.UISettings {
 		},
 	}
 }
+
+func DefaultSGLayerTigeraNamespaces() map[string]bool {
```
rene-dekker, line 999: Maybe you can add a note that a map is used for deduplication only and that bools have no meaning (since you are not filtering out namespaces that are false).

vara2504: Done

pkg/controller/clusterconnection/clusterconnection_controller.go
vara2504: Done

pkg/controller/clusterconnection/clusterconnection_controller.go
vara2504: Done

pkg/controller/clusterconnection/clusterconnection_controller.go
vara2504: Done

vara2504:
@rene-dekker Addressed the review comments.

pkg/controller/clusterconnection/clusterconnection_controller.go
```diff
@@ -331,16 +331,28 @@ func (r *ReconcileConnection) Reconcile(ctx context.Context, request reconcile.R
 		}
 	}
 
+	// Populate a list of namespaces to be displayed in the service graph Tigera infrastructure layer.
+	sgLayerTigeraNameSpaces := render.DefaultSGLayerTigeraNamespaces()
+	sgLayerTigeraNameSpaces[render.GuardianNamespace] = true
+
+	amz, err := utils.GetAmazonCloudIntegration(ctx, r.Client)
+	if err != nil && !k8serrors.IsNotFound(err) {
+		r.status.SetDegraded(operatorv1.ResourceReadError, "Failed to get the GetAmazonCloudIntegration configuration", err, reqLogger)
```
rene-dekker, line 340: You should return after this.

pkg/controller/manager/manager_controller.go
```diff
@@ -521,6 +526,13 @@ func (r *ReconcileManager) Reconcile(ctx context.Context, request reconcile.Requ
 		replicas = &mcmReplicas
 	}
 
+	amz, err := utils.GetAmazonCloudIntegration(ctx, r.client)
+	if err != nil && !errors.IsNotFound(err) {
+		r.status.SetDegraded(operatorv1.ResourceReadError, "Failed to get the GetAmazonCloudIntegration configuration", err, reqLogger)
```
rene-dekker, line 531: You should return after this.

pkg/controller/clusterconnection/clusterconnection_controller.go
vara2504: Done

pkg/controller/manager/manager_controller.go
vara2504: Done

pkg/controller/manager/manager_controller.go
```diff
@@ -359,6 +359,10 @@ func (r *ReconcileManager) Reconcile(ctx context.Context, request reconcile.Requ
 		trustedSecretNames = append(trustedSecretNames, render.ComplianceServerCertSecret)
 	}
 
+	// Populate a list of namespaces to be displayed in the service graph Tigera infrastructure layer.
+	sgLayerTigeraNameSpaces := render.DefaultSGLayerTigeraNamespaces()
+	sgLayerTigeraNameSpaces[render.ManagerNamespace] = true
```
rene-dekker, line 364: I just realize that compliance is an optional namespace as well. The CR is already fetched in this controller.

rene-dekker (approved):
lgtm

PR #2645, opened 2023-05-17
Use correct calico-node UID when running in non-privileged mode.

sridhartigera:
## Description

When enabling non-privileged mode, the ownership of directories that `calico-node` uses should be changed to UID of calico-node. 

In non-privileged mode, the UID of calico-node is 10001. 

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/common/securitycontext/security_context.go
+++ b/pkg/render/common/securitycontext/security_context.go
@@ -76,8 +76,3 @@ func NewNonRootPodContext() *corev1.PodSecurityContext {
 		},
 	}
 }
-
-// GetNonRootUID returns the non-root UID
-func GetNonRootUID() int64 {
-	return runAsUserID
-}
--- a/pkg/render/common/securitycontext/security_context.go
+++ b/pkg/render/common/securitycontext/security_context.go
@@ -76,3 +76,8 @@ func NewNonRootPodContext() *corev1.PodSecurityContext {
 		},
 	}
 }
+
+// GetNonRootUID returns the non-root UID
+func GetNonRootUID() int64 {
+	return runAsUserID
+}
--- a/pkg/render/node.go
+++ b/pkg/render/node.go
@@ -1820,9 +1820,9 @@ func (c *nodeComponent) hostPathInitContainer() corev1.Container {
 		ImagePullPolicy: ImagePullPolicy(),
 		Command:         []string{"sh", "-c", "calico-node -hostpath-init"},
 		Env: []corev1.EnvVar{
-			{Name: "NODE_USER_ID", Value: "999"},
+			{Name: "NODE_USER_ID", Value: fmt.Sprintf("%d", securitycontext.GetNonRootUID())},
 		},
-		SecurityContext: securitycontext.NewRootContext(false),
+		SecurityContext: securitycontext.NewRootContext(true),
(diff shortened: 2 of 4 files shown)


pkg/render/node.go
```diff
@@ -1820,9 +1820,9 @@ func (c *nodeComponent) hostPathInitContainer() corev1.Container {
 		ImagePullPolicy: ImagePullPolicy(),
 		Command:         []string{"sh", "-c", "calico-node -hostpath-init"},
 		Env: []corev1.EnvVar{
-			{Name: "NODE_USER_ID", Value: "999"},
+			{Name: "NODE_USER_ID", Value: fmt.Sprintf("%d", securitycontext.GetNonRootUID())},
```
hjiawei, line 1823: You can use [strconv.FormatInt](https://pkg.go.dev/strconv#FormatInt).

pkg/render/node.go
```diff
@@ -1820,9 +1820,9 @@ func (c *nodeComponent) hostPathInitContainer() corev1.Container {
...
 		},
-		SecurityContext: securitycontext.NewRootContext(false),
+		SecurityContext: securitycontext.NewRootContext(true),
```
hjiawei, line 1825: Why is this changed to `true` (privileged)?

pkg/render/node_test.go
```diff
@@ -896,7 +897,7 @@ var _ = Describe("Node rendering tests", func() {
 
 				// hostpath init container should have the correct env and security context.
 				hostPathContainer := rtest.GetContainer(ds.Spec.Template.Spec.InitContainers, "hostpath-init")
-				rtest.ExpectEnv(hostPathContainer.Env, "NODE_USER_ID", "999")
+				rtest.ExpectEnv(hostPathContainer.Env, "NODE_USER_ID", fmt.Sprintf("%d", securitycontext.GetNonRootUID()))
```
hjiawei, line 900: You can use `10001` explicitly here and remove the new `GetNonRootUID()` function.

pkg/render/common/securitycontext/security_context.go
```diff
@@ -76,3 +76,8 @@ func NewNonRootPodContext() *corev1.PodSecurityContext {
 		},
 	}
 }
+
+// GetNonRootUID returns the non-root UID
```
hjiawei, line 80: You can revert this change and use `10001` directly in your test. We do this in other places as well, for example, https://github.com/tigera/operator/blob/88d6475f1f63e0249fc6acdd10bd52c018951bf4/pkg/render/node_test.go#L881.

hjiawei (changes requested):
See inline comments.

pkg/render/node.go
sridhartigera: I see issues when the init container does `mkdirAll /var/log/calico/cni`.

hjiawei (approved):
LGTM with the latest changes. Thanks.

Brian-McM (approved):

pkg/render/node.go
caseydavenport: @sridhartigera do you know why this was working before then? Or was it never working? Or is this only needed for BPF mode? Or something about your setup?

PR #2653, opened 2023-05-18
Remove GeoIP downloader from Elasticsearch as we have no use for it.

rene-dekker:
Currently we have a denied flow in brand new clusters, which does not give a very nice impression.


tmjd (approved):
LGTM

PR #2657, opened 2023-05-19
rs-979: [applicationlayer] configurable envoy xff settings

electricjesus:
## Description

[ApplicationLayer] We want to add configurable XFF settings to our Envoy-as-a-daemonset config template to allow customers to add extra trusted hops to preserve remote IPs in XFF headers in the case of them using extra load balancers. 

For load balancers like NLB with Preserve Client IP Addresses setting on and Proxy Protocol disabled, this is a required change.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


pkg/render/applicationlayer/envoy-config.yaml.template
```diff
@@ -39,6 +39,12 @@
                       transport_api_version: V3
 {{- end -}}
 
+{{- define "XFF" -}}
+                xff_num_trusted_hops: {{.NumTrustedHopsXFF}}
+                use_remote_address: true
```
peterkellydev, line 44: did you want `use_remote_address` and `skip_xff_append` to be configurable too @electricjesus ?

peterkellydev:
LGTM from a RS and L7 perspective to address the issue with XFF propagation (not a review from an operator maintainers perspective).

pkg/controller/applicationlayer/applicationlayer_controller.go
```diff
@@ -172,7 +172,7 @@ func (r *ReconcileApplicationLayer) Reconcile(ctx context.Context, request recon
 	reqLogger := log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
 	reqLogger.Info("Reconciling ApplicationLayer")
 
-	applicationLayer, err := getApplicationLayer(ctx, r.client)
```
electricjesus, line 175: I renamed this variable as `alp` because I wasn't having such a great time scanning the code because there's a package name referenced '`applicationlayer`' and there's this variable that was named '`applicationLayer`' with basically just a capitalisation the difference 😭

tmjd: Another option would be to switch its name to be `instance`. At least about half of the controllers fetch their main controlling CR as `instance`. If you don't want to switch to instance (which I'm not totally sold on as a good name) I'd at least prefer to switch to `al` as the `p` for policy is just one aspect that the resource controls.

pkg/controller/applicationlayer/applicationlayer_controller.go
```diff
@@ -271,22 +267,29 @@ func (r *ReconcileApplicationLayer) Reconcile(ctx context.Context, request recon
 		}
 	}
 
-	lcSpec := applicationLayer.Spec.LogCollection
+	// prior client.Patch may nil out missing fields (in cases of outdated crd spec).
+	// We need to re-fill out defaults again. We don't need to do this if CRD is updated to match.
+	updateApplicationLayerWithDefaults(alp)
```
tmjd, line 272: I don't think this is needed. We can assume if there are updated CRDs that go along with an updated operator, they will be delivered together.

pkg/controller/applicationlayer/applicationlayer_controller.go
```diff
@@ -271,22 +267,29 @@ func (r *ReconcileApplicationLayer) Reconcile(ctx context.Context, request recon
...
-		ALPEnabled:             r.isALPEnabled(&applicationLayer.Spec),
+		WAFEnabled:             r.isWAFEnabled(&alp.Spec),
+		LogsEnabled:            r.isLogsCollectionEnabled(&alp.Spec),
+		ALPEnabled:             r.isALPEnabled(&alp.Spec),
 		LogRequestsPerInterval: lcSpec.LogRequestsPerInterval,
 		LogIntervalSeconds:     lcSpec.LogIntervalSeconds,
 		ModSecurityConfigMap:   modSecurityRuleSet,
+		UseRemoteAddressXFF:    alp.Spec.EnvoySettings.UseRemoteAddress,
+		NumTrustedHopsXFF:      alp.Spec.EnvoySettings.XFFNumTrustedHops,
+		SkipAppendXFF:          alp.Spec.EnvoySettings.SkipXFFAppend,
```
tmjd, line 287: This should be removed I think, since we were cutting down from the 3 new fields to only 2.

pkg/controller/applicationlayer/applicationlayer_controller.go
```diff
@@ -359,6 +362,10 @@ func updateApplicationLayerWithDefaults(al *operatorv1.ApplicationLayer) {
 	if al.Spec.ApplicationLayerPolicy == nil {
 		al.Spec.ApplicationLayerPolicy = &defaultApplicationLayerPolicyStatusType
 	}
+
+	if al.Spec.EnvoySettings == nil {
+		al.Spec.EnvoySettings = &operatorv1.EnvoySettings{}
```
tmjd, line 367: The design document suggested there were defaults for the new fields. We should set those defaults here so if in the future we decide to change them we won't change behavior for existing installs.

pkg/controller/applicationlayer/applicationlayer_controller.go
electricjesus: @tmjd great suggestion! i went with that, and will keep that in mind moving forward

api/v1/applicationlayer_types.go
```diff
@@ -45,8 +47,20 @@ const (
 	ApplicationLayerPolicyDisabled ApplicationLayerPolicyStatusType = "Disabled"
 )
 
-type LogCollectionSpec struct {
+type EnvoySettings struct {
+	// The number of additional ingress proxy hops from the right side of the
+	// x-forwarded-for HTTP header to trust when determining the origin client’s
+	// IP address. 0 is permitted, but >=1 is the typical setting.
+	// +optional
```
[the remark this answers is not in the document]
electricjesus: alright, np.. I addressed this one. I also thought adding the other kubebuilder fields. Hope that's ok, thanks!

pkg/render/applicationlayer/applicationlayer.go
```diff
@@ -102,6 +102,11 @@ type Config struct {
 	dikastesEnabled bool
 	envoyConfigMap  *corev1.ConfigMap
 
+	// envoy user-configurable overrides
+	UseRemoteAddressXFF bool
+	NumTrustedHopsXFF   int32
+	SkipAppendXFF       bool
```
tmjd, line 108: Sorry one more very minor thing. We're not making use of this field anymore so please remove it.

tmjd (approved):
I've changed the the milestone on this PR to v1.31.0. If this needs to be picked to any older versions then those PRs will need the appropriate milestones.

Also there are several commits in this PR and I think they need to be squashed down when merging.

LGTM

PR #2658, opened 2023-05-22
IP pool controller for more powerful IP pool UX

caseydavenport:
## Description

More capable IP pool management controller in the operator.

It is meant to enable a few key use-cases:
- Installing multiple IP pools at install time via the operator.
- Incorporating IP pools not created via the operator into the operator's logic.
- Allow modification of IP pools post-install-time via the operator API.

The standard install-time flow is this:

- The core controller still gets the Installation and performs defaulting on it. It then waits for IP pools to exist.
- The new IP pool controller gets the installation, defaults and/or extracts IP pools from it, and creates them.
- The core controller sees the new pools and continues.

The IP pool controller will refuse to update existing IP pools, or delete IP pools, if the Calico API server is not running. This ensures we have proper validation and defaulting enforced, while allowing a small window using the crd.projectcalico.org API to bootstrap the cluster.

On upgrade, the IP pool controller will attempt to take over control of any IP pools in the cluster that match those in the Installation resource. This allows the operator to assume control of IP pools it previously had created. If there are IP pools present that are NOT in the Installation (and not owned by the controller), they will be ignored by the controller.

Other changes in this PR:
- Add a "Name" field to allow configuration of IP pool name.

## For PR author

- [x] Tests for change.

This PR includes a handful of termination / uninstall changes that are made necessary by the new FV tests added for the IP pool controller exposing an existing bug that is fixed by:

- Adjustments to the termination logic in this PR
- This API server change: https://github.com/projectcalico/calico/pull/8586

## For PR reviewers

- [x] Appropriate labels:


pkg/controller/ippool/pool_controller.go
```diff
@@ -0,0 +1,287 @@
...
+	// - Verify that the pool doesn't overlap with existing IPAM blocks.
+	// We should add this validation to productize this. Ideally, we'd go through the API server (and maybe we should at steady-state).
+	// We only need to use the crd.projectcalico.org API for bootstrapping.
+	passThru := render.NewPassthrough(toCreate...)
+	handler := utils.NewComponentHandler(log, r.client, r.scheme, nil)
+	if err := handler.CreateOrUpdateOrDelete(ctx, passThru, nil); err != nil {
+		r.status.SetDegraded(operator.ResourceUpdateError, "Error creating / updating IPPools", err, log)
+		return reconcile.Result{}, err
+	}
+	delPassThru := render.NewDeletionPassthrough(toDelete...)
```
caseydavenport, line 211: Deletion is even more tricky - we need to disable, wait, then delete the pool, then release affinities for all the blocks within that pool. 

This is fairly complicated logic that is run within the API server today and isn't easily accessible from this controller. 

That said, there might be a way to either:
- Move that code to this controller in a maintainable way.
- Introduce a new concept of "marked for deletion", so that something like kube-controllers can actually perform the deletion and this controller would just mark the pools without needing to do the complicated parts.

caseydavenport: In the latest commit, I update the controller so it will only ever send deletes via the Calico API server - I think this is probably OK.

In the case that the apiserver isn't installed or functioning, it will simply disable the IP pool as a sort of stop-gap deletion until the API is available.

pkg/controller/ippool/pool_controller.go
```diff
@@ -0,0 +1,327 @@
...
+	// For each pool that is desired, but doesn't exist, create it.
+	toCreate := []client.Object{}
+	for _, p := range instance.Spec.CalicoNetwork.IPPools {
+		if pool, ok := ourPools[p.CIDR]; !ok || !reflect.DeepEqual(pool, p) {
+			// Create a new pool, or update an existing one.
+			res := p.ToCRD()
+			if apiAvailable {
+				toCreate = append(toCreate, crdToV3(res))
+			} else {
+				toCreate = append(toCreate, res)
```
Brian-McM, line 190: Is the reason we need to bypass the api server if it's not running because the api server can't be brought up initially without some IPPool configured for pod networking?

After the initial networking is setup, if the apiserver is down the user wouldn't be able to create or delete IPPools currently as it requires the apiserver, or is this incorrect?

pkg/controller/ippool/pool_controller.go
```diff
@@ -0,0 +1,327 @@
...
+			reqLogger.WithValues("cidr", cidr).Info("Pool needs to be deleted")
+			if apiAvailable {
+				// No match. Needs delete. We only ever send deletes via the API server,
+				// since deletion requires rather complex logic. If the API server isn't available,
+				// we'll instead just mark the pool as disabled temporarily.
+				toDelete = append(toDelete, crdToV3(&pool))
+			} else {
+				// API server is not available. Just mark the pool as disabled so that new allocations
+				// don't come from this pool. We'll delete it once the API server is available.
+				pool.Spec.Disabled = true
```
Brian-McM, line 217: Similar to my comment about creating IPPools, would we really only need this during initial installation, where maybe the IPPools we configured incorrectly and the apiserver couldn't be brought up?

pkg/controller/ippool/pool_controller.go
caseydavenport: > Is the reason we need to bypass the api server if it's not running because the api server can't be brought up initially without some IPPool configured for pod networking?

Yeah, the Calico apiserver generally runs using pod networking and so in order to access the v3 API you need to have an IP pool created first. This is a bit of a chicken/egg situation. 

> After the initial networking is setup, if the apiserver is down the user wouldn't be able to create or delete IPPools currently as it requires the apiserver, or is this incorrect?

In the current PR, creation of IP pools will still succeed even if the apiserver is down. Deletion will never occur unless it can use the v3 API, but if the apiserver is down the pool will just be disabled instead (and eventually deleted when the apiserver is running again).

pkg/controller/ippool/pool_controller.go
caseydavenport: It should be rare. I expect this will only be hit in cases where the apiserver is either accidentally not running, or because of some networking issue related to IP pools preventing the apiserver from working properly. In either case, we want to enable users to turn off IP pools in order to fix the system. Disabling the pool is a good compromise because it should effectively cordon the pool, but will wait for the apiserver to be running without finalizing the deletion (which is a complicated process that we don't want to replicate here)

pkg/controller/installation/core_controller.go
```diff
@@ -963,6 +964,12 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 		}
 	}
 
+	// Wait for IP pools to exist before continuing to install remaining components.
+	if len(instanceStatus.IPPools) == 0 {
```
caseydavenport, line 968: There's an argument to be made to switch this to just reading IPPools directly instead of waiting for the Installation status to be populated by the IP pools controller, and I feel like we should have that discussion.

I think the main point in favor of doing it this way is that it leaves us able to do whatever logic we want in the IP pools controller - validation, modification, etc - before telling this core controller that it's good to go and it allows us to do it all in one location. 

I also think that in terms of code readability, doing it this way makes it more clear that this controller is a dependent of the IP pool controller.

Adding this to the status also gives a central location to see all relevant IP pool information - the desired state as well as the actual cluster state - instead of needing to compare output from `kubectl get installation` and `kubectl get ippools`.

Using the IPPools in the data store directly means one less status field (a potentially large status field for clusters that have 100s of IP pools) and a slightly simpler to understand architecture with a single source of truth for which IP pools are actually installed in the cluster.

caseydavenport: I think I am actually changing my stance here and thinking we should just switch to reading these from the crd.projectcalico.org API instead of from the Installation status, mainly driven by thinking about really large clusters and how many IP pools may end up in the status.

pkg/controller/ippool/pool_controller.go
```diff
@@ -0,0 +1,332 @@
...
+				// v3 API is not available. Just mark the pool as disabled so that new allocations
+				// don't come from this pool. We'll delete it once the API server is available.
+				pool.Spec.Disabled = true
+				toCreate = append(toCreate, &pool)
+			}
+		}
+	}
+
+	// Update pools in the cluster.
+	// TODO: This is bypassing a lot of validaiton! For example:
```
tmjd, line 230: Since this is `TODO` do you expect that we'll add the validation listed here at some point to the operator?
Because of this perhaps we should only update with the v3 api so that validation can be done and therefore require the apiserver being available.
That would block fixing IPPools without the apiserver but that is the same situation a user would have been in before the operator could manage IPPools.

pkg/controller/ippool/pool_controller.go
```diff
@@ -0,0 +1,332 @@
...
+	r.status.ClearDegraded()
+
+	if !r.status.IsAvailable() {
+		// Schedule a kick to check again in the near future. Hopefully by then
+		// things will be available.
+		return reconcile.Result{RequeueAfter: 30 * time.Second}, nil
+	}
+
+	// Update status to include the full set of IP pools. This tells the core controller
+	// that it's good to start installing Calico.
```
tmjd, line 262: The core controller could get the status of the IPPool controller to know if it is "good to start installing Calico".

pkg/controller/installation/core_controller.go
tmjd: I've asked about using the IPPool resource in another comment. I think I like that better. The IPPools resource is the source of truth for what the cluster will be using, this status field would be as the ippool controllers sees things and not necessarily what the cluster is using. These will match +99% of the time so at that point I would ask why duplicate that information.

pkg/controller/installation/core_controller.go
```diff
@@ -1852,10 +1860,9 @@ func mergePlatformPodCIDRs(i *operator.Installation, platformCIDRs []string) err
 		v4found := false
 		v6found := false
 
-		// Currently we only support a single IPv4 and a single IPv6 CIDR configured via the operator.
-		// So, grab the 1st IPv4 and IPv6 cidrs we find and use those. This will allow the
-		// operator to work in situations where there are more than one of each.
+		// For each platform CIDR, add it as an IP pool.
 		for _, c := range platformCIDRs {
+			log.Info("Adding IP pool for platform CIDR", "cidr", c)
```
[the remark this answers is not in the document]
tmjd: Are you suggesting that we'd move all defaulting of the Installation to the IPPool controller? That doesn't make sense to me. I could get on board with moving the IPPool defaulting to the ippool controller. Doesn't some of the IPPool defaulting depend on other configuration in the Installation resource?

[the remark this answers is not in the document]
caseydavenport: > Are you suggesting that we'd move all defaulting of the Installation to the IPPool controller? That doesn't make sense to me.

I was suggesting this, although I agree it's not ideal. The tradeoff is that if we only move the pool defaulting, the logic is split and there is some small bit of cross-pollination here. Namely, that the IP pool defaulting relies on the CNI plugin defaulting.

We could:
- Have core controller default the CNI and KubernetesProvider fields.
- Have IP Pool controller pick up and default pools, then create pools.
- Have core controller further default fields that rely on IP pools (like autodetection).

That's also a bit awkward and complicated.

pkg/controller/ippool/pool_controller.go
caseydavenport: I've updated so that the core controller now waits for IPPools to actually exist in the API server.

pkg/controller/ippool/pool_controller.go
caseydavenport: Yep, I've done this now - modifications to pools of any form can only go through the API server.

The only exception is creating new pools when there are none.

pkg/controller/installation/validation.go
```diff
@@ -136,42 +137,84 @@ func validateCustomResource(instance *operatorv1.Installation) error {
...
+			render.GetIPv4Pool(instance.Spec.CalicoNetwork.IPPools) != nil &&
+			render.GetIPv6Pool(instance.Spec.CalicoNetwork.IPPools) != nil {
+			return fmt.Errorf("BPF dataplane does not support dual stack")
 		}
 
-		v4pool := render.GetIPv4Pool(instance.Spec.CalicoNetwork.IPPools)
-		v6pool := render.GetIPv6Pool(instance.Spec.CalicoNetwork.IPPools)
+		for _, pool := range instance.Spec.CalicoNetwork.IPPools {
+			// Verify the CIDR is valid.
+			isIPv4 := !strings.Contains(pool.CIDR, ":")
```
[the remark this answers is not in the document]
caseydavenport: We're not parsing the address here, and I think this was how the existing logic did it IIRC.

pkg/controller/installation/core_controller.go
```diff
@@ -380,27 +406,6 @@ func updateInstallationWithDefaults(ctx context.Context, client client.Client, i
 		return err
 	}
 
-	var openshiftConfig *configv1.Network
```
caseydavenport, line 383: This logic was only used for defaulting IP pools, so it has moved to the IP pool controller.

pkg/controller/installation/core_controller.go
```diff
@@ -411,38 +416,32 @@ func updateInstallationWithDefaults(ctx context.Context, client client.Client, i
...
+	}
+
+	err = MergeAndFillDefaults(instance, awsNode, currentPools)
 	if err != nil {
 		return err
 	}
 	return nil
 }
 
-// mergeAndFillDefaults merges in configuration from the Kubernetes provider, if applicable, and then
```
caseydavenport, line 421: This is public now for testing purposes.

pkg/controller/installation/core_controller.go
```diff
@@ -411,38 +416,32 @@ func updateInstallationWithDefaults(ctx context.Context, client client.Client, i
...
 		}
 	}
 
-	return fillDefaults(i)
+	return fillDefaults(i, currentPools)
 }
 
 // fillDefaults populates the default values onto an Installation object.
-func fillDefaults(instance *operator.Installation) error {
+func fillDefaults(instance *operator.Installation, currentPools *crdv1.IPPoolList) error {
```
caseydavenport, line 444: We need to pass the current set of IP pools, as read from the cluster's API server, so that Installation defaulting can fill in things like IP auto detection which we do based on configured pools.

This is a notable change in behavior - previously, we only did this based on the pools in the installation. But now we do it based on the IP pools in the cluster as a whole. i.e., if a pool was created by calicoctl, etc. 

I think this is OK enough, given we have written previous defaults back to the Installation on any existing clusters so it won't break upgrade, and any IP pools created after Calico is started will be ignored by this defaulting anyway for the same reason.

caseydavenport: We could of course switch this to continue using the IP pools in the Installation though, in order to maintain the behavior that the operator simply ignores IP pools that it itself didn't create.

pkg/controller/installation/core_controller.go
```diff
@@ -563,26 +561,6 @@ func fillDefaults(instance *operator.Installation) error {
 		}
 	}
 
-	// Only default IP pools if explicitly nil; we use the empty slice to mean "no IP pools".
```
caseydavenport, line 566: This defaulting moved to the IP pool controller.

pkg/controller/installation/core_controller.go
```diff
@@ -896,7 +845,7 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 			return reconcile.Result{}, err
 		}
 		for _, x := range crb.Finalizers {
-			if x == render.NodeFinalizer {
```
caseydavenport, line 899: This was an incidental name change as a result of my investigation into some uninstall errors I was seeing with this change. I think this is a better name anyway, since it's the CNI resources we're actually finalizing.

pkg/controller/installation/core_controller.go
```diff
@@ -1347,6 +1308,7 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 	nodeCfg := render.NodeConfiguration{
 		K8sServiceEp:            k8sapi.Endpoint,
 		Installation:            &instance.Spec,
+		IPPools:                 ippool.CRDPoolsToOperator(currentPools.Items),
```
caseydavenport, line 1311: The same minor behavioral change here -  we use all configured pools, not just the one in the Installation CR.

I do think this is the "correct" behavior. These pools are used to determine if the CNI plugin should assign an IPv4 and/or an IPv6 address when building the CNI config JSON. Using the full set of "enabled" IP pools in the cluster seems like the right way to do that.

pkg/controller/installation/validation.go
```diff
@@ -187,102 +181,31 @@ func validateCustomResource(instance *operatorv1.Installation) error {
...
-
-				// Verify that the CIDR contains the blocksize.
-				ones, _ := cidr.Mask.Size()
-				if int32(ones) > *v4pool.BlockSize {
-					return fmt.Errorf("IP pool size is too small. It must be equal to or greater than the block size.")
-				}
-			}
-		}
-
-		if v6pool != nil {
```
caseydavenport, line 243: Refactored this into a loop above, which saved a bunch of lines of code since many of these checks were duplicated between IPv4 and IPv6 previously.

pkg/controller/ippool/pool_controller.go
```diff
@@ -0,0 +1,789 @@
...
+			if apiAvailable {
+				// v3 API is available - send a delete request.
+				v3res, err := v1ToV3(&v1res)
+				if err != nil {
+					r.status.SetDegraded(operator.ResourceValidationError, "error handling IP pool", err, reqLogger)
+					return reconcile.Result{}, err
+				}
+				toDelete = append(toDelete, v3res)
+			} else {
+				// v3 API is not available. Just mark the pool as disabled so that new allocations
```
caseydavenport, line 688: This was an attempt at a clever UX improvement, but I am wondering if it's perhaps too complicated and it's better to just error in this case. I can imagine this tripping folks up when they see IP pools in their cluster but they aren't functioning. Maybe better to just say "Can't delete IP pools without the API server" ?

api/v1/installation_types.go
```diff
@@ -435,6 +436,7 @@ type CalicoNetworkSpec struct {
 	// IPPools contains a list of IP pools to create if none exist. At most one IP pool of each
 	// address family may be specified. If omitted, a single pool will be configured if needed.
 	// +optional
+	// +kubebuilder:validation:MaxItems=25
```
caseydavenport, line 439: I think this is probably an OK limitation to put in to start? TBD. Most clusters shouldn't need 25+ pools. Better to start somewhat strict and relax later if the need arises?

pkg/controller/installation/core_controller.go
Brian-McM: > It's also probably on average the more desirable behavior.

Could you elaborate on why this is more desirable?

pkg/controller/ippool/pool_controller.go
Brian-McM: I think failing if the API server isn't available could make sense. 

The only reason I can think to not do that is if failing interfered with the API server becoming available. For instance, if we fail here then we wouldn't create / update the IPPools configured in the installation CR and those changes could contain a needed networking fix that would bring the API server back up. Is this a situation that could happen?

We could always deal with attempting to remove IPPools after we've add / updated the ones in the installation CR though.

pkg/controller/installation/core_controller.go
caseydavenport: In most cases, I think this is functionally equivalent to how it behaved before - the IP pools in the cluster should match those in the Installation exactly.

However, the one scenario where this is more desirable is the case where users want to manage IP pools by hand - e.g., using calicoctl - and disable IP pool management in the operator entirely. Without this, that scenario doesn't default the Installation resource as correctly. It will see no IP pools in the InstallationSpec and assume that IP autodetection should be disabled and IPAM assignment should be disabled.

It's niche, but I think the right behavior in that niche case.

For all other cases, I believe this is largely the same. If at least one IP pool of all address families in-use on the cluster is defined in the Installation, then the defaulting result will be the same regardless of this change.

api/v1/installation_types.go
caseydavenport: My original motivation was that I wanted to add more powerful kubebuilder validation here - e.g., uniqueness validaiton - but that it was rejected when the slice was boundless as too complicated of a validation rule.

I ended up not doing that for other reasons, but I kept this around since it's a hint to users that they might be doing something wrong if they need lots of IP pools (which is usually the case, but not necessarily always).

Maybe I should just remove it.

pkg/apis/crd.projectcalico.org/v1/ippool.go
```diff
@@ -70,8 +70,19 @@ type IPPoolSpec struct {
...
 	AWSSubnetID string `json:"awsSubnetID,omitempty" validate:"omitempty"`
+
+	// AllowedUse controls what the IP pool will be used for.  If not specified or empty, defaults to
+	// ["Tunnel", "Workload"] for back-compatibility
+	AllowedUses []IPPoolAllowedUse `json:"allowedUses,omitempty" validate:"omitempty"`
 }
 
+type IPPoolAllowedUse string
+
+const (
```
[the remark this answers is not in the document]
caseydavenport: These are manual updates

pkg/controller/installation/core_controller.go
```diff
@@ -88,19 +89,21 @@ const (
...
+// prevents pod-networked pods (e.g., kube-controllers) from teminating because the CNI plugin no longer has permissions.
+//
 // To ensure this problem does not happen we add a finalizer to the Installation resource and to the
 // calico-node ClusterRoleBinding, ClusterRole, and ServiceAccount.
-// The finalizer on the Installation resource is so that the controller knows that it is time to tear down
+//
+// - The finalizer on the Installation resource is so that the controller knows that it is time to tear down
 // and cleanup. This also allows the Installation resource to remain while the controller cleans up.
-// The finalizer on the calico-node resources is to ensure those resources remain when the Installation
+// - The finalizer on the calico-node resources ensures those resources remain when the Installation
```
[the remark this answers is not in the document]
caseydavenport: I suppose if I make it "The finalizers" (which it probably should be) then "ensures" should become "ensure"

pkg/controller/installation/core_controller.go
```diff
@@ -767,8 +763,8 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 		return reconcile.Result{}, err
 	}
 
-	terminating := (instance.DeletionTimestamp != nil)
-	if terminating {
+	installationDeleted := (instance.DeletionTimestamp != nil)
```
tmjd, line 766: I don't like this name change, if the installation was deleted then this assignment would be accessing invalid memory and the assignment value could be changed to `instance == nil`.

tmjd: I also don't think it is accurate, as it isn't deleted yet. I think `terminating` is terse but descriptive enough but if you don't think so then I'd suggest something like `installationTerminating` or `installationCleanup`.

pkg/controller/ippool/defaults.go
```diff
@@ -0,0 +1,330 @@
...
+func fillDefaults(ctx context.Context, client client.Client, instance *operator.Installation, currentPools *crdv1.IPPoolList) error {
+	if instance.Spec.CNI == nil || instance.Spec.CNI.IPAM == nil {
+		// These fields are needed for IP pool defaulting but defaulted themselves by the core Installation controller, which this controller waits for before
+		// running. We should never hit this branch, but handle it just in case.
+		return fmt.Errorf("Cannot perform IP pool defaulting until CNI configuration is available")
+	}
+
+	// Only add default CIDRs if there are no existing pools in the cluster. If there are existing pools in the cluster,
+	// then we assume that the user has configured them correctly out-of-band and we should not install any others.
+	if currentPools == nil || len(currentPools.Items) == 0 {
```
tmjd, line 73: Since this includes `len(currentPools.Items) == 0` is there a way to have the operator wait for pools to be created externally?

pkg/controller/ippool/pool_controller.go
```diff
@@ -0,0 +1,378 @@
...
+	// This label is used to track which IP pools are managed by this controller. Any IP pool
+	// with this label key/value pair is assumed to be solely managed and reconciled by this controller.
+	// We can't use OwnerReferences until this issue is resolved: https://github.com/projectcalico/calico/issues/5715
+	managedByLabel = "app.kubernetes.io/managed-by"
+	managedByValue = "tigera-operator"
+)
+
+// hasOwner returns true if the given IP pool is owned by the given Installation object, and
+// false otherwise.
+func hasOwner(pool *crdv1.IPPool, installation *operator.Installation) bool {
```
tmjd, line 133: Should we change this to `hasOwnerLabel` to make sure no one gets confused and looks for an OwnerReference?

pkg/controller/ippool/pool_controller.go
```diff
@@ -0,0 +1,378 @@
...
+		reqLogger.Error(err, "An error occurred when querying the Installation resource")
+		return reconcile.Result{}, err
+	}
+	r.status.OnCRFound()
+	defer r.status.SetMetaData(&installation.ObjectMeta)
+
+	// This controller relies on the core Installation controller to perform initial defaulting before it can continue.
+	// The core installation controller adds a specific finalizer as part of performing defaulting,
+	// so wait for that before we continue.
+	finalizerExists := false
```
tmjd, line 170: I think perhaps we should name this `installationDefaulted` to indicate what we're using this variable for and since we don't really care if the finalizer exists.
If we changed how we determined the Installation resource has been defaulted in the future then the variable wouldn't need to change.
I think this helps make the code self documenting.

pkg/controller/ippool/pool_controller.go
```diff
@@ -0,0 +1,378 @@
...
+	// Create a lookup map of pools owned by this controller for easy access.
+	// This controller will ignore any IP pools that it itself did not create.
+	ourPools := map[string]crdv1.IPPool{}
+	for _, p := range currentPools.Items {
+		if hasOwner(&p, installation) {
+			// This pool is owned by the Installation object, so consider it ours.
+			ourPools[p.Spec.CIDR] = p
+		} else if p.Name == "default-ipv4-ippool" || p.Name == "default-ipv6-ippool" {
+			// For legacy installs prior to the existence of this controller, IP pools created by
+			// the operator would have these names. Assume they are ours.
```
tmjd, line 236: I think this is a dangerous assumption. In this case I think we should compare what is in the Installation resource to the IPPool resource and if they match then we can add the reference and manage the pool from then on.
We could even extend this for any pool and that could be the mechanism for taking an externally managed pool and allowing the operator to take it over.

pkg/controller/ippool/pool_controller.go
tmjd: My only concern with disabling a pool if it is removed from the Installation is that (assuming apiserver isn't available/ready) there is different behavior if the user deletes one pool vs deleting one pool and modifying one. In the delete only case it would be disabled as is logical. In the case of deleting and modifying, the deleted one would not be disabled because the controller would return on finding that a pool needed modification.

tmjd (commented):
Since the IPPools created by the operator don't have an OwnerReference, should we have the ippool controller clean them up at some point if the apiserver CR is deleted? It might have to wait for the Installation CR to be deleted too.

pkg/controller/installation/core_controller.go
Brian-McM: Makes sense, sounds like an improvement overall.

pkg/controller/installation/core_controller.go
caseydavenport: @tmjd strictly speaking, I think this is actually accurate - the Installation has had a "delete" API action called against it and it succeeded. The presence of a DeletionTimestamp implies that the object has been deleted.

That said, I can see why this is a confusing name. I think "terminating" is worse - it has always caused me to misinterpret this code because "terminating" to me implies the Pod API state in Kubernetes, so I always read this as "The operator pod is terminating", which is wrong.

I think the correct name for this is probably something like "installationFinalizing" or "finalizeInstallation" as that is what the code is doing when this field is "true". 

What do you think?

pkg/controller/ippool/defaults.go
caseydavenport: The way to do that is to create IP pools before creating the Installation, or to create the installation with an empty IP pools list like this:

```yaml
spec:
  calicoNetwork:
    ipPools: []
```

pkg/controller/ippool/pool_controller.go
caseydavenport: Ideally, this gets switched to using an OwnerReference very soon. I think it's fine to keep the name of this function generic so that the implementation can evolve without constant renaming.

Most likely, once my API bug fix goes in to enable it, we will need this function to check both the label and the owner reference for a small period of time.

pkg/controller/ippool/pool_controller.go
caseydavenport: I considered but didn't name it that way because the installation goes through more than one round of defaulting now, so `installationDefaulted` is ambiguous. We're waiting for the core controller to perform its defaulting, so that this controller can perform additional defaulting.

I think this variable is used only within these several lines, so it's not especially confusing. I could rename it to something like "coreInstallationDefaultingComplete" or "readyToDefaultIPPools" instead, but both of those seem a bit wordy given the limited scope of this variable. 

Maybe a helper function that hides it all away would make this feel better.

caseydavenport:
@tmjd 

> Since the IPPools created by the operator don't have an OwnerReference, should we have the ippool controller clean them up at some point if the apiserver CR is deleted? It might have to wait for the Installation CR to be deleted too.

Yeah, I've been turning this one over in my head as well. I have so far avoided it because it feels complex and error prone - in order to delete IP pools, we _need_ the API server to be running, so we'd need the pool controller to gate termination of the API server, which in turn gates the Installation.

And what if the API server was never installed? I suppose this might just fall out since we'd never apply a finalizer to the APIServer API, but then we'd never be able to delete the IP pools anyway and so we'd have written a bunch of extra logic that doesn't even catch every scenario.

I am currently of the mindset that it's best to just have the exact same behavior regardless of what was installed, which means we need the lowest common denominator of "don't remove the IP pools ever".

If someone is uninstalling Calico, they'll remove the CRDs themselves which will trigger this. If they are doing an uninsatll / reinstall, then the same IP pool might be used again, although if they change the IP pool in their Installation in-between the controller will eventually delete the old one and things will get in sync (albeit with potential for allocations from the old pool sticking around for a bit).

It's icky either way I suppose. Not sure what the best solution is.

pkg/controller/installation/core_controller.go
tmjd: I disagree that it is accurate. If the resource was deleted then it would be gone, it has a deletetion timestamp exactly because it hasn't been able to be deleted yet but there is an intention for it to be deleted.

I can see the confusion with the term "terminating".

I understand why you've included `finalize` in your name suggestion but I think that might be too generic, that could also mean we are setting the finalizers on the Installation. Also I don't know that in general, people will understand what "finalizing" would mean in a variable name because I don't know that finalizers are commonly understood.

How about `installationDeleting`? I think `installationCleanup` would be a viable option too.

pkg/controller/ippool/pool_controller.go
tmjd: Yeah I guess this is fine as it is.
I still feel like it would be better to have an ambiguous name rather than one that indicates something we don't really care about (you could say finalizerExists is ambiguous too), but I don't feel strongly about this if you like the way it currently is.

pkg/controller/ippool/pool_controller.go
tmjd: > Could you expand on this?

I was thinking the operator could take ownership of any IPPool resource that matches the spec of any pool specified in the Installation resource.

> I am not sure how to do that in a more "safe" way. I suppose you might just mean adding additional checks? i.e., if the name matches one of these two and the spec fields match?

Yes that is exactly what I meant.

api/v1/installation_types.go
Brian-McM: I'd personally err on the side of caution, we don't want to cause problems for some customer who has some use case, however odd the use case is, for having a very large number of IPPools. I'd say it's probably only worth setting the cap if there is some bound to which we or kubernetes cannot reasonably handle more than that maximum amount.

pkg/controller/installation/core_controller.go
caseydavenport: > people will understand what "finalizing" would mean in a variable name

I don't think it's necessarily a good idea to prioritize making this variable more understandable to people who don't understand what finalizers are. This code is fundamentally related to the process of finalizing the resource and we use finalizes extensively as part of this logic, so anyone trying to understand this logic better understand what finalizers are. Putting even a tiny bit of distance between the variable name and the mechanism it represents just makes the logic harder to understand IMO, as evidenced by the "terminating" confusion I've always encountered.

pkg/controller/ippool/pool_controller.go
caseydavenport: Yeah, it introduces some unwanted intermediate states that you might get stuck in. For example, in my testing I encountered situations where the pool was correctly disabled, but that means you cannot ever get out of that state without manually re-enabling the IP pool because the API server will never launch when there are no enabled IP pools.

I think it's best to just error here.

api/v1/installation_types.go
caseydavenport: @Brian-McM maybe to reframe this - today, we have a hard limit of 2 IP pools. This PR increases that limit to 25.

Does the fact that this is a 10x increase in the number of pools you can specify change how you feel about this at all?

I can remove the limitation, but once it's removed there's no adding it back without breaking users.

pkg/controller/installation/core_controller.go
caseydavenport: @tmjd I went with `installationMarkedForDeletion` as a compromise.

pkg/controller/ippool/pool_controller.go
caseydavenport: > I was thinking the operator could take ownership of any IPPool resource that matches the spec of any pool specified in the Installation resource.

This is probably the way to go, although it does add a little bit of extra work to the defaulting logic. Namely, older Installations may have the IP pool specified but no "name" field, since that is introduced in this PR. We will need to, as part of defaulting, set the "Name" field based on any matching IP pools we find within the cluster.

We probably only want to do this for "default-ipvx-ippool" though - anyone adding new pools to their Installation can fill in the name explicitly so we don't need to guess. This is tricky though, because we need to default the "name" field since it's optional and hasn't existed forever... so perhaps our defaulting _should_ look at existing IP pools as a first attempt at defaulting the name.

caseydavenport: Hah, turns out I did already implement defaulting IP pool names based on whether the CIDR matches... go figure! That's probably good:

- Defaulting updates any Installation pools with a name based on in cluster pools if they exist, followed by a generated name.
- We assume any IP pools in the cluster that match exactly the IP pools in the Installation are owned by the operator.

This logic has the upside of more easily allowing users to opt-in to operator ownership of their IP pools by simply adding them to the Installation. 

The downside is that it's a bit fuzzy and might be a bit awkward to implement - if you don't match the IP pool exactly, the operator will be confused and attempt to create a second IP pool with the same CIDR as an existing one, resulting in an error. Probably OK - it's easily recoverable and doesn't harm the cluster.

api/v1/installation_types.go
Brian-McM: > Does the fact that this is a 10x increase in the number of pools you can specify change how you feel about this at all?

Oh absolutely, I some how wasn't realising / remembering that you could only specify two before, I thought this somehow affected that cap on all IP pool in the cluster (which is ridiculous).

Seems like a good idea to cap it 😀!

pkg/controller/ippool/pool_controller.go
Brian-McM: @caseydavenport, do you think it would be a good idea to handle deleting (and subsequently returning an error) for the delete case after the create / modify cases have been handled (since we can successfully bypass the API server if it's down)?

If there was something wrong with the IPPool (I don't know what exactly) that we were trying to delete and it was the cause of the API server being down then could restrict a user from "replacing" the bad IPPool? Maybe there's no way to handle this type of theoretical error without manual intervention?

caseydavenport: > If there was something wrong with the IPPool (I don't know what exactly) that we were trying to delete and it was the cause of the API server being down then could restrict a user from "replacing" the bad IPPool?

This is desired behavior - if the API server is down we absolutely can't modify or delete IP pools, since we'd be flying blind so to speak.

Brian-McM: Right, just to double check my understanding we only want to add a slight exception for the start up case where there are no IPPools defined (so allow them to be created), is that correct?

Brian-McM (approved):

pkg/controller/ippool/pool_controller.go
caseydavenport: Yep, exactly. The only time this controller should do anything with IP pools is:
- If the API server is running.
- There is no API server, and there are no IP pools in the cluster (bootstrapping).

Issue #2031, opened 2022-06-21
Tigera Operator unable to uninstall cleanly


prikesh-patel:
## Expected Behavior

When uninstalling the _Tigera Operator_ helm chart, it should be able to remove all resource and clean itself up. Then we should be able to reinstall _Tigera Operator_ without having to restart the nodes in an EKS cluster.

## Current Behavior

Uninstalling the helm release locally (using `helm delete -n tigera-operator tigera-operator`) allows the chart to be removed successfully. But when reinstalling the same chart, the `tigera-operator` pod (keeps crashing) and no `calico-node` pods get created. The following are logs in the `tigera-operator` container.

```shell
2022/06/20 15:13:44 [INFO] Version: v1.27.5
2022/06/20 15:13:44 [INFO] Go Version: go1.17.9b7
2022/06/20 15:13:44 [INFO] Go OS/Arch: linux/amd64
2022/06/20 15:14:04 [ERROR] Get "https://XXX.gr7.eu-west-1.eks.amazonaws.com:443/api?timeout=32s": dial tcp: lookup 1D36CABDB9FD90B6E9FB961600E4045D.gr7.eu-west-1.eks.amazonaws.com on 10.52.X.X:53: read udp 100.65.X.X:50421->10.52.X.X:53: i/o timeout
```

This breaks all network connectivity in the cluster. Any new images can't be pulled.

When deleting the _Tigera Operator_ helm release through [terraform](https://registry.terraform.io/providers/hashicorp/helm/latest/docs/resources/release), the resource is unable to delete and times out. This may be due to this finaliser on the `Installation` resource, which installs calico.

```terraform
finalizers:
  - tigera.io/operator-cleanup
```

This problem has begun occurring on _Tigera Operator_ helm chart [v2.2.0](https://github.com/stevehipwell/helm-charts/releases/tag/tigera-operator-2.2.0) (_Tigera_ image [v1.27.0](https://github.com/tigera/operator/releases/tag/v1.27.0)  & _Calico_ image [v3.23.0](https://github.com/projectcalico/calico/releases/tag/v3.23.0)).

## Possible Solution

A temporary workaround is to restart the nodes. The `calico-node` pods then begin running on these new nodes, and the `tigera-operator` pod starts running without any restarts.

## Steps to Reproduce (for bugs)

1. Install chart using command `helm install tigera-operator stevehipwell/tigera-operator -n tigera-operator --version 2.2.4 --values tigera-values.yaml`. The following values can be used

[...]

PR #2662, opened 2023-05-24
RFC: Graceful termination of the operator

caseydavenport:
## Description

Allow for graceful termination of the operator to occur when both the operator and the Installation resource have been deleted simultaneously.

This was causing issues such as the following:
- https://github.com/projectcalico/calico/issues/6210
- https://github.com/tigera/operator/issues/2031

I am not sure this is a perfect solution, but in very limited testing it seems to be doing the trick for me!

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


main.go
```diff
@@ -240,6 +247,49 @@ func main() {
...
+		// Cancel the main context when we are done.
+		defer cancel()
+
+		// Wait for a signal.
+		<-sigHandler.Done()
+
+		// Check if we need to do any cleanup.
+		client := mgr.GetClient()
+		instance := &v1.Installation{}
+		if err := client.Get(ctx, utils.DefaultInstanceKey, instance); err != nil {
```
caseydavenport, line 261: A limited retry of some sort here would increase reliability of this code.

stevehipwell (commented):
This looks like a good solution. Would it make sense to add a customisable timeout?

caseydavenport:
Looks like even with foreground deletion, tigera-operator loses RBAC for the cluster before it can clean up after itself. So, we'll need to make sure the ClusterRole and Binding don't get deleted until the operator deployment itself is gone. I can't seem to find a way to do that in helm itself, so we probably need to resort to _even more finalizers_ :sob:

caseydavenport:
Probably what we should have done from the beginning is manage RBAC outside of helm (like the namespace, and CRDs). But that starts to mean a lot of yaml is managed outside of the chart.

stevehipwell:
@caseydavenport would finalizers work in this context and if so what would be the increased RBAC surface area?

Alternatively a pre delete Helm hook to delete the installation should work.

caseydavenport:
> Alternatively a pre delete Helm hook to delete the installation should work.

Right, so this would be something along the lines of executing a Job (+ClusterRole +Binding) to delete the Installation prior to actually uninstalling the chart. That seems reasonable, I'll give that a try next.

stevehipwell:
> Right, so this would be something along the lines of executing a Job (+ClusterRole +Binding) to delete the Installation prior to actually uninstalling the chart. That seems reasonable, I'll give that a try next.

You can probably use the RBAC from the chart as a pre-delete hook would be run before they're removed.

caseydavenport:
It wasn't quite as trivial as I thought it might be, but I did get a pre-delete hook working for a basic Calico + apiserver install, in combination with this PR: https://github.com/projectcalico/calico/pull/7859

One thing with this is that helm won't proceed cleaning up its CRs until the pre-delete hook is complete, and we don't want the pre-delete hook to completed until we're sure the Installation object has been dealt with.

So, my solution is to:
- have the pre-delete hook delete both the `Installation` and `APIServer` CRs, and wait for the Installation to disappear. 
- have the finalizer on the Installation gated on the cleanup of the apiserver pods

caseydavenport:
> have the finalizer on the Installation gated on the cleanup of the apiserver pods

Perhaps a smarter approach would be to have the apiserver controller put its own finalizer on the Installation (and any other controllers that depend on the Installation existing in order to perform their cleanup)

main.go
```diff
@@ -240,6 +260,68 @@ func main() {
...
+		}
+
+		if instance.DeletionTimestamp == nil {
+			// Installation isn't terminating, so we can exit immediately.
+			return
+		}
+
+		// We need to wait for termination to complete. We can do this by checking if the Installation
+		// resource has been cleaned up or not.
+		to := 60 * time.Second
```
tmjd, line 302: Should we set `terminationGracePeriodSeconds` to 60s to match this since the default is 30s?

tmjd (commented):
Looks good. One question about a change we might want to make to the operator manifest.

tmjd (approved):
LGTM

joaobettu:
Any news about this issue?

tmjd:
@joaobettu hopefully you've seen the the messages on the issues this is addressing. If you're looking for some info specific to this PR please be more specific about what you're looking for.

PR #2670, opened 2023-05-26
Add temporary role for calico-node during namespace migration

MichalFupso:
## Description
Calico fails to upgrade from manifest installation < v3.26 to operator installation > v3.26. 

The failure is caused by the introduction of new service account for cni-plugin. When calico performs upgrade and migration to the operator at the same time, during the migration calico-node looses the ability to create service token for calico-node cni-plugin in the kube-system and the migration fails. For migration we give calico-node ability to create servicetoken in kube-system as prior to v3.26, and remove the ability once all services are migrated to calico-system namespace

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/migration/namespace_migration.go
+++ b/pkg/controller/migration/namespace_migration.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2019-2022 Tigera, Inc. All rights reserved.
+// Copyright (c) 2019-2023 Tigera, Inc. All rights reserved.
 
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -57,6 +57,7 @@ const (
 	typhaDeploymentName          = "calico-typha"
 	nodeDaemonSetName            = "calico-node"
 	kubeControllerDeploymentName = "calico-kube-controllers"
+	calicoNodeMigrationName      = "calico-node-migration"
 
 	k8sServicesEndpointConfigMap = "kubernetes-services-endpoint"
 
@@ -186,6 +187,55 @@ func AddBindingForKubeSystemNode(crb *rbacv1.ClusterRoleBinding) {
 		Kind:     "Group",
 		Name:     "system:nodes",
 	})
+}
+
+// We create a cluster role and cluster role binding to give the kube-system calico-node
+// permissions to create serviceaccount tokens. This is needed to make sure that the down-level calico-node maintains
+// the permissions it needs in order to launch, since the calico-node clusterrole is shared with the
+// operator resource and the operator removes this permission on upgrade (instead
+// delegating this power to the cni-plugin in the calico-system namespace).
+//
+// This logic can likely be removed after Calico v3.35, as we'll no longer need to support upgrading from versions
+// where calico-node needed this permission.
+func ClusterRoleBindingForKubeSystemNode() *rbacv1.ClusterRoleBinding {
+	crb := &rbacv1.ClusterRoleBinding{
+		TypeMeta: metav1.TypeMeta{Kind: "ClusterRoleBinding", APIVersion: "rbac.authorization.k8s.io/v1"},
+		ObjectMeta: metav1.ObjectMeta{
+			Name: calicoNodeMigrationName,
+		},
+		RoleRef: rbacv1.RoleRef{
+			APIGroup: "rbac.authorization.k8s.io",
+			Kind:     "ClusterRole",
+			Name:     calicoNodeMigrationName,
+		},
+		Subjects: []rbacv1.Subject{
+			{
+				Kind:      "ServiceAccount",
+				Name:      "calico-node",
+				Namespace: kubeSystem,
+			},
+		},
+	}
+	return crb
+}
+
+func ClusterRoleForKubeSystemNode() *rbacv1.ClusterRole {
+	role := &rbacv1.ClusterRole{
+		TypeMeta: metav1.TypeMeta{Kind: "ClusterRole", APIVersion: "rbac.authorization.k8s.io/v1"},
+		ObjectMeta: metav1.ObjectMeta{
+			Name: calicoNodeMigrationName,
+		},
+
+		Rules: []rbacv1.PolicyRule{
+			{
+				APIGroups:     []string{""},
+				Resources:     []string{"serviceaccounts/token"},
+				ResourceNames: []string{"calico-node"},
+				Verbs:         []string{"create"},
+			},
+		},
+	}
+	return role
 }
 
 // SetTyphaAntiAffinity updates the Deployment passed in with a PodAntiAffinity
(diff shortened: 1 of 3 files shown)


pkg/controller/migration/namespace_migration.go
```diff
@@ -188,6 +189,47 @@ func AddBindingForKubeSystemNode(crb *rbacv1.ClusterRoleBinding) {
 	})
 }
 
+func AddClusterRoleBindingForKubeSystemNode() *rbacv1.ClusterRoleBinding {
```
rene-dekker, line 192: nit: The word add is unnecessary for the func name.
Maybe it would be good to explain in a comment that this is needed during migrations.

rene-dekker (approved):
lgtm

pkg/controller/migration/namespace_migration.go
mgleung: +1 on adding a comment to explain that this and the cluster role below is only used for migrations.

pkg/controller/migration/namespace_migration.go
```diff
@@ -746,3 +791,16 @@ func (m *CoreNamespaceMigration) removeNodeLabel(ctx context.Context, nodeName,
 		return true, nil
 	})
 }
+
+// remove temporary ClusterRole and CLusterRoleBinding that gives calico-node permission to create servicetoken in kube-system
+func (m *CoreNamespaceMigration) deleteTemporaryCalicoNodeClusterRole(ctx context.Context) error {
+	err := m.client.RbacV1().ClusterRoleBindings().Delete(ctx, calicoNodeMigrationName, metav1.DeleteOptions{})
+	if err != nil {
+		return err
```
mgleung, line 799: Instead of exiting here, can we collect the errors in a slice and return the errors at the end of this function? As is, if something prevents the role binding from being deleted, we will not delete the cluster role.

pkg/controller/migration/namespace_migration.go
caseydavenport: I do think we need a big comment explaining why this service account is needed, since it's not obvious why we create / delete it, and when it can be removed (i.e., we no longer need this after Calico v3.30, as we'll no longer need to support upgrade from much earlier versions, right?).

e.g.,

>// We create a cluster role and cluster role binding to give the kube-system calico-node permissions to create serviceaccount tokens. This is needed to make sure that the down-level calico-node maintains the permissions it needs in order to launch, since the calico-node clusterrole is shared with the operator resource and the operator removes this permission on upgrade (instead delegating this power to the cni-plugin in the calico-system namespace).
>//
>// This logic can likely be removed after Calico v1.30, as we'll no longer need to support upgrading from versions
>// where calico-node needed this permission.

tmjd: In the example comment you have "Calico v1.30" I assume you mean "Calico v3.30". I'm not sure I agree with getting rid of this until we know that we can't support a migration from v3.25.1, otherwise we're unnecessarily limiting a migration when we don't need to limit it.

pkg/render/node.go
```diff
@@ -231,6 +231,11 @@ func (c *nodeComponent) Objects() ([]client.Object, []client.Object) {
 		objs = append(objs, certificatemanagement.CSRClusterRole())
 	}
 
+	if c.cfg.MigrateNamespaces {
+		objs = append(objs, migration.AddClusterRoleForKubeSystemNode())
```
tmjd, line 235: To clean these up, you should add an `else` statement to this code and add the ClusterRole and ClusterRoleBinding to the list of `objsToDelete`.

pkg/render/node.go
```diff
@@ -819,6 +824,11 @@ func (c *nodeComponent) clusterAdminClusterRoleBinding() *rbacv1.ClusterRoleBind
 				Name:      CalicoNodeObjectName,
 				Namespace: common.CalicoNamespace,
 			},
+			{
+				Kind:      "ServiceAccount",
```
tmjd, line 828: Is this needed for the migration or is this unrelated to the migration issue?
If it is needed for the migration then we should conditionally add this when NamespaceMigration is true.
If this isn't related to the migration I would suggest splitting it to another PR or at least in a separate commit so if we do revert the NamespaceMigration changes in the future we don't revert this accidentally.

pkg/controller/migration/namespace_migration.go
```diff
@@ -252,6 +294,9 @@ func (m *CoreNamespaceMigration) Run(ctx context.Context, log logr.Logger) error
 	if err := m.deleteKubeSystemServiceEndPointConfigMap(ctx, log); err != nil {
 		return fmt.Errorf("failed to delete kube-system k8sServicesEndpoint ConfigMap: %s", err.Error())
 	}
+	if err := m.deleteTemporaryCalicoNodeClusterRole(ctx); err != nil {
```
tmjd, line 297: This shouldn't be needed with my recommended change in `pkg/render/node.go`.

tmjd (commented):
Thank you. Looking good, just a few comments.

tmjd (changes requested):
I'm adding a review as request changes to make sure the changes I raised in my review are addressed before merging this.

pkg/controller/migration/namespace_migration.go
caseydavenport: Yep, v3.30 :facepalm: 

But yeah, we can set that target for whatever we want. We should set it now though so that this code isn't just lying around forever, since it doesn't need to be. If you want to make it v3.35 that's fine too.

pkg/render/node.go
MichalFupso: It is needed during the migration for the new service account, but it is also needed for operator installation in the future. It's been split into it's own commit as you recommended

caseydavenport: I'm not sure I understand this change either (mostly because I'm not familiar with what this binding does).

Could you explain why this is needed long-term for the CNI plugin?

pkg/render/node.go
MichalFupso: I don't think that will work in the same way that we need. The role is only needed if we are migrating from the kube-system to calico-system namespace. It's no longer needed after the migration. Adding the `else` would try to remove non-existing `ClusterRole` and `ClusterRoleBinding `if resources are already in calico-system namespace

pkg/render/node.go
MichalFupso: The calico-cni-plugin accesses directories on the container host, without this role it's not allowed to do so on MKE.

caseydavenport: Hm. CNI plugin has needed to access directories on the host for a long long time. Have we needed this all of that time and it has just never worked? Or has something recently changed that resulted in needing this permission?

caseydavenport: OH nevermind, it previously used the calico-node serviceaccount and so it just worked. Got it!

pkg/render/node.go
tmjd: As I mentioned I expect that always returning the special ClusterRole and ClusterRoleBinding should be fine to do when the namespace migration is not needed (finished).

tmjd (approved):
LGTM

PR #2678, opened 2023-05-30
Anomaly detection jobs should query Linseed

asincu:
## Description

Migrate anomaly detection to use Linseed instead of es-gateway. In order to access Linseed, we need:
- x509 certificate and private pair and Linseed ca (to establish mTLS connection), populated using env variables:
LINSEED_CLIENT_CERT, LINSEED_CLIENT_KEY, LINSEED_CA 
- service account token mounted inside the ad-detector pod, using environment variable LINSEED_TOKEN
- Linseed k8s service, using environment variable LINSEED_URL
- RBAC at cluster level to read flowlogs, dnslogs, l7logs and create events
- rules in network policy to allow access to Linseed

Acess via es-gateway is deprecated.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/intrusion_detection.go
+++ b/pkg/render/intrusion_detection.go
@@ -236,7 +236,9 @@ func (c *intrusionDetectionComponent) Objects() ([]client.Object, []client.Objec
 			c.adDetectorServiceAccount(),
 			c.adDetectorSecret(),
 			c.adDetectorAccessRole(),
+			c.adDetectorAccessClusterRole(),
 			c.adDetectorRoleBinding(),
+			c.adDetectorClusterRoleBinding(),
 		)
 		adObjs = append(adObjs, c.adDetectorPodTemplates()...)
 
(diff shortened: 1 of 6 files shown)


pkg/render/intrusion_detection.go
```diff
@@ -1760,6 +1810,26 @@ func (c *intrusionDetectionComponent) getBaseADDetectorsPodTemplate(podTemplateN
 			Name:  "FIPS_MODE_ENABLED",
 			Value: operatorv1.IsFIPSModeEnabledString(c.cfg.Installation.FIPSMode),
 		},
+		{
+			Name:  "LINSEED_URL",
+			Value: relasticsearch.LinseedEndpoint(c.SupportedOSType(), c.cfg.ClusterDomain),
+		},
+		{
+			Name:  "LINSEED_CA",
```
rene-dekker, line 1818: I think the variable names could be improved upon:
LINSEED_CA -> LINSEED_CERT (We happen to mount a CA bundle, but it is used to validate Linseed's cert)
LINSEED_CLIENT_CERT -> TLS_CRT (This is not specific to linseed, it is the tls keypair of intrusion detection pods)
LINSEED_CLIENT_KEY -> TLS_KEY

asincu: I understand, but I would rather keep it consistent across components. These are the names we have been using in our configurations for all components and in the Linseed client.

rene-dekker (approved):
lgtm

rene-dekker (approved):
lgtm

PR #2684, opened 2023-06-02
Remove tigera-prometheus-api deployment that may have been installed …

rene-dekker:
I tested this by installing an nginx deploy in the tigera-prometheus ns, since it was not feasible to go all the way from a v1.23 cluster to master due to k8s incompatibilities.
Seems harmless enough though.


tmjd (approved):
LGTM

PR #2688, opened 2023-06-07
Modify Linseed deployment to set Elastic credentials

Josh-Tigera:
based on non-admin kube-controllers created secret rather than Elastic admin secret

## Description

Relates to Enterprise commit: https://github.com/tigera/calico-private/pull/6326

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/logstorage.go
+++ b/pkg/render/logstorage.go
@@ -91,6 +91,7 @@ const (
 	ElasticsearchInternalPort       = 9300
 	ElasticsearchOperatorUserSecret = "tigera-ee-operator-elasticsearch-access"
 	ElasticsearchAdminUserSecret    = "tigera-secure-es-elastic-user"
+	ElasticsearchLinseedUserSecret  = "tigera-ee-linseed-elasticsearch-access-gateway"
 	ElasticsearchPolicyName         = networkpolicy.TigeraComponentPolicyPrefix + "elasticsearch-access"
 	ElasticsearchInternalPolicyName = networkpolicy.TigeraComponentPolicyPrefix + "elasticsearch-internal"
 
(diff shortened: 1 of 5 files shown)


pkg/render/logstorage/linseed/linseed_test.go
```diff
@@ -527,17 +525,25 @@ func expectedContainers() []corev1.Container {
 				},
 				{
 					Name:  "ELASTIC_USERNAME",
-					Value: "elastic",
+					Value: "",
```
caseydavenport, line 528: You can just omit this, since `""` is the default value for a string field.

pkg/render/logstorage.go
```diff
@@ -91,6 +91,7 @@ const (
 	ElasticsearchInternalPort       = 9300
 	ElasticsearchOperatorUserSecret = "tigera-ee-operator-elasticsearch-access"
 	ElasticsearchAdminUserSecret    = "tigera-secure-es-elastic-user"
+	ElasticsearchLinseedUserSecret  = "tigera-ee-linseed-elasticsearch-access-gateway"
```
caseydavenport, line 94: I don't think this var name needs to end with `gateway` - Linseed shouldn't be going via es-gateway.

caseydavenport (commented):
Couple minor comments but otherwise LGTM

pkg/render/logstorage.go
Josh-Tigera: Fixed in my newest enterprise change. Added a "DirectConnection" field to the `User` struct that when set to true will result in the suffix `user-secret` instead of `access-gateway`.

rene-dekker (approved):
lgtm

PR #2693, opened 2023-06-14
Update and clarify release instructions

danudey:
Do a clarification pass on operator release instructions.

These instructions will need to be tested before merge.


tmjd (approved):
LGTM

PR #2695, opened 2023-06-19
Update versions for v3.26.1 release

danudey:


sujeet-kr (approved):

rene-dekker:
Did we need a newer tag for key-cert-provisioner? Didn't new golang CVEs come out since v1.1.9?

PR #2700, opened 2023-06-22
Set DNS trusted servers using FelixConfig instead of FELIX_ env var

nelljerram:
For Rancher and OpenShift the operator needs to set a non-default value for DNS trusted servers, and it was previously doing that by setting a FELIX_ env var in the node daemonset.  But that makes it impossible for any further tweaks to that using FelixConfiguration - because FelixConfiguration settings are always overridden by FELIX_ env vars.

To allow later tweaks, change the operator to create or patch FelixConfiguration instead.

References:
https://tigera.atlassian.net/browse/RS-897
https://tigera.atlassian.net/browse/EV-3193

## For PR author

- [x] Tests for change.

## For PR reviewers

- [x] Appropriate labels:


nelljerram:
I think the CI failure is unrelated.  Will wait for review feedback, in case there are other issues to address, before hitting Rerun.

caseydavenport (approved):
LGTM - I really like the common helper for performing FC patches. Thanks @nelljerram!

PR #2708, opened 2023-06-28
CNI tuning plugin configuration added to Installation CRD

Tamas-Biro1:
## Description

We would like to configure `sysctl` keepalive parameters using Calico CNI with Tigera Operator. This PR introduces a `sysctlTuning` parameter under `calicoNetwork` in `Installation` manifest (https://docs.tigera.io/calico/latest/reference/installation/api#operator.tigera.io/v1.CalicoNetworkSpec).
Resolves https://github.com/projectcalico/calico/issues/5886.

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


fasaxc:
@Tamas-Biro1 do you mind explaining the use case for this, what sort of settings are you wanting to tune?  We feel a little uncomfortable allowing any and all sysctls to be set inside the pod because they have pretty far-reaching powers.  Is there a particular subset we could limit it to (for example, the interface-specific sysctls)?

Tamas-Biro1:
> @Tamas-Biro1 do you mind explaining the use case for this, what sort of settings are you wanting to tune? We feel a little uncomfortable allowing any and all sysctls to be set inside the pod because they have pretty far-reaching powers. Is there a particular subset we could limit it to (for example, the interface-specific sysctls)?
> 
> You mentioned the keepalive settings, are those per-interface? Could you just set those system-wide or do you have a particular use case for setting them inside pods only.

@fasaxc We would like to set defaults for keepalive (namespaced sysctl) values for containers. As we see, when `containerd` environment starts it does not inherit some network kernel parameters from the host instead it picks default values similarly as it described here: https://stackoverflow.com/questions/54552379/tcp-keepalive-time-in-docker-container/54564456#54564456.
For `CRI-O` there is an option to change default sysctl values in the underlying container orchestrator, see: https://github.com/openshift/machine-config-operator/blob/release-4.11/templates/worker/01-worker-container-runtime/_base/files/crio.yaml
but there is no such option for `containerd` at the moment (see: https://github.com/containerd/containerd/issues/6647). The Calico CNI tuning option would be the only solution for that (https://www.cni.dev/plugins/current/meta/tuning/) and the unmanaged manifest based Calico install could change these parameters in the past. That is why I came up with this idea for Tigera Operator as well.

Here is a case study why we need this setting:
https://www.ibm.com/cloud/blog/troubleshooting-outgoing-connection-issues-with-ibm-vpc-public-and-service-gateways

Tamas-Biro1:
@fasaxc @caseydavenport can you take a look at this PR?

tmjd:
I guess another option instead of this would be to use a mutating web hook to modify all (non-hostNetworked) pods with a securityContext that sets the sysctls. At least I think that would work from reading https://kubernetes.io/docs/tasks/administer-cluster/sysctl-cluster/. I'm only pointing that out since I came across that as an option, I'm not suggesting it would be a better option.

Part of the reason I was reading about the sysctls was because I was concerned that we would be exposing node level sysctl configuration if we started using the tuning plugin. But from the first sentence in the [CNI Tuning plugin](https://www.cni.dev/plugins/current/meta/tuning/#overview) it says
> This plugin can change some system controls (sysctls) and several interface attributes (promiscuous mode, all-multicast mode, MTU and MAC address) in the network namespace.

and from what I've read it sounds like most of the `net.*` sysctls are network namespaced. So also along with the allow list you've got I'm starting to feel comfortable with adding this.

I think high level I'm not sure about the location in the Installation resource for the new field. Recently we've been chatting (internally) about OS specific configuration and I think sysctl tuning would fall into that category, so we might need to provide some direction there. Also I think there should be some validation (which I think should be added [here](https://github.com/tigera/operator/blob/master/pkg/controller/installation/validation.go) ) on the settings so if a user provided an unsupported setting they would be alerted to that instead of silently ignoring the unsupported value.

Tamas-Biro1:
> I guess another option instead of this would be to use a mutating web hook to modify all (non-hostNetworked) pods with a securityContext that sets the sysctls. At least I think that would work from reading https://kubernetes.io/docs/tasks/administer-cluster/sysctl-cluster/. I'm only pointing that out since I came across that as an option, I'm not suggesting it would be a better option.
> 
> Part of the reason I was reading about the sysctls was because I was concerned that we would be exposing node level sysctl configuration if we started using the tuning plugin. But from the first sentence in the [CNI Tuning plugin](https://www.cni.dev/plugins/current/meta/tuning/#overview) it says
> 
> > This plugin can change some system controls (sysctls) and several interface attributes (promiscuous mode, all-multicast mode, MTU and MAC address) in the network namespace.
> 
> and from what I've read it sounds like most of the `net.*` sysctls are network namespaced. So also along with the allow list you've got I'm starting to feel comfortable with adding this.
> 
> I think high level I'm not sure about the location in the Installation resource for the new field. Recently we've been chatting (internally) about OS specific configuration and I think sysctl tuning would fall into that category, so we might need to provide some direction there. Also I think there should be some validation (which I think should be added [here](https://github.com/tigera/operator/blob/master/pkg/controller/installation/validation.go) ) on the settings so if a user provided an unsupported setting they would be alerted to that instead of silently ignoring the unsupported value.

hey @tmjd ,

Thanks for you detailed response. To be honest, webhooks could put more complexity to our system in addition that would be more error prone. So based in this we'd better prefer to stay with `sysctl` way.

On the other hand I totally agree with your concerns on exposing node level sysctls. So I've made some changes on this PR to restrict setting sysctls that are not allowed and throw an alert to user, as you suggested. Please have a look at it, and also if you have different direction on the location of this setting in `Installation`, please let me know. Thank you!

abasitt:
+1 please add net.* to the allowed list.

rtheis:
@Tamas-Biro1 ping

api/v1/installation_types.go
```diff
@@ -463,6 +463,10 @@ type CalicoNetworkSpec struct {
 	// +optional
 	// +kubebuilder:validation:Enum=Enabled;Disabled
 	ContainerIPForwarding *ContainerIPForwardingType `json:"containerIPForwarding,omitempty"`
+
+	// SysctlTuning configures sysctl parameters for tuning plugin
+	// +optional
+	SysctlTuning *map[string]string `json:"sysctlTuning,omitempty"`
```
tmjd, line 469: After chatting with some other folks it would be better to replace this with:
```
Sysctl []Sysctl `json:"sysctl,omitempty"`
```
and add
```
type Sysctl struct {             
    // +kubebuilder:validation:Enum=net.ipv4.tcp_keepalive_intvl;net.ipv4.tcp_keepalive_probes;net.ipv4.tcp_keepalive_time
    Key   string `json:"key"`  
    Value string `json:"value"`
} 
```

This also provides apply time validation rather than operator load time validation. In other words it will block `kubectl apply` if there is an incorrect value instead of waiting for the error to be found when the operator processes the field.

pkg/controller/installation/validation.go
```diff
@@ -331,6 +331,30 @@ func validateCustomResource(instance *operatorv1.Installation) error {
...
+				"net.ipv4.tcp_keepalive_time":   {},
+			}
+
+			for key, value := range *pluginData {
+				if _, allowed := allowedKeys[key]; !allowed {
+					return fmt.Errorf("value %s not allowed in spec.calicoNetwork.sysctlTuning", value)
+				}
+			}
+
+		}
```
tmjd, line 357: This can all be removed since we'll depend on Kubebuilder validation.

tmjd (changes requested):
Looking good

pkg/controller/installation/validation.go
caseydavenport: Let's make the change to use kubebuilder validation, but also leave this here as a second line of defense since kubebuilder validation is circumventable by modifying the CRDs.

api/v1/installation_types.go
Tamas-Biro1: thanks!

Tamas-Biro1:
@tmjd @caseydavenport added the suggested changes to the PR, please have a look at it. thank you!

api/v1/installation_types.go
tmjd: Sorry for not being clear but part of this suggestion was to stop using a map as that is not K8s idomatic (except in annotations and labels).
With my suggestion the config would look like
```
sysctl:
  - key: net.ipv4.tcp_keepalive_intvl
    value: "15"
  - key: net.ipv4.tcp_keepalive_probes
    value: 6
```

Tamas-Biro1: got it, changed as requested

pkg/controller/installation/validation.go
```diff
@@ -331,6 +331,30 @@ func validateCustomResource(instance *operatorv1.Installation) error {
 				return fmt.Errorf("spec.calicoNetwork.containerIPForwarding is supported only for Calico CNI")
 			}
 		}
+
+		type TuningSpec struct {
+			Sysctl []operatorv1.Sysctl `json:"sysctl,omitempty"`
+			Type   string              `json:"type"`
+		}
```
tmjd, line 338: Sorry I hadn't mentioned this before because I had suggested all the changes in this file be dropped but now that we're keeping the validation, this code is unused here so should be removed. I think it probably got copied from somewhere else and missed that it was not used.

pkg/crds/operator/operator.tigera.io_installations.yaml
```diff
@@ -1483,6 +1483,17 @@ spec:
                           on interfaces that do not match the given regex.
                         type: string
                     type: object
+                      sysctl:
```
tmjd, line 1486: You should run `make manifests`, that should update the installation CRD here and then you'll also get the enums that will cause apply time validation to be created.

Tamas-Biro1: thanks

pkg/controller/installation/validation.go
Tamas-Biro1: you're right, i missed it. resolved

pkg/controller/installation/validation.go
```diff
@@ -336,6 +336,25 @@ func validateCustomResource(instance *operatorv1.Installation) error {
...
+
+			// sysctl settings
+			allowedKeys := map[string]bool{
+				"net.ipv4.tcp_keepalive_intvl":  true,
+				"net.ipv4.tcp_keepalive_probes": true,
+				"net.ipv4.tcp_keepalive_time":   true,
+			}
+
+			for _, setting := range pluginData {
+				if allowedKeys[setting.Key] {
```
tmjd, line 352: I think this validation is incorrect.
I think currently it would fail validation if one of the allowedKeys were specified.

pkg/controller/installation/validation_test.go
```diff
@@ -411,6 +411,24 @@ var _ = Describe("Installation validation tests", func() {
...
+		instance.Spec.CalicoNetwork.Sysctl = []operator.Sysctl{
+			{
+				Key:   "net.ipv4.tcp_keepalive_intvl",
+				Value: "15",
+			}, {
+				Key:   "net.ipv4.tcp_keepalive_probes",
+				Value: "6",
+			},
+			{
+				Key:   "net.ipv4.tcp_keepalive_time",
```
tmjd, line 424: These 4 keys currently match the "Allowed" tuning config, so I think this test is wrong. I'd suggest renaming this test to "should validate the allowed tuning plugin config" and adding another test with a Key that is not one of the allowed keys.

pkg/controller/migration/convert/network.go
```diff
@@ -198,6 +199,42 @@ func handleCalicoCNI(c *components, install *operatorv1.Installation) error {
...
+
+		// sysctl settings
+		allowedKeys := map[string]bool{
+			"net.ipv4.tcp_keepalive_intvl":  true,
+			"net.ipv4.tcp_keepalive_probes": true,
+			"net.ipv4.tcp_keepalive_time":   true,
+		}
+		sysctlTuning := []operatorv1.Sysctl{}
+		for _, setting := range tuningSpecData.Sysctl {
+			if allowedKeys[setting.Key] {
```
tmjd, line 228: This is  the same as the validation code, I think this errors if the allowed Keys are found, which is the opposite of what is expected.

pkg/controller/migration/convert/network_test.go
```diff
@@ -755,6 +755,149 @@ var _ = Describe("Convert network tests", func() {
 		})
 
 		Context("Calico CNI config flags", func() {
+			Describe("migrate tuning setting", func() {
+				It("sysctl tuning in config", func() {
```
tmjd, line 759: I think this test is written correct but because of the inverted validation logic it fails.

tmjd (changes requested):

pkg/controller/installation/validation.go
Tamas-Biro1: yeah sry, you are right, the whole stuff was inverted and missing a "!". Thanks for the heads up

pkg/render/node_test.go
```diff
@@ -2681,6 +2681,90 @@ var _ = Describe("Node rendering tests", func() {
...
+				sysctl := []operatorv1.Sysctl{
+					{
+						Key:   "net.ipv4.tcp_keepalive_intvl",
+						Value: "15",
+					}, {
+						Key:   "net.ipv4.tcp_keepalive_probes",
+						Value: "6",
+					},
+					{
+						Key:   "net.ipv4.tcp_keepalive_time",
```
tmjd, line 2694: Is there a reason this key should be missing in the `cniCM.Data["config"]` check below?

Tamas-Biro1: do not need to check all, but to be in sync i added this as well

pkg/controller/migration/convert/network_test.go
```diff
@@ -755,6 +755,149 @@ var _ = Describe("Convert network tests", func() {
...
+							Value: "6",
+						},
+						{
+							Key:   "net.ipv4.tcp_keepalive_time",
+							Value: "40",
+						},
+					}))
+				})
+
+				It("no sysctl tuning in config, cfg must be nil", func() {
```
tmjd, line 823: This test is not named correctly. The name of this needs to be swapped with the name of the next test.

pkg/controller/migration/convert/network_test.go
```diff
@@ -755,6 +755,149 @@ var _ = Describe("Convert network tests", func() {
...
+		  "value": "40"
+		}
+	]
+  }
+]
+}`,
+					}}
+					c := fake.NewClientBuilder().WithScheme(scheme).WithObjects(ds, emptyKubeControllerSpec(), v4pool, emptyFelixConfig()).Build()
+					cfg, err := Convert(ctx, c)
+					Expect(err).ToNot(HaveOccurred())
```
tmjd, line 861: This should have errored because the key `net.ipv4.not_allowed` is in the config.

pkg/controller/installation/validation.go
```diff
@@ -336,6 +336,25 @@ func validateCustomResource(instance *operatorv1.Installation) error {
...
+				"net.ipv4.tcp_keepalive_intvl":  true,
+				"net.ipv4.tcp_keepalive_probes": true,
+				"net.ipv4.tcp_keepalive_time":   true,
+			}
+
+			for _, setting := range pluginData {
+				if !allowedKeys[setting.Key] {
+					return fmt.Errorf("key %s is not allowed in spec.calicoNetwork.sysctl", setting.Key)
+				}
+			}
```
tmjd, line 355: I'd like to see this extracted to a utility function that we can use here and in the migration code.
I'd suggest adding a function to `pkg/controller/installation/utils`, either to a new file or an existing one if you find a good place for it.

tmjd (changes requested):

pkg/controller/migration/convert/network_test.go
Tamas-Biro1: thanks, swapped

pkg/controller/migration/convert/network_test.go
Tamas-Biro1: fixed

pkg/controller/installation/validation.go
Tamas-Biro1: new function created in utils to check allowed keys. However I found another bug: the migration logic were refactored in a way that finally did nothing useful.. i also fixed that, so now it is not needed to replace it with checker function.

pkg/controller/migration/convert/network.go
```diff
@@ -198,6 +200,40 @@ func handleCalicoCNI(c *components, install *operatorv1.Installation) error {
...
+			return ErrIncompatibleCluster{
+				err:       "error parsing CNI config plugin type 'tuning'",
+				component: ComponentCNIConfig,
+				fix:       "fix CNI config",
+			}
+		}
+
+		sysctlTuning := []operatorv1.Sysctl{}
+		for _, setting := range tuningSpecData.Sysctl {
+			if _, allowed := utils.AllowedSysctlKeys[setting.Key]; allowed {
```
tmjd, line 223: There should be an else statement for this because having an unallowed Key should result in the migration being blocked/error.  This is because if we cannot duplicate the configuration that was present in the cluster before the migration then we want to block and not proceed with the migration.

tmjd: I'd suggest just putting all settings into `sysctlTuning` and then call VerifySysctl on the slice.

tmjd (changes requested):

pkg/controller/migration/convert/network.go
Tamas-Biro1: ok, changed

fasaxc:
@Tamas-Biro1 there was a failure in a unit test; I think one that you added.

Tamas-Biro1:
> @Tamas-Biro1 there was a failure in a unit test; I think one that you added.

thanks, could you help me which unit test? Or could you provide a doc how to run unit tests locally? Thank you!

PR #2713, opened 2023-07-04
Postpone policy recommendation scope watches until apiserver is up an…

rene-dekker:
Affects: EV-3878 EV-3861

### DIFF
--- a/pkg/controller/policyrecommendation/policyrecommendation_controller.go
+++ b/pkg/controller/policyrecommendation/policyrecommendation_controller.go
@@ -67,7 +67,7 @@ func Add(mgr manager.Manager, opts options.AddOptions) error {
 	tierWatchReady := &utils.ReadyFlag{}
 	policyRecScopeWatchReady := &utils.ReadyFlag{}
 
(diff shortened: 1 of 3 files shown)


dimitri-nicolo (approved):
lgtm

pkg/controller/policyrecommendation/policyrecommendation_controller.go
```diff
@@ -65,6 +65,7 @@ func Add(mgr manager.Manager, opts options.AddOptions) error {
 	}
 	licenseAPIReady := &utils.ReadyFlag{}
 	tierWatchReady := &utils.ReadyFlag{}
+	policyRecScopeWatchReady := &utils.ReadyFlag{}
```
tmjd, line 68: It seems like we should check this ready flag (like we do with the license and tier ones) before fetching the PolicyRecommendationScopes in the Reconcile.

rene-dekker: Done, thank you.

tmjd (approved):
LGTM

PR #2716, opened 2023-07-06
fix: WAF rules should be intialized as DetectionOnly post-install

electricjesus:
## Description

Use detection only to start with, because that minimises the chances of post-installation, post-feature-activation disruption.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


tmjd:
What does this mean for anyone that is already using WAF? Will it now be disabled when previously it was enabled.

tmjd (approved):
> What does this mean for anyone that is already using WAF? Will it now be disabled when previously it was enabled.

Discussed this with @electricjesus and this won't change anything for an existing user of WAF.

LGTM

PR #2726, opened 2023-07-10
Fix "make fv" target from incorrectly running UTs. Also remove

Josh-Tigera:
non-existent "vendor" directory skipPackage directive.

## Description

This will speed up CI since it will remove the duplicate tests when "make ut" and "make fv" are run as separate jobs

## For PR author

~~- [ ] Tests for change.~~

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -285,21 +285,22 @@ clean:
 ###############################################################################
 # Tests
 ###############################################################################
-WHAT?=.
+UT_DIR?=./pkg
+FV_DIR?=./test
 GINKGO_ARGS?= -v
 GINKGO_FOCUS?=.*
 
 ut:
 	-mkdir -p .go-pkg-cache report
 	$(CONTAINERIZED) $(CALICO_BUILD) sh -c '$(GIT_CONFIG_SSH) \
-	ginkgo -r --skipPackage "./test" -focus="$(GINKGO_FOCUS)" $(GINKGO_ARGS) "$(WHAT)"'
+	ginkgo -r -focus="$(GINKGO_FOCUS)" $(GINKGO_ARGS) "$(UT_DIR)"'
 
 ## Run the functional tests
 fv: cluster-create run-fvs cluster-destroy
 run-fvs:
 	-mkdir -p .go-pkg-cache report
 	$(CONTAINERIZED) $(CALICO_BUILD) sh -c '$(GIT_CONFIG_SSH) \
-	ginkgo -r --skipPackage "./pkg" -focus="$(GINKGO_FOCUS)" $(GINKGO_ARGS) "$(WHAT)"'
+	ginkgo -r -focus="$(GINKGO_FOCUS)" $(GINKGO_ARGS) "$(FV_DIR)"'
 
 ## Create a local kind dual stack cluster.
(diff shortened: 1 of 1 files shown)


Brian-McM:
It's probably fine to remove the vendor package, but I don't think it hurts to leave it in. There could always be a case where we have to re introduce vendoring go dependencies in the future, and if that happened then this command will blow up.

Makefile
```diff
@@ -292,14 +292,14 @@ GINKGO_FOCUS?=.*
...
-	ginkgo -r --skipPackage "./vendor,./test" -focus="$(GINKGO_FOCUS)" $(GINKGO_ARGS) "$(WHAT)"'
+	ginkgo -r --skipPackage "./test" -focus="$(GINKGO_FOCUS)" $(GINKGO_ARGS) "$(WHAT)"'
 
 ## Run the functional tests
 fv: cluster-create run-fvs cluster-destroy
 run-fvs:
 	-mkdir -p .go-pkg-cache report
 	$(CONTAINERIZED) $(CALICO_BUILD) sh -c '$(GIT_CONFIG_SSH) \
-	ginkgo -pkgdir test -r --skipPackage ./vendor -focus="$(GINKGO_FOCUS)" $(GINKGO_ARGS) "$(WHAT)"'
+	ginkgo -r --skipPackage "./pkg" -focus="$(GINKGO_FOCUS)" $(GINKGO_ARGS) "$(WHAT)"'
```
Brian-McM, line 302: Do you know why pkgdir wasn't working? It should have forced only the packages in the test directory to load, as per this description:
```
  -pkgdir string
    	install and load all packages from the given dir instead of the usual locations.

```

We have code in other folder locations than just pkg and there's no reason we could have tests there, so I think an explicit "inclusion" of the directories where the fvs are located would probably be better if that can be achieved.

I notice that `--skipPackage` is used and not `-skipPackage`, could it be that we need to specify `--pkgdir` instead?

Josh-Tigera: I suspected `-pkgdir` was being overridden by `-r`:
`-r    Find and run test suites under the current directory recursively.`

I agree that an explicit inclusion strategy is better so I could change each UT/FV command to explicitly point at the appropriate directory. This would remove any risk with re-introduced vendor code as well. Something like:

```suggestion
UT_DIR?="./pkg"
FV_DIR?="./test"

ut:
	ginkgo -focus="$(GINKGO_FOCUS)" $(GINKGO_ARGS) "$(UT_DIR)"'
	
run-fvs:
        ginkgo -focus="$(GINKGO_FOCUS)" $(GINKGO_ARGS) "$(FV_DIR)"'
```

What do you think?

Brian-McM: That's weird that the recursive flag would override the pkgdir, you'd think they'd work together (obviously the author thought that).

I just tried what you suggested and that looks good, I wonder why we were trying to use the pkgdir flag instead of just adding the directory as you've done...

Josh-Tigera: Made the changes as discussed and CI passing with newly expected FV time of only a few minutes.

Brian-McM (approved):

PR #2732, opened 2023-07-11
Add support for Windows hostprocess containers

coutinhop:
## Description

Add support for Windows hostprocess containers

- Remove calico windows upgrader code
- Add windows calico-node daemonset with hostprocess containers
  (rendered only if enabled by setting WindowsDataplane to 'HNS' in the Installation).

Manually tested with a clean installation and VXLAN setup and images
from https://github.com/projectcalico/calico/pull/7857.

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


api/v1/installation_types.go
```diff
@@ -171,8 +171,8 @@ type InstallationSpec struct {
 	// ComponentResources or TyphaAffinity, then these overrides take precedence.
 	TyphaDeployment *TyphaDeployment `json:"typhaDeployment,omitempty"`
 
-	// CalicoWindowsUpgradeDaemonSet configures the calico-windows-upgrade DaemonSet.
```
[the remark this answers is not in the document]
coutinhop: Sorry @tmjd, I missed your comment and am only seeing it now... One question: is there any impact of removing this field that I'm missing? I had thought that, since this PR (and the design) removes the windows upgrade daemonset itself (which is the only component that would use this field), it would be OK. I'll add it back, but I'm curious. Thanks!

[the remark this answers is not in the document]
caseydavenport: Yeah, I'm 50/50 on this one. Removing the field from the API is itself a breaking change since anyone who has GitOpts or other workflows in place to install a CR with this field set will upgrade and see that fail. But, it's disingenuous to accept this config if we know it's not going to be respected.

Safest is to leave it, release note that it is deprecated and does nothing, and then remove it from the code at a pre-determined time in the future - even if it's just the next release.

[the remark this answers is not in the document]
coutinhop: Understood. I added it back, will add a comment stating it is deprecated and will be ignored (and removed in the future). Thanks!

caseydavenport (changes requested):
First pass, will definitely need a few looks at this one!

Let me know what you think. I also have a slight concern about duplication of code between the pkg/render files for linux and windows. I am mainly worried they might get out of sync since updates to one will miss the other. Is there a good way to share that code without making it too complex?

pkg/render/windows.go
```diff
@@ -71,123 +102,1276 @@ func (c *windowsComponent) SupportedOSType() rmeta.OSType {
...
+		{Name: "CNI_BIN_DIR", Value: "/host/opt/cni/bin"},
+		{Name: "CNI_CONF_NAME", Value: "10-calico.conflist"},
+		{Name: "CNI_NET_DIR", Value: "/host/etc/cni/net.d"},
+	}
+
+	return envVars
+}
+
+// uninstallContainer creates the node's init container that uninstalls non-HPC Calico from the host.
+func (c *windowsComponent) uninstallContainer() corev1.Container {
```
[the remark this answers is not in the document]
coutinhop: The uninstall initcontainer runs this script: https://github.com/coutinhop/calico/blob/f28243adccef34b9baa523e8f48d09181f93839e/node/windows-packaging/CalicoWindows/uninstall-calico-hpc.ps1

It only removes the windows services if they are present. As far as I can tell, the only cost of having it when HPC is already there is that it cleans up calico CNI config and binaries (which are then reinstalled by the install-cni initcontainer).

Let me know what you think about this, as I think the extra logic to run this conditionally will probably add complexity, but I don't oppose if you think it's worth it...

pkg/render/windows.go
```diff
@@ -71,123 +102,1276 @@ func (c *windowsComponent) SupportedOSType() rmeta.OSType {
...
+		if c.cfg.Installation.Variant == operatorv1.TigeraSecureEnterprise {
+			windowsEnv = append(windowsEnv, corev1.EnvVar{Name: "FELIX_DNSTRUSTEDSERVERS", Value: "k8s-service:kube-system/rke2-coredns-rke2-coredns"})
+		}
+	}
+
+	if c.cfg.Installation.CNI.Type != operatorv1.PluginCalico {
+		windowsEnv = append(windowsEnv, corev1.EnvVar{Name: "FELIX_ROUTESOURCE", Value: "WorkloadIPs"})
+	}
+
+	if c.cfg.AmazonCloudIntegration != nil {
```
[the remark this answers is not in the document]
coutinhop: I didn't verify it directly, but I was thinking it would work... To be fair, I'm not too familiar with this functionality...

pkg/render/windows.go
```diff
@@ -71,123 +102,1276 @@ func (c *windowsComponent) SupportedOSType() rmeta.OSType {
...
+	kubeNetwork := "Calico.*"
+	switch c.cfg.Installation.KubernetesProvider {
+	case operatorv1.ProviderAKS:
+		kubeNetwork = "azure.*"
+	case operatorv1.ProviderEKS:
+		kubeNetwork = "vpc.*"
+	}
+	windowsEnv = append(windowsEnv, corev1.EnvVar{Name: "KUBE_NETWORK", Value: kubeNetwork})
+
+	// If there are no IP pools specified, then configure no default IP pools.
```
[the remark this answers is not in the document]
coutinhop: This is translating IP pool configuration into env variables, we'd still need this here. Granted, I should probably deduplicate the logic. Also, hear me out: Windows doesn't support a lot of this (only IPv4 VXLAN non-crosssubnet or no encap are supported), so I could also not duplicate the linux logic, and cut a lot of the windows logic out. Which do you think is best?

pkg/render/windows.go
```diff
@@ -71,123 +102,1276 @@ func (c *windowsComponent) SupportedOSType() rmeta.OSType {
...
+		SecurityContext: securitycontext.NewWindowsHostProcessContext(),
+		Env:             c.windowsEnvVars(),
+		VolumeMounts:    c.windowsVolumeMounts(),
+	}
+}
+
+// windowsEnvVars creates the node's envvars.
+func (c *windowsComponent) windowsEnvVars() []corev1.EnvVar {
+	// Set the clusterType.
+	clusterType := "k8s,operator"
```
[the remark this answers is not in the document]
coutinhop: Good point! Will do, thanks!

pkg/render/windows.go
```diff
@@ -71,123 +102,1276 @@ func (c *windowsComponent) SupportedOSType() rmeta.OSType {
...
+func (c *windowsComponent) felixContainer() corev1.Container {
+
+	lp, rp := c.windowsLivenessReadinessProbes()
+
+	return corev1.Container{
+		Name:            "felix",
+		Image:           c.nodeImage,
+		Args:            []string{"$env:CONTAINER_SANDBOX_MOUNT_POINT/CalicoWindows/felix-service.ps1"},
+		WorkingDir:      "$env:CONTAINER_SANDBOX_MOUNT_POINT/CalicoWindows/",
+		Resources:       c.windowsResources(),
```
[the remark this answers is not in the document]
coutinhop: Right, thanks for catching this! I'll update with different component names for each container

api/v1/installation_types.go
```diff
@@ -764,3 +790,25 @@ func IsFIPSModeEnabled(mode *FIPSMode) bool {
...
+	// CNIConfDir is the path to the CNI configuration directory on Windows, it must match what is on ContainerD on the Windows nodes.
+	// +optional
+	CNIConfDir string `json:"cniConfDir,omitempty"`
+
+	// CNILogDir is the path to the Calico CNI logs directory on Windows.
+	// +optional
+	CNILogDir string `json:"cniLogDir,omitempty"`
+
+	// VXLANMACPrefix is the prefix used when generating MAC addresses for virtual NICs
+	// +optional
```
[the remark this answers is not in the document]
coutinhop: You mean for `VXLANMACPrefix`, right? I think that's doable, let me try it out

api/v1/installation_types.go
```diff
@@ -390,6 +408,14 @@ type CalicoNetworkSpec struct {
 	// +kubebuilder:validation:Enum=Iptables;BPF;VPP
 	LinuxDataplane *LinuxDataplaneOption `json:"linuxDataplane,omitempty"`
 
+	// WindowsDataplane is used to select the dataplane used for Windows nodes. In particular, it
+	// causes the operator to add required mounts and environment variables for the particular dataplane.
+	// If not specified, HNS mode is used. If disabled, the operator will not render the Calico Windows nodes daemonset.
+	// Default: HNS
+	// +optional
+	// +kubebuilder:validation:Enum=HNS;disabled
```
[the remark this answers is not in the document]
coutinhop: Will do! Thanks

hack/gen-versions/components.go
```diff
@@ -40,9 +42,9 @@ var defaultImages = map[string]string{
 	"coreos-alertmanager":        "unused/image",
 	"guardian":                   "tigera/guardian",
 	"tigera-cni":                 "tigera/cni",
+	"tigera-cni-windows":         "tigera/cni-windows",
```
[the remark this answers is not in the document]
coutinhop: I couldn't find one, but it seems like there should be, right? I'll add both windows and non-windows

pkg/common/common.go
```diff
@@ -19,6 +19,8 @@ const (
 	TyphaDeploymentName           = "calico-typha"
 	NodeDaemonSetName             = "calico-node"
 	KubeControllersDeploymentName = "calico-kube-controllers"
+	WindowsDaemonSetName          = "calico-node-windows"
+	WindowsKubeProxyDaemonSetName = "kube-proxy-windows"
```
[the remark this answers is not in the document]
coutinhop: Oh, I missed this when cleaning up, thanks for catching it!

pkg/controller/installation/core_controller.go
```diff
@@ -185,29 +183,25 @@ func newReconciler(mgr manager.Manager, opts options.AddOptions) (*ReconcileInst
...
+		status:               statusManager,
+		typhaAutoscaler:      typhaScaler,
+		namespaceMigration:   nm,
+		amazonCRDExists:      opts.AmazonCRDExists,
+		enterpriseCRDsExist:  opts.EnterpriseCRDExists,
+		clusterDomain:        opts.ClusterDomain,
+		manageCRDs:           opts.ManageCRDs,
+		usePSP:               opts.UsePSP,
+		tierWatchReady:       &utils.ReadyFlag{},
+		clientset:            cs,
```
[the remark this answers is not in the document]
coutinhop: Missed cleaning this up too, thanks!

pkg/controller/installation/core_controller.go
```diff
@@ -1653,6 +1643,11 @@ func (r *ReconcileInstallation) setDefaultsOnFelixConfiguration(install *operato
 		fc.Spec.HealthPort = &felixHealthPort
 		updated = true
 	}
+	vxlanVNI := 4096
```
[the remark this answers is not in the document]
coutinhop: This is the default in FelixConfig: https://docs.tigera.io/calico/latest/reference/resources/felixconfig
Windows just needs it populated. This wouldn't break anything, would it?

pkg/controller/installation/core_controller_test.go
```diff
@@ -373,9 +371,9 @@ var _ = Describe("Testing core-controller installation", func() {
 			mockStatus.On("RemoveCertificateSigningRequests", mock.Anything)
 			mockStatus.On("ReadyToMonitor")
 			mockStatus.On("SetMetaData", mock.Anything).Return()
+			mockStatus.On("SetDegraded", mock.Anything, mock.Anything, mock.Anything, mock.Anything).Return()
```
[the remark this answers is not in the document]
coutinhop: Oh, it isn't anymore, I again missed cleaning it up... Thanks for catching it!

pkg/render/node.go
```diff
@@ -750,16 +750,16 @@ func (c *nodeComponent) nodeCNIConfigMap() *corev1.ConfigMap {
 	}
 }
 
-func (c *nodeComponent) getCalicoIPAM() map[string]interface{} {
```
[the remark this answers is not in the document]
coutinhop: Not needed, but at some point I was reusing this in render/windows.go, but ended up not using this... I can revert this

pkg/render/windows.go
```diff
@@ -71,123 +102,1276 @@ func (c *windowsComponent) SupportedOSType() rmeta.OSType {
...
-		c.windowsUpgradeDaemonset(),
+		c.windowsRole(),
+		c.windowsRoleBinding(),
+		c.cniPluginServiceAccount(),
+		c.cniPluginRole(),
+		c.cniPluginRoleBinding(),
+	}
+
+	// These are objects to keep even when we're terminating
+	objsToKeep := []client.Object{}
```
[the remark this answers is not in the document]
coutinhop: TBH I was heavily influenced by the linux node daemonset and ended up thinking this might be necessary. I'll remove it

pkg/render/windows.go
```diff
@@ -71,123 +102,1276 @@ func (c *windowsComponent) SupportedOSType() rmeta.OSType {
...
+				Name:      WindowsNodeObjectName,
+				Namespace: common.CalicoNamespace,
+			},
+		},
+	}
+	return crb
+}
+
+// windowsRole creates the clusterrole containing policy rules that allow the windows node daemonset to operate normally.
+func (c *windowsComponent) windowsRole() *rbacv1.ClusterRole {
```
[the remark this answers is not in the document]
coutinhop: >Is this meaningfully different from the default linux cluster role? Should we just bind to that?
It's mostly the same, other than a different name for the `cni-plugin` `serviceaccounts/token` rule. I'm not sure, can we sync up offline on this one?

pkg/render/windows.go
```diff
@@ -71,123 +102,1276 @@ func (c *windowsComponent) SupportedOSType() rmeta.OSType {
...
-	return corev1.Container{
-		Name:            common.CalicoWindowsUpgradeResourceName,
-		Image:           c.windowsUpgradeImage,
-		ImagePullPolicy: ImagePullPolicy(),
-		VolumeMounts:    mounts,
+	return calicoPluginConfig
+}
+
+// cniDirectories returns the CNI binary, network config and log directories and the CNI conf filename for the configured platform.
+// FIXME: populate with known default for other providers
```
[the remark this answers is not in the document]
coutinhop: Yes, I still haven't figured out the defaults for all providers (but that's not a dealbreaker as the users can configure them manually if necessary)

pkg/render/windows.go
```diff
@@ -71,123 +102,1276 @@ func (c *windowsComponent) SupportedOSType() rmeta.OSType {
...
+func (c *windowsComponent) cniDirectories() (string, string, string) {
+	var cniBinDir, cniNetDir, cniLogDir string
+	switch c.cfg.Installation.KubernetesProvider {
+	case operatorv1.ProviderAKS:
```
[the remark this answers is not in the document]
coutinhop: I see, got it! Where exactly would I place this defaulting? To `updateInstallationWithDefaults()` in https://github.com/tigera/operator/blob/master/pkg/controller/installation/core_controller.go#L397 ?

pkg/render/windows.go
```diff
@@ -71,123 +102,1276 @@ func (c *windowsComponent) SupportedOSType() rmeta.OSType {
...
+				Namespace: common.CalicoNamespace,
+			},
+		},
+	}
+	return crb
+}
+
+// clusterAdminClusterRoleBinding returns a ClusterRoleBinding for DockerEE to give
+// the cluster-admin role to calico-node-windows, this is needed for calico-node-windows to be
+// able to use hostNetwork in Docker Enterprise.
```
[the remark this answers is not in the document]
coutinhop: I'm not sure on the details of this, haven't tested on DockerEE/MKE, I admit I took this from linux node rendering as I thought it was necessary for windows as well... Can we sync up on this one offline too? Thanks!

pkg/controller/installation/validation.go
```diff
@@ -399,6 +398,15 @@ func validateCustomResource(instance *operatorv1.Installation) error {
...
 
+	// Verify the CalicoNodeWindowsDaemonSet overrides, if specified, is valid.
+	if ds := instance.Spec.CalicoNodeWindowsDaemonSet; ds != nil {
+		err := validation.ValidateReplicatedPodResourceOverrides(ds, node.ValidateCalicoNodeWindowsDaemonSetContainer, node.ValidateCalicoNodeWindowsDaemonSetInitContainer)
+		if err != nil {
+			return fmt.Errorf("Installation spec.CalicoNodeWindowsDaemonSet is not valid: %w", err)
+
+		}
+	}
+
```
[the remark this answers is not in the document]
coutinhop: I'm not sure, the fields in that struct are mostly dir paths, an interface adapter name and `VXLANMACPrefix` which I'll add a regex kubebuilder validation. Do you have any checks to suggest?

api/v1/installation_types.go
```diff
@@ -764,3 +790,25 @@ func IsFIPSModeEnabled(mode *FIPSMode) bool {
...
+	// +optional
+	CNIBinDir string `json:"cniBinDir,omitempty"`
+
+	// CNIConfDir is the path to the CNI configuration directory on Windows, it must match what is on ContainerD on the Windows nodes.
+	// +optional
+	CNIConfDir string `json:"cniConfDir,omitempty"`
+
+	// CNILogDir is the path to the Calico CNI logs directory on Windows.
+	// +optional
+	CNILogDir string `json:"cniLogDir,omitempty"`
```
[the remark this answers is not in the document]
coutinhop: I'm not entirely sure on the inner workings, but it seems to me as the configs in CNILogging should be the same on Windows (log level, file max sizes, etc). Does this require any extra work on cni-plugin in the calico repo? Otherwise I think I should add these to the windows rendering...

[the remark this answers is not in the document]
caseydavenport: I'm really not sure what our Windows CNI plugin supports in terms of logging, so you might need to dig into that one a bit :sweat_smile:

pkg/render/windows.go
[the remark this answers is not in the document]
coutinhop: That's what I did in my latest push, please let me know if that's not where you expected it to be. Thanks!

pkg/controller/installation/validation.go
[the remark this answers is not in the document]
caseydavenport: One thing I just thought of - we do need to validate that for clusters with WIndows, the user specifies Service CIDRs - we should reject that as invalid configuration.

pkg/controller/installation/windows_controller.go
```diff
@@ -0,0 +1,597 @@
...
+
+	// Watch for changes to BGPConfiguration.
+	err = c.Watch(&source.Kind{Type: &crdv1.BGPConfiguration{}}, &handler.EnqueueRequestForObject{})
+	if err != nil {
+		return fmt.Errorf("tigera-windows-controller failed to watch BGPConfiguration resource: %w", err)
+	}
+
+	if ri.enterpriseCRDsExist {
+		// Watch for changes to primary resource ManagementCluster
+		err = c.Watch(&source.Kind{Type: &operator.ManagementCluster{}}, &handler.EnqueueRequestForObject{})
```
caseydavenport, line 170: I don't think this controller actually uses these CRDs:

- ManagementCluster
- ManagementClusterConnection
- ManagerInternalLSSecretName
- PrometheusTLSSecretName (It does use NodePrometheusTLSServerSecret and PrometheusClientTLSSecretName)
- LogCollector

So, I think we can remove the watches for them.

pkg/controller/installation/windows_controller.go
```diff
@@ -0,0 +1,597 @@
...
+			return fmt.Errorf("tigera-windows-controller failed to watch secret '%s' in '%s' namespace: %w", render.ManagerInternalTLSSecretName, common.OperatorNamespace(), err)
+		}
+
+		// watch for change to primary resource LogCollector
+		err = c.Watch(&source.Kind{Type: &operator.LogCollector{}}, &handler.EnqueueRequestForObject{})
+		if err != nil {
+			return fmt.Errorf("tigera-windows-controller failed to watch primary resource: %v", err)
+		}
+
+		if ri.manageCRDs {
```
caseydavenport, line 197: I don't think the windows controller needs to add CRD watches either - the CRD management is already handled in the core controller.

pkg/controller/installation/windows_controller.go
```diff
@@ -0,0 +1,597 @@
...
+	}
+
+	preDefaultPatchFrom := client.MergeFrom(instance.DeepCopy())
+
+	// Mark CR found so we can report converter problems via tigerastatus
+	r.status.OnCRFound()
+	// SetMetaData in the TigeraStatus such as observedGenerations.
+	defer r.status.SetMetaData(&instance.ObjectMeta)
+
+	// Changes for updating Installation status conditions.
```
caseydavenport, line 280: This logic is going to cause problems because we have two controllers fighting over the status of the installation. I think we need logic to merge the conditions from both controllers.

For now, we can probably get away with just remove this block from this controller. But we should track fixing this logic in a future PR.

pkg/controller/installation/windows_controller.go
```diff
@@ -0,0 +1,597 @@
...
+	if err := validateCustomResource(instance); err != nil {
+		r.status.SetDegraded(operatorv1.InvalidConfigurationError, "Invalid Installation provided", err, reqLogger)
+		return reconcile.Result{}, err
+	}
+
+	// Write the discovered configuration back to the API. This is essentially a poor-man's defaulting, and
+	// ensures that we don't surprise anyone by changing defaults in a future version of the operator.
+	// Note that we only write the 'base' installation back. We don't want to write the changes from 'overlay', as those should only
+	// be stored in the 'overlay' resource.
+	if err := r.client.Patch(ctx, instance, preDefaultPatchFrom); err != nil {
```
caseydavenport, line 308: Since we don't do any defaulting here, we can remove this.

pkg/controller/installation/windows_controller.go
```diff
@@ -0,0 +1,597 @@
...
+		if err := validateCustomResource(instance); err != nil {
+			r.status.SetDegraded(operatorv1.InvalidConfigurationError, "Invalid computed config", err, reqLogger)
+			return reconcile.Result{}, err
+		}
+	}
+
+	// A status is needed at this point for operator scorecard tests.
+	// status.variant is written later but for some tests the reconciliation
+	// does not get to that point.
+	if reflect.DeepEqual(instanceStatus, operatorv1.InstallationStatus{}) {
```
caseydavenport, line 335: Instead of doing this, we should short-circuit. The core controller already has this logic, so we should just gate execution of this controller on the core controller filling in the status.

pkg/controller/installation/windows_controller.go
```diff
@@ -0,0 +1,597 @@
...
+		instance.Status = operatorv1.InstallationStatus{}
+		if err := r.client.Status().Update(ctx, instance); err != nil {
+			r.status.SetDegraded(operatorv1.ResourceUpdateError, "Failed to write default status", err, reqLogger)
+			return reconcile.Result{}, err
+		}
+	}
+
+	// The operator supports running in a "Calico only" mode so that it doesn't need to run TSEE specific controllers.
+	// If we are switching from this mode to one that enables TSEE, we need to restart the operator to enable the other controllers.
+	if !r.enterpriseCRDsExist && instance.Spec.Variant == operatorv1.TigeraSecureEnterprise {
```
caseydavenport, line 345: This logic is also already handled in the core_controller, so can just be removed from this one. No need to duplicate it.

pkg/controller/installation/windows_controller.go
```diff
@@ -0,0 +1,597 @@
...
+		}
+	}
+
+	certificateManager, err := certificatemanager.Create(r.client, &instance.Spec, r.clusterDomain, common.OperatorNamespace(), certificatemanager.AllowCACreation())
+	if err != nil {
+		r.status.SetDegraded(operatorv1.ResourceCreateError, "Unable to create the Tigera CA", err, reqLogger)
+		return reconcile.Result{}, err
+	}
+
+	typhaNodeTLS, err := GetOrCreateTyphaNodeTLSConfig(r.client, certificateManager)
```
caseydavenport, line 385: We should make this `GetTyphaNodeTLSConfig` - the core controller should already be creating these secrets, so let's just wait until they are ready instead of having two controllers racing to create it.

pkg/controller/installation/windows_controller.go
```diff
@@ -0,0 +1,597 @@
...
+			nodeReporterMetricsPort = *felixConfiguration.Spec.PrometheusReporterPort
+		}
+
+		if nodeReporterMetricsPort == 0 {
+			err := errors.New("felixConfiguration prometheusReporterPort=0 not supported")
+			r.status.SetDegraded(operatorv1.InvalidConfigurationError, "invalid metrics port", err, reqLogger)
+			return reconcile.Result{}, err
+		}
+
+		nodePrometheusTLS, err = certificateManager.GetOrCreateKeyPair(r.client, render.NodePrometheusTLSServerSecret, common.OperatorNamespace(), dns.GetServiceDNSNames(render.CalicoNodeMetricsService, common.CalicoNamespace, r.clusterDomain))
```
caseydavenport, line 436: This should be `GetKeyPair` instead of `GetOrCreate` - the core controller should create this for us (and we should add a comment saying as much)

pkg/controller/installation/windows_controller.go
```diff
@@ -0,0 +1,597 @@
...
+			return reconcile.Result{}, err
+		}
+		if prometheusClientCert != nil {
+			typhaNodeTLS.TrustedBundle.AddCertificates(prometheusClientCert)
+		}
+	}
+
+	// Secure calico kube controller metrics.
+	var kubeControllerTLS certificatemanagement.KeyPairInterface
+	if instance.Spec.Variant == operatorv1.TigeraSecureEnterprise {
```
caseydavenport, line 456: We don't do anything with kube-controllers for Windows, so this branch can just be deleted.

pkg/controller/installation/windows_controller.go
```diff
@@ -0,0 +1,597 @@
...
+		reqLogger.Info("Calico Windows daemonset is disabled in the operator installation")
+	} else {
+		// Build a configuration for rendering calico-node-windows, but only if there are Windows nodes in the cluster
+		hasWindowsNodes, err = common.HasWindowsNodes(r.client)
+		if err != nil {
+			r.status.SetDegraded(operatorv1.ResourceReadError, "Unable to detect if there are Windows nodes present", err, reqLogger)
+		}
+		if hasWindowsNodes {
+			if len(instance.Spec.ServiceCIDRs) == 0 {
+				r.status.SetDegraded(operatorv1.ResourceValidationError, "Calico for Windows requires at least one Kubernetes service CIDR to be configured in InstallationSpec.ServiceCIDRs", fmt.Errorf("InstallationSpec.ServiceCIDRs must be configured for Calico for Windows"), reqLogger)
```
caseydavenport, line 500: Ah, you do have this validation present it's just not where I expected it.

I was expecting this validation to be performed based on whether or not the Installation has WindowsNodes, not whether the _cluster_ has Windows nodes. 

Those are two similar but subtly different checks. We shouldn't allow the user to specify WindowsNodes without service cidrs, but I think it's OK for a cluster to have Windows nodes without specifying WindowsNodes, right?

pkg/controller/installation/windows_controller.go
```diff
@@ -0,0 +1,597 @@
...
+	bgpConfiguration := &crdv1.BGPConfiguration{}
+	err = r.client.Get(ctx, types.NamespacedName{Name: "default"}, bgpConfiguration)
+	if err != nil && !apierrors.IsNotFound(err) {
+		r.status.SetDegraded(operatorv1.ResourceReadError, "Unable to read BGPConfiguration", err, reqLogger)
+		return reconcile.Result{}, err
+	}
+
+	var hasWindowsNodes bool
+	// Don't render calico-node-windows if it's disabled in the installation
+	if instance.Spec.CalicoNetwork != nil && (instance.Spec.CalicoNetwork.WindowsDataplane == nil || *instance.Spec.CalicoNetwork.WindowsDataplane == operatorv1.WindowsDataplaneDisabled) {
```
caseydavenport, line 490: I think this check should happen way earlier in this controller - there's no need to do all this other work before we get here.

Checking if Windows is enabled in the Installation, and whether there are even windows nodes in the cluster, should be one of the first things this controller does - if these checks come up "false", then we can simply short-circuit because there's nothing for this controller to do.

pkg/controller/installation/windows_controller.go
```diff
@@ -0,0 +1,597 @@
...
+				K8sDNSServers:           kubeDNSIPs,
+				Installation:            &instance.Spec,
+				ClusterDomain:           r.clusterDomain,
+				TLS:                     typhaNodeTLS,
+				AmazonCloudIntegration:  aci,
+				PrometheusServerTLS:     nodePrometheusTLS,
+				NodeReporterMetricsPort: nodeReporterMetricsPort,
+				VXLANVNI:                *felixConfiguration.Spec.VXLANVNI,
+			}
+			components = append(components, render.Windows(&windowsCfg))
```
caseydavenport, line 550: This controller ever only creates a single component - we don't need to have a slice and use append.

pkg/controller/installation/windows_controller.go
```diff
@@ -0,0 +1,597 @@
...
+	err = r.client.Get(ctx, types.NamespacedName{Name: "default"}, bgpConfiguration)
+	if err != nil && !apierrors.IsNotFound(err) {
+		r.status.SetDegraded(operatorv1.ResourceReadError, "Unable to read BGPConfiguration", err, reqLogger)
+		return reconcile.Result{}, err
+	}
+
+	var hasWindowsNodes bool
+	// Don't render calico-node-windows if it's disabled in the installation
+	if instance.Spec.CalicoNetwork != nil && (instance.Spec.CalicoNetwork.WindowsDataplane == nil || *instance.Spec.CalicoNetwork.WindowsDataplane == operatorv1.WindowsDataplaneDisabled) {
+		reqLogger.Info("Calico Windows daemonset is disabled in the operator installation")
```
caseydavenport, line 491: Instead of just logging, we should `return reconcile.Result{}, nil` - that way we simply stop execution of the controller, and we don't need an "else" branch to this "if"

pkg/controller/installation/windows_controller.go
```diff
@@ -0,0 +1,597 @@
...
+				if apierrors.IsNotFound(err) {
+					r.status.SetDegraded(operatorv1.ResourceNotFound, fmt.Sprintf("%s service not found", kubeDNSServiceName.Name), err, reqLogger)
+				} else {
+					r.status.SetDegraded(operatorv1.ResourceReadError, fmt.Sprintf("Error querying %s service", kubeDNSServiceName.Name), err, reqLogger)
+				}
+				return reconcile.Result{RequeueAfter: 10 * time.Second}, err
+			}
+			kubeDNSIPs := kubeDNSService.Spec.ClusterIPs
+
+			if felixConfiguration.Spec.VXLANVNI == nil {
```
caseydavenport, line 533: This is, at face value, a bit confusing since this field isn't usually set by default.

Worth adding a small comment saying that this field is always defaulted by the core controller.

pkg/controller/installation/windows_controller.go
```diff
@@ -0,0 +1,597 @@
...
+			IsController: true,
+			OwnerType:    &operator.Installation{},
+		}, pred)
+		if err != nil {
+			return fmt.Errorf("tigera-windows-controller failed to watch %s: %w", t, err)
+		}
+	}
+
+	// Watch for changes to KubeControllersConfiguration.
+	err = c.Watch(&source.Kind{Type: &crdv1.KubeControllersConfiguration{}}, &handler.EnqueueRequestForObject{})
```
caseydavenport, line 151: Does this controller actually care about this?

pkg/controller/installation/windows_controller.go
```diff
@@ -0,0 +1,597 @@
...
+		kubecontrollerprometheusTLS, err := certificateManager.GetCertificate(r.client, monitor.PrometheusClientTLSSecretName, common.OperatorNamespace())
+		if err != nil {
+			r.status.SetDegraded(operatorv1.ResourceReadError, "Failed to get certificate for kube controllers", err, reqLogger)
+			return reconcile.Result{}, err
+		} else if kubecontrollerprometheusTLS != nil {
+			typhaNodeTLS.TrustedBundle.AddCertificates(kubeControllerTLS, kubecontrollerprometheusTLS)
+		}
+	}
+
+	components := []render.Component{}
```
caseydavenport, line 478: per other comment, no need for a slcie here - there's only one component.

pkg/controller/installation/windows_controller.go
caseydavenport: In general, we want to be very careful about which controllers own certain pieces of configuration, and which controllers consume that configuration. We only want a single owner for each object, so if there's something being created by another controller we should be careful to only ever GET that object, and never CREATE it.

pkg/controller/installation/windows_controller.go
```diff
@@ -0,0 +1,597 @@
...
+			reqLogger.Info("Scheduling a retry in 30 seconds")
+			return reconcile.Result{RequeueAfter: 30 * time.Second}, nil
+		}
+		if amazonCRDRequired {
+			logw.Info("Rebooting to enable AWS controllers")
+			os.Exit(0)
+		}
+	}
+
+	certificateManager, err := certificatemanager.Create(r.client, &instance.Spec, r.clusterDomain, common.OperatorNamespace(), certificatemanager.AllowCACreation())
```
caseydavenport, line 379: We shouldn't have `AllowCACreation` - the cluster CA is created by the main secrets controller and all other controllers should just wait for it.

pkg/controller/installation/windows_controller.go
coutinhop: Right, I think this was left over from a previous iteration in which we wouldn't have explicit windows config and thus had to detect windows nodes... Will fix, thanks for catching it!

pkg/controller/installation/windows_controller.go
coutinhop: I came up with a fix but it seems to be a problem for tests here: https://github.com/tigera/operator/blob/c8e145ecb6a371f08b56427934d9b01f789d8314/pkg/controller/intrusiondetection/intrusiondetection_controller.go#L576
Working on that one too

caseydavenport: I think you don't need to change the DPI controller. Yes, it is a bug that it's also using GetOrCreate, but that's outside the scope of your changes I would say.

pkg/controller/installation/windows_controller.go
```diff
@@ -0,0 +1,455 @@
...
+			r.status.OnCRNotFound()
+			return reconcile.Result{}, nil
+		}
+		reqLogger.Error(err, "An error occurred when querying the Installation resource")
+		return reconcile.Result{}, err
+	}
+
+	// Don't render calico-node-windows if it's disabled in the installation
+	if !common.WindowsEnabled(instance.Spec) {
+		reqLogger.Info("Calico Windows daemonset is disabled in the operator installation")
```
caseydavenport, line 226: Should probably be `V(1).Info`

pkg/controller/installation/windows_controller.go
```diff
@@ -0,0 +1,455 @@
...
+	// FIXME: add logic to merge Installation status metadata
+
+	// Changes for updating Installation status conditions.
+	if request.Name == InstallationName && request.Namespace == "" {
+		ts := &operatorv1.TigeraStatus{}
+		err := r.client.Get(ctx, types.NamespacedName{Name: InstallationName}, ts)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+		instance.Status.Conditions = status.UpdateStatusCondition(instance.Status.Conditions, ts.Status.Conditions)
```
caseydavenport, line 241: I think we still need to get rid of this block for now - it's going to fight with the same code from core_controller.go

pkg/controller/installation/windows_controller.go
```diff
@@ -0,0 +1,455 @@
...
+		}
+	}
+
+	// A status is needed at this point for operator scorecard tests.
+	// status.variant is written later but for some tests the reconciliation
+	// does not get to that point.
+	if reflect.DeepEqual(instanceStatus, operatorv1.InstallationStatus{}) {
+		err := fmt.Errorf("InstallationStatus is empty")
+		r.status.SetDegraded(operator.ResourceUpdateError, "InstallationStatus is empty", err, reqLogger)
+		return reconcile.Result{RequeueAfter: 10 * time.Second}, err
```
caseydavenport, line 283: We shouldn't need to requeue explicitly here, since we'll get a Reconcile() trigger when the Installation status is updated.

pkg/controller/installation/windows_controller.go
```diff
@@ -0,0 +1,455 @@
...
+		reqLogger.V(2).Info("loaded final computed config", "config", instance)
+
+		// Validate the configuration.
+		if err := validateCustomResource(instance); err != nil {
+			r.status.SetDegraded(operatorv1.InvalidConfigurationError, "Invalid computed config", err, reqLogger)
+			return reconcile.Result{}, err
+		}
+	}
+
+	// A status is needed at this point for operator scorecard tests.
```
caseydavenport, line 277: This comment should say "We rely on the core controller for defaulting, so wait until it has done so before continuing"

pkg/controller/installation/windows_controller.go
```diff
@@ -0,0 +1,455 @@
...
+		}
+
+		// The key pair is created by the core controller, so if it isn't set, requeue to wait until it is
+		nodePrometheusTLS, err = certificateManager.GetKeyPair(r.client, render.NodePrometheusTLSServerSecret, common.OperatorNamespace(), dns.GetServiceDNSNames(render.CalicoNodeMetricsService, common.CalicoNamespace, r.clusterDomain))
+		if err != nil {
+			r.status.SetDegraded(operatorv1.ResourceCreateError, "Error getting TLS certificate", err, reqLogger)
+			return reconcile.Result{RequeueAfter: 10 * time.Second}, err
+		}
+		if nodePrometheusTLS != nil {
+			typhaNodeTLS.TrustedBundle.AddCertificates(nodePrometheusTLS)
```
caseydavenport, line 350: We shouldn't do any adding of certificates to the TrustedBundle, and we shouldn't create the TrustedBundle in this controller - this is another thing that the core controller is already doing, and doing it here as well will cause them to fight with each other.

We do need to pass it to the render function, but that's OK because it just uses it for volume mount determination.

pkg/controller/installation/windows_controller.go
```diff
@@ -0,0 +1,455 @@
...
+		kubeDNSIPs := kubeDNSService.Spec.ClusterIPs
+
+		// felixConfiguration.Spec.VXLANVNI is defaulted by fillDefaults() in the core controller, so if it isn't set, requeue to wait until it is
+		if felixConfiguration.Spec.VXLANVNI == nil {
+			err = fmt.Errorf("VXLANVNI not specified in FelixConfigurationSpec")
+			r.status.SetDegraded(operatorv1.ResourceReadError, "Error reading VXLANVNI from FelixConfiguration", err, reqLogger)
+			return reconcile.Result{RequeueAfter: 10 * time.Second}, err
+		}
+
+		windowsCfg := render.WindowsConfiguration{
```
caseydavenport, line 404: Does this belong within the `if instance.Spec.CalicoNetwork != nil {` block?

Generally I'd expect we want to render the component regardless - the render package can decide to return no objects, though.

caseydavenport (changes requested):

caseydavenport (approved):

PR #2738, opened 2023-07-14
Elasticsearch controller refactor

caseydavenport:
## Description

This PR contains a major refactor of the Elasticsearch controller within
the tigera-operator, as well as changes intended to lay the groundwork
for multiple tenant's sharing the same management cluster in a
multi-cluster management configuration.

This PR splits the monolithic LogStorage controller into the following independent
controllers:

- logstorage controller - performs defaulting / validation on the LogStorage resource.
- logstorage secrets controller - provisions keypairs, per-tenant CA, and trusted bundle.
- logstorage elastic controller - installs ECK and Elasticsearch itself.
- logstorage access controller - installs Linseed, es-gateway, es-kube-controllers.
- logstorage managedcluster controller - installs minimal logstorage components on managed clusters.

Notes and questions:

- some components might be in the wrong place. e.g., es-metrics - does that need its own controller?

New APIs:

- operator.tigera.io/Tenant - represents a Tenant in a multi-tenant system.

## TODO

- [ ] Tests for change.
- [ ] Clean up old es-kube-controllers deployment on upgrade (i.e., delete from calico-system)
- [ ] Lots of TODOs in the code

## For PR reviewers

- [ ] Appropriate labels:


config/enterprise_versions.yml
```diff
@@ -16,8 +16,9 @@ components:
     image: tigera/cnx-queryserver
     version: master
   cnx-kube-controllers:
-    image: tigera/kube-controllers
-    version: master
+    registry: gcr.io/unique-caldron-775/
```
caseydavenport, line 19: Need to revert this prior to merge, but have been testing with modified images.

pkg/controller/certificatemanager/certificatemanager.go
```diff
@@ -86,6 +87,11 @@ type CertificateManager interface {
 // Create creates a signer of new certificates and has methods to retrieve existing KeyPairs and Certificates. If a user
 // brings their own secrets, CertificateManager will preserve and return them.
 func Create(cli client.Client, installation *operatorv1.InstallationSpec, clusterDomain, ns string) (CertificateManager, error) {
+	return CreateWithLogger(cli, installation, clusterDomain, ns, log)
```
caseydavenport, line 90: This is a small change to allow controllers to pass their own logger when creating a certificate manager - it makes it easier in the logs to see which controller is triggering various certificate manager calls and was useful as part of my development on this PR.

I'm considering switching all controllers over to this as out-of-scope for now, though.

pkg/controller/installation/core_controller.go
```diff
@@ -1017,11 +1017,9 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 		}
 
 		managementCluster, err = utils.GetManagementCluster(ctx, r.client)
-		if managementCluster != nil {
```
caseydavenport, line 1020: I think this was just a bug - `managementCluster` would never be non-nil if `err` was also non-nil, so we were just swallowing errors from this function.

pkg/controller/logcollector/logcollector_controller.go
```diff
@@ -542,6 +551,13 @@ func (r *ReconcileLogCollector) Reconcile(ctx context.Context, request reconcile
 	// Create a component handler to manage the rendered component.
 	handler := utils.NewComponentHandler(log, r.client, r.scheme, instance)
 
+	// Determine the namespace in which Linseed is running. For managed and standalone clusters, this is always the elasticsearch
+	// namespace. For multi-tenant management clusters, this may vary.
+	linseedNS := render.ElasticsearchNamespace
+	if multiTenantManagement {
+		linseedNS = "tigera-tenant" // TODO: Hardcoding for testing.
```
caseydavenport, line 558: The idea here is that for management clusters that house multiple tenants, we want a tenant that represents the management cluster itself. The LogCollector should have a way of being told which tenant it should consider itself, and thus be configured to send logs to that tenant's Linseed.

pkg/controller/logstorage/common/common.go
```diff
@@ -46,8 +45,8 @@ const (
 // are generated and stored in the user secret, a hashed version of the credentials is stored in the tigera-elasticsearch namespace for ES Gateway to retrieve and use to compare
 // the gateway credentials, and a secret containing real admin level credentials is created and stored in the tigera-elasticsearch namespace to be swapped in once
 // ES Gateway has confirmed that the gateway credentials match.
-func CreateKubeControllersSecrets(ctx context.Context, esAdminUserSecret *corev1.Secret, esAdminUserName string, cli client.Client) (*corev1.Secret, *corev1.Secret, *corev1.Secret, error) {
```
caseydavenport, line 49: I think we ultimately want to get rid of this, and create secrets via the ECK API instead of having kube-controllers do it. Then, we can get rid of es-kube-controllers access to Elasticsearch entirely (and also get rid of es-gateway).

pkg/controller/logstorage/elastic/elastic_controller.go
```diff
@@ -0,0 +1,725 @@
...
+		return err
+	}
+
+	if err := r.client.Get(ctx, types.NamespacedName{Name: render.OIDCUsersEsSecreteName, Namespace: render.ElasticsearchNamespace}, &corev1.Secret{}); err != nil {
+		return err
+	}
+	return nil
+}
+
+// TODO: This shouldn't be done in the operator. Linseed should probably hanle this for single-tenant setups, and multi-tenant should
```
caseydavenport, line 656: For this PR, I have essentially commented this out for multi-tenant scenarios. I did this in order to work around the way credentials have changed in this PR - es-kube-controllers runs per-tenant, and installs secrets in per-tenant namespaces, and so right now the operator doesn't have credentials to talk to ES in this PR.

We have a few potential options for resolving this:

- Move operator secret provisioning out of es-kube-controllers, and into a cluster-scoped controller
- Make es-kube-controllers cluster-scoped instead of per-tenant
- Operator uses admin credentials for talking to ES, since it has those available already without any help.
- Remove ILM programming from the operator, move it to Linseed and/or another place.

I am partial to the latter.

pkg/controller/logstorage/linseed/esgateway.go
```diff
@@ -0,0 +1,132 @@
...
+	gatewayKeyPair, err := cm.GetKeyPair(r.client, render.TigeraElasticsearchGatewaySecret, request.TruthNamespace())
+	if err != nil {
+		r.status.SetDegraded(operatorv1.ResourceNotFound, "Error getting TLS certificate", err, log)
+		return err
+	} else if gatewayKeyPair == nil {
+		r.status.SetDegraded(operatorv1.ResourceNotFound, "es-gateway key pair not yet available", err, log)
+		return err
+	}
+
+	kubeControllersGatewaySecret, kubeControllersVerificationSecret, kubeControllersSecureUserSecret, err := lscommon.CreateKubeControllersSecrets(ctx, esAdminUserSecret, esAdminUserName, r.client, request)
```
caseydavenport, line 90: This is one of those dependencies that needs to be unpicked... 

The kube-controllers secrets are created here right now because kube-controllers talks via es-gateway in order to reach ES so that it can provision users. In order to do that, it needs these secrets to exist.

The dependency chain right now looks like this:

```
es --> esgw --> kube-controllers --> linseed
```

We could probably just move the generation of these secrets into the kube-controllers function instead though - the render component for ESGW is simply installing the secrets into the cluster, so not truly needed right here.

pkg/controller/logstorage/linseed/eskubecontrollers.go
```diff
@@ -29,37 +30,54 @@ import (
 	"github.com/tigera/operator/pkg/render"
 	"github.com/tigera/operator/pkg/render/kubecontrollers"
 	"github.com/tigera/operator/pkg/tls/certificatemanagement"
-	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"k8s.io/apimachinery/pkg/api/errors"
 )
 
-func (r *ReconcileLogStorage) createESKubeControllers(
+// TODO: Move this out of this controller.
```
caseydavenport, line 36: What else does es-kube-controllers do?

- Provisions users to ES, and generates usernames / passwords for accessing ES via ESGW
- Copies a public certificate to managed clusters so they can verify ESGW as part of the TLS handshake.
- ???

I think that means if ESGW gets removed, we can get rid of ESKC as well! But looking for input in case anyone is aware of other es-kube-controllers responsibilities...

pkg/controller/logstorage/linseed/eskubecontrollers.go
```diff
@@ -29,37 +30,54 @@ import (
...
 		svcDNSNames := append(dns.GetServiceDNSNames(render.ManagerServiceName, render.ManagerNamespace, r.clusterDomain), render.ManagerServiceIP)
 		managerInternalTLSSecret, err = certificateManager.GetOrCreateKeyPair(r.client, render.ManagerInternalTLSSecretName, common.CalicoNamespace, svcDNSNames)
 		if err != nil {
 			r.status.SetDegraded(operatorv1.ResourceValidationError, fmt.Sprintf("Error ensuring internal manager TLS certificate %q exists and has valid DNS names", render.ManagerInternalTLSSecretName), err, reqLogger)
-			return reconcile.Result{}, false, err
+			return err
+		}
+	}
+
+	namespaces := []string{request.InstallNamespace()}
```
caseydavenport, line 76: I'm not 100% on this approach yet, but it's a pattern I've been rolling out in this PR.

Some components have ClusterRoles that need to be bound to a namespaced service account. For single-tenant scenarios, this is simply - there's a one-to-one mapping.

For multi-tenant scenarios, we need to bind that ClusterRole to every serviceaccount for the service across all tenants. Essentially, we're sharing the ClusterRole and ClusterRoleBinding for all tenants (since the permissions are identical). 

This code determines which namespaces we need to bind to, and passes them to the render package. By default, we just bind to the namespace we're installing into. But for multi-tenant setups, we query all tenants and discover the full set of namespaces that need to be bound to.

pkg/controller/logstorage/linseed/eskubecontrollers.go
```diff
@@ -72,30 +90,32 @@ func (r *ReconcileLogStorage) createESKubeControllers(
 		Authentication:               authentication,
 		KubeControllersGatewaySecret: kubeControllersUserSecret,
 		LogStorageExists:             true,
-		TrustedBundle:                certificateManager.CreateTrustedBundle(),
+		TrustedBundle:                certificateManager.CreateTrustedBundle(), // We don't care about the contents.
+		Namespace:                    request.InstallNamespace(),               // TODO: This will give tigera-elasticsearch in single-tenant systems. Is this OK?
```
caseydavenport, line 94: Currently, es-kube-controllers runs in calico-system. I've always thought this was a little weird...

a side-effect of what I'm doing here is to run it in tigera-elasticsearch (for single-tenant clusters). I think this is OK? But would like input.

If we do manage to get rid of es-kube-controllers entirely for multi-tenant management clusters, we could more easily keep it in calico-system and just simply not install it for multi-tenant setups.

pkg/controller/logstorage/linseed/linseed_controller.go
```diff
@@ -0,0 +1,389 @@
...
+	c, err := controller.New("log-storage-access-controller", mgr, controller.Options{Reconciler: r})
+	if err != nil {
+		return err
+	}
+
+	// Determine how to handle watch events for cluster-scoped resources. For multi-tenant clusters,
+	// we should update all tenants whenever one changes. For single-tenatn clusters, we can just queue the object.
+	var eventHandler handler.EventHandler = &handler.EnqueueRequestForObject{}
+	if opts.MultiTenant {
+		eventHandler = utils.EnqueueAllTenants(mgr.GetClient())
```
caseydavenport, line 88: This PR introduces a new event handler type, that means "Trigger a reconcile for all tenants" - it's pretty convenient. I wonder what other simplifications we can make by developing some smarter EventHandler implementations... :thinking:

pkg/controller/logstorage/linseed/linseed_controller.go
```diff
@@ -0,0 +1,389 @@
...
+	if err = utils.AddTigeraStatusWatch(c, "log-storage-access"); err != nil {
+		return fmt.Errorf("logstorage-controller failed to watch logstorage Tigerastatus: %w", err)
+	}
+	if opts.MultiTenant {
+		if err = c.Watch(&source.Kind{Type: &operatorv1.Tenant{}}, eventHandler); err != nil {
+			return fmt.Errorf("log-storage-controller failed to watch Tenant resource: %w", err)
+		}
+	}
+
+	// The namespace(s) we need to monitor depend upon what tenancy mode we're running in.
```
caseydavenport, line 113: This pattern is pretty ugly - I suspect there is a better way. Maybe with a helper interface that hides the namespace selection logic somehow?

pkg/controller/logstorage/linseed/linseed_controller.go
```diff
@@ -0,0 +1,389 @@
...
+	reqLogger.Info("Reconciling LogStorage - Linseed")
+
+	// When running in multi-tenant mode, we need to install Linseed in tenant Namespaces. However, the LogStorage
+	// resource is still cluster-scoped (since ES is a cluster-wide resource), so we need to look elsewhere to determine
+	// which tenant namespaces require a Linseed instance. We use the tenant API to determine the set of namespaces that should have a Linseed.
+	var tenant *operatorv1.Tenant
+	var err error
+	if r.multiTenant {
+		if request.Namespace == "" {
+			// In multi-tenant mode, we only handle namespaced reconcile triggers.
```
caseydavenport, line 173: With the introduction of the EnqueueAllTenants EventHandler, I might be able to get rid of this branch actually. Need to look into that.

pkg/controller/logstorage/linseed/linseed_controller.go
```diff
@@ -0,0 +1,389 @@
...
+		pullSecrets,
+		hdler,
+		reqLogger,
+		r.usePSP,
+	); err != nil {
+		return reconcile.Result{}, err
+	}
+
+	// TODO: This is a temporary location for this. Ideally this goes in its own controller, as it likely
+	// needs to be multi-tenant!
```
caseydavenport, line 354: Update: ideally this isn't needed at all for multi-tenant clusters so it can just go into the single-tenant "elastic" controller or somewhere else.

pkg/controller/logstorage/secrets/secret_controller.go
```diff
@@ -0,0 +1,619 @@
...
+	r.status.Run(opts.ShutdownContext)
+
+	// Create a controller using the reconciler and register it with the manager to receive reconcile calls.
+	c, err := controller.New("log-storage-secrets-controller", mgr, controller.Options{Reconciler: r})
+	if err != nil {
+		return err
+	}
+
+	// Configure watches for operator.tigera.io APIs this controller cares about.
+	if err = c.Watch(&source.Kind{Type: &operatorv1.LogStorage{}}, &handler.EnqueueRequestForObject{}); err != nil {
```
caseydavenport, line 90: Should use `EnqueueAllTenants` here

pkg/controller/logstorage/secrets/secret_controller.go
```diff
@@ -0,0 +1,619 @@
...
+	truthNS := common.OperatorNamespace()
+	if opts.MultiTenant {
+		// For multi-tenant, the manager could be installed in any number of namespaces.
+		// So, we need to watch the resources we care about across all namespaces.
+		installNS = ""
+		truthNS = ""
+	}
+
+	// Watch all the elasticsearch user secrets in the operator namespace.
+	// TODO: In the future, we may want put this logic in the utils folder where the other watch logic is.
```
caseydavenport, line 122: This TODO pre-dates my changes. but I will probably do it in this PR anyway. Not sure why this wasn't done in the first place.

pkg/controller/logstorage/secrets/secret_controller.go
```diff
@@ -0,0 +1,619 @@
...
+			// No tenant in this namespace. Ignore the update.
+			reqLogger.Info("No Tenant in this Namespace, skip")
+			return reconcile.Result{}, nil
+		} else if err != nil {
+			r.status.SetDegraded(operatorv1.ResourceReadError, "An error occurred while querying Tenant", err, reqLogger)
+			return reconcile.Result{}, err
+		}
+	}
+
+	// Generate keys for Tigera components if we're running in multi-tenant mode and this is a reconcile
```
caseydavenport, line 218: This controller handles both cluster-scoped and tenant-scoped secrets at the moment. This is potentially a little bit confusing, so this might be a good place to even further decompose - e.g., a cluster-scoped secret controller and a tenant-scoped secret controller.

pkg/controller/logstorage/secrets/secret_controller.go
```diff
@@ -0,0 +1,619 @@
...
+			log.Info(msg)
+		} else {
+			log.V(2).Info("Adding upstream certificate", "cert", certName)
+			collection.upstreamCerts = append(collection.upstreamCerts, cert)
+		}
+	}
+
+	// Add the certificate manager's keypair. This ensures that the secret used to sign certificates and keys
+	// in this tenant's namespace is saved for other controllers to use.
+	// TODO: This might not be the right place for this.
```
caseydavenport, line 512: Right now, for single-tenant clusters, the root CA is installed into the tigera-operator namespace by the core controller. However, the core controller is single-tenant, and so we need to do this somewhere else for multi-tenant systems, and ideally wherever we do it also works for single-tenant systems so we're not duplicating logic.

I am thinking this might just be a new "Root CA controller" (it may be obvious that I am on a 'write new controllers' kick right now).

pkg/controller/manager/manager_controller.go
```diff
@@ -641,13 +648,6 @@ func (r *ReconcileManager) reconcileInstance(ctx context.Context, logc logr.Logg
 			TruthNamespace:  request.TruthNamespace(),
 			ServiceAccounts: []string{render.ManagerServiceAccount},
 			KeyPairOptions: []rcertificatemanagement.KeyPairOption{
-				// We need to render the certificate manager CA cert again in case we're running in multi-tenant mode.
-				// For single-tenant, this cert is created once by the core controller. For multi-tenant, we
-				// provision a unique CA per-tenant, and so we need to make sure to create it here.
-				//
-				// TODO: We probably want a separate tenant controller managing the creation of of this instead, so
```
caseydavenport, line 648: This was a hack I introduced in my last prototype PR that I am now replace with another hack in this PR (in the secrets_controller).

I think this probably belongs in a "Root CA controller" instead of either of these places.

pkg/controller/utils/utils.go
```diff
@@ -549,12 +582,17 @@ func createPredicateForObject(objMeta metav1.Object) predicate.Predicate {
...
+				// No name or namespace match was specified. Match everything.
 				return true
 			}
 			if objMeta.GetName() != "" && e.Object.GetName() != objMeta.GetName() {
+				// A name match was specified, and the object doesn't match.
 				return false
 			}
-			return e.Object.GetNamespace() == objMeta.GetNamespace()
+
+			// A name match was specified and the name matches, or this is just a namespace match.
```
caseydavenport, line 593: Updated this so that if the caller doesn't pass a namespace, it performs a cluster-scoped watch.

pkg/render/common/networkpolicy/networkpolicy.go
```diff
@@ -177,21 +181,51 @@ var KubeAPIServerEntityRule = v3.EntityRule{
...
 		Namespace: "default",
 		Name:      "kubernetes",
 	},
 }
 
-// The entity rules below are extracted from render subpackages to prevent cyclic dependencies.
-var ESGatewayEntityRule = CreateEntityRule("tigera-elasticsearch", "tigera-secure-es-gateway", 5554)
-var ESGatewaySourceEntityRule = CreateSourceEntityRule("tigera-elasticsearch", "tigera-secure-es-gateway")
-var ESGatewayServiceSelectorEntityRule = CreateServiceSelectorEntityRule("tigera-elasticsearch", "tigera-secure-es-gateway-http")
+func Helper(mt bool, ns string) *NetworkPolicyHelper {
```
caseydavenport, line 192: This is meant to make it easier to write policy rules that work in both single-tenant and multi-tenant setups.

The main difference is that the namespace selector in these rules needs to match either the "default" single-tenant namespace (i.e., how it is today) or the dynamically configured tenant namespace.

pkg/render/render.go
```diff
@@ -47,5 +47,5 @@ func SetClusterCriticalPod(t *corev1.PodTemplateSpec) {
 
 // ImagePullPolicy returns the image pull policy to use for all components.
 func ImagePullPolicy() corev1.PullPolicy {
-	return corev1.PullIfNotPresent
+	return corev1.PullAlways
```
caseydavenport, line 50: For testing purposes - need to revert prior to merge.

Brian-McM:
> es-kube-controllers has moved from calico-system to tigera-elasticsearch in single-tenant setups

The reason es kube controllers is in the calico system namespace is so that it would have permissions in the managed clusters, as it uses the regular kube controllers service account.

If we're to separate this out, you need to make sure the managed clusters have the SA for es kubecontrollers / RBAC, which would require an upgrade to the managed clusters.

pkg/controller/logstorage/linseed/esgateway.go
```diff
@@ -0,0 +1,132 @@
...
+	variant operatorv1.ProductVariant,
+	pullSecrets []*corev1.Secret,
+	hdler utils.ComponentHandler,
+	reqLogger logr.Logger,
+	usePSP bool,
+) error {
+	// Create a trusted bundle to pass to the render pacakge. The actual contents of this bundle don't matter - the ConfigMap
+	// itself will be managed by the Secret controller. But, we need an interface to use as an argument to render in order
+	// to configure volume mounts properly.
+	trustedBundle := certificatemanagement.CreateTrustedBundle()
```
caseydavenport, line 49: As it turns out, this comment is completely false - we need the _real_ trusted bundle here in order to properly populate the hash annotations on the objects that depends on this bundle.

The problem is that the bundle is created in the secrets_controller, and so this controller doesn't have access to it. 

This is further exacerbated by the fact that the trusted bundle that is written to the cluster has lost key information by the time it is stored in a ConfigMap - for example, which namespace the certificates in the bundle originated. This makes it impossible for other controllers to reach the bundle and reverse engineer it by re-loading those certificates.

I think the solution to this is for the secrets controller to write the trusted bundle with the hash annotations pre-calcualted and included in the annotations on the configmap.

Then, controllers can simply load that configmap and copy the annotations straight from the secret without ever needing to deal with the contents of the configmap itself. Would be interested in feedback on this before I implement it - probably one for @rene-dekker

caseydavenport: Render does need more than just the annotations - it also needs the volume mount path, and the volumes.

These can _almost_ be hard-coded, with the exception of needing to know whether or not the system certs need to be mounted.

caseydavenport: My latest commit includes a very quick and dirty example of what this might look like: https://github.com/tigera/operator/pull/2738/commits/c3f76da9855a853c89d0adf667b3c4347f8d2da9

pkg/controller/certificatemanager/certificatemanager.go
```diff
@@ -365,3 +437,47 @@ func (cm *certificateManager) CreateTrustedBundle(certificates ...certificateman
 func (cm *certificateManager) CreateTrustedBundleWithSystemRootCertificates(certificates ...certificatemanagement.CertificateInterface) (certificatemanagement.TrustedBundle, error) {
 	return certificatemanagement.CreateTrustedBundleWithSystemRootCertificates(append([]certificatemanagement.CertificateInterface{cm.keyPair}, certificates...)...)
 }
+
+type annotationPassthru struct {
```
caseydavenport, line 441: Need to clean this struct up and comment it better - it's an implementation of `TrustedBundleRO` that can be used to load an existing bundle for read-only operations (which is what most controllers need). 

This split allows us to have one controller provisioning bundles, and another loading them in order to pass to render so that they can properly get annotations / volumes / etc without any chance of actually modifying the bundle.

pkg/controller/installation/core_controller.go
```diff
@@ -1054,17 +1052,21 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
...
+	// in management clusters. kube-controllers copies the es-gateway and linseed certificates to managed clusters.
+	// TODO: We should ideally move the copying of these secrets elsehwere, since kube-controllers isn't installed in multi-tenant
+	// clusters right now. Without something doing this, fluentd in managed clusters won't be able to verify Voltron.
 	var managerInternalTLSSecret certificatemanagement.KeyPairInterface
 	if instance.Spec.Variant == operator.TigeraSecureEnterprise && managementCluster != nil {
-		dnsNames := append(dns.GetServiceDNSNames(render.ManagerServiceName, render.ManagerNamespace, r.clusterDomain), render.ManagerServiceIP)
-		managerInternalTLSSecret, err = certificateManager.GetOrCreateKeyPair(r.client, render.ManagerInternalTLSSecretName, common.OperatorNamespace(), dnsNames)
-		if err != nil {
-			r.status.SetDegraded(operator.CertificateError, fmt.Sprintf("Error ensuring internal manager TLS certificate %q exists and has valid DNS names", render.ManagerInternalTLSSecretName), err, reqLogger)
+		managerInternalTLSSecret, err = certificateManager.GetKeyPair(r.client, render.ManagerInternalTLSSecretName, common.OperatorNamespace())
```
caseydavenport, line 1067: This keypair is now created by the manager controller - seemed weird that it was being created here. The logic for installing manager was thus leaking into other unrelated controllers.

pkg/controller/logcollector/logcollector_controller.go
```diff
@@ -405,36 +409,40 @@ func (r *ReconcileLogCollector) Reconcile(ctx context.Context, request reconcile
...
-		r.status.SetDegraded(operatorv1.ResourceNotReady, "Linseed certificate are not available yet, waiting until they become available", nil, reqLogger)
+	} else if linseedCertificate == nil {
+		msg := fmt.Sprintf("Linseed certificate (%s/%s) is not yet available", linseedCertNamespace, linseedCertName)
+		log.Info(msg)
+		r.status.SetDegraded(operatorv1.ResourceNotReady, msg, nil, reqLogger)
 		return reconcile.Result{}, nil
 	}
 
 	// Fluentd needs to mount system certificates in the case where Splunk, Syslog or AWS are used.
-	trustedBundle, err := certificateManager.CreateTrustedBundleWithSystemRootCertificates(prometheusCertificate, esgwCertificate, linseedCertificate)
```
caseydavenport, line 437: logcollector is single tenant for both single-tenant mgmt clusters as well as multi-tenant mgmt clusters, which is why I think we can leave the secret and bundle generation within this controller.

The main thing is that some services which run differently in multi-tenant vs. single-tenant clusters have very different logic for how the trusted bundle is managed - those should be factored out into a separate bundle controller.

pkg/controller/logstorage/common/common.go
caseydavenport: For now I have actually skipped this for multi-tenant, so I could theoretically revert the changes to this function and only call it for single-tenant clusters.

pkg/controller/logstorage/elastic/elastic_controller.go
```diff
@@ -0,0 +1,723 @@
...
+	"github.com/tigera/operator/pkg/render/logstorage/linseed"
+	"github.com/tigera/operator/pkg/render/monitor"
+	"github.com/tigera/operator/pkg/tls/certificatemanagement"
+)
+
+var log = logf.Log.WithName("controller_logstorage_elastic")
+
+// ElasticSubController is a sub-controller of the main LogStorage controller
+// responsible for managing the Elasticsearch service used by Calico.
+type ElasticSubController struct {
```
caseydavenport, line 71: Will probably name this "InternalElasticController" when we introduce external ES - I think it makes sense for those to be separate controllers altogether that act in a mutually exclusive way.

pkg/controller/logstorage/linseed/esgateway.go
```diff
@@ -0,0 +1,128 @@
...
+	lscommon "github.com/tigera/operator/pkg/controller/logstorage/common"
+	"github.com/tigera/operator/pkg/controller/utils"
+	"github.com/tigera/operator/pkg/controller/utils/imageset"
+	"github.com/tigera/operator/pkg/render"
+	"github.com/tigera/operator/pkg/render/common/secret"
+	"github.com/tigera/operator/pkg/render/logstorage/esgateway"
+	"github.com/tigera/operator/pkg/tls/certificatemanagement"
+)
+
+func (r *LinseedSubController) createESGateway(
```
caseydavenport, line 36: For now I've left this as a tag-along in the linseed controller... they do similar stuff / fill a similar role in the system, but EGW is on the way out and is just kept for legacy purposes. 

Potentially could split this out, though.

pkg/controller/logstorage/linseed/linseed_controller.go
```diff
@@ -0,0 +1,413 @@
...
+	opts := []certificatemanager.Option{
+		certificatemanager.WithLogger(reqLogger),
+		certificatemanager.WithTenant(args.Tenant),
+	}
+	cm, err := certificatemanager.CreateWithOptions(r.client, install, r.clusterDomain, request.TruthNamespace(), opts...)
+	if err != nil {
+		r.status.SetDegraded(operatorv1.ResourceCreateError, "Unable to create the Tigera CA", err, reqLogger)
+		return reconcile.Result{}, err
+	}
+	linseedKeyPair, err := cm.GetKeyPair(r.client, render.TigeraLinseedSecret, request.InstallNamespace())
```
caseydavenport, line 278: I think this should be TruthNamespace

pkg/controller/logstorage/linseed/linseed_controller.go
```diff
@@ -0,0 +1,413 @@
...
+		hdler = utils.NewComponentHandler(reqLogger, r.client, r.scheme, args.Tenant)
+	} else {
+		hdler = utils.NewComponentHandler(reqLogger, r.client, r.scheme, args.LogStorage)
+	}
+	if err := hdler.CreateOrUpdateOrDelete(ctx, linseedComponent, r.status); err != nil {
+		r.status.SetDegraded(operatorv1.ResourceUpdateError, "Error creating / updating / deleting resource", err, reqLogger)
+		return reconcile.Result{}, err
+	}
+
+	if !r.multiTenant {
```
caseydavenport, line 361: These should probably all just be in another controller. Clean separation of controllers that need to handle multi-tenancy, and those that don't.

caseydavenport: Assuming it ends up that we don't need these in multi-tenant mgmt clusters.

pkg/controller/logstorage/users/user_controller.go
```diff
@@ -0,0 +1,261 @@
...
+	reqLogger = reqLogger.WithValues("installNS", req.InstallNamespace(), "truthNS", req.TruthNamespace())
+	return r.reconcile(ctx, reqLogger, args, req)
+}
+
+type ReconcileArgs struct {
+	LogStorage *operatorv1.LogStorage
+	Tenant     *operatorv1.Tenant
+}
+
+func (r *UserController) reconcile(ctx context.Context, reqLogger logr.Logger, args ReconcileArgs, request octrl.Request) (reconcile.Result, error) {
```
caseydavenport, line 159: I started this pattern of splitting `Reconcile` and `reconcile` early on because I thought I might need it but think it's not actually providing much, so might undo this.

pkg/controller/logstorage/users/user_controller.go
```diff
@@ -0,0 +1,261 @@
...
+	if args.Tenant != nil {
+		tenantID = args.Tenant.Spec.ID
+		if tenantID == "" {
+			r.status.SetDegraded(operatorv1.ResourceNotReady, "Tenant resource does not specify an ID", nil, reqLogger)
+			return reconcile.Result{}, nil
+		}
+	}
+	linseedUser := utils.LinseedUser(tenantID)
+
+	// Query any existing username and password for this Linseed instance. If one already exists, we'll simply
```
caseydavenport, line 176: I think we want a way to specify manually provided user credentials so that they can be provisioned out of band.

we'll also want a way for this controller to clean up credentials that are no longer needed, probably.

pkg/controller/logstorage/linseed/esgateway.go
Brian-McM: Is this still an accurate comment? I think you've made a comment somewhere else suggesting that we could do away with kube controllers / es gateway (in which case let's just leave this messy for now).

I'd be down for either using the ECK operator to create the necessary creds for linseed, or if that wasn't possible, just having the operator create the Elasticsearch credentials directly (and still removing needing kube controllers).

pkg/controller/logstorage/linseed/esgateway.go
```diff
@@ -0,0 +1,132 @@
...
+
+	esGatewayComponent := esgateway.EsGateway(cfg)
+	if err = imageset.ApplyImageSet(ctx, r.client, variant, esGatewayComponent); err != nil {
+		r.status.SetDegraded(operatorv1.ResourceUpdateError, "Error with images from ImageSet", err, reqLogger)
+		return err
+	}
+
+	// Copy the es admin user secret to the namespace so that es-gateway has access to it. For single-tenant, this is essentially a no-op.
+	// For multi-tenant, this means copying the es-gateway admin secret to the tenant's namespace.
+	// TODO: We shouldn't use the admin secret here, this is just for prototyping purposes. Instead, each tenant's esgw should get its own credentials.
```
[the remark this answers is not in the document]
Brian-McM: I like this idea, even though it might be a little annoying that we have this divergence between EE and Cloud operators. But we can just suck it up for a few releases.

pkg/controller/logstorage/linseed/eskubecontrollers.go
Brian-McM: > Copies a public certificate to managed clusters so they can verify ESGW as part of the TLS handshake.

The controller you created for linseed would replace this logic once we remove ESGW, right?

I think that kube controllers also passes a config map containing logstorage configuration needed in the managed cluster. It modifies (or adds, can't remember if it exists before it's modified) the cluster name so the managed cluster knows what "midfix" it should use for flow logs.

Look [here](https://github.com/tigera/calico-private/blob/master/kube-controllers/pkg/controllers/elasticsearchconfiguration/reconcile_elasticsearch_configuration.go#L90) and [here](https://github.com/tigera/calico-private/blob/master/kube-controllers/pkg/controllers/elasticsearchconfiguration/commonbuild.go).

pkg/controller/logstorage/linseed/eskubecontrollers.go
Brian-McM: I'm not sure if you saw my response on the PR comment section, but I posted this:

> The reason es kube controllers is in the calico system namespace is so that it would have permissions in the managed clusters, as it uses the regular kube controllers service account.
>
> If we're to separate this out, you need to make sure the managed clusters have the SA for es kubecontrollers / RBAC, which would require an upgrade to the managed clusters.

pkg/controller/logstorage/users/user_controller.go
```diff
@@ -0,0 +1,261 @@
...
+	"github.com/tigera/operator/pkg/controller/status"
+	"github.com/tigera/operator/pkg/controller/utils"
+	"github.com/tigera/operator/pkg/render"
+	relasticsearch "github.com/tigera/operator/pkg/render/common/elasticsearch"
+	"github.com/tigera/operator/pkg/render/common/secret"
+)
+
+var log = logf.Log.WithName("controller_logstorage_users")
+
+type UserController struct {
```
Brian-McM, line 50: This is only creating linseed users, right?

If so, it might make sense to put this in the linseed controller folder so this is clear.

Brian-McM: oops never mind, that's not what's happeing.

pkg/controller/installation/core_controller.go
Brian-McM: I don't think this kube controllers actually needs the manager tls cert. The regular kube controllers created by `kubecontrollers.NewCalicoKubeControllers` doesn't have the managedcluster controller enabled. That's in the es-kube-controllers.

This might be legacy code that was never cleaned up when we split out kube controllers, since none of the enabled controllers aren't for MCM. Might make your life a little easier if this is true...

pkg/controller/logstorage/linseed/eskubecontrollers.go
caseydavenport: > It modifies (or adds, can't remember if it exists before it's modified) the cluster name so the managed cluster knows what "midfix" it should use for flow logs.

This is no longer needed with Linseed.

pkg/controller/installation/core_controller.go
caseydavenport: Yeah I think I agree with you. That's nice!

pkg/crds/operator/operator.tigera.io_tenants.yaml
```diff
@@ -0,0 +1,45 @@
...
+            description: 'Kind is a string value representing the REST resource this
+              object represents. Servers may infer this from the endpoint the client
+              submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
+            type: string
+          metadata:
+            type: object
+          spec:
+            properties:
+              id:
+                description: ID is the unique identifier for this tenant.
```
Josh-Tigera, line 36: Do we need a validator somewhere for these unique identifiers? Perhaps a test case that can check if someone tries to create a tenant with the same ID as an already existing one?

pkg/controller/namespace_helper.go
```diff
@@ -0,0 +1,92 @@
...
+type NamespaceHelper interface {
+	// InstallNamespace returns the namespace that components will be installed into.
+	// for single-tenant clusters, this is tigera-manager. For multi-tenancy, this
+	// will be the tenant's namespace.
+	InstallNamespace() string
+
+	// TruthNamespace returns the namespace to use as the source of truth for storing data.
+	// For single-tenant installs, this is the tigera-operator namespace.
+	// For multi-tenant installs, this is tenant's namespace.
+	TruthNamespace() string
```
rene-dekker, line 49: I'm not a big fan of the name, but Im not sure if I know a better option.

To me the truth namespace is not so much about being the truth, but more about the namespace where users can provide input. So perhaps userInputNamespace, configurationNamespace or inputNamespace.

Perhaps the installNamespace could be called componentNamespace or managerNamespace(). The installation CR is "installing" in the calico-system namespace, making installNamespace all the more confusing.

pkg/controller/utils/utils.go
```diff
@@ -381,6 +417,43 @@ func GetAuthentication(ctx context.Context, cli client.Client) (*operatorv1.Auth
 	return authentication, nil
 }
 
+// Get the Tenant instance in the given namespace.
```
rene-dekker, line 420: The comments in this file do not match the pattern and give warnings in my IDE.

pkg/controller/logstorage/elastic/elastic_controller.go
```diff
@@ -0,0 +1,710 @@
...
+	}
+
+	// Watch all the elasticsearch user secrets in the operator namespace. In the future, we may want put this logic in
+	// the utils folder where the other watch logic is.
+	if err = utils.AddSecretWatchWithLabel(c, common.OperatorNamespace(), logstoragecommon.TigeraElasticsearchUserSecretLabel); err != nil {
+		return fmt.Errorf("log-storage-controller failed to watch Secret resource: %w", err)
+	}
+
+	// Watch all the secrets created by this controller so we can regenerate any that are deleted
+	for _, secretName := range []string{
```
rene-dekker, line 198: There are secretwatches that can be added to this loop.

CASecretName is missing.

pkg/controller/logstorage/elastic/elastic_controller.go
```diff
@@ -0,0 +1,710 @@
...
+		}
+		ls = nil
+		r.status.OnCRNotFound()
+	} else {
+		// We found the LogStorage instance.
+		r.status.OnCRFound()
+	}
+
+	// Wait for the initializing controller to indicate that the LogStorage object is actionable.
+	if ls == nil || ls.Status.State != operatorv1.TigeraStatusReady {
```
rene-dekker, line 285: For managed clusters (that have no logstorage), would this message be misleading? And probably we would not want to requeue

pkg/controller/logstorage/elastic/elastic_controller.go
```diff
@@ -0,0 +1,710 @@
...
+	}
+
+	// Query the trusted bundle from the namespace.
+	trustedBundle, err := cm.LoadTrustedBundle(ctx, r.client, render.ElasticsearchNamespace)
+	if err != nil {
+		r.status.SetDegraded(operatorv1.ResourceReadError, "Error getting trusted bundle", err, reqLogger)
+		return reconcile.Result{}, err
+	}
+
+	// If this is a Managed cluster ls must be nil to get to this point (unless the DeletionTimestamp is set) so we must
```
rene-dekker, line 534: We won't get to this point . See comment ^

pkg/controller/logstorage/initializer/initializing_controller.go
```diff
@@ -0,0 +1,304 @@
...
+	r.status.Run(opts.ShutdownContext)
+
+	// Create a controller using the reconciler and register it with the manager to receive reconcile calls.
+	c, err := controller.New("log-storage-controller", mgr, controller.Options{Reconciler: r})
+	if err != nil {
+		return err
+	}
+
+	// Configure watches for operator.tigera.io APIs this controller cares about.
+	if err = c.Watch(&source.Kind{Type: &operatorv1.LogStorage{}}, &handler.EnqueueRequestForObject{}); err != nil {
```
rene-dekker, line 77: I think this controller is also dependent on Installation{} and should watch it.

pkg/controller/logstorage/linseed/esgateway.go
rene-dekker: Ugh, the bundle for this controller has continuously caused issues. Glad you are getting to the root of this.

pkg/controller/namespace_helper.go
```diff
@@ -0,0 +1,92 @@
...
+
+	// BothNamespaces returns both the truth namespace and the install namespace.
+	BothNamespaces() []string
```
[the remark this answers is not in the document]
caseydavenport: This interface is actually used quite a bit in this PR - curious what your objections are to it?

pkg/controller/namespace_helper.go
caseydavenport: It's not really about whether the config was done by the user or not, it's about whether the config is the canonical representation or if it has been copied from the canonical representation.

We could call it the "SpecNamespace" to be more consistent with Kubernetes terms - e.g., spec vs. status. - but I think it's less clear than "TruthNamespace", which itself is less clear (and less verbose) than "SourceOfTruthNamespace"

pkg/controller/logstorage/linseed/esgateway.go
caseydavenport: Yeah, this has been updated now so that the operator creates the credentials for Linseed directly and there's no need for kube-controllers / esgw in multi-tenant mgmt clusters any more.

es-kube-controllers still functions as it always has for single-tenant, though.

pkg/controller/logstorage/linseed/linseed_controller.go
```diff
@@ -0,0 +1,413 @@
...
+				return fmt.Errorf("log-storage-controller failed to watch Secret: %w", err)
+			}
+		}
+	}
+
+	// Catch if something modifies the resources that this controller consumes.
+	if err := utils.AddServiceWatch(c, render.ElasticsearchServiceName, installNS); err != nil {
+		return fmt.Errorf("log-storage-controller failed to watch the Service resource: %w", err)
+	}
+	if err := utils.AddConfigMapWatch(c, certificatemanagement.TrustedCertConfigMapName, installNS, &handler.EnqueueRequestForObject{}); err != nil {
```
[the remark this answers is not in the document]
caseydavenport: The trusted cert configmap is installed into the application namespace, and so will always be a per-tenant resource. That means we can just trigger the one tenant that the update is relevant for

pkg/crds/operator/operator.tigera.io_tenants.yaml
caseydavenport: Yeah that's a good call - we should make sure there are no duplicate IDs.

pkg/controller/logstorage/elastic/elastic_controller.go
caseydavenport: @rene-dekker yeah, this is definitely a bit funky.

I am wondering if we should decouple the management / managed implementation so that this controller can confidently short-circuit if there is no LogStorage. That seems like the cleanest implementation to me, and is less likely to result in accidentally writing bugs into this controller.

pkg/controller/certificatemanager/certificatemanager.go
```diff
@@ -81,11 +90,36 @@ type CertificateManager interface {
...
+	return func(cm *certificateManager) error {
+		cm.tenant = t
+		return nil
+	}
 }
 
 // Create creates a signer of new certificates and has methods to retrieve existing KeyPairs and Certificates. If a user
 // brings their own secrets, CertificateManager will preserve and return them.
-func Create(cli client.Client, installation *operatorv1.InstallationSpec, clusterDomain, ns string) (CertificateManager, error) {
+func Create(cli client.Client, installation *operatorv1.InstallationSpec, clusterDomain, ns string, opts ...Option) (CertificateManager, error) {
```
Brian-McM, line 122: Do you know where the `clusterDomain` is used? I see it set to an exported variable, but I can't find where it's actually used.

caseydavenport: Hm, yeah it seems like that variable is never used.

pkg/controller/secrets/cluster_ca_controller.go
```diff
@@ -0,0 +1,111 @@
...
+		return reconcile.Result{}, err
+	}
+
+	// Create the cluster CA. This is done implicitly by initializing a certificate manager instance
+	// and passing the "AllowCACreation" option.
+	opts := []certificatemanager.Option{
+		certificatemanager.AllowCACreation(),
+		certificatemanager.WithLogger(logc),
+	}
+	cm, err := certificatemanager.Create(r.client, &instance.Spec, r.clusterDomain, common.OperatorNamespace(), opts...)
```
Brian-McM, line 94: Could you add a description on what this CA is used for, just to help with understanding. I'm sort of guessing, but it's used for non tenant specific communication in the management cluster, is that right? Any component that's tenanted would use the CA created by the tenant controller, is that correct?

pkg/controller/certificatemanager/certificatemanager.go
Brian-McM: @rene-dekker can you remember what the intention for that variable was?

pkg/controller/certificatemanager/certificatemanager.go
```diff
@@ -81,11 +90,36 @@ type CertificateManager interface {
 	AddToStatusManager(manager status.StatusManager, namespace string)
 	// KeyPair Returns the CA KeyPairInterface, so it can be rendered in the operator namespace.
 	KeyPair() certificatemanagement.KeyPairInterface
+	// Loads an existing trusted bundle to pass to render.
+	LoadTrustedBundle(context.Context, client.Client, string) (certificatemanagement.TrustedBundleRO, error)
+}
+
+type Option func(cm *certificateManager) error
+
+func AllowCACreation() Option {
```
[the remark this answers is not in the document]
caseydavenport: The CA is used by a number of callers for Get() operations to make sure the retrieved cert is signed by the expected authority. I think it makes sense for this to be an option.

> Would almost seem like anything that shouldn't create it shouldn't call this function (they really want Get).

Yeah, we could have a Create() and a Get() call that effectively do this same thing :shrug:

pkg/controller/certificatemanager/certificatemanager.go
```diff
@@ -365,3 +435,53 @@ func (cm *certificateManager) CreateTrustedBundle(certificates ...certificateman
...
 }
+
+func (cm *certificateManager) LoadTrustedBundle(ctx context.Context, client client.Client, ns string) (certificatemanagement.TrustedBundleRO, error) {
+	obj := &corev1.ConfigMap{}
+	k := types.NamespacedName{Name: certificatemanagement.TrustedCertConfigMapName, Namespace: ns}
+	if err := client.Get(ctx, k, obj); err != nil {
+		return nil, err
+	}
+	a := newReadOnlyTrustedBundle(cm, len(obj.Data[certificatemanagement.RHELRootCertificateBundleName]) > 0)
+	for key, val := range obj.Annotations {
```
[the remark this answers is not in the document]
caseydavenport: The reason this structure was introduced was because it's impossible (or really hard, at least) to re-build a `TrustedBundle` interface from the ConfigMap loaded into the cluster. The fact that it's RO isn't the goal of this interface, just a (useful) consequence of needing to be able to load metadata from an existing bundle configmap. 

The annotations on the ConfigMap are a new change I added in this PR in order to enable this particular struct - they should be exactly the same as what would otherwise be created, assuming I implemented this correctly.

pkg/controller/certificatemanager/certificatemanager.go
rene-dekker: I think in my initial implementation the manager was trying to compute the dns names for the leaf certs until I realized that this could not be done.

PR #2739, opened 2023-07-14
Use internal manager tls certificate for es-proxy and linseed

asincu:
communication

## Description

Manager-TLS is configured to also have K8S tigera-manager services as a SAN.

```
X509v3 Subject Alternative Name: 
                DNS:localhost, DNS:tigera-manager, DNS:tigera-manager.tigera-manager, DNS:tigera-manager.tigera-manager.svc, DNS:tigera-manager.tigera-manager.svc.cluster.local
    Signature Algorithm: sha256WithRSAEncryption
```

When users provide their own certificates, they do not have the K8S service declared, which causes errors for communication between Linseed and Es proxy.

This PR proposes that we use the internal manager tls that was created to be used for inter cluster communication (es-proxy and voltron) in MCM use case.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/manager/manager_controller.go
+++ b/pkg/controller/manager/manager_controller.go
@@ -444,9 +444,22 @@ func (r *ReconcileManager) Reconcile(ctx context.Context, request reconcile.Requ
 		return reconcile.Result{}, err
 	}
 
-	var tunnelSecret certificatemanagement.KeyPairInterface
-	var internalTrafficSecret certificatemanagement.KeyPairInterface
+	// We expect that the secret that holds the certificates for internal communication within the management
+	// K8S cluster is already created by the KubeControllers
+	internalTrafficSecret, err := certificateManager.GetKeyPair(r.client, render.ManagerInternalTLSSecretName, common.OperatorNamespace())
+	if internalTrafficSecret == nil {
+		r.status.SetDegraded(operatorv1.ResourceNotReady, fmt.Sprintf("Waiting for secret %s in namespace %s to be available", render.ManagerInternalTLSSecretName, common.OperatorNamespace()), nil, reqLogger)
+		return reconcile.Result{}, err
+	} else if err != nil {
+		r.status.SetDegraded(operatorv1.ResourceReadError, fmt.Sprintf("Error fetching TLS secret %s in namespace %s", render.ManagerInternalTLSSecretName, common.OperatorNamespace()), err, reqLogger)
+		return reconcile.Result{}, nil
+	}
+	// Es-proxy needs to trust Voltron for cross-cluster requests.
+	trustedBundle.AddCertificates(internalTrafficSecret)
+
 	var linseedVoltronSecret certificatemanagement.KeyPairInterface
+	var tunnelSecret certificatemanagement.KeyPairInterface
+
 	if managementCluster != nil {
 		preDefaultPatchFrom := client.MergeFrom(managementCluster.DeepCopy())
 		fillDefaults(managementCluster)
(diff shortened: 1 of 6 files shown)


pkg/render/manager.go
```diff
@@ -516,9 +516,9 @@ func (c *managerComponent) voltronContainer() corev1.Container {
 // managerEsProxyContainer returns the ES proxy container
 func (c *managerComponent) managerEsProxyContainer() corev1.Container {
 	var keyPath, certPath string
-	if c.cfg.TLSKeyPair != nil {
+	if c.cfg.InternalTrafficSecret != nil {
```
caseydavenport, line 519: Please rename this to something like `InternalTLSKeyPair` instead of "TrafifcSecret" - the field is a `KeyPairInterface` and not a `corev1.Secret`

pkg/render/manager.go
```diff
@@ -139,7 +139,7 @@ type ManagerConfiguration struct {
 	// KeyPair used for establishing mTLS tunnel with Guardian.
 	TunnelSecret certificatemanagement.KeyPairInterface
 
-	// TLS KeyPair used by Voltron within the cluster when operating as a management cluster.
+	// TLS KeyPair used by Voltron/EsProxy within the cluster when operating as a management cluster.
```
caseydavenport, line 142: I think we need to improve this comment more - it's not super clear what the different uses are for this, and I believe as part of this PR it's no longer just used in management clusters.

Is this righ? 

```
// TLS KeyPair used by both Voltron and es-proxy, presented by each as part of the mTLS handshake with other services within the cluster. This is used in both management and standalone clusters. 
```

pkg/controller/manager/manager_controller.go
```diff
@@ -447,6 +447,20 @@ func (r *ReconcileManager) Reconcile(ctx context.Context, request reconcile.Requ
 	var tunnelSecret certificatemanagement.KeyPairInterface
 	var internalTrafficSecret certificatemanagement.KeyPairInterface
 	var linseedVoltronSecret certificatemanagement.KeyPairInterface
+
+	// We expect that the secret that holds the certificates for internal communication within the management
```
caseydavenport, line 451: We shouldn't need this at all in managed clusters - don't we need to put this within a branch like this?

```
if !managed {
}
```

?

pkg/render/manager.go
```diff
@@ -107,8 +107,8 @@ func Manager(cfg *ManagerConfiguration) (Component, error) {
 		}
 	}
 
+	tlsAnnotations[cfg.InternalTrafficSecret.HashAnnotationKey()] = cfg.InternalTrafficSecret.HashAnnotationValue()
```
rene-dekker, line 110: This should be guarded with a nil-check for managed clusters if the other comment is implemented.

pkg/controller/manager/manager_controller.go
asincu: Manager is not installed on the managed clusters, only standalone and management.

rene-dekker: It is created by the kube-controllers controller. Historically that has always been the case IIRC.
See https://github.com/tigera/operator/blob/master/pkg/controller/logstorage/eskubecontrollers.go#L59

pkg/controller/manager/manager_controller.go
```diff
@@ -444,9 +444,22 @@ func (r *ReconcileManager) Reconcile(ctx context.Context, request reconcile.Requ
 		return reconcile.Result{}, err
 	}
 
-	var tunnelSecret certificatemanagement.KeyPairInterface
-	var internalTrafficSecret certificatemanagement.KeyPairInterface
+	// We expect that the secret that holds the certificates for internal communication within the management
+	// K8S cluster is already created by the KubeControllers
+	internalTrafficSecret, err := certificateManager.GetKeyPair(r.client, render.ManagerInternalTLSSecretName, common.OperatorNamespace())
```
rene-dekker, line 449: I believe that we should now do GetOrCreate in the manager_controller and change the esubecontrollers to just GetCertificate() or GetKeyPair() depending on whether we need the private key for that controller.

pkg/controller/manager/manager_controller.go
caseydavenport: Strange. Why is this certificate different from the others provisioned by the operator?

caseydavenport: > It is created by the kube-controllers controller

AH, I see - this is very different from "created by the KubeControllers" - which to my implies it is created by the kube-controllers pod itself. 

We should clarify this comment to make this more accurate.

rene-dekker: > Strange. Why is this certificate different from the others provisioned by the operator?

I don't quite remember. I thought @asincu was actually the person who first introduced this some years ago. My recollection of it, was that the kube controllers is passing the secret down the tunnel, but that would still not explain why it should be created here.

pkg/controller/manager/manager_controller.go
asincu: What if we install MCM without going through Enterprise ? Manager is the last component rendered.

rene-dekker: I'd like that to be in master, but not in v1.30 (where this will be picked to)

rene-dekker (approved):
lgtm

pkg/controller/manager/manager_controller.go
caseydavenport: > . I think the logic here was to have it created in kube-controllers because it is the first component that needs it (for MCM) and then propagate this to the rest of the components.

kube-controllers will only need this if it is talking to the manager pod though. This should be created and owned by the manager, and then kube-controllers can watch for it and update its bundle as needed if and when the secret is eventually created.

PR #2767, opened 2023-07-24
Recreate certificates that are only specified to be used as server certs

caseydavenport:
## Description

Old clusters from before Calico Enterprise v3.13 created certificates
that were only valid as server certs.

In v3.17, we have updated these connections to use mTLS, and so these
certs are also used as client certs.

This PR updates the operator to spot when certificates issued by the
tigera-operator-signer do not have the necessary ExtUsageKeyClientAuth
usage, and regenerates them if so.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


pkg/controller/certificatemanager/certificatemanager.go
```diff
@@ -237,44 +240,84 @@ func (cm *certificateManager) getKeyPair(cli client.Client, secretName, secretNa
 		return nil, nil, err
 	}
 
+	// First, check that it's not expired.
```
nelljerram, line 243: nit: "or not yet valid"

pkg/controller/certificatemanager/certificatemanager.go
```diff
@@ -237,44 +240,84 @@ func (cm *certificateManager) getKeyPair(cli client.Client, secretName, secretNa
...
+		// This certificate wasn't issued using external certificate management.
+		if !ValidForClientAndServer(x509Cert) {
+			// The certificate isn't valid as both a client and server cert. Older versions of the operator
+			// generated certificates that were only valid for use as a server certificate. However, some components now
+			// use these as client certificates in mTLS connections as well. So, regenerate any that need this capability.
+			return nil, nil, nil
+		}
+
 		if string(x509Cert.AuthorityKeyId) == string(cm.AuthorityKeyId) {
+			// The certifiate's authority matches the current CA.
```
nelljerram, line 275: typo here in certifiate

pkg/controller/certificatemanager/certificatemanager.go
```diff
@@ -237,44 +240,84 @@ func (cm *certificateManager) getKeyPair(cli client.Client, secretName, secretNa
...
+}
+
+func X509FromSecret(secret *corev1.Secret) (*x509.Certificate, error) {
+	_, certPEM := getKeyCertPEM(secret)
+	x509Cert, err := certificatemanagement.ParseCertificate(certPEM)
+	if err != nil {
+		return nil, err
+	}
+	return x509Cert, nil
+}
```
nelljerram, line 307: Would it be better to define this in a test-only code file, as IIUC it is only used for UT?

nelljerram (commented):
LGTM, with a few possible nits.  (But I am not an operator dev, so I am sure you will want an approval from one of them, and hence have not marked as approved myself.)

pkg/controller/certificatemanager/certificatemanager.go
```diff
@@ -237,44 +240,84 @@ func (cm *certificateManager) getKeyPair(cli client.Client, secretName, secretNa
...
 		if cm.keyPair.CertificateManagement != nil {
 			return certificateManagementKeyPair(cm, secretName, nil), nil, nil
 		}
+
+		// This certificate wasn't issued using external certificate management.
+		if !ValidForClientAndServer(x509Cert) {
+			// The certificate isn't valid as both a client and server cert. Older versions of the operator
+			// generated certificates that were only valid for use as a server certificate. However, some components now
+			// use these as client certificates in mTLS connections as well. So, regenerate any that need this capability.
+			return nil, nil, nil
```
tmjd, line 271: Should we be returning an error here to convey why the certificate isn't valid?

pkg/controller/certificatemanager/certificatemanager.go
```diff
@@ -237,44 +240,84 @@ func (cm *certificateManager) getKeyPair(cli client.Client, secretName, secretNa
...
 			return nil, nil, nil
 		}
-		// We return an error for byo secrets.
+
+		// The certificate was not issued by tigera-operator, or the caller indicated that certs shouldn't be modified.
+		// Consider this a BYO certificate, and error out rather than generate a new certificate.
 		return nil, nil, fmt.Errorf("secret %s is not valid at this date", secretName)
 	}
 	var issuer certificatemanagement.KeyPairInterface
 	if x509Cert.Issuer.CommonName == rmeta.TigeraOperatorCAIssuerPrefix {
```
tmjd, line 261: Do we need to do the `ValidForClientAndServer(x509Cert)` check for the else case of this? I'm thinking we need to do that for BYO certificates and let the user know if the provided certificates aren't sufficient.

pkg/controller/certificatemanager/certificatemanager.go
caseydavenport: I would have liked to, although the specific functionality that I wanted from here was the `getKeyCertPEM` which is currently private.

I suppose I could just make that public though!

nelljerram: Ah I see - whatever you prefer then!

rene-dekker (approved):
lgtm

pkg/controller/certificatemanager/certificatemanager.go
rene-dekker: That's right.

pkg/controller/certificatemanager/certificatemanager_test.go
```diff
@@ -484,15 +538,14 @@ var _ = Describe("Test CertificateManagement suite", func() {
 			By("counting the number of pem blocks in the configmap")
 			bundle := configMap.Data[certificatemanagement.TrustedCertConfigMapKeyName]
 			numBlocks := strings.Count(bundle, "certificate name:")
-			// While we have the ca + 4 certs, we expect 3 cert blocks (+ 2):
+			// While we have the ca + 4 certs, we expect just 2 cert blocks (+ 2):
```
rene-dekker, line 541: Have you considered adding a 5th cert when making the bundle, so we cover both the modern legacy secret use case and the legacy usage legacy secret?

rene-dekker (approved):
lgtm

pkg/controller/certificatemanager/certificatemanager.go
```diff
@@ -237,18 +237,24 @@ func (cm *certificateManager) getKeyPair(cli client.Client, secretName, secretNa
...
 			// We return nil, so a new secret will be created for expired (legacy) operator signed secrets.
 			return nil, nil, nil
 		}
+
 		// We return an error for byo secrets.
-		return nil, nil, fmt.Errorf("secret %s is not valid at this date", secretName)
+		if timeInvalid {
+			return nil, nil, fmt.Errorf("secret %s/%s is not valid at this date", secretNamespace, secretName)
+		}
+		return nil, nil, fmt.Errorf("secret %s/%s must specify ExtKeyusageClientAuth", secretNamespace, secretName)
```
tmjd, line 257: This means we won't ever get this secret recreated since we're returning an error here, at least that is my understanding from previous comments on this PR.
Also I tried these changes in cluster and the operator kept logging `secret <ns>/<name> must specify ExtKeyusageClientAuth` for several secrets.
Should we create an error type that we can return here and then check for that error type in the caller and handle the error as another case for when we need to create the secret?

tmjd: Oh I see we shouldn't get here because of the line `if timeInvalid || !ValidForClientAndServer(x509Cert) {` should handle the case I'm concerned with.

PR #2772, opened 2023-07-26
FV test improvements

Josh-Tigera:
-Add localized caching of OSS Calico images to speed up performance and improve test reliability by removing variability of image pull timing per test. Also has the added benefit of not ripping through docker pull request limits
-Fix race condition in CRD management tests
-Fix race condition in mainline tests
-Use non-caching k8s client and remove use of unstructured types

Test timing on local laptop went from 25 - 30 minutes to <2.5 minutes.
Ran ~250 iterations with no failures for a failure rate of <0.5%

## For PR author

~~- [ ] Tests for change.~~

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -111,6 +111,7 @@ CALICO_BUILD?=calico/go-build:$(GO_BUILD_VER)-$(ARCH)
 SRC_FILES=$(shell find ./pkg -name '*.go')
 SRC_FILES+=$(shell find ./api -name '*.go')
 SRC_FILES+=$(shell find ./controllers -name '*.go')
+SRC_FILES+=$(shell find ./test -name '*.go')
 SRC_FILES+=main.go
 
 EXTRA_DOCKER_ARGS += -e GO111MODULE=on -e GOPRIVATE=github.com/tigera/*
@@ -299,7 +300,7 @@ ut:
 	ginkgo -trace -r -focus="$(GINKGO_FOCUS)" $(GINKGO_ARGS) "$(UT_DIR)"'
 
 ## Run the functional tests
-fv: cluster-create run-fvs cluster-destroy
+fv: cluster-create load-container-images run-fvs cluster-destroy
 run-fvs:
 	-mkdir -p .go-pkg-cache report
 	$(CONTAINERIZED) $(CALICO_BUILD) sh -c '$(GIT_CONFIG_SSH) \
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -111,6 +111,7 @@ CALICO_BUILD?=calico/go-build:$(GO_BUILD_VER)-$(ARCH)
 SRC_FILES=$(shell find ./pkg -name '*.go')
 SRC_FILES+=$(shell find ./api -name '*.go')
 SRC_FILES+=$(shell find ./controllers -name '*.go')
+SRC_FILES+=$(shell find ./test -name '*.go')
```
Josh-Tigera, line 114: This change was so that "make fix" would apply to the files in ./test. I was getting a commit check failure but make fix wasn't actually applying the fix.

test/active_test.go
```diff
@@ -74,19 +72,8 @@ var _ = Describe("pkg/active with apiserver", func() {
...
-			u := &unstructured.Unstructured{}
-			u.SetGroupVersionKind(schema.GroupVersionKind{
-				Group:   "",
-				Version: "v1",
-				Kind:    "Namespace",
-			})
-
-			k := client.ObjectKey{Name: "calico-system"}
-			err := c.Get(context.Background(), k, u)
+			err := GetResource(c, ns)
```
caseydavenport, line 76: nit: you could just do `return GetResource(c, ns)`

test/mainline_test.go
```diff
@@ -339,6 +349,21 @@ func removeInstallResourceCR(c client.Client, name string, ctx context.Context)
...
+	Eventually(func() error {
+		installation := &operator.Installation{
+			TypeMeta:   metav1.TypeMeta{Kind: "Installation", APIVersion: "operator.tigera.io/v1"},
+			ObjectMeta: metav1.ObjectMeta{Name: name},
+		}
+		err = GetResource(c, installation)
+		if kerror.IsNotFound(err) || kerror.IsGone(err) {
+			return nil
+		}
+		return fmt.Errorf("\"%s\" installation still exists", name)
```
caseydavenport, line 365: I think we should return the err if it's non-nil but also not a 404

test/active_test.go
Josh-Tigera: Took it once step further and removed the func definition entirely so value of `GetResource` could be evaluated directly.

test/mainline_test.go
Josh-Tigera: Done

test/mainline_test.go
```diff
@@ -339,6 +349,24 @@ func removeInstallResourceCR(c client.Client, name string, ctx context.Context)
 	Expect(err).NotTo(HaveOccurred())
 	err = c.Delete(ctx, instance)
 	Expect(err).NotTo(HaveOccurred())
+	// Need to wait here for Installation resource to be fully deleted prior to cancelling the context
+	// which will in turn terminate the operator. Race conditions can occur otherwise that will leave the
+	// Installation resource intact while the operator is no longer running which will result in test failures
+	// that try to create an Installation resource of their own
+	Eventually(func() error {
+		installation := &operator.Installation{
```
caseydavenport, line 357: would it make sense to encap this into a helper function rather than duplicate it here and also above?

Josh-Tigera: Yeah good call. There's actually a helper function that already exists so I've repurposed it a bit and added a timeout parameter, then replaced the relevant blocks in my diff to use it.

test/crd_management_test.go
```diff
@@ -118,63 +105,45 @@ var _ = Describe("CRD management tests", func() {
 
 		mgr = nil
 
-		_ = c.Delete(context.Background(), npCRD.DeepCopy())
+		//Need to make sure the operator is shut down prior to deleting and recreating the networkpolicies CRD otherwise
```
caseydavenport, line 108: nit: need a space after `//`

test/crd_management_test.go
```diff
@@ -118,63 +105,45 @@ var _ = Describe("CRD management tests", func() {
...
-		}, 20*time.Second).Should(BeNil())
+			err := GetResource(c, npCRD)
+			if kerror.IsNotFound(err) || kerror.IsGone(err) {
+				return nil
+			} else if err != nil {
+				return err
+			} else {
+				return fmt.Errorf("NetworkPolicy CRD still exists")
+			}
+		}, 20*time.Second).MustPassRepeatedly(50).ShouldNot(HaveOccurred())
```
caseydavenport, line 134: Hm, do we really need to run this 50 times?

test/mainline_test.go
```diff
@@ -72,19 +75,6 @@ var _ = Describe("Mainline component function tests", func() {
 		if err != nil && !kerror.IsAlreadyExists(err) {
 			Expect(err).NotTo(HaveOccurred())
 		}
-		//
```
caseydavenport, line 75: Do we need an ExpectResourceDestroyed here?

caseydavenport: I guess it's handled in `cleanupResources`

test/utils.go
```diff
@@ -52,12 +52,18 @@ func ExpectResourceCreated(c client.Client, obj client.Object) {
...
 	EventuallyWithOffset(1, func() error {
 		err = GetResource(c, obj)
-		return err
-	}, 10*time.Second).ShouldNot(BeNil(), fmt.Sprintf("GetResource %s should return error", obj.GetName()))
+		if errors.IsNotFound(err) || errors.IsGone(err) {
+			return nil
+		} else if err != nil {
+			return err
+		} else {
+			return fmt.Errorf("Resource \"%s\" should no longer exist", obj.GetName())
```
caseydavenport, line 64: You can use single quotes here mixed with the double quotes to avoid needing to escape the string.

test/crd_management_test.go
Josh-Tigera: 50 might be a touch aggressive. When I was testing I observed that sometimes the first ~10 iterations could toggle back and forth between success and failure so I just wanted to ensure that the value had truly stabilized prior to advancing. I could try reducing this and running longevity tests but ultimately this is only requiring a 500 ms stable result in exchange for completely eliminating any flakiness.

test/crd_management_test.go
Josh-Tigera: Fixed.

test/utils.go
Josh-Tigera: Fixed.

marvin-tigera:

caseydavenport (approved):

PR #2775, opened 2023-07-28
Cert fix

tmjd:
## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/compliance/compliance_controller.go
+++ b/pkg/controller/compliance/compliance_controller.go
@@ -442,6 +442,9 @@ func (r *ReconcileCompliance) Reconcile(ctx context.Context, request reconcile.R
 	}
 
 	reqLogger.V(3).Info("rendering components")
+
+	namespaceComp := render.NewPassthrough(render.CreateNamespace(render.ComplianceNamespace, network.KubernetesProvider, render.PSSPrivileged))
+
 	hasNoLicense := !utils.IsFeatureActive(license, common.ComplianceFeature)
 	openshift := r.provider == operatorv1.ProviderOpenShift
 	complianceCfg := &render.ComplianceConfiguration{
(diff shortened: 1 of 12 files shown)


pkg/controller/compliance/compliance_controller.go
```diff
@@ -442,6 +442,9 @@ func (r *ReconcileCompliance) Reconcile(ctx context.Context, request reconcile.R
 	}
 
 	reqLogger.V(3).Info("rendering components")
+
+	namespaceComp := render.NewPassthrough(render.CreateNamespace(render.ComplianceNamespace, network.KubernetesProvider, render.PSSPrivileged))
```
caseydavenport, line 446: I'm not sure I understand the relevance of this change

caseydavenport (approved):

pkg/controller/compliance/compliance_controller.go
tmjd: The certificates need to be created/updated before the compliance deployments are updated and to do that we need the namespace created first.

pkg/render/compliance.go
```diff
@@ -85,6 +87,14 @@ var (
 	ComplianceReporterSourceEntityRule    = networkpolicy.CreateSourceEntityRule(ComplianceNamespace, ComplianceReporterName)
 )
 
+// Register secret/certs that need Server and Client Key usage
+func init() {
+	certkeyusage.SetCertKeyUsage(ComplianceServerCertSecret, []x509.ExtKeyUsage{x509.ExtKeyUsageClientAuth, x509.ExtKeyUsageServerAuth})
```
rene-dekker, line 92: When considering the init function, what made you decide to do it in the render file, rather than the controller?

tmjd: Mainly because that is where the secret name is defined.

Issue #2783, opened 2023-08-01
Duplicate FELIX_HEALTHPORT environment variable causing problems in Openshift


dzacball:
When installing Calico with Tigera operator on Openshift, the environment variable `FELIX_HEALTHPORT` is duplicated in `calico-node` pods.

## Expected Behavior
`FELIX_HEALTHPORT` should not be duplicated in the calico-node Daemonset for Openshift clusters.

## Current Behavior
`FELIX_HEALTHPORT` is duplicated in the calico-node Daemonset or Openshift clusters.

## Possible Solution
I'm not exactly sure how this should be handled/fixed. 
Ever since the PR https://github.com/tigera/operator/pull/1955 was merged, there have been two instances of this environment variable present for `calico-node` on Openshift. 
First one is inserted here: https://github.com/tigera/operator/blob/af15d9ba47d3c4a772030ea61e9a9570ddaa81c1/pkg/render/node.go#L1388 
second one is inserted here (in case of `operatorv1.ProviderOpenShift`): https://github.com/tigera/operator/blob/af15d9ba47d3c4a772030ea61e9a9570ddaa81c1/pkg/render/node.go#L1657-L1659

## Steps to Reproduce (for bugs)
1. Install Calico on an Openshift 4.13 cluster using the Tigere Operator.

## Context
The duplication of the environment variable alone wouldn't be a huge problem. However, there is a K8s bug present in 1.26 (which Openshift 4.13 is based on) which causes the Pod Garbage Collector to fail for pods that have duplicated environment variables. This is well described in https://github.com/kubernetes/kubernetes/issues/118261.
This bug, combined with the duplicated `FELIX_HEALTHPORT` environment variable for `calico-node` will result in `calico-node` pods not being cleaned up after a k8s node is deleted, and this eventually can keep calico-node rolling update from succeeding (due to the `maxUnavailable` rollingUpdate setting).

## Your Environment
Openshift 4.13 + latest Tigera Operator

dzacball:
I created a PR with a possible fix for this at https://github.com/tigera/operator/pull/2785.

rene-dekker:
Thank you for raising this issue and for helping with a solution. I did some delving into the history of this repo. From what I understand:
This commit was present in 2019: https://github.com/tigera/operator/blame/eda3df71a96ec8d4aab6665feddfb41a8dc8a1c7/pkg/render/node.go#L482
[...]

PR #2785, opened 2023-08-02
replace value of FELIX_HEALTHPORT instead of appending it

dzacball:
## Description
As described in https://github.com/tigera/operator/issues/2783, when installing Calico with Tigera operator on Openshift, the environment variable `FELIX_HEALTHPORT` is duplicated in calico-node pods, which can cause problems on specific Openshift clusters that are affected by a Kubernetes bug described at https://github.com/kubernetes/kubernetes/issues/118261.
This PR aims to fix this by overwriting the value for `FELIX_HEALTHPORT` instead of appending a new version of it (in case of Openshift).

For more details, please take a look at https://github.com/tigera/operator/issues/2783.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/node.go
+++ b/pkg/render/node.go
@@ -1656,7 +1656,12 @@ func (c *nodeComponent) nodeEnvVars() []corev1.EnvVar {
 	switch c.cfg.Installation.KubernetesProvider {
 	case operatorv1.ProviderOpenShift:
 		// For Openshift, we need special configuration since our default port is already in use.
-		nodeEnv = append(nodeEnv, corev1.EnvVar{Name: "FELIX_HEALTHPORT", Value: "9199"})
+		for i := range nodeEnv {
+			if nodeEnv[i].Name == "FELIX_HEALTHPORT" {
+				nodeEnv[i].Value = "9199"
+				break
+			}
+		}
 	// For AKS/AzureVNET and EKS/VPCCNI, we must explicitly ask felix to add host IP's to wireguard ifaces
 	case operatorv1.ProviderAKS:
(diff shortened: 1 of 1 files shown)


pkg/render/node.go
```diff
@@ -1656,7 +1656,12 @@ func (c *nodeComponent) nodeEnvVars() []corev1.EnvVar {
 	switch c.cfg.Installation.KubernetesProvider {
 	case operatorv1.ProviderOpenShift:
 		// For Openshift, we need special configuration since our default port is already in use.
-		nodeEnv = append(nodeEnv, corev1.EnvVar{Name: "FELIX_HEALTHPORT", Value: "9199"})
```
caseydavenport, line 1659: Interesting, I wonder if this was already encountering another bu which was that it wasn't respecting `c.cfg.FelixHealthPort` 

> 		{Name: "FELIX_HEALTHPORT", Value: fmt.Sprintf("%d", c.cfg.FelixHealthPort)},

@tmjd or @mgleung might know better?

To me, it seems like we should probably be addressing this earlier in this function and accounting for the user specified health port - otherwise we're overwriting user input in this branch.

caseydavenport:
@dzacball actually, think this is the right fix: https://github.com/tigera/operator/pull/2795#pullrequestreview-1561534993

Turns out we were already defaulting `cfg.FelixHealthPort` for OpenShift in the calling code, so simply removing that branch does the trick.

caseydavenport:
Going to close this in favor of the other PR, but thank you for the heads up and for proposing a fix!

dzacball:
> 

@caseydavenport understood! Thanks!

PR #2786, opened 2023-08-02
WIP: Refix cert

tmjd:
## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/certificatemanager/certificatemanager.go
+++ b/pkg/controller/certificatemanager/certificatemanager.go
@@ -207,6 +207,51 @@ func (cm *certificateManager) GetOrCreateKeyPair(cli client.Client, secretName, 
 	}, nil
 }
 
+// This type will be returned for errors that do not have the correct Ext Key usage types
+// for a specific secert certificate.
+type CertExtKeyUsageError struct {
+	msg string
+}
+
+func (cue *CertExtKeyUsageError) Error() string {
+	return cue.msg
+}
+
+var _ error = &CertExtKeyUsageError{}
+
+// Returns true if the error is a CertExtKeyUsageError
+func IsCertExtKeyUsageError(err error) bool {
+	_, ok := err.(*CertExtKeyUsageError)
+	return ok
+}
+
+func extKeyUsageToString(requiredExtKeyUsages []x509.ExtKeyUsage) string {
+	out := ""
+	for i, x := range requiredExtKeyUsages {
+		if i != 0 {
+			out = out + ", "
+		}
+		switch x {
+		case x509.ExtKeyUsageServerAuth:
+			out = out + "ExtKeyUsageServerAuth"
+		case x509.ExtKeyUsageClientAuth:
+			out = out + "ExtKeyUsageClientAuth"
+		default:
+			out = out + "unknown"
+		}
+
+	}
+	return out
+}
+
+func newCertExtKeyUsageError(name, ns string, requiredExtKeyUsages []x509.ExtKeyUsage) error {
+	return &CertExtKeyUsageError{
+		fmt.Sprintf(
+			"secret %s/%s must specify ext key usages: %+v",
+			ns, name, extKeyUsageToString(requiredExtKeyUsages)),
+	}
+}
+
 // getKeyPair is an internal convenience method to retrieve a keypair or a certificate.
 func (cm *certificateManager) getKeyPair(cli client.Client, secretName, secretNamespace string, readCertOnly bool) (certificatemanagement.KeyPairInterface, *x509.Certificate, error) {
 
@@ -261,7 +306,7 @@ func (cm *certificateManager) getKeyPair(cli client.Client, secretName, secretNa
 		if timeInvalid {
 			return nil, nil, fmt.Errorf("secret %s/%s is not valid at this date", secretNamespace, secretName)
 		}
-		return nil, nil, fmt.Errorf("secret %s/%s must specify ext key usages: %+v", secretNamespace, secretName, requiredKeyUsages)
+		return nil, nil, newCertExtKeyUsageError(secretName, secretNamespace, requiredKeyUsages)
 	}
 
 	var issuer certificatemanagement.KeyPairInterface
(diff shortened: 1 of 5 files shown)


pkg/controller/certificatemanager/certificatemanager.go
```diff
@@ -207,6 +207,26 @@ func (cm *certificateManager) GetOrCreateKeyPair(cli client.Client, secretName,
...
+
+func (cue *CertUsageError) Error() string {
+	return cue.msg
+}
+
+var _ error = &CertUsageError{}
+
+// Returns true if the error is a CertUsageError
+func IsCertUsageError(err error) bool {
+	if _, ok := err.(*CertUsageError); ok {
```
rene-dekker, line 224: nit: we don't need an if-statement and can return ok.

pkg/controller/certificatemanager/certificatemanager.go
```diff
@@ -261,7 +281,7 @@ func (cm *certificateManager) getKeyPair(cli client.Client, secretName, secretNa
 		if timeInvalid {
 			return nil, nil, fmt.Errorf("secret %s/%s is not valid at this date", secretNamespace, secretName)
 		}
-		return nil, nil, fmt.Errorf("secret %s/%s must specify ext key usages: %+v", secretNamespace, secretName, requiredKeyUsages)
+		return nil, nil, &CertUsageError{fmt.Sprintf("secret %s/%s must specify ext key usages: %+v", secretNamespace, secretName, requiredKeyUsages)}
```
rene-dekker, line 284: It will print:

> Failed to get certificate: secret tigera-operator/tigera-compliance-server-tls must specify ext key usages: [2 1]

Which are the enum indices, not the value. I'm ok to defer though.

pkg/controller/certificatemanager/certificatemanager.go
```diff
@@ -207,6 +207,26 @@ func (cm *certificateManager) GetOrCreateKeyPair(cli client.Client, secretName,
 	}, nil
 }
 
+// This type will be returned for errors that do not have the correct Key usage types
+// for a specific secert certificate.
+type CertUsageError struct {
+	msg string
```
caseydavenport, line 213: It would be nicer to build the error within the struct, since we have that capability. 

e.g., 

```
func NewCertUsage(name, ns string, requiredKeyUsages []x509.ExtKeyUsage) error {
  ...
}
```

pkg/controller/certificatemanager/certificatemanager.go
```diff
@@ -207,6 +207,26 @@ func (cm *certificateManager) GetOrCreateKeyPair(cli client.Client, secretName,
 	}, nil
 }
 
+// This type will be returned for errors that do not have the correct Key usage types
+// for a specific secert certificate.
+type CertUsageError struct {
```
caseydavenport, line 212: We should name this all `CertExtKeyUsage` - `CertUsage` is ambiguous

pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -464,8 +464,13 @@ func (r *ReconcileLogStorage) generateSecrets(
 	for _, certName := range certs {
 		cert, err := cm.GetCertificate(r.client, certName, common.OperatorNamespace())
 		if err != nil {
-			r.status.SetDegraded(operatorv1.ResourceReadError, "Failed to get certificate", err, log)
-			return nil, err
+			if certificatemanager.IsCertUsageError(err) {
+				msg := fmt.Sprintf("skipping tigera-operator/%s secret it will be added when it is updated: %s", certName, err)
```
caseydavenport, line 468: We shouldn't hard-code the namespace in the log - we have `common.OperatorNamespace()`

pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -464,8 +464,13 @@ func (r *ReconcileLogStorage) generateSecrets(
 	for _, certName := range certs {
 		cert, err := cm.GetCertificate(r.client, certName, common.OperatorNamespace())
 		if err != nil {
-			r.status.SetDegraded(operatorv1.ResourceReadError, "Failed to get certificate", err, log)
-			return nil, err
+			if certificatemanager.IsCertUsageError(err) {
```
caseydavenport, line 467: I think we need a big comment here explaining why we have this check, and why it's not needed everywhere else that `GetCertificate` is called

rene-dekker (approved):
lgtm

pkg/controller/certificatemanager/certificatemanager.go
```diff
@@ -207,6 +207,51 @@ func (cm *certificateManager) GetOrCreateKeyPair(cli client.Client, secretName,
...
+			out = out + "ExtKeyUsageClientAuth"
+		default:
+			out = out + "unknown"
+		}
+
+	}
+	return out
+}
+
+func newCertExtKeyUsageError(name, ns string, requiredExtKeyUsages []x509.ExtKeyUsage) error {
```
caseydavenport, line 247: I should have been more clear - this is fine for now, but I think a better pattern in Go when dealing with custom error types is to have typed fields in the error itself, and then build the error message in the `Error()` method itself, rather than constructing the message in the constructor.

It allows potential callers to actually access relevant details. For example, if calling code wanted to switch on error type and pull out the secret name and namespace, that would be easy to do with `e.Namespace` instead of needing to parse an error string.

We're not doing anything like that in this PR so it's moot, more or just a best-practice thing.

pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -464,10 +464,23 @@ func (r *ReconcileLogStorage) generateSecrets(
 	for _, certName := range certs {
 		cert, err := cm.GetCertificate(r.client, certName, common.OperatorNamespace())
 		if err != nil {
-			r.status.SetDegraded(operatorv1.ResourceReadError, "Failed to get certificate", err, log)
-			return nil, err
+			// This check is needed because there is a circular dependency between the certs/secrets
+			// used here and with the linseed certs that are created by linseed which this controller
```
caseydavenport, line 468: > linseed certs that are created by linseed 

Linseed itself doesn't create any certificates - which secret is this referring to?

pkg/controller/certificatemanager/certificatemanager.go
tmjd: Ah I see what you mean.

tmjd:
Replaced by #2790

PR #2802, opened 2023-08-08
ev-4025 release-v1.30 Create internal manager tls inside of manager_controller and fix certificate management

rene-dekker:
To fix certificate management, the following things were currently broken:
- the tunnel secret was never created, looks like a regression
- The internal-manager-tls secret should be created in the manager_controller such that certificate management can create a csr with the correct dns names
- Now mounting the ca-bundle in the kubecontrollers, rather than the internal-manager-tls secret.
- Some linseed secrets did not yet have init containers created for them.

### DIFF
--- a/pkg/controller/apiserver/apiserver_controller.go
+++ b/pkg/controller/apiserver/apiserver_controller.go
@@ -278,7 +278,7 @@ func (r *ReconcileAPIServer) Reconcile(ctx context.Context, request reconcile.Re
 	}
 
 	// Query enterprise-only data.
-	var tunnelCASecret certificatemanagement.KeyPairInterface
+	var tunnelCAKeyPair certificatemanagement.KeyPairInterface
 	var trustedBundle certificatemanagement.TrustedBundle
 	var amazon *operatorv1.AmazonCloudIntegration
 	var managementCluster *operatorv1.ManagementCluster
@@ -305,19 +305,20 @@ func (r *ReconcileAPIServer) Reconcile(ctx context.Context, request reconcile.Re
 		}
 
 		if managementCluster != nil {
-			tunnelCASecret, err = certificateManager.GetKeyPair(r.client, render.VoltronTunnelSecretName, common.OperatorNamespace())
-			if tunnelCASecret == nil {
-				tunnelSecret, err := certificatemanagement.CreateSelfSignedSecret(render.VoltronTunnelSecretName, common.OperatorNamespace(), "tigera-voltron", []string{"voltron"})
-				if err == nil {
-					tunnelCASecret = certificatemanagement.NewKeyPair(tunnelSecret, nil, "")
-					// Creating the voltron tunnel secret is not (yet) supported by certificate mananger.
-					tunnelSecretPassthrough = render.NewPassthrough(tunnelCASecret.Secret(common.OperatorNamespace()))
-				}
-			}
+			tunnelCASecret, err := utils.GetSecret(ctx, r.client, render.VoltronTunnelSecretName, common.OperatorNamespace())
 			if err != nil {
-				r.status.SetDegraded(operatorv1.ResourceCreateError, "Unable to get or create the tunnel secret", err, reqLogger)
+				r.status.SetDegraded(operatorv1.ResourceReadError, "Unable to fetch the tunnel secret", err, reqLogger)
 				return reconcile.Result{}, err
 			}
+			if tunnelCASecret == nil {
+				tunnelCASecret, err = certificatemanagement.CreateSelfSignedSecret(render.VoltronTunnelSecretName, common.OperatorNamespace(), "tigera-voltron", []string{"voltron"})
+				if err != nil {
+					r.status.SetDegraded(operatorv1.ResourceCreateError, "Unable to create the tunnel secret", err, reqLogger)
+					return reconcile.Result{}, err
+				}
+			}
+			tunnelSecretPassthrough = render.NewPassthrough(tunnelCASecret)
+			tunnelCAKeyPair = certificatemanagement.NewKeyPair(tunnelCASecret, nil, "")
 		}
 
 		if r.amazonCRDExists {
@@ -378,7 +379,7 @@ func (r *ReconcileAPIServer) Reconcile(ctx context.Context, request reconcile.Re
 		TLSKeyPair:                  tlsSecret,
 		PullSecrets:                 pullSecrets,
 		Openshift:                   r.provider == operatorv1.ProviderOpenShift,
-		TunnelCASecret:              tunnelCASecret,
+		TunnelCASecret:              tunnelCAKeyPair,
 		TrustedBundle:               trustedBundle,
 		UsePSP:                      r.usePSP,
 	}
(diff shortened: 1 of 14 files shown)


asincu (approved):

pkg/controller/installation/core_controller.go
```diff
@@ -1067,11 +1067,10 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 	}
 	var managerInternalTLSSecret certificatemanagement.KeyPairInterface
 	if instance.Spec.Variant == operator.TigeraSecureEnterprise {
-		dnsNames := append(dns.GetServiceDNSNames(render.ManagerServiceName, render.ManagerNamespace, r.clusterDomain), render.ManagerServiceIP)
-		managerInternalTLSSecret, err = certificateManager.GetOrCreateKeyPair(r.client, render.ManagerInternalTLSSecretName, common.OperatorNamespace(), dnsNames)
+		managerInternalTLSSecret, err = certificateManager.GetKeyPair(r.client, render.ManagerInternalTLSSecretName, common.OperatorNamespace())
```
Brian-McM, line 1070: With this change, won't `managerInternalTLSSecret` for a new installation will be nil (since the manager controller needs to wait for the core controller to setup networking)? It looks like the Objects function can handle nil key pairs, but it might still be better to not add this as an option if it's nil.

You may also want to investigate if you even need this tls secret in the core controller. I don't think anything here uses voltron anymore, it might just be left over from before we split kube controllers into two components (i.e. the es kube controllers).

Brian-McM (approved):

pkg/tls/certificatemanagement/keypair.go
```diff
@@ -123,6 +123,19 @@ func (k *KeyPair) VolumeMount(osType rmeta.OSType) corev1.VolumeMount {
 
 // InitContainer contains an init container for making a CSR. is only applicable when certificate management is enabled.
 func (k *KeyPair) InitContainer(namespace string) corev1.Container {
+	if len(k.DNSNames) == 0 {
```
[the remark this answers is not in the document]
rene-dekker: Ow, I was not supposed to commit the changes to this file..... Thanks for spotting this.

pkg/controller/installation/core_controller.go
rene-dekker: Very insightful, I had not thought of that.
- Now mounting the ca-bundle in the kubecontrollers, rather than the internal-manager-tls secret.
- We still need to watch the secret in order to add it to the bundle in the calico-system namespace.

pkg/controller/installation/core_controller.go
```diff
@@ -1081,6 +1073,16 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 		return reconcile.Result{}, err
 	}
 
+	if instance.Spec.Variant == operator.TigeraSecureEnterprise {
+		managerInternalTLSSecret, err := certificateManager.GetCertificate(r.client, render.ManagerInternalTLSSecretName, common.OperatorNamespace())
+		if err != nil {
+			r.status.SetDegraded(operator.ResourceReadError, fmt.Sprintf("Error fetching TLS secret %s in namespace %s", render.ManagerInternalTLSSecretName, common.OperatorNamespace()), err, reqLogger)
+			return reconcile.Result{}, nil
+		} else if managerInternalTLSSecret != nil {
+			typhaNodeTLS.TrustedBundle.AddCertificates(managerInternalTLSSecret)
```
Brian-McM, line 1082: What does typha need the manager tls secret for? Does it use voltron?

pkg/controller/apiserver/apiserver_controller.go
```diff
@@ -305,19 +305,20 @@ func (r *ReconcileAPIServer) Reconcile(ctx context.Context, request reconcile.Re
...
-				}
-			}
+			tunnelCASecret, err := utils.GetSecret(ctx, r.client, render.VoltronTunnelSecretName, common.OperatorNamespace())
 			if err != nil {
-				r.status.SetDegraded(operatorv1.ResourceCreateError, "Unable to get or create the tunnel secret", err, reqLogger)
+				r.status.SetDegraded(operatorv1.ResourceReadError, "Unable to fetch the tunnel secret", err, reqLogger)
 				return reconcile.Result{}, err
 			}
+			if tunnelCASecret == nil {
+				tunnelCASecret, err = certificatemanagement.CreateSelfSignedSecret(render.VoltronTunnelSecretName, common.OperatorNamespace(), "tigera-voltron", []string{"voltron"})
```
Brian-McM, line 314: Something seems funny about the fact that nothing was broken with MCM without the operator creating this secret. If the operator was supposed to be responsible for creating this, wouldn't every new MCM cluster be broken on startup?

Are you able to point to where the regression happen, and the code to create this was removed?

rene-dekker: First of all, the issue only occurs when installation.spec.certificatemanagement is enabled. This feature is probably only ever tested by me and 1 other person doing test cycles with both MCM and installation.spec.certificatemanagement. On top of that if you first create the management cluster and then enable certificate management, the secret will have already been created.

Looking back, I think the issue was present when we first merged b0fce5783d50c41c30942494fa1977b2d2d9325d

rene-dekker: If my memory serves me well during code reviews and testing, this line has changed: https://github.com/tigera/operator/commit/b0fce5783d50c41c30942494fa1977b2d2d9325d#diff-f9dc3b7ddca285064281a90259a65ddf0b5949a3ea1ba03ada3bda3e12810a0fR190

pkg/controller/installation/core_controller.go
rene-dekker: We render only one bundle per namespace.
The way the code evolved over time before the bundle was introduced is that this typhaNodeTLS struct is passed along for configuring node, typha and other components. This is why the trustedBundle is also in this struct. It's a bit misleading.

rene-dekker: The cert internal-manager-tls needs to be in the bundle to have the kube controllers trust voltron for MCM requests, but Typha has no need for it.

Brian-McM: OK, so it's needed for kube controllers. I think I posed the question before about this, but does kube controllers need this? What requests is kube controllers (not es kube controllers) making to voltron?

pkg/controller/apiserver/apiserver_controller.go
Brian-McM: > On top of that if you first create the management cluster and then enable certificate management, the secret will have already been created.

I'm still a little confused about what exactly is going on here. What's creating the secret when certificate management is not enabled? Is that in the operator?

rene-dekker:
```
tunnelSecretPassthrough = render.NewPassthrough(tunnelCASecret)
...
	if tunnelSecretPassthrough != nil {
		components = append(components, tunnelSecretPassthrough)
	}
```
It is added by rendering a passthrough component.

Brian-McM: I think I had a brain fart when I first asked these questions, I thought previous there was no `certificatemanagement.CreateSelfSignedSecret` call before your changes if `certificateManager.GetKeyPair` return an nil, so I was wondering what the heck used to create this. I had just missed the glaringly obvious section that you just moved.

You've only renamed the variable and handled the errors properly.

Forgot _**all**_ of this...

Brian-McM (approved):

pkg/controller/installation/core_controller.go
rene-dekker: es-kube-controllers is the pod requiring access, not kube-controllers. Because it resides in the calico-system namespace, it is using the trusted bundle created by the core_controller.

PR #2821, opened 2023-08-23
update coreruleset to 3.3.5

electricjesus:
## Description

CVE link: https://nvd.nist.gov/vuln/detail/CVE-2023-38199
- affects CRS v3.3.4 and below

On this PR:
1. update CRS rules data and modsecdefault.conf to v3.3.5/ engine v3.0.10
2. change default anomaly score thresholds to 100 so we can see more rule matches (about 20 rules max) on a processed request.

Fixes RS-1242

## For PR author

- [ ] Tests for change.
- [ ] If changing pkg/apis/, run `make gen-files`
- [ ] If changing versions, run `make gen-versions`

## For PR reviewers

A note for code reviewers - all pull requests must have the following:

- [ ] Milestone set according to targeted release.
- [ ] Appropriate labels:
  - `kind/bug` if this is a bugfix.
  - `kind/enhancement` if this is a a new feature.
  - `enterprise` if this PR applies to Calico Enterprise only.

PR #2823, opened 2023-08-23
replace gen.go with embed.FS coreruleset package

electricjesus:
## Description

old gen.go method was easy to miss, hard to update with a lot of manual intervention

- add new and configurable Makefile target(s) to pull required files from various sources
  - usage: `make generate-crs`
- remove references to manually-generated file map
- new filemap doesn't need to be b64. this improves CM rendering (less `\n` rendered when using `kubectl edit`)
- remove old embedded files, too
- added tests to verify embedded files, at least the important ones

retains past work in tigera/operator#2821

1. CRS version v3.3.5
2. ModSec setup files (modsecdefault, unicode.mapping) v3.0.10
3. manual anomaly scores bump to 100 (see `.hack-crs-custom-epilogue` Makefile target)
	- this should be fine appended to the end of `crs-setup.conf` since it is a `SecAction` override.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


Makefile
```diff
@@ -866,3 +866,49 @@ install-git-hooks:
...
 	$(CONTAINERIZED) $(CALICO_BUILD) git-hooks/pre-commit-in-container
+
+CRS_VERSION := 3.3.5
+MODSEC_VERSION := v3.0.10
+
+CRS_EMBED_FOLDER := pkg/render/applicationlayer/embed/coreruleset
+CRS_ARCHIVE_SRC := https://github.com/coreruleset/coreruleset/archive/refs/tags/v$(CRS_VERSION).tar.gz
+CRS_ARCHIVE_SHA := b552005471446a4a2454a6b6ee7e65f864219ce0
+CRS_ARCHIVE_TAR := v$(CRS_VERSION).tar.gz
+CRS_ARCHIVE_TMP := $(shell mktemp -d)
```
tmjd, line 877: I don't like that this will create a tmp directory on every make invocation. It would be much better if you had this done in the `$(CRS_EMBED_FOLDER)` target since it is only used in there and could be cleaned it up in there too.

Makefile
```diff
@@ -866,3 +866,49 @@ install-git-hooks:
...
+MODSEC_CONF_SRC := https://raw.githubusercontent.com/SpiderLabs/ModSecurity/v3.0.10
+MODSEC_CONF_FILES :=  modsecurity.conf-recommended unicode.mapping
+
+CHECKSUM_CMD := shasum
+
+.PHONY: generate-crs
```
tmjd, line 883: Please make this more specific, maybe `generate-applayer-ruleset`, `CRs` are also how we refer to instances of CRDs so could be confusing.

Makefile
```diff
@@ -866,3 +866,49 @@ install-git-hooks:
...
+
+CRS_VERSION := 3.3.5
+MODSEC_VERSION := v3.0.10
+
+CRS_EMBED_FOLDER := pkg/render/applicationlayer/embed/coreruleset
+CRS_ARCHIVE_SRC := https://github.com/coreruleset/coreruleset/archive/refs/tags/v$(CRS_VERSION).tar.gz
+CRS_ARCHIVE_SHA := b552005471446a4a2454a6b6ee7e65f864219ce0
+CRS_ARCHIVE_TAR := v$(CRS_VERSION).tar.gz
+CRS_ARCHIVE_TMP := $(shell mktemp -d)
+MODSEC_CONF_SRC := https://raw.githubusercontent.com/SpiderLabs/ModSecurity/v3.0.10
```
tmjd, line 878: Seems like you should use the `MODSEC_VERSION` defined above here instead of having it twice.

tmjd (changes requested):
IIUC this will now be reading, at runtime, from the file system for the default ruleset. Assuming I'm understanding correctly, there are some missing Dockerfile changes to include the ruleset files in the operator docker image. Have you tested out an operator image with these changes?

I'd also like to see the new generate make target added to CI and verify there aren't any "dirty" files (no changes) to make sure everything is updated as it should be.

PR #2831, opened 2023-08-26
Relax Linseed Policy in case we have DPI installed

asincu:
DPI runs on the Host network and it is hard to create a policy that targerts its endpoints. Thus, for Linseed we will relax the ingress policy to allow all traffic in case we have DPI enabled.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/logstorage/linseed.go
+++ b/pkg/controller/logstorage/linseed.go
@@ -17,6 +17,8 @@ package logstorage
 import (
 	"context"
 
+	v3 "github.com/tigera/api/pkg/apis/projectcalico/v3"
+
 	relasticsearch "github.com/tigera/operator/pkg/render/common/elasticsearch"
 
 	"github.com/go-logr/logr"
@@ -44,8 +46,14 @@ func (r *ReconcileLogStorage) createLinseed(
 	managementCluster bool,
 	usePSP bool,
 	esClusterConfig *relasticsearch.ClusterConfig,
-	hasDPIResource bool,
 ) (reconcile.Result, bool, error) {
+
+	dpiList := &v3.DeepPacketInspectionList{}
+	if err := r.client.List(ctx, dpiList); err != nil {
+		r.status.SetDegraded(operatorv1.ResourceReadError, "Failed to retrieve DeepPacketInspection resource", err, reqLogger)
+		return reconcile.Result{}, false, err
+	}
+	hasDPIResource := len(dpiList.Items) != 0
 
 	cfg := &linseed.Config{
 		Installation:      install,
--- a/pkg/controller/logstorage/linseed.go
+++ b/pkg/controller/logstorage/linseed.go
@@ -44,6 +44,7 @@ func (r *ReconcileLogStorage) createLinseed(
 	managementCluster bool,
 	usePSP bool,
 	esClusterConfig *relasticsearch.ClusterConfig,
+	hasDPIResource bool,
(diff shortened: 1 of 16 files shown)


pkg/controller/logstorage/linseed.go
```diff
@@ -44,6 +44,7 @@ func (r *ReconcileLogStorage) createLinseed(
 	managementCluster bool,
 	usePSP bool,
 	esClusterConfig *relasticsearch.ClusterConfig,
+	hasDPIResource bool,
```
caseydavenport, line 47: Rather than pass this in, you should just have the client query this from within this function.

We have way too many arguments here, and really the only things that should be arguments are variables that are shared between multiple components.

pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -786,6 +801,13 @@ func (r *ReconcileLogStorage) Reconcile(ctx context.Context, request reconcile.R
 		return reconcile.Result{}, err
 	}
 
+	dpiList := &v3.DeepPacketInspectionList{}
```
caseydavenport, line 804: You need to check on whether the DPI watch is ready before attempting to query it - if we don't wait, then we'll potentially hit the API server with lots of Get() requests.

The operator is designed to serve Get() requests from its cache, and only ever perform Watches against the API server.

pkg/controller/logstorage/logstorage_controller_test.go
```diff
@@ -125,6 +126,8 @@ var _ = Describe("LogStorage controller", func() {
 
 		readyFlag = &utils.ReadyFlag{}
 		readyFlag.MarkAsReady()
```
caseydavenport, line 128: If we're going to have two of these, we should rename the other one as well to be more clear.

`readyFlag` -> `tiersReadyflag`

pkg/controller/logstorage/logstorage_controller.go
```diff
@@ -243,6 +252,10 @@ func add(mgr manager.Manager, c controller.Controller) error {
 		return fmt.Errorf("logstorage-controller failed to watch logstorage Tigerastatus: %w", err)
 	}
 
+	if err = utils.AddAPIServerWatch(c); err != nil {
+		return fmt.Errorf("intrusiondetection-controller failed to watch APIServer resource: %v", err)
```
asincu, line 256: Copy-paste error.

pkg/render/logstorage/linseed/linseed.go
```diff
@@ -429,6 +432,112 @@ func (l *linseed) linseedAllowTigeraPolicy() *v3.NetworkPolicy {
...
+			Source:      render.ECKOperatorSourceEntityRule,
+			Destination: linseedIngressDestinationEntityRule,
+		},
+		{
+			Action:      v3.Allow,
+			Protocol:    &networkpolicy.TCPProtocol,
+			Source:      esmetrics.ESMetricsSourceEntityRule,
+			Destination: linseedIngressDestinationEntityRule,
+		},
+		// TODO: Should we delete this ?
```
caseydavenport, line 515: Yeah, I think we probably should delete this and anywhere else that is referencing the DPISourceEntityRule - it's not doing what it thinks it is doing.

caseydavenport (approved):

PR #2833, opened 2023-08-28
Multi-tenant management clusters

caseydavenport:
## Description

For Calico Enterprise / Cloud, MVP of the ability to run management clusters housing multiple tenants. This PR includes a major refactor of the Elasticsearch controller in order to enable this.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


pkg/controller/apiserver/apiserver_controller.go
```diff
@@ -305,6 +307,7 @@ func (r *ReconcileAPIServer) Reconcile(ctx context.Context, request reconcile.Re
 		}
 
 		if managementCluster != nil {
+			// TODO: Does this belong here or in the manager?
```
rene-dekker, line 310: Both the apiserver and the manager have use for this secret. The apiserver allows users to create a managed cluster and obtain an installation manifests. I believe since Manager has a dependency on the ApiServer, it would make sense to create it here to prevent another rollout of this deployment.

pkg/controller/certificatemanager/certificatemanager.go
```diff
@@ -111,17 +166,25 @@ func Create(cli client.Client, installation *operatorv1.InstallationSpec, cluste
 		}
 		certificatePEM = certificateManagement.CACert
 	} else {
+		// Using operator-managed certificates. Check to see if we have already provisioned one.
```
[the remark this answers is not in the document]
rene-dekker: On second thought, how about we make the log statement better, so a comment is redundant.

pkg/controller/certificatemanager/certificatemanager.go
```diff
@@ -132,29 +195,38 @@ func Create(cli client.Client, installation *operatorv1.InstallationSpec, cluste
...
+			cm.log.V(2).Info("Found an existing CA secret")
 			privateKeyPEM, certificatePEM = caSecret.Data[corev1.TLSPrivateKeyKey], caSecret.Data[corev1.TLSCertKey]
 			cryptoCA, err = crypto.GetCAFromBytes(certificatePEM, privateKeyPEM)
 			if err != nil {
 				return nil, err
 			}
 		}
 	}
+
+	// At this point, we've located an existing CA or genrated a new one. Build a certificateManager
```
rene-dekker, line 208: generated*

pkg/controller/certificatemanager/certificatemanager.go
```diff
@@ -284,22 +360,23 @@ func (cm *certificateManager) getKeyPair(cli client.Client, secretName, secretNa
...
 				// the old secret will be deleted automatically.
-				return certificateManagementKeyPair(cm, secretName, nil), nil, nil
+				return certificateManagementKeyPair(cm, secretName, secretNamespace, dnsNames), nil, nil
 			}
 
 			if invalidKeyUsage {
 				log.Info("secret %s/%s must specify ext key usages: %+v", secretNamespace, secretName, requiredKeyUsages)
 			}
 			// We return nil, so a new secret will be created for expired (legacy) operator signed secrets.
+			cm.log.V(2).Info("KeyPair is an expired legacy operator cert, make a new one", "name", secretName)
```
rene-dekker, line 379: How about bumping the log level to warning? It may actually be of great value in some edge cases.

pkg/controller/certificatemanager/certificatemanager.go
```diff
@@ -312,13 +389,14 @@ func (cm *certificateManager) getKeyPair(cli client.Client, secretName, secretNa
...
 		if cm.keyPair.CertificateManagement != nil {
-			return certificateManagementKeyPair(cm, secretName, nil), nil, nil
+			return certificateManagementKeyPair(cm, secretName, secretNamespace, dnsNames), nil, nil
 		}
 		if string(x509Cert.AuthorityKeyId) == string(cm.AuthorityKeyId) {
 			issuer = cm.keyPair
 		} else {
 			if !readCertOnly {
 				// We want to return nothing, so a new secret will be created to overwrite this one.
+				cm.log.V(2).Info("KeyPair's authority key id doesn't match", "name", secretName)
```
rene-dekker, line 399: Similarly, perhaps make it warning level, and improve the comment to say that we make a new one.

pkg/controller/installation/core_controller.go
```diff
@@ -1179,15 +1171,17 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
...
+		// es-kube-controllers needs to trust the ESGW certificate. We'll fetch it here and add it to the trusted bundle.
+		// Note that although we're adding this to the typhaNodeTLS trusted bundle, it will be used by es-kube-controllers. This is because
+		// all components within this namespace share a trusted CA bundle.
 		esgwCertificate, err := certificateManager.GetCertificate(r.client, relasticsearch.PublicCertSecret, common.OperatorNamespace())
 		if err != nil {
 			r.status.SetDegraded(operator.CertificateError, fmt.Sprintf("Failed to retrieve / validate  %s", relasticsearch.PublicCertSecret), err, reqLogger)
 			return reconcile.Result{}, err
-		}
-		if esgwCertificate != nil {
+		} else if esgwCertificate != nil {
```
rene-dekker, line 1181: I guess this is a difference of style, but I've learnt that else should not be used if the if-case returns/breaks.

pkg/controller/logcollector/logcollector_controller.go
```diff
@@ -405,36 +409,49 @@ func (r *ReconcileLogCollector) Reconcile(ctx context.Context, request reconcile
...
+		// Instead, look for a Tenant instance that represent's the management cluster's own tenant.
+		if instance.Spec.MultiTenantManagementClusterNamespace == "" {
+			r.status.SetDegraded(operatorv1.ResourceValidationError, "multiTenantManagementClusterNamespace is not set", nil, reqLogger)
+			return reconcile.Result{}, nil
+		}
+		tenant, _, err = utils.GetTenant(ctx, r.multiTenant, r.client, instance.Spec.MultiTenantManagementClusterNamespace)
+		if err != nil {
+			r.status.SetDegraded(operatorv1.ResourceNotReady, fmt.Sprintf("Failed to retrieve tenant in ns %s", instance.Spec.MultiTenantManagementClusterNamespace), err, reqLogger)
+			return reconcile.Result{}, err
+		}
```
rene-dekker, line 439: Isn't tenant.Namespace alsways equal to instance.Spec.MultiTenantManagementClusterNamespace?

pkg/controller/namespace_helper.go
```diff
@@ -0,0 +1,92 @@
...
+	// will be the tenant's namespace.
+	InstallNamespace() string
+
+	// TruthNamespace returns the namespace to use as the source of truth for storing data.
+	// For single-tenant installs, this is the tigera-operator namespace.
+	// For multi-tenant installs, this is tenant's namespace.
+	TruthNamespace() string
+
+	// BothNamespaces returns both the truth namespace and the install namespace.
+	BothNamespaces() []string
```
rene-dekker, line 52: It is used twice, the name is ambiguous. I think the reader of the code will be much better off if this would not exist, so it does not have to read this description.

pkg/controller/namespace_helper.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+}
+
+type NamespaceHelper interface {
+	// InstallNamespace returns the namespace that components will be installed into.
+	// for single-tenant clusters, this is tigera-manager. For multi-tenancy, this
+	// will be the tenant's namespace.
+	InstallNamespace() string
+
+	// TruthNamespace returns the namespace to use as the source of truth for storing data.
```
rene-dekker, line 46: I've always thought of this namespace as the place to read user provided data from. Mostly secrets, as other input is typically and ideally in our CRDs.
I still don't like the name  TruthNamespace, I cannot (yet) get used to it.

pkg/controller/logstorage/elastic/elastic_controller.go
```diff
@@ -0,0 +1,734 @@
...
+
+package elastic
+
+import (
+	"context"
+	"fmt"
+	"net/url"
+	"time"
+
+	cmnv1 "github.com/elastic/cloud-on-k8s/v2/pkg/apis/common/v1"
```
rene-dekker, line 23: Imports should be reformatted.

pkg/controller/logstorage/elastic/elastic_controller.go
```diff
@@ -0,0 +1,734 @@
...
+		return nil
+	}
+
+	// Create the reconciler
+	r := &ElasticSubController{
+		client:         mgr.GetClient(),
+		scheme:         mgr.GetScheme(),
+		esCliCreator:   utils.NewElasticClient,
+		tierWatchReady: &utils.ReadyFlag{},
+		status:         status.New(mgr.GetClient(), "log-storage-elastic", opts.KubernetesVersion),
```
rene-dekker, line 98: "log-storage-elastic" should be a const

pkg/controller/logstorage/elastic/elastic_controller.go
```diff
@@ -0,0 +1,734 @@
...
+	for _, secretName := range []string{
+		render.TigeraElasticsearchGatewaySecret, render.TigeraKibanaCertSecret,
+		render.OIDCSecretName, render.DexObjectName, esmetrics.ElasticsearchMetricsServerTLSSecret,
+		render.TigeraLinseedSecret, certificatemanagement.CASecretName,
+	} {
+		if err = utils.AddSecretsWatch(c, secretName, common.OperatorNamespace()); err != nil {
+			return fmt.Errorf("log-storage-elastic-controller failed to watch Secret resource: %w", err)
+		}
+	}
+	if err = utils.AddSecretsWatch(c, relasticsearch.PublicCertSecret, common.OperatorNamespace()); err != nil {
```
rene-dekker, line 192: I think all the secrets with OperatorNamespace can be added to the for-loop.

pkg/controller/logstorage/elastic/elastic_controller.go
```diff
@@ -0,0 +1,734 @@
...
+	if opts.MultiTenant {
+		if err = utils.AddSecretsWatch(c, certificatemanagement.TenantCASecretName, ""); err != nil {
+			return fmt.Errorf("log-storage-elastic-controller failed to watch Secret resource: %w", err)
+		}
+	}
+
+	for _, name := range []string{
+		render.OIDCUsersConfigMapName,
+		render.OIDCUsersEsSecreteName,
+		render.ElasticsearchAdminUserSecret,
```
rene-dekker, line 222: This is a secret, not a cm

pkg/controller/logstorage/elastic/elastic_controller.go
```diff
@@ -0,0 +1,734 @@
...
+	}
+	if opts.MultiTenant {
+		if err = utils.AddSecretsWatch(c, certificatemanagement.TenantCASecretName, ""); err != nil {
+			return fmt.Errorf("log-storage-elastic-controller failed to watch Secret resource: %w", err)
+		}
+	}
+
+	for _, name := range []string{
+		render.OIDCUsersConfigMapName,
+		render.OIDCUsersEsSecreteName,
```
rene-dekker, line 221: This is a secret, also could you correct the typo?

pkg/controller/logstorage/elastic/elastic_controller.go
rene-dekker: In general, I think the watches in this func can be refactored into a few for-loops

pkg/controller/manager/manager_controller.go
```diff
@@ -81,123 +82,116 @@ func Add(mgr manager.Manager, opts options.AddOptions) error {
 		return err
 	}
 
-	err = utils.AddSecretsWatch(controller, render.VoltronLinseedTLS, render.ManagerNamespace)
-	if err != nil {
-		return err
+	// Determine how to handle watch events for cluster-scoped resources. For multi-tenant clusters,
+	// we should update all tenants whenever one changes. For single-tenatn clusters, we can just queue the object.
```
rene-dekker, line 86: typo: tenant*

pkg/controller/utils/tenant_event_handler.go
```diff
@@ -0,0 +1,89 @@
...
+func EnqueueAllTenants(c client.Client) handler.EventHandler {
+	return &enqueueAllTenants{client: c}
+}
+
+// enqueueAllTenants queues updates for all tenants in the cluster.
+type enqueueAllTenants struct {
+	client client.Client
+}
+
+// Create is called in response to an create event - e.g. Pod Creation.
```
rene-dekker, line 42: typo: a*

pkg/controller/utils/tenant_event_handler.go
```diff
@@ -0,0 +1,89 @@
...
+
+// Generic is called in response to an event of an unknown type or a synthetic event triggered as a cron or
+// external trigger request - e.g. reconcile Autoscaling, or a Webhook.
+func (h *enqueueAllTenants) Generic(e event.GenericEvent, q workqueue.RateLimitingInterface) {
+	elog.V(2).Info("Generic event triggered reconciliation for all tenants", "event", e)
+	h.enqueue(q)
+}
+
+func (h *enqueueAllTenants) enqueue(q workqueue.RateLimitingInterface) {
+	tenants := h.getAllTenants()
```
rene-dekker, line 68: unnecessary variable declaration

pkg/controller/utils/tenant_event_handler.go
```diff
@@ -0,0 +1,89 @@
...
+	allTenants := operatorv1.TenantList{}
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+	err := h.client.List(ctx, &allTenants)
+	if err != nil {
+		elog.Error(err, "Error querying tenants, cannot trigger Reconcile")
+		return nil
+	}
+
+	names := []types.NamespacedName{}
```
rene-dekker, line 84: Goland: "empty slice declaration using literal", best to replace with
var names []types.NamespacedName

pkg/controller/logstorage/users/user_controller.go
```diff
@@ -0,0 +1,254 @@
...
+import (
+	"context"
+	"crypto/rand"
+	"encoding/base64"
+	"fmt"
+	"time"
+
+	"github.com/go-logr/logr"
+
+	esv1 "github.com/elastic/cloud-on-k8s/v2/pkg/apis/elasticsearch/v1"
```
rene-dekker, line 26: Imports could be reformatted.

pkg/controller/logstorage/users/user_controller.go
```diff
@@ -0,0 +1,254 @@
...
+	status      status.StatusManager
+	multiTenant bool
+}
+
+func Add(mgr manager.Manager, opts options.AddOptions) error {
+	if !opts.EnterpriseCRDExists {
+		return nil
+	}
+	if !opts.MultiTenant {
+		// TODO: For now, the operator only creates users in multi-tenant mode.
```
rene-dekker, line 64: Is there really something we intend to do here?

pkg/controller/logstorage/users/user_controller.go
```diff
@@ -0,0 +1,254 @@
...
+	user.Password = password
+	if err = esClient.CreateUser(ctx, user); err != nil {
+		r.status.SetDegraded(operatorv1.ResourceUpdateError, "Failed to create or update Elasticsearch user", err, reqLogger)
+		return err
+	}
+
+	return nil
+}
+
+func randomPassword(length int) (string, error) {
```
rene-dekker, line 249: We have a function in the code already for generating a password: GeneratePassword() I think it is best to have only one.

! Passwords should be created using crypto/rand

pkg/controller/logstorage/users/user_controller.go
```diff
@@ -0,0 +1,254 @@
...
+func (r *UserController) createLinseedLogin(ctx context.Context, tenantID string, secret *corev1.Secret, reqLogger logr.Logger) error {
+	// ES should be in ready phase when execution reaches here, apply ILM polices
+	esClient, err := utils.NewElasticClient(r.client, ctx, relasticsearch.ElasticEndpoint())
+	if err != nil {
+		r.status.SetDegraded(operatorv1.ResourceCreateError, "Failed to connect to Elasticsearch - failed to create the Elasticsearch client", err, reqLogger)
+		return err
+	}
+
+	// Determine the password from the secret.
+	password := secret.StringData["password"]
```
rene-dekker, line 230:
```
	// stringData allows specifying non-binary secret data in string form.
	// It is provided as a write-only input field for convenience.
	// All keys and values are merged into the data field on write, overwriting any existing values.
	// The stringData field is never output when reading from the API.
```

pkg/controller/secrets/tenant_controller.go
```diff
@@ -0,0 +1,184 @@
...
+	cm, err := certificatemanager.Create(r.client, installation, r.clusterDomain, tenant.Namespace, opts...)
+	if err != nil {
+		r.status.SetDegraded(operatorv1.ResourceCreateError, "Unable to create CA", err, logc)
+		return reconcile.Result{}, err
+	}
+	cm.AddToStatusManager(r.status, tenant.Namespace)
+
+	// Create the CA in the tenant's namespace.
+	keyPairOptions := []rcertificatemanagement.KeyPairOption{
+		rcertificatemanagement.NewKeyPairOption(cm.KeyPair(), true, true),
```
rene-dekker, line 147: I think this will render the secret twice in the same namespace., since TenantNamepace == TruthNamespace

pkg/controller/secrets/tenant_controller.go
```diff
@@ -0,0 +1,184 @@
...
+	}
+
+	// Create a trusted bundle for this tenant. This bundle is provided to the tenant pods so that they can verify
+	// each other's certificates. Each tenant needs to trust:
+	// - Certificates signed by its own CA
+	// - Certificates signed by the cluster-scoped Tigera CA
+	trustedBundle := cm.CreateTrustedBundle()
+	trustedBundle.AddCertificates(clusterCA)
+
+	// TODO: Provision a trusted bundle that includes system certificates for components that need it.
```
rene-dekker, line 164: This seems too important to address before merging.

pkg/controller/apiserver/apiserver_controller.go
caseydavenport: In multi-tenancy, this is a per-tenant secret - it's not just one per management cluster, and so it needs to be done in the manager controller which is what led me to this question.

I think it's OK to restart this pod if needed, though obviously would be nice to avoid it if possible. Either way, I think some cleanup can be done here in a future PR even if we keep provisioning it in two places

pkg/controller/logcollector/logcollector_controller.go
caseydavenport: yeah, this check is to confirm that a real Tenant CR actually exists - it doesn't expect the value to be any different.

pkg/controller/namespace_helper.go
caseydavenport: I guess so. It will be used in more places when we roll out multi-tenancy to more subsystems, and avoids reproducing some boilerplate. I'll get rid of it for now anyway.

pkg/controller/namespace_helper.go
caseydavenport: I think it's the best description of the purpose of the tigera-operator namespace. It's the namespace users can put configuration, sure. But it's also the namespace that we populate if the user didn't provide any information, and that information always flows from tigera-operator -> any namespace that needs it.

Controllers compare the value of secrets in the tigera-operator namespace with the "live" version of that secret in the namespace, and if they don't match they replace it with the one from the tigera-operator namespace. It really is the "source of truth" namespace.

pkg/controller/logstorage/elastic/elastic_controller.go
caseydavenport: How so? goimports is telling me this is correct.

pkg/controller/logstorage/elastic/elastic_controller.go
caseydavenport: To be honest, I was planning on getting rid of the for loop anyway. This mix/match pre-dates my changes and I'd like to refactor to make it all consistent.

pkg/controller/logstorage/users/user_controller.go
caseydavenport: Yeah, there's a backlog item for it.

pkg/controller/logstorage/users/user_controller.go
caseydavenport: Yep. The code above may pass either a queried secret (in which case the data is in `Data` or a freshly created one, in which case the data is in `StringData`

pkg/controller/secrets/tenant_controller.go
caseydavenport: Nope - we're merging to prevent rotting of this branch vs. master. This codepath only impacts the multi-tenant scenario, which is largely working in this PR but still has several TODOs (including this one) before it is production ready.

The main goal is to get the refactor in and testing the mainline single-tenant path.

pkg/controller/logstorage/users/user_controller.go
rene-dekker: I think we best refrain from using this "convenience" function if we want to read the contents before writing.

pkg/controller/logstorage/users/user_controller.go
```diff
@@ -0,0 +1,254 @@
...
+		return reconcile.Result{RequeueAfter: 10 * time.Second}, nil
+	}
+
+	// Query any existing username and password for this Linseed instance. If one already exists, we'll simply
+	// use that. Otherwise, generate a new one.
+	linseedUser := utils.LinseedUser(tenantID)
+	basicCreds := corev1.Secret{}
+	credentialSecrets := []client.Object{}
+	key = types.NamespacedName{Name: render.ElasticsearchLinseedUserSecret, Namespace: helper.TruthNamespace()}
+	if err := r.client.Get(ctx, key, &basicCreds); err != nil && !errors.IsNotFound(err) {
```
rene-dekker, line 172: We have utils.GetSecret() in this repo already that would make this code a bit cleaner.

pkg/controller/logstorage/users/user_controller.go
```diff
@@ -0,0 +1,240 @@
...
+		reqLogger.Info("No Tenant in this Namespace, skip")
+		return reconcile.Result{}, nil
+	} else if err != nil {
+		r.status.SetDegraded(operatorv1.ResourceReadError, "An error occurred while querying Tenant", err, reqLogger)
+		return reconcile.Result{}, err
+	}
+
+	// Get LogStorage resource.
+	logStorage := &operatorv1.LogStorage{}
+	key := utils.DefaultTSEEInstanceKey
```
rene-dekker, line 132: No need for a variable declaration here.

pkg/controller/logcollector/logcollector_controller.go
rene-dekker: It took me  a little while to understand this block of code. If we would move the code to get the tenant above this if-else block, I think  it could improve the readability.

pkg/controller/logstorage/users/user_controller.go
caseydavenport: We're not reading the contents before writing. We do this:

1. Get the secret if it exists.
2. If it doesn't exist, make a new one.
3. Write it to the API.
4. Assuming it was successful, create the user in ES

This function is step 4 - we're just avoiding the need to wait for our watches to sync / re-query the secret, since we just wrote it, and so we know exactly what the values are.

If we reach step 4 via step 2, then `StringData` will have the fields present.

If we reach step 4 via step 1, then `Data` will have them present.

pkg/controller/logstorage/users/user_controller.go
caseydavenport: If I had my way, `utils.GetSecret` would be removed and we'd do everything this way... 

I find it an unnecessary abstraction that just intercepts and hides a particular type of error leading to code that is confusingly different from all of the other similar places that we query information from the k8s API. Many many times I have found or written in bugs myself as a result of this helper function, because I expect `errors.IsNotFound` to work, but it doesn't. 

We haven't written that helper function for all of the other types that we might want to query (ConfigMap, ManagementClusterConnection, DaemonSet, etc., etc.,) and of course we shouldn't.

TL;DR - utils.GetSecret shouldn't exist IMO, and I may make a PR to actually eliminate it in the near future assuming I can get people on board with it.

pkg/controller/logcollector/logcollector_controller.go
caseydavenport: I'm not sure I understand the suggestion. I don't think we can move the tenant querying code because we need to query for the tenant referenced in the `instance.Spec.MultiTenantManagementClusterNamespace`. To do that, we must first check that field is actually set.

We only want to query the tenant and check that field if this is indeed a multi-tenant management cluster. Which sort of places some limitations on where we can actually perform this query.

Maybe a better comment explaining all of this would help?

caseydavenport: I moved it around a little bit and I think improved the comment. LMK what you think

pkg/controller/logstorage/elastic/elastic_controller.go
rene-dekker: If you remove the empty lines and sort, and then place empty lines back in the right places, you'll see what I mean.

caseydavenport: I can do that, but I prefer to keep the API packages all together. I'm separating by function rather than strictly which repository they come from.

pkg/controller/logstorage/users/user_controller.go
rene-dekker: My concern remains that the use of StringData has made the code more complicated.

pkg/controller/logstorage/users/user_controller.go
rene-dekker: I suppose we can agree to disagree about the convenience of the function.

But one immediate question that I have when reading the code the way it is, is why is basicCreds redefined? Isn't that redundant?

pkg/controller/logstorage/users/user_controller.go
caseydavenport: It is marginally more complicated in this particular function - it's an extra three lines of code in a helper function - to avoid complexity elsewhere. Namely, we would need additional logic in the calling code to handle this instead.

pkg/controller/logstorage/users/user_controller.go
```diff
@@ -0,0 +1,239 @@
...
+	if elasticsearch == nil || elasticsearch.Status.Phase != esv1.ElasticsearchReadyPhase {
+		r.status.SetDegraded(operatorv1.ResourceNotReady, "Waiting for Elasticsearch cluster to be operational", nil, reqLogger)
+		return reconcile.Result{RequeueAfter: 10 * time.Second}, nil
+	}
+
+	// Query any existing username and password for this Linseed instance. If one already exists, we'll simply
+	// use that. Otherwise, generate a new one.
+	linseedUser := utils.LinseedUser(tenantID)
+	basicCreds := corev1.Secret{}
+	credentialSecrets := []client.Object{}
```
rene-dekker, line 169: Should be replaced with empty slice declaration.

pkg/controller/logstorage/users/user_controller.go
```diff
@@ -0,0 +1,239 @@
...
+		return err
+	}
+
+	// Determine the password from the secret.
+	password := secret.StringData["password"]
+	if password == "" {
+		password = string(secret.Data["password"])
+	}
+	if password == "" {
+		return fmt.Errorf("Unable to find password in secret")
```
rene-dekker, line 227: Message should start with lowercase

pkg/controller/logstorage/users/user_controller.go
caseydavenport: Other options:

1. Just error out the first time this is called, logging an error and degrading status, wait for the retry to fix it.

2. Update the calling code logic to short-circuit if this is the first time this has been called. For example:

```go

creatingSecret  := false
if err := r.client.Get(ctx, key, &basicCreds); err != nil && !errors.IsNotFound(err) {
  r.status.SetDegraded(operatorv1.ResourceReadError, fmt.Sprintf("Error getting Secret %s", key), err, reqLogger)
} else if errors.IsNotFound(err) {
  creatingSecret = true
}
...

if err := hdler.CreateOrUpdateOrDelete(ctx, credentialComponent, r.status); err != nil {
  r.status.SetDegraded(operatorv1.ResourceUpdateError, "Error creating / updating Linseed user secret", err, reqLogger)
}

if creatingSecret {
  // We just created the secret. Return early, and we'll provision it to ES the next time reconcile is called.
  return reconcile.Result{Requeue: 5 * time.Second}, nil
}

// Create the user in ES
...

```

3. Pass the username and password to this function explicitly instead of the secret

```go

username := linseedUser.Username
password := "" // To be filled in below.
if err := r.client.Get(ctx, key, &basicCreds); err != nil && !errors.IsNotFound(err) {
  r.status.SetDegraded(operatorv1.ResourceReadError, fmt.Sprintf("Error getting Secret %s", key), err, reqLogger)
} else if errors.IsNotFound(err) {
  // Create the secret
  password = RandomPassword(16)
} else {
  // The secret already exists. Get the password from it.
  password = string(basicCreds.Data["password"])
}
...

// Create the user in ES
createLinseedLogin(username, password)
...

```

Of these, option 3 is the only one that I think could be considered less complicated than how it is currently implemented. I can make that switch, but I still very slightly prefer the way it is now because the complexity is limited to one small code block within a helper function.

pkg/controller/logstorage/users/user_controller.go
caseydavenport: Yeah, we could probably do without the reassignment. I usually do that sort of things to be safe with regards to side-effects of the calling code, as I don't know what guarantees it provides regarding inputs. In this case, I think we can be pretty confident that it's not modifying fields on `basicCreds` in the case of an error, though.

pkg/render/common/test/testing.go
```diff
@@ -88,7 +90,55 @@ func ExpectResourceInList(objs []client.Object, name, ns, group, version, kind s
...
+			))
+	}
+	if ns == "" {
+		return fmt.Errorf("%T %s not found in:\n\n%s", expected, name, strings.Join(items, "\n"))
+	}
+	return fmt.Errorf("%T %s/%s not found in:\n\n%s", expected, ns, name, strings.Join(items, "\n"))
+}
+
+// CompareResource checks that the given resource matches the expected name, namespace, group, version, and kind.
+func CompareResource(resource runtime.Object, name, ns, group, version, kind string) {
```
tmjd, line 141: I'm not a big fan of this name change. Compare* suggests to me that it would be providing a result. Expect* suggests it will result in an Expect call. Like I think ExpectResourceMeta or something like that would be more informative.

pkg/render/dex.go
```diff
@@ -426,13 +426,13 @@ func (c *dexComponent) allowTigeraNetworkPolicy() *v3.NetworkPolicy {
 				{
 					Action:      v3.Allow,
 					Protocol:    &networkpolicy.TCPProtocol,
-					Source:      ManagerSourceEntityRule,
+					Source:      networkpolicy.Helper(false, "").ManagerSourceEntityRule(),
```
tmjd, line 429: I think this is the same as the DefaultHelper, right?

pkg/render/logstorage/esgateway/esgateway.go
```diff
@@ -401,7 +404,7 @@ func (e *esGateway) esGatewayAllowTigeraPolicy() *v3.NetworkPolicy {
 				{
 					Action:      v3.Allow,
 					Protocol:    &networkpolicy.TCPProtocol,
-					Source:      render.ManagerSourceEntityRule,
+					Source:      networkpolicy.Helper(false, "").ManagerSourceEntityRule(),
```
tmjd, line 407: I think this is the same as the DefaultHelper.

tmjd (commented):
I'd like you to consider handling status differently. I don't think it is great that were adding a bunch of new tigerastatus for very specific components. I think I'd like to see a sub status interface that feeds into an aggregate status. Then the controllers/logstorage_controller.go could instantiate the aggregate and pass sub status interfaces to each of the "sub" controllers.
There are probably other possible options too.

I also don't think we should be so liberal with the RequeueAfter usage, 1 minute seems excessive for any of our controllers for the happy state. I think a minimum we should use is 5 minutes. And I think anywhere we Requeue we should aim to have only one requeue running for a single tenant reconcile or one per tenant for a multi-tenant reconcile. It is difficult but we can probably get close by using RequeueAfter only when a certain named resource is being handled.

api/v1/tenant_types.go
```diff
@@ -0,0 +1,54 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type TenantSpec struct {
+	// ID is the unique identifier for this tenant.
+	// +required
+	ID string `json:"id,omitempty"`
+}
+
+type TenantStatus struct{}
```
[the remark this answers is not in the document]
caseydavenport: We should probably have a detailed discussion for status handling here before I implement anything too far. There are a few ways we could go:

- Namespaced TigeraStatus objects, inside the tenant's namespace.
- Global TigeraStatus objects, no "tenant" TS, but adding status to the Tenant object itself
- Just don't worry about status for the Tenant API since it's covered elsewhere (I think?)

pkg/controller/certificatemanager/certificatemanager.go
```diff
@@ -82,18 +91,64 @@ type CertificateManager interface {
 	AddToStatusManager(manager status.StatusManager, namespace string)
 	// KeyPair Returns the CA KeyPairInterface, so it can be rendered in the operator namespace.
 	KeyPair() certificatemanagement.KeyPairInterface
+	// Loads an existing trusted bundle to pass to render.
+	LoadTrustedBundle(context.Context, client.Client, string) (certificatemanagement.TrustedBundleRO, error)
+}
+
+type Option func(cm *certificateManager) error
+
+func AllowCACreation() Option {
```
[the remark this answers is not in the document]
caseydavenport: I suppose it might be cleaner to have a "CreateCA" function that is distinct from this, and just always prevent creation of the CA from this function. That's probably better.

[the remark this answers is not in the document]
caseydavenport: I think I will try this as a separate PR from other feedback though - this one is a bit gnarly to unpick and this code scares me a bit.

api/v1/logcollector_types.go
```diff
@@ -36,6 +36,11 @@ type LogCollectorSpec struct {
 	// +optional
 	// +kubebuilder:validation:Enum=Enabled;Disabled
 	CollectProcessPath *CollectProcessPathOption `json:"collectProcessPath,omitempty"`
+
+	// If running as a multi-tenant management cluster, the namespace in which
+	// the management cluster's tenant services are running.
+	// +optional
+	MultiTenantManagementClusterNamespace string `json:"multiTenantManagementClusterNamespace,omitempty"`
```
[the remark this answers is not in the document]
caseydavenport: > we already expect the tenant name to match the tenant namespace so I don't think adding namespace adds anything here

This is certainly not true - we expect the tenant name to always be "default".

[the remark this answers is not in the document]
caseydavenport: https://github.com/tigera/operator/blob/7a5426ff11ffa2fdf7a814aa07a87d5830bfe33d/pkg/controller/utils/utils.go#L427-L432

pkg/controller/logstorage/elastic/elastic_controller.go
caseydavenport: @tmjd I must be out of the loop then, because this is the way I have tried to organize imports for a long long time. What is the organization scheme you are referring to because it's never been communicated to me!

pkg/controller/namespace_helper.go
```diff
@@ -0,0 +1,82 @@
...
+	return &namespacer{
+		multiTenant:           mt,
+		singleTenantNamespace: singleTenantNS,
+		multiTenantNamespace:  multiTenantNS,
+	}
+}
+
+type NamespaceHelper interface {
+	// InstallNamespace returns the namespace that components will be installed into.
+	// for single-tenant clusters, this is tigera-manager. For multi-tenancy, this
```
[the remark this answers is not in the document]
caseydavenport: The comment is out of date from a much much earlier version, will fix.

pkg/controller/logstorage/initializer/initializing_controller.go
```diff
@@ -0,0 +1,252 @@
...
+type LogStorageInitializer struct {
+	client      client.Client
+	scheme      *runtime.Scheme
+	status      status.StatusManager
+	provider    operatorv1.Provider
+	multiTenant bool
+}
+
+// FillDefaults populates the default values onto an LogStorage object.
+func FillDefaults(opr *operatorv1.LogStorage) {
```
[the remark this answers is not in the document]
caseydavenport: I think because it is used in several other test suites

pkg/controller/logstorage/secrets/secret_controller.go
```diff
@@ -0,0 +1,610 @@
...
+		return reconcile.Result{}, nil
+	}
+	trustedBundle := keyPairs.trustedBundle(appCM)
+
+	if err = hdler.CreateOrUpdateOrDelete(ctx, keyPairs.component(trustedBundle, helper), r.status); err != nil {
+		r.status.SetDegraded(operatorv1.ResourceUpdateError, "Error creating / updating resource", err, reqLogger)
+		return reconcile.Result{}, err
+	}
+	r.status.ClearDegraded()
+	return reconcile.Result{RequeueAfter: 60 * time.Second}, nil
```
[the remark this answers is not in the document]
caseydavenport: > and currently at 1 minute it seems like a level that we expect the requeue to be needed for normal operation.

@tmjd This particular rhetoric doesn't makes sense to me. 1 minute isn't any more or less "for normal operation" than 5 or 30 minutes is. The value here should suggest only one thing - how quickly do we want to react in the face of a bug?

So, instead of debating magic numbers, I think we have a few questions to answer first:

- Do we care about this case at all? If not, we just shouldn't requeue.
- If we do care, what is a reasonable amount of time that we can retry that will actually mitigate cluster issues?
- What value is too small, and what is the reason it is too small?

From there, assuming we want to keep this at all, we should pick the smallest value for this number that we possibly can without incurring unreasonable downsides.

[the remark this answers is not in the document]
caseydavenport: I think I am in favor of simply removing RequeueAfter altogether, given the above.

pkg/controller/logstorage/initializer/initializing_controller.go
```diff
@@ -0,0 +1,252 @@
...
+const (
+	DefaultElasticsearchStorageClass = "tigera-elasticsearch"
+	TigeraStatusName                 = "log-storage"
+	defaultEckOperatorMemorySetting  = "512Mi"
+)
+
+// Add creates a new LogStorage Controller and adds it to the Manager. The Manager will set fields on the Controller
+// and Start it when the Manager is Started.
+func Add(mgr manager.Manager, opts options.AddOptions) error {
+	if !opts.EnterpriseCRDExists {
```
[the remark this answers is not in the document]
caseydavenport: Yeah, I think it makes sense to do it here.

pkg/controller/logstorage/linseed/linseed_controller.go
```diff
@@ -0,0 +1,420 @@
...
+	if err = r.client.Get(ctx, key, &corev1.Secret{}); err != nil && !errors.IsNotFound(err) {
+		r.status.SetDegraded(operatorv1.ResourceReadError, fmt.Sprintf("Error getting Secret %s", key), err, reqLogger)
+		return reconcile.Result{}, err
+	} else if errors.IsNotFound(err) {
+		r.status.SetDegraded(operatorv1.ResourceNotFound, fmt.Sprintf("Waiting for Linseed credential Secret %s", key), err, reqLogger)
+		return reconcile.Result{RequeueAfter: 10 * time.Second}, nil
+	}
+
+	dpiList := &v3.DeepPacketInspectionList{}
+	if err := r.client.List(ctx, dpiList); err != nil {
```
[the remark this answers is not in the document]
caseydavenport: Some day, certain parts of it probably will be. But not for this iteration.

pkg/controller/logstorage/elastic/elastic_controller.go
```diff
@@ -0,0 +1,741 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+var log = logf.Log.WithName("controller_logstorage_elastic")
+
+const (
+	LogStorageFinalizer = "tigera.io/eck-cleanup"
+	tigeraStatusName    = "log-storage=-elastic"
```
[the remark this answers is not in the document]
caseydavenport: yeah this has already been fixed in master

pkg/controller/logstorage/kubecontrollers/es_kube_controllers.go
```diff
@@ -0,0 +1,330 @@
...
+	}
+
+	// Get the Authentication resource.
+	authentication, err := utils.GetAuthentication(ctx, r.client)
+	if err != nil && !errors.IsNotFound(err) {
+		r.status.SetDegraded(operatorv1.ResourceReadError, "Error while fetching Authentication", err, reqLogger)
+		return reconcile.Result{}, err
+	}
+
+	pullSecrets, err := utils.GetNetworkingPullSecrets(install, r.client)
```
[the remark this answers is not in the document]
caseydavenport: This has already been fixed in master

pkg/controller/utils/utils.go
```diff
@@ -79,6 +82,7 @@ func IgnoreObject(obj runtime.Object) bool {
 	return false
 }
 
+// TODO: Deprecate and delete these functions.
```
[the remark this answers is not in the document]
caseydavenport: At a minimum these need to be updated to include the particular handler that should be used, instead of assuming EnqueueRequestForObject.

So the new signatures will be:

```go
if err := AddInstallationWatch(c, &handler.EnqueueRequestForObject{})
```

instead of, for example:

```go
if err := c.Watch(&source.Kind{Type: &operatorv1.Installation{}), &handler.EnqueueRequestForObject{})
```

I guess there is a small readability difference, at the cost of it being tedious to make changes (especially per-controller changes) when needed. I added this TODO because the use of these particular "helpers" wasted time for me at some point throughout this changeset. 

I'll remove the TODO for now since we're not in agreement, but I'm in general pretty unhappy with the proliferation of functions that simply wrap another function. I've said it before in the context of the "GetSecret" and "GetConfigMap" functions that just hide certain errors as well.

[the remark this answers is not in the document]
caseydavenport: They're not even consistently implemented - there's only `AddNetworkWatch`, `AddComplianceWatch`, and `AddAPIServerWatch`

I definitely would not say there is value in going through the code and adding equivalent functions for every other thing that we watch.

pkg/render/common/test/testing.go
caseydavenport: Yeah I agree. I'd like to move all the tests off of this function eventually but for now I will change the name to `ExpectResourceTypeAndObjectMeta` or something.

pkg/controller/logstorage/elastic/elastic_controller.go
tmjd: I thought the organization was 3 groups in the following order:
1. language built-in packages
2. 3rd party packages
3. tigera/calico packages
I thought we separated each of those section with a new line. Currently this PR has 5 groups of packages.

api/v1/tenant_types.go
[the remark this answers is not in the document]
tmjd: I'm totally on board with having a discussion on the status handling.

pkg/controller/logstorage/secrets/secret_controller.go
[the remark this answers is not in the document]
tmjd: I am good with removing retries. The only reason in my mind is to ensure that any resources that operator creates, that we don't watch, will be corrected eventually if the user changes them.

To clarify my "normal operation" comment. When I said that I meant as part of normal cluster setup/installation. What that excludes is when a user (or something) modifies a resource that the operator deploys, sorry for not being clear with what I meant.

pkg/controller/logstorage/elastic/elastic_controller.go
caseydavenport: That's certainly not consistently implemented, or even the majority implemented throughout the codebase. I started going through files to clean them up in my other PR but it was overwhelming the number of files that were not correct, at least according to Rene's suggestion above (which I don't think is the same as what you are saying?)

> If you remove the empty lines and sort, and then place empty lines back in the right places, you'll see what I mean.

All of this comes down to my current and continuing believe that we have no standard here other than what goimports and "make fix" are happy with, and it's probably not reasonable to suddenly start requiring an arbitrary standard before we agree what it is and ideally go through and make existing code match.

tmjd: I have been requesting the standard I've stated when I see people adjusting or reconfiguring the includes in the operator. So I wouldn't say I'm suddenly starting to require an arbitrary standard, it has been the one I've always expected in operator.
This is the format that I think Calico has been using which was carried over from when Calico was implemented in Python.

caseydavenport: Sure, but one person requesting a format does not a project standard make.

I'm happy to have a discussion on what the standard should be - we have never had that discussion as far as I am aware - and until we  do I just don't think it's fair to say that we have a project standard.

PR #2840, opened 2023-08-29
Use tenant ID as SAN for tunnel connection

asincu:
## Description

This PR will change the DNS name used for tunnel certificate for Voltron. For a management cluster single-tenant, it will remain "voltron". For multi-tenant it will be be set to the tenant ID. In addition, the managed controller will be the one that created the secret and distributes accordingly. The API server will read the secrets from the corresponding namespaces ("tigera-system" for single-tenant and the tenant namespace for multi-tenant mode). Thus, the api server controller only needs to render it in tigera-system namespace, but not mount it inside the API server. Any parameters needed by the API sever to specify the location are not needed any more. Since the secret name can change, we will also be passing the secret named used for store this credentials.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/apiserver/apiserver_controller.go
+++ b/pkg/controller/apiserver/apiserver_controller.go
@@ -274,6 +274,7 @@ func (r *ReconcileAPIServer) Reconcile(ctx context.Context, request reconcile.Re
 	}
 
 	// Query enterprise-only data.
+	var tunnelCAKeyPair certificatemanagement.KeyPairInterface
 	var trustedBundle certificatemanagement.TrustedBundle
 	var amazon *operatorv1.AmazonCloudIntegration
 	var managementCluster *operatorv1.ManagementCluster
@@ -296,6 +297,26 @@ func (r *ReconcileAPIServer) Reconcile(ctx context.Context, request reconcile.Re
 			err = fmt.Errorf("having both a ManagementCluster and a ManagementClusterConnection is not supported")
 			r.status.SetDegraded(operatorv1.ResourceValidationError, "", err, reqLogger)
 			return reconcile.Result{}, err
+		}
+
+		if managementCluster != nil && !r.multiTenant {
+			// The secret that contains the CA x509 certificate to create client certificates for the managed cluster
+			// is created by the Manager controller in tigera-operator namespace. We will read this secret and make
+			// sure it is available in the same namespace as the API server (tigera-system)
+			// This secret is only created for a management cluster in a multi-cluster setup for a single tenant.
+			// Other cluster types do not require this secret. (Standalone configuration do not need it and multi-tenant
+			// configuration create secrets inside the tenant namespaces)
+			tunnelSecretName := managementCluster.Spec.TLS.SecretName
+			tunnelCASecret, err := utils.GetSecret(ctx, r.client, tunnelSecretName, common.OperatorNamespace())
+			if err != nil {
+				r.status.SetDegraded(operatorv1.ResourceReadError, "Unable to fetch the tunnel secret", err, reqLogger)
+				return reconcile.Result{}, err
+			}
+			if tunnelCASecret == nil {
+				r.status.SetDegraded(operatorv1.ResourceReadError, "Unable to fetch the tunnel secret", err, reqLogger)
+				return reconcile.Result{}, err
+			}
+			tunnelCAKeyPair = certificatemanagement.NewKeyPair(tunnelCASecret, nil, "")
 		}
 
 		if r.amazonCRDExists {
@@ -373,6 +394,7 @@ func (r *ReconcileAPIServer) Reconcile(ctx context.Context, request reconcile.Re
 			ServiceAccounts: []string{render.APIServerServiceAccountName(variant)},
 			KeyPairOptions: []rcertificatemanagement.KeyPairOption{
 				rcertificatemanagement.NewKeyPairOption(tlsSecret, true, true),
+				rcertificatemanagement.NewKeyPairOption(tunnelCAKeyPair, false, true),
 			},
 			TrustedBundle: trustedBundle,
 		}),
--- a/pkg/controller/apiserver/apiserver_controller.go
+++ b/pkg/controller/apiserver/apiserver_controller.go
@@ -144,6 +144,12 @@ func add(c controller.Controller, r *ReconcileAPIServer) error {
 			return fmt.Errorf("apiserver-controller failed to watch primary resource: %v", err)
 		}
 
+		for _, namespace := range []string{common.OperatorNamespace(), rmeta.APIServerNamespace(operatorv1.TigeraSecureEnterprise)} {
+			if err = utils.AddSecretsWatch(c, render.VoltronTunnelSecretName, namespace); err != nil {
+				return fmt.Errorf("apiserver-controller failed to watch the Secret resource: %v", err)
+			}
+		}
+
 		// Watch for changes to authentication
 		err = c.Watch(&source.Kind{Type: &operatorv1.Authentication{}}, &handler.EnqueueRequestForObject{})
 		if err != nil {
--- a/pkg/controller/apiserver/apiserver_controller.go
+++ b/pkg/controller/apiserver/apiserver_controller.go
@@ -144,12 +144,6 @@ func add(c controller.Controller, r *ReconcileAPIServer) error {
 			return fmt.Errorf("apiserver-controller failed to watch primary resource: %v", err)
 		}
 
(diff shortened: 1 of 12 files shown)


pkg/controller/apiserver/apiserver_controller.go
```diff
@@ -306,21 +305,23 @@ func (r *ReconcileAPIServer) Reconcile(ctx context.Context, request reconcile.Re
...
 			if tunnelCASecret == nil {
-				tunnelCASecret, err = certificatemanagement.CreateSelfSignedSecret(render.VoltronTunnelSecretName, common.OperatorNamespace(), "tigera-voltron", []string{"voltron"})
-				if err != nil {
-					r.status.SetDegraded(operatorv1.ResourceCreateError, "Unable to create the tunnel secret", err, reqLogger)
-					return reconcile.Result{}, err
-				}
+				r.status.SetDegraded(operatorv1.ResourceReadError, "Unable to fetch the tunnel secret", err, reqLogger)
+				return reconcile.Result{}, err
 			}
-			tunnelSecretPassthrough = render.NewPassthrough(tunnelCASecret)
```
caseydavenport, line 323: Don't we need this passthrough in order to copy the secret into the right namespace?

pkg/render/apiserver.go
```diff
@@ -197,12 +196,18 @@ func (c *apiServerComponent) Objects() ([]client.Object, []client.Object) {
...
+		if c.cfg.MultiTenant {
+			// Multi-tenant management cluster API servers need access to per-tenant CA secrets in order to sign
+			// per-tenant guardian certificates when creating ManagedClusters.
+			globalObjects = append(globalObjects, c.secretsClusterRole()...)
+		} else {
+			globalObjects = append(globalObjects, c.secretsRole()...)
+		}
 	} else {
-		objsToDelete = append(objsToDelete, c.secretsClusterRole()...)
+		if c.cfg.MultiTenant {
```
caseydavenport, line 208: I don't think this branch is valid - multi-tenant clusters _must_ be management clusters, so you should never hit this else if case

pkg/render/apiserver.go
```diff
@@ -702,6 +707,52 @@ func (c *apiServerComponent) secretsClusterRole() []client.Object {
 	}
 }
 
+// secretsRole provides the tigera API server with the ability to read secrets for tigera-system.
+func (c *apiServerComponent) secretsRole() []client.Object {
+	name := "tigera-extension-apiserver-secrets-access"
```
caseydavenport, line 712: Probably should make this a constant

pkg/render/apiserver.go
```diff
@@ -702,6 +707,52 @@ func (c *apiServerComponent) secretsClusterRole() []client.Object {
...
+	rules := []rbacv1.PolicyRule{
+		{
+			APIGroups: []string{""},
+			Resources: []string{"secrets"},
+			Verbs:     []string{"get"},
+		},
+	}
+
+	return []client.Object{
+		// Return the cluster role itself.
```
caseydavenport, line 723: typo: "cluster role"

caseydavenport (commented):
Looking good - a few minor comments / questions though.

pkg/controller/certificatemanager/certificatemanager.go
```diff
@@ -176,7 +176,7 @@ func Create(cli client.Client, installation *operatorv1.InstallationSpec, cluste
 			cm.log.V(2).Info("No existing CA secret")
 		}
 
-		if len(caSecret.Data) == 0 ||
+		if caSecret == nil || len(caSecret.Data) == 0 ||
```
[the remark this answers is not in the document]
asincu: I do not think so either. I think I was debugging something and just trying out stuff.

pkg/render/apiserver.go
asincu: Not sure I follow. If we are running multi-tenant, we install a role instead of a cluster role. Are you saying that the role is not needed ?

caseydavenport: No, I mean that this branch is logically only ever hit if `c.cfg.ManagementCluster == nil && c.cfg.MultiTenant` - which is not a valid combination. There's no such thing as a "multi-tenant standalone cluster"

pkg/controller/apiserver/apiserver_controller.go
caseydavenport: Where? I don't see that in this file.

pkg/controller/apiserver/apiserver_controller_test.go
```diff
@@ -686,5 +688,123 @@ var _ = Describe("apiserver controller tests", func() {
...
+						Name:      "tigera-secure",
+						Namespace: common.OperatorNamespace(),
+					},
+				})
+				Expect(err).NotTo(HaveOccurred())
+			})
+
+			It("Should reconcile multi-cluster setup for a management cluster for a single tenant", func() {
+				// Create the tunnel secret in operator namespace
+				// This will be created by Manager controller in tigera-operator namespace
```
caseydavenport, line 720: "In a production cluster, this would be created... " instead of "This will be created"

pkg/render/apiserver.go
caseydavenport: In other words, I think this should be this:

```go
} else {
  // If we're not a management cluster, the API server doesn't need permissions to access secrets.
  objsToDelete = append(objsToDelete, c.secretsClusterRole()...)
  objsToDelete = append(objsToDelete, c.secretsRole()...)
}
```

pkg/render/apiserver.go
```diff
@@ -675,7 +679,7 @@ func (c *apiServerComponent) secretsClusterRole() []client.Object {
 		&rbacv1.ClusterRole{
 			TypeMeta: metav1.TypeMeta{Kind: "ClusterRole", APIVersion: "rbac.authorization.k8s.io/v1"},
 			ObjectMeta: metav1.ObjectMeta{
-				Name: name,
+				Name: APIServerSecretsRBACName,
```
caseydavenport, line 682: I know this is pre-existing and not part of your changes, but I think the rules should include a list of valid resource names that the API server is allowed to "get" - we can know this by looking at the full set of namespaces that have tenants in it (which should be passed in from the controller as an argument). Otherwise, the apiserver suddenly has access to any secret in the cluster.

We do a similar thing for cluster role bindings for Linseed: https://github.com/tigera/operator/blob/09be800995e4745154c9cff1250b3ff623be994b/pkg/controller/logstorage/linseed/linseed_controller.go#L365-L369

Feel free to defer this to another PR / a backlog item though - I won't make you fix my own flawed code as part of this one :grin:

pkg/render/apiserver.go
```diff
@@ -684,11 +688,55 @@ func (c *apiServerComponent) secretsClusterRole() []client.Object {
...
+	}
+}
+
+// secretsRole provides the tigera API server with the ability to read secrets from the API server's namespace.
+func (c *apiServerComponent) secretsRole() []client.Object {
+	rules := []rbacv1.PolicyRule{
+		{
+			APIGroups: []string{""},
+			Resources: []string{"secrets"},
+			Verbs:     []string{"get"},
```
caseydavenport, line 715: Same here - we should call out the exact name of the secret that is allowed using ResourceNames

pkg/render/apiserver.go
```diff
@@ -684,11 +688,55 @@ func (c *apiServerComponent) secretsClusterRole() []client.Object {
...
+					Name:      APIServerServiceAccountName(c.cfg.Installation.Variant),
+					Namespace: rmeta.APIServerNamespace(c.cfg.Installation.Variant),
+				},
+			},
+		},
+	}
+}
+
+// secretsRole provides the tigera API server with the ability to read secrets from the API server's namespace.
+func (c *apiServerComponent) secretsRole() []client.Object {
```
caseydavenport, line 710: I think this code would be more clear if we named this `secretsRBAC` and the other one `multiTenantSecretsRBAC` - that way it's clear the distinction between the two.

caseydavenport (commented):
Looks pretty good - just some minor comments

pkg/controller/apiserver/apiserver_controller.go
asincu: @caseydavenport I think it already is in master: https://github.com/tigera/operator/blob/master/pkg/controller/apiserver/apiserver_controller.go#L403

pkg/render/apiserver.go
asincu: I think we can get the secret name from the managementcluster resource. From what I see, linseed uses these namespaces to bind to multiple service accounts. But since there will always be a single API server for all tenants, not sure we need it.

caseydavenport (approved):

PR #2842, opened 2023-08-30
Fix certificate management after multi-tenant changes

caseydavenport:
## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/logstorage/linseed/linseed_controller.go
+++ b/pkg/controller/logstorage/linseed/linseed_controller.go
@@ -43,6 +43,7 @@ import (
 	"github.com/tigera/operator/pkg/controller/status"
 	"github.com/tigera/operator/pkg/controller/utils"
 	"github.com/tigera/operator/pkg/controller/utils/imageset"
+	"github.com/tigera/operator/pkg/dns"
 	"github.com/tigera/operator/pkg/render"
 	relasticsearch "github.com/tigera/operator/pkg/render/common/elasticsearch"
 	"github.com/tigera/operator/pkg/render/common/networkpolicy"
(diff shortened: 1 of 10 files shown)


pkg/controller/logstorage/linseed/linseed_controller.go
```diff
@@ -286,15 +287,16 @@ func (r *LinseedSubController) Reconcile(ctx context.Context, request reconcile.
...
+	linseedKeyPair, err := cm.GetKeyPair(r.client, render.TigeraLinseedSecret, helper.TruthNamespace(), linseedDNSNames)
 	if err != nil {
 		r.status.SetDegraded(operatorv1.ResourceReadError, "Error getting Linseed KeyPair", err, reqLogger)
 		return reconcile.Result{}, err
 	} else if linseedKeyPair == nil {
 		r.status.SetDegraded(operatorv1.ResourceNotFound, fmt.Sprintf("Waiting for Linseed key pair (%s/%s) to exist", helper.TruthNamespace(), render.TigeraLinseedSecret), err, reqLogger)
 		return reconcile.Result{}, nil
 	}
-	tokenKeyPair, err := cm.GetKeyPair(r.client, render.TigeraLinseedTokenSecret, helper.TruthNamespace())
+	tokenKeyPair, err := cm.GetKeyPair(r.client, render.TigeraLinseedTokenSecret, helper.TruthNamespace(), []string{"localhost"})
```
rene-dekker, line 299: I think it is better for the SAN to be render.TigeraLinseedTokenSecret over "localhost".  render.TigeraLinseedTokenSecret is used as the common name already. A signed certificate with localhost in the SAN may pose a larger security risk when breached.

pkg/render/apiserver_test.go
```diff
@@ -1111,7 +1111,7 @@ var _ = Describe("API server rendering tests (Calico Enterprise)", func() {
 			// - 1 added by the operator by default
 			// - 1 added by the calicoNodeDaemonSet override
 			Expect(d.Spec.Template.Annotations).To(HaveLen(2))
-			Expect(d.Spec.Template.Annotations).To(HaveKey("hash.operator.tigera.io/tigera-apiserver-certs"))
+			Expect(d.Spec.Template.Annotations).To(HaveKey("tigera-operator.hash.operator.tigera.io/tigera-apiserver-certs"))
```
rene-dekker, line 1114: I see that in certificatebundle L114 we have the following: `annotations[fmt.Sprintf("%s.hash.operator.tigera.io/%s", cert.GetNamespace(), cert.GetName())] = hash`
And that leads to this value on this line.

I was wondering if a better key for the hash would be: `hash.operator.tigera.io/%s.%s"`

pkg/controller/logstorage/linseed/linseed_controller.go
caseydavenport: yeah, that's fair. The existing implementation in master uses localhost, which is why I kept it as such.

PR #2844, opened 2023-08-30
Operator changes for guardian

MichalFupso:
## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/clusterconnection/clusterconnection_controller.go
+++ b/pkg/controller/clusterconnection/clusterconnection_controller.go
@@ -60,10 +60,6 @@ var log = logf.Log.WithName(controllerName)
 // Add creates a new ManagementClusterConnection Controller and adds it to the Manager. The Manager will set fields on the Controller
 // and start it when the Manager is started. This controller is meant only for enterprise users.
 func Add(mgr manager.Manager, opts options.AddOptions) error {
-	if !opts.EnterpriseCRDExists {
-		// No need to start this controller.
-		return nil
-	}
 	statusManager := status.New(mgr.GetClient(), "management-cluster-connection", opts.KubernetesVersion)
 
 	// Create the reconciler
@@ -191,6 +187,7 @@ func (r *ReconcileConnection) Reconcile(ctx context.Context, request reconcile.R
 	if err != nil {
 		return result, err
 	}
+	isCalicoEnterprise := variant == operatorv1.TigeraSecureEnterprise
 
 	managementCluster, err := utils.GetManagementCluster(ctx, r.Client)
 	if err != nil {
@@ -278,7 +275,13 @@ func (r *ReconcileConnection) Reconcile(ctx context.Context, request reconcile.R
 		trustedCertBundle = certificateManager.CreateTrustedBundle()
 	}
 
-	for _, secretName := range []string{render.PacketCaptureCertSecret, monitor.PrometheusTLSSecretName, render.ProjectCalicoAPIServerTLSSecretName(instl.Variant)} {
+	var certs = []string{}
+	if isCalicoEnterprise {
+		certs = []string{render.PacketCaptureCertSecret, monitor.PrometheusTLSSecretName, render.ProjectCalicoAPIServerTLSSecretName(instl.Variant)}
+	} else {
+		certs = []string{render.ProjectCalicoAPIServerTLSSecretName(instl.Variant)}
+	}
+	for _, secretName := range certs {
 		secret, err := certificateManager.GetCertificate(r.Client, secretName, common.OperatorNamespace())
 		if err != nil {
 			r.status.SetDegraded(operatorv1.ResourceReadError, fmt.Sprintf("Failed to retrieve %s", secretName), err, reqLogger)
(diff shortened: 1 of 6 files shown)


pkg/controller/clusterconnection/clusterconnection_controller.go
```diff
@@ -292,22 +295,22 @@ func (r *ReconcileConnection) Reconcile(ctx context.Context, request reconcile.R
...
-	if !r.tierWatchReady.IsReady() {
+	if !r.tierWatchReady.IsReady() && isCalicoEnterprise {
 		r.status.SetDegraded(operatorv1.ResourceNotReady, "Waiting for Tier watch to be established", nil, reqLogger)
 		return reconcile.Result{RequeueAfter: 10 * time.Second}, nil
 	}
 
 	// Ensure the allow-tigera tier exists, before rendering any network policies within it.
 	includeV3NetworkPolicy := false
-	if err := r.Client.Get(ctx, client.ObjectKey{Name: networkpolicy.TigeraComponentTierName}, &v3.Tier{}); err != nil {
+	if err := r.Client.Get(ctx, client.ObjectKey{Name: networkpolicy.TigeraComponentTierName}, &v3.Tier{}); err != nil && isCalicoEnterprise {
```
caseydavenport, line 305: We actually need to put this within an if branch that prevents the `Get` call from being made at all. Otherwise, we will be sending unnecessary requests to the Kubernetes API server.

Normally, we block on the `tierWatchReady` - this ensures that requests made with the client are served from our local cache and don't actually generate API requests. But without that, this is going to generate a lot of requests that otherwise would not have happened.

pkg/controller/clusterconnection/clusterconnection_controller.go
```diff
@@ -349,7 +352,7 @@ func (r *ReconcileConnection) Reconcile(ctx context.Context, request reconcile.R
 	// In managed clusters, the clusterconnection controller is a dependency for the License to be created. In case the
 	// License is unavailable and reconciliation of non-NetworkPolicy resources in the clusterconnection controller
 	// would resolve it, we render network policies last to prevent a chicken-and-egg scenario.
-	if includeV3NetworkPolicy {
+	if includeV3NetworkPolicy && isCalicoEnterprise {
```
caseydavenport, line 355: we shouldn't need this extra check here - `includeV3NetworkPolicy` will be false if `isCalicoEnterprise` is false.

pkg/render/guardian.go
```diff
@@ -81,13 +81,14 @@ func GuardianPolicy(cfg *GuardianConfiguration) (Component, error) {
...
-	TrustedCertBundle certificatemanagement.TrustedBundle
-	TunnelCAType      operatorv1.CAType
+	URL                 string
+	PullSecrets         []*corev1.Secret
+	Openshift           bool
+	Installation        *operatorv1.InstallationSpec
+	TunnelSecret        *corev1.Secret
+	TrustedCertBundle   certificatemanagement.TrustedBundle
+	TunnelCAType        operatorv1.CAType
+	EnterpriseCRDExists bool
```
caseydavenport, line 91: Instead of using the presence of the enterprise CRD, we should use `c.cfg.Installation.Variant`, which is already available in the config.

pkg/render/guardian.go
```diff
@@ -125,17 +130,22 @@ func (c *GuardianComponent) Objects() ([]client.Object, []client.Object) {
...
-		managerClusterWideSettingsGroup(),
-		managerUserSpecificSettingsGroup(),
-		managerClusterWideTigeraLayer(),
-		managerClusterWideDefaultView(),
 	)
 
+	if c.cfg.EnterpriseCRDExists {
+		// Add tigera-manager service account for impersonation
+		objs = append(objs,
+			CreateNamespace(ManagerNamespace, c.cfg.Installation.KubernetesProvider, PSSRestricted),
```
caseydavenport, line 138: We may need the namespace / serviceaccount / and RBAC resources.

Managed clusters need to be able to authorize requests from management cluster components. So if `system:serviceaccount:tigera-manager:tigera-manager` is making requests to the managed cluster API server, we will need to make sure that service account exists and is bound to the correct permissions.

pkg/controller/clusterconnection/clusterconnection_controller.go
```diff
@@ -191,6 +187,7 @@ func (r *ReconcileConnection) Reconcile(ctx context.Context, request reconcile.R
 	if err != nil {
 		return result, err
 	}
+	isCalicoEnterprise := variant == operatorv1.TigeraSecureEnterprise
```
caseydavenport, line 190: This is really nitty, but I think we should just use `variant == operatorv1.TigeraSecureEnterprise` throughout instead of aliasing it as a new variable. This is consistent with how we do this same check in other similar controllers.

pkg/render/guardian.go
```diff
@@ -125,17 +129,22 @@ func (c *GuardianComponent) Objects() ([]client.Object, []client.Object) {
...
 		managerClusterRole(false, true, c.cfg.UsePSP, c.cfg.Installation.KubernetesProvider),
 		managerClusterRoleBinding(),
-		managerClusterWideSettingsGroup(),
-		managerUserSpecificSettingsGroup(),
-		managerClusterWideTigeraLayer(),
-		managerClusterWideDefaultView(),
 	)
 
+	if c.cfg.Installation.Variant == operatorv1.TigeraSecureEnterprise {
+		// Add tigera-manager service account for impersonation
```
caseydavenport, line 139: This comment is in the wrong place

PR #2860, opened 2023-09-07
Fix! Tenant CA secrets needs to be rendered

asincu:
## Description

We need this in order to allow tenant controller to create the secrets that are not `tigera-ca-private` in the tenant namespace.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/certificatemanagement/certificatemanagement.go
+++ b/pkg/render/certificatemanagement/certificatemanagement.go
@@ -98,7 +98,7 @@ func (c component) Objects() (objsToCreate, objsToDelete []client.Object) {
 			}
 			needsCSRRoleAndBinding = true
 		} else {
-			if keyPairCreator.renderInTruthNamespace && (!keyPair.BYO() || keyPair.GetName() == certificatemanagement.CASecretName) {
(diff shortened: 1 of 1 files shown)


pkg/render/certificatemanagement/certificatemanagement.go
```diff
@@ -98,7 +98,7 @@ func (c component) Objects() (objsToCreate, objsToDelete []client.Object) {
 			}
 			needsCSRRoleAndBinding = true
 		} else {
-			if keyPairCreator.renderInTruthNamespace && (!keyPair.BYO() || keyPair.GetName() == certificatemanagement.CASecretName) {
+			if keyPairCreator.renderInTruthNamespace && (!keyPair.BYO() || keyPair.GetName() != certificatemanagement.CASecretName) {
```
rene-dekker, line 101: Doesn't this break non-multitenancy?

caseydavenport: Yeah, this isn't the branch that is intended to handle multi-tenancy (although I can see why it would seem that way based on the name check).

asincu: Talked offline. The CA generated per tenant does not have an issuer, but we still need to whitelist this in order to allow for multi-tenacy installation to happen. I have enhanced the condition.

caseydavenport (approved):

PR #2861, opened 2023-09-08
[EV-3914] Set the controller manager cache RESTMapper back to the DynamicRESTMapper (default changed on k8s upgrade)

Brian-McM:
The controller runtime adding this https://github.com/kubernetes-sigs/controller-runtime/commit/b38c4a526b4cf9ffbcaebd0fe9363d9d64182dd2 which changes how the cache builder and the controller cache options are combined.

Previously if the cache builders REST Mapper was nil, it would default to the controllers rest mapper, which defaults to the DynamicRESTMapper. This rest mapper is not a static mapper, it looks up types that were previously missing when consulted, like the Tier or Licence types.

With the commit mentioned, if the Mapper in the cache builder config is nil that code defaults it to the DiscoveryRESTMapper, which does not look up types that were previously missing. We use the cache builder here https://github.com/tigera/operator/blob/master/main.go#L232 to do some special things for tiers, so the upgrade from controller runtime 0.13 to 0.14 switched the mapper. This is why removing the custom cache fixes the issue, since it means the default is changed back to the DynamicRestMapper.

This should probably be fixed upstream, but for now we can just set the Mapper in the builder to the DynamicRESTMapper as I've done in this commit.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


main.go
```diff
@@ -201,6 +202,12 @@ func main() {
 		os.Exit(0)
 	}
 
+	mapper, err := apiutil.NewDynamicRESTMapper(cfg)
+	if err != nil {
+		log.Error(err, "Failed stuff.")
```
rene-dekker, line 207: Maybe this needs some polishing : )

rene-dekker (approved):
lgtm

main.go
Brian-McM: Haha! I disagree :smile:

tmjd (approved):
LGTM, the real test for me will be if downstream CI stays clean, but it certainly seems very reasonable that the RESTMapper change was the issue.

PR #2866, opened 2023-09-13
Fix certificate management due to missing volumemount

rene-dekker:


pkg/controller/logstorage/secrets/secret_controller.go
```diff
@@ -226,6 +227,12 @@ func (r *SecretSubController) Reconcile(ctx context.Context, request reconcile.R
 		return reconcile.Result{}, nil
 	}
 
+	r.managementCluster, err = utils.GetManagementCluster(ctx, r.client)
```
caseydavenport, line 230: I think it's a little bit fishy to store a per-reconcile object as a field on the reconciler to avoid passing it as an argument to a function. I'd rather we just pass this as an argument to `generateNamespacedSecrets`

Worried that storing it might lead to unexpected behaviors as this controller evolves.

caseydavenport (approved):
One comment but otherwise this LGTM - thanks for fixing this @rene-dekker !

pkg/controller/logstorage/kubecontrollers/esgateway.go
```diff
@@ -73,7 +73,10 @@ func (r *ESKubeControllersController) createESGateway(
 		r.status.SetDegraded(operatorv1.ResourceCreateError, "Unable to create the Tigera CA", err, reqLogger)
 		return err
 	}
-	gatewayDNSNames := dns.GetServiceDNSNames(render.ElasticsearchServiceName, helper.InstallNamespace(), r.clusterDomain)
+	gatewayDNSNames := append(
+		dns.GetServiceDNSNames(render.ElasticsearchServiceName, helper.InstallNamespace(), r.clusterDomain),
```
caseydavenport, line 77: Could you explain why we need both of these sets of DNS names?

caseydavenport (approved):

PR #2883, opened 2023-09-19
Add RBAC permissions so Linseed can operate on secrets and configmaps

Josh-Tigera:
where necessary

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/logstorage.go
+++ b/pkg/render/logstorage.go
@@ -70,6 +70,10 @@ const (
 	// The secret contains server key and certificate.
 	TigeraLinseedSecret = "tigera-secure-linseed-cert"
 
+	// TigeraLinseedSecretsClusterRole is the name of the ClusterRole used to make RoleBindings in namespaces where Linseed
+	// needs to be able to manipulate secrets
+	TigeraLinseedSecretsClusterRole = "tigera-linseed-secrets"
+
 	// TigeraLinseedTokenSecret is the name of the secret that holds the access token signing key for Linseed.
 	TigeraLinseedTokenSecret = "tigera-secure-linseed-token-tls"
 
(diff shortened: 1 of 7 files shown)


pkg/render/logstorage.go
```diff
@@ -2122,17 +2132,52 @@ func (m managedClusterLogStorage) linseedExternalRoleAndBinding() (*rbacv1.Clust
 		},
 	}
 
-	// Bind the permission to the tigera-fluentd namespace. Other controllers may also bind
+	// These permissions are necessary so that we can fetch the operator namespace of the managed cluster from the
+	// management cluster so that we're copying secrets into the right place in a multi-tenant environment.
+	configMapsRole := &rbacv1.ClusterRole{
```
caseydavenport, line 2137: Just to be clear - this configmap role is only needed in managed clusters, right? It isn't needed in a management cluster?

pkg/render/logstorage.go
```diff
@@ -2122,17 +2132,52 @@ func (m managedClusterLogStorage) linseedExternalRoleAndBinding() (*rbacv1.Clust
...
+			{
+				Kind:      "ServiceAccount",
+				Name:      "tigera-linseed",
+				Namespace: ElasticsearchNamespace,
+			},
+		},
+	}
+
+	// Bind the configmaps permission to the calico-system namespace.
+	configMapBinding := &rbacv1.RoleBinding{
```
caseydavenport, line 2172: This as well is only needed in managed clusters?

pkg/render/logstorage.go
Josh-Tigera: That's right. We use the managed cluster's k8s client to call `FetchOperatorNamespace` which is what necessitates this permission.

pkg/render/logstorage.go
Josh-Tigera: This specific one yes, we do have an equivalent for management clusters by way of linseed.go:188

pkg/render/logstorage/linseed/linseed.go
```diff
@@ -185,6 +185,13 @@ func (l *linseed) linseedClusterRole() *rbacv1.ClusterRole {
 			Resources: []string{"managedclusters"},
 			Verbs:     []string{"list", "watch"},
 		},
+		// These permissions are necessary to allow the management cluster to monitor secrets that we want to propagate
+		// through to the managed cluster for identity verification such as the Voltron Linseed public certificate
+		{
+			APIGroups: []string{""},
+			Resources: []string{"secrets"},
+			Verbs:     []string{"get", "list", "watch"},
```
caseydavenport, line 193: I think we can include a `ResourceNames` section here to limit specifically to the secret we want to copy, no? It should always be a well-known secret name + namespace.

caseydavenport (approved):

pkg/render/logstorage/linseed/linseed.go
caseydavenport: I think it's OK to do that in parallel and we can merge this to at least get things working. 

It should definitely be possible to have the informer only watch a particular resource, it might require multiple in order to watch multiple resources though.

marvin-tigera:

marvin-tigera:

PR #2884, opened 2023-09-20
Add RBAC for DPI when a management cluster is spin up

asincu:
## Description

In the following setup:
-  managed cluster has DPI configured
- management cluster does not have any DPI configured

All calls to ingest DPI alerts will fails with 401 User not authorized to create events. This is happening because DPI is installed only conditionally. Thus, its RBAC to create events via Linseed is only setup if DPI is already installed in the management cluster.

```
2023-09-19 23:32:06.599 [INFO][1] token.go 226: User not authorized cluster="saba-bz-uoc7" group="linseed.tigera.io" resource="events" user="system:serviceaccount:tigera-dpi:tigera-dpi" verb="create"
2023-09-19 23:32:07.087 [INFO][1] token.go 226: User not authorized cluster="saba-bz-uoc7" group="linseed.tigera.io" resource="events" user="system:serviceaccount:tigera-dpi:tigera-dpi" verb="create"
2023-09-19 23:32:08.913 [INFO][1] token.go 226: User not authorized cluster="saba-bz-uoc7" group="linseed.tigera.io" resource="events" user="system:serviceaccount:tigera-dpi:tigera-dpi" verb="create"
2023-09-20 01:22:41.634 [INFO][1] token.go 226: User not authorized cluster="saba-bz-uoc7" group="linseed.tigera.io" resource="events" user="system:serviceaccount:tigera-dpi:tigera-dpi" verb="create"

```

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/intrusiondetection/dpi/dpi.go
+++ b/pkg/render/intrusiondetection/dpi/dpi.go
@@ -38,13 +38,14 @@ import (
 )
 
 const (
-	DeepPacketInspectionNamespace  = "tigera-dpi"
-	DeepPacketInspectionName       = "tigera-dpi"
-	DeepPacketInspectionPolicyName = networkpolicy.TigeraComponentPolicyPrefix + DeepPacketInspectionName
-	DefaultMemoryLimit             = "1Gi"
-	DefaultMemoryRequest           = "100Mi"
-	DefaultCPULimit                = "1"
-	DefaultCPURequest              = "100m"
+	DeepPacketInspectionNamespace       = "tigera-dpi"
+	DeepPacketInspectionName            = "tigera-dpi"
+	DeepPacketInspectionPolicyName      = networkpolicy.TigeraComponentPolicyPrefix + DeepPacketInspectionName
+	DefaultMemoryLimit                  = "1Gi"
+	DefaultMemoryRequest                = "100Mi"
+	DefaultCPULimit                     = "1"
+	DefaultCPURequest                   = "100m"
+	DeepPacketInspectionLinseedRBACName = "tigera-dpi-linseed-permissions"
 )
 
 type DPIConfig struct {
@@ -54,6 +55,7 @@ type DPIConfig struct {
 	PullSecrets        []*corev1.Secret
 	Openshift          bool
 	ManagedCluster     bool
+	ManagementCluster  bool
 	HasNoLicense       bool
 	HasNoDPIResource   bool
 	ESClusterConfig    *relasticsearch.ClusterConfig
@@ -113,6 +115,25 @@ func (d *dpiComponent) Objects() (objsToCreate, objsToDelete []client.Object) {
 			d.dpiClusterRole(),
 			d.dpiClusterRoleBinding(),
 			d.dpiDaemonset(),
+		)
+	}
+	if d.cfg.ManagementCluster {
+		// We always want to create these permissions when a management
+		// cluster is configured, to allow any DPI running inside a
+		// managed cluster to write data
+		toCreate = append(toCreate, d.dpiLinseedAccessClusterRole())
+		toCreate = append(toCreate, d.dpiLinseedAccessClusterRoleBinding())
+	} else if !d.cfg.ManagedCluster && !d.cfg.HasNoDPIResource && !d.cfg.HasNoLicense {
+		// We want to create these permissions when a standalone
+		// cluster is configured to run DPI
+		toCreate = append(toCreate, d.dpiLinseedAccessClusterRole())
+		toCreate = append(toCreate, d.dpiLinseedAccessClusterRoleBinding())
+	} else {
+		// We want to remove these permissions when a standalone
+		// cluster is no longer configured, to run DPI or for managed clusters
+		toDelete = append(toDelete,
+			d.dpiLinseedAccessClusterRole(),
+			d.dpiLinseedAccessClusterRoleBinding(),
 		)
 	}
 	if d.cfg.ManagedCluster {
@@ -321,6 +342,28 @@ func (d *dpiComponent) dpiClusterRoleBinding() *rbacv1.ClusterRoleBinding {
 	}
 }
 
+func (d *dpiComponent) dpiLinseedAccessClusterRoleBinding() *rbacv1.ClusterRoleBinding {
+	return &rbacv1.ClusterRoleBinding{
+		TypeMeta: metav1.TypeMeta{Kind: "ClusterRoleBinding", APIVersion: "rbac.authorization.k8s.io/v1"},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:   DeepPacketInspectionLinseedRBACName,
+			Labels: map[string]string{},
+		},
+		RoleRef: rbacv1.RoleRef{
+			APIGroup: "rbac.authorization.k8s.io",
+			Kind:     "ClusterRole",
+			Name:     DeepPacketInspectionLinseedRBACName,
+		},
+		Subjects: []rbacv1.Subject{
+			{
+				Kind:      "ServiceAccount",
+				Name:      DeepPacketInspectionName,
+				Namespace: DeepPacketInspectionNamespace,
+			},
+		},
+	}
+}
+
 func (d *dpiComponent) dpiClusterRole() *rbacv1.ClusterRole {
 	role := &rbacv1.ClusterRole{
 		TypeMeta: metav1.TypeMeta{Kind: "ClusterRole", APIVersion: "rbac.authorization.k8s.io/v1"},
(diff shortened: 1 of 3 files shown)


pkg/render/intrusiondetection/dpi/dpi.go
```diff
@@ -321,6 +344,28 @@ func (d *dpiComponent) dpiClusterRoleBinding() *rbacv1.ClusterRoleBinding {
 	}
 }
 
+func (d *dpiComponent) dpiLinseedAccessClusterRoleBinding() *rbacv1.ClusterRoleBinding {
+	return &rbacv1.ClusterRoleBinding{
+		TypeMeta: metav1.TypeMeta{Kind: "ClusterRoleBinding", APIVersion: "rbac.authorization.k8s.io/v1"},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:   fmt.Sprintf("%s-extra", DeepPacketInspectionNamespace),
```
caseydavenport, line 351: I don't thinks this name is right - it doesn't explain what this CRB does. For example, `tigera-dpi-linseed-permissions`

I'd also like to see a comment on this function explaining when this CRB is expected to exist and why. I think the answer to that is "This binding should exist in management and standalone clusters - essentially any cluster running Linseed.  We need this for authz checks performed by Linseed when receiving requests from DPI"

pkg/render/intrusiondetection/dpi/dpi.go
```diff
@@ -114,6 +124,19 @@ func (d *dpiComponent) Objects() (objsToCreate, objsToDelete []client.Object) {
 			d.dpiClusterRoleBinding(),
 			d.dpiDaemonset(),
 		)
+		// We also want these permissions present for a standalone
+		// cluster that has DPI resource configured
+		if !d.cfg.ManagementCluster && !d.cfg.ManagedCluster {
```
caseydavenport, line 129: Instead of duplicating this logic here and in the above branch, can't we just do this below?

```
if d.cfg.ManagementCluster || !d.cfg.ManagedCluster {
  // Always create permission for DPI to write to Linseed in management and standalone clusters.
} else {
  // Remoe the permissions otherwise. These only need to exist in clusters that are running Linseed.
}
```

pkg/render/intrusiondetection/dpi/dpi.go
```diff
@@ -370,6 +409,24 @@ func (d *dpiComponent) dpiClusterRole() *rbacv1.ClusterRole {
 	return role
 }
 
+func (d *dpiComponent) dpiLinseedAccessClusterRole() *rbacv1.ClusterRole {
+	return &rbacv1.ClusterRole{
+		TypeMeta: metav1.TypeMeta{Kind: "ClusterRole", APIVersion: "rbac.authorization.k8s.io/v1"},
+		ObjectMeta: metav1.ObjectMeta{
+			Name: fmt.Sprintf("%s-extra", DeepPacketInspectionName),
```
caseydavenport, line 416: Same here - this name should explain the purpose of the role. For example: 

```
tigera-dpi-linseed-permissions
```

pkg/render/intrusiondetection/dpi/dpi.go
asincu: This means that we will have these roles installed always for standalone regardless whether we have DPI CR configured or not. Also, when removing DPI, these roles will still be present. It does simplify the logic and if it is something we can live with, I have no problem with this.

asincu: Talked offine and decided on a if else if statement to cover all this.

pkg/render/intrusiondetection/dpi/dpi.go
```diff
@@ -115,6 +117,25 @@ func (d *dpiComponent) Objects() (objsToCreate, objsToDelete []client.Object) {
 			d.dpiDaemonset(),
 		)
 	}
+	// We always want to create these permissions when a management
+	// cluster is configured to allow any DPI running inside a
```
caseydavenport, line 121: nitty, but you're missing a comma here:

"when a management cluster is configured, to allow any DPI"

has a different meaning than:

"when a management cluster is configured to allow any API"

caseydavenport: Also nitty, but this comment should be within the first branch of the if / elseif / else

caseydavenport: Logic looks good though :+1:

pkg/render/intrusiondetection/dpi/dpi_test.go
```diff
@@ -87,6 +87,14 @@ var (
 		},
 	}
 
+	expectedExtraClusterRoleRules = []rbacv1.PolicyRule{
```
caseydavenport, line 90: "expectedLinseedClusterRoleRules"

pkg/render/intrusiondetection/dpi/dpi_test.go
```diff
@@ -102,6 +110,21 @@ var (
 		},
 	}
 
+	expectedExtraCRB = rbacv1.RoleBinding{
```
caseydavenport, line 113: "expectedLinseedCRB"

pkg/render/intrusiondetection/dpi/dpi_test.go
```diff
@@ -243,6 +268,47 @@ var _ = Describe("DPI rendering tests", func() {
...
+			HasNoLicense:       false,
+			HasNoDPIResource:   true,
+			ManagementCluster:  true,
+			ESClusterConfig:    esConfigMap,
+			ClusterDomain:      dns.DefaultClusterDomain,
+			DPICertSecret:      dpiCertSecret,
+		}
+		component := dpi.DPI(cfg)
+
+		resources, _ := component.Objects()
```
caseydavenport, line 287: Do we have a test that confirms the `objectsToDelete` is populated correctly for managed clusters and standalone clusters without DPI?

pkg/render/intrusiondetection/dpi/dpi_test.go
```diff
@@ -243,6 +268,47 @@ var _ = Describe("DPI rendering tests", func() {
 		validateDPIComponents(resources, false)
 	})
 
+	It("should render all resources for deep packet inspection with default resource requirements for a management cluster without DPI resources", func() {
```
caseydavenport, line 271: this test name is a bit much - does the test care at all about resource requirements? 

let's focus it in on just the thing we want to verify here:

"Should render Linseed permissions for management clusters, even without DPI enabled"

pkg/render/intrusiondetection/dpi/dpi_test.go
asincu: For standalone, we already have tests for it. 
https://github.com/tigera/operator/blob/master/pkg/render/intrusiondetection/dpi/dpi_test.go#L294
https://github.com/tigera/operator/blob/master/pkg/render/intrusiondetection/dpi/dpi_test.go#L319

I can add one for managed also.

pkg/render/intrusiondetection/dpi/dpi_test.go
```diff
@@ -243,6 +268,47 @@ var _ = Describe("DPI rendering tests", func() {
...
+		expectedResources := []resourceTestObj{
+			{name: dpi.DeepPacketInspectionNamespace, ns: "", group: "", version: "v1", kind: "Namespace"},
+			{name: dpi.DeepPacketInspectionLinseedRBACName, ns: "", group: "rbac.authorization.k8s.io", version: "v1", kind: "ClusterRole"},
+			{name: dpi.DeepPacketInspectionLinseedRBACName, ns: "", group: "rbac.authorization.k8s.io", version: "v1", kind: "ClusterRoleBinding"},
+		}
+
+		Expect(len(resources)).To(Equal(len(expectedResources)))
+
+		for i, expectedRes := range expectedResources {
+			rtest.CompareResource(resources[i], expectedRes.name, expectedRes.ns, expectedRes.group, expectedRes.version, expectedRes.kind)
```
caseydavenport, line 298: This is a better function to use in tests: https://github.com/tigera/operator/blob/0c1c706b30dc8ab66247bdc911d005a71b4283ee/pkg/render/common/test/testing.go#L107-L138

I know this is just copied from the other tests in this package, but for new tests we should be using this function IMO - it produces much better output and is more flexible to changes in ordering, etc.

caseydavenport: If you want 100% coverage, this function is even better: https://github.com/tigera/operator/blob/0c1c706b30dc8ab66247bdc911d005a71b4283ee/pkg/render/common/test/testing.go#L95

pkg/render/intrusiondetection/dpi/dpi_test.go
asincu: I think for a managed cluster we never create the cluster role and cluster rolebinding, but we include them for deletion.

caseydavenport (approved):

PR #2887, opened 2023-09-20
Add IPAMConfig StrictAffinity validation to windows controller

coutinhop:
Also update tests accordingly and clean up duplicated import in windows_controller.go.

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/installation/windows_controller.go
+++ b/pkg/controller/installation/windows_controller.go
@@ -321,8 +321,8 @@ func (r *ReconcileWindows) Reconcile(ctx context.Context, request reconcile.Requ
 			return reconcile.Result{}, err
 		}
 		if !ipamConfiguration.Spec.StrictAffinity {
-			err := fmt.Errorf("StrictAffinity is false, it must be set to 'true' in the default IPAMConfiguration when using Calico CNI on Windows")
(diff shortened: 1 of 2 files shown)


pkg/controller/installation/windows_controller.go
```diff
@@ -151,6 +151,12 @@ func AddWindowsController(mgr manager.Manager, opts options.AddOptions) error {
 		return fmt.Errorf("tigera-windows-controller failed to watch FelixConfiguration resource: %w", err)
 	}
 
+	// Watch for changes to IPAMConfiguration.
+	err = c.Watch(&source.Kind{Type: &apiv3.IPAMConfiguration{}}, &handler.EnqueueRequestForObject{})
```
caseydavenport, line 155: Any time you're watching for a projectcalico.org/v3 resource, you need to use a special pattern because the watch won't establish until the apiserver is running.

Here is an example of starting a watch: https://github.com/tigera/operator/blob/0c1c706b30dc8ab66247bdc911d005a71b4283ee/pkg/controller/manager/manager_controller.go#L99-L104

And then your controller needs to wait for the Watch to establish before performing a Get() on that resource, like this: https://github.com/tigera/operator/blob/master/pkg/controller/manager/manager_controller.go#L296-L300

coutinhop: Got it, thanks @caseydavenport! Done

caseydavenport (approved):

PR #2894, opened 2023-09-25
Ability to configure per-tenant index names

asincu:
## Description

Define data retention for each tenant. For all stored data, we will defined the ILM policy and index name to be used. Linseed will be configured with these values via environment variables.

ILM policies, index templates, write aliases and boostrap indices will be created out of band when running in multi-tenant mode.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


pkg/render/logstorage/linseed/linseed.go
```diff
@@ -310,6 +310,11 @@ func (l *linseed) linseedDeployment() *appsv1.Deployment {
 		envVars = append(envVars, corev1.EnvVar{Name: "LINSEED_EXPECTED_TENANT_ID", Value: l.cfg.Tenant.Spec.ID})
 		envVars = append(envVars, corev1.EnvVar{Name: "BACKEND", Value: "elastic-single-index"})
 		envVars = append(envVars, corev1.EnvVar{Name: "LINSEED_MULTI_CLUSTER_FORWARDING_ENDPOINT", Value: fmt.Sprintf("https://tigera-manager.%s.svc:9443", l.cfg.Tenant.Namespace)})
+		envVars = append(envVars, corev1.EnvVar{Name: "ELASTIC_DYNAMIC_INDEX_CREATION_ENABLED", Value: "true"})
```
caseydavenport, line 313: Why do we only set this if a tenant is provided?

pkg/render/logstorage/linseed/linseed.go
```diff
@@ -310,6 +310,11 @@ func (l *linseed) linseedDeployment() *appsv1.Deployment {
...
+		for _, data := range l.cfg.Tenant.Spec.DataRetention {
+			envVars = append(envVars, corev1.EnvVar{Name: fmt.Sprintf("ELASTIC_%s_INDEX_NAME", strings.ToUpper(data.Type)), Value: data.IndexName})
```
caseydavenport, line 315: Generally it's a not-so-great pattern to link up the user-facing API (the Tenant CR) with the implementation detail (the env var). 

Even if they can be programatically generated, it's better to not do that in case we come upon a case where the strings.ToUpper isn't a good enough translation.

Tying API -> implementation detail is almost always a bad thing.

One option would be to do something like this:

```go
type DataType string

const (
  DataTypeFlowLog DataType = "FlowLogs"
)

func (t DataType) EnvVars() []corev1.EnvVar {
  return []corev1.EnvVar{
    {Name: t.IndexEnvName(), Value: data.IndexName},
  }
}

func (t DataType) IndexEnvName() string {
  switch t {
    case DataTypeFlowLog:
      return "ELASTIC_FLOW_LOG_INDEX_NAME"
    default:
      panic("Unexpected data type")
  }
}
```

WDYT?

api/v1/tenant_types.go
```diff
@@ -24,6 +24,25 @@ type TenantSpec struct {
 	// ID is the unique identifier for this tenant.
 	// +required
 	ID string `json:"id,omitempty"`
+
+	// DataRetention defines the how to store a tenant's data
+	DataRetention []DataRetention `json:"dataRetention"`
```
caseydavenport, line 29: I think `DataRetrention` is no longer the right name for this - it really is just configuring indices. Maybe just call the field `Indices`? 

I think we should set validation on this field that ensures:
- Each element is unique, and there are no duplicates in the array.
- If any indices are specified, then _all_ indices should be specified. I don't think it's valid to just include one index here, and it will be easy to accidentally forget one.

api/v1/tenant_types.go
```diff
@@ -24,6 +24,25 @@ type TenantSpec struct {
...
+	// store data in Elastic
+	PolicyName string `json:"policyName"`
+
+	// IndexName defines the name of the index
+	// that will be used to store data (this name
+	// excludes the numerical identifier suffix)
+	IndexName string `json:"indexName"`
+
+	// +kubebuilder:validation:Enum=alerts;audit;bgp;compliance_benchmarks;compliance_reports;compliance_snapshots;dns;flow;l7;runtime;threat_feeds_domain_name;thread_feeds_ip_set;waf
+	Type string `json:"type"`
```
caseydavenport, line 45: Maybe call this "DataType" to match up how we refer to it elsewhere?

api/v1/tenant_types.go
caseydavenport: A good test that shows one way in which the naming isn't quite right - `DataRetention.Type` implies "What type of data retention?" but actually the field means "What type of data does this refer to"

pkg/render/logstorage/linseed/linseed.go
asincu: We should not set this.

api/v1/tenant_types.go
caseydavenport: I think we can do the uniqueness check via kubebuilder (but not 100% sure).

Otherwise, it should probably be the Linseed controller for now, given that's where this field is used.

asincu: I did try the kubebuilder tag, but then I saw FV failures: (https://tigera.semaphoreci.com/jobs/26a53ce8-5407-481d-aed2-8150f322bfd9)

```
customresourcedefinition.apiextensions.k8s.io/tigerastatuses.operator.tigera.io created 01:51
The CustomResourceDefinition "tenants.operator.tigera.io" is invalid: spec.validation.openAPIV3Schema.properties[spec].properties[indices].uniqueItems: Forbidden: uniqueItems cannot be set to true since the runtime complexity becomes quadratic 01:51
Makefile:403: recipe for target 'deploy-crds' failed
```

api/v1/tenant_types.go
```diff
@@ -17,13 +17,67 @@ limitations under the License.
...
+// Index defines how to store a tenant's data
+type Index struct {
+	// IndexName defines the name of the index
+	// that will be used to store data (this name
+	// excludes the numerical identifier suffix)
+	IndexName string `json:"indexName"`
+
+	// DataType represents the type of data stored
+	// in the defined index
+	// +kubebuilder:validation:Enum=Alerts;AuditLogs;BGPLogs;ComplianceBenchmarks;ComplianceReports;ComplianceSnapshots;DNSLogs;FlowLogs;L7Logs;RuntimeReports;ThreatFeedsDomainSet;ThreadFeedsIPSet;WAFLogs
```
caseydavenport, line 79: If you put this on the `type DataType string` line instead of here, anywhere that inherits this enum will get the same validation. Not really important, but a neat trick of kubebuilder.

api/v1/tenant_types.go
```diff
@@ -17,13 +17,67 @@ limitations under the License.
...
+}
+
+// Index defines how to store a tenant's data
+type Index struct {
+	// IndexName defines the name of the index
+	// that will be used to store data (this name
+	// excludes the numerical identifier suffix)
+	IndexName string `json:"indexName"`
+
+	// DataType represents the type of data stored
```
caseydavenport, line 77: This can just be one line. It's not long enough to force a second line.

api/v1/tenant_types.go
```diff
@@ -52,3 +106,16 @@ type TenantList struct {
...
 }
+
+func (i *Index) EnvVar() corev1.EnvVar {
+	return corev1.EnvVar{Name: i.DataType.IndexEnvName(), Value: i.IndexName}
+}
+
+func (t DataType) IndexEnvName() string {
+	envName, ok := DataTypes[t]
+	if !ok {
+		panic("Unexpected data type")
```
caseydavenport, line 117: We should include the value in this panic

pkg/controller/logstorage/linseed/linseed_controller.go
```diff
@@ -205,6 +205,11 @@ func (r *LinseedSubController) Reconcile(ctx context.Context, request reconcile.
 		r.status.SetDegraded(operatorv1.ResourceReadError, "An error occurred while querying Tenant", err, reqLogger)
 		return reconcile.Result{}, err
 	}
+	err = validateTenant(tenant)
```
caseydavenport, line 208: I wonder if we should move this `validateTenant` inside the implementation of `GetTenant` - WDYT?

pkg/controller/logstorage/linseed/linseed_controller.go
```diff
@@ -426,3 +431,30 @@ func (r *LinseedSubController) Reconcile(ctx context.Context, request reconcile.
...
+func validateTenant(tenant *operatorv1.Tenant) error {
+	if tenant == nil {
+		return nil
+	}
+
+	declaredDataTypes := make(map[operatorv1.DataType]struct{})
+	for _, index := range tenant.Spec.Indices {
+		_, found := declaredDataTypes[index.DataType]
+		if found {
+			return fmt.Errorf("index %s was been declared multiple times", index.DataType)
```
caseydavenport, line 444: "was been" -> "is"

caseydavenport (commented):
Minor comments, otherwise LGTM

pkg/controller/logstorage/linseed/linseed_controller.go
asincu: Sure, but should we bunch up all errors under `operatorv1.ResourceReadError` ?

caseydavenport: I suppose we can leave it like so.

pkg/controller/logstorage/linseed/linseed_controller_test.go
```diff
@@ -269,6 +269,21 @@ var _ = Describe("LogStorage Linseed controller", func() {
 			tenant.Name = "default"
 			tenant.Namespace = tenantNS
 			tenant.Spec.ID = "test-tenant-id"
+			tenant.Spec.Indices = []operatorv1.Index{
```
caseydavenport, line 272: Is this needed? I would have expected we leave this test as is, but then add a new test which checks when non-standard names are given.

tmjd: I was thinking that we will have validation to check that indices are always set. And that it will be a validation failure if they are not. Since this is not expected to be used for Calico or Calico Enterprise users then I was thinking it would be reasonable to not have any defaults and require the fields be set, it will force us to provide the configuration rather than possibly using some defaults when it was not expected.

caseydavenport: Sounds good to me

caseydavenport (approved):

PR #2903, opened 2023-09-28
Remove ResourceNames on tigera-linseed ClusterRole

Josh-Tigera:
## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


caseydavenport (approved):
LGTM - would be nice to make this a `Role` instead of a `ClusterRole` at some point.

tmjd:
Do we need this change in general or is this only for the multi-tenant case? If it is only for multi-tenant then this change should be restricted only to that case then.

Josh-Tigera:
> Do we need this change in general or is this only for the multi-tenant case? If it is only for multi-tenant then this change should be restricted only to that case then.

@tmjd it's needed even in the single-tenant case since we moved the logic responsible for migrating VoltronLinseed certificate from management to managed cluster from es-kube-controllers to linseed. See [this](https://github.com/tigera/calico-private/pull/6553) PR for more info.

PR #2922, opened 2023-10-10
Fix windows_controller/core_controller race condition

coutinhop:
Make windows_controller wait for Installation.Spec.WindowsNodes to be defaulted by core_controller before continuing.

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/installation/windows_controller.go
+++ b/pkg/controller/installation/windows_controller.go
@@ -292,7 +292,12 @@ func (r *ReconcileWindows) Reconcile(ctx context.Context, request reconcile.Requ
 	// We rely on the core controller for defaulting, so wait until it has done so before continuing
 	if reflect.DeepEqual(instanceStatus, operatorv1.InstallationStatus{}) {
 		err := fmt.Errorf("InstallationStatus is empty")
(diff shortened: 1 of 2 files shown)


caseydavenport (approved):
LGTM - presumably this is for the upgrade case, because for new clusters I'd assume the check on line 293 right above would handle this?

caseydavenport:
@coutinhop do we need a UT for this?

coutinhop:
> LGTM - presumably this is for the upgrade case, because for new clusters I'd assume the check on line 293 right above would handle this?
@lwr20 hit this and sent me logs, so I'm not 100% sure what the scenario was, but couldn't there be a race condition where the installation status is not empty, but the core_controller hasn't filled in the defaults yet? (or is that just not possible?)

> @coutinhop do we need a UT for this?
absolutely, will add!

lwr20:
It was absolutely for a new cluster, but the banzai code is starting from linu

coutinhop:
> It was absolutely for a new cluster, but the banzai code is starting from Dataplane: linux and changing to Dataplane: WindowsHPC, which I could believe you're calling an upgrade case

yes that would definitely happen in this scenario... thanks @lwr20!

PR #2923, opened 2023-10-11
feat: add flag --cluster-type to disable auto discovery

### LINKED ISSUE #2920: AutoDiscoverProvider leads to wrong result
The code in pkg/controller/utils/discovery.go in the function `autodetectFromGroup` results in wrong results.

We run tigera-operater on EKS and the installation uses: `kubernetesProvider: EKS` but we observed the following error: 

```bash
Installation spec.kubernetesProvider 'EKS' does not match auto-detected value 'GKE'
```

We found out that the code 

```go
if g.Name == "networking.gke.io" {
	// Running on GKE.
	return operatorv1.ProviderGKE, nil
}
```

is responsible for our Problem. We are using `fqdnnetworkpolicies` in the group `networking.gke.io` and therefore `autodetectFromGroup` results in the wrong value.

## Expected Behavior

Tigera Operator should detect the correct cluster type

## Current Behavior

tigera-operator blives to be in GKE and complains about the `kubernetesProvider: EKS` setting in the installation

## Possible Solution

Allow to disable `AutoDiscoverProvider` by using a ENV variable or a argument for the operator

1. install fqdnnetworkpolicies.networking.gke.io on a EKS cluster
2. start tigera-operator

## Your Environment

* Operating System and version:

akesser:
## Description

This pull request adds a flag --cluster-type that allows users to set the used cluster type and disable auto discovery, that may lead to false results. This is a new feature to fix #2920 

Without this fix tigera cannot be used in a EKS cluster that uses fqdnnetworkpolicies.networking.gke.io, as the autodiscovery mechanism of tigera believes to be on a GKE cluster in this case. I tested this change locally without and with the flag using different valid and invalid values.

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/main.go
+++ b/main.go
@@ -94,6 +94,8 @@ func main() {
 	var printImages string
 	var printCalicoCRDs string
 	var printEnterpriseCRDs string
+	var userDefinedProviderInput string
+
 	var sgSetup bool
 	var manageCRDs bool
 	var preDelete bool
@@ -117,6 +119,8 @@ func main() {
 		"Operator should manage the projectcalico.org and operator.tigera.io CRDs.")
 	flag.BoolVar(&preDelete, "pre-delete", false,
 		"Run helm pre-deletion hook logic, then exit.")
+	flag.StringVar(&userDefinedProviderInput, "cluster-type", "",
+		"Use the given cluster type instead of using auto discovery. Valid values are EKS, GKE, AKS, RKE2, OpenShift, DockerEnterprise")
 
 	opts := zap.Options{}
 	opts.BindFlags(flag.CommandLine)
@@ -167,6 +171,28 @@ func main() {
 			setupLog.Error(err, "Terminating")
 			os.Exit(1)
 		}
+	}
+
+	userDefinedProvider := v1.ProviderNone
+	if userDefinedProviderInput != "" {
+		switch strings.ToLower(userDefinedProviderInput) {
+		case "eks":
+			userDefinedProvider = v1.ProviderEKS
+		case "gke":
+			userDefinedProvider = v1.ProviderGKE
+		case "aks":
+			userDefinedProvider = v1.ProviderAKS
+		case "rke2":
+			userDefinedProvider = v1.ProviderRKE2
+		case "openshift":
+			userDefinedProvider = v1.ProviderOpenShift
+		case "dockerenterprise":
+			userDefinedProvider = v1.ProviderDockerEE
+		default:
+			fmt.Println("Invalid option for --cluster-type flag ", userDefinedProviderInput, ". Valid values are EKS, GKE, AKS, RKE2, OpenShift, DockerEnterprise")
+			os.Exit(1)
+		}
+
 	}
 
 	printVersion()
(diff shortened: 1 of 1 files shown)


caseydavenport (commented):
This seems like a good escape hatch to have. @tmjd what do you think? We had talked about simply re-ordering the execution of the auto-discovery checks, but this is comparatively more robust (albeit, requiring more work on the part of the user).

Perhaps we want to do this, and also move the EKS check above the GKE check so that you won't have to use this flag in your particular scenario anyway?

tmjd:
I think this is an acceptable solution. I'd have concern with changing the auto discovery ordering that others hit an issue because of reordering. I'd probably have that concern with any changes to the auto discovery code.

tmjd:
I think another option I would consider is to remove the use of the auto discovered provider except for the purpose of adding it to the Installation if there is none set and remove the mismatch error.

It would also require changing many of the controllers to use the provider in the Installation rather than the auto-detected one.

I think something like this would be more in line with the rest of the operator configuration. In general it audo-detects settings or uses a default and saves the auto-detected setting to the config if it was unset, otherwise it uses what is configured.

caseydavenport:
@tmjd what do you think - should we go ahead with this approach or take on the slightly larger task of properly allowing the Installation spec field to override the auto-detected value?

tmjd:
I think we should go with
> properly allowing the Installation spec field to override the auto-detected value

That is how the rest of the API works and I don't think there is any particular reason we should differ with the provider.

So the change being suggested is to remove the validation that the auto-detected provider matches the provider in the Installation resource. Also it will be necessary to adjust any place that used the auto-detected provider (which I think it is most if not all controllers) to now use the provider as specified in the Installation resource.

@caseydavenport in main.go we check that the provider is not OpenShift before checking to set UsePSP. I guess checking if PSPs should be used will need to move into all the controllers also. Do you know if there anything we should particularly be concerned about since it uses the clientset which we don't usually use?

@akesser is it clear what we're suggesting and would you be interested in making updates to implement those changes?

akesser:
Using the auto detection and allowing the `kubernetesProvider` of the installation to overwrite the value would force everybody to always set the  field in every installation object because you would never know, if a) you need this field at the moment and b) if you will need this in the future. 

In our case, we installed first the operator and created the installation, after that we in stalled fqdnnetworkpolicies.networking.gke.io. The operator was running without problems till the first restart of the pod. Why should one repeat to add a field when the person installing the provider could simply add a flag? You could simply merge this   pull request and later deprecate and remove the `kubernetesProvider` field.

tmjd:
> Using the auto detection and allowing the kubernetesProvider of the installation to overwrite the value would force everybody to always set the field in every installation object because you would never know, if a) you need this field at the moment and b) if you will need this in the future.

The operator defaults the kubernetesProvider field when it is auto detected, so everybody that has already ran the operator has the kubernetesProvider field set if a provider was auto detected, so no one would need to explicitly set the field.

We would keep the auto detection of the provider for the purpose of defaulting the kubernetesProvider field, but when the Installation is reconciled it would set the kubernetesProvider to the auto detected provider only if it was not already set.
But from then on the auto detected value would be ignored and the kubernetesProvider would be the authority for the provider.

> In our case, we installed first the operator and created the installation, after that we in stalled fqdnnetworkpolicies.networking.gke.io. The operator was running without problems till the first restart of the pod.

In this scenario, what Casey and I are suggesting is that the operator would have updated the Installation resource and set the kubernetesProvider to the auto detected value of EKS (this is how it functions currently). Then when the operator was restarted, even though you installed `fqdnnetworkpolicies.networking.gke.io` and auto detection would have determined GKE, the operator would ignore the auto detected GKE and use the kubernetesProvider already set in the Installation resource which was EKS. So the suggested solution would have ensured you never would have had an issue.

Lets consider if you had the `fqdnnetworkpolicies...` before deploying the operator case, then the operator would have incorrectly detected GKE, with your current suggestion you would need to update the kuberentesProvider field and modify the deployment of the operator. Then in the future when you update the operator you would need to make the same modification to the operator deployment. With what I'm proposing you would only need to update the kubernetesProvider to EKS, then the operator would see that and use that as the provider and any future updates to the operator deployment would need no modifications.

PR #2938, opened 2023-10-20
Add a label to the prometheus service, so it can be referenced in Ser…

rene-dekker:
Add a label to the prometheus service, so it can be referenced in ServiceMonitors. This allows users to scrape our prometheus instance in federated mode.

### DIFF
--- a/pkg/render/monitor/monitor.go
+++ b/pkg/render/monitor/monitor.go
@@ -684,6 +684,9 @@ func (mc *monitorComponent) prometheusServiceService() *corev1.Service {
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      PrometheusServiceServiceName,
 			Namespace: common.TigeraPrometheusNamespace,
+			Labels: map[string]string{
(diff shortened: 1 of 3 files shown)


pkg/render/monitor/monitor.go
```diff
@@ -684,6 +684,9 @@ func (mc *monitorComponent) prometheusServiceService() *corev1.Service {
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      PrometheusServiceServiceName,
 			Namespace: common.TigeraPrometheusNamespace,
+			Labels: map[string]string{
+				"k8s-app": TigeraPrometheusObjectName,
```
tmjd, line 688: Does this need to be `k8s-app`? Or is this just one we've chosen?
I'm hesitant to use this since it is a standard k8s label for pods, is it also a standard one for Services?

tmjd: Looks like we use `k8s-app` for other service labels already.

tmjd (approved):

PR #2939, opened 2023-10-20
Automated cherry pick of #2938: Add a label to the prometheus service, so it can be

rene-dekker:
Cherry pick of #2938 on release-v1.32.

#2938: Add a label to the prometheus service, so it can be


tmjd (approved):
This is only picking part of 2938 since 2938 includes a CRD update that should not be picked back.

Issue #2886, opened 2023-09-20
feature request: calico-apiserver PodDisruptionBudget


dpiddock:
No PodDisruptionBudget is created for the `calico-apiserver` deployment.

## Expected Behavior

PodDisruptionBudget created for `calico-apiserver`, similar to the `calico-typha` deployment.

## Current Behavior

No PDB

## Possible Solution

Generate a PodDisruptionBudget when creating the Deployment

## Steps to Reproduce (for bugs)

1. Install operator
2. `kubectl -n calico-apiserver get pdb`
3. no result

## Context

The `calico-apiserver` deployment is configured with 2 replicas. To avoid service interruption when replacing a lot of nodes at once (e.g. during version upgrade) the deployment should be configured with a PodDisruptionBudget. Similar to the `calico-typha` deployment.

## Your Environment

* Operating System and version: AWS EKS Kubernetes 1.26
* Link to your project (optional):

Thank you

MuhtasimTanmoy:
I would like to work on this issue.

MuhtasimTanmoy:
Requesting review for https://github.com/tigera/operator/pull/2940.

Would like to mention one issue for testing though. 

 When applying this command.
`kubectl create -f ./config/samples/operator_v1_apiserver.yaml`

It was failing to pull `calico/apiserver:master` or `calico/apiserver:latest` from docker. 
Though `calico/apiserver:master-amd64` worked in my case.

So I changed the version to `master-amd64` to make it work though it is not included in pull request. https://github.com/tigera/operator/blob/077f7483633898b43050b7cfcca9935eea34ebf0/pkg/components/calico.go#L92

PR #2940, opened 2023-10-20
[Feature Request] Add podDisruptionBudget in API server

MuhtasimTanmoy:
## Description

Closes https://github.com/tigera/operator/issues/2886

This PR adds `podDisruptionBudget` in the API Server and affects the `calico-apiserver` component. 

## Testing
For testing this feature :
1. Install operator
2. Run `kubectl create -f ./config/samples/operator_v1_apiserver.yaml`
3. Run `kubectl -n calico-system get pdb` 
4. Shows `podDisruptionBudget` for `calico-apiserver` component. 

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:
  - `kind/enhancement` if this is a new feature.

### DIFF
--- a/pkg/common/common.go
+++ b/pkg/common/common.go
@@ -17,6 +17,7 @@ package common
 const (
 	CalicoNamespace               = "calico-system"
 	TyphaDeploymentName           = "calico-typha"
+	APIServerDeploymentName       = "calico-apiserver"
 	NodeDaemonSetName             = "calico-node"
 	KubeControllersDeploymentName = "calico-kube-controllers"
 	WindowsDaemonSetName          = "calico-node-windows"
--- a/pkg/common/common.go
+++ b/pkg/common/common.go
@@ -17,7 +17,6 @@ package common
 const (
 	CalicoNamespace               = "calico-system"
 	TyphaDeploymentName           = "calico-typha"
-	APIServerDeploymentName       = "calico-apiserver"
 	NodeDaemonSetName             = "calico-node"
 	KubeControllersDeploymentName = "calico-kube-controllers"
 	WindowsDaemonSetName          = "calico-node-windows"
--- a/pkg/render/apiserver.go
+++ b/pkg/render/apiserver.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2019-2023 Tigera, Inc. All rights reserved.
+// Copyright (c) 2019-2024 Tigera, Inc. All rights reserved.
 
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -31,7 +31,6 @@ import (
 
 	v3 "github.com/tigera/api/pkg/apis/projectcalico/v3"
 	operatorv1 "github.com/tigera/operator/api/v1"
-	"github.com/tigera/operator/pkg/common"
 	"github.com/tigera/operator/pkg/components"
 	"github.com/tigera/operator/pkg/controller/k8sapi"
 	rcomp "github.com/tigera/operator/pkg/render/common/components"
@@ -326,7 +325,7 @@ func (c *apiServerComponent) apiServerPodDisruptionBudget() *policyv1.PodDisrupt
 		TypeMeta: metav1.TypeMeta{Kind: "PodDisruptionBudget", APIVersion: "policy/v1"},
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      name,
-			Namespace: common.CalicoNamespace,
+			Namespace: rmeta.APIServerNamespace(c.cfg.Installation.Variant),
 		},
 		Spec: policyv1.PodDisruptionBudgetSpec{
 			MaxUnavailable: &maxUnavailable,
--- a/pkg/render/apiserver.go
+++ b/pkg/render/apiserver.go
@@ -21,6 +21,7 @@ import (
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	netv1 "k8s.io/api/networking/v1"
+	policyv1 "k8s.io/api/policy/v1"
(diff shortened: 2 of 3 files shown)


pkg/common/common.go
```diff
@@ -17,6 +17,7 @@ package common
 const (
 	CalicoNamespace               = "calico-system"
 	TyphaDeploymentName           = "calico-typha"
+	APIServerDeploymentName       = "calico-apiserver"
```
caseydavenport, line 20: Let's not add this here.

Instead, I think we should take this logic: https://github.com/tigera/operator/blob/4eff4995da85134f91a4a7c3bebf072b1095f35e/pkg/render/apiserver.go#L954-L960

And make a new function in `pkg/render/apiserver.go` like so:

```go
func deploymentName() string { . . .}
```

pkg/render/apiserver.go
```diff
@@ -290,6 +294,25 @@ func (c *apiServerComponent) Ready() bool {
...
+		TypeMeta: metav1.TypeMeta{Kind: "PodDisruptionBudget", APIVersion: "policy/v1"},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      common.APIServerDeploymentName,
+			Namespace: common.CalicoNamespace,
+		},
+		Spec: policyv1.PodDisruptionBudgetSpec{
+			MaxUnavailable: &maxUnavailable,
+			Selector: &metav1.LabelSelector{
+				MatchLabels: map[string]string{
+					AppLabelName: APIServerK8sAppName,
```
caseydavenport, line 309: I think this needs to match the labels set on the deployment selector here: https://github.com/tigera/operator/blob/4eff4995da85134f91a4a7c3bebf072b1095f35e/pkg/render/apiserver.go#L997

Maybe we should commonize this as well into a new private function in this file - `deploymentSelector` - so that both places can use it and make sure they have the same values.

caseydavenport (changes requested):
@MuhtasimTanmoy thanks for the PR! I've left a couple of small comments on it.

pkg/common/common.go
MuhtasimTanmoy: Can we generalize the deployment naming like this?

```go
// Determine names based on the configured variant
// It takes two name as parameters, enterpriseName and ossName, and returns name and nameToDelete.
func (c *apiServerComponent) deploymentName(enterpriseName, ossName string) (string, string) {
    var name, nameToDelete string
    switch c.cfg.Installation.Variant {
    case operatorv1.TigeraSecureEnterprise:
        name = enterpriseName
        nameToDelete = ossName
    case operatorv1.Calico:
        name = ossName
        nameToDelete = enterpriseName
    }
    return name, nameToDelete
}

// Usage
// name, nameToDelete := c.deploymentName("tigera-apiserver", "calico-apiserver")
// name, nameToDelete := c.deploymentName("tigera-apiserver-delegate-auth", "calico-apiserver-delegate-auth")
```
This should reduce around 60 lines in `pkg/render/apiserver.go`.

The final output should look like this commit https://github.com/MuhtasimTanmoy/operator/commit/f31415c21914b695abd158641c51d2665353e37c made on a temporary branch for review. Looking for better intuitive naming btw.

caseydavenport: Seems like a nice cleanup to me!

The only thing I would say is that the function shouldn't be called `deploymentName` if we're going to make it generalized, since not all of those objects are Deployments. 

Maybe `r.getResourceName`?

MuhtasimTanmoy: Yes, the function definitely should not be `deploymentName`. 

As the sole responsibility of the function is resolving name based on the configured variant regardless of it's usage, how about `c.resolveNameByVariant` or `c.resourceNameBasedOnVariant`?

caseydavenport: Sounds perfect :+1:

pkg/render/apiserver.go
MuhtasimTanmoy: Please review changes @caseydavenport

caseydavenport (approved):

caseydavenport:
@MuhtasimTanmoy looks like `make ut` is failing because it's not expecting this new resource to exist.

Here's an example failure:

```
pkg/render Suite
Failure
/go/src/github.com/tigera/operator/pkg/render/apiserver_test.go:1653
Expected
    <int>: 16
to equal
    <int>: 15
/go/src/github.com/tigera/operator/pkg/render/apiserver_test.go:1568
```

You should be able to run these locally with `make ut`

pkg/render/apiserver.go
```diff
@@ -290,6 +293,47 @@ func (c *apiServerComponent) Ready() bool {
...
+}
+
+func (c *apiServerComponent) apiServerPodDisruptionBudget() *policyv1.PodDisruptionBudget {
+	maxUnavailable := intstr.FromInt(1)
+	name, _ := c.resourceNameBasedOnVariant("tigera-apiserver", "calico-apiserver")
+	return &policyv1.PodDisruptionBudget{
+		TypeMeta: metav1.TypeMeta{Kind: "PodDisruptionBudget", APIVersion: "policy/v1"},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: rmeta.APIServerNamespace(c.cfg.Installation.Variant),
```
tmjd, line 328: This use to be `Namespace: common.CalicoNamespace`, but I think it needs to depend on the variant.

tmjd (commented):
I've pushed a commit to this branch, I found one bit I think needed changing (I've put a comment there) and I've fixed up the unit tests, they passed for me locally.

pkg/render/apiserver.go
MuhtasimTanmoy: Thanks. Had issues with accessing the CI error log or generating locally.

caseydavenport (approved):

PR #2945, opened 2023-10-24
Changes to support ebpf and ipv6

sridhartigera:
## Description

Remove the check to disallow eBPF and IPv6.

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/installation/validation.go
+++ b/pkg/controller/installation/validation.go
@@ -152,10 +152,18 @@ func validateCustomResource(instance *operatorv1.Installation) error {
 			}
 		}
 
+		if bpfDataplane && v4pool != nil && v6pool != nil {
+			return fmt.Errorf("bpf dataplane does not support dual stack")
+		}
+
 		if v4pool != nil {
 			_, cidr, err := net.ParseCIDR(v4pool.CIDR)
 			if err != nil {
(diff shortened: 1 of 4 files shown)


tomastigera (commented):
LGTM but I will let operator-maintainers to approve :)

pkg/controller/installation/validation.go
```diff
@@ -291,8 +292,16 @@ func validateCustomResource(instance *operatorv1.Installation) error {
 			}
 		}
 
-		if bpfDataplane && instance.Spec.CalicoNetwork.NodeAddressAutodetectionV4 == nil {
-			return fmt.Errorf("spec.calicoNetwork.nodeAddressAutodetectionV4 is required for the BPF dataplane")
+		if bpfDataplane {
+			if bpfIPv6Enabled {
```
tmjd, line 296: Is dual stack supported with bpf? If not then I think we should have a check for that.
I think I'd also suggest moving these validations to under the `if v4pool != nil {` and `if v6pool != nil {` blocks.

sridhartigera: Dual stack is not supported. Added a check for that.

tmjd (approved):
LGTM

PR #2947, opened 2023-10-24
Management cluster installation fixes

pasanw:
## Description
Fixes two new issues when installing a management cluster.

**1. Chicken-and-egg between APIServer and Manager.**
APIServer controller will not become ready until Manager controller defaults ManagementCluster CR and creates tunnel secret. Manager controller will not perform these actions unless the APIServer controller is ready.

This was a side-effect of the changes to the API server and operator that were involved in making the API server read the tunnel secret dynamically.

**2. Role in tigera-system created before tigera-system itself is created.**
This was a side-effect of establishing a Role specific to single tenant scenarios (previously was always a ClusterRole).

These side-effects arose from [here](https://github.com/tigera/operator/commit/2d6c9babadf43b5c8e458f67e3193c5ad5003249).

Fixes validated in a cluster that was failing with the current operator.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/apiserver/apiserver_controller.go
+++ b/pkg/controller/apiserver/apiserver_controller.go
@@ -144,8 +144,10 @@ func add(c controller.Controller, r *ReconcileAPIServer) error {
 		}
 
 		for _, namespace := range []string{common.OperatorNamespace(), rmeta.APIServerNamespace(operatorv1.TigeraSecureEnterprise)} {
-			if err = utils.AddSecretsWatch(c, render.VoltronTunnelSecretName, namespace); err != nil {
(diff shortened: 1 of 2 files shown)


rene-dekker (approved):
lgtm

asincu (approved):

pasanw:
@tmjd I think a better way for me to phrase that would have been: the APIServer controller will panic in it's current state unless the Manager controller has defaulted the ManagementCluster CR, and it will degrade unless the Manager controller has created the tunnel secret.

Putting a guard on the potential nil dereference, and not degrading if the tunnel secret is missing avoids the panic and breaks the circular dependency respectively

PR #2954, opened 2023-10-27
Remove usage of ES ConfigMap and PublicCertSecret

Josh-Tigera:
These resources are no longer used now that Linseed has taken over this functionality

This has a companion commit in Calico Enterprise [here](https://github.com/tigera/calico-private/pull/6821)
## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


pkg/render/fluentd.go
```diff
@@ -1040,10 +1042,40 @@ func (c *fluentdComponent) eksLogForwarderDeployment() *appsv1.Deployment {
 		{Name: "AWS_REGION", Value: c.cfg.EKSConfig.AwsRegion},
 		{Name: "AWS_ACCESS_KEY_ID", ValueFrom: secret.GetEnvVarSource(EksLogForwarderSecret, EksLogForwarderAwsId, false)},
 		{Name: "AWS_SECRET_ACCESS_KEY", ValueFrom: secret.GetEnvVarSource(EksLogForwarderSecret, EksLogForwarderAwsKey, false)},
+		relasticsearch.ElasticIndexSuffixEnvVar(c.cfg.ESClusterConfig.ClusterName()),
```
caseydavenport, line 1045: I think all of these are only needed in the env when EKS log forwarding is enabled, so we should only add them conditionally on that feature being enabled.

pkg/render/fluentd.go
```diff
@@ -1040,10 +1042,40 @@ func (c *fluentdComponent) eksLogForwarderDeployment() *appsv1.Deployment {
...
+		Name:            eksLogForwarderName + "-startup",
+		Image:           c.image,
+		ImagePullPolicy: ImagePullPolicy(),
+		Command:         []string{c.path("/bin/eks-log-forwarder-startup")},
+		Env:             envVars,
+		SecurityContext: c.securityContext(false),
+		VolumeMounts:    c.eksLogForwarderVolumeMounts(),
+	}
+
+	initContainer.Env = append(initContainer.Env, envVars...)
```
caseydavenport, line 1066: I don't understand this - this is appending `envVars` onto `envVars`, is it not?

pkg/render/intrusion_detection.go
```diff
@@ -701,10 +707,15 @@ func (c *intrusionDetectionComponent) deploymentPodTemplate() *corev1.PodTemplat
 			})
 	}
 
-	container := relasticsearch.ContainerDecorateIndexCreator(
-		relasticsearch.ContainerDecorate(c.intrusionDetectionControllerContainer(), c.cfg.ESClusterConfig.ClusterName(),
-			ElasticsearchIntrusionDetectionUserSecret, c.cfg.ClusterDomain, rmeta.OSTypeLinux),
-		c.cfg.ESClusterConfig.Replicas(), c.cfg.ESClusterConfig.Shards())
+	container := c.intrusionDetectionControllerContainer()
+
+	envVars := []corev1.EnvVar{
```
caseydavenport, line 712: I don't think there's any need to keep these appended separately from the other env vars defined in `intrustionDetectionControllerContainer()`

pkg/render/intrusiondetection/dpi/dpi.go
```diff
@@ -95,10 +93,6 @@ func (d *dpiComponent) Objects() (objsToCreate, objsToDelete []client.Object) {
 	}
 	if d.cfg.HasNoDPIResource || d.cfg.HasNoLicense {
 		toDelete = append(toDelete, d.dpiAllowTigeraPolicy())
-		toDelete = append(toDelete, &corev1.Secret{
```
caseydavenport, line 98: Why don't we want to delete this any more?

pkg/render/intrusiondetection/dpi/dpi_test.go
```diff
@@ -531,7 +520,6 @@ var _ = Describe("DPI rendering tests", func() {
 
 		expectedDeleteResources := []client.Object{
 			&v3.NetworkPolicy{ObjectMeta: metav1.ObjectMeta{Name: dpi.DeepPacketInspectionPolicyName, Namespace: dpi.DeepPacketInspectionNamespace}},
-			&corev1.Secret{ObjectMeta: metav1.ObjectMeta{Name: relasticsearch.PublicCertSecret, Namespace: dpi.DeepPacketInspectionNamespace}},
```
caseydavenport, line 534: @asincu can you confirm whether or not we switched DPI to use Linseed yet?

pkg/render/kubecontrollers/kube-controllers.go
```diff
@@ -528,16 +527,6 @@ func (c *kubeControllersComponent) controllersDeployment() *appsv1.Deployment {
 		VolumeMounts:    c.kubeControllersVolumeMounts(),
 	}
 
-	if c.kubeControllerName == EsKubeController {
```
caseydavenport, line 531: I am pretty sure es-calico-kube-controllers still talks to elasticsearch - we still use it in a few scenarios. To name a few:

- For Kibana configuration in any cases that Kibana is enabled.
- Support for managed clusters <= v3.16 (pre linseed)
- Installing the LInseed user in single-tenant clusters

pkg/render/logstorage.go
```diff
@@ -1450,6 +1452,28 @@ func (es elasticsearchComponent) curatorCronJob() *batchv1.CronJob {
...
+		ImagePullPolicy: ImagePullPolicy(),
+		Env:             es.curatorEnvVars(),
+		LivenessProbe:   elasticCuratorLivenessProbe,
+		SecurityContext: securitycontext.NewNonRootContext(),
+		VolumeMounts:    es.cfg.TrustedBundle.VolumeMounts(es.SupportedOSType()),
+	}
+
+	_, esHost, esPort, _ := tigeraurl.ParseEndpoint(relasticsearch.GatewayEndpoint(es.SupportedOSType(), es.cfg.ClusterDomain, ElasticsearchNamespace))
+
+	envVars := []corev1.EnvVar{
```
caseydavenport, line 1467: Why not put these inside `curatorEnvVars()`?

pkg/render/logstorage/esgateway/esgateway.go
```diff
@@ -119,12 +118,6 @@ func (e *esGateway) Objects() (toCreate, toDelete []client.Object) {
 	toCreate = append(toCreate, e.esGatewayServiceAccount())
 	toCreate = append(toCreate, e.esGatewayDeployment())
 
-	// The following secret is used by kube controllers and sent to managed clusters. It is also used by manifests in our docs.
```
caseydavenport, line 122: I think this is also still needed to support clusters <= v3.16 (that aren't connecting to Linseed)

pkg/render/logstorage/esmetrics/elasticsearch_metrics.go
```diff
@@ -209,6 +211,39 @@ func (e elasticsearchMetrics) metricsDeployment() *appsv1.Deployment {
...
+		VolumeMounts: append(
+			e.cfg.TrustedBundle.VolumeMounts(e.SupportedOSType()),
+			e.cfg.ServerTLS.VolumeMount(e.SupportedOSType()),
+		),
+		Env: []corev1.EnvVar{
+			{Name: "FIPS_MODE_ENABLED", Value: operatorv1.IsFIPSModeEnabledString(e.cfg.Installation.FIPSMode)},
+		},
+	}
+
+	container.Env = append(container.Env, envVars...)
```
caseydavenport, line 245: I think throughout we should clean up this split env declaration - it's a bit confusing the way this is laid out here.

- Define envVars
- Create a container with a separate set of hard-coded env vars
- Append the previously created env vars to the container's env.

We should just be able to define all of the env vars in one place.

pkg/render/logstorage/esmetrics/elasticsearch_metrics.go
```diff
@@ -228,36 +263,14 @@ func (e elasticsearchMetrics) metricsDeployment() *appsv1.Deployment {
 					ServiceAccountName: ElasticsearchMetricsName,
 					InitContainers:     initContainers,
 					Containers: []corev1.Container{
-						relasticsearch.ContainerDecorate(
```
caseydavenport, line 231: Changing how we define the env vars as suggested above means we can also revert this change and keep defining the container inline. Better to touch only the lines of code we need (in this case, the env vars)

pkg/render/manager.go
```diff
@@ -567,6 +562,21 @@ func (c *managerComponent) managerEsProxyContainer() corev1.Container {
...
+			relasticsearch.ElasticCAEnvVar(c.SupportedOSType()),
+			relasticsearch.ElasticSchemeEnvVar(esScheme),
+			relasticsearch.ElasticHostEnvVar(esHost),
+			relasticsearch.ElasticPortEnvVar(esPort),
+			relasticsearch.ElasticUsernameEnvVar(ElasticsearchManagerUserSecret),
+			relasticsearch.ElasticPasswordEnvVar(ElasticsearchManagerUserSecret),
+			relasticsearch.ElasticIndexSuffixEnvVar(c.cfg.ClusterConfig.ClusterName()),
+		}
+
+		env = append(env, esEnvVars...)
```
caseydavenport, line 577: We don't need to define the intermediate variable

```go
env = append(env, []corev1.EnvVar{
  relasticsearch.ElasticCAEnvVar(c.SupportedOSType()),
  . . .
})

File: pkg/render/manager_test.go [hunk 12]
```diff
@@ -695,9 +703,13 @@ var _ = Describe("Tigera Secure Manager rendering tests", func() {
 	// renderManager passes in as few parameters as possible to render.Manager without it
 	// panicing. It accepts variations on the installspec for testing purposes.
 	renderManager := func(i *operatorv1.InstallationSpec) *appsv1.Deployment {
+		var esConfigMap *relasticsearch.ClusterConfig
+		if !operatorv1.IsFIPSModeEnabled(i.FIPSMode) {
```
caseydavenport, line 707: It's not obvious why we pass this in only for FIPS disabled - a comment would help here.

pkg/render/manager_test.go
```diff
@@ -1050,7 +1062,10 @@ func renderObjects(roc renderConfig) []client.Object {
 		roc.bindingNamespaces = []string{roc.ns}
 	}
 
-	esConfigMap := relasticsearch.NewClusterConfig("clusterTestName", 1, 1, 1)
+	var esConfigMap *relasticsearch.ClusterConfig
+	if roc.tenant == nil && !operatorv1.IsFIPSModeEnabled(roc.installation.FIPSMode) {
```
caseydavenport, line 1066: Same here - the `if` statement isn't self-explanatory

pkg/render/fluentd.go
Josh-Tigera: These are added unconditionally but we're already inside the `eksLogForwarderDeployment` function which does the checks necessary to ensure we're only doing this in the case of EKS log forwarding being enabled.

pkg/render/intrusion_detection.go
Josh-Tigera: Agreed and this was a change I already had in my next set of changes to be pushed while I was fixing some issues I found once I loaded my images into a cluster.

pkg/render/intrusiondetection/dpi/dpi.go
Josh-Tigera: I removed all creations of it so if it's no longer being created it seems redundant to have deletion logic for something that isn't created. I'm realizing now that those creations were never in the `DeepPacketInspectionNamespace` so I'm not actually sure if or even how this is ever created. I can try to track it down.

pkg/render/kubecontrollers/kube-controllers.go
Josh-Tigera: Both my standalone and single-tenant management-managed clusters seem ok without this and I can't find any references in existing kube-controllers code to either raw `ELASTIC_*` env vars or LMA calls to fetch/load them. I can revert it for the version skew situation alone though. I'll verify their usage in the versions we'll still have to support though, would 3.16 be the oldest we have to worry about? Or do we have to support even further back?

pkg/render/logstorage.go
Josh-Tigera: Can do.

pkg/render/logstorage/esgateway/esgateway.go
Josh-Tigera: Ok I'll verify.

pkg/render/logstorage/esmetrics/elasticsearch_metrics.go
Josh-Tigera: Agreed that's what I have in my current iteration.

pkg/render/manager_test.go
Josh-Tigera: True, I'll port the comment I had in the controller here as well.

pkg/render/intrusiondetection/dpi/dpi.go
caseydavenport: Stopping the creation of something can be done independently from deleting it - clusters that have had it created by a past version may still need for it to be deleted on upgrade, even if the newer version no longer creates it.

If anything, we should start deleting this unconditionally so that it's always removed if it's no longer needed at all (and then at some future point we can remove the deletion once we're confident that we don't need to support direct upgrade from a version that had previously created this)

pkg/controller/installation/core_controller.go
```diff
@@ -1198,18 +1196,6 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 			typhaNodeTLS.TrustedBundle.AddCertificates(prometheusClientCert)
 		}
 
-		// es-kube-controllers needs to trust the ESGW certificate. We'll fetch it here and add it to the trusted bundle.
```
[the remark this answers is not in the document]
Josh-Tigera: I have a setup with a management cluster running this operator code, with successfully connected 316 and master managed clusters, as well as a managed cluster running this operator code. I can access kibana from the management cluster. Is there something specific within Kibana I can check for that would indicate whether or not this is causing any failures?

pkg/render/kubecontrollers/kube-controllers.go
Josh-Tigera: Added back only the specific values we need to support the 3.16 managed cluster case.

pkg/controller/installation/core_controller.go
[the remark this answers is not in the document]
caseydavenport: We support a few different modes of certificate provisioning in enterprise.

1. All certificates signed by the tigera-operator CA
2. All certificates signed by an external CSR, with a unique certificate per-pod.
3. What we call "BYO" certificates, which means users can provide their own hand-spun certificate for each component.

I suspect the reason this works in your environment is because the default is option (1), which doesn't require we add the individual certs to the trusted bundle because the tigera-operator CA will already be in the bundle.

While I'm not aware of anyone actually doing option (3), this is the mode in which this code is needed - if the user provides a certificate for es-kube-controllers that isn't actually signed by the tigera-operator signer, then we need to add that certificate to the trusted bundle explicitly.

CC @rene-dekker since we've spoken about this a few times in various contexts.... I really want to stop supporting BYO somehow because it's just so confusing.

pkg/controller/intrusiondetection/intrusiondetection_controller.go
```diff
@@ -393,6 +388,14 @@ func (r *ReconcileIntrusionDetection) Reconcile(ctx context.Context, request rec
 		return reconcile.Result{}, err
 	}
 
+	if isManagedCluster {
+		if esClusterConfig.ClusterName() == render.DefaultElasticsearchClusterName {
+			err = fmt.Errorf("Elasticsearch cluster name must be non-default value in managed clusters")
+			r.status.SetDegraded(operatorv1.InvalidConfigurationError, "", err, reqLogger)
```
caseydavenport, line 394: This is an interesting new assertion - any background on this one? It seems safe to me, but curious what prompted you to add it!

I do think we need a user-facing message in the SetDegraded call instead of `""` though.

Also, this isnt' the Elasticsearch cluster name. Rather this is the just the name of this managed cluster (which is primarily used by our elastic log ingestion pipeline).

pkg/controller/intrusiondetection/intrusiondetection_controller.go
```diff
@@ -422,16 +425,6 @@ func (r *ReconcileIntrusionDetection) Reconcile(ctx context.Context, request rec
 		return reconcile.Result{}, err
 	}
 
-	esgwCertificate, err := certificateManager.GetCertificate(r.client, relasticsearch.PublicCertSecret, common.OperatorNamespace())
```
caseydavenport, line 425: From what I can tell, intrusion-detection-controller is one of the few managed cluster components that still talks directly to elasticsearch via es-gateway. 

It does talk to Linseed in most codepaths, but when searching for security events it seems it still talks to ES.

pkg/controller/logcollector/logcollector_controller.go
```diff
@@ -141,7 +142,7 @@ func add(mgr manager.Manager, c controller.Controller) error {
 
 	for _, secretName := range []string{
 		render.ElasticsearchEksLogForwarderUserSecret,
-		relasticsearch.PublicCertSecret, render.S3FluentdSecretName, render.EksLogForwarderSecret,
```
caseydavenport, line 144: fluentd still talks to elasticsearch, but @vara2504 is working on removing that dependency as we speak so we should be good to keep this change once her PR goes in.

pkg/controller/manager/manager_controller.go
```diff
@@ -458,14 +458,19 @@ func (r *ReconcileManager) Reconcile(ctx context.Context, request reconcile.Requ
...
 	}
 
-	clusterConfig, err := utils.GetElasticsearchClusterConfig(context.Background(), r.client)
-	if err != nil {
-		if errors.IsNotFound(err) {
-			r.status.SetDegraded(operatorv1.ResourceNotFound, "Elasticsearch cluster configuration is not available, waiting for it to become available", err, logc)
-			return reconcile.Result{}, nil
+	var clusterConfig *relasticsearch.ClusterConfig
+	// We only require Elastic cluster configuration when Kibana is enabled. We infer whether Kibana is enabled by checking
+	// FIPS configuration mode and multi-tenancy mode. See manager.go function kibanaEnabled for more details.
```
caseydavenport, line 463: We should probably make `kibnanEnabled` public and use the same implementation everywhere that we need to check if Kibana is enabled, so that they don't get out of sync.

pkg/controller/manager/manager_controller_test.go
```diff
@@ -566,6 +560,12 @@ var _ = Describe("Manager controller tests", func() {
 		})
 
 		Context("image reconciliation", func() {
+			BeforeEach(func() {
+				Expect(c.Create(
```
caseydavenport, line 564: This can be a single line - no need to split it up.

Alternatively, we could keep this in the main `BeforeEach` and then any tests that need something different can just call `Update` in their `BeforeEach` to modify the installation.

pkg/render/fluentd.go
```diff
@@ -1040,10 +1042,38 @@ func (c *fluentdComponent) eksLogForwarderDeployment() *appsv1.Deployment {
 		{Name: "AWS_REGION", Value: c.cfg.EKSConfig.AwsRegion},
 		{Name: "AWS_ACCESS_KEY_ID", ValueFrom: secret.GetEnvVarSource(EksLogForwarderSecret, EksLogForwarderAwsId, false)},
 		{Name: "AWS_SECRET_ACCESS_KEY", ValueFrom: secret.GetEnvVarSource(EksLogForwarderSecret, EksLogForwarderAwsKey, false)},
+		relasticsearch.ElasticIndexSuffixEnvVar(c.cfg.ESClusterConfig.ClusterName()),
+		relasticsearch.ElasticUserEnvVar(ElasticsearchEksLogForwarderUserSecret),
+		relasticsearch.ElasticPasswordEnvVar(ElasticsearchEksLogForwarderUserSecret),
+		relasticsearch.ElasticSchemeEnvVar(esScheme),
+		relasticsearch.ElasticHostEnvVar(esHost),
```
caseydavenport, line 1049: One to keep an eye on - once @vara2504 is done with her work for EKS log forwarder, we can remove most (all?) of these.

pkg/render/fluentd.go
```diff
@@ -1040,10 +1042,38 @@ func (c *fluentdComponent) eksLogForwarderDeployment() *appsv1.Deployment {
...
+		relasticsearch.ElasticPortEnvVar(esPort),
+		relasticsearch.ElasticCAEnvVar(c.cfg.OSType),
 	}
 
 	var eksLogForwarderReplicas int32 = 1
 
+	initContainer := corev1.Container{
```
caseydavenport, line 1056: I don't think there's any reason we need to pull these out here - can we leave them where they were and just remove the `ContainerDecorate` calls just to keep the changes minimal?

pkg/render/fluentd.go
```diff
@@ -1040,10 +1042,38 @@ func (c *fluentdComponent) eksLogForwarderDeployment() *appsv1.Deployment {
...
+	container := corev1.Container{
+		Name:            eksLogForwarderName,
+		Image:           c.image,
+		ImagePullPolicy: ImagePullPolicy(),
+		Env:             envVars,
+		SecurityContext: c.securityContext(false),
+		VolumeMounts:    c.eksLogForwarderVolumeMounts(),
+	}
+
+	container.Env = append(container.Env, envVars...)
```
caseydavenport, line 1075: This is a bug I think - looks like we're appending `envVars...` on top of `envVars` so we'll have two of each.

pkg/render/fluentd_test.go
```diff
@@ -884,6 +895,15 @@ var _ = Describe("Tigera Secure Fluentd rendering tests", func() {
 			}))
 
 		Expect(envs).To(ContainElement(corev1.EnvVar{Name: "EKS_CLOUDWATCH_LOG_FETCH_INTERVAL", Value: "900"}))
+		Expect(envs).To(ContainElements([]corev1.EnvVar{
```
caseydavenport, line 898: Probably want the reverse check as well to catch any cases where there are unexpected env vars. And maybe a length check to catch bugs like the one above where we added env vars twice.

e.g.,

```go
expected := []corev1.EnvVar{...}

Expect(envs).Should(ContainElements(expected))
Expect(expected).Should(ContainElements(envs))
Expect(envs).Should(HaveLength(len(expected))
```

Also, we can merge the assertion on the line right above this.

pkg/controller/installation/core_controller.go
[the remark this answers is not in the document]
rene-dekker: I think we need to continue to support it for the manager-tls secret. I would propose simply un-documenting the rest.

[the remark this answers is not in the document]
rene-dekker: Also note that this code being mistakenly removed here is necessary also for certificate issued by the operator before v3.13.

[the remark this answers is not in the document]
caseydavenport: Sounds like we should:

- Remove documentation for all but manager-tls
- Leave the code in place, and add a big comment explaining that this is needed because prior to v3.13 secrets were not signed by a single CA and so we need to include each individually.

That said, we can remove this block of code when kube-controllers no longer talks to es-gateway in any capabity, which hopefully will happen in the future.

pkg/controller/logstorage/elastic/elastic_controller.go
```diff
@@ -227,10 +225,6 @@ func Add(mgr manager.Manager, opts options.AddOptions) error {
 		}
 	}
 
-	if err = utils.AddConfigMapWatch(c, relasticsearch.ClusterConfigConfigMapName, common.OperatorNamespace(), &handler.EnqueueRequestForObject{}); err != nil {
```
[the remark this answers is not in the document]
caseydavenport: > Josh: Apparently I can modify other people's comments?

:laughing:

caseydavenport:
@Josh-Tigera a few conflicts to resolve, though

PR #2976, opened 2023-11-02
[RS-1209] Webhooks processor deployment in the operator.

bartolini:
This PR is to deploy `webhooks-processor` alongside with `intrussion-detection-controller` (same pod deployment).

### DIFF
--- a/pkg/render/intrusion_detection.go
+++ b/pkg/render/intrusion_detection.go
@@ -437,6 +437,10 @@ func (c *intrusionDetectionComponent) intrusionDetectionJobServiceAccount() *cor
 }
 
 func (c *intrusionDetectionComponent) intrusionDetectionClusterRole() *rbacv1.ClusterRole {
+	stdApiResources := []string{"podtemplates"}
+	if c.deployWebhooksController() {
+		stdApiResources = append(stdApiResources, "secrets", "configmaps")
+	}
 	rules := []rbacv1.PolicyRule{
 		{
 			APIGroups: []string{
@@ -466,7 +470,7 @@ func (c *intrusionDetectionComponent) intrusionDetectionClusterRole() *rbacv1.Cl
 		},
 		{
 			APIGroups: []string{""},
-			Resources: []string{"podtemplates", "secrets", "configmaps"},
+			Resources: stdApiResources,
 			Verbs:     []string{"get"},
 		},
 		{
@@ -739,8 +743,7 @@ func (c *intrusionDetectionComponent) deploymentPodTemplate() *corev1.PodTemplat
 		intrusionDetectionContainer,
 	}
 
-	// deploy webhooks controller container only for managed clusters or stand-alone enterprise clusters
-	if c.cfg.ManagedCluster || (!c.cfg.ManagedCluster && !c.cfg.ManagementCluster) {
+	if c.deployWebhooksController() {
 		containers = append(containers, c.webhooksControllerContainer())
 	}
 
@@ -760,6 +763,12 @@ func (c *intrusionDetectionComponent) deploymentPodTemplate() *corev1.PodTemplat
 			Volumes:            volumes,
 		},
 	}, c.cfg.ESClusterConfig, c.cfg.ESSecrets).(*corev1.PodTemplateSpec)
+}
+
+func (c *intrusionDetectionComponent) deployWebhooksController() bool {
+	// deploy webhooks controller container only for:
+	return c.cfg.ManagedCluster || // managed clusters or ...
+		(!c.cfg.ManagedCluster && !c.cfg.ManagementCluster) // stand-alone enterprise clusters
 }
 
 func (c *intrusionDetectionComponent) webhooksControllerContainer() corev1.Container {
--- a/pkg/render/intrusion_detection.go
+++ b/pkg/render/intrusion_detection.go
@@ -782,6 +782,7 @@ func (c *intrusionDetectionComponent) webhooksControllerContainer() corev1.Conta
 	}
 
 	volumeMounts := c.cfg.TrustedCertBundle.VolumeMounts(c.SupportedOSType())
+	volumeMounts = append(volumeMounts, c.cfg.IntrusionDetectionCertSecret.VolumeMount(c.SupportedOSType()))
 	if c.cfg.ManagedCluster {
 		volumeMounts = append(volumeMounts,
 			corev1.VolumeMount{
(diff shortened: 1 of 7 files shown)


gantony (approved):
LGTM

pkg/render/intrusion_detection.go
```diff
@@ -459,7 +465,7 @@ func (c *intrusionDetectionComponent) intrusionDetectionClusterRole() *rbacv1.Cl
 		},
 		{
 			APIGroups: []string{""},
-			Resources: []string{"podtemplates"},
+			Resources: []string{"podtemplates", "secrets", "configmaps"},
 			Verbs:     []string{"get"},
```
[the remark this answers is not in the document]
bartolini: I'm resolving this comment, we discussed possible solutions over a zoom call.

caseydavenport (changes requested):
One bit of feedback would be to only install this container conditionally when it is actually needed.

- Is it needed in Management clusters?
- Managed clusters?
- Standalone enterprise clusters?

bartolini:
@caseydavenport thank you for you review.

The new container needs to be deployed only on managed clusters and stand-alone enterprise clusters. I have missed that bit, relevant changes are now in place. Please have a look whenever you get a chance.

pkg/render/intrusion_detection.go
```diff
@@ -430,6 +437,10 @@ func (c *intrusionDetectionComponent) intrusionDetectionJobServiceAccount() *cor
 }
 
 func (c *intrusionDetectionComponent) intrusionDetectionClusterRole() *rbacv1.ClusterRole {
+	stdApiResources := []string{"podtemplates"}
+	if c.deployWebhooksController() {
+		stdApiResources = append(stdApiResources, "secrets", "configmaps")
```
caseydavenport, line 442: In golang, acryonyms shoudl be same-case.  Meaning this should be `stdAPIResources`.

While here though, I think this should be renamed entirely since it's not clear what `std` means in this context. Maybe this shoudl be called `coreGetResources` since these are the resources in the core API group that need get permissions?

pkg/render/intrusion_detection.go
```diff
@@ -492,6 +503,17 @@ func (c *intrusionDetectionComponent) intrusionDetectionClusterRole() *rbacv1.Cl
 			},
 			Verbs: []string{"get"},
 		},
+		{
```
caseydavenport, line 506: is this always needed? Or just when running the webhook controller?

pkg/render/intrusion_detection.go
```diff
@@ -710,13 +732,21 @@ func (c *intrusionDetectionComponent) deploymentPodTemplate() *corev1.PodTemplat
...
 		}
-		container.Env = append(container.Env, envVars...)
+		intrusionDetectionContainer.Env = append(intrusionDetectionContainer.Env, envVars...)
 	}
 	var initContainers []corev1.Container
 	if c.cfg.IntrusionDetectionCertSecret != nil && c.cfg.IntrusionDetectionCertSecret.UseCertificateManagement() {
 		initContainers = append(initContainers, c.cfg.IntrusionDetectionCertSecret.InitContainer(IntrusionDetectionNamespace))
 	}
 
+	containers := []corev1.Container{
```
caseydavenport, line 742: This can just be one line

pkg/render/intrusion_detection.go
```diff
@@ -729,14 +759,62 @@ func (c *intrusionDetectionComponent) deploymentPodTemplate() *corev1.PodTemplat
...
-			Volumes: volumes,
+			Containers:         containers,
+			Volumes:            volumes,
 		},
 	}, c.cfg.ESClusterConfig, c.cfg.ESSecrets).(*corev1.PodTemplateSpec)
 }
 
+func (c *intrusionDetectionComponent) deployWebhooksController() bool {
+	// deploy webhooks controller container only for:
+	return c.cfg.ManagedCluster || // managed clusters or ...
```
caseydavenport, line 770: This should also be one line and the comment should be written out in its entirety above the return statement. At least in this codebase, but also more generally, we usually avoid comments at the end of lines.

This can also be simplified to this:
```go
c.cfg.ManagedCluster || !c.cfg.ManagementCluster
```

Since the `!c.cfg.ManagedCluster` is implied due to the `||` condition.

pkg/render/intrusion_detection_test.go
```diff
@@ -154,7 +154,7 @@ var _ = Describe("Intrusion Detection rendering tests", func() {
 		// Should mount ManagerTLSSecret for non-managed clusters
 		idc := rtest.GetResource(resources, "intrusion-detection-controller", render.IntrusionDetectionNamespace, "apps", "v1", "Deployment").(*appsv1.Deployment)
 		idji := rtest.GetResource(resources, "intrusion-detection-es-job-installer", render.IntrusionDetectionNamespace, "batch", "v1", "Job").(*batchv1.Job)
-		Expect(idc.Spec.Template.Spec.Containers).To(HaveLen(1))
+		Expect(idc.Spec.Template.Spec.Containers).To(HaveLen(2))
```
caseydavenport, line 157: I think we need a test for management clusters that asserts the webhook controller isn't installed.

From what I can tell, this test file is missing ManagementCluster test cases altogether.

pkg/render/intrusion_detection.go
bartolini: Good point, I have reorganized this function a bit so this variable is no longer required.

pkg/render/intrusion_detection.go
bartolini: Done.

pkg/render/intrusion_detection.go
bartolini: Good point, changed.

pkg/render/intrusion_detection.go
bartolini: This is required only when running the webhook controller. Changed

pkg/render/intrusion_detection_test.go
bartolini: I was thinking about that too. I will create a new test case for management clusters. Although I'm not sure if it should belong here or in the `operator-cloud` codebase. I'll let you know when this is completed.

caseydavenport: It belongs here (operator-cloud is on freeze for new code)

bartolini: @caseydavenport Tests are now implemented. Pleas have another look.

caseydavenport (approved):
Looks good to me, thanks @bartolini !

caseydavenport:
@bartolini FYI I squashed your commits since they seemed mostly dev commits rather than distinct bits of functionality.

PR #2978, opened 2023-11-02
Remove old controller ref for Voltron secret

caseydavenport:
## Description

We recently moved ownership of this secret from the APIServer controller
to the Manager controller. However, on upgrade, older versions of this
secret will still have the old owner reference set, resulting in two
owners and thus an error.

If we see this case, simply remove the old owner reference.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/manager/manager_controller.go
+++ b/pkg/controller/manager/manager_controller.go
@@ -554,6 +554,17 @@ func (r *ReconcileManager) Reconcile(ctx context.Context, request reconcile.Requ
 				r.status.SetDegraded(operatorv1.ResourceCreateError, "Unable to create the tunnel secret", err, logc)
 				return reconcile.Result{}, err
 			}
+		} else {
+			// Check controller references and remove any old APIServer ownership, since ownership of this resource has moved
+			// to the manager controller instead. Without this, we will hit an error when trying to update the secret as it will
+			// have two controllers set.
+			for i := 0; i < len(tunnelCASecret.OwnerReferences); i++ {
+				ref := tunnelCASecret.OwnerReferences[i]
+				if ref.Kind == "APIServer" && ref.Controller != nil && *ref.Controller {
+					tunnelCASecret.OwnerReferences = append(tunnelCASecret.OwnerReferences[:i], tunnelCASecret.OwnerReferences[i+1:]...)
+					i--
+				}
+			}
 		}
 
 		// We use the CA as the server cert.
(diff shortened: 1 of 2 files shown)


pkg/controller/manager/manager_controller.go
```diff
@@ -554,6 +554,17 @@ func (r *ReconcileManager) Reconcile(ctx context.Context, request reconcile.Requ
...
 			}
+		} else {
+			// Check controller references and remove any old APIServer ownership, since ownership of this resource has moved
+			// to the manager controller instead. Without this, we will hit an error when trying to update the secret as it will
+			// have two controllers set.
+			for i := 0; i < len(tunnelCASecret.OwnerReferences); i++ {
+				ref := tunnelCASecret.OwnerReferences[i]
+				if ref.Kind == "APIServer" && ref.Controller != nil && *ref.Controller {
+					tunnelCASecret.OwnerReferences = append(tunnelCASecret.OwnerReferences[:i], tunnelCASecret.OwnerReferences[i+1:]...)
+					i--
```
rene-dekker, line 565: Any reason we don't break the loop?

pkg/controller/manager/manager_controller.go
```diff
@@ -554,6 +554,17 @@ func (r *ReconcileManager) Reconcile(ctx context.Context, request reconcile.Requ
...
 				return reconcile.Result{}, err
 			}
+		} else {
+			// Check controller references and remove any old APIServer ownership, since ownership of this resource has moved
+			// to the manager controller instead. Without this, we will hit an error when trying to update the secret as it will
+			// have two controllers set.
+			for i := 0; i < len(tunnelCASecret.OwnerReferences); i++ {
+				ref := tunnelCASecret.OwnerReferences[i]
+				if ref.Kind == "APIServer" && ref.Controller != nil && *ref.Controller {
+					tunnelCASecret.OwnerReferences = append(tunnelCASecret.OwnerReferences[:i], tunnelCASecret.OwnerReferences[i+1:]...)
```
Brian-McM, line 564: Wouldn't it be more correct to remove any reference that isn't the manager, or find the manager reference and overwrite the owner reference if it's found (or set it to empty if it wasn't found)? Technically this code would still allow for multiple owners to be set, it's just banking on the fact that there shouldn't be any other owner other than the API.

caseydavenport: @Brian-McM If there's another owner that doesn't match the condition laid out here, then I want it to fail. Because this is the only known scenario that we want to upgrade from.

Other cases would be something unexpected, so I'm OK with not handling those.

rene-dekker (approved):
lgtm

pkg/controller/manager/manager_controller.go
Brian-McM: Gotcha, sounds reasonable.

Brian-McM (approved):

PR #2989, opened 2023-11-07
Enable OIDC for multi-tenant setup

asincu:
## Description

Enable OIDC when configuring a multi-tenant management cluster.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/manager/manager_controller.go
+++ b/pkg/controller/manager/manager_controller.go
@@ -386,7 +386,7 @@ func (r *ReconcileManager) Reconcile(ctx context.Context, request reconcile.Requ
 	// Build a trusted bundle containing all of the certificates of components that communicate with the manager pod.
 	// This bundle contains the root CA used to sign all operator-generated certificates, as well as the explicitly named
 	// certificates, in case the user has provided their own cert in lieu of the default certificate.
-	var authenticationCR *operatorv1.Authentication
+
 	var trustedSecretNames []string
 	if !r.multiTenant {
 		// For multi-tenant systems, we don't support user-provided certs for all components. So, we don't need to include these,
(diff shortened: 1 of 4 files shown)


pkg/controller/manager/manager_controller.go
```diff
@@ -404,19 +404,20 @@ func (r *ReconcileManager) Reconcile(ctx context.Context, request reconcile.Requ
...
-			r.status.SetDegraded(operatorv1.ResourceReadError, "Error while fetching Authentication", err, logc)
-			return reconcile.Result{}, err
-		}
-		if authenticationCR != nil && authenticationCR.Status.State != operatorv1.TigeraStatusReady {
-			r.status.SetDegraded(operatorv1.ResourceNotReady, fmt.Sprintf("Authentication is not ready authenticationCR status: %s", authenticationCR.Status.State), nil, logc)
-			return reconcile.Result{}, nil
-		} else if authenticationCR != nil {
-			trustedSecretNames = append(trustedSecretNames, render.DexTLSSecretName)
-		}
+	var authenticationCR *operatorv1.Authentication
```
caseydavenport, line 409: I'm not sure this is the right way to do this - I think we will need to namespace this CR, no?

Otherwise, all tenants will share the same Authentication configuration?

asincu: I was thinking the same thing. At the moment, I was simply trying to get the environment populated for a single tenant.

asincu: Talked offline. It seems that Cloud uses a TigeraOIDC and does not enable it per each tenant. ManagerURL is used to configure Kibana and IssuerURL and Groups claims have the same value across tenants.

caseydavenport: I think we should add validation so that when running in multi-tenant mode, we reject an Authentication resource that isn't supported (e.g., Dex mode and other unsupported fields in multi-tenant)

pkg/controller/manager/manager_controller.go
```diff
@@ -404,19 +404,24 @@ func (r *ReconcileManager) Reconcile(ctx context.Context, request reconcile.Requ
...
+	authenticationCR, err = utils.GetAuthentication(ctx, r.client)
+	if err != nil && !errors.IsNotFound(err) {
+		r.status.SetDegraded(operatorv1.ResourceReadError, "Error while fetching Authentication", err, logc)
+		return reconcile.Result{}, err
+	}
+	if authenticationCR != nil && authenticationCR.Status.State != operatorv1.TigeraStatusReady {
+		r.status.SetDegraded(operatorv1.ResourceNotReady, fmt.Sprintf("Authentication is not ready authenticationCR status: %s", authenticationCR.Status.State), nil, logc)
+		return reconcile.Result{}, nil
+	} else if authenticationCR != nil && !r.multiTenant {
+		// A multi-tenant setup use TigeraOIDC and do not spin up Dex
```
caseydavenport, line 420: This comment is confusing because it says this is a multi-tenant setup but the branch is explicitly for NOT multi-tenant clusters.

pkg/controller/manager/manager_controller.go
```diff
@@ -404,19 +404,24 @@ func (r *ReconcileManager) Reconcile(ctx context.Context, request reconcile.Requ
...
-			trustedSecretNames = append(trustedSecretNames, render.DexTLSSecretName)
-		}
+	var authenticationCR *operatorv1.Authentication
+	// Fetch the Authentication spec. If present, we use to configure user authentication.
+	authenticationCR, err = utils.GetAuthentication(ctx, r.client)
+	if err != nil && !errors.IsNotFound(err) {
+		r.status.SetDegraded(operatorv1.ResourceReadError, "Error while fetching Authentication", err, logc)
+		return reconcile.Result{}, err
+	}
+	if authenticationCR != nil && authenticationCR.Status.State != operatorv1.TigeraStatusReady {
```
caseydavenport, line 416: I think the validation part of this should be moved to the existing validation function for authentication here: https://github.com/tigera/operator/blob/3919c836a4518f6ff5569fcef9551f99930f3933/pkg/controller/authentication/authentication_controller.go#L409-L474 - this will probably require adding a `multiTenant` field to the authentication reconciler, but that's fairly easy. 

The was this is designed, the authentication controller defaults and validates the CR, performs any necessary actions, and then marks the CR as "Ready".

This controller waits for the CR to be "Ready" before continuing.

pkg/controller/manager/manager_controller.go
caseydavenport: That said, I don't think `r.multiTenant` is the correct check to make here. This is the logic we use to determine if dex should be installed: https://github.com/tigera/operator/blob/3919c836a4518f6ff5569fcef9551f99930f3933/pkg/controller/authentication/authentication_controller.go#L320-L323

We should use the same logic (perhaps with a common helper function?) to decide whether or not to include the Dex TLS secret in the trusted bundle.

caseydavenport (changes requested):

caseydavenport (approved):

PR #2990, opened 2023-11-08
Use trusted bundle with root CAs for multi-tenant manager pods

caseydavenport:
## Description

Manager pods need to use a trusted bundle that includes the system
certificates, but other pods do not. 

This PR introduces the ability in a multi-tenant system to deploy both
types of bundle, and enables individiual components to decide which of
the two they need to load.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/tls/certificatemanagement/interface.go
+++ b/pkg/tls/certificatemanagement/interface.go
@@ -35,7 +35,7 @@ const (
 
 	// TrustedCertConfigMapNamePublic is the name of the trusted certificate bundle ConfigMap that includes public CAs, used
 	// only in multi-tenant environments as a single namespace requires both a trusted bundle with public CAs as well as one without.
-	TrustedCertConfigMapNamePublic = "tigera-ca-bundle-public"
+	TrustedCertConfigMapNamePublic = "tigera-ca-bundle-system-certs"
 )
 
 // KeyPairInterface wraps a Secret object that contains a private key and a certificate. Whether CertificateManagement is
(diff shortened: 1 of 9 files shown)


pkg/tls/certificatemanagement/interface.go
```diff
@@ -23,12 +23,19 @@ import (
...
 	TrustedCertBundleMountPath        = "/etc/pki/tls/certs/tigera-ca-bundle.crt"
 	TrustedCertBundleMountPathWindows = "c:/etc/pki/tls/certs/tigera-ca-bundle.crt"
+
+	// TrustedCertConfigMapName is the name of the trusted certificate bundle ConfigMap. This value is used
+	// for all single-tenant trusted bundles, as well as multi-tenant trusted bundles that do not include public CAs.
+	TrustedCertConfigMapName = "tigera-ca-bundle"
+
+	// TrustedCertConfigMapNamePublic is the name of the trusted certificate bundle ConfigMap that includes public CAs, used
+	// only in multi-tenant environments as a single namespace requires both a trusted bundle with public CAs as well as one without.
+	TrustedCertConfigMapNamePublic = "tigera-ca-bundle-public"
```
caseydavenport, line 38: Probably want to rename this to `tigera-ca-bundle-with-system-roots`

asincu: I see that for a single tenant, we should use the same config map `tigera-ca-bundle`, but with the following fields
- `tigera-ca-bundle.crt` holds certificates generated by tigera ca
- `ca-bundle.crt`holds public certificates

Why can't we use the same mechanism ?

caseydavenport: We don't want to mount the public certificates into ever pod, and this approach just seemed the most in line with how we currently manage these for enterprise. I think it's a bit easier to grasp conceptually.

asincu (commented):
Looks good! I think some static checks are failing, but the correct bundle is mounted now.

asincu (approved):

PR #2991, opened 2023-11-08
update CoreRuleset payload for dikastes' new WAF engine

electricjesus:
## Description

Now that we're changing the WAF engine in Dikastes from Modsecurity to Coraza, this PR provides mainly two changes that ties into that changel: 

1. a few changes to command-line invocation, and
2.  an update to the coreruleset version

Here is a more detailed breakdown on the changes in this PR

- update dikastes command-line options to reflect new cli options for dikastes that runs coraza-based waf (pkg/render/applicationlayer/applicationlayer.go)
- bumped coreruleset version v3.3.5 to v4.0.0-rc2
- move and reorganise coreruleset delivery
  - removed the 'go:generate'-based workflow, this removes the generated artifacts and reduces the possibility of not updating it everytime we need an update to coreruleset. This also means we replace generated files with embedded references in applicationlayer renderer and controller.
  - reorganise how we do the customisations: all our custom CRS behaviour in its own `tigera.conf` file
  - coreruleset files now follow a rules subdirectory format, which is the original file/directory structure in github.com/coreruleset/coreruleset

Depends on https://github.com/tigera/calico-private/pull/6862 but that PR also depends on this one.

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


pkg/render/applicationlayer/applicationlayer.go
```diff
@@ -295,17 +295,24 @@ func (c *component) containers() []corev1.Container {
 		}
 
 		if c.config.WAFEnabled {
-			commandArgs = append(commandArgs, "--rules", "/etc/modsecurity-ruleset")
+			commandArgs = append(
+				commandArgs,
+				"--waf-enabled",
```
sigsegv1989, line 300: I think we need to pass --dial as /var/run/nodeagent/socket command line argument to Dikastes

sigsegv1989: It's already added to line# 288

caseydavenport (approved):
LGTM - one minor nit about variables being public.

PR #2992, opened 2023-11-10
[EV-4356] createPredicateForObject should react if the generation isn…

dimitri-nicolo:
…'t set

## Description

Updates to watched objects do not propagate a call to the `Reconcile`, if the `ObjectMeta::Generation` isn't defined. The change checks the old object in an update and if that is zero sets `generationChanged` to 0.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/utils/utils.go
+++ b/pkg/controller/utils/utils.go
@@ -647,7 +647,9 @@ func createPredicateForObject(objMeta metav1.Object) predicate.Predicate {
 			return e.Object.GetNamespace() == objMeta.GetNamespace() || objMeta.GetNamespace() == ""
 		},
 		UpdateFunc: func(e event.UpdateEvent) bool {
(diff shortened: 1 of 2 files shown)


pkg/controller/utils/utils.go
```diff
@@ -647,7 +647,8 @@ func createPredicateForObject(objMeta metav1.Object) predicate.Predicate {
 			return e.Object.GetNamespace() == objMeta.GetNamespace() || objMeta.GetNamespace() == ""
 		},
 		UpdateFunc: func(e event.UpdateEvent) bool {
-			generationChanged := e.ObjectOld.GetGeneration() != e.ObjectNew.GetGeneration()
+			// React if the generation isn't set (== 0).
```
rene-dekker, line 650: Could you add a comment saying that not all objects use/have a generation and that therefore, this addition is required?

rene-dekker (approved):
lgtm

PR #3007, opened 2023-11-21
[RS-1431] Create webhooks-secret and allow UI admin user to patch it

gantony:
## Description

This adds and empty secret (`webhooks-secret`) in the `tigera-intrusion-detection` namespace to store user's Jira credentials for webhooks that interact with Jira.

The webhook definitions (and secrets) live in the managed clusters. When creating webhooks with `kubectl`, users can store various configs in the webhook itself, or reference config maps and secrets for more sensitive information. 

In our UI, we allow admin users to easily create/update/delete webhooks. When creating a Jira webhook, we store the Jira credentials they provide (email + token) into a secret. When we delete a webhook through the UI, it also cleans up the information in the secret.

In the initial [poc](https://github.com/tigera/ui-modules/pull/220), we allowed admin users to view and edit those credentials through the UI, which would require adding `create/view/update/delete` permissions for `secrets` in `tigera-netwok-admin` cluster role. This felt like too much permissions (even for admin users) so we decided to:

 - Only specify Jira credentials when creating a webhook, no way to view them ("UI" does not read secrets). If creds need to change, delete webhook and create a new one.
 - Have 1 secret that contains Jira credentials for all the webhooks of a managedcluster created through the UI, with key in secret data using format `<webhok-name>.username` 
 - Let the UI `patch` the required secret when secret when creating/adding a jira webhook (update/remove secret key accordingly).
 - Only allow admin users to `patch` `webhooks-secret` secret using the `ResourceNames` parameter. 
 - Annotate the secret with new annotation so that the operator ignores the secret's data, but re-creates an empty secret if it does not exists.

This PR implements the required changes for this in the operator.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/utils/component.go
+++ b/pkg/controller/utils/component.go
@@ -157,6 +157,10 @@ func (c componentHandler) createOrUpdateObject(ctx context.Context, obj client.O
 			}
 			return nil
 		case *v1.Secret:
+			if IgnoreSecretData(cur) {
+				logCtx.Info("Ignoring secret data of annotated object")
+				return nil
+			}
 			objSecret := obj.(*v1.Secret)
 			curSecret := cur.(*v1.Secret)
 			// Secret types are immutable, we need to delete the old version if the type has changed. If the
--- a/pkg/controller/utils/utils.go
+++ b/pkg/controller/utils/utils.go
@@ -53,6 +53,7 @@ import (
 	"github.com/tigera/operator/pkg/common"
 	"github.com/tigera/operator/pkg/controller/k8sapi"
 	"github.com/tigera/operator/pkg/render"
+	"github.com/tigera/operator/pkg/render/common/secret"
 )
 
 const (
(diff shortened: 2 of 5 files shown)


Brian-McM:
This doesn't seem like something the operator should be responsible for:
1. You're explicitly trying to ignore reconciliation and the purpose of the operator is to reconcile the items it creates
2. This is not installation configuration, this is post installation configuration for each webhook and it's not even mounted on any component (I'm guessing intrusion detection reads this from k8s directly).

It seems like your trying to use the operator to over come some technical difficulties. If you are going to go with having a single secret to store multiple sets of credentials, then you can just limit the UI to being able to create / delete that specific secret in the namespace (you can restrict RBAC to an exact resource, not just the type).

If that's not enough, I'd probably advocate for having a dedicated "backend" component for the UI to communicate with over adding it to the operator (but I'm sure there's simpler solutions).

gantony:
> This doesn't really seem like something the operator should be responsible for:
> 
> 1. You're explicitly trying to ignore reconciliation and the purpose of the operator is to reconcile the items it creates
> 2. This is not installation configuration, this is post installation configuration for each webhook and it's not even mounted on any component (I'm guessing intrusion detection reads this from k8s directly).
> 
> It seems like your trying to use the operator to over come some technical difficulties. If you are going to go with having a single secret to store multiple sets of credentials, then you can just limit the UI to being able to create / delete that specific secret in the namespace (you can restrict RBAC to an exact resource, not just the type).

Thanks for the feedback! Got similar feedback in https://tigera.slack.com/archives/CN5UQVAQG/p1700610656326959?thread_ts=1700566565.072459&cid=CN5UQVAQG and I'll now just handle the RBAC in this PR and let the UI create the secret if needed. Much simpler :+1:.

I approached this task in the mindset of "just create the secret as part of the install script" but there is no "install script" and things got a bit too complicated :sweat_smile:.

Brian-McM:
> I approached this task in the mindset of "just create the secret as part of the install script" but there is no "install script" and things got a bit too complicated 😅.

Ya that's understandable, the operator's an easy place to put a lot of different things. It can be tough figuring out if it's actually the appropriate place to put these things though (even among the maintainers lol).

Brian-McM (approved):

gantony:
Will be replaced by #3014

PR #3010, opened 2023-11-21
Ev 4104 csr controller

rene-dekker:
There will be a hard-coded finite list in the operator of service accounts and the secrets that they are allowed to make. If a CSR is issued by one of these service accounts, we will validate the contents. All other CSRs with signerName tigera.io/operator-signer will be rejected.

For now we only added the prometheus server TLS secret. But once done, all of our TLS secrets will be added this way.

The validation criteria:
- Verify that the service account is allowed to request the common name and/or SANs.
- Verify that the issuer of the CSR (the pod) indeed is the pod that belongs to the IP in the CSR. 
- Verify that the requested key ext usages are valid
- Verify that the CSR was not previously denied or failed
- Verify that the public key matches the signature on the CSR for the provider alg

Other:
- Serial number will be a random int64.
- The validity is currently fixed (825d from current time onwards).
- No info other than CN will be added to the subject of the issued cert (we don’t need it)
- Version field will be set to 3.

### DIFF
--- a/pkg/controller/certificatemanager/certificatemanager.go
+++ b/pkg/controller/certificatemanager/certificatemanager.go
@@ -43,6 +43,10 @@ import (
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 )
 
+// OperatorCSRSignerName when this value is set as a signer on a CSR, the CSR controller will handle
+// the request.
+const OperatorCSRSignerName = "tigera.io/operator-signer"
+
 var log = logf.Log.WithName("tls")
 
 func ErrInvalidCertDNSNames(secretName, secretNamespace string) error {
--- a/pkg/controller/certificatemanager/certificatemanager.go
+++ b/pkg/controller/certificatemanager/certificatemanager.go
@@ -17,7 +17,9 @@ package certificatemanager
 import (
 	"bytes"
 	"context"
+	"crypto/rand"
 	"crypto/x509"
+	"encoding/pem"
 	"fmt"
 	"strings"
 	"time"
@@ -100,6 +102,9 @@ type CertificateManager interface {
 	LoadTrustedBundle(context.Context, client.Client, string) (certificatemanagement.TrustedBundleRO, error)
 	// LoadMultiTenantTrustedBundleWithRootCertificates loads an existing trusted bundle with system root certificates to pass to render.
 	LoadMultiTenantTrustedBundleWithRootCertificates(context.Context, client.Client, string) (certificatemanagement.TrustedBundleRO, error)
+	// SignCertificate signs a certificate using its own private key. The currently supported public key types are
+	// *rsa.PublicKey, *ecdsa.PublicKey and ed25519.PublicKey. (See x509 package)
+	SignCertificate(certificate *x509.Certificate, publicKey any) ([]byte, error)
 }
 
 type Option func(cm *certificateManager) error
@@ -131,6 +136,7 @@ func Create(cli client.Client, installation *operatorv1.InstallationSpec, cluste
 	var (
 		cryptoCA                      *crypto.CA
 		csrImage                      string
+		privateKey                    any
 		privateKeyPEM, certificatePEM []byte
 		certificateManagement         *operatorv1.CertificateManagement
 		err                           error
@@ -202,11 +208,22 @@ func Create(cli client.Client, installation *operatorv1.InstallationSpec, cluste
 			if err := cryptoCA.Config.WriteCertConfig(crtContent, keyContent); err != nil {
 				return nil, err
 			}
-			privateKeyPEM, certificatePEM = keyContent.Bytes(), crtContent.Bytes()
+			privateKey, privateKeyPEM, certificatePEM = cryptoCA.Config.Key, keyContent.Bytes(), crtContent.Bytes()
 		} else {
 			// Found an existing CA - use that.
 			cm.log.V(2).Info("Found an existing CA secret")
 			privateKeyPEM, certificatePEM = caSecret.Data[corev1.TLSPrivateKeyKey], caSecret.Data[corev1.TLSCertKey]
+			privateKeyDER, _ := pem.Decode(privateKeyPEM)
+			if privateKeyDER == nil {
+				return nil, fmt.Errorf("cannot parse private tls.key PEM from the CA bundle")
+			}
+			if privateKey, err = x509.ParsePKCS1PrivateKey(privateKeyDER.Bytes); err != nil {
+				if privateKey, err = x509.ParsePKCS8PrivateKey(privateKeyDER.Bytes); err != nil {
+					if privateKey, err = x509.ParseECPrivateKey(privateKeyDER.Bytes); err != nil {
+						return nil, fmt.Errorf("cannot parse private key from the CA bundle")
+					}
+				}
+			}
 			cryptoCA, err = crypto.GetCAFromBytes(certificatePEM, privateKeyPEM)
 			if err != nil {
 				return nil, err
@@ -227,6 +244,7 @@ func Create(cli client.Client, installation *operatorv1.InstallationSpec, cluste
 	cm.keyPair = &certificatemanagement.KeyPair{
 		Name:                  caSecretName,
 		Namespace:             ns,
+		PrivateKey:            privateKey,
 		PrivateKeyPEM:         privateKeyPEM,
 		CertificatePEM:        certificatePEM,
 		CSRImage:              csrImage,
@@ -272,7 +290,7 @@ func (cm *certificateManager) GetOrCreateKeyPair(cli client.Client, secretName, 
 	}
 
 	// If we reach here, it means we need to create a new KeyPair.
-	tlsCfg, err := cm.MakeServerCertForDuration(sets.NewString(dnsNames...), rmeta.DefaultCertificateDuration, tls.SetServerAuth, tls.SetClientAuth)
+	tlsCfg, err := cm.MakeServerCertForDuration(sets.NewString(dnsNames...), tls.DefaultCertificateDuration, tls.SetServerAuth, tls.SetClientAuth)
 	if err != nil {
 		return nil, fmt.Errorf("unable to create signed cert pair: %s", err)
 	}
@@ -289,6 +307,22 @@ func (cm *certificateManager) GetOrCreateKeyPair(cli client.Client, secretName, 
 		CertificatePEM: crtContent.Bytes(),
 		DNSNames:       dnsNames,
 	}, nil
+}
+
+// SignCertificate signs a certificate using its own private key. The currently supported public key types are
+// *rsa.PublicKey, *ecdsa.PublicKey and ed25519.PublicKey. (See x509 package)
+func (cm *certificateManager) SignCertificate(certificateTemplate *x509.Certificate, publicKey any) ([]byte, error) {
+	derBytes, err := x509.CreateCertificate(rand.Reader, certificateTemplate, cm.Certificate, publicKey, cm.keyPair.PrivateKey)
+	if err != nil {
+		return nil, err
+	}
+	pemBytes := bytes.NewBuffer([]byte{})
+	err = pem.Encode(pemBytes, &pem.Block{Type: "CERTIFICATE", Bytes: derBytes})
+	if err != nil {
+		log.Error(err, "error encoding certificate PEM")
+		return nil, err
+	}
+	return pemBytes.Bytes(), nil
 }
 
 // This type will be returned for errors that do not have the correct Ext Key usage types
@@ -467,6 +501,7 @@ func certificateManagementKeyPair(ca *certificateManager, secretName, ns string,
 		DNSNames:              dnsNames,
 		CSRImage:              ca.keyPair.CSRImage,
 		Namespace:             ns,
+		CertificatePEM:        ca.CertificateManagement().CACert,
 	}
 }
 
--- a/pkg/controller/certificatemanager/certificatemanager.go
+++ b/pkg/controller/certificatemanager/certificatemanager.go
@@ -102,9 +102,9 @@ type CertificateManager interface {
 	LoadTrustedBundle(context.Context, client.Client, string) (certificatemanagement.TrustedBundleRO, error)
 	// LoadMultiTenantTrustedBundleWithRootCertificates loads an existing trusted bundle with system root certificates to pass to render.
 	LoadMultiTenantTrustedBundleWithRootCertificates(context.Context, client.Client, string) (certificatemanagement.TrustedBundleRO, error)
-	// SignCertificate signs a certificate using its own private key. The currently supported public key types are
-	// *rsa.PublicKey, *ecdsa.PublicKey and ed25519.PublicKey. (See x509 package)
-	SignCertificate(certificate *x509.Certificate, publicKey any) ([]byte, error)
+	// SignCertificate signs a certificate using the certificate manager's private key. The function is assuming that the
+	// public key of the requestor is already set in the certificate template.
+	SignCertificate(certificate *x509.Certificate) ([]byte, error)
 }
 
 type Option func(cm *certificateManager) error
@@ -134,8 +134,9 @@ func WithTenant(t *operatorv1.Tenant) Option {
 // brings their own secrets, CertificateManager will preserve and return them.
 func Create(cli client.Client, installation *operatorv1.InstallationSpec, clusterDomain, ns string, opts ...Option) (CertificateManager, error) {
 	var (
-		cryptoCA                      *crypto.CA
-		csrImage                      string
+		cryptoCA *crypto.CA
+		csrImage string
+		// The private key is of type any, as this is the interface used in the x509 package for all private key types.
 		privateKey                    any
 		privateKeyPEM, certificatePEM []byte
 		certificateManagement         *operatorv1.CertificateManagement
@@ -217,6 +218,8 @@ func Create(cli client.Client, installation *operatorv1.InstallationSpec, cluste
 			if privateKeyDER == nil {
 				return nil, fmt.Errorf("cannot parse private tls.key PEM from the CA bundle")
 			}
+			// Parse in order of likelihood of format. If the tigera-ca-private secret is not replaced with a custom one,
+			// the certificate is PKCS1 formatted. (The x509 package also uses parsing as the way to identifying the type.)
 			if privateKey, err = x509.ParsePKCS1PrivateKey(privateKeyDER.Bytes); err != nil {
 				if privateKey, err = x509.ParsePKCS8PrivateKey(privateKeyDER.Bytes); err != nil {
 					if privateKey, err = x509.ParseECPrivateKey(privateKeyDER.Bytes); err != nil {
@@ -309,10 +312,10 @@ func (cm *certificateManager) GetOrCreateKeyPair(cli client.Client, secretName, 
 	}, nil
 }
 
-// SignCertificate signs a certificate using its own private key. The currently supported public key types are
-// *rsa.PublicKey, *ecdsa.PublicKey and ed25519.PublicKey. (See x509 package)
-func (cm *certificateManager) SignCertificate(certificateTemplate *x509.Certificate, publicKey any) ([]byte, error) {
-	derBytes, err := x509.CreateCertificate(rand.Reader, certificateTemplate, cm.Certificate, publicKey, cm.keyPair.PrivateKey)
+// SignCertificate signs a certificate using the certificate manager's private key. The function is assuming that the
+// public key of the requestor is already set in the certificate template.
+func (cm *certificateManager) SignCertificate(certificateTemplate *x509.Certificate) ([]byte, error) {
+	derBytes, err := x509.CreateCertificate(rand.Reader, certificateTemplate, cm.Certificate, certificateTemplate.PublicKey, cm.keyPair.PrivateKey)
 	if err != nil {
 		return nil, err
 	}
@@ -325,7 +328,7 @@ func (cm *certificateManager) SignCertificate(certificateTemplate *x509.Certific
 	return pemBytes.Bytes(), nil
 }
 
-// This type will be returned for errors that do not have the correct Ext Key usage types
+// CertExtKeyUsageError This type will be returned for errors that do not have the correct Ext Key usage types
 // for a specific secert certificate.
 type CertExtKeyUsageError struct {
 	msg string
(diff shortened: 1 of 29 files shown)


pkg/controller/csr/csr_controller.go
```diff
@@ -0,0 +1,223 @@
...
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+var log = logf.Log.WithName("controller_csr")
+var controllerName = "csr-controller"
+
+// LabelName label that we set on our CSRs, so it helps us narrow down our list somewhat.
+const LabelName = "k8s-app"
```
caseydavenport, line 41: Maybe this should be something less generic? like `operator.tigera.io/csr`? 

Does this need to be public?

pkg/controller/csr/csr_controller.go
```diff
@@ -0,0 +1,223 @@
...
+
+	instance := &operatorv1.Installation{}
+	if err := r.client.Get(ctx, utils.DefaultInstanceKey, instance); err != nil {
+		if apierrors.IsNotFound(err) {
+			return reconcile.Result{}, nil
+		}
+		return reconcile.Result{}, err
+	}
+
+	certificateManager, err := certificatemanager.Create(r.client, &instance.Spec, r.clusterDomain, common.OperatorNamespace(), certificatemanager.WithLogger(reqLogger))
```
caseydavenport, line 119: This will likely need to be tenant aware, and use the correct CA to sign requests for each tenant.

pkg/controller/csr/csr_controller.go
```diff
@@ -0,0 +1,223 @@
...
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	for _, csr := range csrList.Items {
+		if csr.Spec.SignerName == utils.OperatorCSRSignerName && csr.Status.Certificate == nil {
```
caseydavenport, line 125: This would probably be better phrased as:

```go
if sigerName != operatorSigner || csr.Status.Certificate != nil {
  // Not for us, or already signed. 
  continue
}
```

To simplify indentation.

pkg/controller/csr/csr_controller.go
```diff
@@ -0,0 +1,223 @@
...
+			if err != nil {
+				csr.Status.Conditions = []certificatesv1.CertificateSigningRequestCondition{
+					{
+						Type:    certificatesv1.CertificateDenied,
+						Message: err.Error(),
+						Reason:  err.Error(),
+						Status:  corev1.ConditionTrue,
+					},
+				}
+				log.Error(err, "Rejecting the CSR.")
```
caseydavenport, line 137: reqLogger?

pkg/controller/csr/csr_controller.go
```diff
@@ -0,0 +1,223 @@
...
+				_, err = r.certificatesClient.CertificateSigningRequests().UpdateStatus(ctx, &csr, metav1.UpdateOptions{})
+				if err != nil {
+					return reconcile.Result{}, err
+				}
+				break
```
caseydavenport, line 142: Should this be a `continue` instead of `break`?

pkg/controller/csr/csr_controller.go
```diff
@@ -0,0 +1,223 @@
...
+		return nil, err
+	}
+
+	if !certificatemanagement.TLSAssets.ValidRequestor(csr.Name, csr.Spec.Username) {
+		return nil, fmt.Errorf("invalid requestor %s for CSR with name %s", csr.Spec.Username, csr.Name)
+	}
+	if !certificatemanagement.TLSAssets.ValidDNSNames(csr.Name, cr) {
+		return nil, fmt.Errorf("invalid dns names found in CSR with name %s", csr.Name)
+	}
+	// todo: validate the ip address by using the podname from the csr and make a GET to the k8s pods api.
```
caseydavenport, line 206: Do we already have permissions to list all pods? Do we really need that?

pkg/tls/certificatemanagement/assets.go
```diff
@@ -0,0 +1,90 @@
...
+}
+
+// TLSAssetMap can look up all assets associated with a serviceaccount and determine if a CSR is requesting allowed
+// certificate contents.
+type TLSAssetMap map[string]TLSAsset
+
+var (
+	// TLSAssets Stores the secret name as key and asset as value.
+	TLSAssets           = make(TLSAssetMap)
+	PrometheusServerTLS TLSAsset
```
rene-dekker, line 46: The idea is that all of our TLS assets are defined here and initiated at startup.

Then the signature of `certificateManager.GetOrCreateKeyPair(...)` can be modified to get simply the TLSAsset as input.

pkg/controller/utils/utils.go
```diff
@@ -174,6 +175,41 @@ func AddSecretWatchWithLabel(c controller.Controller, ns, label string) error {
...
+			}
+			csr, ok := e.ObjectNew.(*certificatesv1.CertificateSigningRequest)
+			if !ok {
+				return false
+			}
+			if csr.Spec.SignerName != OperatorCSRSignerName {
+				return false
+			}
+			for _, condition := range csr.Status.Conditions {
+				if (condition.Type == certificatesv1.CertificateDenied || condition.Type == certificatesv1.CertificateFailed) && condition.Status == corev1.ConditionTrue {
```
caseydavenport, line 198: It's a bit complicated that we have this check in the watch as well as in the CSR controller itself.

Can we commonize this logic into something like `func DoWeCareAboutThisCSR(obj) bool` that we can use in both places? That would make it much harder for these to get out of sync.

pkg/controller/utils/utils.go
```diff
@@ -174,6 +175,41 @@ func AddSecretWatchWithLabel(c controller.Controller, ns, label string) error {
...
+					// These request statuses are deterministic/non-recoverable.
+					return false
+				}
+			}
+			// We are only interested in CSRs that have not been signed. This also excludes
+			// us from watching our own updates and reconciling for no reason.
+			return csr.Status.Certificate == nil
+		},
+		DeleteFunc: func(e event.DeleteEvent) bool {
+			return false // We are not interested in these events.
```
caseydavenport, line 208: For future readers, it might be worth elaborating on why we're not interested. From experience, best not to assume the reader knows why. In other words, "We are not interested in these events" is self-evident by the fact that we return "false" here unconditionally. The interesting thing is "why?"

pkg/tls/certificatemanagement/assets.go
```diff
@@ -0,0 +1,90 @@
...
+		}
+		if request.Subject.CommonName != "" && !permittedNames[request.Subject.CommonName] {
+			return false
+		}
+		return true
+	}
+	return false
+}
+
+// InstantiateTLSAssets Called when the operator initializes.
```
caseydavenport, line 80: The fact this is called when the operator starts is obvious from the code, so this comment doesn't add anything. Why is it called then? What does this function actually do and when should I modify it?

pkg/tls/certificatemanagement/assets.go
```diff
@@ -0,0 +1,90 @@
...
+
+var (
+	// TLSAssets Stores the secret name as key and asset as value.
+	TLSAssets           = make(TLSAssetMap)
+	PrometheusServerTLS TLSAsset
+)
+
+// ValidRequestor returns true if the TLS may be created by the service account by means of a CSR.
+func (t TLSAssetMap) ValidRequestor(csrName, requestor string) bool {
+	secretName := strings.Split(csrName, ":")[0] // The CSRName is formatted as "<secretName>:<podName>".
```
caseydavenport, line 51: I usually ask folks to put comments on top of the line they are referring to instead of next to - I find it improves readability  and increases consistency.

pkg/tls/certificatemanagement/assets.go
```diff
@@ -0,0 +1,90 @@
...
+	return false
+}
+
+// InstantiateTLSAssets Called when the operator initializes.
+func InstantiateTLSAssets(clusterDomain string) {
+	PrometheusServerTLS = TLSAsset{
+		PrometheusServerTLSSecretName,
+		"tigera-prometheus",
+		"tigera-prometheus",
+		dns.GetServiceDNSNames("prometheus-http-api", common.TigeraPrometheusNamespace, clusterDomain),
```
caseydavenport, line 86: Seems like we're jumping through an extra hoop here to dynamically build some global variables that should just be `const` in order to incorporate the `clusterDomain` at startup.

Wouldn't it be nicer if we declared this as constants, removed the need for `Initialize`, and then just passed the clusterDomain as an argument to the `DNSNames` function when calling it? 

It would be simple to just do this all as part of `init()` on program load.

pkg/tls/tls.go
```diff
@@ -24,6 +24,8 @@ import (
 	"github.com/openshift/library-go/pkg/crypto"
 )
 
+const DefaultCertificateDuration = 825 * 24 * time.Hour
```
caseydavenport, line 27: is this true for all secrets? Or just those signed by our CSR controller?

caseydavenport (commented):
This is very cool! I gave it a first skim with initial notes on it. Haven't done a deep dive yet though

pkg/controller/csr/csr_controller.go
rene-dekker: I think we could probably do it, or perhaps just remove the whole label name check altoghether.

If we wanna change it, we have to also make a change in https://github.com/tigera/key-cert-provisioner/blob/3abd79340d97354aa6b24789b302664795c324e0/pkg/k8s/certificate.go#L104

pkg/tls/tls.go
rene-dekker: Currently, this is true of all TLS secrets we create, other than the CA itself.

pkg/controller/csr/csr_controller.go
rene-dekker: Added to a PR for key-cert-provisioner: https://github.com/tigera/key-cert-provisioner/pull/89

pkg/controller/csr/csr_controller.go
rene-dekker: Can we postpone multi-tenancy for a later iteration? (After we are done with the prometheus work.)

pkg/controller/csr/csr_controller.go
rene-dekker: The operator currently has these permissions.

If we do not validate the IP, then technically an attacker that obtains one of our service account tokens can issue a cert for any IPSAN it wants. I don't think there is anything other than prometheus that would be making IP based requests, so the attack has very limited potential.

pkg/controller/csr/csr_controller.go
```diff
@@ -0,0 +1,321 @@
...
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+		reqLogger.Info("Signed CSR with name : %v.", csr.Name)
+	}
+
+	return reconcile.Result{}, nil
+}
+
+type systemError struct {
```
caseydavenport, line 221: This is a fairly generic name for an error type - think we need a big comment explaining what this is for and/or a rename to make it more clear when code authors should consider using this vs. a generic error. 

My understanding is that this is meant to allow callers of `validate` to determine if there was a validation error or a generic error when talking to the apiserver.

I'd suggest either:
- Not performing any API calls in the `validate` function, removing the need for this altogether. That seems the nicest conceptually, since "validate" should really just be validating a resource, not generating API calls.
- Switching this to "InvalidCSRError" which is pretty tightly scoped conceptually in comparison.

There's also an option to use the existing APIStatus error type: https://github.com/kubernetes/apimachinery/blob/master/pkg/api/errors/errors.go#L39-L43 

But I can't say for sure that covers exactly what this does...

pkg/controller/csr/csr_controller.go
```diff
@@ -0,0 +1,321 @@
...
+		podList := &corev1.PodList{}
+		err = r.client.List(ctx, podList, &client.ListOptions{
+			LabelSelector: labels.SelectorFromSet(map[string]string{
+				render.AppLabelName: label,
+			}),
+			Namespace: asset.serviceaccountNamespace})
+		if err != nil {
+			return nil, systemError{fmt.Sprintf("unexpected server error while listing pods by label for CSR %s", csr.Name)}
+		}
+		var foundIp bool
```
caseydavenport, line 289: should be `foundIP`

pkg/controller/csr/csr_controller.go
```diff
@@ -0,0 +1,321 @@
...
+		label, ok := csr.Labels[render.AppLabelName]
+		if !ok {
+			return nil, fmt.Errorf("unable to find expected k8s-app label for CSR with name %s", csr.Name)
+		}
+		podList := &corev1.PodList{}
+		err = r.client.List(ctx, podList, &client.ListOptions{
+			LabelSelector: labels.SelectorFromSet(map[string]string{
+				render.AppLabelName: label,
+			}),
+			Namespace: asset.serviceaccountNamespace})
```
caseydavenport, line 285: if we're going to spread this across multiple lines, we should do so for the ending brackets as well:

```go
  Namespace: asset.serviceaccountNamespace,
 })
```

Or just put it all on one line because it doesn't actually seem _that_ long.

pkg/controller/certificatemanager/certificatemanager.go
```diff
@@ -100,6 +102,9 @@ type CertificateManager interface {
 	LoadTrustedBundle(context.Context, client.Client, string) (certificatemanagement.TrustedBundleRO, error)
 	// LoadMultiTenantTrustedBundleWithRootCertificates loads an existing trusted bundle with system root certificates to pass to render.
 	LoadMultiTenantTrustedBundleWithRootCertificates(context.Context, client.Client, string) (certificatemanagement.TrustedBundleRO, error)
+	// SignCertificate signs a certificate using its own private key. The currently supported public key types are
+	// *rsa.PublicKey, *ecdsa.PublicKey and ed25519.PublicKey. (See x509 package)
+	SignCertificate(certificate *x509.Certificate, publicKey any) ([]byte, error)
```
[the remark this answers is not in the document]
rene-dekker: TIL that interface methods cannot have parameterized types. It would not compile. I actually found a way to remove the public key as a param altogether, so I will solve the conversation that way.

pkg/controller/certificatemanager/certificatemanager.go
```diff
@@ -131,6 +136,7 @@ func Create(cli client.Client, installation *operatorv1.InstallationSpec, cluste
 	var (
 		cryptoCA                      *crypto.CA
 		csrImage                      string
+		privateKey                    any
```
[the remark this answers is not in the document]
rene-dekker: The use of `any` comes from how it is defined in golang's x509 package. It predates the generics, my guess is that otherwise the implementation would have been different.
The x509 functions and methods need `any` to be passed in. At the time of writing, the three supported private key types are rsa, ecdsa, ed25519. We could implement a new interface containing these constraints, but it would introduce more code + casting / parsing, without any practical implications that I can think of. If someone goes out of his/her way to replace the tigera-ca-private secret with a broken private key, we can still leverage the same code in the x509 package to return the right error.

I'll add a comment.

pkg/controller/csr/csr_controller.go
rene-dekker: I opted for `invalidCertificateRequestError`. It is indeed used to differentiate between errors that lead to denying a CSR vs errors that should end the reconcile loop, due to an unexpected issue.

Motivated by:
- Fetching the pods also requires validation.
- Preferably we first do the "cheap checks" that require no api calls. 
- Keeping all validation related logic contained in this function allows us to unit test all of it comprehensively, while keeping the code relatively readable.

pkg/controller/csr/csr_controller.go
```diff
@@ -0,0 +1,321 @@
...
+		}
+	} else if len(certificateRequest.IPAddresses) > 1 {
+		return nil, fmt.Errorf("unable request more than 1 IP for CSR with name %s", csr.Name)
+	}
+
+	bigint, _ := rand.Int(rand.Reader, big.NewInt(math.MaxInt64))
+	certTemplate := &x509.Certificate{
+		// We don't rely on any other part of the subject. Common name is validated already.
+		Subject:            pkix.Name{CommonName: certificateRequest.Subject.CommonName},
+		SerialNumber:       bigint,
```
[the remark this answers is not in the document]
rene-dekker: I think we could even just hard-code and set it to "1" in our current setup. I considered that and consulted with our security team.
> "What happens if we do somehow spit out the same number twice?"
The primary use case of the serial number would be if certificates are being tracked in a database. The uniqueness, and validity of the certificate is determined by the signature.

pkg/controller/csr/csr_controller.go
```diff
@@ -0,0 +1,321 @@
...
+		for _, pod := range podList.Items {
+			if pod.Status.PodIP == certificateRequest.IPAddresses[0].String() {
+				foundIp = true
+				break
+			}
+		}
+		if !foundIp {
+			return nil, fmt.Errorf("invalid pod IP for CSR with name %s", csr.Name)
+		}
+	} else if len(certificateRequest.IPAddresses) > 1 {
```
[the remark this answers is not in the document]
rene-dekker: If there are no IP addresses, we issue a cert without IP SAN on them. This is a valid use-case, but with the way we configure our init container it will currently never happen.

pkg/controller/csr/csr_controller.go
```diff
@@ -0,0 +1,344 @@
...
+		rmonitor.PrometheusServerTLSSecretName: {
+			serviceaccountName:      rmonitor.PrometheusServiceAccountName,
+			serviceaccountNamespace: rmonitor.TigeraPrometheusObjectName,
+			validDNSNames:           monitor.PrometheusTLSServerDNSNames(clusterDomain),
+		},
+	}
+}
+
+// blank assignment to verify that ReconcileCompliance implements reconcile.Reconciler
+var _ reconcile.Reconciler = &ReconcileCSR{}
```
caseydavenport, line 119: Here you're asserting that a pointer to this structure implements the interface, but the bound functions are all defined against the struct and not a pointer to the struct. I don't think it matters a ton, but we usually default to pointer receivers unless there's a specific reason not to so probably best to do that here as well.

`func (r ReconcileCSR)` -> `func (r *ReconcileCSR)`

pkg/controller/csr/csr_controller.go
```diff
@@ -0,0 +1,344 @@
...
+
+// ReconcileCSR Components created by the operator may submit certificate signing requests against k8s under certain
+// conditions for signer name "tigera.io/operator-signer". This is the controller that monitors, approves and signs
+// these CSRs. It will only sign requests that are pre-defined and reject others in order to avoid malicious requests.
+type ReconcileCSR struct {
```
caseydavenport, line 124: Is there are reason this needs to be a public struct?

Seems to me like this is only referenced from the `csr` package so should be private.

pkg/controller/csr/csr_controller_test.go
```diff
@@ -0,0 +1,330 @@
...
+			Expect(csr.Status.Certificate).To(BeEmpty())
+		})
+	})
+
+	table.DescribeTable("csr validation", func(csrfunc func() *certificatesv1.CertificateSigningRequest, expectError bool) {
+		csr := csrfunc()
+		certificate, err := r.validate(csr, &corev1.Pod{ObjectMeta: metav1.ObjectMeta{Name: "prometheus-calico-node-prometheus-0"}, Status: corev1.PodStatus{PodIP: "1.2.3.4"}})
+		if expectError {
+			Expect(err).To(HaveOccurred())
+		} else if relevantCSR(csr) {
```
caseydavenport, line 162: I think this test has potential to silently pass if we don't craft the inputs correctly - i.e., if they somehow return `false` from relevantCSR, it will look like `validate` succeeded but actually we won't be executing any of the assertions.

I think we should split out `relevantCSR` into its own set of tests and remove the check from here, and every input to this function should be simply testing the validate() function (and not other units). That will keep the scope of the test focused and less prone to accidental success / failure.

pkg/controller/csr/csr_controller_test.go
```diff
@@ -0,0 +1,330 @@
...
+			Expect(err).ShouldNot(HaveOccurred())
+			Expect(r.client.Get(ctx, client.ObjectKey{Name: csr.Name}, csr)).NotTo(HaveOccurred())
+			Expect(csr.Status.Conditions).To(HaveLen(1))
+			Expect(csr.Status.Conditions[0].Type).To(Equal(certificatesv1.CertificateDenied))
+			Expect(csr.Status.Conditions[0].Status).To(Equal(corev1.ConditionTrue))
+			Expect(csr.Status.Certificate).To(BeEmpty())
+		})
+	})
+
+	table.DescribeTable("csr validation", func(csrfunc func() *certificatesv1.CertificateSigningRequest, expectError bool) {
```
caseydavenport, line 157: I'm trying to follow this test and not sure why it takes a function that generates a CSR and not just a CSR object as input. Feels like an unnecessary layer here?

pkg/controller/csr/csr_controller_test.go
```diff
@@ -0,0 +1,330 @@
...
+		csr := csrfunc()
+		certificate, err := r.validate(csr, &corev1.Pod{ObjectMeta: metav1.ObjectMeta{Name: "prometheus-calico-node-prometheus-0"}, Status: corev1.PodStatus{PodIP: "1.2.3.4"}})
```
caseydavenport, line 159: Should we allow the testcases to pass in the pod so that we can verify different / nil pods being passed?

pkg/controller/csr/csr_controller_test.go
```diff
@@ -0,0 +1,330 @@
...
+			Expect(err).To(HaveOccurred())
+		} else if relevantCSR(csr) {
+			Expect(err).ToNot(HaveOccurred())
+			Expect(certificate.ExtKeyUsage).To(Equal(extKeyUsage))
+			Expect(certificate.DNSNames).To(Equal(monitor.PrometheusTLSServerDNSNames(dns.DefaultClusterDomain)))
+			Expect(certificate.Subject.CommonName).To(Equal(monitor.PrometheusTLSServerDNSNames(dns.DefaultClusterDomain)[0]))
+			Expect(certificate.IPAddresses).To(Equal([]net.IP{net.ParseIP("1.2.3.4").To4()}))
+			Expect(certificate.IsCA).To(BeFalse())
+		}
+	},
```
caseydavenport, line 170: I think we're missing entries to verify various different `Extra` cases (multiple pods, no pods, etc) as well as SignerName

caseydavenport (approved):
@rene-dekker A few minor comments - mostly nits, a few missing test cases, and some test code that might be more complicated than it needs to be (?) 

Otherwise don't feel I need to review again.

PR #3012, opened 2023-11-22
remove cniLogging when cni is not calico

ti-afra:
## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/installation/core_controller.go
+++ b/pkg/controller/installation/core_controller.go
@@ -727,6 +727,8 @@ func fillDefaults(instance *operator.Installation) error {
 			instance.Spec.Logging.CNI.LogFileMaxSize = new(resource.Quantity)
 			*instance.Spec.Logging.CNI.LogFileMaxSize = resource.MustParse("100Mi")
 		}
+	} else {
+		instance.Spec.Logging = nil
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/controller/installation/core_controller.go
```diff
@@ -727,6 +727,8 @@ func fillDefaults(instance *operator.Installation) error {
 			instance.Spec.Logging.CNI.LogFileMaxSize = new(resource.Quantity)
 			*instance.Spec.Logging.CNI.LogFileMaxSize = resource.MustParse("100Mi")
 		}
+	} else {
```
caseydavenport, line 730: I don't think we can do this - we can't just default away the configuration that a user might have set.

One of the core principles we follow is that we _never_ overwrite user input. Our code cannot be smart enough to know if this is the correct choice to make.

caseydavenport (commented):
I don't think we can add this code, and added a comment explaining why. What's the motivation?

ti-afra:
> I don't think we can add this code, and added a comment explaining why. What's the motivation?

we discussed this offline and I am convinced that we should not re-default cniLogging values when CNI Type is changing away from Calico and changes should explicitly be provided by the user.

PR #3018, opened 2023-11-27
Add Linseed RBAC,envvars for eks log forwarding

vara2504:
Related PR:
fluentd-docker: https://github.com/tigera/fluentd-docker/pull/402

For token generation:
Linseed:https://github.com/tigera/calico-private/pull/6946

RBAC, env variable to connect to linseed
https://github.com/tigera/operator/pull/3018

image
image

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


pkg/render/fluentd.go
```diff
@@ -94,7 +94,9 @@ const (
 	FluentdNodeName        = "fluentd-node"
 	fluentdNodeWindowsName = "fluentd-node-windows"
 
-	eksLogForwarderName = "eks-log-forwarder"
+	EKSNodeName                  = "eks-log-forwarder-node"
```
caseydavenport, line 97: I think we can just use `EKSLogForwarderName` here - the `NodeName` suffix is mostly used for daemonset components that run on every node (e.g., calico-node, fluentd-node, etc) so we don't need to copy that pattern here.

pkg/controller/logcollector/logcollector_controller.go
```diff
@@ -592,6 +600,14 @@ func (r *ReconcileLogCollector) Reconcile(ctx context.Context, request reconcile
 			},
 			TrustedBundle: trustedBundle,
 		}),
+		rcertificatemanagement.CertificateManagement(&rcertificatemanagement.Config{
```
caseydavenport, line 603: I think the standard pattern is to merge both of these CertificateManagement components into one that has both serviceaccounts and keypair options, instead of using two - otherwise they will both try to write the trusted bundle (which isn't necessarily a problem, but does add additional writes)

Worth a double-check with @rene-dekker though.

pkg/render/fluentd.go
```diff
@@ -170,6 +173,8 @@ type FluentdConfiguration struct {
 	UsePSP bool
 	// Whether to use User provided certificate or not.
 	UseSyslogCertificate bool
+
+	EKSLogForwarderKeyPair certificatemanagement.KeyPairInterface
```
caseydavenport, line 177: Worth a comment explaining what this is - i.e., "This is the keypair containing the certificate presented by EKS LogForwarder when communicating with Linseed"

pkg/render/fluentd_test.go
```diff
@@ -884,6 +886,13 @@ var _ = Describe("Tigera Secure Fluentd rendering tests", func() {
 			}))
 
 		Expect(envs).To(ContainElement(corev1.EnvVar{Name: "EKS_CLOUDWATCH_LOG_FETCH_INTERVAL", Value: "900"}))
+
+		Expect(envs).To(ContainElement(corev1.EnvVar{Name: "LINSEED_ENABLED", Value: "true"}))
+		Expect(envs).To(ContainElement(corev1.EnvVar{Name: "LINSEED_ENDPOINT", Value: "https://tigera-linseed.tigera-elasticsearch.svc"}))
+		Expect(envs).To(ContainElement(corev1.EnvVar{Name: "LINSEED_CA_PATH", Value: "/etc/pki/tls/certs/tigera-ca-bundle.crt"}))
+		Expect(envs).To(ContainElement(corev1.EnvVar{Name: "TLS_CRT_PATH", Value: "/tigera-eks-log-forwarder-tls/tls.crt"}))
+		Expect(envs).To(ContainElement(corev1.EnvVar{Name: "TLS_KEY_PATH", Value: "/tigera-eks-log-forwarder-tls/tls.key"}))
+		Expect(envs).To(ContainElement(corev1.EnvVar{Name: "LINSEED_TOKEN", Value: "/var/run/secrets/kubernetes.io/serviceaccount/token"}))
```
caseydavenport, line 895: I think it would be worth adding a multi-tenant unit-test case here as well that asserts on just the different env vars we expect there.

Also, a managed cluster case that asserts the Linseed volume mount is created would also be good to have.

pkg/controller/logcollector/logcollector_controller.go
```diff
@@ -389,6 +389,13 @@ func (r *ReconcileLogCollector) Reconcile(ctx context.Context, request reconcile
 		return reconcile.Result{}, err
 	}
 
+	// eksLogForwarderKeyPair is the key pair eks-log-forwarder presents to identify itself
+	eksLogForwarderKeyPair, err := certificateManager.GetOrCreateKeyPair(r.client, render.EKSLogForwarderTLSSecretName, common.OperatorNamespace(), []string{render.EKSLogForwarderTLSSecretName})
```
caseydavenport, line 393: I think we should only do this inside of the `	if instance.Spec.AdditionalSources.EksCloudwatchLog != nil {` branch below on line 545. 

No need to create this secret if we're not running on EKS, and EKS log forwarding isn't enabled.

caseydavenport (changes requested):
@vara2504 this looks good! Just a few minor comments so far.

pkg/controller/logcollector/logcollector_controller.go
caseydavenport: Yep, same would apply for the CertificateManagementComponent - we'd probably want to conditionally include the serviceaccount and KPO depending on whether EKS log forwarding is enabled.

vara2504:
@caseydavenport Ready for review

pkg/controller/monitor/monitor_controller.go
```diff
@@ -150,6 +150,7 @@ func add(mgr manager.Manager, c controller.Controller) error {
 		render.FluentdPrometheusTLSSecretName,
 		render.NodePrometheusTLSServerSecret,
 		kubecontrollers.KubeControllerPrometheusTLSSecret,
+		render.EKSLogForwarderTLSSecretName,
```
vara2504, line 153: Is it OK to watch for EKS secret even in non EKS environment? :thinking:

vara2504:
@caseydavenport Made additional changes to fix "tigera-eks-log-forwarder-tls secret not available yet, will add it if/when it becomes available" Info log in operator pod for non EKS environment,

https://github.com/tigera/operator/pull/3018/commits/2933dd3ea21ad9ccac8a6f699368f812ee500d9a

Please take a look.

pkg/render/fluentd.go
```diff
@@ -1025,6 +1030,13 @@ func (c *fluentdComponent) eksLogForwarderDeployment() *appsv1.Deployment {
 		eksCloudwatchLogCredentialHashAnnotation: rmeta.AnnotationHash(c.cfg.EKSConfig),
 	}
 
+	// Determine the namespace in which Linseed is running. For managed and standalone clusters, this is always the elasticsearch
+	// namespace. For multi-tenant management clusters, this may vary.
+	linseedNS := ElasticsearchNamespace
+	if c.cfg.Tenant != nil {
```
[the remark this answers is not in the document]
caseydavenport: `c.cfg.Tenant != nil` is also true for single-tenant management clusters in Calico Cloud, and for those we still want to use the ES namespace.

pkg/render/fluentd_test.go
caseydavenport: @vara2504  I think the multi-tenant case should be a distinct `It()` clause rather than embedding it in this test.

pkg/render/fluentd_test.go
```diff
@@ -884,6 +887,40 @@ var _ = Describe("Tigera Secure Fluentd rendering tests", func() {
...
+
+		// Adding test for Multi tenannt use case
```
caseydavenport, line 898: typo: "tenannt"

pkg/controller/logcollector/logcollector_controller.go
```diff
@@ -594,6 +603,23 @@ func (r *ReconcileLogCollector) Reconcile(ctx context.Context, request reconcile
 		}),
 	}
 
+	if installation.KubernetesProvider == operatorv1.ProviderEKS {
```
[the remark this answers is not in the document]
vara2504: Done

pkg/render/fluentd_test.go
```diff
@@ -917,104 +920,133 @@ var _ = Describe("Tigera Secure Fluentd rendering tests", func() {
...
 
-		fetchInterval := int32(900)
-		cfg.EKSConfig = &render.EksCloudwatchLogConfig{
-			AwsId:         []byte("aws-id"),
-			AwsKey:        []byte("aws-key"),
-			AwsRegion:     "us-west-1",
-			GroupName:     "dummy-eks-cluster-cloudwatch-log-group",
-			FetchInterval: fetchInterval,
-		}
+		// we need to insert the linseedResource at index 11 to assert the resources in order
```
caseydavenport, line 1004: This type of test is very very brittle, and is going to result in unnecessary failures when we modify the Linseed render code in the future for unrelated changes that happen to adjust resource ordering.

We should be using this helper function here instead: https://github.com/caseydavenport/operator/blob/3919c836a4518f6ff5569fcef9551f99930f3933/pkg/render/common/test/testing.go#L93-L105

It performs an order agnostic check and emits much better diagnostic output on failure.

pkg/render/fluentd_test.go
```diff
@@ -1051,3 +1083,106 @@ var _ = Describe("Tigera Secure Fluentd rendering tests", func() {
...
+		{name: "fluentd-node", ns: "tigera-fluentd", group: "", version: "v1", kind: "ServiceAccount"},
+		{name: render.PacketCaptureAPIRole, ns: render.LogCollectorNamespace, group: "rbac.authorization.k8s.io", version: "v1", kind: "Role"},
+		{name: render.PacketCaptureAPIRoleBinding, ns: render.LogCollectorNamespace, group: "rbac.authorization.k8s.io", version: "v1", kind: "RoleBinding"},
+		// Daemonset
+		{name: "fluentd-node", ns: "tigera-fluentd", group: "apps", version: "v1", kind: "DaemonSet"},
+	}
+}
+
+func assertExpectedResourcesForEKS(deploy *appsv1.Deployment, t corev1.Toleration, cfg *render.FluentdConfiguration) {
+	Expect(deploy.Spec.Template.Spec.InitContainers).To(HaveLen(1))
```
caseydavenport, line 1140: I'm not a fan of these types of helper functions - it makes it harder to debug the tests when they go wrong because line numbers now point here instead of the test itself.

You could use `ExpectWithOffset` instead, but then we lose the exact assertion that triggered it and instead only see the line in the calling test function.

I think we can leave this for now, but in the future I'd much rather we just have one or two mainline tests that verify all of these assertions, and then much smaller tests that assert on specific things.

There's no need for every multi-tenant env var test to check the SecurityContext, for example - we already have UTs that cover that. This is a good example: https://github.com/caseydavenport/operator/blob/3919c836a4518f6ff5569fcef9551f99930f3933/pkg/render/logstorage/linseed/linseed_test.go#L399-L423

^ That test only asserts the things that it is specifically looking for, keeping the test itself focused and simple.

vara2504: Done

pkg/render/fluentd_test.go
```diff
@@ -1051,3 +1086,44 @@ var _ = Describe("Tigera Secure Fluentd rendering tests", func() {
...
+	return &render.EksCloudwatchLogConfig{
+		AwsId:         []byte("aws-id"),
+		AwsKey:        []byte("aws-key"),
+		AwsRegion:     "us-west-1",
+		GroupName:     "dummy-eks-cluster-cloudwatch-log-group",
+		FetchInterval: fetchInterval,
+	}
+}
+
+func setupInstallationConfig(t corev1.Toleration) *operatorv1.InstallationSpec {
```
caseydavenport, line 1101: This is the type of function that I think we want to get rid of - it's wrapping one line of code.

There's really not much point in having a function that only wraps a single line of code - it just makes the calling code harder to read because it requires that you go read this function in order to understand it.

caseydavenport: It's much better to be verbose and duplicate lines of code than it is to be over-aggressive in commonizing into helper functions. I'd rather have readable code that takes a couple of more lines than minimized code that is fewer lines but harder to read.

pkg/render/fluentd_test.go
```diff
@@ -1051,3 +1086,44 @@ var _ = Describe("Tigera Secure Fluentd rendering tests", func() {
 		)
 	})
 })
+
+func setupEKSConfig() *render.EksCloudwatchLogConfig {
```
caseydavenport, line 1090: nit: this should be `setupEKSCloudwatchLogConfig` (if we want to have a helper function at all)

vara2504:
@caseydavenport Done :+1:

caseydavenport (approved):

PR #3024, opened 2023-11-30
Allow linseed to read secrets in tigera-operator namespace

asincu:
This permission is needed because Linseed Controller needs to read secrets before copying tigera-voltron-linseed-certs-public certificate to the managed cluster.

## Description

This permission is needed because Linseed Controller needs to read secrets before copying tigera-voltron-linseed-certs-public certificate to the managed cluster.

Error logs:
```
2023-11-28 19:44:27.694 [ERROR][1] token_controller.go 622: Error writing secret to managed cluster cluster="tenant-w99surc9/mihaela-test-create" error=secrets "tigera-voltron-linseed-certs-public" is forbidden: User "system:serviceaccount:tigera-elasticsearch:tigera-linseed" cannot get resource "secrets" in API group "" in the namespace "tigera-operator" tenant="zmv8jboc"
```

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/logstorage.go
+++ b/pkg/render/logstorage.go
@@ -22,6 +22,8 @@ import (
 	"net/url"
 	"strings"
 
+	"github.com/tigera/operator/pkg/common"
+
 	cmnv1 "github.com/elastic/cloud-on-k8s/v2/pkg/apis/common/v1"
 	esv1 "github.com/elastic/cloud-on-k8s/v2/pkg/apis/elasticsearch/v1"
 	kbv1 "github.com/elastic/cloud-on-k8s/v2/pkg/apis/kibana/v1"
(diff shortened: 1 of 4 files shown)


pkg/render/logstorage.go
```diff
@@ -2147,12 +2149,12 @@ func (m managedClusterLogStorage) linseedExternalRolesAndBindings() ([]*rbacv1.C
 		},
 	}
 
-	// Bind the secrets permission to the tigera-fluentd namespace. Other controllers may also bind
+	// Bind the secrets permission to the operator namespace. Other controllers may also bind
```
caseydavenport, line 2152: We should update the comment a bit more here - this binding now adds permissions for Linseed to create its public cert secret in the tigera-operator namespace, and doesn't have anything to do with access tokens, so we should rephrase it to say that.

caseydavenport: We can add to the comment on line 2123 above the clusterrole itself to say "controllers that need Linseed access tokens can bind to this using a role binding in their own namespace"

caseydavenport (approved):

PR #3030, opened 2023-12-02
Version updates for v3.18.0-2.0

danudey:
Update versions for v3.18.0-2.0 release

### DIFF
--- a/pkg/components/images.go
+++ b/pkg/components/images.go
@@ -16,7 +16,7 @@ package components
 // Default registries for Calico and Tigera.
 const (
 	CalicoRegistry = "docker.io/"
-	TigeraRegistry = "gcr.io/unique-caldron-775/cnx/"
+	TigeraRegistry = "quay.io/"
 	// For production InitRegistry should match TigeraRegistry.
 	// For the master branch and other testing scenarios we switch TigeraRegistry to
 	// point to a testing repo but the init image will be pushed to quay, so having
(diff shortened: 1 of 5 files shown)


caseydavenport (approved):

pkg/components/images.go
```diff
@@ -16,7 +16,7 @@ package components
 // Default registries for Calico and Tigera.
 const (
 	CalicoRegistry = "docker.io/"
-	TigeraRegistry = "gcr.io/unique-caldron-775/cnx/"
+	TigeraRegistry = "quay.io/"
```
Brian-McM, line 19: Just wanted to double check this, it's intentional, right?

matthewdupre: The comment on the line below suggests that it could be, although I'd want confirmation.

matthewdupre (approved):

pkg/components/images.go
danudey: @Brian-McM @matthewdupre Yes, this is intentional. TigeraRegistry is where Operator will expect published images to come from; before we release a minor version, we expect any images we generate to come from our dev repo, since obviously we aren't publishing them, but after release if we don't change this then customers' operators will attempt to pull from gcr.io.

(In actuality this won't happen, because it's always caught by our post-release validations, but the issue is the same)

You can see the instructions referenced [in the release docs](https://github.com/tigera/operator/blob/release-v1.32/RELEASING.md#updating-versions), which IIRC we had to add the first time we did a release without updating this field.

PR #3031, opened 2023-12-02
Add validation to ldap secret to catch misconfigurations

rene-dekker:
ev-4161 ci-1331

The secret for LDAP is mounted into the dex pod via ENV and there, Dex is using os.epandenv to insert it into a json config. This means that certain characters need extra escaping. Currently, the user does not see an error in the TigeraStatus when Dex fails to start. This PR adds that validation by performing similar operations.

### DIFF
--- a/pkg/controller/clusterconnection/clusterconnection_controller.go
+++ b/pkg/controller/clusterconnection/clusterconnection_controller.go
@@ -288,7 +288,19 @@ func (r *ReconcileConnection) Reconcile(ctx context.Context, request reconcile.R
 		trustedCertBundle = certificateManager.CreateTrustedBundle()
 	}
 
-	for _, secretName := range []string{render.PacketCaptureServerCert, monitor.PrometheusServerTLSSecretName, render.ProjectCalicoAPIServerTLSSecretName(instl.Variant)} {
+	secretsToTrust := []string{render.PacketCaptureServerCert, render.ProjectCalicoAPIServerTLSSecretName(instl.Variant)}
+	// If external prometheus is enabled, the secret will be signed by the Calico CA and won't get rendered. We can skip
+	// adding it to the bundle, as trusting the CA will suffice.
+	monitorCR := &operatorv1.Monitor{}
+	if err := r.Client.Get(ctx, utils.DefaultTSEEInstanceKey, monitorCR); err != nil {
+		r.status.SetDegraded(operatorv1.ResourceReadError, "Error querying required Monitor resource: ", err, reqLogger)
+		return reconcile.Result{}, err
+	}
+	if monitorCR.Spec.ExternalPrometheus == nil {
+		secretsToTrust = append(secretsToTrust, monitor.PrometheusServerTLSSecretName)
+	}
+
+	for _, secretName := range secretsToTrust {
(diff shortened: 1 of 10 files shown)


pkg/controller/utils/auth.go
```diff
@@ -99,12 +102,36 @@ func GetIDPSecret(ctx context.Context, client client.Client, authentication *ope
...
-		if field == render.BindDNSecretField {
-			if _, err := ldap.ParseDN(string(data)); err != nil {
-				return nil, fmt.Errorf("secret %s/%s field %s: should have be a valid LDAP DN", common.OperatorNamespace(), secretName, field)
-			}
+	// Validate LDAP fields.
+	bindDN := secret.Data[render.BindDNSecretField]
+	bindPW := secret.Data[render.BindPWSecretField]
+	if len(bindDN) > 0 && len(bindPW) > 0 {
+		if _, err := ldap.ParseDN(string(bindDN)); err != nil {
+			return nil, fmt.Errorf("secret %s/%s: should have be a valid LDAP DN", common.OperatorNamespace(), secretName)
```
Brian-McM, line 112: The wording seems a little off here `should have be a`

pkg/controller/utils/auth.go
```diff
@@ -99,12 +102,36 @@ func GetIDPSecret(ctx context.Context, client client.Client, authentication *ope
...
+	}
 
-		if field == render.BindDNSecretField {
-			if _, err := ldap.ParseDN(string(data)); err != nil {
-				return nil, fmt.Errorf("secret %s/%s field %s: should have be a valid LDAP DN", common.OperatorNamespace(), secretName, field)
-			}
+	// Validate LDAP fields.
+	bindDN := secret.Data[render.BindDNSecretField]
+	bindPW := secret.Data[render.BindPWSecretField]
+	if len(bindDN) > 0 && len(bindPW) > 0 {
```
Brian-McM, line 110: Is it valid to have one and not the other? If not, should that be validated too?

pkg/controller/utils/auth.go
```diff
@@ -99,12 +102,36 @@ func GetIDPSecret(ctx context.Context, client client.Client, authentication *ope
...
+		// the input in the operator, the user will catch the configuration issue sooner.
+		const config = `{"bindDN": "$BIND_DN", "bindPW": "$BIND_PW"}`
+		data := []byte(os.Expand(config, func(s string) string {
+			return map[string]string{
+				"BIND_DN": string(bindDN),
+				"BIND_PW": string(bindPW),
+			}[s]
+		}))
+		if err := json.Unmarshal(data, &map[string]any{}); err != nil {
+			return nil, fmt.Errorf("secret %s/%s should have properly escaped data for bindDN and bindPW", common.OperatorNamespace(), secretName)
```
Brian-McM, line 125: Nit, would the error message be better worded stating just that it's not escaped "Fields bindDN and bindPW in secret %s/%s are not properly escaped"?

pkg/controller/utils/auth.go
rene-dekker: They are both part of the `requiredFields` variable, which the loop above checks are all present. So at this point we know that if this is an ldap secret, both of these are present.

PR #3032, opened 2023-12-02
Change rollout settings for Fluentd to speed up rollout, while avoidi…

rene-dekker:
…ng too much contention on huge clusters.


tmjd (approved):
LGTM

PR #3033, opened 2023-12-02
EV-4307 Allow configuration of replicas per-tenant

vara2504:
## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


pkg/render/logstorage/linseed/linseed.go
```diff
@@ -342,6 +342,7 @@ func (l *linseed) linseedDeployment() *appsv1.Deployment {
 		)
 	}
 
+	linseedReplica := l.cfg.Installation.ControlPlaneReplicas
```
asincu, line 345: I think this value can be nil. You we need to consider the following values for control plane replicas:
- l.cfg.Installation.ControlPlaneReplicas or 2 for a non-multitenant environment
- l.cfg.Tenant.Spec.ControlPlaneReplicas or 2 for a multi-tenant environment.

vara2504: @asincu if it is nil (not set by user) we are setting the value to 2 by default in core_controller.
https://github.com/tigera/operator/blob/1fe6e42bb37846ce340c10781a1c70678cb9ee07/pkg/controller/installation/core_controller.go#L735

caseydavenport: We should test nil regardless in case that defaulting changes in the future. It's better to avoid having code like this rely on a subtle bit of logic in another piece of code that lives in a different place.

caseydavenport: nit, but this should be plural - `linseedReplicas` or really just `replicas` (we already know we're talking about Linseed here because of the package and file names)

pkg/render/logstorage/linseed/linseed_test.go
```diff
@@ -483,6 +483,25 @@ var _ = Describe("Linseed rendering tests", func() {
 			Expect(envs).To(ContainElement(corev1.EnvVar{Name: "ELASTIC_THREAT_FEEDS_IP_SET_BASE_INDEX_NAME", Value: "calico_threat_feeds_ip_set_standard"}))
 			Expect(envs).To(ContainElement(corev1.EnvVar{Name: "ELASTIC_WAF_LOGS_BASE_INDEX_NAME", Value: "calico_waflogs_standard"}))
 		})
+		It("should override replica with the value from TenantSpec's controlPlaneReplica when available", func() {
```
caseydavenport, line 486: nit: newline before a new test case

pkg/render/logstorage/linseed/linseed.go
vara2504: @caseydavenport We are dereferencing it while checking for >1
*l.cfg.Installation.ControlPlaneReplicas > 1 but with the nil check before

```
l.cfg.Installation.ControlPlaneReplicas != nil && *l.cfg.Installation.ControlPlaneReplicas > 1) ||
		(l.cfg.Tenant.MultiTenant() && l.cfg.Tenant.Spec.ControlPlaneReplicas != nil && *l.cfg.Tenant.Spec.ControlPlaneReplicas > 1)
```

vara2504:
@asincu @caseydavenport Addressed the review comments. Please take a look.

caseydavenport (approved):

PR #3038, opened 2023-12-04
Fix bug in v1.32.1 that caused the fluentd namespace to be deleted im…

rene-dekker:
caused by: https://github.com/tigera/operator/pull/3027


rene-dekker:
Another PR was already made to address this: https://github.com/tigera/operator/pull/3029

PR #3043, opened 2023-12-06
fix kubefwd URL

binarysta:
just fixing the URL, currently clicking `kubefwd` goes to https://github.com/tigera/operator/blob/master/kubefwd.com


tmjd (approved):
Thank you. LGTM

PR #3049, opened 2023-12-07
Ev 4104 monitor changes

rene-dekker:


pkg/render/monitor/monitor.go
```diff
@@ -1199,3 +1212,143 @@ func (mc *monitorComponent) serviceMonitorCalicoKubeControllers() *monitoringv1.
...
+func (mc *monitorComponent) externalServiceAccount() client.Object {
+	return &corev1.ServiceAccount{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      TigeraExternalPrometheus,
+			Namespace: mc.cfg.Monitor.ExternalPrometheusConfiguration.Namespace,
+		},
+	}
+}
+
+// externalConfigMap creates the configmap with tls certificate scrape our prometheus server.
```
caseydavenport, line 1291: "with tls certificate scrape our"

pkg/render/monitor/monitor.go
```diff
@@ -1199,3 +1212,143 @@ func (mc *monitorComponent) serviceMonitorCalicoKubeControllers() *monitoringv1.
...
+					},
+				},
+			},
+			BearerTokenSecret:    ep.BearerTokenSecret,
+			HonorLabels:          ep.HonorLabels,
+			HonorTimestamps:      ep.HonorTimestamps,
+			MetricRelabelConfigs: ep.MetricRelabelConfigs,
+			RelabelConfigs:       ep.RelabelConfigs,
+		}
+		if ep.BearerTokenSecret.LocalObjectReference.Name == TigeraExternalPrometheus {
```
caseydavenport, line 1331: Making a note for myself to go back and understand this better

caseydavenport (changes requested):
First pass, will keep looking this afternoon though.

api/v1/monitor_types.go
```diff
@@ -17,11 +17,78 @@ limitations under the License.
...
+	// that you configure labels if you want your prometheus instance to pick up the configuration automatically.
+	// The operator will configure 1 endpoint by default:
+	// - Params to scrape all metrics available in Calico Enterprise.
+	// - BearerTokenSecret (If not overridden, the operator will also create corresponding RBAC that allows authz to the metrics.)
+	// - TLSConfig, containing the caFile and serverName.
+	// +optional
+	ServiceMonitor *ServiceMonitor `json:"serviceMonitor,omitempty"`
+
+	// Namespace is the namespace where the operator will create resources for your Prometheus instance. As long as the
+	// namespace does not exist, nothing will be rendered.
```
caseydavenport, line 44: I think this wording should be "The namespace must be created in advance before the Prometheus resources will be created" - "rendered" isn't a term that will resonate with end users, I think.

pkg/controller/monitor/monitor_controller.go
```diff
@@ -225,6 +231,30 @@ func (r *ReconcileMonitor) Reconcile(ctx context.Context, request reconcile.Requ
...
+			if !errors.IsNotFound(err) {
+				r.status.SetDegraded(operatorv1.ResourceReadError, "Failed to get external prometheus namespace", err, reqLogger)
+				return reconcile.Result{}, err
+			}
+			// We set ExternalPrometheus to nil, and proceed to render the other configuration.
+			// When the namespace is created, we will reconcile the ExternalPrometheus.
+			reqLogger.V(3).
+				WithValues("namespace", instance.Spec.ExternalPrometheus.Namespace).
+				Info("will skip external prometheus configuration, namespace does not exist")
+			instance.Spec.ExternalPrometheus = nil
```
caseydavenport, line 253: I think I commented on this in the previous iteration of this PR, but I'm not comfortable overriding user input like this.

We should represent this state differently without trying to piggy-back on the API resource by modifying it (if we even want to do this at all). The more I think about it, the more I wonder why we don't just treat this as invalid configuration - the user is providing us input that isn't actionable, and that is generally considered an error.

pkg/controller/monitor/monitor_controller.go
```diff
@@ -225,6 +231,30 @@ func (r *ReconcileMonitor) Reconcile(ctx context.Context, request reconcile.Requ
...
+	}
+	var externalPrometheus bool
+	if instance.Spec.ExternalPrometheus != nil {
+		if err = r.client.Get(ctx, client.ObjectKey{Name: instance.Spec.ExternalPrometheus.Namespace}, &corev1.Namespace{}); err != nil {
+			if !errors.IsNotFound(err) {
+				r.status.SetDegraded(operatorv1.ResourceReadError, "Failed to get external prometheus namespace", err, reqLogger)
+				return reconcile.Result{}, err
+			}
+			// We set ExternalPrometheus to nil, and proceed to render the other configuration.
+			// When the namespace is created, we will reconcile the ExternalPrometheus.
```
caseydavenport, line 249: Is this extra complexity needed? Why not just block progression if an ExternalPrometheus namespace was provided, but the namespace doesn't exist?

At a minimum, that feels like something we should be emitting in TigeraStatus to alert the user of invalid configuration (which it is).

pkg/controller/monitor/monitor_controller.go
```diff
@@ -262,10 +292,19 @@ func (r *ReconcileMonitor) Reconcile(ctx context.Context, request reconcile.Requ
 		r.status.SetDegraded(operatorv1.ResourceCreateError, "Unable to create the Tigera CA", err, reqLogger)
 		return reconcile.Result{}, err
 	}
-	serverTLSSecret, err := certificateManager.GetOrCreateKeyPair(r.client, monitor.PrometheusServerTLSSecretName, common.OperatorNamespace(), PrometheusTLSServerDNSNames(r.clusterDomain))
-	if err != nil {
-		r.status.SetDegraded(operatorv1.ResourceCreateError, "Error creating TLS certificate", err, reqLogger)
-		return reconcile.Result{}, err
+
+	var serverTLSSecret certificatemanagement.KeyPairInterface
+	if !externalPrometheus || install.CertificateManagement != nil {
```
caseydavenport, line 297: I think this branch could use a comment.

```
// We're either not using an external prometheus in which case we simply sign the KeyPair ourselves, or we're configured to use an external CSR signer to provision certificates which is also handled under the covers by `GetOrCreateKeyPair`.
```

I can see the distinction between `CreateCSRKeyPair` and the CertificateManagement code that is implemented in `GetOrCreateKeyPair` (that also uses CSRs ultimately) to be pretty confusing.

pkg/controller/monitor/monitor_controller.go
```diff
@@ -415,6 +456,37 @@ func (r *ReconcileMonitor) Reconcile(ctx context.Context, request reconcile.Requ
 	return reconcile.Result{}, nil
 }
 
+func fillDefaults(instance *operatorv1.Monitor) {
+	if instance.Spec.ExternalPrometheus != nil && instance.Spec.ExternalPrometheus.ServiceMonitor != nil {
```
caseydavenport, line 460: Is it ever valid to have `Spec.ExternalPrometheus != nil` and `Spec.ExternalPrometheus.ServiceMonitor == nil` ?

Do we need a validation function for this resource as well? I think probably yes

pkg/render/common/networkpolicy/networkpolicy.go
```diff
@@ -254,7 +254,7 @@ func (h *NetworkPolicyHelper) ManagerSourceEntityRule() v3.EntityRule {
 	return CreateSourceEntityRule(h.namespace("tigera-manager"), "tigera-manager")
 }
 
-const PrometheusSelector = "(app == 'prometheus' && prometheus == 'calico-node-prometheus') || (app.kubernetes.io/name == 'prometheus' && prometheus == 'calico-node-prometheus')"
```
caseydavenport, line 257: I'm not following this change - is this related to something in this PR or is it just a cleanup you noticed?

pkg/render/monitor/monitor.go
```diff
@@ -222,7 +226,24 @@ func (mc *monitorComponent) Objects() ([]client.Object, []client.Object) {
 		toCreate = append(toCreate, mc.prometheusOperatorPodSecurityPolicy())
 	}
 
+	if mc.cfg.Monitor.ExternalPrometheus != nil {
+		toCreate = append(toCreate, mc.externalConfigMap())
+		if mc.cfg.ExternalPrometheus && mc.cfg.Monitor.ExternalPrometheus.ServiceMonitor != nil {
+			externalServiceMonitor, needsRBAC := mc.externalServiceMonitor()
+			toCreate = append(toCreate, externalServiceMonitor)
+			if needsRBAC {
+				toCreate = append(toCreate, mc.externalPrometheusRole(), mc.externalPrometheusRoleBinding(), mc.externalServiceAccount(), mc.externalPrometheusTokenSecret())
```
caseydavenport, line 235: Do we need to `toDelete` these if ExternalPrometheus is nil?

pkg/render/monitor/monitor.go
```diff
@@ -1088,6 +1109,18 @@ func allowTigeraPrometheusPolicy(cfg *Config) *v3.NetworkPolicy {
 			Destination: render.DexEntityRule,
 		},
 	}...)
+	typhaMetricsPort := cfg.Installation.TyphaMetricsPort
```
caseydavenport, line 1112: nit: newline before this

pkg/render/monitor/monitor.go
```diff
@@ -1199,3 +1232,181 @@ func (mc *monitorComponent) serviceMonitorCalicoKubeControllers() *monitoringv1.
...
+func (mc *monitorComponent) externalPrometheusRole() client.Object {
+	return &rbacv1.ClusterRole{
+		TypeMeta: metav1.TypeMeta{Kind: "ClusterRole", APIVersion: "rbac.authorization.k8s.io/v1"},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      TigeraExternalPrometheus,
+			Namespace: mc.cfg.Monitor.ExternalPrometheus.Namespace,
+		},
+		Rules: []rbacv1.PolicyRule{
+			{
+				// When prometheus was first introduced it was accessed through k8s services/proxy and so to this day,
```
caseydavenport, line 1246: "When prometheus was first introduced" implies to me that it is no longer accessed through services/proxy?

pkg/render/monitor/monitor.go
```diff
@@ -1199,3 +1232,181 @@ func (mc *monitorComponent) serviceMonitorCalicoKubeControllers() *monitoringv1.
...
+			APIGroup: "rbac.authorization.k8s.io",
+			Kind:     "ClusterRole",
+			Name:     TigeraExternalPrometheus,
+		},
+	}
+}
+
+// externalPrometheusTokenSecret creates the bearer token on which behalf requests will be made from the external prometheus
+// server to ours.
+func (mc *monitorComponent) externalPrometheusTokenSecret() client.Object {
```
caseydavenport, line 1284: https://kubernetes.io/docs/concepts/security/service-accounts/#get-a-token

I think we should be using token projection instead of an explicitly created secret here - is that feasible?

caseydavenport: If not, we'll probably want to do a follow-on to use the token request API instead of a long-lived token with no expiry.

pkg/render/monitor/monitor.go
caseydavenport: > But if the user chooses to override the bearer token

Do we really have a requirement to allow end-users to override bearer tokens?

pkg/render/typha.go
```diff
@@ -700,3 +703,31 @@ func (c *typhaComponent) affinity() (aff *corev1.Affinity) {
 	}
 	return aff
 }
+
+// typhaPrometheusService service for scraping typha metrics.
```
caseydavenport, line 707: Our docs currently include steps to create this service: https://docs.tigera.io/calico-enterprise/latest/operations/monitor/prometheus/byo-prometheus

Just adding a reminder here that we'll need to submit a corresponding docs PR to remove this step.

pkg/tls/certificatemanagement/csr.go
```diff
@@ -101,7 +100,7 @@ func ResolveCSRInitImage(inst *operatorv1.InstallationSpec, is *operatorv1.Image
 }
 
 // CSRClusterRole returns a role with the necessary permissions to create certificate signing requests.
-func CSRClusterRole() client.Object {
+func CSRClusterRole() *rbacv1.ClusterRole {
```
caseydavenport, line 103: Did we need to change the return type here?

pkg/controller/csr/csr_controller.go
```diff
@@ -222,8 +231,27 @@ func (r *reconcileCSR) Reconcile(ctx context.Context, request reconcile.Request)
...
 
-	return reconcile.Result{}, nil
+	var passthrough render.Component
+	if needsCSRRole {
+		// This controller creates the cluster role for any pod in the cluster that requires certificate management.
+		passthrough = render.NewPassthrough(certificatemanagement.CSRClusterRole())
+	} else {
+		passthrough = render.NewDeletionPassthrough(certificatemanagement.CSRClusterRole())
+	}
+	componentHandler := utils.NewComponentHandler(log, r.client, r.scheme, instance)
```
caseydavenport, line 253: I think we probably want to create this prior to the loop above?

It's a bit of a niche scenario, but couldn't an error in the above loop result in us not creating / updating / deleting this, resulting in a deadlock? 

I think in practice it's unlikely because we skip CSRs that aren't for us, but if there's a CSR that _is_ for us somehow and is causing problems in the loop above, then we'll never execute this code until that CSR is resolved. Maybe that's OK, but I like to avoid unnecessary dependencies like this because they introduce complexity into the debugging process. It seems a more reasonable ordering of dependencies that we'd never handle any CSRs until the ClusterRole that gives our components permissions to create CSRs had been created / updated.

caseydavenport: Also, I think for now `needsCSRRole` is the same as "Should this controller do anything?" so we can probably short-circuit the controller if we move this earlier too?

pkg/controller/monitor/monitor_controller.go
rene-dekker: A user may want to get the TLS bundle without Tigera creating the Service monitor. In that case ExternalPrometheus is not nil.

Since we no longer embed all of the prometheus ServiceMonitor any longer, but only a subset we no longer have any need for validation, so I removed it. Fields related to being able to access our Prometheus are controlled and defaulted by us. I don't think there are any contradictory configurations possible any longer.

pkg/render/common/networkpolicy/networkpolicy.go
rene-dekker: The feature branch this is targeting now specifically sets our own label k8s-app=tigera-prometheus. Previously, we relied on what labels prometheus put on our pods, which are defined in the above selector. So it is a cleanup, but related to the feature branch. Maybe it could have been in a separate PR.

pkg/render/monitor/monitor.go
rene-dekker: This is not possible, because if ExternalPrometheus is nil, we don't know what namespace to delete it from.

pkg/render/monitor/monitor.go
rene-dekker: That's right. But the RBAC is still enforced in our backend servers.

pkg/render/monitor/monitor.go
caseydavenport: The way to do this would be with a finalizer, so that we have an opportunity to clean up before the object disappears.

caseydavenport: We do this in the user controller already and it's fairly straightforward: https://github.com/tigera/operator/blob/d54d499ca96085a38878a21996da2d5eff1f3cf5/pkg/controller/logstorage/users/users_controller.go#L400-L406

pkg/render/monitor/monitor.go
caseydavenport: Which code enforces this? I'm still not quite following - if it's no longer accessed through services/proxy, then why do we need RBAC for that?

pkg/render/monitor/monitor.go
rene-dekker: I only see authn/authz options that involve reading a file or a secret in the ServiceMonitorSpec: BearerTokenFile, BearerTokenSecret, Authorization, which relies on a secretKeySelector. So I think today there is not yet an option to use the token api.
We do allow the user to override our BearerTokenSecret, so we would not render this secret, but that is all that I was able to figure out.

pkg/render/typha.go
rene-dekker: Great catch. External prometheus and the current steps for byo prometheus individual settings are supposed to be mutually exclusive. And if they are used in combination then you are right that that service is no longer necessary. The docs changes can be found here: https://github.com/tigera/docs/pull/1192

pkg/render/monitor/monitor.go
rene-dekker: What would have been best is if we create our own tigera.io rbac resource for authz. However, when we first moved away from using services/proxy, we kept the same RBAC attributes. Today both the tigera/prometheus and tigera/queryserver are still checking that the request is authorized based on these attributes.  I can make a ticket for implementing this.

caseydavenport (approved):
Approving for now since I think this is in a good state. A few follow-on actions:

- Expiry for serviceaccount tokens
- Finalizers and cleanup of resources on deletion (and perhaps some other way to do this for non-deletion case)

PR #3052, opened 2023-12-09
PolicyRecommendation needs to access resource within a tenant namespace

asincu:
## Description

Enable access per tenant for PolicyRecommendation:
- Access per tenant Linseed (use the tenant namespace service and modify networkpolicy to access resources in the tenant namespace)
- Access per tenant Manager (use the tenant namespace service and modify networkpolicy to access resources in the tenant namespace)
- Bind RBAC using the tenant service account
- Add impersonation rights for multi-tenant
- Allow to get managed cluster in from tigera-policy-recommendation in a multi-tenant environment

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/policyrecommendation/policyrecommendation_controller.go
+++ b/pkg/controller/policyrecommendation/policyrecommendation_controller.go
@@ -335,12 +335,12 @@ func (r *ReconcilePolicyRecommendation) Reconcile(ctx context.Context, request r
 	handler := utils.NewComponentHandler(log, r.client, r.scheme, policyRecommendation)
 
 	// Determine the namespaces to which we must bind the cluster role.
-	var bindNamespaces = []string{ResourceName}
-	if tenant.MultiTenant() {
-		bindNamespaces, err = helper.TenantNamespaces(r.client)
-		if err != nil {
-			return reconcile.Result{}, err
-		}
+	// For multi-tenant, the cluster role will be bind to the service account in the tenant namespace
+	// For single-tenant or zero-tenant, the cluster role will be bind to the service account in the tigera-policy-recommendation
+	// namespace
+	bindNamespaces, err := helper.TenantNamespaces(r.client)
+	if err != nil {
+		return reconcile.Result{}, err
 	}
 
 	logc.V(3).Info("rendering components")
--- a/pkg/controller/policyrecommendation/policyrecommendation_controller.go
+++ b/pkg/controller/policyrecommendation/policyrecommendation_controller.go
@@ -334,16 +334,23 @@ func (r *ReconcilePolicyRecommendation) Reconcile(ctx context.Context, request r
 	// Create a component handler to manage the rendered component.
 	handler := utils.NewComponentHandler(log, r.client, r.scheme, policyRecommendation)
 
+	// Determine the namespaces to which we must bind the cluster role.
+	bindNamespaces, err := helper.TenantNamespaces(r.client)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
 	logc.V(3).Info("rendering components")
 	policyRecommendationCfg := &render.PolicyRecommendationConfiguration{
-		ClusterDomain:  r.clusterDomain,
-		Installation:   installation,
-		ManagedCluster: isManagedCluster,
-		PullSecrets:    pullSecrets,
-		Openshift:      r.provider == operatorv1.ProviderOpenShift,
-		UsePSP:         r.usePSP,
-		Namespace:      helper.InstallNamespace(),
-		Tenant:         tenant,
+		ClusterDomain:     r.clusterDomain,
+		Installation:      installation,
+		ManagedCluster:    isManagedCluster,
+		PullSecrets:       pullSecrets,
+		Openshift:         r.provider == operatorv1.ProviderOpenShift,
+		UsePSP:            r.usePSP,
+		Namespace:         helper.InstallNamespace(),
+		Tenant:            tenant,
+		BindingNamespaces: bindNamespaces,
 	}
 
 	// Render the desired objects from the CRD and create or update them.
--- a/pkg/controller/policyrecommendation/policyrecommendation_controller.go
+++ b/pkg/controller/policyrecommendation/policyrecommendation_controller.go
@@ -335,9 +335,12 @@ func (r *ReconcilePolicyRecommendation) Reconcile(ctx context.Context, request r
 	handler := utils.NewComponentHandler(log, r.client, r.scheme, policyRecommendation)
 
 	// Determine the namespaces to which we must bind the cluster role.
-	bindNamespaces, err := helper.TenantNamespaces(r.client)
-	if err != nil {
-		return reconcile.Result{}, err
+	var bindNamespaces = []string{ResourceName}
+	if tenant.MultiTenant() {
+		bindNamespaces, err = helper.TenantNamespaces(r.client)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
 	}
 
 	logc.V(3).Info("rendering components")
--- a/pkg/render/fluentd.go
+++ b/pkg/render/fluentd.go
@@ -656,13 +656,11 @@ func (c *fluentdComponent) metricsService() *corev1.Service {
 }
 
 func (c *fluentdComponent) envvars() []corev1.EnvVar {
-	// Determine the namespace in which Linseed is running. For managed and standalone clusters, this is always the elasticsearch
-	// namespace. For multi-tenant management clusters, this may vary.
-	linseedNS := LinseedNamespace(c.cfg.Tenant)
-
 	envs := []corev1.EnvVar{
 		{Name: "LINSEED_ENABLED", Value: "true"},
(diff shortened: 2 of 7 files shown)


pkg/controller/policyrecommendation/policyrecommendation_controller.go
```diff
@@ -334,16 +334,23 @@ func (r *ReconcilePolicyRecommendation) Reconcile(ctx context.Context, request r
 	// Create a component handler to manage the rendered component.
 	handler := utils.NewComponentHandler(log, r.client, r.scheme, policyRecommendation)
 
+	// Determine the namespaces to which we must bind the cluster role.
+	bindNamespaces, err := helper.TenantNamespaces(r.client)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
```
caseydavenport, line 341: I think we will need a clause like this:

```go
if tenant.MultiTenant() {
  bindNamespaces = append(bindNamespaces, render.PolicyRecommendationNamespace)
}
```

To support impersonation.

pkg/render/policyrecommendation.go
```diff
@@ -17,6 +17,8 @@ package render
 import (
 	"crypto/x509"
 
+	rcomponents "github.com/tigera/operator/pkg/render/common/components"
```
caseydavenport, line 20: import should be grouped with the other operator imports.

pkg/render/policyrecommendation.go
```diff
@@ -109,8 +112,8 @@ func (pr *policyRecommendationComponent) Objects() ([]client.Object, []client.Ob
 	objs := []client.Object{
 		CreateNamespace(pr.cfg.Namespace, pr.cfg.Installation.KubernetesProvider, PSSRestricted),
 		pr.serviceAccount(),
-		pr.clusterRole(),
-		pr.clusterRoleBinding(),
+		pr.clusterRole(pr.cfg.ManagedCluster),
```
caseydavenport, line 115: We shouldn't need to pass in the cfg as an argument - these functions are already bound to the `pr` instance so can access the cfg directly.

pkg/render/policyrecommendation_test.go
```diff
@@ -347,11 +353,30 @@ var _ = Describe("Policy recommendation rendering tests", func() {
...
-			envs = d.Spec.Template.Spec.Containers[0].Env
-			Expect(envs).To(ContainElement(corev1.EnvVar{Name: "TENANT_NAMESPACE", Value: tenantBNamespace}))
-			Expect(envs).To(ContainElement(corev1.EnvVar{Name: "TENANT_ID", Value: "tenant-b-id"}))
-
+			verifyTenantResources(tenantBResources, tenantBNamespace, "tenant-b-id")
 		})
 	})
 })
+
+func verifyTenantResources(createdResources []client.Object, namespace string, tenantID string) {
```
caseydavenport, line 361: This is somewhat personal preference, but I would really like if we avoid these types of catch-all helper functions in tests for a few reasons:

- It makes test output and line numbers convoluted and harder to follow (ExpectWithOffset helps a bit, but not much)
- It assumes every test wants to test the same thing, which isn't the case. Especially in the future we may alter expectations which will require making this function even more complicated, or would require that we write tests that don't use this function anyway.
- It makes it harder to read through a test and understand what the test is actually trying to verify.

I would _much_ rather we write targeted tests. There's no need for _every_ test to assert that the clusterrole / binding is done correctly - let's just have one test that asserts that's correct in a multi-tenant environment. Then it becomes really clear what's going on when it fails. With this model, if we introduce a bug into the RBAC resources, every multi-tenant UT will fail and it will be less obvious what the problem is.

pkg/render/utils.go
```diff
@@ -0,0 +1,42 @@
...
+import (
+	"fmt"
+
+	operatorv1 "github.com/tigera/operator/api/v1"
+)
+
+// LinseedNamespace determine the namespace in which Linseed is running.
+// For management and standalone clusters, this is always the tigera-elasticsearch
+// namespace. For multi-tenant management clusters, this is the tenant namespace
+func LinseedNamespace(tenant *operatorv1.Tenant) string {
```
caseydavenport, line 26: Any reason not to put this in the `linseed` package?

pkg/render/utils.go
```diff
@@ -0,0 +1,42 @@
...
+		return tenant.Namespace
+	}
+	return ElasticsearchNamespace
+}
+
+// ManagerService determine the name of the tigera manager service.
+// For management and standalone clusters, this is always the tigera-manager.tigera-manager
+// namespace. For multi-tenant management clusters, this is a service that resides within the
+// tenant namespace
+func ManagerService(tenant *operatorv1.Tenant) string {
```
caseydavenport, line 37: Same here - can we just put this in the `manager.go` file?

pkg/render/policyrecommendation_test.go
```diff
@@ -321,6 +326,7 @@ var _ = Describe("Policy recommendation rendering tests", func() {
 					ID: "tenant-b-id",
 				},
 			}
+			cfg.BindingNamespaces = []string{tenantBNamespace}
```
caseydavenport, line 329: We should include both namespaces here so that we properly verify the binding includes two subjects.

pkg/render/policyrecommendation.go
```diff
@@ -197,28 +205,8 @@ func (pr *policyRecommendationComponent) clusterRole() client.Object {
...
-		},
-		Subjects: []rbacv1.Subject{
-			{
-				Kind:      "ServiceAccount",
-				Name:      PolicyRecommendationName,
-				Namespace: pr.cfg.Namespace,
-			},
-		},
-	}
+func (pr *policyRecommendationComponent) clusterRoleBinding(namespaces []string) client.Object {
```
caseydavenport, line 208: I think this is only used in this file, so we can get rid of this function and just call `rcomponents.ClusterRoleBinding` directly. 

The only reason we have a helper for the manager one is because it's also called from the guardian render code, IIRC

pkg/render/policyrecommendation.go
```diff
@@ -179,6 +172,21 @@ func (pr *policyRecommendationComponent) clusterRole() client.Object {
 		},
 	}
 
+	if !managedCluster {
```
caseydavenport, line 175: How sure are you about this change? Was the existing RBAC causing a problem?

pkg/render/fluentd.go
```diff
@@ -1063,10 +1060,7 @@ func (c *fluentdComponent) eksLogForwarderDeployment() *appsv1.Deployment {
 
 	// Determine the namespace in which Linseed is running. For managed and standalone clusters, this is always the elasticsearch
 	// namespace. For multi-tenant management clusters, this may vary.
-	linseedNS := ElasticsearchNamespace
-	if c.cfg.Tenant.MultiTenant() {
-		linseedNS = c.cfg.Tenant.Namespace
-	}
+	linseedNS := LinseedNamespace(c.cfg.Tenant)
```
caseydavenport, line 1063: We don't need to declare an intermediate variable here now that we have a helper function

caseydavenport (changes requested):

pkg/render/policyrecommendation_test.go
asincu: Np. I can create a new test to check RBAC. Maybe we should rename this test from `should render expected components inside expected namespace for each policyrecommendation instance` to something more specific then ? Seems like we are checking for components and some environment variables.

pkg/render/utils.go
asincu:
```
package github.com/tigera/operator/pkg/render/logstorage/linseed
	imports github.com/tigera/operator/pkg/render
	imports github.com/tigera/operator/pkg/render/logstorage/linseed: import cycle not allowed
```

pkg/render/policyrecommendation.go
asincu: ManagedCluster API is not enabled on a managed cluster and policy recommendation only reads the licensekey from the management cluster.

pkg/controller/policyrecommendation/policyrecommendation_controller.go
caseydavenport: This isn't quite what I meant - here's an example for the manager controller:

https://github.com/tigera/operator/blob/713ed247a1d1f0e921970d03d620a0bb6f700f65/pkg/controller/manager/manager_controller.go#L622-L632

caseydavenport: That said, you have handled this differently here: https://github.com/tigera/operator/pull/3052/files#diff-28bf4bd22da8886729331acba1bb2f7c4e423502ac4aed05ce35ac05e1e8e257R240-R281

So I think there's no need to do _both_ of these - you can leave this how it was before if you are already granting those permissions via that special binding.

```go
		bindNamespaces, err := helper.TenantNamespaces(r.client)
		if err != nil {
		}
```

pkg/render/policyrecommendation_test.go
caseydavenport: I think it's good / fine to have one test that verifies the high level stuff - the correct resources were rendered - but then we should have fine-grained tests that verify specific inputs map to specific outputs.

caseydavenport (approved):

PR #3053, opened 2023-12-11
[RS-1574] Linseed misconfiguration for Webhooks in MCM configuration

bartolini:
This PR fixes the issue with missing `LINSEED_CLUSTER` environment variable set for the `webhooks-processor` in the IDC deployment for managed clusters.

Original Jira ticket + before/after fix details: https://tigera.atlassian.net/browse/RS-1574


electricjesus (approved):

pkg/render/intrusion_detection.go
```diff
@@ -789,6 +789,10 @@ func (c *intrusionDetectionComponent) webhooksControllerContainer() corev1.Conta
 			Name:  "LINSEED_TOKEN",
 			Value: GetLinseedTokenPath(c.cfg.ManagedCluster),
 		},
+		{
+			Name:  "LINSEED_CLUSTER",
```
asincu, line 793: @bartolini I would suggest we do not introduce a new variable to set the cluster name. We are currently trying to move away from using this config map. For managed clusters, the managed cluster ID will be set by Voltron. I think you probably need to leave cluster name blank when using the Linseed client.

PR #3054, opened 2023-12-11
Remove curator

Josh-Tigera:
## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


pkg/render/logstorage.go
```diff
@@ -425,7 +376,6 @@ func (es *elasticsearchComponent) Objects() ([]client.Object, []client.Object) {
 			toCreate = append(toCreate, secret.ToRuntimeObjects(secret.CopyToNamespace(ElasticsearchNamespace, es.cfg.KeyStoreSecret)...)...)
 		}
 		toDelete = append(toDelete, es.kibanaCR())
-		toDelete = append(toDelete, es.curatorCronJob())
```
caseydavenport, line 428: We might still want to include logic to delete curator from the cluster, since otherwise I think upgraded clusters will continue to have a downlevel curator component installed.

caseydavenport: - service account
- any rbac
- the job itself
- any other supporting resources

caseydavenport (commented):
Looking good, the main thing is that I think we should still keep some logic around to clean up es-curator resources from a cluster on upgrade.

caseydavenport (approved):
So happy

PR #3059, opened 2023-12-13
add unit testing for WAF render

electricjesus:
## Description
- add unit-tests for WAF, specifically for flags control and expected vols/volMounts
- added embed.AsConfigMap so that it's simpler to inject a config map, just an dry approach.. possibly an optional change

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/applicationlayer/applicationlayer_test.go
+++ b/pkg/render/applicationlayer/applicationlayer_test.go
@@ -564,7 +564,7 @@ var _ = Describe("Tigera Secure Application Layer rendering tests", func() {
 			"--waf-ruleset-base-dir", applicationlayer.ModSecurityRulesetVolumePath,
 			"--waf-directive", "Include modsecdefault.conf",
 			"--waf-directive", "Include crs-setup.conf",
-			"--waf-directive", "Include REQUEST-*.conf",
+			"--waf-directive", "Include rules/*.conf",
 			"--waf-directive", "Include tigera.conf",
 		}
 		for _, element := range expectedDikastesArgs {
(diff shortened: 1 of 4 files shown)


pkg/render/applicationlayer/applicationlayer_test.go
```diff
@@ -410,4 +413,176 @@ var _ = Describe("Tigera Secure Application Layer rendering tests", func() {
 			Expect(dikastesVolMounts).To(ContainElement(expected))
 		}
 	})
+
+	It("should render with default l7 WAF configuration", func() {
+		expectedResources := []struct {
```
caseydavenport, line 418: Highly recommend moving away from this model of test.

Use `ExpectResource` instead: https://github.com/tigera/operator/blob/e5450ce876012b3e2c4392b2e36f3fe71750849d/pkg/render/common/test/testing.go#L93-L105

Note the comment on `ExpectResourceTypeAndObjectMetadata`: https://github.com/tigera/operator/blob/e5450ce876012b3e2c4392b2e36f3fe71750849d/pkg/render/common/test/testing.go#L141-L142

caseydavenport: I'd expect this to be a `[]client.Object` instead of a `[]struct`, in order for ExpectResource / ExpectResources to behave corretly.

caseydavenport: Part of the intention is that we can use the actual structs - `v1.ServiceAccount`, etc - instead of "fake" structs that are more error prone.

electricjesus:
closed - already addressed in another PR

PR #3069, opened 2023-12-15
Ev-4161

rene-dekker:


pkg/controller/utils/auth.go
```diff
@@ -99,12 +102,36 @@ func GetIDPSecret(ctx context.Context, client client.Client, authentication *ope
...
+		// the input in the operator, the user will catch the configuration issue sooner.
+		const config = `{"bindDN": "$BIND_DN", "bindPW": "$BIND_PW"}`
+		data := []byte(os.Expand(config, func(s string) string {
+			return map[string]string{
+				"BIND_DN": string(bindDN),
+				"BIND_PW": string(bindPW),
+			}[s]
+		}))
+		if err := json.Unmarshal(data, &map[string]any{}); err != nil {
+			return nil, fmt.Errorf("fields bindDN and bindPW in secret %s/%s are not properly escaped", common.OperatorNamespace(), secretName)
```
Brian-McM, line 125: I don't know that this 100% means that the fields are malformed, all we know is that there was a json error (and this is a likely suspect). You should probably include the original error by wrapping it with `%w`.

Brian-McM (approved):

pkg/controller/utils/auth.go
rene-dekker: Maybe I should have clarified, but the error messages are extremely cryptic. I don't believe they are helpful. For example:
```
in: CN=example,OU="finance",DC=com
err:  invalid character 'f' after object key:value pair

in: CN=example,OU=finance",DC=com
err: invalid character 'D' looking for beginning of object key string
```

PR #3075, opened 2023-12-19
Break infinite reconciliation loop

rene-dekker:
Without this change, the core_controller stays in an infinite loop. The old and new service account objects are not recognized as being equivalents.

image

Since I don't see any related release notes in controller-runtime, I think opening a PR is the best way forward.


caseydavenport (approved):
LGTM as a fix.

I do think this is another case for a longer term investigation of a more intelligent field tracking, e.g., using managed fields: https://kubernetes.io/docs/reference/using-api/server-side-apply/#field-management

PR #3077, opened 2023-12-20
Ensure that at least one of ApplicationLayer features are enabled

radixo:
## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/crds/operator/operator.tigera.io_applicationlayers.yaml
+++ b/pkg/crds/operator/operator.tigera.io_applicationlayers.yaml
@@ -85,6 +85,11 @@ spec:
                   opt-in to having ingress traffic examed by ModSecurity.
                 type: string
             type: object
+            x-kubernetes-validations:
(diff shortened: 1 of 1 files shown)


tmjd:
You need to figure out how to set this by adding a comment in `api/v1/applicationlayer_types.go` or else the changes will be removed when the CRDs are regenerated.

Though I don't think that we should make this change, it would mean that a previous valid resource would now fail/be invalid.
@caseydavenport WDYT?

caseydavenport:
Need to check this one.. I seem to recall requesting this type of validation in the past for this CR, so if we don't currently have it there's probably a reason we opted against it :sweat_smile: 

@radixo could you share some context as to why you're opting to make this change now? What prompted this PR?

electricjesus:
closed due to age / staleness -- please reopen when there's new updates!

PR #3090, opened 2024-01-06
Fix FLOW_LOG_FILE fluentd env var for Windows

coutinhop:
Felix on Windows actually writes flow logs to "c:/TigeraCalico/flowlogs/flows.log", so use that value when rendering fluentd windows pods.

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/crds/enterprise/crd.projectcalico.org_felixconfigurations.yaml
+++ b/pkg/crds/enterprise/crd.projectcalico.org_felixconfigurations.yaml
@@ -81,12 +81,25 @@ spec:
                   [Default: unset - means logs are emitted when BPFLogLevel id debug
                   and BPFLogFilters not set.]'
                 type: string
+              bpfConnectTimeLoadBalancing:
+                description: 'BPFConnectTimeLoadBalancing when in BPF mode, controls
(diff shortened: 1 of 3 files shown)


rene-dekker:
Why did you choose to change it here, rather than in the windows node? It seems a bit inconsistent to me to have everything in c:/var/log, except this.

coutinhop:
> Why did you choose to change it here, rather than in the windows node? It seems a bit inconsistent to me to have everything in c:/var/log, except this.

To be honest, I wasn't 100% certain I'd understood the flow log writing code in Felix, so I went with changing as least as possible to see if that fixes the Windows flow logs issue. (I'm a bit stumped as I couldn't see a difference when running the tests with a custom operator images that had these changes, but the cluster had already been created and I'm not sure if there's any old config that was persisting).

@rene-dekker if that is OK with you, I'd rather get this in and see if it fixes things, then look into major changes to make Windows paths more consistent, WDYT? (For instance, it doesn't show in the operator config, but the flow logs position file path is "c:/TigeraCalico/flowlogs/flows.log.pos")

rene-dekker (approved):
lgtm

PR #3115, opened 2024-01-12
[RS-1691] Webhooks-controller should now be rendered on management clusters.

bartolini:
Original Jira ticket: https://tigera.atlassian.net/browse/RS-1691

Previously (by design) `webhooks-controller` was not rendered on management cluster in MCM set-up. The PR changes this behavior as it is now expected to have the controller up-and-running on the management clusters as well.


gantony (approved):
LGTM

PR #3117, opened 2024-01-12
[cherrypick] [RS-1691] Webhooks-controller should now be rendered on management clusters.

bartolini:
Original PR: https://github.com/tigera/operator/pull/3115


gantony (approved):

pkg/render/intrusion_detection_test.go
```diff
@@ -397,152 +397,6 @@ var _ = Describe("Intrusion Detection rendering tests", func() {
 			}))
 	})
 
-	It("should not render webhooks-controller on management clusters", func() {
```
caseydavenport, line 400: We need to implement a corresponding test "It should render webhooks controller on management clusters" right?

caseydavenport: And it "should not render webhooks controller on multi-tenant management clusters"

bartolini: @caseydavenport this is covered in the `should render all resources for a default configuration` test case. Granted - it does not cover multi-tenant management clusters, however we need to establish the base-ground for this as we are only discussing it in the thread here: https://tigera.slack.com/archives/CN5UQVAQG/p1705073696492089

caseydavenport: I suppose the actually resource is covered there, although the intent isn't (i.e., a default configuration isn't a management cluster). 

For multi-tenancy I agree - I think I spoke a bit soon on this one and we can ignore that test case for now.

electricjesus (approved):

PR #3118, opened 2024-01-12
Consolidate tigera status for all log storage controllers

vara2504:
## Description

```
  status:
    conditions:
    - lastTransitionTime: "2024-01-25T22:17:31Z"
      message: ""
      reason: Unknown
      status: "False"
      type: Degraded
    - lastTransitionTime: "2024-01-25T22:17:31Z"
      message: All sub-controllers are available
      reason: AllObjectsAvailable
      status: "True"
      type: Ready
    - lastTransitionTime: "2024-01-25T22:17:31Z"
      message: ""
      reason: Unknown
      status: "False"
      type: Progressing

```

```
  status:
    conditions:
    - lastTransitionTime: "2024-01-25T22:55:42Z"
      message: ""
      reason: Unknown
      status: "False"
      type: Progressing
    - lastTransitionTime: "2024-01-25T23:02:49Z"
      message: 'The following sub-controllers are in this condition: [log-storage-access
        log-storage-elastic log-storage-esmetrics log-storage-kubecontrollers]'
      reason: ResourceNotReady
      status: "True"
      type: Degraded
    - lastTransitionTime: "2024-01-25T22:55:03Z"
      message: ""
      reason: ResourceNotReady
      status: "False"
      type: Ready

```

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/logstorage/initializer/conditions_controller.go
+++ b/pkg/controller/logstorage/initializer/conditions_controller.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2020-2023 Tigera, Inc. All rights reserved.
+// Copyright (c) 2020-2024 Tigera, Inc. All rights reserved.
 
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -17,10 +17,11 @@ package initializer
 import (
 	"context"
 	"fmt"
+	"time"
 
-	operatorv1 "github.com/tigera/operator/api/v1"
-
+	"github.com/tigera/operator/pkg/controller/logstorage/esmetrics"
 	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -30,8 +31,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 
+	operatorv1 "github.com/tigera/operator/api/v1"
 	"github.com/tigera/operator/pkg/controller/options"
-	"github.com/tigera/operator/pkg/controller/status"
 	"github.com/tigera/operator/pkg/controller/utils"
 )
 
@@ -90,18 +91,127 @@ func (r *LogStorageConditions) Reconcile(ctx context.Context, request reconcile.
 		return reconcile.Result{}, err
 	}
 
-	// Update conditions on the LogStorage object based on the status of the various components.
-	// TODO: Incorporate status from other sub controllers as well.
-	ts := &operatorv1.TigeraStatus{}
-	if err := r.client.Get(ctx, types.NamespacedName{Name: TigeraStatusName}, ts); err != nil {
+	// Fetch the current status condition for log storage
+	currentConditions := getCurrentConditions(ls.Status.Conditions)
+
+	// Aggregate TigeraStatus conditions from all logstorage subcontrollers into a map,
+	// using the condition type (e.g., Available, Progressing, and Degraded) as the key.
+	desiredConditions, err := r.getDesiredConditions(ctx)
+	if err != nil {
 		return reconcile.Result{}, err
 	}
 
-	ls.Status.Conditions = status.UpdateStatusCondition(ls.Status.Conditions, ts.Status.Conditions)
+	// Compare and update the current StatusCondition if there are any new changes
+	ls.Status.Conditions = updateConditions(currentConditions, desiredConditions)
+
 	if err := r.client.Status().Update(ctx, ls); err != nil {
 		log.WithValues("reason", err).Info("Failed to update LogStorage status conditions")
 		return reconcile.Result{}, err
 	}
 
 	return reconcile.Result{}, nil
+}
+
+func getCurrentConditions(lsConditions []metav1.Condition) map[string]metav1.Condition {
+	statusConditions := make(map[string]metav1.Condition)
+	for _, condition := range lsConditions {
+		statusConditions[condition.Type] = condition
+	}
+	return statusConditions
+}
+
+func (r *LogStorageConditions) getDesiredConditions(ctx context.Context) (map[string]metav1.Condition, error) {
+
+	expectedInstances := []string{TigeraStatusName, TigeraStatusLogStorageAccess, TigeraStatusLogStorageElastic, TigeraStatusLogStorageSecrets}
+	if r.multiTenant {
+		expectedInstances = append(expectedInstances, TigeraStatusLogStorageUsers)
+	} else {
+		expectedInstances = append(expectedInstances, esmetrics.TigeraStatusLogStorageESMetrics, TigeraStatusLogStorageKubeController)
+	}
+
+	// Keep track of which instances are in which state.
+	states := map[string][]string{
+		string(operatorv1.ComponentDegraded):    {},
+		string(operatorv1.ComponentAvailable):   {},
+		string(operatorv1.ComponentProgressing): {},
+	}
+
+	// We also keep track of the oldest observed generation here so we can add it to the conditions later.
+	var observedGeneration int64
+
+	// Build up the lists of which components are in which state.
+	for _, instance := range expectedInstances {
+		ts := &operatorv1.TigeraStatus{}
+		if err := r.client.Get(ctx, types.NamespacedName{Name: instance}, ts); err != nil && errors.IsNotFound(err) {
+			// Expected status not found, treat it as an error.
+			states[string(operatorv1.ComponentDegraded)] = append(states[string(operatorv1.ComponentDegraded)], instance)
+			continue
+		} else if err != nil {
+			return nil, err
+		}
+
+		// Found it - add to states map, which we use to track which TigeraStatuses are in which state(s).
+		for _, condition := range ts.Status.Conditions {
+			if condition.Status == operatorv1.ConditionTrue {
+				states[string(condition.Type)] = append(states[string(condition.Type)], instance)
+			}
+			if observedGeneration == 0 || condition.ObservedGeneration < observedGeneration {
+				observedGeneration = condition.ObservedGeneration
+			}
+		}
+	}
+
+	// Convert the states map to a set of conditions.
+	conditions := map[string]metav1.Condition{}
+	for statusType, instances := range states {
+
+		condition := metav1.Condition{
+			Type:               statusType,
+			ObservedGeneration: observedGeneration,
+			Reason:             string(operatorv1.Unknown),
+		}
+		if statusType != string(operatorv1.ComponentAvailable) && len(instances) != 0 {
+			// This condition is true.
+			condition.Status = metav1.ConditionTrue
+			condition.Reason = string(operatorv1.ResourceNotReady)
+			condition.Message = fmt.Sprintf("The following sub-controllers are in this condition: %+v", instances)
+		} else if statusType != string(operatorv1.ComponentAvailable) {
+			// This condition is false.
+			condition.Status = metav1.ConditionFalse
+		} else {
+			// This is the available condition, which is only true if all statuses are available.
+			condition.Type = string(operatorv1.ComponentReady)
+			// Available will be mapped to Ready while storing in ls Conditions. Update the key type to Ready for Available
+			statusType = condition.Type
+			if len(instances) == len(expectedInstances) {
+				condition.Status = metav1.ConditionTrue
+				condition.Reason = string(operatorv1.AllObjectsAvailable)
+				condition.Message = "All sub-controllers are available"
+			} else {
+				condition.Status = metav1.ConditionFalse
+			}
+		}
+		// Store the condition.
+		conditions[statusType] = condition
+	}
+	return conditions, nil
+}
+
+func updateConditions(currentConditions, desiredConditions map[string]metav1.Condition) []metav1.Condition {
+
+	statusConditions := []metav1.Condition{}
+
+	//Update the current log storage condition only when the aggregate desired status have new changes
+	for _, desired := range desiredConditions {
+
+		current, ok := currentConditions[desired.Type]
+		if !ok || current.Status != desired.Status || current.Message != desired.Message {
+			desired.LastTransitionTime = metav1.NewTime(time.Now())
+		} else {
+			desired.LastTransitionTime = current.LastTransitionTime
+		}
+
+		statusConditions = append(statusConditions, desired)
+	}
+	return statusConditions
 }
--- a/pkg/controller/logstorage/initializer/conditions_controller.go
+++ b/pkg/controller/logstorage/initializer/conditions_controller.go
@@ -19,7 +19,6 @@ import (
 	"fmt"
 	"time"
 
-	"github.com/tigera/operator/pkg/controller/logstorage/esmetrics"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -126,7 +125,7 @@ func (r *LogStorageConditions) getDesiredConditions(ctx context.Context) (map[st
 	if r.multiTenant {
 		expectedInstances = append(expectedInstances, TigeraStatusLogStorageUsers)
 	} else {
-		expectedInstances = append(expectedInstances, esmetrics.TigeraStatusLogStorageESMetrics, TigeraStatusLogStorageKubeController)
+		expectedInstances = append(expectedInstances, TigeraStatusLogStorageESMetrics, TigeraStatusLogStorageKubeController)
 	}
 
 	// Keep track of which instances are in which state.
--- a/pkg/controller/logstorage/initializer/conditions_controller_test.go
+++ b/pkg/controller/logstorage/initializer/conditions_controller_test.go
@@ -138,6 +138,17 @@ var _ = Describe("LogStorage Conditions controller", func() {
 		Expect(readyCondition.Message).To(Equal("All sub-controllers are available"))
 		Expect(readyCondition.ObservedGeneration).To(Equal(generation))
 
+		degCondition, ok := actualConditions["Degraded"]
+		Expect(ok).To(BeTrue())
+		Expect(string(degCondition.Status)).To(Equal(string(operatorv1.ConditionFalse)))
+		Expect(degCondition.Reason).To(Equal(string(operatorv1.Unknown)))
+		Expect(degCondition.Message).To(Equal(""))
+		progCondition, ok := actualConditions["Progressing"]
+		Expect(ok).To(BeTrue())
+		Expect(string(progCondition.Status)).To(Equal(string(operatorv1.ConditionFalse)))
+		Expect(progCondition.Reason).To(Equal(string(operatorv1.Unknown)))
+		Expect(progCondition.Message).To(Equal(""))
+
 	})
 
 	It("should reconcile with empty tigerastatus conditions", func() {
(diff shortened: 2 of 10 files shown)


pkg/controller/logstorage/initializer/conditions_controller.go
```diff
@@ -90,18 +92,80 @@ func (r *LogStorageConditions) Reconcile(ctx context.Context, request reconcile.
...
+		}
+	}
+}
+
+func updateAggregatedConditions(tsConditions []operatorv1.TigeraStatusCondition, aggTigeraStatusConditions []operatorv1.TigeraStatusCondition, statusMap map[operatorv1.StatusConditionType]bool, logStorageType string) {
+	for _, condition := range tsConditions {
+		for i := range aggTigeraStatusConditions {
+			if aggTigeraStatusConditions[i].Type == condition.Type {
+				aggTigeraStatusConditions[i].Status = condition.Status
+				aggTigeraStatusConditions[i].Message = fmt.Sprintf("%s%s for %s;", aggTigeraStatusConditions[i].Message, condition.Message, logStorageType)
```
caseydavenport, line 158: I think it would be good if we did something like this:

- Track which conditions are true, and which sub-controllers they are true for.
- Build a Message based on that after iterating all of them.

Then, we can produce a simple message like this:

> "The following sub-controllers are degraded: log-storage-access, log-storage-users"

or 

> "The following-sub-controllers are progressing: log-storage-elastic"

caseydavenport (changes requested):
First pass - It's a bit of a tricky one, but I suggested some changes to the logic to simplify things. I think there are a couple of small bugs regarding how we set certain fields as well.

pkg/controller/logstorage/initializer/conditions_controller.go
```diff
@@ -90,18 +92,80 @@ func (r *LogStorageConditions) Reconcile(ctx context.Context, request reconcile.
...
+			return reconcile.Result{}, err
+		}
+
+		// Update aggregated conditions based on fetched TigeraStatus
+		updateAggregatedConditions(ts.Status.Conditions, aggTigeraStatusConditions, statusMap, logStorage)
+	}
+
+	// Update tigera status conditions
+	switch {
+	case statusMap[operatorv1.ComponentDegraded]:
```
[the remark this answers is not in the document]
vara2504: "Okay, so what could be the combinations?

Available = true, only when aggregated Progressing and Degraded conditions are false.
When Progressing and/or Degraded is true: What if we have Available=true for one of the components? Do we need to set true with the message "the following components are available," or should Available be simply marked as false?"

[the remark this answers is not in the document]
caseydavenport: I just looked back at the semantics that we implement for our other TigeraStatus resources: https://github.com/tigera/operator-cloud/blob/9aabd16a94b6a5c7306cb8e1f7caf85f909b8159/pkg/controller/status/status.go#L377-L429

I believe that it is something like this:

- Degraded if any sub-component/deployment is crashing, or explicitly marked degraded.
- Progressing if any sub-component/deployment is progressing, and none are crashing.
- Available if no sub-components are progressing, not explicitly marked degraded, and nothing is crashing.

For this particular controller, we don't really have the ability to distinguish between "explicitly marked degraded by the controller" and "some pods are crashing", so in practice it may actually be closer to mutually exclusive here.

I think it would be good to aggregate the Degraded and Progressing fields, and then Available will be true if every sub-controller is Available, and false otherwise.

pkg/controller/logstorage/initializer/conditions_controller.go
```diff
@@ -90,18 +92,80 @@ func (r *LogStorageConditions) Reconcile(ctx context.Context, request reconcile.
...
+	}
+}
+
+func updateAggregatedConditions(tsConditions []operatorv1.TigeraStatusCondition, aggTigeraStatusConditions []operatorv1.TigeraStatusCondition, statusMap map[operatorv1.StatusConditionType]bool, logStorageType string) {
+	for _, condition := range tsConditions {
+		for i := range aggTigeraStatusConditions {
+			if aggTigeraStatusConditions[i].Type == condition.Type {
+				aggTigeraStatusConditions[i].Status = condition.Status
+				aggTigeraStatusConditions[i].Message = fmt.Sprintf("%s%s for %s;", aggTigeraStatusConditions[i].Message, condition.Message, logStorageType)
+				if aggTigeraStatusConditions[i].LastTransitionTime.Time.Before(condition.LastTransitionTime.Time) {
```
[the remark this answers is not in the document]
vara2504: Done.
Updating the latest transition time here. However reverting it when there is no change in the object status in resetLastTransitionTimeStamp. So that i can compare the status and message once all the subcontrollers are aggregated.

pkg/controller/logstorage/initializer/conditions_controller.go
```diff
@@ -90,18 +91,116 @@ func (r *LogStorageConditions) Reconcile(ctx context.Context, request reconcile.
...
 	}
 
-	// Update conditions on the LogStorage object based on the status of the various components.
-	// TODO: Incorporate status from other sub controllers as well.
-	ts := &operatorv1.TigeraStatus{}
-	if err := r.client.Get(ctx, types.NamespacedName{Name: TigeraStatusName}, ts); err != nil {
-		return reconcile.Result{}, err
+	// Log storage instances to fetch TigeraStatus
+	logStorageInstances := []string{TigeraStatusName, TigeraStatusLogStorageAccess, TigeraStatusLogStorageElastic, TigeraStatusLogStorageSecrets, TigeraStatusLogStorageUsers}
+
```
vara2504, line 96: Will all the sub-controller's TigeraStatus exist, or is there a possibility that one of them might not exist, causing an error in the line r.client.Get(ctx, controller, ts) at line 110?
In such a case, is it ok to return reconcile.Result{} without proceeding further as in the current logic, or do we need to handle it differently?

caseydavenport: It is possible that any of them might be not found. We probably ideally should treat them differently depending on (1) whether they are optional and (2) whether we expect them to exist given the input configuration. e.g., `LogStorageUsers` only runs for multi-tenant clusters.

caseydavenport: If a status is expected but not found, that is probably "degraded" territory.

vara2504:
@caseydavenport 
1) When the TigeraStatus object contains only 1 item (e.g., Available):
Currently, the LogStorage condition array will also have the same 1 item (e.g., Available). With the aggregation changes, the condition array will have all 3 items, marking the unavailable items (Progressing, Degraded) in TigeraStatus as false.

2) When the TigeraStatus condition is empty:
Currently, the LogStorage condition array will be empty. With the aggregation changes, the condition array will have all 3 items, marking Available and Progressing as false, and Degraded as true, with the following message.
```"The following sub-controllers are degraded:log-storage-access,log-storage-elastic,log-storage-secrets"```

`	  []v1.Condition: [
          {
              Type: "Ready",
              Status: "False",
              ObservedGeneration: 0,
              LastTransitionTime: {
                  Time: 2024-01-19T11:40:10-08:00,
              },
              Reason: "ResourceNotReady",
              Message: "",
          },
          {
              Type: "Progressing",
              Status: "False",
              ObservedGeneration: 0,
              LastTransitionTime: {
                  Time: 2024-01-19T11:40:10-08:00,
              },
              Reason: "ResourceNotReady",
              Message: "",
          },
          {
              Type: "Degraded",
              Status: "True",
              ObservedGeneration: 0,
              LastTransitionTime: {
                  Time: 2024-01-19T11:40:10-08:00,
              },
              Reason: "ResourceNotReady",
              Message: "The following sub-controllers are degraded:log-storage-access,log-storage-elastic,log-storage-secrets",
          },`

File: pkg/controller/logstorage/initializer/conditions_controller.go [hunk 5]
```diff
@@ -90,18 +91,138 @@ func (r *LogStorageConditions) Reconcile(ctx context.Context, request reconcile.
...
+func updateAggregatedConditions(tsConditions []operatorv1.TigeraStatusCondition, aggTigeraStatusConditions []operatorv1.TigeraStatusCondition, statusMap map[operatorv1.StatusConditionType]bool, logStorageType string) {
+	for _, condition := range tsConditions {
+		for i := range aggTigeraStatusConditions {
+			if aggTigeraStatusConditions[i].Type == condition.Type {
+
+				statusMap[condition.Type] = statusMap[condition.Type] || (condition.Status == operatorv1.ConditionTrue)
+				// Available should be marked true when all of the subcontrollers are available.
+				if aggTigeraStatusConditions[i].Type == operatorv1.ComponentAvailable {
+					statusMap[condition.Type] = statusMap[condition.Type] && (condition.Status == operatorv1.ConditionTrue)
+				}
```
vara2504, line 209: Add this additional check since i noticed. LogStorage User had tigerastatus with all conditions set to False.
`log-storage-users     False       False         False       6m58s`

vara2504:
@caseydavenport Incorporate the suggested changes. Please take a look

pkg/controller/logstorage/initializer/conditions_controller.go
```diff
@@ -90,18 +91,272 @@ func (r *LogStorageConditions) Reconcile(ctx context.Context, request reconcile.
...
+	desiredTigeraStatusMap := make(map[operatorv1.StatusConditionType]operatorv1.TigeraStatusCondition)
+
+	logStorageInstances := []string{TigeraStatusName, TigeraStatusLogStorageAccess, TigeraStatusLogStorageElastic, TigeraStatusLogStorageSecrets}
+	if r.multiTenant {
+		logStorageInstances = append(logStorageInstances, TigeraStatusLogStorageUsers)
+	} else {
+		logStorageInstances = append(logStorageInstances, TigeraStatusLogStorageESMetrics, TigeraStatusLogStorageKubeController)
+	}
+
+	for _, logStorage := range logStorageInstances {
```
caseydavenport, line 136: @vara2504 I think this particular function can probably still be simplified. The thing I'd like to avoid is the many layers of helper function that modify the conditions. I think we should be able to factor this so that we build each condition exactly once in once place. Let me know what you think of logic like this:

```go

// Keep track of which instances are in which state.
states := map[string][]string {
  ComponentDegraded: []string{},
  ComponentAvailable: []string{},
  ComponentProgressing: []string{},
}

// We also keep track of the oldest observed generation here so we can add it to the conditions later.
var observedGeneration int64

// Build up the lists of which components are in which state.
for _, instance := range expectedInstances {
  ts := &operatorv1.TigeraStatus{}
  if err := r.client.Get(ctx, namespacedName, ts); errors.IsNotFound(err) {
    // Expected status not found, treat it as an error.
    states[ComponentDegraded] = append(states[ComponentDegraded], instance.Name)
    continue
  } else if err != nil {
    return nil, err
  }

  // Found it - add to states map, which we use to track which TigeraStatuses are in which state(s).
  for _, condition := range ts.Status.Conditions {
    if condition.Status == ConditionTrue {
      states[condition.Status] = append(states[condition.Status], instance.Name)
    }
    if observedGeneration == 0 || condition.ObservedGeneration < observedGeneration {
      observedGeneration = condition.ObservedGeneration
    }
  }
}

// Convert the states map to a set of conditions.
conditions := map[string]metav1.Condition{}
for _, statusType, instances := range states {

  condition := metav1.Condition{
    Type: statusType, 
    ObservedGeneration: observedGeneration,
  }  

  if statusType != ComponentAvailable && len(instances) != 0 {
    // This condition is true.
    condition.Status = operatorv1.ConditionTrue
    conditoin.Reason = operatorv1.NotReady
    condition.Message = fmt.Sprintf("The following sub-controllers are in this condition: %+v", instances)
  } else if statusType != ConditionAvailable {
    // This condition is false.
    condition.Status = operatorv1.ConditionFalse
  } else {
    // This is the available condition, which is only true if all statuses are available.
    if len(instances) == len(expectedInstances) {
      condition.Status = operatorv1.ConditionTrue
      condition.Message = "All sub-controllers are available"
    } else {
      condition.Status = operatorv1.ConditionFalse
    }
  }

  // Store the condition.
  conditions[statusType] = condition
} 
return conditions
```

pkg/controller/logstorage/initializer/conditions_controller.go
```diff
@@ -90,18 +91,272 @@ func (r *LogStorageConditions) Reconcile(ctx context.Context, request reconcile.
...
+
+	return lsConditions
+}
+
+func updateConditions(currentConditions, desiredConditions map[string]metav1.Condition) []metav1.Condition {
+
+	statusCondition := []metav1.Condition{}
+
+	//Update the current log storage condition only when the aggregate desired status have new changes
+	for _, desired := range desiredConditions {
```
caseydavenport, line 230: One of the reasons I suggested using a `map` type here is that it avoids the need for the inner loop here (and also makes it easier to avoid duplicate entries of the same type)

For example:

```go
for conditionType, desired := range desiredConditions {

  // Get the current condition, if it exists.
  current, ok := currentConditions[conditionType]

  // Update the desired state with the correct transition timestamp. 
  // If the current Status doesn't match the desired,
  // or if there is no current Status, use the current time.
  // Otherwise, use the existing transition timestamp.
  if !ok || current.Status != desired.Status {
    desired.LastTransitionTime = time.Now()
  } else {
    desired.LastTransitionTime = current.LastTransitionTime
  }
  statusConditions = append(statusConditions, desired)
}

File: pkg/controller/logstorage/initializer/conditions_controller.go (hunk 6 above)
Comment (caseydavenport) [reply pkg/controller/logstorage/initializer/conditions_controller.go:136]: I think this should remove the need for a number of the helper functions, like:

- transformIntoLogStorageConditions
- setAvailable, setProgressing, setDegraded
- clearAvailable, clearProgressing, clearDegraded
- mergeCondition

As always, I could be missing an edge case though so let me know if this is too simple for what we need!

File: pkg/controller/logstorage/initializer/conditions_controller_test.go [hunk 8]
```diff
@@ -340,6 +358,252 @@ var _ = Describe("LogStorage Conditions controller", func() {
...
+		}
+
+		// set Degraded TigeraStatus for LogStorageUsers
+		createTigeraStatus(cli, ctx, TigeraStatusLogStorageUsers, generation, transitionTime, []operatorv1.TigeraStatusCondition{
+			{
+				Type:               operatorv1.ComponentAvailable,
+				Status:             operatorv1.ConditionFalse,
+				Reason:             string(operatorv1.ResourceNotReady),
+				Message:            "",
+				ObservedGeneration: generation,
```
caseydavenport, line 514: Do we have any tests that handle different ObservedGeneration values for different sub-controllers?

pkg/controller/logstorage/initializer/initializing_controller.go
```diff
@@ -42,9 +42,15 @@ import (
 var log = logf.Log.WithName("controller_logstorage")
 
 const (
-	DefaultElasticsearchStorageClass = "tigera-elasticsearch"
-	TigeraStatusName                 = "log-storage"
-	defaultEckOperatorMemorySetting  = "512Mi"
+	DefaultElasticsearchStorageClass     = "tigera-elasticsearch"
+	TigeraStatusName                     = "log-storage"
+	TigeraStatusLogStorageAccess         = "log-storage-access"
```
caseydavenport, line 47: These look like they are duplicating the constants from the sub-controllers. I think we want to define a single constant for each so that they don't get out-of-sync.

Either these should each be defined in the sub-packages where they are used, or we should remove the constants from those sub-packages as part of this PR and instead use the values from here.

pkg/controller/status/status.go
```diff
@@ -965,7 +965,7 @@ func UpdateStatusCondition(statuscondition []metav1.Condition, conditions []oper
 			if condition.Type == operator.ComponentAvailable && c.Type == string(operator.ComponentReady) ||
 				condition.Type == operator.ComponentDegraded && c.Type == string(operator.ComponentDegraded) ||
 				condition.Type == operator.ComponentProgressing && c.Type == string(operator.ComponentProgressing) {
-				if !reflect.DeepEqual(c.Status, condition.Status) {
+				if string(c.Status) != string(condition.Status) {
```
caseydavenport, line 968: Is this change needed? If so, I'm curious why.

caseydavenport (changes requested):

pkg/controller/logstorage/initializer/conditions_controller.go
```diff
@@ -90,18 +90,128 @@ func (r *LogStorageConditions) Reconcile(ctx context.Context, request reconcile.
...
+			condition.Reason = string(operatorv1.ResourceNotReady)
+			condition.Message = fmt.Sprintf("The following sub-controllers are in this condition: %+v", instances)
+		} else if statusType != string(operatorv1.ComponentAvailable) {
+			// This condition is false.
+			condition.Status = metav1.ConditionFalse
+			condition.Reason = string(operatorv1.ResourceNotReady)
+		} else {
+			// This is the available condition, which is only true if all statuses are available.
+			condition.Type = string(operatorv1.ComponentReady)
+			condition.Reason = string(operatorv1.ResourceNotReady)
```
vara2504, line 183: Reason can not be empty (// +kubebuilder:validation:MinLength=1) hence setting some defaults here.

caseydavenport: Right now this code does defaulting in two places:
- Here
- A few lines below, in the "else" branch

I think we should pick one of them and do all of the defaulting there instead of splitting it.

pkg/controller/logstorage/initializer/conditions_controller.go
```diff
@@ -90,18 +90,128 @@ func (r *LogStorageConditions) Reconcile(ctx context.Context, request reconcile.
...
+		}
+		// Store the condition.
+		conditions[statusType] = condition
+	}
+	return conditions, nil
+}
+
+func updateConditions(currentConditions, desiredConditions map[string]metav1.Condition) []metav1.Condition {
+
+	statusCondition := []metav1.Condition{}
```
caseydavenport, line 202: nit, but this should be `statusConditions` since it's a slice, so we want it to be plural.

pkg/controller/logstorage/initializer/conditions_controller.go
```diff
@@ -90,18 +90,128 @@ func (r *LogStorageConditions) Reconcile(ctx context.Context, request reconcile.
 		return reconcile.Result{}, err
 	}
 
-	// Update conditions on the LogStorage object based on the status of the various components.
-	// TODO: Incorporate status from other sub controllers as well.
-	ts := &operatorv1.TigeraStatus{}
-	if err := r.client.Get(ctx, types.NamespacedName{Name: TigeraStatusName}, ts); err != nil {
+	// Fetch the current status condition for log storage
+	currentCondition := getCurrentConditions(ls.Status.Conditions)
```
caseydavenport, line 94: nit here as well: currentConditions (plural)

pkg/controller/logstorage/initializer/conditions_controller.go
```diff
@@ -90,18 +90,128 @@ func (r *LogStorageConditions) Reconcile(ctx context.Context, request reconcile.
...
 	if err := r.client.Status().Update(ctx, ls); err != nil {
 		log.WithValues("reason", err).Info("Failed to update LogStorage status conditions")
 		return reconcile.Result{}, err
 	}
 
 	return reconcile.Result{}, nil
 }
+
+func getCurrentConditions(lsConditions []metav1.Condition) map[string]metav1.Condition {
+	statusCondition := make(map[string]metav1.Condition)
```
caseydavenport, line 115: and here: maps / slices are typically named plural, so should be `statusConditions`

pkg/controller/logstorage/initializer/conditions_controller_test.go
vara2504: Nope will add it.

pkg/controller/logstorage/initializer/conditions_controller.go
```diff
@@ -90,18 +90,127 @@ func (r *LogStorageConditions) Reconcile(ctx context.Context, request reconcile.
...
+	}
+
+	// Convert the states map to a set of conditions.
+	conditions := map[string]metav1.Condition{}
+	for statusType, instances := range states {
+
+		condition := metav1.Condition{
+			Type:               statusType,
+			ObservedGeneration: observedGeneration,
+			Reason:             string(operatorv1.Unknown),
```
vara2504, line 170: Setting Unknown as a default reason. Because It was straightforward while handling use case with multiple condition set as true.

eg: deg = true(one component), prog =false, available= true(all component).
Then , we would be setting as such in resultant ls storage conditions deg=true, prog=false and available=true.

If we need to set reason for all condition types as AllObjectsAvailable when available=true, we might need to handle deg!=true similarly for progressing.

I can add those check and set all reasons as "AllObjectsAvailable if it is a preferred option.

caseydavenport (approved):

PR #3135, opened 2024-01-18
Fix trusted-bundle conflict in ES secret controller

caseydavenport:
## Description

This controller is a bit complicated, but it tries to manage a few things:

- Creation of KeyPairs for ES and ES related components
- Creation of a trusted bundle to use for ES and ES related components.

It needs to do this differently depending on:

- If this is a single-tenant or multi-tenant system.
- If using internal or external ES

The bug this PR fixes occurred when using external ES in a single-tenant system. In this case, we would attempt to write two trusted bundles into the tigera-elasticsearch Namespace with conflicting contents, resulting in the tigera-ca-bundle ConfigMap oscillating between two states.

We need to make sure that this code:

- Generates only a single trusted bundle in single-tenant mode (regardless of internal / external ES)
- Generates both a trusted bundle for the tigera-elasticsearch NS, as well as for each tenant in multi-tenant mode (only supported with external ES)

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


pkg/controller/logstorage/secrets/secret_controller.go
```diff
@@ -570,7 +570,10 @@ func (c *elasticKeyPairCollection) component(bundle certificatemanagement.Truste
 			// enabled. Instead, it will be replaced with a dummy secret that serves merely to pass ECK's validation checks.
 			rcertificatemanagement.NewKeyPairOption(c.elastic, true, c.elastic != nil && !c.elastic.UseCertificateManagement()),
 		},
-		TrustedBundle: bundle,
+
+		// We don't create a trusted bundle as part of the elastic component - this is just to install the keypair secret.
+		// The trusted bundle will be installed later, and include other upstream certificates as well.
+		TrustedBundle: nil,
```
caseydavenport, line 576: @tmjd I think this was effectively the problem line - we were creating a trusted bundle here that only included the tigera-ca and Linseed certificate.

Then, on  line 316 we would create another trusted bundle that included the tigera-ca, Linseed certificate, as well as many other things - critically, the external ES public certificate.

pkg/controller/logstorage/secrets/secret_controller.go
```diff
@@ -321,78 +334,64 @@ func (r *SecretSubController) Reconcile(ctx context.Context, request reconcile.R
...
+			cert, err := cm.GetCertificate(r.client, render.ManagerInternalTLSSecretName, render.ManagerNamespace)
+			if err != nil && !errors.IsNotFound(err) {
+				return nil, err
+			} else {
+				collection.upstreamCerts = append(collection.upstreamCerts, cert)
+			}
 		}
 	}
 
-	for _, ns := range linseedNamespaces {
```
caseydavenport, line 382: Removed this logic - it only applied to multi-tenant, but was nestled away within code that should only run for internal ES (which doesn't support multi-tenant).

pkg/controller/logstorage/secrets/secret_controller.go
```diff
@@ -321,78 +334,64 @@ func (r *SecretSubController) Reconcile(ctx context.Context, request reconcile.R
...
 
-	linseedNamespaces := []string{render.ElasticsearchNamespace}
-	if r.multiTenant {
-		// For multi-tenant systems, linseed runs in multiple namespaces, so we need to collect the certificates from all namespaces
-		// that have a tenant.
-		var err error
-		linseedNamespaces, err = utils.TenantNamespaces(context.Background(), r.client)
-		if err != nil {
-			return nil, err
+	if !r.multiTenant {
```
caseydavenport, line 357: I think I can remove this check - internal ES (which this branch is meant for) doesn't support multi-tenant clusters.

pkg/controller/logstorage/secrets/secret_controller.go
```diff
@@ -276,23 +267,45 @@ func (r *SecretSubController) Reconcile(ctx context.Context, request reconcile.R
...
-	if kibanaEnabled {
-		// Render the key pair and trusted bundle into the Kibana namespace.
-		if err = hdler.CreateOrUpdateOrDelete(ctx, elasticKeys.kibanaComponent(esTrustedBundle), r.status); err != nil {
+	// Internal ES modes:
+	// - Zero-tenant: everything installed in tigera-elasticsearch/tigera-kibana Namespaces. We need a single trusted bundle in each.
+	// - Single-tenant: everything installed in tigera-elasticsearch/tigera-kibana Namespaces. We need a single trusted bundle in each.
+	//
+	// External ES modes:
+	// - Single-tenant: everything installed in tigera-elasticsearch Namespace. We need a single trusted bundle.
+	// - Multi-tenant: nothing installed in tigera-elasticsearch Namespace, we need a trusted bundle per tenant.
```
caseydavenport, line 279: This is an assumption that I think is true? For multi-tenant external ES, there is no need for a trusted bundle in the tigera-elasticsearch Namespace because we don't run any components there. Hopefully a reviewer can confirm or deny that assertion.

pkg/controller/logstorage/secrets/secret_controller.go
caseydavenport: I've changed it so that the internal ES component only creates KeyPairs in the tiger-elasticsearch Namespace (it still creates a bundle in the tigera-kibana namespace, though).

Then, a single trusted bundle is created shortly after, avoiding having two fighting representations of what should be in it.

caseydavenport: I might be able to simplify this even further by just getting rid of these separate keyPair components and doing it all in one place, but that's a larger refactor.

Most of this logic was added originally because we were testing multi-tenancy against an internal ES. But, since that's not actually a supported mode of operation, a lot of the hoops I had to jump through can now be removed. Hooray!

pkg/controller/logstorage/secrets/secret_controller.go
caseydavenport: @asincu @tmjd would like your confirmation on this one

asincu: I would expect this to be true. Also, for external ES mode we always have mTLS enabled, while for internal ES mode we do not have it enabled. We need to load `tigera-secure-es-http-certs-public` inside the trusted bundle because it contains the CA for external elastic.

pkg/controller/logstorage/secrets/secret_controller.go
```diff
@@ -238,70 +244,71 @@ func (r *SecretSubController) Reconcile(ctx context.Context, request reconcile.R
...
-	// Cluster-scoped certificate manager, used for managing Elasticsearch secrets.
-	clusterCM, err = certificatemanager.Create(r.client, install, r.clusterDomain, common.OperatorNamespace(), certificatemanager.WithLogger(reqLogger))
+	// Create a cluster-scoped certificate manager from the tigera-operator CA, used for signing KeyPairs for use by Elasticsearch.
+	operatorSigner, err = certificatemanager.Create(r.client, install, r.clusterDomain, common.OperatorNamespace(), certificatemanager.WithLogger(reqLogger))
 	if err != nil {
 		r.status.SetDegraded(operatorv1.ResourceReadError, "Error building certificate manager", err, reqLogger)
 		return reconcile.Result{}, err
 	}
-	clusterCM.AddToStatusManager(r.status, render.ElasticsearchNamespace)
+	operatorSigner.AddToStatusManager(r.status, render.ElasticsearchNamespace)
```
asincu, line 255: Does this apply for multi-tenant clusters ?

pkg/controller/logstorage/secrets/secret_controller.go
```diff
@@ -238,70 +244,71 @@ func (r *SecretSubController) Reconcile(ctx context.Context, request reconcile.R
...
+	cm = operatorSigner
+	if r.multiTenant {
+		// Override with a tenant-scoped certificate manager which uses the CA in the tenant's namespace.
+		opts := []certificatemanager.Option{certificatemanager.WithLogger(reqLogger), certificatemanager.WithTenant(tenant)}
+		cm, err = certificatemanager.Create(r.client, install, r.clusterDomain, helper.InstallNamespace(), opts...)
+		if err != nil {
+			r.status.SetDegraded(operatorv1.ResourceReadError, "Error building certificate manager", err, reqLogger)
+			return reconcile.Result{}, err
+		}
+		cm.AddToStatusManager(r.status, render.ElasticsearchNamespace)
```
asincu, line 307: Is the namespace correct here ?

pkg/controller/logstorage/secrets/secret_controller.go
caseydavenport: > Also, for external ES mode we always have mTLS enabled, while for internal ES mode we do not have it enabled. We need to load tigera-secure-es-http-certs-public inside the trusted bundle because it contains the CA for external elastic.

yes, that is already handled in this code later on but I don't this is relevant to this particular comment.

You can see it here: https://github.com/tigera/operator/blob/5cdf4100c68d2dbac70affdfe7f87ab2611be58f/pkg/controller/logstorage/secrets/secret_controller.go#L498-L503

pkg/controller/logstorage/secrets/secret_controller.go
```diff
@@ -154,6 +151,11 @@ func Add(mgr manager.Manager, opts options.AddOptions) error {
 	if err := utils.AddServiceWatch(c, render.LinseedServiceName, helper.InstallNamespace()); err != nil {
 		return fmt.Errorf("log-storage-secrets-controller failed to watch Service: %w", err)
 	}
+	if opts.MultiTenant {
+		if err = utils.AddSecretsWatchWithHandler(c, certificatemanagement.TenantCASecretName, "", &handler.EnqueueRequestForObject{}); err != nil {
```
rene-dekker, line 155: nit: Looks like we already have a convenience func that creates the handler: `utils.AddSecretsWatch`

rene-dekker (approved):
lgtm

pkg/controller/logstorage/secrets/secret_controller.go
caseydavenport: I think it is essentially irrelevant for multi-tenant clusters. It's only relevant when using the CSR API, which we don't ever do for multi-tenant.

pkg/controller/logstorage/secrets/secret_controller.go
caseydavenport: It's effectively ignored, but I'll change it for clarity.

pkg/controller/logstorage/secrets/secret_controller.go
tmjd: I'm surprised you're asking me because I would have thought you would know better than I would.
I'm not aware of a current need for a bundle in multi-tenant mode.

caseydavenport: @tmjd I should! I'm doubting my own memory of things given the state of some of this code... so wanting to double and triple check.

pkg/render/common/test/testing.go
```diff
@@ -253,3 +255,53 @@ func CreateCertSecret(name, namespace string, dnsNames ...string) *corev1.Secret
 		},
 	}
 }
+
+func ExpectBundleContents(bundle *corev1.ConfigMap, secrets ...types.NamespacedName) {
```
rene-dekker, line 259: One thing to note about the certificate bundle is that if you have CA A and A is the signer of B and C. Then if A is already in the bundle, B and C don't need to be added any longer. The Bundle implementation does contain this logic. This helper function is not aware of this and may result in wrong assertions. Therefore, I am not sure it belongs in a generic testing package.

caseydavenport: Yeah, that was something I was aware of - I think it's just up to the test writer to put the correct inputs to this function, and I think that's the way it should work - we shouldn't try to duplicate that logic as part of our test assertions.

caseydavenport: Personally I think it's totally appropriate in a generic testing package, and it does exactly what it says it does - it asserts that the bundle contents contain the expected values.

PR #3136, opened 2024-01-19
Update intrusion detection for multi-tenancy

caseydavenport:
## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/intrusiondetection/intrusiondetection_controller.go
+++ b/pkg/controller/intrusiondetection/intrusiondetection_controller.go
@@ -42,7 +42,6 @@ import (
 	"github.com/tigera/operator/pkg/tls/certificatemanagement"
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
-	storagev1 "k8s.io/api/storage/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -57,7 +56,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/source"
 )
 
-const ResourceName = "intrusion-detection"
+const tigeraStatusName = "intrusion-detection"
 
 var log = logf.Log.WithName("controller_intrusiondetection")
 
@@ -73,11 +72,11 @@ func Add(mgr manager.Manager, opts options.AddOptions) error {
 	dpiAPIReady := &utils.ReadyFlag{}
 	tierWatchReady := &utils.ReadyFlag{}
 
-	// create the reconciler
+	// Create the reconciler
 	reconciler := newReconciler(mgr, opts, licenseAPIReady, dpiAPIReady, tierWatchReady)
 
 	// Create a new controller
-	controller, err := controller.New("intrusiondetection-controller", mgr, controller.Options{Reconciler: reconcile.Reconciler(reconciler)})
+	c, err := controller.New("intrusiondetection-controller", mgr, controller.Options{Reconciler: reconcile.Reconciler(reconciler)})
 	if err != nil {
 		return fmt.Errorf("failed to create intrusiondetection-controller: %v", err)
 	}
@@ -88,13 +87,23 @@ func Add(mgr manager.Manager, opts options.AddOptions) error {
 		return err
 	}
 
-	go utils.WaitToAddLicenseKeyWatch(controller, k8sClient, log, licenseAPIReady)
+	// Make a helper for determining which namespaces to use based on tenancy mode.
+	helper := utils.NewNamespaceHelper(opts.MultiTenant, render.ManagerNamespace, "")
 
-	go utils.WaitToAddResourceWatch(controller, k8sClient, log, dpiAPIReady,
+	// Determine how to handle watch events for cluster-scoped resources. For multi-tenant clusters,
+	// we should update all tenants whenever one changes. For single-tenant clusters, we can just queue the object.
+	var eventHandler handler.EventHandler = &handler.EnqueueRequestForObject{}
+	if opts.MultiTenant {
+		eventHandler = utils.EnqueueAllTenants(mgr.GetClient())
+	}
+
+	go utils.WaitToAddLicenseKeyWatch(c, k8sClient, log, licenseAPIReady)
+
+	go utils.WaitToAddResourceWatch(c, k8sClient, log, dpiAPIReady,
 		[]client.Object{&v3.DeepPacketInspection{TypeMeta: metav1.TypeMeta{Kind: v3.KindDeepPacketInspection}}})
 
-	go utils.WaitToAddTierWatch(networkpolicy.TigeraComponentTierName, controller, k8sClient, log, tierWatchReady)
-	go utils.WaitToAddNetworkPolicyWatches(controller, k8sClient, log, []types.NamespacedName{
+	go utils.WaitToAddTierWatch(networkpolicy.TigeraComponentTierName, c, k8sClient, log, tierWatchReady)
+	go utils.WaitToAddNetworkPolicyWatches(c, k8sClient, log, []types.NamespacedName{
 		{Name: render.IntrusionDetectionControllerPolicyName, Namespace: render.IntrusionDetectionNamespace},
 		{Name: render.IntrusionDetectionInstallerPolicyName, Namespace: render.IntrusionDetectionNamespace},
 		{Name: render.ADAPIPolicyName, Namespace: render.IntrusionDetectionNamespace},
@@ -103,35 +112,27 @@ func Add(mgr manager.Manager, opts options.AddOptions) error {
 		{Name: dpi.DeepPacketInspectionPolicyName, Namespace: dpi.DeepPacketInspectionNamespace},
 	})
 
-	return add(mgr, controller)
-}
-
-// newReconciler returns a new reconcile.Reconciler
-func newReconciler(mgr manager.Manager, opts options.AddOptions, licenseAPIReady *utils.ReadyFlag, dpiAPIReady *utils.ReadyFlag, tierWatchReady *utils.ReadyFlag) reconcile.Reconciler {
-	r := &ReconcileIntrusionDetection{
-		client:          mgr.GetClient(),
-		scheme:          mgr.GetScheme(),
-		provider:        opts.DetectedProvider,
-		status:          status.New(mgr.GetClient(), "intrusion-detection", opts.KubernetesVersion),
-		clusterDomain:   opts.ClusterDomain,
-		licenseAPIReady: licenseAPIReady,
-		dpiAPIReady:     dpiAPIReady,
-		tierWatchReady:  tierWatchReady,
-		usePSP:          opts.UsePSP,
-		elasticExternal: opts.ElasticExternal,
-	}
-	r.status.Run(opts.ShutdownContext)
-	return r
-}
-
-// add adds watches for resources that are available at startup
-func add(mgr manager.Manager, c controller.Controller) error {
-	var err error
-
-	// Watch for changes to primary resource IntrusionDetection
-	err = c.Watch(&source.Kind{Type: &operatorv1.IntrusionDetection{}}, &handler.EnqueueRequestForObject{})
-	if err != nil {
+	// Watch for changes to operator.tigera.io APIs.
+	if err = c.Watch(&source.Kind{Type: &operatorv1.IntrusionDetection{}}, eventHandler); err != nil {
 		return fmt.Errorf("intrusiondetection-controller failed to watch primary resource: %v", err)
+	}
+	if err = c.Watch(&source.Kind{Type: &operatorv1.LogCollector{}}, eventHandler); err != nil {
+		return fmt.Errorf("intrusiondetection-controller failed to watch LogCollector resource: %v", err)
+	}
+	if err = c.Watch(&source.Kind{Type: &operatorv1.ManagementCluster{}}, eventHandler); err != nil {
+		return fmt.Errorf("intrusiondetection-controller failed to watch primary resource: %w", err)
+	}
+	if err = c.Watch(&source.Kind{Type: &operatorv1.Installation{}}, eventHandler); err != nil {
+		return fmt.Errorf("intrusiondetection-controller failed to watch Installation resource: %w", err)
+	}
+	if err = c.Watch(&source.Kind{Type: &operatorv1.APIServer{}}, eventHandler); err != nil {
+		return fmt.Errorf("intrusiondetection-controller failed to watch APIServer resource: %w", err)
+	}
+	if err = c.Watch(&source.Kind{Type: &operatorv1.ImageSet{}}, eventHandler); err != nil {
+		return fmt.Errorf("intrusiondetection-controller failed to watch ImageSet: %w", err)
+	}
+	if err = utils.AddTigeraStatusWatch(c, tigeraStatusName); err != nil {
+		return fmt.Errorf("intrusiondetection-controller failed to watch intrusion-detection Tigerastatus: %w", err)
 	}
 
 	err = c.Watch(&source.Kind{Type: &batchv1.Job{ObjectMeta: metav1.ObjectMeta{
@@ -142,40 +143,8 @@ func add(mgr manager.Manager, c controller.Controller) error {
 		return fmt.Errorf("intrusiondetection-controller failed to watch installer job: %v", err)
 	}
 
-	// Watch for changes to to primary resource LogCollector, to determine if syslog forwarding is
-	// turned on or off.
-	err = c.Watch(&source.Kind{Type: &operatorv1.LogCollector{}}, &handler.EnqueueRequestForObject{})
-	if err != nil {
-		return fmt.Errorf("intrusiondetection-controller failed to watch LogCollector resource: %v", err)
-	}
-
-	// Watch for changes to primary resource ManagementCluster
-	err = c.Watch(&source.Kind{Type: &operatorv1.ManagementCluster{}}, &handler.EnqueueRequestForObject{})
-	if err != nil {
-		return fmt.Errorf("intrusiondetection-controller failed to watch primary resource: %w", err)
-	}
-
-	if err = utils.AddInstallationWatch(c); err != nil {
-		return fmt.Errorf("intrusiondetection-controller failed to watch Installation resource: %v", err)
-	}
-
-	if err = imageset.AddImageSetWatch(c); err != nil {
-		return fmt.Errorf("intrusiondetection-controller failed to watch ImageSet: %w", err)
-	}
-
-	// Watch for changes in storage classes to queue changes if new storage classes may be made available for AD API.
-	if err = c.Watch(&source.Kind{Type: &storagev1.StorageClass{}}, &handler.EnqueueRequestForObject{}); err != nil {
-		return fmt.Errorf("intrusiondetection-controller failed to watch StorageClass resource: %w", err)
-	}
-
-	if err = utils.AddAPIServerWatch(c); err != nil {
-		return fmt.Errorf("intrusiondetection-controller failed to watch APIServer resource: %v", err)
-	}
-
 	for _, secretName := range []string{
-		render.ElasticsearchIntrusionDetectionUserSecret,
 		render.ElasticsearchIntrusionDetectionJobUserSecret,
-		render.ElasticsearchPerformanceHotspotsUserSecret,
 		render.ManagerInternalTLSSecretName,
 		render.NodeTLSSecretName,
 		render.TyphaTLSSecretName,
@@ -183,37 +152,50 @@ func add(mgr manager.Manager, c controller.Controller) error {
 		render.VoltronLinseedPublicCert,
 		certificatemanagement.CASecretName,
 	} {
-		if err = utils.AddSecretsWatch(c, secretName, common.OperatorNamespace()); err != nil {
+		if err = utils.AddSecretsWatch(c, secretName, helper.TruthNamespace()); err != nil {
 			return fmt.Errorf("intrusiondetection-controller failed to watch the Secret resource: %v", err)
 		}
 	}
 
-	if err = utils.AddSecretsWatch(c, render.ManagerInternalTLSSecretName, render.IntrusionDetectionNamespace); err != nil {
+	if err = utils.AddSecretsWatch(c, render.ManagerInternalTLSSecretName, helper.InstallNamespace()); err != nil {
 		return fmt.Errorf("intrusiondetection-controller failed to watch the Secret resource: %v", err)
 	}
 
-	if err = utils.AddSecretsWatch(c, render.TigeraLinseedSecret, render.IntrusionDetectionNamespace); err != nil {
+	if err = utils.AddSecretsWatch(c, render.TigeraLinseedSecret, helper.InstallNamespace()); err != nil {
 		return fmt.Errorf("intrusiondetection-controller failed to watch the Secret resource: %v", err)
 	}
 
-	if err = utils.AddConfigMapWatch(c, relasticsearch.ClusterConfigConfigMapName, common.OperatorNamespace(), &handler.EnqueueRequestForObject{}); err != nil {
+	if err = utils.AddConfigMapWatch(c, relasticsearch.ClusterConfigConfigMapName, helper.TruthNamespace(), eventHandler); err != nil {
 		return fmt.Errorf("intrusiondetection-controller failed to watch the ConfigMap resource: %v", err)
 	}
 
-	if err = utils.AddConfigMapWatch(c, render.ECKLicenseConfigMapName, render.ECKOperatorNamespace, &handler.EnqueueRequestForObject{}); err != nil {
+	if err = utils.AddConfigMapWatch(c, render.ECKLicenseConfigMapName, render.ECKOperatorNamespace, eventHandler); err != nil {
 		return fmt.Errorf("intrusiondetection-controller failed to watch the ConfigMap resource: %v", err)
 	}
 
-	if err = utils.AddConfigMapWatch(c, render.TyphaCAConfigMapName, common.OperatorNamespace(), &handler.EnqueueRequestForObject{}); err != nil {
+	if err = utils.AddConfigMapWatch(c, render.TyphaCAConfigMapName, helper.TruthNamespace(), eventHandler); err != nil {
 		return fmt.Errorf("intrusiondetection-controller failed to watch the ConfigMap resource: %v", err)
-	}
-
-	// Watch for changes to TigeraStatus.
-	if err = utils.AddTigeraStatusWatch(c, ResourceName); err != nil {
-		return fmt.Errorf("intrusiondetection-controller failed to watch intrusion-detection Tigerastatus: %w", err)
 	}
 
 	return nil
+}
+
+// newReconciler returns a new reconcile.Reconciler
+func newReconciler(mgr manager.Manager, opts options.AddOptions, licenseAPIReady *utils.ReadyFlag, dpiAPIReady *utils.ReadyFlag, tierWatchReady *utils.ReadyFlag) reconcile.Reconciler {
+	r := &ReconcileIntrusionDetection{
+		client:          mgr.GetClient(),
+		scheme:          mgr.GetScheme(),
+		provider:        opts.DetectedProvider,
+		status:          status.New(mgr.GetClient(), tigeraStatusName, opts.KubernetesVersion),
+		clusterDomain:   opts.ClusterDomain,
+		licenseAPIReady: licenseAPIReady,
+		dpiAPIReady:     dpiAPIReady,
+		tierWatchReady:  tierWatchReady,
+		multiTenant:     opts.MultiTenant,
+		elasticExternal: opts.ElasticExternal,
+	}
+	r.status.Run(opts.ShutdownContext)
+	return r
 }
 
 // blank assignment to verify that ReconcileIntrusionDetection implements reconcile.Reconciler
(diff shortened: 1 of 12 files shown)


pkg/controller/intrusiondetection/intrusiondetection_controller.go
```diff
@@ -231,28 +213,54 @@ type ReconcileIntrusionDetection struct {
 	licenseAPIReady *utils.ReadyFlag
 	dpiAPIReady     *utils.ReadyFlag
 	tierWatchReady  *utils.ReadyFlag
-	usePSP          bool
+	multiTenant     bool
 	elasticExternal bool
 }
 
+func GetIntrusionDetection(ctx context.Context, cli client.Client, mt bool, ns string) (*operatorv1.IntrusionDetection, error) {
```
Brian-McM, line 220: Is this used outside of this file? If it is, it might be better in the utils or some other library package (and if not it probably doesn't need to be exported.

pkg/controller/intrusiondetection/intrusiondetection_controller.go
```diff
@@ -231,28 +213,54 @@ type ReconcileIntrusionDetection struct {
...
+	reqLogger := log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name, "installNS", helper.InstallNamespace(), "truthNS", helper.TruthNamespace())
 	reqLogger.Info("Reconciling IntrusionDetection")
 
+	// We skip requests without a namespace specified in multi-tenant setups.
+	if r.multiTenant && request.Namespace == "" {
+		return reconcile.Result{}, nil
+	}
+
+	// Check if this is a tenant-scoped request.
+	tenant, _, err := utils.GetTenant(ctx, r.multiTenant, r.client, request.Namespace)
```
Brian-McM, line 250: I'm a little behind on multi tenancy, but how does this work if a controller needs to reconcile when something changes in a central namespace, like the operator namespace (which I didn't think was multi tenant).

pkg/controller/intrusiondetection/intrusiondetection_controller.go
```diff
@@ -103,119 +112,92 @@ func Add(mgr manager.Manager, opts options.AddOptions) error {
...
-	// Watch for changes in storage classes to queue changes if new storage classes may be made available for AD API.
-	if err = c.Watch(&source.Kind{Type: &storagev1.StorageClass{}}, &handler.EnqueueRequestForObject{}); err != nil {
-		return fmt.Errorf("intrusiondetection-controller failed to watch StorageClass resource: %w", err)
+	if err = utils.AddTigeraStatusWatch(c, tigeraStatusName); err != nil {
+		return fmt.Errorf("intrusiondetection-controller failed to watch intrusion-detection Tigerastatus: %w", err)
 	}
 
-	if err = utils.AddAPIServerWatch(c); err != nil {
-		return fmt.Errorf("intrusiondetection-controller failed to watch APIServer resource: %v", err)
+	err = c.Watch(&source.Kind{Type: &batchv1.Job{ObjectMeta: metav1.ObjectMeta{
```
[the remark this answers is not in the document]
caseydavenport: Yeah, I think that pattern is not appropriate for these. The only reason "AddSecretsWatch" is useful is because it does stuff with predicates, which we woudln't be doing here. I am mostly against adding helpers that don't save us any lines of code or repeated complexity.

In other places, where we already had these helper functions, it made the update to multi-tenancy a lot more tedious than it needed to be if they weren't there.

pkg/controller/intrusiondetection/intrusiondetection_controller.go
caseydavenport: The pattern we have established for other things (I think) is that these live within the controller that owns that CR. Pretty sure it doesn't need to be exported though

pkg/controller/intrusiondetection/intrusiondetection_controller.go
caseydavenport: In the `Add` function, we use a special even handler called `EneuqueAllTenants` for those resources.

The `Reconcile()` function will be called with the `Tenant` CR instead of the object that changed.

pkg/render/intrusion_detection.go
```diff
@@ -1385,12 +1078,12 @@ func (c *intrusionDetectionComponent) intrusionDetectionControllerAllowTigeraPol
 		egressRules = append(egressRules, v3.Rule{
 			Action:      v3.Allow,
 			Protocol:    &networkpolicy.TCPProtocol,
-			Destination: networkpolicy.DefaultHelper().ESGatewayEntityRule(),
+			Destination: networkpolicy.Helper(c.cfg.Tenant.MultiTenant(), c.cfg.Namespace).ESGatewayEntityRule(),
```
Josh-Tigera, line 1081: Instead of instantiating a new instance of a helper per rule we can just instantiate a single one and re-use it for each rule. Probably paranoid but may avoid bugs if the rules are all using the same helper vs unique instances.

pkg/render/intrusion_detection.go
```diff
@@ -1453,22 +1146,204 @@ func (c *intrusionDetectionComponent) intrusionDetectionElasticsearchAllowTigera
...
-			Namespace: IntrusionDetectionNamespace,
+// adComponentsToDelete returns a list of objects to delete. Anomaly detection used to be installed here,
+// but has since been removed. This function is kept around to clean up any old objects that may be left.
+func (c intrusionDetectionComponent) adComponentsToDelete() []client.Object {
+	objs := []client.Object{
+		&corev1.ServiceAccount{
+			TypeMeta: metav1.TypeMeta{Kind: "ServiceAccount", APIVersion: "v1"},
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      ADAPIObjectName,
+				Namespace: c.cfg.Namespace,
```
Josh-Tigera, line 1157: Since AD is old and deprecated should these namespaces stay hardcoded to the one where they were installed originally? Seems unlikely but also doesn't make sense to me to have a namespace referenced here where they may have never existed anyway.

pkg/controller/intrusiondetection/intrusiondetection_controller.go
```diff
@@ -378,70 +389,61 @@ func (r *ReconcileIntrusionDetection) Reconcile(ctx context.Context, request rec
...
-		if esClusterConfig.ClusterName() == render.DefaultElasticsearchClusterName {
-			msg := fmt.Sprintf("%s/%s ConfigMap must contain a 'clusterName' field that is not '%s'", common.OperatorNamespace(), relasticsearch.ClusterConfigConfigMapName, render.DefaultElasticsearchClusterName)
-			err = fmt.Errorf("Elasticsearch cluster name must be non-default value in managed clusters")
-			r.status.SetDegraded(operatorv1.InvalidConfigurationError, msg, err, reqLogger)
+	// For now, these are only needed in single-tenant mode, since for multi-tenant we don't install Kibana.
+	var esgwCertificate certificatemanagement.CertificateInterface
+	var esClusterConfig *relasticsearch.ClusterConfig
+	var esSecrets []*corev1.Secret
+	if !r.multiTenant && !isManagedCluster {
+		esClusterConfig, err = utils.GetElasticsearchClusterConfig(context.Background(), r.client)
```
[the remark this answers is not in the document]
caseydavenport: I think this is only needed when there are components that don't go via Linseed - in the case of IDS, the only remaining component is the Kibana dashboard installer which is only installed in management clusters.

My other PR has switched the intrusion detection controller over to Linseed, so this configmap isn't needed any more. I think? Unless I'm forgetting a case?

pkg/controller/intrusiondetection/intrusiondetection_controller.go
```diff
@@ -378,70 +389,61 @@ func (r *ReconcileIntrusionDetection) Reconcile(ctx context.Context, request rec
...
-	secrets := []string{
-		render.ElasticsearchIntrusionDetectionUserSecret,
-		render.ElasticsearchPerformanceHotspotsUserSecret,
-	}
+		secrets := []string{render.ElasticsearchIntrusionDetectionJobUserSecret}
 
-	if !isManagedCluster {
-		secrets = append(secrets, render.ElasticsearchIntrusionDetectionJobUserSecret)
-	}
+		esSecrets, err = utils.ElasticsearchSecrets(context.Background(), secrets, r.client)
```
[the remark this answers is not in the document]
caseydavenport: Yep, probably some more cleanup I need to do - good catch. 

IDC shouldn't need these secrets any more, as everything goes via Linseed.

pkg/controller/intrusiondetection/intrusiondetection_controller.go
```diff
@@ -378,70 +389,61 @@ func (r *ReconcileIntrusionDetection) Reconcile(ctx context.Context, request rec
...
 
-	esSecrets, err := utils.ElasticsearchSecrets(
-		context.Background(),
-		secrets,
-		r.client,
-	)
-	if err != nil {
-		if errors.IsNotFound(err) {
-			r.status.SetDegraded(operatorv1.ResourceNotFound, "Elasticsearch secrets are not available yet, waiting until they become available", err, reqLogger)
+		esgwCertificate, err = certificateManager.GetCertificate(r.client, relasticsearch.PublicCertSecret, helper.TruthNamespace())
```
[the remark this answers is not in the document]
caseydavenport: IDC should no longer need any ESGW secrets as part of my recent work to switch it to Linseed.

pkg/controller/intrusiondetection/intrusiondetection_controller.go
```diff
@@ -479,18 +483,35 @@ func (r *ReconcileIntrusionDetection) Reconcile(ctx context.Context, request rec
...
+		bundleMaker = nil
 	}
 
 	// Create a component handler to manage the rendered component.
 	handler := utils.NewComponentHandler(log, r.client, r.scheme, instance)
 
+	// Determine the namespaces to which we must bind the cluster role.
+	namespaces, err := helper.TenantNamespaces(r.client)
+	if err != nil {
+		return reconcile.Result{}, err
```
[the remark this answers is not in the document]
caseydavenport: Yep, good catch.

pkg/render/intrusion_detection.go
```diff
@@ -830,13 +736,6 @@ func (c *intrusionDetectionComponent) intrusionDetectionControllerContainer() co
 			Name:  "LINSEED_TOKEN",
 			Value: GetLinseedTokenPath(c.cfg.ManagedCluster),
 		},
-		relasticsearch.ElasticIndexSuffixEnvVar(c.cfg.ESClusterConfig.ClusterName()),
```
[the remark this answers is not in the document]
caseydavenport: Yep!

pkg/render/intrusion_detection.go
caseydavenport: Yep, they should stay hardcoded I think. Strictly speaking the behavior should be the same, but it's more clear if we just leave them hard-coded. This was a find / replace error I think.

pkg/controller/intrusiondetection/intrusiondetection_controller.go
```diff
@@ -509,79 +530,87 @@ func (r *ReconcileIntrusionDetection) Reconcile(ctx context.Context, request rec
...
-		PullSecrets:        pullSecrets,
-		Openshift:          r.provider == operatorv1.ProviderOpenShift,
-		ManagedCluster:     isManagedCluster,
-		ManagementCluster:  isManagementCluster,
-		HasNoLicense:       hasNoLicense,
-		HasNoDPIResource:   hasNoDPIResource,
-		ClusterDomain:      r.clusterDomain,
-		DPICertSecret:      dpiKeyPair,
-	})
+	if !hasNoDPIResource && r.multiTenant {
```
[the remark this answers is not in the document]
caseydavenport: The DeepPacketInspection resource doesn't go via our API server, so we don't have an opportunity to check things like that until we get to this code.

pkg/controller/intrusiondetection/intrusiondetection_controller.go
```diff
@@ -88,134 +88,121 @@ func Add(mgr manager.Manager, opts options.AddOptions) error {
 		return err
 	}
 
-	go utils.WaitToAddLicenseKeyWatch(controller, k8sClient, log, licenseAPIReady)
+	// Make a helper for determining which namespaces to use based on tenancy mode.
+	helper := utils.NewNamespaceHelper(opts.MultiTenant, render.ManagerNamespace, "")
 
-	go utils.WaitToAddResourceWatch(controller, k8sClient, log, dpiAPIReady,
-		[]client.Object{&v3.DeepPacketInspection{TypeMeta: metav1.TypeMeta{Kind: v3.KindDeepPacketInspection}}})
+	installNS, _, _ := tenancy.GetWatchNamespaces(opts.MultiTenant, render.IntrusionDetectionNamespace)
```
Brian-McM, line 94: If this is in multi tenant mode, then this value seems to be empty. What does that mean in terms of the policy / job watches? Wouldn't this make it so you're technically watching across all namespaces for these policies / jobs (which would still be filtered by name).

Future improvements are going to improve our caching mechanisms to cache by the namespace for namespace resources so we don't load cache resources from other namespaces. If we don't have the specific namespace we want to watch for the resource in, we can't achieve this and we'll continue to load all resources into the cache.

caseydavenport: Yes, in multi-tenant mode we currently watch all namespaces. This is because we don't know the total set of namespaces that we need to watch at startup.

if / when we make that change, we will need to handle this case appropriately - either by omitting multi-tenancy from that improvement (which would be sad) or perhaps by implementing a more dynamic watch system that enables us to add / remove namespaces to be watched as tenants in the cluster come and go.

For now, this is the pattern that we're already using throughout the rest of the code.

Brian-McM: This is unfortunate, but I suppose I understand the difficulty.

pkg/controller/intrusiondetection/intrusiondetection_controller.go
```diff
@@ -88,134 +88,121 @@ func Add(mgr manager.Manager, opts options.AddOptions) error {
 		return err
 	}
 
-	go utils.WaitToAddLicenseKeyWatch(controller, k8sClient, log, licenseAPIReady)
+	// Make a helper for determining which namespaces to use based on tenancy mode.
+	helper := utils.NewNamespaceHelper(opts.MultiTenant, render.ManagerNamespace, "")
```
Josh-Tigera, line 92: This should be `render.IntrusionDetectionNamespace` instead of `render.ManagerNamespace` right (copy-paste error)? Looking at this and the next line I wonder if it would make sense to just make `GetWatchNamespaces` a method of `NamespaceHelper` rather than an entirely separate package and function.

pkg/controller/intrusiondetection/intrusiondetection_controller.go
```diff
@@ -88,134 +88,121 @@ func Add(mgr manager.Manager, opts options.AddOptions) error {
...
+	go utils.WaitToAddNetworkPolicyWatches(c, k8sClient, log, policiesToWatch)
+	go utils.WaitToAddLicenseKeyWatch(c, k8sClient, log, licenseAPIReady)
+	go utils.WaitToAddTierWatch(networkpolicy.TigeraComponentTierName, c, k8sClient, log, tierWatchReady)
 
-	// Watch for changes to to primary resource LogCollector, to determine if syslog forwarding is
-	// turned on or off.
-	err = c.Watch(&source.Kind{Type: &operatorv1.LogCollector{}}, &handler.EnqueueRequestForObject{})
-	if err != nil {
+	// Watch for changes to operator.tigera.io APIs.
+	if err = c.Watch(&source.Kind{Type: &operatorv1.IntrusionDetection{}}, eventHandler); err != nil {
```
Josh-Tigera, line 121: I think we want this to be `EnqueueRequestForObject` (I remember we talked about this case specifically on my compliance PR) so that we only reconcile the tenant namespace in which the `IntrusionDetection` CR was modified rather than all other tenants.

pkg/controller/intrusiondetection/intrusiondetection_controller.go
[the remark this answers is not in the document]
Josh-Tigera: Ok yeah if Linseed has its own way of injecting differentiating info definitely fine to remove this.

pkg/controller/intrusiondetection/intrusiondetection_controller.go
caseydavenport: Yep, good catch

pkg/controller/intrusiondetection/intrusiondetection_controller.go
caseydavenport: tbh, I feel like I originally had the GetWatchNamespaces on the helper and then changed it to be a separate function for some reason... I can't recall why though.

pkg/controller/intrusiondetection/intrusiondetection_controller.go
```diff
@@ -88,134 +88,118 @@ func Add(mgr manager.Manager, opts options.AddOptions) error {
...
-	var err error
 
-	// Watch for changes to primary resource IntrusionDetection
-	err = c.Watch(&source.Kind{Type: &operatorv1.IntrusionDetection{}}, &handler.EnqueueRequestForObject{})
-	if err != nil {
-		return fmt.Errorf("intrusiondetection-controller failed to watch primary resource: %v", err)
+	policiesToWatch := []types.NamespacedName{
+		{Name: render.IntrusionDetectionControllerPolicyName, Namespace: installNS},
+		{Name: render.IntrusionDetectionInstallerPolicyName, Namespace: installNS},
+		{Name: render.ADAPIPolicyName, Namespace: installNS},
```
asincu, line 103: Are the AD policies still needed ?

pkg/controller/intrusiondetection/intrusiondetection_controller.go
```diff
@@ -88,134 +88,118 @@ func Add(mgr manager.Manager, opts options.AddOptions) error {
...
-func add(mgr manager.Manager, c controller.Controller) error {
-	var err error
 
-	// Watch for changes to primary resource IntrusionDetection
-	err = c.Watch(&source.Kind{Type: &operatorv1.IntrusionDetection{}}, &handler.EnqueueRequestForObject{})
-	if err != nil {
-		return fmt.Errorf("intrusiondetection-controller failed to watch primary resource: %v", err)
+	policiesToWatch := []types.NamespacedName{
+		{Name: render.IntrusionDetectionControllerPolicyName, Namespace: installNS},
+		{Name: render.IntrusionDetectionInstallerPolicyName, Namespace: installNS},
```
asincu, line 102: Do we still need to watch for the installer policy ?

pkg/controller/intrusiondetection/intrusiondetection_controller.go
```diff
@@ -88,134 +88,118 @@ func Add(mgr manager.Manager, opts options.AddOptions) error {
...
+		Namespace: installNS,
+		Name:      render.IntrusionDetectionInstallerJobName,
+	}}}, &handler.EnqueueRequestForObject{})
+	if err != nil {
+		return fmt.Errorf("intrusiondetection-controller failed to watch installer job: %v", err)
 	}
 
 	for _, secretName := range []string{
-		render.ElasticsearchIntrusionDetectionUserSecret,
 		render.ElasticsearchIntrusionDetectionJobUserSecret,
```
asincu, line 149: Not sure if we also need to watch for this secret.

pkg/controller/intrusiondetection/intrusiondetection_controller.go
```diff
@@ -88,134 +88,118 @@ func Add(mgr manager.Manager, opts options.AddOptions) error {
...
 		return fmt.Errorf("intrusiondetection-controller failed to watch the Secret resource: %v", err)
 	}
 
-	if err = utils.AddConfigMapWatch(c, relasticsearch.ClusterConfigConfigMapName, common.OperatorNamespace(), &handler.EnqueueRequestForObject{}); err != nil {
+	if err = utils.AddConfigMapWatch(c, relasticsearch.ClusterConfigConfigMapName, truthNS, eventHandler); err != nil {
 		return fmt.Errorf("intrusiondetection-controller failed to watch the ConfigMap resource: %v", err)
 	}
 
-	if err = utils.AddConfigMapWatch(c, render.ECKLicenseConfigMapName, render.ECKOperatorNamespace, &handler.EnqueueRequestForObject{}); err != nil {
+	if err = utils.AddConfigMapWatch(c, render.ECKLicenseConfigMapName, render.ECKOperatorNamespace, eventHandler); err != nil {
```
asincu, line 174: Why do need to check for ECK license ?

pkg/controller/intrusiondetection/intrusiondetection_controller.go
```diff
@@ -231,28 +215,54 @@ type ReconcileIntrusionDetection struct {
 	licenseAPIReady *utils.ReadyFlag
 	dpiAPIReady     *utils.ReadyFlag
 	tierWatchReady  *utils.ReadyFlag
-	usePSP          bool
+	multiTenant     bool
```
asincu, line 218: Why are we removing usePSP ?

pkg/controller/intrusiondetection/intrusiondetection_controller.go
```diff
@@ -231,28 +215,54 @@ type ReconcileIntrusionDetection struct {
...
 	elasticExternal bool
 }
 
+func getIntrusionDetection(ctx context.Context, cli client.Client, mt bool, ns string) (*operatorv1.IntrusionDetection, error) {
+	key := client.ObjectKey{Name: "tigera-secure"}
+	if mt {
+		key.Namespace = ns
```
asincu, line 225: We also need to make some changes to make the CRD Namespaced instead of Global for crd helm chart.

pkg/controller/intrusiondetection/intrusiondetection_controller.go
```diff
@@ -378,70 +386,61 @@ func (r *ReconcileIntrusionDetection) Reconcile(ctx context.Context, request rec
...
 	}
 
-	if isManagedCluster {
-		if esClusterConfig.ClusterName() == render.DefaultElasticsearchClusterName {
-			msg := fmt.Sprintf("%s/%s ConfigMap must contain a 'clusterName' field that is not '%s'", common.OperatorNamespace(), relasticsearch.ClusterConfigConfigMapName, render.DefaultElasticsearchClusterName)
-			err = fmt.Errorf("Elasticsearch cluster name must be non-default value in managed clusters")
-			r.status.SetDegraded(operatorv1.InvalidConfigurationError, msg, err, reqLogger)
+	// For now, these are only needed in single-tenant mode, since for multi-tenant we don't install Kibana.
+	var esgwCertificate certificatemanagement.CertificateInterface
+	var esClusterConfig *relasticsearch.ClusterConfig
```
asincu, line 402: Is this still needed at the moment ?

pkg/controller/intrusiondetection/intrusiondetection_controller.go
```diff
@@ -378,70 +386,61 @@ func (r *ReconcileIntrusionDetection) Reconcile(ctx context.Context, request rec
...
 		return reconcile.Result{}, err
 	}
 
-	if isManagedCluster {
-		if esClusterConfig.ClusterName() == render.DefaultElasticsearchClusterName {
-			msg := fmt.Sprintf("%s/%s ConfigMap must contain a 'clusterName' field that is not '%s'", common.OperatorNamespace(), relasticsearch.ClusterConfigConfigMapName, render.DefaultElasticsearchClusterName)
-			err = fmt.Errorf("Elasticsearch cluster name must be non-default value in managed clusters")
-			r.status.SetDegraded(operatorv1.InvalidConfigurationError, msg, err, reqLogger)
+	// For now, these are only needed in single-tenant mode, since for multi-tenant we don't install Kibana.
+	var esgwCertificate certificatemanagement.CertificateInterface
```
asincu, line 401: Do we still need this at the moment ?

pkg/controller/intrusiondetection/intrusiondetection_controller.go
```diff
@@ -378,70 +386,61 @@ func (r *ReconcileIntrusionDetection) Reconcile(ctx context.Context, request rec
...
+			r.status.SetDegraded(operatorv1.ResourceReadError, "Failed to get the elasticsearch cluster configuration", err, reqLogger)
 			return reconcile.Result{}, err
 		}
-	}
 
-	secrets := []string{
-		render.ElasticsearchIntrusionDetectionUserSecret,
-		render.ElasticsearchPerformanceHotspotsUserSecret,
-	}
+		secrets := []string{render.ElasticsearchIntrusionDetectionJobUserSecret}
```
asincu, line 415: I think we do not need this secret anymore.

pkg/controller/intrusiondetection/intrusiondetection_controller.go
```diff
@@ -509,79 +528,87 @@ func (r *ReconcileIntrusionDetection) Reconcile(ctx context.Context, request rec
...
-		UsePSP:                       r.usePSP,
+		Namespace:                    helper.InstallNamespace(),
+		BindNamespaces:               namespaces,
+		Tenant:                       tenant,
 	}
-	comp := render.IntrusionDetection(intrusionDetectionCfg)
+	intrusionDetectionComponent := render.IntrusionDetection(intrusionDetectionCfg)
 
-	if err = imageset.ApplyImageSet(ctx, r.client, variant, comp); err != nil {
+	if err = imageset.ApplyImageSet(ctx, r.client, variant, intrusionDetectionComponent); err != nil {
```
asincu, line 537: I have never thought about this, but at the moment, if we define an ImageSet, we apply all those images to all tenants, right ? Does this need to be per tenant ?

pkg/render/intrusion_detection.go
```diff
@@ -177,14 +156,23 @@ func (c *intrusionDetectionComponent) Objects() ([]client.Object, []client.Objec
...
-		CreateNamespace(IntrusionDetectionNamespace, c.cfg.Installation.KubernetesProvider, PodSecurityStandard(pss)),
-		c.intrusionDetectionControllerAllowTigeraPolicy(),
-		networkpolicy.AllowTigeraDefaultDeny(IntrusionDetectionNamespace),
+	objs := []client.Object{}
+	if !c.cfg.Tenant.MultiTenant() {
+		// In multi-tenant environments, the namespace is pre-created. So, only create it if we're not in a multi-tenant environment.
+		objs = append(objs, CreateNamespace(c.cfg.Namespace, c.cfg.Installation.KubernetesProvider, PodSecurityStandard(pss)))
+
+		// ES Secrets are only used for the Kibana dashboard installer Job, which is not run in multi-tenant mode.
+		objs = append(objs, secret.ToRuntimeObjects(secret.CopyToNamespace(c.cfg.Namespace, c.cfg.ESSecrets...)...)...)
```
asincu, line 165: Are these still needed ? We are actually not using esgateway and we do not need the users to be copied over.

pkg/render/intrusion_detection.go
```diff
@@ -177,14 +156,23 @@ func (c *intrusionDetectionComponent) Objects() ([]client.Object, []client.Objec
...
+
+		// GlobalAlertTemplates are not used in multi-tenant management clusters.
+		objs = append(objs, c.globalAlertTemplates()...)
```
asincu, line 168: Do we also need to disable this feature from the UI for multi-tenancy ? :laughing:

pkg/render/intrusion_detection.go
```diff
@@ -1275,12 +972,12 @@ func (c *intrusionDetectionComponent) intrusionDetectionControllerAllowTigeraPol
 		egressRules = append(egressRules, v3.Rule{
 			Action:      v3.Allow,
 			Protocol:    &networkpolicy.TCPProtocol,
-			Destination: networkpolicy.DefaultHelper().ESGatewayEntityRule(),
+			Destination: helper.ESGatewayEntityRule(),
```
asincu, line 975: Should we also remove this ?

pkg/render/intrusiondetection/dpi/dpi.go
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2021-2023 Tigera, Inc. All rights reserved.
```
asincu, line 1: There are no changes to this file it seems.

pkg/render/intrusiondetection/dpi/dpi_test.go
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2021-2023 Tigera, Inc. All rights reserved.
```
asincu, line 1: There are no changes to this file it seems.

pkg/controller/intrusiondetection/intrusiondetection_controller.go
caseydavenport: PodSecurityPolicies have been deprecated since k8s 1.21 and removed since v1.25, and we now have supported their replacement (PodSecurityStandards) for several releases, so I removed it rather than refactor it for multi-tenancy.

pkg/controller/intrusiondetection/intrusiondetection_controller.go
caseydavenport: Nope, forgot to remove this after rebasing on your PR (we needed it before that)

pkg/controller/intrusiondetection/intrusiondetection_controller.go
caseydavenport: Maybe, although I don't think we ever use ImageSets for multi-tenancy right now. Probably one to consider later.

pkg/render/intrusion_detection.go
caseydavenport: Nope, don't think so - this is still supported in managed clusters, just not on the management cluster itself.

pkg/render/intrusiondetection/dpi/dpi.go
caseydavenport: Yeah, but the CI was complaining about it so I had to change it. I'll try reverting...

caseydavenport: I think I just need to leave it, shouldn't be any harm. The CI and pre-commit hooks won't let me adjust this.

asincu (approved):

PR #3137, opened 2024-01-20
Migrate job installer to run inside elasticsearch namespace

asincu:
## Description

Migrate intrusion detection installer to a job inside elasticsearch namespace that access Kibana directly.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


controllers/logstorage_controller.go
```diff
@@ -84,6 +85,11 @@ func (r *LogStorageReconciler) SetupWithManager(mgr ctrl.Manager, opts options.A
 		return err
 	}
 
+	// The settings controller installs
```
caseydavenport, line 88: typo: comment seems incomplete, and refers to the wrong controller.

pkg/controller/logstorage/dashboards/dashboards_controller.go
```diff
@@ -0,0 +1,376 @@
...
+	usePSP          bool
+	multiTenant     bool
+	elasticExternal bool
+	tierWatchReady  *utils.ReadyFlag
+}
+
+func Add(mgr manager.Manager, opts options.AddOptions) error {
+	if !opts.EnterpriseCRDExists {
+		return nil
+	}
```
caseydavenport, line 74: I think for the moment, we want to skip adding this controller for multi-tenant systems (since they don't support Kibana yet)

pkg/controller/logstorage/dashboards/dashboards_controller.go
```diff
@@ -0,0 +1,376 @@
...
+			if err != nil {
+				reqLogger.Error(err, "Failed to read external Kibana client certificate secret")
+				d.status.SetDegraded(operatorv1.ResourceReadError, "Waiting for external Kibana client certificate secret to be available", err, reqLogger)
+				return reconcile.Result{}, err
+			}
+		}
+	}
+
+	// Query the username and password this Dashboards Installer instance should use to authenticate with Elasticsearch.
+	// ALINA : To check the statement below
```
caseydavenport, line 307: ALINA comment left over :) 

Your comments below are correct for Linseed, but we don't have any logic in the operator to create a user for the dashboard installer job. I think that means we either need to add logic to the operator to create such a secret, or piggy-back on Linseed's.

pkg/render/logstorage.go
```diff
@@ -1760,6 +1760,11 @@ func (es *elasticsearchComponent) kibanaAllowTigeraPolicy() *v3.NetworkPolicy {
 			Protocol:    &networkpolicy.TCPProtocol,
 			Destination: networkpolicy.DefaultHelper().ESGatewayEntityRule(),
 		},
+		{
```
caseydavenport, line 1763: Does it really need egress access to the Job?

pkg/render/logstorage/dashboards/dashboards.go
```diff
@@ -0,0 +1,379 @@
...
+	}
+}
+
+type dashboards struct {
+	image    string
+	csrImage string
+	cfg      *Config
+
+	// Namespace in which to provision the namespaced resources.
+	namespace string
```
caseydavenport, line 61: I don't think we need this field - it's accessible via `c.cfg.Namespace` so we should just use that one, no?

pkg/render/logstorage/dashboards/dashboards.go
```diff
@@ -0,0 +1,379 @@
...
+	Installation *operatorv1.InstallationSpec
+
+	// Pull secrets provided by the user.
+	PullSecrets []*corev1.Secret
+
+	// Trusted bundle to use when validating client certificates.
+	TrustedBundle certificatemanagement.TrustedBundleRO
+
+	// ClusterDomain to use when building service URLs.
+	ClusterDomain string
```
caseydavenport, line 76: Is ClusterDomain ever used?

caseydavenport (changes requested):
Looks good! A few thoughts

pkg/render/logstorage.go
asincu: Not sure ... I was simply following the pattern used for esgateway.

caseydavenport: es-gateway shouldn't be making egress calls to the Job AFAIK, so I think we don't need it.

pkg/components/overrides.go
```diff
@@ -55,3 +55,9 @@ type ReplicatedPodResourceOverrides interface {
 
 	GetDeploymentStrategy() *appsv1.DeploymentStrategy
 }
+
+type PodResourceOverrides interface {
```
caseydavenport, line 59: Is there a reason we needed to introduce a new interface for this?

As far as I can tell, `ReplicatedPodResourceOverrides` should do the trick. Sure, some of the fields (GetDeploymentStrategy) don't make sense for a job. But we just return `nil` for the methods that don't apply and they are skipped.

pkg/controller/logstorage/users/users_controller.go
```diff
@@ -314,7 +340,7 @@ func (r *UserController) createLinseedLogin(ctx context.Context, elasticEndpoint
 	}
 
 	// Create the user in ES.
-	user := utils.LinseedUser(clusterID, tenantID)
+	//user := utils.LinseedUser(clusterID, tenantID)
```
caseydavenport, line 343: Commented code left in here

pkg/render/common/components/components.go
```diff
@@ -39,6 +40,13 @@ type replicatedPodResource struct {
 	deploymentStrategy *appsv1.DeploymentStrategy // Deployments only
 }
 
+// replicatedPodResource contains the overridable data for a Deployment or DaemonSet.
+type podResource struct {
```
caseydavenport, line 44: same here - feel like we could just use `replicatedPodResource` instead?

pkg/render/common/components/components.go
```diff
@@ -145,6 +162,29 @@ func ApplyDeploymentOverrides(d *appsv1.Deployment, overrides components.Replica
 	d.Spec.Strategy = *r.deploymentStrategy
 }
 
+// ApplyJobOverrides applies the overrides to the given Job.
+// Note: overrides must not be nil pointer.
+func ApplyJobOverrides(job *batchv1.Job, overrides components.PodResourceOverrides) {
+	// Catch if caller passes in an explicit nil.
```
caseydavenport, line 168: I know this is just copied from existing code, but the comments here are sort of wrong. The function level comment says "Must not be nil" and then the code actually handles `nil` explicitly, so seems to contradict the previous comment.

pkg/components/overrides.go
asincu: K8S Jobs do not have `MinReadySeconds` and `DeploymentStrategy`. Altough, the rest of the methods are applicable to a resource that contains PodTemplate objects.

pkg/controller/logstorage/dashboards/dashboards_controller.go
```diff
@@ -0,0 +1,374 @@
...
+	if err = c.Watch(&source.Kind{Type: &operatorv1.LogStorage{}}, eventHandler); err != nil {
+		return fmt.Errorf("log-storage-dashboards-controller failed to watch LogStorage resource: %w", err)
+	}
+	if err = c.Watch(&source.Kind{Type: &operatorv1.Installation{}}, eventHandler); err != nil {
+		return fmt.Errorf("log-storage-dashboards-controller failed to watch Installation resource: %w", err)
+	}
+	if err = c.Watch(&source.Kind{Type: &operatorv1.ManagementClusterConnection{}}, eventHandler); err != nil {
+		return fmt.Errorf("log-storage-dashboards-controller failed to watch ManagementClusterConnection resource: %w", err)
+	}
+	if err = utils.AddTigeraStatusWatch(c, "log-storage-dashboards"); err != nil {
```
caseydavenport, line 109: we should use `tigeraStatusName` name here

caseydavenport: Once @vara2504's PR is in, we'll need tdo add this status to the aggregated one as well. https://github.com/tigera/operator/pull/3118/

pkg/controller/logstorage/dashboards/dashboards_controller_test.go
```diff
@@ -0,0 +1,511 @@
...
+			}
+			err = cli.Get(ctx, types.NamespacedName{Name: dashboardJob.Name, Namespace: dashboardJob.Namespace}, &dashboardJob)
+			Expect(err).Should(HaveOccurred())
+			Expect(errors.IsNotFound(err)).Should(BeTrue())
+
+			// Check that OnCRFound was not called.
+			mockStatus.AssertNotCalled(GinkgoT(), "OnCRFound")
+		})
+
+		It("should reconcile resources for a tenant", func() {
```
caseydavenport, line 376: I don't think this test is correct - it seems to claim that this works for a tenant, but actually this controller doesn't reconcile for multi-tenant envs.

caseydavenport (approved):
A couple of small comments but otherwise LGTM

pkg/controller/logstorage/dashboards/dashboards_controller_test.go
asincu: We are disabling multi-tenancy in Add() and testing out Reconcile(). Seems like there is mismatch between this two.

asincu: We disable multi-tenancy in Add() and test calling Reconcile(). Seems like the is a mismatch between these two.

pkg/controller/logstorage/dashboards/dashboards_controller_test.go
```diff
@@ -0,0 +1,511 @@
...
+			// Check that Dashboards was created as expected. We don't need to check every resource in detail, since
+			// the render package has its own tests which cover this in more detail.
+			dashboardJob := batchv1.Job{
+				TypeMeta: metav1.TypeMeta{Kind: "Job", APIVersion: "batch/v1"},
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      dashboards.Name,
+					Namespace: tenantNS,
+				},
+			}
+			Expect(test.GetResource(cli, &dashboardJob)).To(BeNil())
```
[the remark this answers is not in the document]
asincu: Reconcile actually renders this resource. We disable multi-tenancy in Add() and test calling Reconcile(). Seems like the is a mismatch between these two.

PR #3140, opened 2024-01-26
Add feature annotations to cluster-service-version

lwr20:
## Description
Openshift certification has added a set of annotations that will be required in operator bundles for certification from 1st March onwards.  This PR adds those annotations.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/hack/gen-bundle/update-bundle.sh
+++ b/hack/gen-bundle/update-bundle.sh
@@ -43,6 +43,19 @@ yq write -i ${CSV} metadata.annotations.containerImage ${OPERATOR_IMAGE_DIGEST}
 TIMESTAMP=$(echo ${OPERATOR_IMAGE_INSPECT} | jq -r .[0].Created)
 yq write -i ${CSV} metadata.annotations.createdAt ${TIMESTAMP}
 
+# Add the features annotations
+FEATURES=metadata.annotations.features.operators.openshift.io
+yq write -i ${CSV} ${FEATURES}/disconnected "false"
+yq write -i ${CSV} ${FEATURES}/fips-compliant "false"
+yq write -i ${CSV} ${FEATURES}/proxy-aware "false"
+yq write -i ${CSV} ${FEATURES}/tls-profiles "false"
+yq write -i ${CSV} ${FEATURES}/token-auth-aws "false"
+yq write -i ${CSV} ${FEATURES}/token-auth-azure "false"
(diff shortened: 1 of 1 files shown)


matthewdupre (approved):
I think this is a correct; albeit slightly pessimistic interpretation.  We may qualify for things like fips-compliant, and disconnected per RH's rules, but since I'm not sure what those require exactly it's probably the right choice to keep it simple.

hack/gen-bundle/update-bundle.sh
```diff
@@ -43,6 +43,19 @@ yq write -i ${CSV} metadata.annotations.containerImage ${OPERATOR_IMAGE_DIGEST}
...
+yq write -i ${CSV} ${FEATURES}/disconnected "false"
+yq write -i ${CSV} ${FEATURES}/fips-compliant "false"
+yq write -i ${CSV} ${FEATURES}/proxy-aware "false"
+yq write -i ${CSV} ${FEATURES}/tls-profiles "false"
+yq write -i ${CSV} ${FEATURES}/token-auth-aws "false"
+yq write -i ${CSV} ${FEATURES}/token-auth-azure "false"
+yq write -i ${CSV} ${FEATURES}/token-auth-gcp "false"
+yq write -i ${CSV} ${FEATURES}/cnf "false"
+yq write -i ${CSV} ${FEATURES}/cni "true"
+yq write -i ${CSV} ${FEATURES}/csi "false"
```
tmjd, line 57: I was looking at the link provided and I thought we did provide a csi plugin. We do have the `csi-node-driver` daemonset.

lwr20: Good point.  @matthewdupre thoughts?

My thoughts on this are that while we _include_ a CSI plugin in our install, we don't _provide_ a CSI plugin with our install.

My worry would be that people would assume that they don't need to install another CSI plugin on their cluster because Calico takes care of that, which isn't quite true.

caseydavenport: We don't provide CSI functionality to the end user. We use CSI for a very specific purpose within our produce to make one of our features (ALP) work. 

We're aren't a general purpose storage driver, so I suspect we shouldn't indicate that we are in this manifest.

lwr20: @tmjd are you happy to approve this now?

tmjd (approved):

PR #3142, opened 2024-01-27
Deploy es-kube-controllers in a multi-tenant environment

asincu:
## Description

Deploy es kube controllers in a tenant namespace to enable license copy in the managed clusters. We will use a new configuration `managedclusterlicensing`, that only deploy the licensing controller. We also need to impersonate kube-controllers service account from calico-system namespace, since this is the one that has rights inside the managed cluster.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/logstorage/kubecontrollers/es_kube_controllers.go
+++ b/pkg/controller/logstorage/kubecontrollers/es_kube_controllers.go
@@ -289,7 +289,7 @@ func (r *ESKubeControllersController) Reconcile(ctx context.Context, request rec
 	}
 
 	// Get secrets needed for kube-controllers to talk to elastic. This is needed for zero-tenants and single-tenants
-	// that deploy eskube controllers and need to talk to es-gateway
+	// that deploy es-kube-controllers and need to talk to es-gateway
 	var kubeControllersUserSecret *core.Secret
 	if !r.multiTenant {
 		kubeControllersUserSecret, err = utils.GetSecret(ctx, r.client, kubecontrollers.ElasticsearchKubeControllersUserSecret, helper.TruthNamespace())
--- a/pkg/controller/logstorage/kubecontrollers/es_kube_controllers.go
+++ b/pkg/controller/logstorage/kubecontrollers/es_kube_controllers.go
@@ -118,12 +118,6 @@ func Add(mgr manager.Manager, opts options.AddOptions) error {
 		}
 	}
 
-	// The namespace(s) we need to monitor depend upon what tenancy mode we're running in.
-	// For single-tenant, everything is installed in the calico-system namespace.
-	// Make a helper for determining which namespaces to use based on tenancy mode.
-	esGatewayHelper := utils.NewNamespaceHelper(opts.MultiTenant, render.ElasticsearchNamespace, "")
-	helper := utils.NewNamespaceHelper(opts.MultiTenant, common.CalicoNamespace, "")
-
 	// Watch secrets this controller cares about.
 	secretsToWatch := []string{
 		render.TigeraElasticsearchGatewaySecret,
@@ -140,23 +134,17 @@ func Add(mgr manager.Manager, opts options.AddOptions) error {
 		}
 	}
 
-	// Catch if something modifies the resources that this controller consumes.
-	if err := utils.AddServiceWatch(c, render.ElasticsearchServiceName, esGatewayHelper.InstallNamespace()); err != nil {
+	// The namespace(s) we need to monitor depend upon what tenancy mode we're running in.
+	// For single-tenant, everything is installed in the calico-system namespace.
+	// Make a helper for determining which namespaces to use based on tenancy mode.
+	esKubeControllersNamespace := utils.NewNamespaceHelper(opts.MultiTenant, common.CalicoNamespace, "")
+	if err := utils.AddConfigMapWatch(c, certificatemanagement.TrustedCertConfigMapName, esKubeControllersNamespace.InstallNamespace(), &handler.EnqueueRequestForObject{}); err != nil {
 		return fmt.Errorf("log-storage-kubecontrollers failed to watch the Service resource: %w", err)
 	}
-	if err := utils.AddServiceWatch(c, esgateway.ServiceName, esGatewayHelper.InstallNamespace()); err != nil {
-		return fmt.Errorf("log-storage-kubecontrollers failed to watch the Service resource: %w", err)
-	}
-	if err := utils.AddConfigMapWatch(c, certificatemanagement.TrustedCertConfigMapName, esGatewayHelper.InstallNamespace(), &handler.EnqueueRequestForObject{}); err != nil {
-		return fmt.Errorf("log-storage-kubecontrollers failed to watch the ConfigMap resource: %w", err)
-	}
-	if err := utils.AddConfigMapWatch(c, certificatemanagement.TrustedCertConfigMapName, helper.InstallNamespace(), &handler.EnqueueRequestForObject{}); err != nil {
-		return fmt.Errorf("log-storage-kubecontrollers failed to watch the Service resource: %w", err)
-	}
-	if err := utils.AddDeploymentWatch(c, esgateway.DeploymentName, helper.InstallNamespace()); err != nil {
+	if err := utils.AddDeploymentWatch(c, esgateway.DeploymentName, esKubeControllersNamespace.InstallNamespace()); err != nil {
 		return fmt.Errorf("log-storage-access-controller failed to watch the Service resource: %w", err)
 	}
-	if err := utils.AddDeploymentWatch(c, kubecontrollers.EsKubeController, helper.InstallNamespace()); err != nil {
+	if err := utils.AddDeploymentWatch(c, kubecontrollers.EsKubeController, esKubeControllersNamespace.InstallNamespace()); err != nil {
 		return fmt.Errorf("log-storage-access-controller failed to watch the Service resource: %w", err)
 	}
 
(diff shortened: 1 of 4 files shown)


pkg/controller/logstorage/kubecontrollers/es_kube_controllers.go
```diff
@@ -109,13 +111,32 @@ func Add(mgr manager.Manager, opts options.AddOptions) error {
...
 	// Watch secrets this controller cares about.
+	// ALINA : - Need to remove all esgateway watches for single tenant
 	secretsToWatch := []string{
 		render.TigeraElasticsearchGatewaySecret,
 		monitor.PrometheusClientTLSSecretName,
 	}
-	for _, ns := range []string{common.OperatorNamespace(), render.ElasticsearchNamespace} {
+
+	// Determine namespaces to watch.
+	namespacesToWatch := []string{esGatewayHelper.TruthNamespace(), esGatewayHelper.InstallNamespace()}
```
caseydavenport, line 134: There's a utility function for this - `tenancy.GetWatchNamespaces`

pkg/controller/logstorage/kubecontrollers/es_kube_controllers.go
```diff
@@ -149,22 +176,40 @@ func Add(mgr manager.Manager, opts options.AddOptions) error {
...
+		return reconcile.Result{}, nil
+	}
+
+	// When running in multi-tenant mode, we need to install es-kubecontrollers in tenant Namespaces. However, the LogStorage
+	// resource is still cluster-scoped (since ES is a cluster-wide resource), so we need to look elsewhere to determine
+	// which tenant namespaces require an es-kubecontrollers instance. We use the tenant API to determine the set of
+	// namespaces that should have an es-kubecontrollers.
+	tenant, _, err := utils.GetTenant(ctx, r.multiTenant, r.client, request.Namespace)
+	if errors.IsNotFound(err) {
+		reqLogger.Info("No Tenant in this Namespace, skip")
```
caseydavenport, line 202: This should be `.V(1)` I think to avoid being too spammy.

pkg/controller/logstorage/kubecontrollers/es_kube_controllers.go
```diff
@@ -280,20 +326,22 @@ func (r *ESKubeControllersController) Reconcile(ctx context.Context, request rec
...
-		pullSecrets,
-		hdler,
-		reqLogger,
-		trustedBundle,
-		r.usePSP,
-	); err != nil {
-		return reconcile.Result{}, err
+	if !r.multiTenant {
+		// ESGateway is required in order for kube-controllers to operate successfully, since es-kube-controllers talks to ES
+		// via this gateway. However, in multi-tenant mode we disable es and so this isn't needed.
```
caseydavenport, line 331: >  in multi-tenant mode we disable es and so this isn't needed.

probably

> in multi-tenant mode, es-kube-controllers doesn't talk to elasticsearch, so this is only needed in single-tenant clusters.

pkg/controller/logstorage/kubecontrollers/es_kube_controllers.go
```diff
@@ -313,6 +361,7 @@ func (r *ESKubeControllersController) Reconcile(ctx context.Context, request rec
 		TrustedBundle:                trustedBundle,
 		Namespace:                    helper.InstallNamespace(),
 		BindingNamespaces:            namespaces,
+		MultiTenant:                  r.multiTenant,
```
caseydavenport, line 364: We should be passing `Tenant: tenant` here instead, so we can use `tenant.MultiTenant()`

pkg/controller/logstorage/kubecontrollers/es_kube_controllers.go
```diff
@@ -111,13 +116,27 @@ func Add(mgr manager.Manager, opts options.AddOptions) error {
...
+	if opts.MultiTenant {
+		if err = c.Watch(&source.Kind{Type: &operatorv1.Tenant{}}, &handler.EnqueueRequestForObject{}); err != nil {
+			return fmt.Errorf("log-storage-kubecontrollers failed to watch Tenant resource: %w", err)
+		}
+	}
+
+	// The namespace(s) we need to monitor depend upon what tenancy mode we're running in.
+	// For single-tenant, everything is installed in the calico-system namespace.
+	// Make a helper for determining which namespaces to use based on tenancy mode.
+	esGatewayHelper := utils.NewNamespaceHelper(opts.MultiTenant, render.ElasticsearchNamespace, "")
```
rene-dekker, line 128: Wouldn't it be better to do:
```
	esgwNamespace := utils.NewNamespaceHelper(opts.MultiTenant, render.ElasticsearchNamespace, "").InstallNamespace()
	kubecontrollersNamespace := utils.NewNamespaceHelper(opts.MultiTenant, common.CalicoNamespace, "").InstallNamespace()

```

And move the lines to right above the place they are used? Imo the code is good enough to do without the comment, but I don't mind keeping it either.

pkg/controller/logstorage/kubecontrollers/es_kube_controllers.go
```diff
@@ -151,22 +176,40 @@ func Add(mgr manager.Manager, opts options.AddOptions) error {
...
 }
 
 func (r *ESKubeControllersController) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
-	helper := utils.NewNamespaceHelper(false, common.CalicoNamespace, request.Namespace)
+	helper := utils.NewNamespaceHelper(r.multiTenant, common.CalicoNamespace, request.Namespace)
 	reqLogger := log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name, "installNS", helper.InstallNamespace(), "truthNS", helper.TruthNamespace())
 	reqLogger.Info("Reconciling LogStorage - ESKubeControllers")
 
+	// We skip requests without a namespace specified in multi-tenant setups.
+	if r.multiTenant && request.Namespace == "" {
```
rene-dekker, line 192: Does that mean we don't need to watch objects without a namespace either?

pkg/render/kubecontrollers/kube-controllers.go
```diff
@@ -171,10 +198,22 @@ func NewElasticsearchKubeControllers(cfg *KubeControllersConfiguration) *kubeCon
...
-	enabledControllers := []string{"authorization", "elasticsearchconfiguration"}
+	var enabledControllers []string
+
+	if !cfg.Tenant.MultiTenant() {
+		enabledControllers = append(enabledControllers, "authorization", "elasticsearchconfiguration")
+		if cfg.ManagementCluster != nil {
+			enabledControllers = append(enabledControllers, "managedcluster")
+		}
+	} else {
+		if cfg.ManagementCluster != nil {
```
rene-dekker, line 209: Could be else-if to simplify the code. But if cfg.Tenant.MultiTenant()=True, doesn't that always mean that cfg.ManagementCluster != nil ?

pkg/render/kubecontrollers/kube-controllers.go
```diff
@@ -578,7 +625,10 @@ func (c *kubeControllersComponent) controllersDeployment() *appsv1.Deployment {
 			},
 		},
 	}
-	render.SetClusterCriticalPod(&(d.Spec.Template))
+
+	if !c.cfg.Tenant.MultiTenant() {
+		render.SetClusterCriticalPod(&(d.Spec.Template))
```
rene-dekker, line 630: I believe these are unnecessary parentheses.

rene-dekker (approved):
lgtm

pkg/controller/logstorage/kubecontrollers/es_kube_controllers.go
caseydavenport: > Does that mean we don't need to watch objects without a namespace either in multi-tenant setups?

We need to watch them, but with the `EnqueueAllTenants` handler instead.

This particular check is probably superfluous to be honest - if all is working as intended, we should never hit this branch. However, I think it's a fine-enough safety valve.

pkg/controller/logstorage/kubecontrollers/es_kube_controllers.go
```diff
@@ -272,40 +322,43 @@ func (r *ESKubeControllersController) Reconcile(ctx context.Context, request rec
...
-		install,
-		variant,
-		pullSecrets,
-		hdler,
-		reqLogger,
-		gwTrustedBundle,
-		r.usePSP,
-	); err != nil {
-		return reconcile.Result{}, err
+	if !r.multiTenant {
```
caseydavenport, line 331: This needs the same comment:

> 	// In zero-tenant clusters, ESGateway is required in order for kube-controllers to operate successfully, since es-kube-controllers talks to ES via this gateway. However, in multi-tenant mode we disable the elasticsearch controller and so this isn't needed.

caseydavenport: I see that you moved it below actually, but I think it's better at the top of the branch. It should also say "In zero-tenant clusters" for clarity.

pkg/render/kubecontrollers/kube-controllers.go
```diff
@@ -100,6 +103,9 @@ type KubeControllersConfiguration struct {
 
 	// List of namespaces that are running a kube-controllers instance that need a cluster role binding.
 	BindingNamespaces []string
+
+	// Whether or not to run the rendered components in multi-tenant mode.
```
caseydavenport, line 107: This comment isn't correct - the Tenant object provides tenant configuration for both single and multi-tenant modes. If this is nil, then we should run in zero-tenant mode.

pkg/render/kubecontrollers/kube-controllers.go
```diff
@@ -171,10 +198,20 @@ func NewElasticsearchKubeControllers(cfg *KubeControllersConfiguration) *kubeCon
...
+		}
+	} else {
+		enabledControllers = append(enabledControllers, "managedclusterlicensing")
+	}
 
-	if cfg.ManagementCluster != nil {
-		enabledControllers = append(enabledControllers, "managedcluster")
+	kubeControllerConfigName := "elasticsearch"
+	if cfg.Tenant.MultiTenant() {
+		kubeControllerConfigName = fmt.Sprintf("tenant-%s", cfg.Tenant.Spec.ID)
```
caseydavenport, line 214: I think this code should be reverted - we're not creating one per tenant any more IIUC

pkg/render/kubecontrollers/kube-controllers.go
```diff
@@ -578,7 +623,10 @@ func (c *kubeControllersComponent) controllersDeployment() *appsv1.Deployment {
 			},
 		},
 	}
-	render.SetClusterCriticalPod(&(d.Spec.Template))
+
+	if !c.cfg.Tenant.MultiTenant() {
```
caseydavenport, line 627: Is this change needed? Why don't we want to set this in multi-tenant?

pkg/render/kubecontrollers/kube-controllers.go
asincu: Yup. I have this removed, forgot to push.

pkg/render/kubecontrollers/kube-controllers.go
asincu: I cannot create multiple instances of es-kubecontrollers with this setting.

caseydavenport: Okie doke.

pkg/controller/logstorage/kubecontrollers/es_kube_controllers.go
```diff
@@ -247,15 +293,19 @@ func (r *ESKubeControllersController) Reconcile(ctx context.Context, request rec
 	}
 
 	// Get secrets needed for kube-controllers to talk to elastic.
-	kubeControllersUserSecret, err := utils.GetSecret(ctx, r.client, kubecontrollers.ElasticsearchKubeControllersUserSecret, helper.TruthNamespace())
-	if err != nil {
-		r.status.SetDegraded(operatorv1.ResourceReadError, "Failed to get kube controllers gateway secret", err, reqLogger)
-		return reconcile.Result{}, err
+	var kubeControllersUserSecret *core.Secret
```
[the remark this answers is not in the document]
caseydavenport: Typo: "eskube controllers" should be "es-kube-controllers"

caseydavenport (approved):
@asincu one typo but otherwise LGTM

PR #3111, opened 2024-01-11
Compliance multi-tenant changes

Josh-Tigera:
## For PR author

- [x] Tests for change.
~- [ ] If changing pkg/apis/, run `make gen-files`~
~- [ ] If changing versions, run `make gen-versions`~

## For PR reviewers

A note for code reviewers - all pull requests must have the following:

- [ ] Milestone set according to targeted release.
- [ ] Appropriate labels:
  - `kind/bug` if this is a bugfix.
  - `kind/enhancement` if this is a a new feature.
  - `enterprise` if this PR applies to Calico Enterprise only.

PR #3145, opened 2024-01-30
Fix the Compliance namespace in Voltron's proxy targets

rene-dekker:
Fixes: https://github.com/tigera/operator/pull/3111/files#diff-a95b82b76390b16587aaf2a6d12d9ff332d5597002643a3846b518a914aebce0R500


caseydavenport (approved):
LGTM, wonder if we should be following this pattern for the Linseed endpoint as well since it's a bit simpler than how we've built the logic for that today.

PR #3147, opened 2024-01-31
Support git alternates in pre-commit hook

danudey:
## Description

If using the git alternates functionality (e.g. `git clone --reference /some/local/repository`) the
reference repositories are not available inside of the git container, causing most or all files to
not be recognized and to show as changed in the pre-commit hook.

This patch simply mounts (read-only) any git alternates into the container at the same path as on the
host, allowing the hook to run as expected.

If there are no alternates configured and/or the alternates file doesn't exist, the script behaves as normal.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -908,7 +908,7 @@ install-git-hooks:
 
 .PHONY: pre-commit
 pre-commit:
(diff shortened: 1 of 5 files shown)


rene-dekker (approved):
lgtm
You need to run `make gen-files gen-versions` and commit the changes.
Don't forget to pick to master.

PR #3156, opened 2024-02-05
should not ignore non-migrated nodes for typha scheduling

mihivagyok:
## Description
kind/bugfix

If there is any problem with the migration procedure (`kube-system` to `calico-system`), the migration does not continue automatically,  As noticed, the `typha_autoscaler` blocks the migration as it waits for nodes without the label `projectcalico.org/operator-node-migration=pre-operator`. But the migration logic puts this label to all the nodes.
That's why if the migration fails then on the next reconcile, the `typha_autoscaler` reports degraded state as it can't find enough nodes and blocks the migration. This makes no sense, as `calico-typha` does not have related node selector and it does not depend on calico-node (it uses hostNetwork).

```bash
➜ kubectl get tigerastatus calico -o yaml
kind: TigeraStatus
metadata:
  creationTimestamp: "2024-02-01T13:11:58Z"
  generation: 1
  name: calico
  resourceVersion: "15240"
  uid: 326eadff-c10a-4971-a617-6ec1b033e7c3
spec: {}
status:
  conditions:
  - lastTransitionTime: "2024-02-01T13:12:08Z"
    message: 'Failed to autoscale typha - not enough linux nodes to schedule typha
      pods on, require 1 and have 0: '
    observedGeneration: 8
    reason: ResourceScalingError
    status: "True"
    type: Degraded
```

Although there is a running `calico-typha` pod already in the `calico-system` namespace.

```bash
➜ kubectl get pods -n calico-system
NAME                                       READY   STATUS    RESTARTS         AGE
calico-kube-controllers-79c99bcfc8-tmgq6   1/1     Running   0                4h
calico-typha-6fb44dd6cd-w9v46              1/1     Running   0                4h
```

With the fix, the status shows a much better error if there is a `NotReady` node:
```
➜ kubectl get tigerastatus calico -o yaml
kind: TigeraStatus
metadata:
  creationTimestamp: "2024-02-03T19:36:03Z"
  generation: 1
  name: calico
  resourceVersion: "8289"
  uid: fc9a19c1-9cc7-4b23-9241-181a1be9e68a
spec: {}
status:
  conditions:
  - lastTransitionTime: "2024-02-03T19:46:03Z"
    message: 'error migrating resources to calico-system: the kube-system node DaemonSet
      is not ready with the updated nodeSelector: timed out waiting for the condition'
    observedGeneration: 4
    reason: ResourceMigrationError
    status: "True"
    type: Degraded
```

Once the node is fixed, the migration coninues automatically as it is expected.

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/installation/typha_autoscaler.go
+++ b/pkg/controller/installation/typha_autoscaler.go
@@ -235,10 +235,6 @@ func (t *typhaAutoscaler) getNodeCounts() (int, int, error) {
 		if n.Spec.Unschedulable {
 			continue
 		}
-		if n.GetObjectMeta().GetLabels()["projectcalico.org/operator-node-migration"] == "pre-operator" {
-			// This node hasn't been migrated to the operator yet. Don't include it in the number of desired Typhas.
-			continue
-		}
 
 		if _, ok := n.Labels["kubernetes.azure.com/cluster"]; ok && n.Labels["type"] == "virtual-kubelet" {
 			// in AKS, there is a feature called 'virtual-nodes' which represent azure's container service as a node in the kubernetes cluster.
--- a/pkg/controller/installation/typha_autoscaler_test.go
+++ b/pkg/controller/installation/typha_autoscaler_test.go
@@ -164,7 +164,7 @@ var _ = Describe("Test typha autoscaler ", func() {
 		verifyTyphaReplicas(c, 2)
 	})
 
-	It("should ignore non-migrated nodes in its count", func() {
+	It("should not ignore non-migrated nodes in its count", func() {
 		typhaMeta := metav1.ObjectMeta{
 			Name:      "calico-typha",
 			Namespace: "calico-system",
@@ -193,9 +193,7 @@ var _ = Describe("Test typha autoscaler ", func() {
 		ta := newTyphaAutoscaler(c, nodeIndexInformer, tlw, statusManager, typhaAutoscalerPeriod(10*time.Millisecond))
 		ta.start(ctx)
 
-		// normally we'd expect to see three replicas for five nodes, but since one node is not migrated,
-		// we should still only expect two
-		verifyTyphaReplicas(c, 2)
+		verifyTyphaReplicas(c, 3)
 	})
 
 	It("should ignore aks virtual nodes in its count", func() {


tmjd:
As I understand from your description, it isn't actually a problem with ignoring nodes with the label `projectcalico.org/operator-node-migration=pre-operator`. But when ignoring those nodes the tigerastatus reported when there is a problem might not indicate if there is a problem with the kube-system calico-node daemonset.

I believe that label was included in the node count for typha was for small cluster migrations with typha. IIRC on small clusters if typha is already used and we try to deploy the new typha without using that label to limit the count there would be more typhas (kube-system plus calico-system) than there are nodes which would mean not all the typhas would be able to be deployed.

It sounds like the correct way to address the issue would be to ensure that the issue `error migrating resources to calico-system: the kube-system node DaemonSet is not ready with the updated nodeSelector: timed out waiting for the condition` is "higher" priority/checked before checking the typha scale check.

mihivagyok:
Hi @tmjd !

Thanks.
During the first reconcile the following is happening:
- `calico-typha` autoscaler is starting - there is no node with migration label, no error
- namespace migration starts, labels all the nodes with `projectcalico.org/operator-node-migration=pre-operator`, but it fails to update `kube-system/calico-node` due to the `NotReady` node - first reconcile ends

Second reconcile:
-  `calico-typha` autoscaler state check - as now all the nodes all labeled with `projectcalico.org/operator-node-migration=pre-operator` label, it reports the error, reconcile ends - and this will happen during the next reconcilation attemtps.

At this point you need manual intervention to make any more progress (removing the label or set it to `migrated`.
So even the `error migrating resources to calico-system: the kube-system node DaemonSet is not ready with the updated nodeSelector: timed out waiting for the condition` would be "higher", it would not help as the code does not reach the namespace-migration function anymore.

We have tested this change with small clusters (1, 2 and 3 nodes), and function `ensureTyphaRoom` makes sure that `kube-system/calico-typha` deployment is scaled down properly - so new `calico-system/calico-typha` pods can start. The typhas are not using any migration labels in their deployments as far as I see.

I hope it makes sense. We are testing this since months and we were thinking a lot how to solve our issues and this change was a helping a lot.

Thanks,
Adam

tmjd (approved):
Thanks for the explanation.

I think we originally added the label as the method to adjust the scale of typha during the migration and originally we ran more typhas. I think now that the number of typhas has been reduced (at least for small clusters) and we have `ensureTyphaRoom` it isn't necessary to scale typha during the migration.

Thanks for this update.

PR #3157, opened 2024-02-05
Wait for calico-node DaemonSet rollout to avoid mass restart during manifest-to-operator migration

mihivagyok:
## Description

kind/bugfix

As we have experienced when we start to use the `tigera-operator`, the migration take down ALL Calico node pods on my cluster such that they were running `0/1` ready status at the same time.  Of course if the migration continues smoothly this is fine.  However, if it does not then your pod network is taken down by the migration.

```
NAMESPACE         NAME                                                  READY   STATUS        RESTARTS      AGE   IP               NODE             NOMINATED NODE   READINESS GATES
calico-system     calico-kube-controllers-9d5b46798-wpddk               1/1     Running       0             33m   172.30.6.12      10.177.136.158   <none>           <none>
calico-system     calico-node-rc6kx                                     0/1     Terminating   0             12s   10.177.136.158   10.177.136.158   <none>           <none>
calico-system     calico-node-tlqwl                                     0/1     Running       0             9s    10.177.136.169   10.177.136.169   <none>           <none>
calico-system     calico-node-vmrdv                                     0/1     Running       0             9s    10.177.136.151   10.177.136.151   <none>           <none>
calico-system     calico-typha-7f78866dc4-5mrzm                         1/1     Running       0             7s    10.177.136.169   10.177.136.169   <none>           <none>
calico-system     calico-typha-7f78866dc4-9grbh                         1/1     Running       0             33m   10.177.136.151   10.177.136.151   <none>           <none>
```

The daemonset has the following setting:
```
  updateStrategy:
    rollingUpdate:
      maxSurge: 0
      maxUnavailable: 1
    type: RollingUpdate
```
This is not respected and it can cause problems.

To avoid that, we need to wait the calico-node daemoset to get properly updated before doing any more change.
- Migration code uses the function `waitUntilNodeCanBeMigrated()` which is checking the `calico-system/calico-node` daemoset status, but that is not waiting the rollout to complete fully. Then the next reconcile will remove the `nodeSelector: projectcalico.org/operator-node-migration=migrated` from the `calico-system/calico-node` (as the migration is completed now), so the `calico-system/calico-node` daemonset will be updated once more. But as the `waitUntilNodeCanBeMigrated()` does not wait the last update to complete, it can be a following state e.g
```
status:
  currentNumberScheduled: 4
  desiredNumberScheduled: 4
  numberAvailable: 3
  numberMisscheduled: 0
  numberReady: 4
  numberUnavailable: 1
  observedGeneration: 2
  updatedNumberScheduled: 1
```
then the second update (removing the nodeSelector from the daemonset) could lead to mass calico-node restart as shown above. That's why let's wait the rollout fully during the migration.

- as I wrote earlier, once the migration logic is done, the next reconcile will remove the `nodeSelector: projectcalico.org/operator-node-migration=migrated` from the `calico-system/calico-node`. Then the same label will be removed from the nodes. If the label is removed earlier from the nodes than the the `nodeSelector` from the calico-node pods, it could lead to similar mass calico-node restart. So before cleaning up the migration related label from the nodes, wait for the calico-node rollout.

These changes may slow down the migration, but it makes it more stable as we have experienced.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/migration/namespace_migration.go
+++ b/pkg/controller/migration/namespace_migration.go
@@ -72,7 +72,7 @@ type NamespaceMigration interface {
 	NeedsCoreNamespaceMigration(ctx context.Context) (bool, error)
 	Run(ctx context.Context, log logr.Logger) error
 	NeedCleanup() bool
-	CleanupMigration(ctx context.Context) error
+	CleanupMigration(ctx context.Context, log logr.Logger) error
 }
 
 type CoreNamespaceMigration struct {
@@ -291,6 +291,11 @@ func (m *CoreNamespaceMigration) Run(ctx context.Context, log logr.Logger) error
 		return fmt.Errorf("failed to migrate all nodes: %s", err.Error())
 	}
 	log.V(1).Info("Nodes migrated")
+	err := m.waitForOperatorCalicoNodeDaemonsetReady(ctx, log)
+	if err != nil {
+		return fmt.Errorf("failed to wait for calico-node daemonset to be ready: %s", err.Error())
+	}
+	log.V(1).Info("calico-system/calico-node daemonset has been rolled out successfully")
 	if err := m.deleteKubeSystemCalicoNode(ctx); err != nil {
 		return fmt.Errorf("failed to delete kube-system node DaemonSet: %s", err.Error())
 	}
@@ -397,9 +402,14 @@ func (m *CoreNamespaceMigration) NeedCleanup() bool {
 
 // CleanupMigration ensures all labels used during the migration are removed
 // and any migration resources are stopped.
-func (m *CoreNamespaceMigration) CleanupMigration(ctx context.Context) error {
+func (m *CoreNamespaceMigration) CleanupMigration(ctx context.Context, log logr.Logger) error {
 	if m.migrationComplete {
 		return nil
+	}
+	// wait for calico-node rollout to avoid mass restart of calico-node pods
+	err := m.waitForOperatorCalicoNodeDaemonsetReady(ctx, log)
+	if err != nil {
+		return fmt.Errorf("failed to wait for calico-node daemonset to be ready: %s", err.Error())
 	}
 	if err := m.removeNodeMigrationLabelFromNodes(ctx); err != nil {
 		return fmt.Errorf("error cleaning up node labels: %s", err)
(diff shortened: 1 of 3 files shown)


pkg/controller/migration/namespace_migration.go
```diff
@@ -291,6 +291,11 @@ func (m *CoreNamespaceMigration) Run(ctx context.Context, log logr.Logger) error
 		return fmt.Errorf("failed to migrate all nodes: %s", err.Error())
 	}
 	log.V(1).Info("Nodes migrated")
+	err := m.waitForOperatorCalicoNodeDaemonsetReady(ctx, log)
```
tmjd, line 294: Is this needed at this point? It seems unnecessary from your description and because it is included as part of the CleanupMigration function.

mihivagyok: Yes, so after all the calico-node pods are migrated to the `calico-system` namespace, the daemonset state could be in a transient state. 
Then as the next reconcile removes the node selector from the daemonset and it finds the daemonset in this state, the update won't respect the `updateStrategy`. This was our experience, we saw mass updates of the calico-node pods.

Although I agree that it is hard to differentiate. It is also sure that if we remove the labels from the nodes first then from the daemonset, that could result in the same.
We have tested this many times (100+), and this combintatiion was working the best so far - otherwise there are so frequent changes, and it results situations like reported.

I can do some re-runs if you want to double-check these statements.

Thanks,
Adam

pkg/controller/migration/namespace_migration.go
```diff
@@ -796,3 +806,24 @@ func (m *CoreNamespaceMigration) removeNodeLabel(ctx context.Context, nodeName,
 		return true, nil
 	})
 }
+
+// waitForOperatorCalicoNodeDaemonsetReady waits until the 'new' calico-node daemonset in
+// the calico-system namespace is ready before continuing, it will wait up to
+// 1 minutes before returning with an error.
+func (m *CoreNamespaceMigration) waitForOperatorCalicoNodeDaemonsetReady(ctx context.Context, log logr.Logger) error {
+	return wait.PollImmediate(5*time.Second, 1*time.Minute, func() (bool, error) {
```
caseydavenport, line 814: Is 1 minute going to be long enough for larger clusters with hundreds of nodes? I suspect it might not - perhaps we need a variable time here based on the size of the DS?

mihivagyok: That's an excellent question. It does not enough. But in this way, the reconcilation is not blocked for long time. And the continues reconcilation just solve the problem - it waits on this 1 minutes, and ends. And an another rounds starts.

Although I agree we can make this better. 
We have tested with 700 nodes, and it took 6hrs 30 mins total with this change (tested with Kubernetes 1.29). The change definitely increases the time, but it makes it more stable based on our testing (this change was requested from our update team). Also, the rolling update settings is important here.

Thanks!

caseydavenport (approved):

caseydavenport:
Looks good to me too, sorry for the long delay!

mihivagyok:
@caseydavenport Thank you!

Can you help me to see the error in semaphoreci? Thanks!

tmjd:
It is the static-checks that are failing.
You should be able to check by running `make static-checks`

Here is the error I copied from the CI system:
```
make static-checks 

mkdir -p .go-pkg-cache /home/semaphore/go/pkg/mod && docker run --rm -v /home/semaphore/operator:/go/src/github.com/tigera/operator:rw -v /home/semaphore/operator/.go-pkg-cache:/go-cache/:rw -e LOCAL_USER_ID=1001 -e GOPATH=/go -e GOCACHE=/go-cache -e GOOS=linux -e GOARCH=amd64 -e KUBECONFIG=/go/src/github.com/tigera/operator/kubeconfig.yaml -e ACK_GINKGO_RC=true -e ACK_GINKGO_DEPRECATIONS=1.16.5 -w /go/src/github.com/tigera/operator --net=host -e GOPRIVATE=github.com/tigera/* -v /tmp/ssh-TlHBfo3ORHSe/agent.2529:/ssh-agent --env SSH_AUTH_SOCK=/ssh-agent -v /home/semaphore/go/pkg/mod:/go/pkg/mod:rw calico/go-build:v0.91-amd64 golangci-lint run --timeout 5m
Starting with UID: 1001
pkg/controller/migration/namespace_migration.go:823:9: SA1019: wait.PollImmediate is deprecated: This method does not return errors from context, use PollUntilContextTimeout. Note that the new method will no longer return ErrWaitTimeout and instead return errors defined by the context package. Will be removed in a future release. (staticcheck)
       return wait.PollImmediate(5*time.Second, 1*time.Minute, func() (bool, error) {

make: *** [Makefile:413: static-checks] Error 1
```

mihivagyok:
Thanks, I take a look asap!

marvin-tigera:

mihivagyok:
@tmjd Okay, this was a good finding. This new, proposed function is only available in a newer version of k8s.io/apimachinery. That's why I made a rebase to get the newer version of it. Thanks!

https://pkg.go.dev/k8s.io/apimachinery/pkg/util/wait#PollUntilContextTimeout

PR #3168, opened 2024-02-13
Fix autodetection of current RKE2

nelljerram:
Previously based on the kube-system/rke2 ConfigMap, but in current RKE2 the above ConfigMap no longer exists.  We leave that code in place in case there are variants where it is useful, but check also for the RKE2 DNS service - which is especially relevant because one of the main uses of the RKE2 autodetection is to set DNS config.

## For PR author

- [ ] Tests for change.
- [ ] If changing pkg/apis/, run `make gen-files`
- [ ] If changing versions, run `make gen-versions`

## For PR reviewers

A note for code reviewers - all pull requests must have the following:

- [ ] Milestone set according to targeted release.
- [ ] Appropriate labels:
  - `kind/bug` if this is a bugfix.
  - `kind/enhancement` if this is a a new feature.
  - `enterprise` if this PR applies to Calico Enterprise only.

PR #3169, opened 2024-02-14
Fix autodetection of current RKE2 [pick#3168 r1.33]

nelljerram:
Previously based on the kube-system/rke2 ConfigMap, but in current RKE2 the above ConfigMap no longer exists.  We leave that code in place in case there are variants where it is useful, but check also for the RKE2 DNS service - which is especially relevant because one of the main uses of the RKE2 autodetection is to set DNS config.

(cherry-pick of #3168 )

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


tmjd (approved):
LGTM

PR #3171, opened 2024-02-14
Omit cnx-manager container in single- and multi-tenant tigera-manager…

Josh-Tigera:
… pods

## For PR author

- [x] Tests for change.
~- [ ] If changing pkg/apis/, run `make gen-files`~
~- [ ] If changing versions, run `make gen-versions`~

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/manager.go
+++ b/pkg/render/manager.go
@@ -292,6 +292,10 @@ func (c *managerComponent) managerDeployment() *appsv1.Deployment {
 		initContainers = append(initContainers, c.cfg.VoltronLinseedKeyPair.InitContainer(ManagerNamespace))
 	}
 
+	managerPodContainers := []corev1.Container{c.managerEsProxyContainer(), c.voltronContainer()}
+	if !c.cfg.Tenant.SingleTenant() && !c.cfg.Tenant.MultiTenant() {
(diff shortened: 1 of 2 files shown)


pkg/render/manager.go
```diff
@@ -292,6 +292,10 @@ func (c *managerComponent) managerDeployment() *appsv1.Deployment {
 		initContainers = append(initContainers, c.cfg.VoltronLinseedKeyPair.InitContainer(ManagerNamespace))
 	}
 
+	managerPodContainers := []corev1.Container{c.managerEsProxyContainer(), c.voltronContainer()}
+	if !c.cfg.Tenant.SingleTenant() && !c.cfg.Tenant.MultiTenant() {
```
asincu, line 296: Can we simplify this check to be `if c.cfg.Tenant != nil`. I think we only need to install this for zero tenants, aka Enterprise installations.

asincu (approved):
WIth a small comment.

pkg/render/manager.go
Josh-Tigera: Sure yeah could do.

caseydavenport (approved):
LGTM but one comment from Alina to address I think

PR #3186, opened 2024-02-21
Adds PolicySetupTimeoutSeconds option to CalicoNetwork

aaaaaaaalex:
[Sibling PR to be merged first](https://github.com/projectcalico/calico/pull/8469)

## Description

Adds option to CalicoNetwork to enable a policy-setup timeout in the CNI/Felix.

Felix's new option `EndpointStatusPathPrefix` gets set to it's /var/run/calico dir to enable the Felix-side feature, where-in it will create a directory "status".

CNI should have its timeout configured according to the CalicoNetwork field, and it's `EndpointStatusDir` should be set to Felix's /var/run/calico dir plus "/status".

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


aaaaaaaalex:
Going to mark this PR as Ready for Review, although there are still two merge tests that are failing, that I'm a little confused about.

api/v1/installation_types.go
```diff
@@ -474,6 +474,17 @@ type CalicoNetworkSpec struct {
 	// Sysctl configures sysctl parameters for tuning plugin
 	// +optional
 	Sysctl []Sysctl `json:"sysctl,omitempty"`
+
+	// PolicySetupTimeoutSeconds delays new pods from going ready
+	// until their policy has been programmed in the dataplane.
+	// The specified delay defines the maximum amount of time
+	// that the Calico CNI will wait for policy to be programmed.
```
caseydavenport, line 481: "the Calico CNI _plugin" or just "Calico"

"CNI plugin" refers to a component, and "Calico" refers to the system as a whole.

api/v1/installation_types.go
```diff
@@ -474,6 +474,17 @@ type CalicoNetworkSpec struct {
 	// Sysctl configures sysctl parameters for tuning plugin
 	// +optional
 	Sysctl []Sysctl `json:"sysctl,omitempty"`
+
+	// PolicySetupTimeoutSeconds delays new pods from going ready
```
caseydavenport, line 478: "going ready" should probably be "running containers" - we actually prevent the code from running altogether until the policy is programmed.

pkg/controller/installation/validation.go
```diff
@@ -349,6 +349,13 @@ func validateCustomResource(instance *operatorv1.Installation) error {
 			}
 
 		}
+
+		if instance.Spec.CalicoNetwork.PolicySetupTimeoutSeconds != nil {
```
caseydavenport, line 353: I think we need validation here to prevent setting this in other invalid scenarios - e.g., when Calico networking isn't being used, or when Calico in eBPF mode is being used.

pkg/render/node.go
```diff
@@ -1716,6 +1727,15 @@ func (c *nodeComponent) nodeEnvVars() []corev1.EnvVar {
 		})
 	}
 
+	if c.cfg.Installation.CalicoNetwork != nil &&
+		c.cfg.Installation.CalicoNetwork.PolicySetupTimeoutSeconds != nil &&
+		*c.cfg.Installation.CalicoNetwork.PolicySetupTimeoutSeconds > 0 {
+		nodeEnv = append(nodeEnv, corev1.EnvVar{
+			Name:  "FELIX_ENDPOINTsTATUSPATHPREFIX",
```
caseydavenport, line 1734: capitalization typo ENDPOINTsTATUS

pkg/render/node.go
```diff
@@ -642,6 +647,8 @@ func (c *nodeComponent) createCalicoPluginConfig() map[string]interface{} {
 		"policy": map[string]interface{}{
 			"type": "k8s",
 		},
+		"policy_setup_timeout_seconds": policySetupTimeoutSeconds,
+		"endpoint_status_dir":          filepath.Join(c.varRunCalicoVolume().VolumeSource.HostPath.Path, "status"),
```
caseydavenport, line 651: `/var/run/calico/status` is a pretty generic path. Can we make this more intuitive?

e.g., 

`/var/run/calico/policy_program_status` ?

pkg/render/node_test.go
```diff
@@ -199,6 +199,8 @@ var _ = Describe("Node rendering tests", func() {
   "plugins": [
     {
       "type": "calico",
+	  "policy_setup_timeout_seconds": 0,
```
caseydavenport, line 202: Indentation is wrong here and doesn't match the other fields

pkg/controller/installation/core_controller.go
```diff
@@ -657,6 +657,14 @@ func fillDefaults(instance *operator.Installation) error {
 		}
 	}
 
+	// Only fill default policy setup timeout if CalicoCNI is used.
```
caseydavenport, line 660: nit, but "CalicoCNI" is neither a valid enum option in our API nor a commonly used representation of "Calico CNI plugin", so saying "if the Calico CNI plugin is used" would be more clear

caseydavenport: Similarly, this comment is self-explanatory by reading the code (or rather, it would be if the code didn't actually do something different from what the comment says!) and doesn't really add any additional context. Would be better if the comment explained _why_ we only default in this case.

> (or rather, it would be if the code didn't actually do something different from what the comment says!) 

We also are checking the LinuxDataplane here, so the comment doesn't match the code :)

pkg/controller/installation/validation.go
caseydavenport: What about Windows? Do we prevent this option in a cluster with Windows nodes? Do we just ignore it?

caseydavenport (changes requested):
Looks pretty good, left a few comments

pkg/render/node.go
caseydavenport: Would be nice to align with the field name too - we seem to be using `endpoint_status` and `endpointStatus` elsewhere, so maybe this should be `/var/run/calico/endpoint_status`

pkg/render/node_test.go
```diff
@@ -199,6 +199,8 @@ var _ = Describe("Node rendering tests", func() {
   "plugins": [
```
caseydavenport, line 199: I think we need some tests that check non-zero values

api/v1/installation_types.go
aaaaaaaalex: How about "[...] from entering the Running phase" ?

pkg/controller/installation/validation.go
aaaaaaaalex: I've been working off the assumption that we would ignore this option on Windows, how does the operator deal with the issue of windows compat? Does it know if a windows node exists in the cluster?

pkg/render/node_test.go
aaaaaaaalex: VSCode acting up, thanks

aaaaaaaalex: Found a few other occurrences of tabs instead of spaces in this file's raw json strings, so fixed those up too.

pkg/controller/installation/core_controller.go
aaaaaaaalex: Thanks - I have a personal goal to properly re-read all my comments after I amend the code beneath them :rofl:

aaaaaaaalex: Considering removing the comment altogether, since as you say, it's redundant really, and the _why_ really only amounts to "because we haven't covered the other cases yet" :sweat_smile:

pkg/render/node.go
aaaaaaaalex: Happy to take this onboard, Shaun also suggested:

    I might use a - instead of _ in a dir name (unless there was precedent in an adjacent dir

If that works for you?

pkg/controller/installation/validation.go
aaaaaaaalex: I suppose I could accept it in the operator, ignore it in felix/cni, if thats a worthy paradigm?

pkg/render/node_test.go
aaaaaaaalex: Added some values to validation_test.go now.
Also checking rendering of custom values in node_test.go

api/v1/installation_types.go
caseydavenport: I think "entering the running phase" is too vague - it doesn't drive the point home that the users application code won't even attempt to be launched. 

"from running" is OK, but IMO still not precise - the pod itself will be running (as in, the pod will have established a network namespace with an  ID and an IP address) - what won't be running are the containers within the pod.

pkg/controller/installation/validation.go
caseydavenport: Typically, we would expect that if a config option isn't supported in a certain configuration, that we return an error and prevent installation of the code in that configuration. We should probably do that for this if Windows is enabled in the configuration.

pkg/render/node.go
caseydavenport: :shrug: should be fine either way, main thing is the name. I'd double check the precedent thing though - what other multi-part dir names do we create?

pkg/controller/installation/validation.go
```diff
@@ -355,6 +355,9 @@ func validateCustomResource(instance *operatorv1.Installation) error {
 			if *instance.Spec.CalicoNetwork.PolicySetupTimeoutSeconds < 0 {
 				return fmt.Errorf("Installation CNI spec.PolicySetupTimeoutSeconds negative value is not valid")
 			}
+			if instance.Spec.CalicoNetwork.LinuxDataplane != nil && *instance.Spec.CalicoNetwork.LinuxDataplane != operatorv1.LinuxDataplaneIptables {
```
caseydavenport, line 358: I think we probably want this?

```go
if instance.Spec.CalicoNetwork.LinuxDataplane == nil || *instance.Spec.CalicoNetwork.LinuxDataplane != operatorv1.LinuxDataplaneIptables {
```

Basically if the data plane is unset, then we don't know if it's OK to enable this feature or not and we need to wait until it is.

pkg/controller/installation/validation_test.go
```diff
@@ -791,6 +791,31 @@ var _ = Describe("Installation validation tests", func() {
...
 	})
 
+	Describe("validate CalicoNetwork PolicySetupTimeoutSeconds", func() {
+		It("should return an error when PolicySetupTimeoutSeconds is negative", func() {
+			negative := int32(-1)
+			instance.Spec.CalicoNetwork.PolicySetupTimeoutSeconds = &negative
+			err := validateCustomResource(instance)
+			Expect(err).To(MatchError("Installation CNI spec.PolicySetupTimeoutSeconds negative value is not valid"))
+		})
+		It("should return an error when PolicySetupTimeoutSeconds is set for a non-IPTables dataplane", func() {
```
caseydavenport, line 801: nit, but it's nice to have a newline between `It()` clauses for clarity.

pkg/controller/installation/validation.go
caseydavenport: Probably also want a check on WindowsDataplane here

pkg/controller/installation/validation.go
aaaaaaaalex: Im unfamiliar with this part of the operator resources, but my first thought was to check if the windows dataplane config option was non-nil. But, that would presumably only tell me that _some_ nodes may be windows, and I still may want the feature on for non-windows nodes :thinking: any guidance appreciated

pkg/render/node.go
aaaaaaaalex: Had a quick skim of some reference docs and occurrences of `Mkdir` in our repo and most I saw seemed to use dashes, so I'll go with that.

pkg/controller/installation/validation_test.go
aaaaaaaalex: Fixed

pkg/controller/installation/validation.go
aaaaaaaalex: Similar to the [question](https://github.com/tigera/operator/pull/3186#discussion_r1508020370) I have above, I think this feature would still be desirable in a mixed cluster with windows and linux, if its possible to continue when that is the case. Maybe I could error if _only_ a Windows dataplane is defined, and no Linux one?

api/v1/installation_types.go
aaaaaaaalex: Updated to "running containers" now

pkg/controller/installation/validation.go
caseydavenport: There's no such thing as a "Windows only" cluster - they all have at least one Linux node since the control plane components don't run on Windows.

The usual ideal we strive for is that if configuration is accepted it means it is going to be implemented fully. We don't really have a mechanism to report back "This configuration is sort of working but not entirely"

I think your options are:
- Reject the config outright if Windows is enabled.
- Add support for this feature to Windows.

pkg/controller/installation/validation.go
caseydavenport: > Maybe I could error if only a Windows dataplane is defined, and no Linux one?

Is there a reason we haven't implemented this for Windows?

pkg/render/node.go
```diff
@@ -628,6 +629,10 @@ func (c *nodeComponent) createCalicoPluginConfig() map[string]interface{} {
 
 	apiRoot := c.cfg.K8sServiceEp.CNIAPIRoot()
 
+	LinuxpolicySetupTimeoutSeconds := int32(0)
```
caseydavenport, line 632: This should be a private variable, and `Policy` should be capitalized

`linuxPolicySetupTimeoutSeconds`

caseydavenport (approved):

marvin-tigera:

caseydavenport (approved):

PR #3191, opened 2024-02-22
PolicyRecommendation controller overwrites tigera-ca bundle per tenant

asincu:
## Description

Tigera-ca-bundle is created by the tenant controller for a multi-tenant setup (https://github.com/tigera/operator/blob/master/pkg/controller/secrets/tenant_controller.go#L187-L212). This gets overwritten by PolicyRecommendation Controller. This causes Linseed to reject connection with Elastic.

```
kubectl logs -ncc-tenant-a1q3vp13 tigera-linseed-d95447fc8-8q6zc -ctigera-linseed
2024-02-22 00:38:02.708 [FATAL][1] elastic.go 59: Failed to create Elastic client error=health check timeout: Head "https://es-lss-gcpdev-01.dev.calicocloud.io:443": tls: failed to verify certificate: x509: certificate signed by unknown authority: no Elasticsearch node available
```

Screenshot from 2024-02-21 16-37-04

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/policyrecommendation/policyrecommendation_controller.go
+++ b/pkg/controller/policyrecommendation/policyrecommendation_controller.go
@@ -124,6 +124,7 @@ func Add(mgr manager.Manager, opts options.AddOptions) error {
 			certificatemanagement.CASecretName,
 			render.ManagerInternalTLSSecretName,
 			render.TigeraLinseedSecret,
+			render.PolicyRecommendationTLSSecretName,
 		} {
 			if err = utils.AddSecretsWatch(c, secretName, namespace); err != nil {
 				return fmt.Errorf("policy-recommendation-controller failed to watch the secret '%s' in '%s' namespace: %w", secretName, namespace, err)
@@ -349,31 +350,9 @@ func (r *ReconcilePolicyRecommendation) Reconcile(ctx context.Context, request r
 	}
 
 	logc.V(3).Info("rendering components")
-	policyRecommendationCfg := &render.PolicyRecommendationConfiguration{
-		ClusterDomain:        r.clusterDomain,
-		Installation:         installation,
-		ManagedCluster:       isManagedCluster,
-		PullSecrets:          pullSecrets,
-		Openshift:            r.provider == operatorv1.ProviderOpenShift,
-		UsePSP:               r.usePSP,
-		Namespace:            helper.InstallNamespace(),
-		Tenant:               tenant,
-		BindingNamespaces:    bindNamespaces,
-		PolicyRecommendation: policyRecommendation,
-		ExternalElastic:      r.externalElastic,
-	}
-
-	// Render the desired objects from the CRD and create or update them.
-	component := render.PolicyRecommendation(policyRecommendationCfg)
-
-	if err = imageset.ApplyImageSet(ctx, r.client, variant, component); err != nil {
-		r.status.SetDegraded(operatorv1.ResourceUpdateError, "Error with images from ImageSet", err, logc)
-		return reconcile.Result{}, err
-	}
-
-	components := []render.Component{
-		component,
-	}
+	var policyRecommendationKeyPair certificatemanagement.KeyPairInterface
+	var trustedBundle certificatemanagement.TrustedBundleRO
+	var components []render.Component
 
 	if !isManagedCluster {
 		opts := []certificatemanager.Option{
@@ -406,8 +385,6 @@ func (r *ReconcilePolicyRecommendation) Reconcile(ctx context.Context, request r
 			return reconcile.Result{}, nil
 		}
 
-		trustedBundle := certificateManager.CreateTrustedBundle(managerInternalTLSSecret, linseedCertificate)
-
 		// policyRecommendationKeyPair is the key pair policy recommendation presents to identify itself
 		policyRecommendationKeyPair, err := certificateManager.GetOrCreateKeyPair(r.client, render.PolicyRecommendationTLSSecretName, helper.TruthNamespace(), []string{render.PolicyRecommendationTLSSecretName})
 		if err != nil {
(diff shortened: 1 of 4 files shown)


pkg/controller/policyrecommendation/policyrecommendation_controller.go
```diff
@@ -124,6 +124,7 @@ func Add(mgr manager.Manager, opts options.AddOptions) error {
 			certificatemanagement.CASecretName,
 			render.ManagerInternalTLSSecretName,
 			render.TigeraLinseedSecret,
+			render.PolicyRecommendationTLSSecretName,
```
Josh-Tigera, line 127: We don't need to watch the PolicyRecommendationTLS Secret anymore? How come?

pkg/controller/policyrecommendation/policyrecommendation_controller.go
```diff
@@ -417,20 +394,38 @@ func (r *ReconcilePolicyRecommendation) Reconcile(ctx context.Context, request r
...
-				Namespace:       helper.InstallNamespace(),
-				TruthNamespace:  helper.TruthNamespace(),
-				ServiceAccounts: []string{render.PolicyRecommendationName},
-				KeyPairOptions: []rcertificatemanagement.KeyPairOption{
-					rcertificatemanagement.NewKeyPairOption(policyRecommendationKeyPair, true, true),
-				},
-				TrustedBundle: trustedBundle,
-			}),
-		)
+		if !r.multiTenant {
```
Josh-Tigera, line 397: Can you add a comment here indicating why we have a RW trusted bundle for non-multitenant but RO for multi-tenant? Just to help us in the future.

pkg/controller/policyrecommendation/policyrecommendation_controller.go
```diff
@@ -417,20 +394,38 @@ func (r *ReconcilePolicyRecommendation) Reconcile(ctx context.Context, request r
...
+			trustedBundleRW := certificateManager.CreateTrustedBundle(managerInternalTLSSecret, linseedCertificate)
+			components = append(components,
+				rcertificatemanagement.CertificateManagement(&rcertificatemanagement.Config{
```
Josh-Tigera, line 400: Could the `certificatemanagement.Config` be factored out of the if-else since everything is identical except `TrustedBundle`? That way it's more clear what exactly is being set conditionally and avoids issues with the 2 cases diverging when they shouldn't.

pkg/controller/policyrecommendation/policyrecommendation_controller_test.go
```diff
@@ -412,6 +413,40 @@ var _ = Describe("PolicyRecommendation controller tests", func() {
 				err = test.GetResource(c, &tenantBServiceAccount)
 				Expect(err).ShouldNot(HaveOccurred())
 			})
+
+			It("It should not create the trusted bundle config map as it will be created by the tenant controller", func() {
```
Josh-Tigera, line 417:
```suggestion
			It("should not create the trusted bundle config map as it will be created by the tenant controller", func() {
```

pkg/controller/policyrecommendation/policyrecommendation_controller_test.go
```diff
@@ -412,6 +413,40 @@ var _ = Describe("PolicyRecommendation controller tests", func() {
...
+				// Create the Tenant resources for tenant-a and tenant-b.
```
Josh-Tigera, line 418:
```suggestion
				// Create the Tenant resources
```
There's only one tenant.

pkg/controller/policyrecommendation/policyrecommendation_controller_test.go
```diff
@@ -412,6 +413,40 @@ var _ = Describe("PolicyRecommendation controller tests", func() {
...
+
+				// Now reconcile only tenant A's namespace and expect an error
+				_, err := r.Reconcile(ctx, reconcile.Request{NamespacedName: types.NamespacedName{Namespace: tenantANamespace}})
+				Expect(err).Should(HaveOccurred())
+				Expect(err.Error()).Should(ContainSubstring("CA secret"))
+
+				// Create a CA secret for the test, and create its KeyPair.
+				certificateManagerTenantA, err := certificatemanager.Create(c, nil, "", tenantANamespace, certificatemanager.AllowCACreation(), certificatemanager.WithTenant(tenantA))
+				Expect(err).NotTo(HaveOccurred())
+				Expect(c.Create(ctx, certificateManagerTenantA.KeyPair().Secret(tenantANamespace)))
```
Josh-Tigera, line 443:
```suggestion
				Expect(c.Create(ctx, certificateManagerTenantA.KeyPair().Secret(tenantANamespace)).NotTo(HaveOccurred())
```

pkg/controller/secrets/tenant_controller.go
```diff
@@ -90,6 +90,17 @@ func AddTenantController(mgr manager.Manager, opts options.AddOptions) error {
 		return fmt.Errorf("tenant-controller failed to watch tenant CA Secret %s in all namespace: %w", certificatemanagement.TenantCASecretName, err)
 	}
 
+	// Make a helper for determining which namespaces to use based on tenancy mode.
+	// In multi-tenant mode, we need to watch all namespaces for secrets. In single-tenant mode,
+	// we only need to watch the elasticsearch namespace. Both need tigera-operator.
```
Josh-Tigera, line 95: This comment doesn't seem to align with the use of the `NamespaceHelper`, unless I've misunderstood what's going on. You're setting both single-tenant and multi-tenant namespaces to `""` so regardless of whether `opts.MultiTenant` is true or false you will use namespace `""` in the below watches. https://github.com/tigera/operator/blob/master/pkg/controller/utils/namespace_helper.go#L69

pkg/controller/policyrecommendation/policyrecommendation_controller.go
```diff
@@ -417,20 +394,38 @@ func (r *ReconcilePolicyRecommendation) Reconcile(ctx context.Context, request r
...
+			}
+			components = append(components,
+				rcertificatemanagement.CertificateManagement(&rcertificatemanagement.Config{
+					Namespace:       helper.InstallNamespace(),
+					TruthNamespace:  helper.TruthNamespace(),
+					ServiceAccounts: []string{render.PolicyRecommendationName},
+					KeyPairOptions: []rcertificatemanagement.KeyPairOption{
+						rcertificatemanagement.NewKeyPairOption(policyRecommendationKeyPair, true, true),
+					},
+					TrustedBundle: nil,
```
Josh-Tigera, line 425: Is nil the right value here? I'm not sure what the downstream result of this value being nil would be when setting these environment variables: https://github.com/tigera/operator/blob/master/pkg/render/policyrecommendation.go#L303

I think the difference you're intending is that in multi-tenant mode we simply load the existing trusted bundle created by the tenant controller, while in single or zero-tenant mode we need to create a trusted bundle, but in both cases I think we still need to populate the `TrustedBundle` field no?

pkg/controller/policyrecommendation/policyrecommendation_controller.go
asincu: Actually, I am adding the watch not removing it.

pkg/controller/policyrecommendation/policyrecommendation_controller.go
asincu: For multi-tenant, we should leave this nil because we want to create the secrets that contain the x509 certificates, but not the actual config map `tigera-ca-bundle`. For zero-tenant/single-tenant we want both the secrets and the config map created, because we install in tigera-policy-recommendation namespace.

Josh-Tigera: Oh so when those environment variables are blank it will trigger creation of the x509 certificates?

asincu: Actually, the secrets that hold the certificates.

pkg/controller/secrets/tenant_controller.go
```diff
@@ -90,6 +90,16 @@ func AddTenantController(mgr manager.Manager, opts options.AddOptions) error {
 		return fmt.Errorf("tenant-controller failed to watch tenant CA Secret %s in all namespace: %w", certificatemanagement.TenantCASecretName, err)
 	}
 
+	// Make a helper for determining which namespaces to use based on tenancy mode.
+	// In multi-tenant mode, we need to watch all namespaces for secrets.
+	helper := utils.NewNamespaceHelper(opts.MultiTenant, "", "")
```
Josh-Tigera, line 95: Maybe just remove the Namespace helper entirely and use `""` in the watches directly since the helper has no effect?

pkg/controller/policyrecommendation/policyrecommendation_controller.go
```diff
@@ -439,6 +432,30 @@ func (r *ReconcilePolicyRecommendation) Reconcile(ctx context.Context, request r
...
+
+	// Render the desired objects from the CRD and create or update them.
+	component := render.PolicyRecommendation(policyRecommendationCfg)
+
+	if err = imageset.ApplyImageSet(ctx, r.client, variant, component); err != nil {
+		r.status.SetDegraded(operatorv1.ResourceUpdateError, "Error with images from ImageSet", err, logc)
+		return reconcile.Result{}, err
+	}
+
+	components = append(components, component)
```
rene-dekker, line 458: By adding the component after the other components, doesn't the operator try to render tls assets before the namespace exists on the first ever iteration? I think that would create an unnecessary error in the logs / tigerastatus.

pkg/controller/secrets/tenant_controller.go
```diff
@@ -90,6 +90,13 @@ func AddTenantController(mgr manager.Manager, opts options.AddOptions) error {
 		return fmt.Errorf("tenant-controller failed to watch tenant CA Secret %s in all namespace: %w", certificatemanagement.TenantCASecretName, err)
 	}
 
+	if err = utils.AddConfigMapWatch(c, certificatemanagement.TrustedCertConfigMapName, "", &handler.EnqueueRequestForObject{}); err != nil {
```
rene-dekker, line 93: First of all, you adding this statement twice.
Secondly, why would we want to watch the things that we create? What is there to gain by this?

rene-dekker (approved):
lgtm

PR #3229, opened 2024-03-08
Disable CGO when build operator

hjiawei:
## Description

This changeset disables CGO when building operator binary for all supported architectures. We shelved FIPS support so bringing back tls module from Go stdlib reenabled TLS 1.3 support.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -41,7 +41,4 @@ COPY --from=source / /
 
 USER 10001:10001
 
-# The exec form of ENTRYPOINT does not invoke a command shell.
-# This means that normal shell processing does not happen, so will not
-# do variable substitution. Using this form instead of passing $OPERATOR.
 ENTRYPOINT ["/usr/bin/operator"]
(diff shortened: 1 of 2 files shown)


build/Dockerfile
```diff
@@ -41,7 +41,4 @@ COPY --from=source / /
 
 USER 10001:10001
 
-# The exec form of ENTRYPOINT does not invoke a command shell.
-# This means that normal shell processing does not happen, so will not
-# do variable substitution. Using this form instead of passing $OPERATOR.
```
hjiawei, line 46: This comment doesn't seem to apply any more so removed.

rene-dekker:
We are removing FIPS for enterprise. I still need to find out the exact requirements for open source. I think that the operator will probably continue to be able to install FIPS for open source.

exfly:
issue: https://github.com/tigera/operator/issues/3461

> What I think we may want to consider is disabling CGO/boringcrypto by default but continue to allow users to build the image with a FIPS=true flag

Can we disable CGO/boringcrypto by default

PR #3251, opened 2024-03-18
Remove special key-cert-provisioner image code, now that it is in the…

rene-dekker:
Both for OSS and EE, the key-cert-provisioner can now be pulled from the same registry and with the same tag as other images (like node, typha, apiserver), removing the need for a special common_versions.yml. This PR removes it and should be picked to release-v1.34.


radTuti (approved):

pkg/components/images_test.go
```diff
@@ -61,20 +63,22 @@ var _ = Describe("test GetReference", func() {
...
+			Entry("a CSR init image correctly", ComponentCSRInitContainer, "calico/key-cert-provisioner"),
+			Entry("a CSR init image correctly", ComponentCSRInitContainerPrivate, "tigera/key-cert-provisioner"),
 		)
 	})
 
 	Context("image prefix override", func() {
 		DescribeTable("should render",
 			func(c component, image string) {
-				Expect(GetReference(c, "quay.io/", "", "pref", nil)).To(Equal(fmt.Sprintf("quay.io/%s:%s", image, c.Version)))
+				Expect(GetReference(c, "quay.io/", "", "pref/", nil)).To(Equal(fmt.Sprintf("quay.io/%s:%s", image, c.Version)))
```
tmjd, line 74: This might work but it doesn't match the expectation for how the prefix field is expected to be used. The prefix is expected to be a string added to the start of the image name. This test change suggests is is for setting the path.
The image format is documented as `<registry><imagePath>/<imagePrefix><imageName>:<image-tag>`

pkg/tls/certificatemanagement/csr.go
```diff
@@ -92,6 +92,15 @@ func CreateCSRInitContainer(
...
 func ResolveCSRInitImage(inst *operatorv1.InstallationSpec, is *operatorv1.ImageSet) (string, error) {
+	if inst.Variant == operatorv1.TigeraSecureEnterprise {
+		return components.GetReference(
+			components.ComponentCSRInitContainerPrivate,
+			inst.Registry,
+			inst.ImagePath,
+			inst.ImagePrefix,
+			is,
+		)
+	}
```
tmjd, line 103: Don't we need an else case here to use the Calico image?

pkg/components/enterprise.go
```diff
@@ -323,6 +329,7 @@ var (
 		ComponentComplianceReporter,
 		ComponentComplianceServer,
 		ComponentComplianceSnapshotter,
+		ComponentCSRInitContainerPrivate,
```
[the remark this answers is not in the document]
rene-dekker: I see there is also:
- ComponentFlexVolumePrivate
- ComponentCSIPrivate,
- ComponentCSINodeDriverRegistrarPrivate,

Do you want me to change all of them?

pkg/components/images_test.go
rene-dekker: Tbh I found it unpleasant to read the code easily, which motivated the change for me. I didn't think it broke the contract with the API, I just picked something that I preferred reading.
That said, I will revert it.

pkg/tls/certificatemanagement/csr.go
rene-dekker: Line 96 returns, making the `else` redundant. This is less code and I'd argue more readable.

pkg/components/enterprise.go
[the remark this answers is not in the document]
tmjd: If you want to, but I wouldn't require you to do that. But we should keep an eye out for these in the future.

pkg/components/images_test.go
tmjd: I'd suggest `prefix-` if you think having a separator is more readable.

pkg/tls/certificatemanagement/csr.go
tmjd: Oh I see :facepalm: . What you have is good.

hack/gen-versions/calico.go.tpl
```diff
@@ -149,6 +156,7 @@ var (
 		ComponentCalicoCNI,
 		ComponentCalicoCNIFIPS,
 		ComponentCalicoCNIWindows,
+        ComponentCalicoCSRInitContainer,
```
tmjd, line 159: Can you check the indentation on this? I've seen github show indentation oddly sometimes so I'm not sure if it is github or the changes.

pkg/components/enterprise.go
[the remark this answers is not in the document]
tmjd: I'm pretty sure you changed these to have Calico or Tigera in the name but it looks like you've changed these back to Private, was that on accident?

tmjd (approved):
LGTM

PR #3280, opened 2024-04-01
Fix IP pool name defaulting

caseydavenport:
## Description

Noticed that in some cases, creating a new controller was generating IP
pools that didn't have the old name. Previously, we would create pools
with `default-ipvX-ippool`. 

This PR aims to be backwards compatible and use that name format for the
following cases (the only possible scenarios prior to v1.34):

- There is only a single IP pool declared.
- This is a dual-stack cluster

Otherwise, IP pool name must either be provided by the user or it will
be defaulted based on the IP pool's CIDR.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


tmjd (approved):
LGTM

PR #3294, opened 2024-04-03
Add egress rule to allow dashboards to connect to external Kibana

asincu:
## Description

Single tenant with external Kibana need to access the external Kibana URL to setup up the dashboards in the tenant namespace. We need to create an egress rule to match that.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/logstorage/dashboards/dashboards_controller.go
+++ b/pkg/controller/logstorage/dashboards/dashboards_controller.go
@@ -17,7 +17,10 @@ package dashboards
 import (
 	"context"
 	"fmt"
+	"math"
 	"net/url"
+	"strconv"
+	"strings"
 
 	"github.com/tigera/operator/pkg/controller/logstorage/initializer"
 
@@ -305,6 +308,13 @@ func (d DashboardsSubController) Reconcile(ctx context.Context, request reconcil
 		}
 	}
 
+	kibanaDomain, kibanaPortAsInt, err := parseDomainAndPort(kibanaHost, kibanaPort)
+	if err != nil {
+		reqLogger.Error(err, "Failed to extract domain or unit16 port for Kibana")
+		d.status.SetDegraded(operatorv1.ResourceValidationError, "Failed to parse kibana domain or port", err, reqLogger)
+		return reconcile.Result{}, nil
+	}
+
 	// Query the username and password this Dashboards Installer instance should use to authenticate with Elasticsearch.
 	// For multi-tenant systems, credentials are created by the elasticsearch users controller.
 	// For single-tenant system, these are created by es-kube-controllers.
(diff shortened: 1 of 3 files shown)


pkg/render/logstorage/dashboards/dashboards.go
```diff
@@ -148,11 +151,26 @@ func (d *dashboards) resources() []client.Object {
 func (d *dashboards) AllowTigeraPolicy() *v3.NetworkPolicy {
 	egressRules := []v3.Rule{}
 	egressRules = networkpolicy.AppendDNSEgressRules(egressRules, d.cfg.Installation.KubernetesProvider == operatorv1.ProviderOpenShift)
-	egressRules = append(egressRules, v3.Rule{
-		Action:      v3.Allow,
-		Protocol:    &networkpolicy.TCPProtocol,
-		Destination: render.KibanaEntityRule,
-	})
+	if d.cfg.ExternalKibanaClientSecret != nil {
+		issuerDomain := strings.TrimPrefix(d.cfg.KibanaHost, "https://")
```
caseydavenport, line 155: What do we mean by "issuer" domain here? Who has issued what in this scenario?

caseydavenport: Maybe better just to call it `kibanaDomain`?

pkg/render/logstorage/dashboards/dashboards.go
```diff
@@ -148,11 +152,33 @@ func (d *dashboards) resources() []client.Object {
...
-		Protocol:    &networkpolicy.TCPProtocol,
-		Destination: render.KibanaEntityRule,
-	})
+	if d.cfg.ExternalKibanaClientSecret != nil {
+		const defaultKibanaPort uint64 = 443
+		kibanaDomain := strings.TrimPrefix(d.cfg.KibanaHost, "https://")
+		kibanaDomain = strings.TrimSuffix(kibanaDomain, "/")
+		kibanaPort, err := strconv.ParseUint(d.cfg.KibanaPort, 10, 16)
+		if err != nil {
+			kibanaPort = defaultKibanaPort
```
caseydavenport, line 161: We definitely need to log out if there is an error here

caseydavenport: If we need to support "no port specified" then we should determine that by parsing the string and deciding that there was no port given - we want to be able to tell when `ParseUint` fails for a legitimate reason.

pkg/controller/logstorage/dashboards/dashboards_controller.go
```diff
@@ -347,6 +357,8 @@ func (d DashboardsSubController) Reconcile(ctx context.Context, request reconcil
 		KibanaHost:                 kibanaHost,
 		KibanaScheme:               kibanaScheme,
 		KibanaPort:                 kibanaPort,
+		KibanaDomain:               kibanaDomain,
+		KibanaPortAsInt:            kibanaPortAsInt,
```
caseydavenport, line 361: We definitely don't need a KibanaPort and KibanaPortAsInt - they should always be the same value, and the render code can just `fmt.Sprintf("%s", cfg.KibanaPort)` where it needs it as a string.

caseydavenport (approved):
One comment but I don't need to review again after

caseydavenport (approved):

PR #3296, opened 2024-04-04
The dashboard name has changed in the installer repo, change it accor…

rene-dekker:
The dashboard name has changed in the installer repo, change it accordingly in here, so that the users are redirected to the right page after logging in.


vara2504 (approved):
lgtm

tmjd (approved):
LGTM

PR #3306, opened 2024-04-09
Shorten kibana session

rene-dekker:
See: https://www.elastic.co/guide/en/kibana/7.17/xpack-security-session-management.html

The UX is as follows: after you don't do anything for a while and you have only a minute left, kibana will display a box in the bottom right informing you that the session expires. You can click a button there to prolong it. If you don't you will be logged out.


tmjd (approved):
LGTM

PR #3322, opened 2024-04-22
Master role deprecation 

lobuhi:
Master role was deprecated in v1.24 and became just control-plane
https://kubernetes.io/blog/2022/04/07/upcoming-changes-in-kubernetes-1-24/


caseydavenport:
Looks like we already have updated to include control-plane

PR #3325, opened 2024-04-23
[RS-1818] Update ILM policy to keep warm tigera_secure_ee_events indices writable

gantony:
## Description

Update ILM policy to keep warm `tigera_secure_ee_events` indices writable to enable, for example, dismissing older events that were created before an index rolloed over (e.g. during upgrade).

Because we don't know the impact of that change on ES performance, we limit this setting to the only index we know this is needed (`tigera_secure_ee_events`).

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/utils/test_files/02_put_policy.json
+++ b/pkg/controller/utils/test_files/02_put_policy.json
@@ -20,7 +20,6 @@
       },
       "warm" : {
         "actions" : {
-          "readonly" : { },
           "set_priority" : {
             "priority" : 50
           }
--- a/pkg/controller/utils/elasticsearch.go
+++ b/pkg/controller/utils/elasticsearch.go
@@ -477,11 +477,7 @@ func buildILMPolicy(totalEsStorage int64, totalDiskPercentage float64, percentOf
 					},
 				},
 				"warm": map[string]interface{}{
-					"actions": map[string]interface{}{
-						"set_priority": map[string]interface{}{
-							"priority": 50,
-						},
-					},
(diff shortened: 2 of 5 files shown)


peterkellydev (approved):

pkg/controller/utils/test_files/02_put_policy.json
tmjd, line 1: Do you know if this file is being used? Because unless it was being used only to check on the events policy I wouldn't expect to see this change.
I think we should have a test that checks the events policy and another policy anyway, so should need a version of this file with readonly and one without.

gantony: Yes and it actually caught an error I had last night with it (I ignored my new parameter :sweat_smile:).

There are 2 files: `test_files/01_put_policy.json` and `test_files/02_put_policy.json`. First one is used when creating a policy for an index and other one used when updating a policy for same index (a test index in this case).

In `pkg/controller/utils/elasticsearch_test.go` I create a policy for test index with `readOnlyAfterRollover` `true` and then update it with `false` so I'm happy the test covers the new logic (it helped me catch this issue: https://github.com/tigera/operator/pull/3325/commits/aaa55150deb0f03cc27c8744f6867e9e6b66aefa)

tmjd: Ah ok, I see the use of those files are embedded in the mockElastiClient and the version is selected based on the `newPolicies` flag.

PR #3336, opened 2024-04-30
[RS-1818] Update ILM policy when warm index readonly setting changes

gantony:
## Description

Further testing around https://github.com/tigera/operator/pull/3325 showed that the operator does not update the ILM policy when the value of `readOnlyAfterRollover` changes. 

The code responsible for updating the ILM policy only do so if there is a change warranting an update. That check does not generically compare the old policy to the new one, but instead only looks at a handful of relevant parameters and tirggers an update only when one of these relevant parameters changes. This PR adds `readOnlyAfterRollover` to that logic and updates the tests to cover that logic.

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/utils/elasticsearch.go
+++ b/pkg/controller/utils/elasticsearch.go
@@ -55,6 +55,11 @@ type Policy struct {
 				}
 			}
 		}
+		Warm struct {
+			Actions struct {
+				Readonly *struct{} `json:"readonly,omitempty"`
+			}
+		}
 		Delete struct {
 			MinAge string `json:"min_age"`
 		}
@@ -62,10 +67,11 @@ type Policy struct {
 }
 
 type policyDetail struct {
-	rolloverAge  string
-	rolloverSize string
-	deleteAge    string
-	policy       map[string]interface{}
+	rolloverAge           string
+	rolloverSize          string
+	deleteAge             string
+	readOnlyAfterRollover bool
+	policy                map[string]interface{}
 }
 
 type logrWrappedESLogger struct{}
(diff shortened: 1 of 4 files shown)


electricjesus (approved):

pkg/controller/utils/elasticsearch.go
```diff
@@ -588,20 +596,21 @@ func getESRoots(ctx context.Context, client client.Client) (*x509.CertPool, erro
...
-		return "", "", "", err
+		return "", "", "", true, err
 	}
 
 	currentMaxAge := existingPolicy.Phases.Hot.Actions.Rollover.MaxAge
 	currentMaxSize := existingPolicy.Phases.Hot.Actions.Rollover.MaxSize
 	currentMinAge := existingPolicy.Phases.Delete.MinAge
-	return currentMaxAge, currentMaxSize, currentMinAge, nil
+	readOnlyAfterRollover := existingPolicy.Phases.Warm.Actions.Readonly != nil
+	return currentMaxAge, currentMaxSize, currentMinAge, readOnlyAfterRollover, nil
```
Brian-McM, line 613: Do you know why we don't just do a rollover if anything has changed, i.e. do some sort of deep equal?

Brian-McM (approved):

pkg/controller/utils/elasticsearch.go
gantony: That was my thought too when I first saw this, seems like it would be a lot simpler and better long term. But a bigger change could expose me to more unforeseen issues and wanted to keep the change straightforward.

The main issue I could foresee would be sensitivity to json formatting or ordering mismatch could cause the comparison to "fail" and the policy to be continuously updated. This could be avoided with proper testing. 

Would be happy to clean that up further if you think that's a good change?

PR #3344, opened 2024-05-01
update script to generate release notes from milestone

radTuti:
## Description

This updates the scripts for generating release notes based on milestone.

Had to do a bit of refactor to:
- use black formatter
- dockerize script
- adhere to pylint: include types to help with docstrings, rename file

**NOTE**: There is work planned in future sprints for switching to Go but this prevent having issues between now and when that change goes in

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -477,8 +477,8 @@ maybe-build-release:
 	./hack/maybe-build-release.sh
 
 release-notes: var-require-all-VERSION-GITHUB_TOKEN clean
-	@python3 -m pip install PyYAML PyGithub==2.3.0 && \
-		GITHUB_TOKEN=$(GITHUB_TOKEN) VERSION=$(VERSION) ./generate-release-notes.py
+	@docker build -t tigera/release-notes -f build/Dockerfile.release-notes .
+	@docker run --rm -v $(CURDIR):/workdir -e	GITHUB_TOKEN=$(GITHUB_TOKEN) -e VERSION=$(VERSION) tigera/release-notes
 
 ## Tags and builds a release from start to finish.
 release: release-prereqs
--- a/Makefile
+++ b/Makefile
@@ -476,6 +476,10 @@ endif
 maybe-build-release:
 	./hack/maybe-build-release.sh
 
+release-notes: var-require-all-VERSION-GITHUB_TOKEN clean
+	@python3 -m pip install PyYAML PyGithub==2.3.0 && \
+		GITHUB_TOKEN=$(GITHUB_TOKEN) VERSION=$(VERSION) ./generate-release-notes.py
+
 ## Tags and builds a release from start to finish.
 release: release-prereqs
 ifneq ($(VERSION), $(GIT_VERSION))
--- a/generate-release-notes.py
+++ b/generate-release-notes.py
@@ -1,20 +1,39 @@
 #!/usr/bin/env python3
-from github import Github  # https://github.com/PyGithub/PyGithub
-import yaml
+# pylint: disable=missing-module-docstring
+# pylint: disable=missing-class-docstring
+# pylint: disable=missing-function-docstring
 import os
 import re
 import io
 import datetime
 
+# To install, run python3 -m pip install PyYAML
+import yaml
+
+# To install, run python3 -m pip install PyGithub==2.3.0
+from github import Github, Auth  # https://github.com/PyGithub/PyGithub
+
+# Validate required environment variables
+GITHUB_TOKEN = os.environ.get("GITHUB_TOKEN")
+assert GITHUB_TOKEN, "GITHUB_TOKEN must be set"
+VERSION = os.environ.get("VERSION")
+assert os.environ.get("VERSION"), "VERSION must be set"
+assert VERSION.startswith("v"), "VERSION must start with 'v'"
+
 # First create a Github instance. Create a token through Github website - provide "repo" auth.
-g = Github(os.environ.get('GITHUB_TOKEN'))
+auth = Auth.Token(os.environ.get("GITHUB_TOKEN"))
+g = Github(auth=auth)
 
 # The milestone to generate notes for.
-assert os.environ.get('VERSION')
-MILESTONE=os.environ.get('VERSION')
+MILESTONE = VERSION
 
 # The file where we'll store the release notes.
-FILENAME="%s-release-notes.md" % MILESTONE
+FILENAME = f"{VERSION}-release-notes.md"
+
+
+class ReleaseNoteError(Exception):
+    pass
+
 
 # Returns a dictionary where the keys are repositories, and the values are
 # a list of issues in the repository which match the milestone and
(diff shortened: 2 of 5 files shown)


generate-release-notes.py
```diff
@@ -60,28 +88,37 @@ def kind(issue):
...
+                file.write(
+                    f" - {note} [#{issue.number}]({issue.html_url}) (@{issue.user.login})\n"
+                )
+    file.write("\n")
+
 
 def calico_version():
-    v = yaml.safe_load(open('config/calico_versions.yml', 'r'))
-    return v['title']
+    v = yaml.safe_load(open("config/calico_versions.yml", "r", encoding="utf-8"))
```
danudey, line 114: Is there a specific need to specify the encoding? Is `utf-8` not the default?

generate-release-notes.py
```diff
@@ -103,25 +140,27 @@ def enterprise_version():
             other.append(i)
 
     # Write release notes out to a file.
-    with io.open(FILENAME, "w", encoding='utf-8') as f:
-        f.write(u"%s\n\n" % date)
+    with io.open(FILENAME, "w", encoding="utf-8") as f:
```
danudey, line 143: If we're changing this line anyway:

1. We can use `open()` instead of `io.open()`
2. We don't need to explicitly set encoding, since `utf-8` is the default

generate-release-notes.py
```diff
@@ -28,14 +47,22 @@ def issues_in_milestone():
...
+            )
+            for issue in milestone_issues:
+                pr = issue.as_pull_request()
                 if pr.merged:
                     # Filter out PRs which are closed but not merged.
-                    issues.append(i)
+                    issues.append(issue)
+                else:
+                    print(f"WARNING: {pr.number} is not merged, skipping")
+            if len(issues) == 0:
```
danudey, line 62: This should just be `if issues:`

Also, I would move this above the loop for clarity's sake — weird to loop over something and then check if it's empty afterwards.

generate-release-notes.py
```diff
@@ -1,20 +1,39 @@
...
+
 # First create a Github instance. Create a token through Github website - provide "repo" auth.
-g = Github(os.environ.get('GITHUB_TOKEN'))
+auth = Auth.Token(os.environ.get("GITHUB_TOKEN"))
+g = Github(auth=auth)
 
 # The milestone to generate notes for.
-assert os.environ.get('VERSION')
-MILESTONE=os.environ.get('VERSION')
+MILESTONE = VERSION
```
danudey, line 28: If `MILESTONE` is always going to == `VERSION` we should replace `MILESTONE` with `VERSION` for clarity.

generate-release-notes.py
```diff
@@ -44,14 +71,15 @@ def extract_release_notes(issue):
     # Look for a release note section in the body.
     matches = None
     if issue.body:
-        matches = re.findall(r'```release-note(.*?)```', issue.body, re.DOTALL)
+        matches = re.findall(r"```release-note(.*?)```", str(issue.body), re.DOTALL)
```
danudey, line 74: Is the `str()` call necessary? What type is `issue.body` normally?

Makefile
```diff
@@ -476,6 +476,10 @@ endif
 maybe-build-release:
 	./hack/maybe-build-release.sh
 
+release-notes: var-require-all-VERSION-GITHUB_TOKEN clean
+	@python3 -m pip install PyYAML PyGithub==2.3.0 && \
```
danudey, line 480: I'm extremely wary about modifying the user's Python install without asking. IIUC this will also break on default installations of Ubuntu >= 23.10 due to the pip issue we've seen with Banzai.

We can create a temporary (?) venv to do this relatively simply, with something like the solution specified here:

https://stackoverflow.com/a/46188210

generate-release-notes.py
```diff
@@ -1,20 +1,39 @@
 #!/usr/bin/env python3
-from github import Github  # https://github.com/PyGithub/PyGithub
-import yaml
+# pylint: disable=missing-module-docstring
```
danudey, line 2: My preference would be to add docstrings to the relevant functions, even though they're mostly self-evident. Not a dealbreaker though.

generate-release-notes.py
```diff
@@ -1,20 +1,39 @@
...
+# pylint: disable=missing-class-docstring
+# pylint: disable=missing-function-docstring
 import os
 import re
 import io
```
danudey, line 7: If we remove use of `io.open()` we can remove this useless import.

generate-release-notes.py
```diff
@@ -28,14 +47,22 @@ def issues_in_milestone():
...
+                milestone=m, state="closed", labels=["release-note-required"]
+            )
+            for issue in milestone_issues:
+                pr = issue.as_pull_request()
                 if pr.merged:
                     # Filter out PRs which are closed but not merged.
-                    issues.append(i)
+                    issues.append(issue)
+                else:
+                    print(f"WARNING: {pr.number} is not merged, skipping")
```
danudey, line 61: Does this need to be a warning? If the PR is closed but not merged then we've thrown it away, we probably don't need to be reminded of that at release time.

generate-release-notes.py
```diff
@@ -1,20 +1,39 @@
...
-import yaml
+# pylint: disable=missing-module-docstring
+# pylint: disable=missing-class-docstring
+# pylint: disable=missing-function-docstring
 import os
 import re
 import io
 import datetime
 
+# To install, run python3 -m pip install PyYAML
```
caseydavenport, line 10: We should be using a requirements.txt to do this, rather than having people install modules individaully. I certainly don't want to read the code in order to figure out how to prep my environment.

Ideally, this uses a virtualenv and/or a docker environment, and doesn't require me to even know I'm using Python.

generate-release-notes.py
```diff
@@ -1,20 +1,39 @@
...
+import yaml
+
+# To install, run python3 -m pip install PyGithub==2.3.0
```
caseydavenport, line 13: Same here

generate-release-notes.py
```diff
@@ -28,14 +47,22 @@ def issues_in_milestone():
         if m.title == MILESTONE:
             # Found the milestone in this repo - look for issues (but only
             # ones that have been closed!)
+            print(f"  found milestone {m.title}")
             issues = []
-            for i in repo.get_issues(milestone=m, state="closed", labels=['release-note-required']):
-                pr = i.as_pull_request()
+            milestone_issues = repo.get_issues(
```
caseydavenport, line 52: Why this change to do this in two lines?

We shouldn't be making changes that don't have a functional impact without good reason - makes it harder to review, and muddles the history up in git.

Makefile
caseydavenport: Yeah, this is a no-go for me. 

We should really be using a venv or a docker container for this

caseydavenport:
@radTuti could you update the description and title so that they explain the goals of this PR? It's hard to review without understanding the motivation. Most of this looks like refactoring without functional changes? So hard to tell what's important and what isn't.

generate-release-notes.py
radTuti: this is from running black formatter

generate-release-notes.py
radTuti: this is to be able to warn about open issues attached to the milestone. I will update to only warn for those

generate-release-notes.py
radTuti: you are conflating `milestone_issues` with `issues`. It will be a noop if before the loop meanwhile at this point is to see if there are issues to generate a release note for

generate-release-notes.py
radTuti: it is necessary

generate-release-notes.py
radTuti: I will rather not rely on defaults in this case

generate-release-notes.py
caseydavenport: If we're going to start to adhere to a style guide we probably want to include it as part of CI to prevent non-conformant changes getting merged.

That said, eventually this should all be Go code if I had my way, so maybe not that important to do now so long as we can see that on the horizon...

Makefile
radTuti: switched to a docker container

generate-release-notes.py
radTuti: I was templted to switch to Go after the latest round of reviews but I wanted to get this script meged with a good working base. I will like to park enforce a format to when we swich to Golang sicne that is part of my objective this quarter

caseydavenport: Golang is shipped with a linter - `gofmt` that is standard and we run against all of our code. It will happen naturally when we switch this to Go.

caseydavenport (approved):

PR #3348, opened 2024-05-06
Install Kibana per tenant

asincu:
## Description

Kibana will be installed per tenant using ECK operator. Both Kibana and ECK will be deployed by external elastic search controller. ECK operator will be configured to watch in all namespaces. In a tenant namespace we will create a Kibana CR that contains an additional container "challenger". Challenger connects to external elastic via mTLS, while Kibana per tenant connects to Elastic using locahost and no TLS. An kibana admin user will be created in external elasticsearch and referenced via SecureSettings field from the Kibana spec. This secret will contain only elasticsearch.password that in mounted by kibana keystore container at startup. This secret will be created by users controller.

Dashboards and ES Proxy will need to access kibana per tenant.

This PR also adds the ability to install or not install kibana per tenant.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


pkg/controller/logstorage/users/users_controller.go
```diff
@@ -351,6 +386,20 @@ func (r *UserController) createUserLogin(ctx context.Context, elasticEndpoint st
 	return nil
 }
 
+func (r *UserController) createUser(ctx context.Context, elasticEndpoint string, user *utils.User, reqLogger logr.Logger) error {
```
tmjd, line 389: I was looking to see if we needed both createUser and createUserLogin and I think the answer is yes but it looks like we could have createUserLogin call createUser, that should allow removing some duplicate code.

pkg/controller/logstorage/users/users_controller.go
```diff
@@ -418,23 +467,37 @@ func (r *UsersCleanupController) cleanupStaleUsers(ctx context.Context, logger l
...
+			}
+		}
 
-				// Remove the finalizer from the tenant to allow it to be deleted.
-				if stringsutil.StringInSlice(userCleanupFinalizer, t.GetFinalizers()) {
-					t.SetFinalizers(stringsutil.RemoveStringInSlice(userCleanupFinalizer, t.GetFinalizers()))
-					if err = r.client.Update(ctx, &t); err != nil {
-						logger.Error(err, "Failed to remove user cleanup finalizer from tenant")
-					}
+		if deletedUsers > 0 {
```
tmjd, line 491: I guess this is maintaining the same behavior as before but it seems like we should only remove the finalizer if all 3 users have been removed. I think if removal of any of them failed then the finalizer should remain. WDYT?

pkg/render/common/networkpolicy/networkpolicy.go
```diff
@@ -290,6 +290,10 @@ func (h *NetworkPolicyHelper) ComplianceReporterSourceEntityRule() v3.EntityRule
 	return CreateSourceEntityRule(h.namespace("tigera-compliance"), "compliance-reporter")
 }
 
+func (h *NetworkPolicyHelper) KibanaEntityRule() v3.EntityRule {
```
tmjd, line 293: I think this function is only used for Multi-tenant kibana and in that case doesn't this rule need to use the tenant namespace?

PR #3349, opened 2024-05-08
set mount type to DirectoryOrCreate for hostPath needed by Calico

mazdakn:
## Description
Fixes: https://github.com/projectcalico/calico/issues/8773

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/node.go
+++ b/pkg/render/node.go
@@ -1056,7 +1056,7 @@ func (c *nodeComponent) nodeVolumes() []corev1.Volume {
 	} else {
 		volumes = append(volumes,
 			c.varRunCalicoVolume(),
-			corev1.Volume{Name: "var-lib-calico", VolumeSource: corev1.VolumeSource{HostPath: &corev1.HostPathVolumeSource{Path: "/var/lib/calico"}}},
+			corev1.Volume{Name: "var-lib-calico", VolumeSource: corev1.VolumeSource{HostPath: &corev1.HostPathVolumeSource{Path: "/var/lib/calico", Type: &dirOrCreate}}},
 		)
 	}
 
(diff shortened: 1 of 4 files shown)


pkg/render/node.go
```diff
@@ -1082,7 +1082,7 @@ func (c *nodeComponent) nodeVolumes() []corev1.Volume {
 	if c.cfg.Installation.CNI.Type == operatorv1.PluginCalico {
 		// Determine directories to use for CNI artifacts based on the provider.
 		cniNetDir, cniBinDir, cniLogDir := c.cniDirectories()
-		volumes = append(volumes, corev1.Volume{Name: "cni-bin-dir", VolumeSource: corev1.VolumeSource{HostPath: &corev1.HostPathVolumeSource{Path: cniBinDir}}})
+		volumes = append(volumes, corev1.Volume{Name: "cni-bin-dir", VolumeSource: corev1.VolumeSource{HostPath: &corev1.HostPathVolumeSource{Path: cniBinDir, Type: &dirOrCreate}}})
```
caseydavenport, line 1085: I'm curious why we need this on the cni bin dir, but not the other CNI volume mounts as well?

mazdakn: I think in this case, the user reporter these 3 dirs as they hit the error with these 3 ones. However, in other setups maybe another user might hit the issue with the other paths. Now, we can either fix these 3 as we know the issue has happened and wait for other incidents, or maybe we can add this property for all directories that we think might be affected. 
It's good to be proactive and set all the really needs it, but it might needs a bit more investigation to realise which ones really needs it. WDYT?

mazdakn: btw, the same question applies to calico node mounts. Maybe we should update all of them.

mazdakn: @caseydavenport I don't know about the purpose of these directories. Given you know more about Calico CNI, does it make sense to set the same `Type` to the other directories? If so, I'll add it.

caseydavenport: I suppose let's leave them as is for now just to keep the changes minimal to fix the originally reported issue. I suspect these probably suffer the same problem in some cases, but we've never heard about it.

caseydavenport (approved):

PR #3352, opened 2024-05-17
Force a Linseed restart when the credentials to connect to Elastic change

asincu:
## Description

We encountered a situation where linseed user was recreated, but Linseed was still referencing the old password causing the following errors to show up on all APIs:

```
[status 500] server error: elastic: Error 401 (Unauthorized): unable to authenticate user [tigera-ee-linseed-yeidobci-secure] for REST request [/tigera_secure_ee_benchmark_results.yeidobci.cluster.%2A/_search] 
```

After restart, Linseed errors were resolved. At further inspection, the container that got restarted was referencing the new password, while the one that did not get restarted was still using the old password.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/logstorage/linseed/linseed_controller.go
+++ b/pkg/controller/logstorage/linseed/linseed_controller.go
@@ -136,6 +136,7 @@ func Add(mgr manager.Manager, opts options.AddOptions) error {
 		render.LinseedTokenSecret,
 		monitor.PrometheusClientTLSSecretName,
 		render.ElasticsearchLinseedUserSecret,
+		render.ElasticsearchLinseedUserSecret,
 	}
 
 	// Determine namespaces to watch.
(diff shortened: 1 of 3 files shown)


tmjd (approved):
LGTM
I don't understand why we didn't see this problem before.

pkg/render/logstorage/linseed/linseed_test.go
```diff
@@ -150,6 +164,7 @@ var _ = Describe("Linseed rendering tests", func() {
 
 			// The deployment should have the hash annotation set, as well as a volume and volume mount for the client secret.
 			Expect(d.Spec.Template.Annotations["hash.operator.tigera.io/elastic-client-secret"]).To(Equal("ae1a6776a81bf1fc0ee4aac936a90bd61a07aea7"))
+			Expect(d.Spec.Template.Annotations["hash.operator.tigera.io/elastic-client-credential-secret"]).To(Equal("465c25d580ea36d8e7cda470a0c34afd05eae6f7:wq"))
```
tmjd, line 167: I didn't know you use vim :smile: . (I see a `:wq` which is causing CI to fail)

pkg/controller/logstorage/linseed/linseed_controller.go
```diff
@@ -136,6 +136,7 @@ func Add(mgr manager.Manager, opts options.AddOptions) error {
 		render.LinseedTokenSecret,
 		monitor.PrometheusClientTLSSecretName,
 		render.ElasticsearchLinseedUserSecret,
+		render.ElasticsearchLinseedUserSecret,
```
Glen-Tigera, line 139: These look the same, is that expected?

tmjd: I just realized I hadn't commented on that, yeah I agree. We shouldn't need that repeated here.

pkg/render/logstorage/linseed/linseed.go
```diff
@@ -427,6 +432,9 @@ func (l *linseed) linseedDeployment() *appsv1.Deployment {
 	if l.cfg.ElasticClientSecret != nil {
 		annotations["hash.operator.tigera.io/elastic-client-secret"] = rmeta.SecretsAnnotationHash(l.cfg.ElasticClientSecret)
 	}
+	if l.cfg.ElasticClientCredentialsSecret != nil {
+		annotations["hash.operator.tigera.io/elastic-client-credential-secret"] = rmeta.SecretsAnnotationHash(l.cfg.ElasticClientCredentialsSecret)
```
rene-dekker, line 436: I suggest the name for this annotation also to be "tigera-ee-linseed-elasticsearch-user-secret", just like the name of the secret that it is using.

rene-dekker:
Could you change the PR name such that it is useful as a release note summary?

pkg/render/logstorage/linseed/linseed_test.go
asincu: u

PR #3367, opened 2024-05-29
Limit TLSRoute lists to the tenant namespace in multi-tenant clusters

gcosgrave:
## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/manager/manager_route_config.go
+++ b/pkg/render/manager/manager_route_config.go
@@ -21,14 +21,15 @@ import (
 	"strconv"
 	"strings"
 
-	operatorv1 "github.com/tigera/operator/api/v1"
-	rmeta "github.com/tigera/operator/pkg/render/common/meta"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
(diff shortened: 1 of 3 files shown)


pkg/render/manager/manager_route_config.go
```diff
@@ -21,14 +21,15 @@ import (
...
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	operatorv1 "github.com/tigera/operator/api/v1"
+	rmeta "github.com/tigera/operator/pkg/render/common/meta"
 )
 
 const (
-	voltronRoutesConfigMapName = "voltron-routes"
+	VoltronRoutesConfigMapName = "voltron-routes"
```
Brian-McM, line 32: Nit, I prefer not exporting things just for tests (I'd rather the test have it's own duplicated value). I was thinking that nothing outside of here would need to even know the name of the config map since the annotation / volume generation is done here, and needing to know this value _might_ be indicative of a code smell (it might not though, but I'd like to evaluate the change to export this with the usage needed).

gcosgrave: export undone.

PR #3368, opened 2024-05-29
Clean up unused variables from manager

asincu:
## Description

UI and ES Proxy is not making calls via elastic client any more. Cluster Config was used to set the elastic index and ES secrets were used to store the elastic user.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/manager/manager_controller.go
+++ b/pkg/controller/manager/manager_controller.go
@@ -500,20 +500,6 @@ func (r *ReconcileManager) Reconcile(ctx context.Context, request reconcile.Requ
 		return reconcile.Result{}, err
 	}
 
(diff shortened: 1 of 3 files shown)


tmjd (approved):
Alina let me know this use to be needed to setup an elastic client for es-proxy to make elastic queries directly and they have been deprecated.
LGTM

rene-dekker:
@asincu How about for logging into Kibana?

asincu:
@rene-dekker When logging into Kibana, we use the credentials from OIDC secrets and do not make use of an user setup in es proxy via environment variables. Also, we do not need any suffix/midfix information for any indices.

PR #3370, opened 2024-05-31
Annotate Linseed clients with a hash of the token

asincu:
## Description

Adding an annotations for Linseed token will force a restart and the volume for the token will be reloaded. This is needed when the connection is down, the token expires and in order for the new value to be picked up, we need to to force a restart. This PR only adds annotations for fluentD tokens, as the rest of the application use Linseed client that reloads when it changes.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


asincu:
Will close this PR in order to avoid any churn for the FluentD pos when the token is refreshed. Any updates to the secret should result in a remounting operation. FluentD reads from disk the token for each request.

The amount of time between an update and remount can take longer. It should account for sync period + cache propagation delay (https://kubernetes.io/docs/concepts/configuration/secret/#using-secrets-as-files-from-a-pod)

PR #3381, opened 2024-06-10
Allow Terminating resources to be recreated when Installation is not finalizing. 

caseydavenport:
## Description

Originally spotted in this comment: https://github.com/projectcalico/calico/issues/8368#issuecomment-2033258096

The operator only removes the CNI protector finalizers on these resources when the Installation is `Terminating`. However, in practice there are other scenarios that might result in these resources being deleted. In those cases, we should allow the deletion to complete and immmediately recreate them. Otherwise, they will be stuck terminating forever and be left in a broken state.

Fixes https://github.com/projectcalico/calico/issues/8368

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/node.go
+++ b/pkg/render/node.go
@@ -245,9 +245,9 @@ func (c *nodeComponent) Ready() bool {
 	return true
 }
 
-// CalicoSystemFinalizedObjects returns a list of objects that use the CNIFinalizer that should be
+// CNIPluginFinalizedObjects returns a list of objects that use the CNIFinalizer that should be
 // removed only after the CNI plugin is removed.
-func CalicoSystemFinalizedObjects() []client.Object {
+func CNIPluginFinalizedObjects() []client.Object {
 	return []client.Object{
 		&corev1.ServiceAccount{ObjectMeta: metav1.ObjectMeta{Name: CalicoNodeObjectName, Namespace: common.CalicoNamespace}},
 		&corev1.ServiceAccount{ObjectMeta: metav1.ObjectMeta{Name: CalicoCNIPluginObjectName, Namespace: common.CalicoNamespace}},
--- a/pkg/render/node.go
+++ b/pkg/render/node.go
@@ -245,6 +245,19 @@ func (c *nodeComponent) Ready() bool {
 	return true
 }
 
+// CalicoSystemFinalizedObjects returns a list of objects that use the CNIFinalizer that should be
+// removed only after the CNI plugin is removed.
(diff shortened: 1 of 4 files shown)


pkg/render/node.go
```diff
@@ -245,6 +245,19 @@ func (c *nodeComponent) Ready() bool {
 	return true
 }
 
+// CalicoSystemFinalizedObjects returns a list of objects that use the CNIFinalizer that should be
+// removed only after the CNI plugin is removed.
+func CalicoSystemFinalizedObjects() []client.Object {
```
Brian-McM, line 250: This is probably the quickest way to get this out the door, but a hard coded list here seems a bit fragile. If we add finalizers like this to any other object, it's probably an easy think to forget / miss in a code review.

I'm also wondering if this is the only place that this would apply, i.e. would it be feasible that we'd have finalizers managed by other controllers, or is the situation with this controller (since it controls networking) unique?

Other options might be to filter the list of objects we're creating to see if there's a finalizer so the map can be dynamically built, or even have something baked into the core operator logic when we render objects.

caseydavenport: Yep, these are all good points. I did briefly consider doing something a bit more holistic but thought it might not apply generically - we really only want to do this on a specific subset of all possible Finalizers, and only when the `Installation` is in a specific state (i.e., not being deleted itself), and so the logic felt less generically applicable.

I'll think on ways to make this a bit cleaner, but so far this felt the least icky to me (at the expensive of needing to keep two places in-sync, but it's hard to imagine what other places we might use this particular finalizer).

Brian-McM: Understandable. I was thinking that one cheap solution would be to iterate over all the objects that we render and check if any have this finalizer that you're looking for. Then, if the installation is not in a deleted state run your finalizer removal logic on those objects. 

There might be an even slicker solution than that (if that works at all) though.

I would be fine with this solution too if something like that doesn't work, the "hard codedness" of this solution certainly isn't my favourite, but it's also not bad (especially considering what you had said).

Brian-McM: In light of the conversation we had outside of this PR (about generalising this to anything other than the CNI plugins resources), this should probably be named something specific to the CNI plugin's resources.

Brian-McM (approved):

pkg/render/node.go
caseydavenport: I was just in the process of doing that :laughing:

PR #3389, opened 2024-06-19
[CORE-10407] Add support for OpenShift HCP

coutinhop:
## Description

Add support for OpenShift HCP (Hosted Control Planes). Main change is on the AWS security group setup job, as the SG names are different in an HCP hosted cluster as opposed to vanilla OpenShift (or a hosting cluster which is essentially the same). Also added auto-detection for an HCP hosted cluster (when ControlPlaneTopology == "External") and propagation of this info to the SG setup job via an env var.

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


go.mod
```diff
@@ -32,11 +32,14 @@ require (
 	k8s.io/apiserver v0.28.9
 	k8s.io/client-go v0.28.9
 	k8s.io/kube-aggregator v0.28.9
+	k8s.io/utils v0.0.0-20230406110748-d93618cff8a2 // indirect
 	sigs.k8s.io/controller-runtime v0.15.3
 	sigs.k8s.io/kind v0.17.0 // Do not remove, not used by code but used by build
 	sigs.k8s.io/yaml v1.3.0
 )
 
+require github.com/openshift/client-go v0.0.0-20200521150516-05eb9880269c
```
caseydavenport, line 41: can we put this in the main require?

Actually, looks like we have two require sections here - not sure why!

pkg/awssgsetup/aws_security_group_setup.go
```diff
@@ -38,7 +38,7 @@ var DEBUG = 5
 // setupAWSSecurityGroups updates the master and worker security groups used in an
 // OpenShift AWS setup. It sets a time that should be checked before attempting
 // to do another call to setup.
-func SetupAWSSecurityGroups(ctx context.Context, client client.Client) error {
+func SetupAWSSecurityGroups(ctx context.Context, client client.Client, hostedOpenShift bool) error {
```
caseydavenport, line 41: nit, but we can just called this `hosted` I think, given this function is OpenShift only already.

caseydavenport: Also worth a function-level comment explaining what "hosted" means.

pkg/controller/installation/core_controller.go
```diff
@@ -1214,9 +1214,16 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 	// If we're on OpenShift on AWS render a Job (and needed resources) to
 	// setup the security groups we need for IPIP, BGP, and Typha communication.
 	if openShiftOnAws {
+		// Detect if this cluster is an OpenShift HPC hosted cluster, as AWS
+		// security group setup is different in this case.
+		isHostedOpenShift, err := utils.IsHostedOpenShift(ctx, r.config)
+		if err != nil {
+			log.Info(err.Error())
```
caseydavenport, line 1221: I think we want to return an error in this branch -  What happens if this fails and we attempt to continue anyway?

pkg/controller/installation/core_controller.go
```diff
@@ -1214,9 +1214,16 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 	// If we're on OpenShift on AWS render a Job (and needed resources) to
 	// setup the security groups we need for IPIP, BGP, and Typha communication.
 	if openShiftOnAws {
+		// Detect if this cluster is an OpenShift HPC hosted cluster, as AWS
+		// security group setup is different in this case.
+		isHostedOpenShift, err := utils.IsHostedOpenShift(ctx, r.config)
```
caseydavenport, line 1219: This is a little bit of a bad pattern - this is going to cause an API call on every reconcile, since it's instantiating a direct non-cached client.

We should instead register a watch for this object, and use the runtime client to read it from the cache. This will avoid the extra API calls.

pkg/render/aws-securitygroup-setup_test.go
```diff
@@ -90,5 +91,52 @@ var _ = Describe("AWS SecurityGroup Setup rendering tests", func() {
...
+				Value: "true",
+			},
+			{
+				Name:  "KUBELET_KUBECONFIG",
+				Value: "/etc/kubernetes/kubeconfig",
+			},
+		}
+		Expect(job.Spec.Template.Spec.Containers[0].Env).To(ConsistOf(expectedEnv))
+	})
+	It("should render Setup Job env correctly when in an HCP hosted cluster", func() {
```
caseydavenport, line 110: nit: should have a newline before the It()

caseydavenport (changes requested):

caseydavenport (dismissed):

pkg/controller/installation/core_controller.go
```diff
@@ -1843,11 +1851,20 @@ func isOpenshiftOnAws(install *operator.Installation, ctx context.Context, clien
...
-		return false, fmt.Errorf("unable to read OpenShift infrastructure configuration: %s", err.Error())
+		return false, fmt.Errorf("unable to read OpenShift infrastructure configuration: %w", err)
 	}
 	return (infra.Status.PlatformStatus.Type == "AWS"), nil
 }
 
+// isHostedOpenShift returns true if this cluster is an OpenShift HCP hosted cluster.
+func isHostedOpenShift(ctx context.Context, client client.Client) (bool, error) {
+	infra := configv1.Infrastructure{}
+	if err := client.Get(ctx, types.NamespacedName{Name: openshiftNetworkConfig}, &infra); err != nil {
```
caseydavenport, line 1862: If you don't start a Watch() for this resource, then there will be nothing in the client's cache and we'll need to go round-trip to the API anyway.

You should add a Watch in this block: https://github.com/tigera/operator/blob/01ec8648080b1da2388df5997564e0a16c9b5646/pkg/controller/installation/core_controller.go#L216-L225

caseydavenport (approved):

PR #3412, opened 2024-07-08
Changes to enable nftables dataplane

caseydavenport:
## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/api/v1/installation_types.go
+++ b/api/v1/installation_types.go
@@ -421,8 +421,8 @@ const (
 
 // LinuxDataplaneOption controls which dataplane is to be used on Linux nodes.
 //
-// One of: Iptables, BPF, VPP, Nftables, BPFWithNftables
-// +kubebuilder:validation:Enum=Iptables;BPF;VPP;Nftables;BPFWithNftables
+// One of: Iptables, BPF, VPP, Nftables
+// +kubebuilder:validation:Enum=Iptables;BPF;VPP;Nftables;
 type LinuxDataplaneOption string
 
 const (
@@ -430,7 +430,6 @@ const (
 	LinuxDataplaneBPF      LinuxDataplaneOption = "BPF"
 	LinuxDataplaneVPP      LinuxDataplaneOption = "VPP"
 	LinuxDataplaneNftables LinuxDataplaneOption = "Nftables"
-	LinuxDataplaneBPFNFT   LinuxDataplaneOption = "BPFWithNftables"
 )
 
 // +kubebuilder:validation:Enum=HNS;Disabled
--- a/api/v1/installation_types.go
+++ b/api/v1/installation_types.go
@@ -421,13 +421,16 @@ const (
 
 // LinuxDataplaneOption controls which dataplane is to be used on Linux nodes.
 //
-// One of: Iptables, BPF
+// One of: Iptables, BPF, VPP, Nftables, BPFWithNftables
+// +kubebuilder:validation:Enum=Iptables;BPF;VPP;Nftables;BPFWithNftables
 type LinuxDataplaneOption string
 
 const (
 	LinuxDataplaneIptables LinuxDataplaneOption = "Iptables"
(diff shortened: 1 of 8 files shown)


api/v1/installation_types.go
```diff
@@ -421,13 +421,16 @@ const (
...
+// One of: Iptables, BPF, VPP, Nftables, BPFWithNftables
+// +kubebuilder:validation:Enum=Iptables;BPF;VPP;Nftables;BPFWithNftables
 type LinuxDataplaneOption string
 
 const (
 	LinuxDataplaneIptables LinuxDataplaneOption = "Iptables"
 	LinuxDataplaneBPF      LinuxDataplaneOption = "BPF"
 	LinuxDataplaneVPP      LinuxDataplaneOption = "VPP"
+	LinuxDataplaneNftables LinuxDataplaneOption = "Nftables"
+	LinuxDataplaneBPFNFT   LinuxDataplaneOption = "BPFWithNftables"
```
caseydavenport, line 433: Submitting for discussion. Not sure the best way to represent this - we can have nftables mode, but also BPF mode with nftables instead of iptables

tmjd: I guess an alternative would be to add a new field that is only applicable in BPF dataplane mode that is either Nftables or Iptables (default).

caseydavenport: Yeah, that's basically the only alternative I thought of. But also feels like an awkward UX.

e.g.,

```go
linuxDataplane: BPF
bpfTablesMode: Nftables
```

I think perhaps for now I stick with just `Iptables`, `BPF`, `VPP`, `Nftables` in order to get this merged, and consider BPF + Nftables separately (it's not even 100% working yet anyway).

pkg/controller/installation/core_controller.go
```diff
@@ -1626,6 +1627,20 @@ func getOrCreateTyphaNodeTLSConfig(cli client.Client, certificateManager certifi
 func (r *ReconcileInstallation) setDefaultsOnFelixConfiguration(ctx context.Context, install *operator.Installation, fc *crdv1.FelixConfiguration, reqLogger logr.Logger) (bool, error) {
 	updated := false
 
+	if install.Spec.CalicoNetwork.LinuxDataplane != nil {
+		if *install.Spec.CalicoNetwork.LinuxDataplane == operatorv1.LinuxDataplaneNftables && fc.Spec.NFTablesMode == nil {
```
tmjd, line 1631: If I'm reading this right we will only set this to enabled if FelixConfig has NFTablesMode nil but we'll set disabled even if FelixConfig has NFTablesMode set.
That doesn't seem like symmetric behavior to me.

caseydavenport: Yeah you're absolutely right.

pkg/controller/installation/core_controller.go
```diff
@@ -1626,6 +1627,20 @@ func getOrCreateTyphaNodeTLSConfig(cli client.Client, certificateManager certifi
 func (r *ReconcileInstallation) setDefaultsOnFelixConfiguration(ctx context.Context, install *operator.Installation, fc *crdv1.FelixConfiguration, reqLogger logr.Logger) (bool, error) {
 	updated := false
 
+	if install.Spec.CalicoNetwork.LinuxDataplane != nil {
```
tmjd, line 1630: Hmmm, this is in a function called `setDefaults...`, so since this always sets the config it seems like we should change the function name or create a new function. I think it'd be good to include a comment about why we're ok setting this compared to how we handle other FelixConfig options where we'll only set defaults.

tmjd (approved):
LGTM, I wonder if there should be some custom migration code for taking a Manifest based install and migrating it to operator managed install.

caseydavenport:
Yeah, migration is a good callout. I suspect we do want that, but I'll raise a backlog item for it.

PR #3424, opened 2024-07-12
Relax ImageSet validation

tmjd:
Only block progress when an ImageSet exists for the same variant but not one with the correct version.

These changes help ensure when a cluster using Calico is upgrade to Enterprise that an ImageSet that had been created for Calico will not block the upgrade to Enterprise.
This helps decouple the management of a Calico install that was managed by a platform where the platform added an ImageSet but the cluster had been upgraded to Enterprise, it ensures the addition of the ImageSet for Calico won't impact the already functioning Enterprise cluster that is being managed separately.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/installation/core_controller.go
+++ b/pkg/controller/installation/core_controller.go
@@ -1394,6 +1394,21 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 		return reconcile.Result{}, err
 	}
 
+	if imageSet == nil {
+		// There is no imageSet for the configured variant, but check to see if there are any
+		// ImageSets with a different variant so we can give the user some kind of indication
+		// to why an existing ImageSet is being ignored.
+		nvis, err := imageset.DoesNonVariantImageSetExist(ctx, r.client, instance.Spec.Variant)
+		if err != nil {
+			r.status.SetDegraded(operator.ResourceReadError, "Error checking for non-variant ImageSet", err, reqLogger)
+			return reconcile.Result{}, err
+		} else {
+			if nvis {
+				reqLogger.Info("An ImageSet exists for a different variant")
+			}
+		}
+	}
+
 	if err = imageset.ValidateImageSet(imageSet); err != nil {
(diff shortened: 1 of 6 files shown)


pkg/controller/installation/core_controller.go
```diff
@@ -1394,6 +1394,18 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 		return reconcile.Result{}, err
 	}
 
+	if imageSet == nil {
```
rene-dekker, line 1397: I had to read this a few times. As a reader, I think I would prefer to have something like:

```
	required, err := imageset.IsRequired(ctx, r.client, imageSet, instance.Spec.Variant);
	if err != nil {
		r.status.SetDegraded(operator.ResourceReadError, "Error checking for non-variant ImageSet", err, reqLogger)
	} else if required {
		r.status.SetDegraded(...)
		return reconcile.Result{}, ...
	}

```

And inside of the imageset.IsRequired(..) explain what the requirements are. I think that would be the best place for those details.

pkg/controller/utils/imageset/imageset.go
```diff
@@ -72,14 +85,49 @@ func GetImageSet(ctx context.Context, cli client.Client, v operator.ProductVaria
...
 
-	return nil, fmt.Errorf("ImageSets exist but none with the expected name %s", setName)
+	if variantISExists {
+		return nil, fmt.Errorf("ImageSets exist but none with the expected name %s", setName)
+	} else {
+		return nil, nil
+	}
+}
+
+func DoesNonVariantImageSetExist(ctx context.Context, cli client.Client, v operator.ProductVariant) (bool, error) {
```
rene-dekker, line 107: Maybe it can have a comment explaining the requirements. Something like:
- ImageSets are optional by default.
- However, once an ImageSet is specified for your ProductVariant, it is required to have an ImageSet for your ProductVariant and version as well.
- ImageSets for other ProductVariants are not considered for this requirement.

rene-dekker (approved):
lgtm

PR #3435, opened 2024-07-17
Set default values for vxlanVNI and BPFHostConntrackBypass for DockerEE

sridhartigera:
## Description

1. DockerEE uses vxlanPort 4789 and this results in a conflict. Hence changing the default vxlanVNI for DockerEE to 10000.
2. When in BPF mode in DockerEE, ```BPFHostConntrackBypass``` needs to be disabled for cluster connectivity to work. 

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/installation/core_controller.go
+++ b/pkg/controller/installation/core_controller.go
@@ -1686,13 +1686,17 @@ func (r *ReconcileInstallation) setDefaultsOnFelixConfiguration(ctx context.Cont
 		updated = true
 	}
 
-	vxlanPort := 4798
+	// MKE uses a vxlanVNI:4096 and vxlanPort:4789 for its docker swarm vxlan.
+	// This results in a conflict with calico's VXLAN and the vxlan.calico interface
+	// gets deleted. To fix this we change the vxlanVNI to 10000 as recommended by
+	// MKE docs (https://docs.mirantis.com/mke/3.7/cli-ref/mke-cli-install.html).
 	if install.Spec.KubernetesProvider == operator.ProviderDockerEE {
-		fc.Spec.VXLANPort = &vxlanPort
 		// Set bpfHostConntrackBypass to false for eBPF dataplane to work with MKE
 		if install.Spec.BPFEnabled() {
 			disableBPFHostConntrackBypass(fc)
 		}
+		vxlanVNI = 10000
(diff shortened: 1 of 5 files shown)


tmjd:
Is there a specific reason 4798 was chosen? It looks very similar to the value we're trying to avoid and might look like a typo. WDYT about using 4790, I think that is a little better.

sridhartigera:
> Is there a specific reason 4798 was chosen? It looks very similar to the value we're trying to avoid and might look like a typo. WDYT about using 4790, I think that is a little better.



> Is there a specific reason 4798 was chosen? It looks very similar to the value we're trying to avoid and might look like a typo. WDYT about using 4790, I think that is a little better.

Might end up changing the VNI. Will update the PR accordingly. MKE uses ```vni:4096, port:4789``` for their swarm overlay. We thought of using a different port but shaun thinks its better to use a different VNI. Will update the PR accordingly.

pkg/controller/installation/core_controller.go
```diff
@@ -1686,6 +1686,20 @@ func (r *ReconcileInstallation) setDefaultsOnFelixConfiguration(ctx context.Cont
...
+	// This results in a conflict with calico's VXLAN and the vxlan.calico interface
+	// gets deleted. To fix this we change the vxlanVNI to 10000 as recommended by
+	// MKE docs (https://docs.mirantis.com/mke/3.7/cli-ref/mke-cli-install.html).
+	if install.Spec.KubernetesProvider == operator.ProviderDockerEE {
+		// Set bpfHostConntrackBypass to false for eBPF dataplane to work with MKE
+		if install.Spec.BPFEnabled() {
+			disableBPFHostConntrackBypass(fc)
+		}
+		vxlanVNI = 10000
+		fc.Spec.VXLANVNI = &vxlanVNI
```
tmjd, line 1699: The code here should only be setting values if they are not already set.
That will also hopefully ensure that this change won't impact functionality in existing clusters.

Does `fc.Spec.VXLANVNI` get defaulted by CalicoNode/Felix? If not that could be an issue with impacting existing clusters, unless it is safe to change VXLANVNI on the fly.

tmjd (approved):
LGTM

fasaxc (approved):

PR #3439, opened 2024-07-19
the calico-apiserver namespace frequently get stuck for a long time in terminating status when upgrading to enterprise

gcosgrave:
…
## Description

The apiserver TigeraStatus error message `Error creating / updating resource: secrets "tigera-apiserver-certs" is forbidden: unable to create new content in namespace calico-apiserver because it is being terminated`  identifies the problem as the certs secret is called "calico-apiserver-certs" when it is in the "calico-apiserver" namespace, and "tigera-apiserver-certs" when in the "tigera-apiserver" namespace

```
kind: TigeraStatus
metadata:
    creationTimestamp: "2024-07-18T11:15:59Z"
    generation: 1
    name: apiserver
    resourceVersion: "2010"
    uid: fcdb9846-3772-4684-afe3-db458486608b
spec: {}
status:
    conditions:
        - lastTransitionTime: "2024-07-18T11:18:06Z"
          message: 'Error creating / updating resource: secrets "tigera-apiserver-certs" is forbidden: unable to create new content in namespace calico-apiserver because it is being terminated'
          observedGeneration: 1
          reason: ResourceUpdateError
          status: "True"
          type: Degraded
        - lastTransitionTime: "2024-07-18T11:17:56Z"
          observedGeneration: 1
          reason: Unknown
          status: "False"
          type: Available
        - lastTransitionTime: "2024-07-18T11:17:56Z"
          message: Deployment "tigera-system/tigera-apiserver" is not available (awaiting 2 replicas)
          observedGeneration: 1
          reason: ResourceNotReady
          status: "True"
          type: Progressing
```

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/apiserver/apiserver_controller.go
+++ b/pkg/controller/apiserver/apiserver_controller.go
@@ -243,7 +243,7 @@ func (r *ReconcileAPIServer) Reconcile(ctx context.Context, request reconcile.Re
 	}
 
 	// Query for the installation object.
-	variant, installationSpec, err := utils.GetInstallation(context.Background(), r.client)
+	_, installationSpec, err := utils.GetInstallation(context.Background(), r.client)
 	if err != nil {
 		if errors.IsNotFound(err) {
 			r.status.SetDegraded(operatorv1.ResourceNotFound, "Installation not found", err, reqLogger)
@@ -252,11 +252,7 @@ func (r *ReconcileAPIServer) Reconcile(ctx context.Context, request reconcile.Re
 		r.status.SetDegraded(operatorv1.ResourceReadError, "Error querying installation", err, reqLogger)
 		return reconcile.Result{}, err
 	}
-	if variant == "" {
-		r.status.SetDegraded(operatorv1.ResourceNotReady, "Waiting for Installation to be ready", nil, reqLogger)
-		return reconcile.Result{}, nil
-	}
-	ns := rmeta.APIServerNamespace(variant)
+	ns := rmeta.APIServerNamespace(installationSpec.Variant)
 
 	certificateManager, err := certificatemanager.Create(r.client, installationSpec, r.clusterDomain, common.OperatorNamespace())
 	if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/controller/apiserver/apiserver_controller.go
```diff
@@ -252,11 +252,7 @@ func (r *ReconcileAPIServer) Reconcile(ctx context.Context, request reconcile.Re
 		r.status.SetDegraded(operatorv1.ResourceReadError, "Error querying installation", err, reqLogger)
 		return reconcile.Result{}, err
 	}
-	if variant == "" {
```
caseydavenport, line 255: @tmjd @gcosgrave I think we still do need this check to handle the tiniest of race conditions.

If a user creates an Installation with no variant, we want to wait for the Installation controller to default it before we continue. It should only add a second of delay.

tmjd: Yeah I agree. To be specific the check can be changed to `if installationSpec.Variant == ""`

gcosgrave: done

pkg/controller/apiserver/apiserver_controller.go
```diff
@@ -252,11 +252,11 @@ func (r *ReconcileAPIServer) Reconcile(ctx context.Context, request reconcile.Re
...
+	if installationSpec.Variant == "" {
 		r.status.SetDegraded(operatorv1.ResourceNotReady, "Waiting for Installation to be ready", nil, reqLogger)
```
tmjd, line 256: I think the message here needs to be changed to something like "Waiting for Installation Variant to be set". I think the ResourceNotReady is good though.

tmjd (approved):
Looks good, I think the unset variant message could use an update.

pkg/controller/apiserver/apiserver_controller.go
gcosgrave: Agreed, message updated

PR #3452, opened 2024-08-02
Improve OpenShift platform detection accuracy

libesz:
## Description

This change improves the accuracy of Openshift platform detection. In some cases (i.e. when one hosts managed Openshift master deployments on a non-Openshift cluster) some portion of the Openshift API resources may be present on a non-Openshift cluster. That tricks Tigera operator atm to think it is operating on an Openshift cluster, leading to various failures.

After the change, the condition becomes more restricted and will be triggered only if the `ClusterVersion` CRD is specifically present, which is the main source of the actual Openshift cluster version, for example:

```shell
$ k get clusterversions.config.openshift.io                                                                                                                         [22:03:33]
NAME      VERSION   AVAILABLE   PROGRESSING   SINCE   STATUS
version   4.15.11   True        False         23d     Cluster version is 4.15.11
```  

This custom resource is extremely unlikely (and makes no sense) to be present on non-Openshift clusters.

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/utils/discovery.go
+++ b/pkg/controller/utils/discovery.go
@@ -20,6 +20,7 @@ import (
 	"strings"
 
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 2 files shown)


caseydavenport (approved):
This LGTM, thanks!

mihivagyok (approved):

PR #3453, opened 2024-08-05
Print image options and ImageSet utility

tmjd:
## Description

Added additional `--print-images` values to print only Calico or Enterprise images.
Also add makefile targets for generating ImageSets from an operator image.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -140,7 +140,11 @@ DOCKER_RUN := $(CONTAINERIZED) $(CALICO_BUILD)
 BUILD_IMAGE?=tigera/operator
 BUILD_INIT_IMAGE?=tigera/operator-init
 
-BINDIR?=build/_output/bin
+BUILD_DIR?=build/_output
+BINDIR?=$(BUILD_DIR)/bin
+
+$(BUILD_DIR):
+	mkdir -p $(BUILD_DIR)
 
 IMAGE_REGISTRY?=quay.io
 PUSH_IMAGE_PREFIXES?=quay.io/
@@ -263,7 +267,7 @@ $(BINDIR)/kind:
 	$(CONTAINERIZED) $(CALICO_BUILD) sh -c "GOBIN=/go/src/$(PACKAGE_NAME)/$(BINDIR) go install sigs.k8s.io/kind"
 
 clean:
-	rm -rf build/_output
+	rm -rf $(BUILD_DIR)
 	rm -rf build/init/bin
 	rm -rf hack/bin
 	rm -rf .go-pkg-cache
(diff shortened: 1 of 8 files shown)


main.go
```diff
@@ -139,18 +139,24 @@ func main() {
...
 			cmpnts = append(cmpnts, components.EnterpriseImages...)
-
-			for _, x := range cmpnts {
-				ref, _ := components.GetReference(x, "", "", "", nil)
-				fmt.Println(ref)
-			}
-			os.Exit(0)
+		} else if strings.ToLower(printImages) == "listcalico" {
+			cmpnts = components.CalicoImages
+		} else if strings.ToLower(printImages) == "listenterprise" {
```
rene-dekker, line 148: Nice addition.

Makefile
```diff
@@ -940,3 +941,99 @@ define github_pr_add_comment
...
+clean-imageset:
+	rm -f $(BUILD_DIR)/*imageset*
+
+
+ifdef VERSION
+OPERATOR_IMAGE ?= $(IMAGE_REGISTRY)/$(BUILD_IMAGE):$(VERSION)
+else
+OPERATOR_IMAGE ?= $(BUILD_IMAGE):latest
+endif
+CRANE_IMAGE ?= calico/go-build:master
```
rene-dekker, line 958: I think we should be using GO_BUILD_VER here. If that crane version is not right, then we should update it there too.

nit: could you add an empty line above to separate the OPERATOR_IMAGE and CRANE_IMAGE variable creations?

rene-dekker: We already have a way of defining the crane command in other repos:
https://github.com/projectcalico/calico/blob/53af4139e9df29001d36712ed900fb963261b064/lib.Makefile#L423-L441

It has the advantage of being tested and used already and not leaving extra files on your disk.

Makefile
```diff
@@ -940,3 +941,99 @@ define github_pr_add_comment
...
+# to generate an ImageSet for an alternate registry.
+.PHONY: gen-imageset gen-enterprise-imageset gen-calico-imageset
+gen-imageset: gen-enterprise-imageset gen-calico-imageset
+	@cat $(BUILD_DIR)/imageset-enterprise.yaml > $(BUILD_DIR)/imageset.yaml
+	@echo "---" >> $(BUILD_DIR)/imageset.yaml
+	@cat $(BUILD_DIR)/imageset-calico.yaml >> $(BUILD_DIR)/imageset.yaml
+	@echo Imageset written to file $(BUILD_DIR)/imageset.yaml
+
+.PHONY: gen-enterprise-imageset
+gen-enterprise-imageset: $(BINDIR)/crane
```
rene-dekker, line 991: For me the command crashes on the operator image, any chance it could be skipped?

```
2024/08/06 11:25:15 HEAD request failed, falling back on GET: HEAD https://quay.io/v2/tigera/operator/manifests/v1.35.0-0.dev-85-g1460069d5243: unexpected status code 404 Not Found (HEAD responses have no body, use GET for details)
Error: GET https://quay.io/v2/tigera/operator/manifests/v1.35.0-0.dev-85-g1460069d5243: MANIFEST_UNKNOWN: manifest unknown; map[]
```

Makefile
tmjd: What do you think should be copied from that because I don't think we should start including that lib.Makefile because that'll be introducing a dependency we don't need.
Do we need the "LOCAL" functionality? Do we need the "DRY_RUN"? I would probably suggest neither are needed since this is adding a convenience utility.

Makefile
tmjd: What operator image did you use?
No it cannot be skipped because we need the digest from the registry to create the ImageSet.
I'm guessing maybe it is trying to use an image that hasn't been built and pushed to the registry. I could add a comment/warning that running the command with an image built from master may not work if the appropriate image has not been built by CI and pushed to the registry.

Makefile
rene-dekker: We do already have this CONFIRM/DRY_RUN logic in our makefile, although I agree that we don't need it. Importing the other makefile could allow us to clean quite some thing up, and on the flip-side introduce things we don't need. Probably worth discussing separately at some point?

I think the thing we would need is `CRANE_CMD` and `double_quote`.

Makefile
rene-dekker: I think a comment is enough.

rene-dekker (approved):
lgtm

PR #3454, opened 2024-08-09
Tenant fail to clean up because of missing connection to Elasticsearch

asincu:
## Description

Tenants are not being deleted because of the user controller tries to first make sure Elastic is ready. In an multi-tenant environment, we never have elastic search deployed in the same cluster.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/logstorage/users/users_controller.go
+++ b/pkg/controller/logstorage/users/users_controller.go
@@ -357,13 +357,16 @@ func (r *UsersCleanupController) Reconcile(ctx context.Context, request reconcil
 		request.Namespace, "Request.Name", request.Name, "installNS", helper.InstallNamespace(), "truthNS", helper.TruthNamespace())
 	reqLogger.Info("Reconciling LogStorage - Cleanup")
 
-	// Wait for Elasticsearch to be installed and available.
-	elasticsearch, err := utils.GetElasticsearch(ctx, r.client)
-	if err != nil {
-		return reconcile.Result{}, err
-	}
-	if elasticsearch == nil || elasticsearch.Status.Phase != esv1.ElasticsearchReadyPhase {
-		return reconcile.Result{}, nil
+	if !r.elasticExternal {
+		// Wait for Elasticsearch to be installed and available.
+		elasticsearch, err := utils.GetElasticsearch(ctx, r.client)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+		if elasticsearch == nil || elasticsearch.Status.Phase != esv1.ElasticsearchReadyPhase {
+			reqLogger.Info("Elasticsearch is not ready")
+			return reconcile.Result{}, nil
+		}
 	}
 
 	// Clean up any stale users that may have been left behind by a previous tenant


pkg/controller/logstorage/users/users_controller.go
```diff
@@ -357,15 +358,51 @@ func (r *UsersCleanupController) Reconcile(ctx context.Context, request reconcil
...
+		return reconcile.Result{}, nil
+	} else if err != nil {
 		return reconcile.Result{}, err
 	}
-	if elasticsearch == nil || elasticsearch.Status.Phase != esv1.ElasticsearchReadyPhase {
+
+	// Get LogStorage resource.
+	logStorage := &operatorv1.LogStorage{}
+	err = r.client.Get(ctx, utils.DefaultTSEEInstanceKey, logStorage)
+	if err != nil {
```
Brian-McM, line 378: Just to double check, this change isn't for the current bug in cc, this is something else you noticed, right?

I think in general we try to do this CR check first since there's not really any point in doing anything else if the CR doesn't exist.

Brian-McM (approved):
One small comment but looks good, thanks for fixing that.

pkg/controller/logstorage/users/users_controller.go
```diff
@@ -66,6 +66,7 @@ type UsersCleanupController struct {
 	scheme          *runtime.Scheme
 	esClientFn      utils.ElasticsearchClientCreator
 	elasticExternal bool
+	multiTenant     bool
```
caseydavenport, line 69: I don't understand why we need this - line 76 disables this controller unless `opts.MultiTenant` is true, so this is _always_ true, no?

pkg/controller/logstorage/users/users_controller.go
```diff
@@ -357,15 +358,51 @@ func (r *UsersCleanupController) Reconcile(ctx context.Context, request reconcil
...
+		if !errors.IsNotFound(err) {
+			return reconcile.Result{}, err
+		}
+		reqLogger.Info("Logstorage is not found")
+		return reconcile.Result{}, nil
+	}
+
+	// Wait for the initializing controller to indicate that the LogStorage object is actionable.
+	if logStorage.Status.State != operatorv1.TigeraStatusReady {
+		reqLogger.Info("Logstorage is not marked as ready")
```
caseydavenport, line 390: We normally SetDegraded when dependencies aren't ready - we should probably do so here as well.

pkg/controller/logstorage/users/users_controller.go
```diff
@@ -357,15 +358,51 @@ func (r *UsersCleanupController) Reconcile(ctx context.Context, request reconcil
...
 		return reconcile.Result{}, err
 	}
-	if elasticsearch == nil || elasticsearch.Status.Phase != esv1.ElasticsearchReadyPhase {
+
+	// Get LogStorage resource.
+	logStorage := &operatorv1.LogStorage{}
+	err = r.client.Get(ctx, utils.DefaultTSEEInstanceKey, logStorage)
+	if err != nil {
+		// Not finding the LogStorage CR is not an error, as a Managed cluster will not have this CR available but
+		// there are still "LogStorage" related items that need to be set up
```
caseydavenport, line 380: Thinks we need two SetDegraded calls in this branch - one for the error case and one for the NotFound case.

pkg/controller/logstorage/users/users_controller.go
```diff
@@ -357,15 +358,51 @@ func (r *UsersCleanupController) Reconcile(ctx context.Context, request reconcil
...
-	// Wait for Elasticsearch to be installed and available.
-	elasticsearch, err := utils.GetElasticsearch(ctx, r.client)
-	if err != nil {
+	// We skip requests without a namespace specified in multi-tenant setups.
+	if r.multiTenant && request.Namespace == "" {
+		return reconcile.Result{}, nil
+	}
+
+	// Check if this is a tenant-scoped request.
+	_, _, err := utils.GetTenant(ctx, r.multiTenant, r.client, request.Namespace)
```
caseydavenport, line 367: Is this right? Shouldn't we still reconcile if there is no tenant since we need to clean up users for that tenant?

pkg/controller/logstorage/users/users_controller.go
```diff
@@ -357,15 +358,51 @@ func (r *UsersCleanupController) Reconcile(ctx context.Context, request reconcil
...
+		return reconcile.Result{}, nil
+	}
+
+	// Wait for the initializing controller to indicate that the LogStorage object is actionable.
+	if logStorage.Status.State != operatorv1.TigeraStatusReady {
+		reqLogger.Info("Logstorage is not marked as ready")
 		return reconcile.Result{}, nil
 	}
 
+	if !r.elasticExternal {
```
caseydavenport, line 394: From the description of the PR, it sounds like this is the only change necessary to fix the bug?

pkg/controller/logstorage/users/users_controller.go
Brian-McM: That's probably a good point, I don't know that it means that this should change here though. We probably want to ensure the tenant resource isn't actually deleted until it's cleaned up after. Otherwise we can't know if the request for a tenant or because of change in a common namespace.

pkg/controller/logstorage/users/users_controller.go
Brian-McM: Do we normally degrade if the CR isn't found? If this controller is tied to the CR and the CR doesn't exist then we should report any sort of status.

caseydavenport: Oh yeah, you're right. We would typically call OnCRNotFound instead I think.

pkg/controller/logstorage/users/users_controller.go
caseydavenport: I think this branch is effectively dead code, thinking about it some more.

We don't watch anything in a common namespace for this controller: https://github.com/tigera/operator/blob/191bc39643441e935e09ff9b544da607f9627ec6/pkg/controller/logstorage/users/users_controller.go#L145-L158

I think we should just remove this check - seems like the presence of this code is misleading, and negates the "periodic reconcile" behavior that we want.

caseydavenport: I suppose we probably want a watch for LogStorage to be added, if we are going to add that check here ( seems like we probably should, but also it's probably not a huge issue).

If we do, I think it is reasonable to reconcile users when the log storage changes.

pkg/controller/logstorage/users/users_controller.go
```diff
@@ -357,15 +358,51 @@ func (r *UsersCleanupController) Reconcile(ctx context.Context, request reconcil
...
+		// Wait for Elasticsearch to be installed and available.
+		elasticsearch, err := utils.GetElasticsearch(ctx, r.client)
+		if err != nil {
```
caseydavenport, line 397: both of these branches also need SetDegraded calls.

pkg/controller/logstorage/users/users_controller.go
Brian-McM: Personally I'm a little hesitant about not handling the case where we watching something from a different namespace, even if it's not currently done.

I think the assumptions for these controllers have been you can watch what you need to from any namespace, so there's really no reason why this couldn't change, then we do have a bug.

Unless we have some sort of standard that this is a specific type of controller that can only ever watch specific namespaces (and also not global values), and we also enforce this (possible by wrapping the controller / watch registration to fail if the correct namespace isn't added to the watch), then this seems like a footgun.

caseydavenport: > Personally I'm a little hesitant about not handling the case where we watching something from a different namespace, even if it's not currently done.

The design of this controller is that it is a full reconcile on every event, regardless of what triggered it. I believe that to mean that the controller already properly handles events triggered from other namespaces. I am against adding code that is effectively unused, on the premise that it might someday be needed - if we add such a watch in the future, then we can add the necessary logic to the body of Reconcile() at the same time (which may or may not involve a tenant check depending on what the actual change is). 

Specifically adding this check breaks the periodic reconcile behavior of this controller, which is important, so we need to at least address that before we can merge this change.

caseydavenport: Either way, I think it should be a second PR - seems unrelated to the fix at hand here.

pkg/controller/logstorage/users/users_controller.go
asincu: Yup. I added the rest of the code to have parity with the user creation controller, but I am not sure if that is really needed since we read all tenant anyway for deletion. I am happy to only have this put in at the moment.

pkg/controller/logstorage/users/users_controller.go
asincu: Yup, we are only calling this for multi tenancy.

pkg/controller/logstorage/users/users_controller.go
asincu: I wanted to avoid a new status since I do not know how we can track it since it refers to deleted tenants.

pkg/controller/logstorage/users/users_controller.go
asincu: I wanted to avoid a new status since I do not know how we can track it since it refers to deleted tenants.

caseydavenport (approved):

Brian-McM (approved):

PR #3455, opened 2024-08-09
Remove FIPS mode for enterprise users. It is no longer supported.

rene-dekker:
Support was removed in v3.18 (release-v1.30+), but the code was never removed.

If you enable FIPS on EE, you will now see the following:

```
$ k get tigerastatus  calico -o yaml
kind: TigeraStatus
metadata:
  creationTimestamp: "2024-08-19T18:02:49Z"
  generation: 1
  name: calico
  resourceVersion: "10643"
  uid: f52ae118-9282-47ca-b362-64865ce7a8fe
spec: {}
status:
  conditions:
  - lastTransitionTime: "2024-08-19T18:22:10Z"
    message: 'Invalid Installation provided: Installation spec.FIPSMode=Enabled combined
      with spec.Variant=TigeraSecureEnterprise is not supported'
    observedGeneration: 4
    reason: InvalidConfigurationError
    status: "True"
    type: Degraded
  - lastTransitionTime: "2024-08-19T18:22:10Z"
    observedGeneration: 4
    reason: Unknown
    status: "False"
    type: Available
  - lastTransitionTime: "2024-08-19T18:03:59Z"
    observedGeneration: 4
    reason: Unknown
    status: "False"
    type: Progressing
```

### DIFF
--- a/pkg/controller/applicationlayer/applicationlayer_controller_test.go
+++ b/pkg/controller/applicationlayer/applicationlayer_controller_test.go
@@ -22,16 +22,6 @@ import (
 	. "github.com/onsi/gomega"
 	"github.com/stretchr/testify/mock"
 
-	operatorv1 "github.com/tigera/operator/api/v1"
-	"github.com/tigera/operator/pkg/apis"
-	crdv1 "github.com/tigera/operator/pkg/apis/crd.projectcalico.org/v1"
-	"github.com/tigera/operator/pkg/common"
-	"github.com/tigera/operator/pkg/components"
-	"github.com/tigera/operator/pkg/controller/status"
-	"github.com/tigera/operator/pkg/controller/utils"
-	"github.com/tigera/operator/pkg/render/applicationlayer"
-	"github.com/tigera/operator/test"
-
 	appsv1 "k8s.io/api/apps/v1"
 	batchv1 "k8s.io/api/batch/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
@@ -41,7 +31,16 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
+	operatorv1 "github.com/tigera/operator/api/v1"
+	"github.com/tigera/operator/pkg/apis"
+	crdv1 "github.com/tigera/operator/pkg/apis/crd.projectcalico.org/v1"
+	"github.com/tigera/operator/pkg/common"
+	"github.com/tigera/operator/pkg/components"
+	"github.com/tigera/operator/pkg/controller/status"
+	"github.com/tigera/operator/pkg/controller/utils"
 	ctrlrfake "github.com/tigera/operator/pkg/ctrlruntime/client/fake"
+	"github.com/tigera/operator/pkg/render/applicationlayer"
+	"github.com/tigera/operator/test"
 )
 
 var _ = Describe("Application layer controller tests", func() {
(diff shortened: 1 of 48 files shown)


pkg/controller/applicationlayer/applicationlayer_controller_test.go
```diff
@@ -41,6 +40,8 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	ctrlrfake "github.com/tigera/operator/pkg/ctrlruntime/client/fake"
```
tmjd, line 44: If you're tweaking imports, please put this one with the other tigera/operator imports.

pkg/controller/logstorage/initializer/initializing_controller.go
```diff
@@ -212,9 +212,6 @@ func (r *LogStorageInitializer) Reconcile(ctx context.Context, request reconcile
 		return reconcile.Result{}, err
 	}
 
-	// Determine if Kibana is enabled for this cluster.
-	kibanaEnabled := !operatorv1.IsFIPSModeEnabled(install.FIPSMode) && !r.multiTenant
```
tmjd, line 216: I'd prefer to keep `kibanaEnabled := !r.multiTenant` so it is documented by code what we're using r.multiTenant being false means.

pkg/controller/logstorage/secrets/secret_controller.go
```diff
@@ -261,9 +261,6 @@ func (r *SecretSubController) Reconcile(ctx context.Context, request reconcile.R
 	// Provision secrets and the trusted bundle into the cluster.
 	hdler := utils.NewComponentHandler(reqLogger, r.client, r.scheme, ls)
 
-	// Determine if Kibana should be enabled for this cluster.
-	kibanaEnabled := !operatorv1.IsFIPSModeEnabled(install.FIPSMode) && !r.multiTenant
```
tmjd, line 265: Same here I'd like to keep `kibanaEnabled := !r.multiTenant` for documentation purposes.

pkg/render/typha.go
```diff
@@ -529,7 +529,6 @@ func (c *typhaComponent) typhaEnvVars() []corev1.EnvVar {
 		{Name: "TYPHA_CAFILE", Value: c.cfg.TLS.TrustedBundle.MountPath()},
 		{Name: "TYPHA_SERVERCERTFILE", Value: c.cfg.TLS.TyphaSecret.VolumeMountCertificateFilePath()},
 		{Name: "TYPHA_SERVERKEYFILE", Value: c.cfg.TLS.TyphaSecret.VolumeMountKeyFilePath()},
-		{Name: "TYPHA_FIPSMODEENABLED", Value: operatorv1.IsFIPSModeEnabledString(c.cfg.Installation.FIPSMode)},
```
tmjd, line 532: Doesn't this need to remain?

rene-dekker: No. In our earliest iteration we would use flags at runtime to create a fips validated tls configuration. In a later iteration we did this at compile time using build flags. However, the unused env variables were not removed until now.

pkg/controller/logstorage/secrets/secret_controller.go
rene-dekker: I addressed this in the code using comments only.

pkg/controller/logstorage/initializer/initializing_controller.go
rene-dekker: I addressed this in the code using comments only.

tmjd (approved):
LGTM

PR #3458, opened 2024-08-12
[EV-5126] Rename and update tigera-crds clusterrole

ti-afra:
Changes:
1. Rename "tigera-crds" clusterrole to "tigera-apiserver"
2. Add "authorizationreviews" rbac to "tigera-apiserver" clusterrole; queryserver needs this rbac to return authorized list of items in the resultset.
3. Rename "tigera-apiserver-access-tigera-crds" clusterrolebinding to "tiger-apiserver"

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


pkg/render/apiserver.go
```diff
@@ -1906,3 +1914,25 @@ func (c *apiServerComponent) multiTenantManagedClusterAccessClusterRoles() []cli
 
 	return objects
 }
+
+func (c *apiServerComponent) getDeprecatedResources() []client.Object {
+	renamedRscList := []client.Object{}
```
rene-dekker, line 1919: nit: it could be replaced with empty slice declaration:

```
var renamedRscList []client.Object
```

pkg/render/apiserver.go
```diff
@@ -279,6 +279,9 @@ func (c *apiServerComponent) Objects() ([]client.Object, []client.Object) {
 		// Explicitly delete any global OSS objects.
 		// Namespaced objects will be handled by namespace deletion.
 		objsToDelete = append(objsToDelete, globalCalicoObjects...)
+
+		// Explicitly delete any renamed/deprecated objects.
+		objsToDelete = append(objsToDelete, c.getDeprecatedResources()...)
```
rene-dekker, line 284: I think this can be outside (under) the if-else statement. This always needs to get deleted, regardless of whether the cluster is enterprise right now.

pkg/render/apiserver.go
```diff
@@ -1258,26 +1261,31 @@ func (c *apiServerComponent) tigeraCustomResourcesClusterRole() *rbacv1.ClusterR
 				"patch",
 			},
 		},
+		{
```
rene-dekker, line 1264: Please if you could add a comment as to why we need this permission, it will help us track over time why we added permissions and also make it easier to remove unused permissions.

pkg/render/apiserver.go
ti-afra: I moved it outside, but for opensource we dont have the "tigera-crds" clusterrole and the apiserver binding. Doesn't it matter or wouldn't it fail if they dont exist?

rene-dekker (approved):
lgtm

PR #3460, opened 2024-08-14
Introducing sidecarInjection to ApplicationLayer resource

radixo:
## Description

Those changes are introduced to make the operator capable to install and control the new Sidecar for L7 features.

Changes:
* Create the SidecarInjection property into the ApplicationLayer resource, which will make the operator to understand when to install or remove the sidecar injection dependencies.
* Install the MutatingWebhookConfiguration into the cluster for sidecar injection when the property is "Enabled"
* Install the L7AdmissionControler web server inside ApiServer pods, as a new container, to control the creation of new pods that has the label "applicationlayer.projectcalico.org/sidecar" = "true", when the sidecar injection property is "Enabled" as well.

Design doc of the new feature as follows:
https://docs.google.com/document/d/1C5VKry_HLTLZq6VmpwaznpSjwYm4aYqeEpeN54S2fas/edit?usp=sharing

P.S. The introduced feature doesn't break any of the current L7 Application Layer features (WAF, Logging and ALP).

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


pkg/render/apiserver.go
```diff
@@ -65,6 +67,12 @@ const (
 	tigeraAPIServerTLSSecretName                    = "tigera-apiserver-certs"
 	APIServerSecretsRBACName                        = "tigera-extension-apiserver-secrets-access"
 	MultiTenantManagedClustersAccessClusterRoleName = "tigera-managed-cluster-access"
+	L7AdmssCtrlContainerName                        = "calico-l7admssctrl"
+	L7AdmssCtrlEnvoyImage                           = "envoyproxy/envoy:v1.31-latest"
+	L7AdmssCtrlPort                                 = 6443
+
+	// Sidecar
```
caseydavenport, line 74:
```suggestion
```

api/v1/applicationlayer_types.go
```diff
@@ -36,11 +36,17 @@ type ApplicationLayerSpec struct {
 	// L7LogCollectorDaemonSet configures the L7LogCollector DaemonSet.
 	// +optional
 	L7LogCollectorDaemonSet *L7LogCollectorDaemonSet `json:"l7LogCollectorDaemonSet,omitempty"`
+
+	// SidecarInjection controls whether or not sidecar injection is enabled for the cluster.
+	// When enabled, sidecar containers can be injected into pods to provide additional L7 functionality.
+	// +optional
+	SidecarInjection *SidecarStatusType `json:"sidecarInjection,omitempty"`
```
caseydavenport, line 43: It's not super clear to me how this field interacts with the other fields in here.

A few questions I have:
- Is it valid to have `SidecarInjection: Enabled` and all the other fields be Disabled? If so, what is the intended behavior of the system under that configuration?
- What is the "additional L7 functionality" this provides? My understanding was that sidecar was just a deployment technique for all of the existing L7 functionality (WAF, logs, ALP, etc).

pkg/controller/applicationlayer/applicationlayer_controller.go
```diff
@@ -252,7 +252,7 @@ func (r *ReconcileApplicationLayer) Reconcile(ctx context.Context, request recon
 
 	var passthroughModSecurityRuleSet bool
 	var modSecurityRuleSet *corev1.ConfigMap
-	if r.isWAFEnabled(&instance.Spec) {
+	if r.isWAFEnabled(&instance.Spec) || r.isSidecarInjectionEnabled(&instance.Spec) {
```
caseydavenport, line 255: This seems to suggest that `SidecarInjection` is an alternative to WAF, whereas I had been thinking of `SidecarInjection` as a config knob on WAF (and other L7 features).

e.g., 

```yaml
spec:
  WebApplicationFirewall: Enabled
  SidecarInjection: Enabled
```

^ This seems valid to me. I would expect WAF to be enabled, and implemented using a sidecar.

```yaml
spec:
  WebApplicationFirewall: Enabled
  SidecarInjection: Disabled
```

^ This also seems valid to me. I would expect WAF to be enabled, but implemented using a per-host proxy.

```yaml
spec:
  WebApplicationFirewall: Disabled
  SidecarInjection: Enabled
```

^ This feels like an invalid configuration to me - from a UX perspective, injecting a sidecar doesn't do anything on its own unless the sidecar is configured to perform WAF / ALP / L7 logging / etc. as well.

caseydavenport (changes requested):
First quick pass - mostly nits, but a couple of API level questions that will help me evaluate some of the other bits of code.

pkg/controller/applicationlayer/applicationlayer_controller.go
caseydavenport: @radixo OK, so I am sort of following that - sounds like it is valid to have both daemonset and sidecar injection enabled at the same time.

> because it will depend on the pod annotations if WAF is enabled for that por or not, 

Right, so there will be a new `applicationlayer.projectcalico.org/waf` annotation that will control whether or not WAF is enabled for sidecar-based pods (and similar for other L7 features)

Does that mean it is viable to have _only_ sidecar injection enabled (with no DaemonSet)?

e.g.,

```yaml
spec:
  sidecarInjection: Enabled
  webApplicationFirewall: Disabled
```

If so, this is a confusing API, since the user is saying "Disable WAF" but what is actually being implemented is "Enable WAF per-pod instead of per-host"

Not sure how we could make that better, but figured it should be called out. Perhaps we just need detailed API comments / documentation to explain that.

pkg/render/applicationlayer/applicationlayer.go
```diff
@@ -397,14 +433,16 @@ func (c *component) volumes() []corev1.Volume {
...
+					Path: "/var/run/dikastes",
+					Type: &hostPathDirectoryOrCreate,
+				},
 			},
 		})
 
 		// Needed for ModSecurity library - contains rule set.
-		if c.config.WAFEnabled { // WAF-only
+		if c.config.WAFEnabled || c.config.SidecarInjectionEnabled {
 			// WAF logs need HostPath volume - logs to be consumed by fluentd.
```
caseydavenport, line 445: Does this mean that the per-host daemonset is still responsible for L7 logging even if the sidecar mode is being used?

pkg/render/apiserver.go
```diff
@@ -2023,3 +2128,51 @@ func (c *apiServerComponent) getDeprecatedResources() []client.Object {
...
+		Env: []corev1.EnvVar{
+			corev1.EnvVar{
+				Name:  "L7ADMCTRL_TLSCERTPATH",
+				Value: c.cfg.TLSKeyPair.VolumeMountCertificateFilePath(),
+			},
+			corev1.EnvVar{
+				Name:  "L7ADMCTRL_TLSKEYPATH",
+				Value: c.cfg.TLSKeyPair.VolumeMountKeyFilePath(),
+			},
+			corev1.EnvVar{
```
caseydavenport, line 2156: Just wanted to confirm - these are the only images that the admission controller is going to inject as sidecars right? 

Need to make sure that any images we deploy are configured via the operator (i.e., it would be an anti-pattern to have the admission controller make its own decisions about what images and versions to install).

caseydavenport: Part of me wonders if this admission controller should have been an endpoint on the tigera/operator, thinking about it a bit more. But it's definitely too late to make that call for this particular release.

pkg/render/applicationlayer/applicationlayer.go
```diff
@@ -92,12 +93,17 @@ type Config struct {
...
-	proxyImage      string
-	collectorImage  string
-	dikastesImage   string
-	dikastesEnabled bool
-	envoyConfigMap  *corev1.ConfigMap
+	proxyImage            string
+	collectorImage        string
+	dikastesImage         string
+	dikastesEnabled       bool
+	envoyEnabled          bool
```
caseydavenport, line 104: I assume `envoyEnabled` means `perHostEnvoyEnabled`? 

I think we should refactor these names to be explicit about whether they are referring to per-host, sidecar, or "any" version of the component for clarity.

pkg/render/applicationlayer/applicationlayer.go
```diff
@@ -149,20 +155,28 @@ func (c *component) Objects() ([]client.Object, []client.Object) {
...
 	// Envoy configuration
-	c.config.envoyConfigMap = c.envoyL7ConfigMap()
-	objs = append(objs, c.config.envoyConfigMap)
+	if c.config.envoyEnabled {
+		c.config.envoyConfigMap = c.envoyL7ConfigMap()
+		objs = append(objs, c.config.envoyConfigMap)
+	}
 
 	// Envoy & Dikastes Daemonset
 	objs = append(objs, c.daemonset())
```
caseydavenport, line 182: Do we need to install the daemonset even if we're only configured to use sidecars (and not per-host envoy)?

If so, we should have a detailed comment explaining why.

pkg/render/applicationlayer/applicationlayer.go
```diff
@@ -311,6 +329,10 @@ func (c *component) containers() []corev1.Container {
 			)
 		}
 
+		if c.config.ALPEnabled {
```
caseydavenport, line 332: Same here - should probably be `PerHostALPEnabled` or `AnyALPEnabled`, for clarity (with field level comments in the struct).

api/v1/applicationlayer_types.go
```diff
@@ -36,11 +36,20 @@ type ApplicationLayerSpec struct {
 	// L7LogCollectorDaemonSet configures the L7LogCollector DaemonSet.
 	// +optional
 	L7LogCollectorDaemonSet *L7LogCollectorDaemonSet `json:"l7LogCollectorDaemonSet,omitempty"`
+
+	// SidecarInjection controls whether or not sidecar injection is enabled for the cluster.
+	// When enabled, containers that has the label
```
caseydavenport, line 41:
```suggestion
	// When enabled, pods with the label
```

api/v1/applicationlayer_types.go
```diff
@@ -36,11 +36,20 @@ type ApplicationLayerSpec struct {
...
+	// "applicationlayer.projectcalico.io/sidecar"="true" will have injected
+	// sidecar, that will replace system firewall rules for L7
+	// functionality, like WAF and ALP.
```
caseydavenport, line 44:
```suggestion
	// "applicationlayer.projectcalico.org/sidecar"="true" will have their L7 functionality 
	// such as WAF and ALP implemented using an injected sidecar instead of a per-host proxy.
	// The per-host proxy will continue to be used for pods without this label.
```

api/v1/applicationlayer_types.go
```diff
@@ -36,11 +36,20 @@ type ApplicationLayerSpec struct {
...
+	// +optional
+	SidecarInjection *SidecarStatusType `json:"sidecarInjection,omitempty"`
 }
 
 type LogCollectionStatusType string
 type WAFStatusType string
 type ApplicationLayerPolicyStatusType string
+type SidecarStatusType string
```
caseydavenport, line 52:
```suggestion

// +kubebuilder:validation:Enum=Enabled;Disabled
type SidecarStatusType string
```

caseydavenport: (really the others should have this too but don't need to make that change in this PR if you don't want)

pkg/render/apiserver.go
```diff
@@ -2023,3 +2128,51 @@ func (c *apiServerComponent) getDeprecatedResources() []client.Object {
...
+		*cfg.ApplicationLayer.Spec.SidecarInjection == operatorv1.SidecarEnabled
+}
+
+func (c *apiServerComponent) l7AdmissionControllerContainer() corev1.Container {
+	volumeMounts := []corev1.VolumeMount{
+		c.cfg.TLSKeyPair.VolumeMount(c.SupportedOSType()),
+	}
+
+	l7AdmssCtrl := corev1.Container{
+		Name:            L7AdmissionControllerContainerName,
```
caseydavenport, line 2144: We need to add this name as a valid option in the API server overrides section: https://github.com/tigera/operator/blob/6eeefb15c96cfabe0ce747b053101726c1276df5/api/v1/apiserver_types.go#L77-L90

We should also add a UT that verifies it's possible to set resource requirements for this new container like these here: https://github.com/tigera/operator/blob/6eeefb15c96cfabe0ce747b053101726c1276df5/pkg/render/apiserver_test.go#L1025-L1034

pkg/render/applicationlayer/applicationlayer.go
```diff
@@ -288,13 +304,15 @@ func (c *component) containers() []corev1.Container {
 			{Name: DikastesSyncVolumeName, MountPath: "/var/run/dikastes"},
 		}
 
-		if c.config.WAFEnabled {
+		if c.config.WAFEnabled || c.config.SidecarInjectionEnabled {
 			commandArgs = append(
 				commandArgs,
-				"--waf-enabled",
 				"--waf-log-file", filepath.Join(CalicologsVolumePath, "waf", "waf.log"),
```
caseydavenport, line 310: should `--waf-log-file` go under the `if c.config.WAFEnabled` section below? What aobut `--waf-ruleset-file`?

At face value these both seem like arguments that are only relevant when WAF is enabled.

pkg/controller/applicationlayer/applicationlayer_controller.go
caseydavenport: I think we leave this as-is for now.

I think we should consider cleaning this API up in a future version though. As a strawman, something like this seems more accurate:

```yaml
hostEnforcement:
  webApplicationFirewall: Disabled
  l7Logging: Enabled
sidecarEnforcement: Enabled
```

caseydavenport (changes requested):
OK, second round done. Will keep looking through this.

Overall I don't see any major bugs in the code, but I want to take a few more looks as I continue to get a better feel for the feature set and desired behaviors.

pkg/controller/applicationlayer/applicationlayer_controller.go
radixo: Yeah, it's confusing... totally agree. I created this issue for Runtime Security backlog
https://tigera.atlassian.net/browse/RS-2061

pkg/render/applicationlayer/applicationlayer.go
radixo: Yes... actually the per-host daemonset is still doing everything for l7 features in EE, on sidecars, only the envoy proxy is injected on the application pods.

pkg/render/apiserver.go
radixo: Yes... only this images. The sidecars are only two initContainers, the first one is dikastes that sets the iptables rules for the pod namespace, create the envoy configuration files and exits 0. The envoy is the an initContainer with the native sidecars k8s pattern. That will send the ext_authz and logs to the dikastes and l7-log-collector through unix sockets.

pkg/render/applicationlayer/applicationlayer.go
radixo: The `--waf-enabled` is telling dikastes that the per-host enforcement is enabled for WAF, needed for internal decisions on requests that can come even from sidecars or TPROXY iptables rules. So the `--waf-log-file` and `--waf-ruleset-file` are the configs that are required by both.

api/v1/apiserver_types.go
```diff
@@ -78,7 +78,7 @@ func init() {
 type APIServerDeploymentContainer struct {
 	// Name is an enum which identifies the API server Deployment container by name.
 	// Supported values are: calico-apiserver, tigera-queryserver
-	// +kubebuilder:validation:Enum=calico-apiserver;tigera-queryserver
+	// +kubebuilder:validation:Enum=calico-apiserver;tigera-queryserver;calico-l7admssctrl
```
caseydavenport, line 81: Given this name shows up on a user-facing API, can we change the container name to avoid abbreviations?

```
calico-l7-admission-controller
```

pkg/render/applicationlayer/applicationlayer.go
caseydavenport: Perfect.

Honestly, the renaming of WAFEnabled -> PerHostWAFEnabled makes this SO much easier to read for me now - this makes sense.

caseydavenport: > The --waf-enabled is telling dikastes that the per-host enforcement is enabled for WA

It is probably a good idea to rename this argument on Dikastes (or add a new one and deprecate this one)....

```
--per-host-waf-enabled
```

^ Would be more accurate.

caseydavenport (approved):
Only minor comments this time - otherwise this LGTM.

Thanks much @radixo !

radixo:
Thanks a million @caseydavenport 

>It is probably a good idea to rename this argument on Dikastes (or add a new one and deprecate this one)....
>
>--per-host-waf-enabled
>^ Would be more accurate.

I will clarify them on Dikastes and open a small PR for operator to fix it.

PR #3464, opened 2024-08-16
Remove Job + SA when it is a managed cluster. Only render SA if Job i…

rene-dekker:


vara2504 (approved):
LGTM. Do we have a PR for the master branch?

PR #3472, opened 2024-08-21
Add IP pool config to disable new allocations

caseydavenport:
## Description

Add new operator API field to IP pools that is equivalent to the exisitng IP pool spec.disabled option.

This is necessary if we want to implement IP pool migration logic from this guide using the operator API: https://docs.tigera.io/calico/3.27/networking/ipam/migrate-pools#tutorial

Separately, I do wonder if we could make pool migraiton even easier in operator mode - could the operator automatically do most of those steps?

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


pkg/controller/ippool/defaults.go
```diff
@@ -203,6 +203,11 @@ func fillDefaults(ctx context.Context, client client.Client, instance *operator.
 			}
 		}
 
+		if pool.DisableNewAllocations == nil {
+			disabled := false
```
Brian-McM, line 207: If you're interested, I did make a convenience method in the ptr package to get points without having to allocate a non pointer variable (and it's generic) `ptr.ToPtr(false)` would return a pointer to a false value.

Brian-McM (approved):

PR #3475, opened 2024-08-23
Respect HTTP proxies when rendering Guardian policy

pasanw:
## Description
The operator will inspect the tigera-guardian pod to determine if a proxy is in use, and update egress policy rules accordingly.

The proxy environment variables are read from the Guardian pod as it is expected they are injected via a mutating admission webhook on the pod spec.

The approach limits to amount of queries done for pods, in anticipation of pods no longer being cached by the operator. The query for pods is only made when the cluster-connection controller has observed a new availability condition and that condition indicates that the deployment is available.
 

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/clusterconnection/clusterconnection_controller.go
+++ b/pkg/controller/clusterconnection/clusterconnection_controller.go
@@ -19,6 +19,12 @@ import (
 	"fmt"
 	"net"
 
+	"github.com/tigera/operator/pkg/url"
+	"golang.org/x/net/http/httpproxy"
+	v1 "k8s.io/api/apps/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+
 	"github.com/go-logr/logr"
 
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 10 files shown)


pkg/controller/clusterconnection/clusterconnection_controller.go
```diff
@@ -19,6 +19,12 @@ import (
 	"fmt"
 	"net"
 
+	"github.com/tigera/operator/pkg/url"
```
rene-dekker, line 22: Needs formatting

pkg/controller/clusterconnection/clusterconnection_controller.go
```diff
@@ -323,6 +335,67 @@ func (r *ReconcileConnection) Reconcile(ctx context.Context, request reconcile.R
...
+			Namespace:     render.GuardianNamespace,
+		})
+		if err != nil {
+			r.status.SetDegraded(operatorv1.ResourceReadError, "Failed to list the pods of the Guardian deployment", err, reqLogger)
+			return reconcile.Result{}, nil
+		}
+
+		// Parse the pod spec to resolve the proxy config.
+		var proxyConfig *httpproxy.Config
+		for _, pod := range pods.Items {
```
Brian-McM, line 378: Since we're already considering it viable that the operator might have a different proxy address than the guardian pods (which is why we're getting the proxy address directly from the guardian pod), it would probably be equally viable that the guardian pods have different proxy addresses from each other (in the event that we did allow multiple guardian pods, which would be a good idea to do eventually).

It probably wouldn't be much more work to collect a list of proxy configs, and apply them all to the policy.

I think with that addition it makes the solution fairly bullet proof.

Brian-McM (changes requested):

PR #3481, opened 2024-08-29
Calico apiserver improvements

Tamas-Biro1:
## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


api/v1/apiserver_types.go
```diff
@@ -87,6 +87,9 @@ type APIServerDeploymentContainer struct {
 	// If used in conjunction with the deprecated ComponentResources, then this value takes precedence.
 	// +optional
 	Resources *v1.ResourceRequirements `json:"resources,omitempty"`
+
+	// +optional
+	Lifecycle *v1.Lifecycle `json:"lifecycle,omitempty" protobuf:"bytes,12,opt,name=lifecycle"`
```
tmjd, line 92: I'd like to hear the use case for adding this. (I hadn't heard of Lifecycle before but looking at the K8s API reference I don't understand what its use case would be for the apiserver.)

api/v1/apiserver_types.go
```diff
@@ -185,6 +198,25 @@ type APIServerDeploymentSpec struct {
...
 	// +optional
 	Template *APIServerDeploymentPodTemplateSpec `json:"template,omitempty"`
+
+	// The deployment strategy to use to replace existing pods with new ones.
+	// +optional
+	// +patchStrategy=retainKeys
+	Strategy *APIServerDeploymentStrategy `json:"strategy,omitempty" patchStrategy:"retainKeys" protobuf:"bytes,4,opt,name=strategy"`
+}
+
+// TyphaDeploymentStrategy describes how to replace existing pods with new ones.  Only RollingUpdate is supported
```
tmjd, line 208: copy paste error in the comment `TyphaDeploymentStrategy`

api/v1/apiserver_types.go
```diff
@@ -185,6 +198,25 @@ type APIServerDeploymentSpec struct {
 	// Template describes the API server Deployment pod that will be created.
 	// +optional
 	Template *APIServerDeploymentPodTemplateSpec `json:"template,omitempty"`
+
+	// The deployment strategy to use to replace existing pods with new ones.
+	// +optional
+	// +patchStrategy=retainKeys
+	Strategy *APIServerDeploymentStrategy `json:"strategy,omitempty" patchStrategy:"retainKeys" protobuf:"bytes,4,opt,name=strategy"`
```
tmjd, line 205: I'd want to hear what the use case is that we need to add this override for the apiserver. In other words why would someone need to change the Strategy?

api/v1/apiserver_types.go
```diff
@@ -173,6 +180,12 @@ type APIServerDeployment struct {
 
 // APIServerDeploymentSpec defines configuration for the API server Deployment.
 type APIServerDeploymentSpec struct {
+	// Replicas defines how many instances of the API server pod will run.
+	// +optional
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:validation:Maximum=2147483647
+	Replicas *int32 `json:"replicas,omitempty"`
```
tmjd, line 187: The replicas for the apiserver can already be controlled through the Installation resource with the `controlPlaneReplicas` field.
I'm not saying this wouldn't be useful but just wanted to point out it is possible to change the number of replicas.

tmjd (commented):
Some thoughts and use case requests around these changes.

api/v1/apiserver_types.go
Tamas-Biro1: thanks, fixed

api/v1/apiserver_types.go
Tamas-Biro1: i guess this could be useful if using `preStop` hook to gracefully manage IPAM cleanup when calico-node shut down. But haven't tested this scenario so i'm removing these lines for now. Maybe i can create a new PR later when properly tested it out. Thanks for your comments.

api/v1/apiserver_types.go
Tamas-Biro1: you are right! Removing these lines, thanks for pointing this out.

api/v1/apiserver_types.go
Tamas-Biro1: removing these lines for now

api/v1/apiserver_types.go
Tamas-Biro1: you are right, no need to allow overriding the strategy. But it would be better to set it to RollingUpdate instead of Recreate to increase HA during an update.

Tamas-Biro1: @tmjd I think the `RollingUpdate` is better because it replaces pods gradually to ensure minimal downtime and new pods are created before old ones are terminated. The `Recreate` just deletes all existing pods simultaneously and then creates new ones, causing downtime.

caseydavenport: Yeah, I think RollingUpdate is preferable most of the time. There are some subset of scenarios where Recreate is necessary - namely for 1 node clusters running on EKS, or similar platforms, that run our API server in host network.

Can we configure RollingUpdate when hostNet: false, and Recreate otherwise? would that work?

pkg/render/apiserver.go
```diff
@@ -1068,8 +1070,10 @@ func (c *apiServerComponent) apiServerDeployment() *appsv1.Deployment {
 					Tolerations:        c.tolerations(),
 					ImagePullSecrets:   secret.GetReferenceList(c.cfg.PullSecrets),
 					InitContainers:     initContainers,
-					Containers:         containers,
-					Volumes:            c.apiServerVolumes(),
+					Containers: []corev1.Container{
```
tmjd, line 1073: I think this needs to go back to `Containers: containers,` or else there are more changes needed here to account for the l7AdmissionControllerContainer that is conditionally added to the containers above.

api/v1/apiserver_types.go
Tamas-Biro1: Yes, configured to use RollingUpdate when `hostNetwork: false`. thanks for the suggestion.

caseydavenport (approved):

caseydavenport:
@Tamas-Biro1 A few tests seem to be failing:

```

Failure
/go/src/github.com/tigera/operator/pkg/render/apiserver_test.go:360
Expected
    <v1.DeploymentStrategyType>: RollingUpdate
to equal
    <v1.DeploymentStrategyType>: Recreate
/go/src/github.com/tigera/operator/pkg/render/apiserver_test.go:197
```

PR #3485, opened 2024-09-01
Add new NonClusterHost resource for log ingestion

hjiawei:
## Description

This change introduces a new `NonClusterHost` custom resource for Calico Enterprise to support non-cluster hosts log ingestion.

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


api/v1/nonclusterhost_types.go
```diff
@@ -0,0 +1,53 @@
...
+package v1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// NonClusterHostSpec enables non-cluster hosts to connect to a cluster.
+type NonClusterHostSpec struct {
+	// Location of the log ingestion point for non-cluster hosts. example: https://1.2.3.4:443
+	Endpoint string `json:"endpoint"`
```
dimitri-nicolo, line 25: This seems like a required field. Should we add:
``json:"endpoint" validate:"required"``

dimitri-nicolo: It might be worth writing a custom validator, so that the string input that does not match the a predefined regex pattern prints an error at the time of creation.

api/v1/nonclusterhost_types.go
```diff
@@ -0,0 +1,53 @@
...
+// +kubebuilder:resource:scope=Cluster
+
+// NonClusterHost installs the components required for non-cluster host log collection.
+// At most one instance of this resource is supported. It must be named "tigera-secure".
+type NonClusterHost struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	// Specification of the desired state for non-cluster host log collection.
+	Spec NonClusterHostSpec `json:"spec,omitempty"`
```
dimitri-nicolo, line 38: What about adding a Status?

If yes, then: add the `// +kubebuilder:subresource:status` tag

api/v1/nonclusterhost_types.go
hjiawei: Validation pattern is added to the endpoint field.

rene-dekker: I wonder if the field name should be made more specific. Like logstorageEndpoint. I imagine we may have a need for a second endpoint for other purposes, for example an endpoint that will be used for connecting to typha.

pkg/render/manager.go
```diff
@@ -549,6 +551,11 @@ func (c *managerComponent) voltronContainer() corev1.Container {
 		mounts = append(mounts, c.cfg.TunnelServerCert.VolumeMount(c.SupportedOSType()))
 		mounts = append(mounts, c.cfg.VoltronLinseedKeyPair.VolumeMount(c.SupportedOSType()))
 	}
+	if c.cfg.NonClusterHost != nil {
+		// We need the internal tls key pair for voltron and linseed mtls connection
+		// when non-cluster host log ingestion is enabled.
+		mounts = append(mounts, c.cfg.InternalTLSKeyPair.VolumeMount(c.SupportedOSType()))
```
rene-dekker, line 557: This creates the possibility for the secret to be mounted twice. You should combine it with the if-statement just above this one.

api/v1/nonclusterhost_types.go
```diff
@@ -0,0 +1,54 @@
...
+	// +kubebuilder:validation:Pattern=`^https://.+$`
+	Endpoint string `json:"endpoint"`
+}
+
+// +kubebuilder:object:root=true
+// +kubebuilder:resource:scope=Cluster
+
+// NonClusterHost installs the components required for non-cluster host log collection.
+// At most one instance of this resource is supported. It must be named "tigera-secure".
+type NonClusterHost struct {
```
caseydavenport, line 34: As far as naming goes, I think we should consider something else.  Since k8s APIs are typically representative of an object, this naming implies that this object represents a single non-cluster host (i.e., there would be one per host) whereas I think this is actually just general configuration. 

What other types of configuration do you foresee us putting in this resource? If it's just a log storage endpoint, then perhaps it belongs in the LogStorage resource? But I suspect there is more planned for this.

pkg/render/manager.go
hjiawei: They are combined.

api/v1/nonclusterhost_types.go
rene-dekker: Since NonClusterHostSpec is not a required field, please mind that it is still possible to create this resource without the Endpoint field. Does that have any implications for the feature? Should the controller still validate that the field exists?

api/v1/nonclusterhost_types.go
rene-dekker: We spoke offline and I think we will proceed with the current name.

pkg/controller/manager/manager_controller.go
```diff
@@ -640,6 +643,17 @@ func (r *ReconcileManager) Reconcile(ctx context.Context, request reconcile.Requ
 		return reconcile.Result{}, err
 	}
 
+	// Check if non-cluster host log ingestion is enabled.
+	// non-cluster hosts will only forward logs to a standalone or a managed cluster.
+	var nonclusterhost *operatorv1.NonClusterHost
+	if managementCluster == nil {
```
rene-dekker, line 649: I was wondering if there is any harm in removing this if-statement.

hjiawei: If a nonclusterhost resource is accidentally applied to a management cluster, this check prevents opening the ingestion endpoint.

rene-dekker: I meant to say that I don't really see the problem in having a management cluster that also has non cluster hosts working at the same time.

api/v1/nonclusterhost_types.go
caseydavenport: NonClusterHostSpec isn't required, but it's also not a pointer, which means it will be present regardless and so this field should always be defaulted by the kubebuilder validation IIUC

pkg/controller/manager/manager_controller.go
caseydavenport: At a minimum, I would prefer if we inverted this check - i.e., "If both are specified, error" instead of silently ignoring an unsupported configuration.

I'm fine with erroring for now if MCM is out of scope, but also fine to allow it in management clusters if we think it will come relatively for free.

pkg/controller/utils/utils.go
```diff
@@ -439,6 +439,17 @@ func GetManagementClusterConnection(ctx context.Context, c client.Client) (*oper
 	return managementClusterConnection, nil
 }
 
+// GetNonClusterHost finds the NonClusterHost CR in your cluster.
+func GetNonClusterHost(ctx context.Context, cli client.Client) (*operatorv1.NonClusterHost, error) {
+	nonclusterhost := &operatorv1.NonClusterHost{}
+	err := cli.Get(ctx, DefaultTSEEInstanceKey, nonclusterhost)
```
caseydavenport, line 445: So, this function isn't following the pattern that the rest of the `GetX` functions follow (namely, they swallow IsNotFound errors).

Personally, I dislike how the other functions swallow the IsNotFound error and find them somewhat confusing. But we probably still want to be consistent with them to avoid further confusion.

rene-dekker: I checked that too, but our getX funcs are very inconsistent. Some swallow, others don't... :/

caseydavenport: :crying_cat_face:

api/v1/nonclusterhost_types.go
rene-dekker: I applied the CRD to my kind cluster:

```
$ k apply -f nonclust.yaml 
The NonClusterHost "default" is invalid: spec.endpoint: Required value

 1   12:17:55   ~/dev/kind    
$ cat nonclust.yaml 
kind: NonClusterHost
metadata:
  name: default
spec: {}
```

But then:

```
$ k apply -f nonclust.yaml 
nonclusterhost.operator.tigera.io/default configured
 0   12:18:26   ~/dev/kind    
$ cat nonclust.yaml 
kind: NonClusterHost
metadata:
  name: default
spec: null
```

caseydavenport: hmm, guess I was wrong about that. Surprised that `spec: null` is valid at all considering the spec isn't a pointer.

rene-dekker: It is identical to this: 
```
kind: NonClusterHost
metadata:
  name: default
```

pkg/controller/utils/utils.go
hjiawei: This function is updated to follow others. It will swallow `IsNotFound` errors.

api/v1/nonclusterhost_types.go
hjiawei: `spec.endpoint` is validated in the manager controller. When NonClusterHost resource is applied to the cluster and endpoint filed is an invalid URL, manager controller will be degraded.

pkg/controller/manager/manager_controller.go
hjiawei: `spec.endpoint` is validated in the manager controller. When `NonClusterHost` resource is applied to the cluster and the `endpoint` filed is an invalid URL, manager controller will be degraded.

The `managementCluster == nil` check is also removed because management cluster should support this feature.

rene-dekker (approved):
lgtm

PR #3487, opened 2024-09-03
Dynamically add network policy for apiserver egress with configured K8SServicEndpoint

tmjd:
For Enterprise, where the operator creates and manages policy to allow components to function in the presence of default deny policy, it was found that there was not appropriate allow policy when the K8s apiserver is defined with a K8SServiceEndpoint. This PR add an Egress rule to allow the needed traffic.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/apiserver.go
+++ b/pkg/render/apiserver.go
@@ -499,6 +499,14 @@ func allowTigeraAPIServerPolicy(cfg *APIServerConfiguration) *v3.NetworkPolicy {
 		},
 	}...)
 
+	if r, err := cfg.K8SServiceEndpoint.DestinationEntityRule(); r != nil && err == nil {
+		egressRules = append(egressRules, v3.Rule{
+			Action:      v3.Allow,
+			Protocol:    &networkpolicy.TCPProtocol,
+			Destination: *r,
+		})
+	}
+
 	// The ports Calico Enterprise API Server and Calico Enterprise Query Server are configured to listen on.
 	ingressPorts := networkpolicy.Ports(443, APIServerPort, QueryServerPort, 10443)
 	if cfg.IsSidecarInjectionEnabled() {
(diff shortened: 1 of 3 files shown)


pkg/render/apiserver_test.go
```diff
@@ -662,6 +664,28 @@ var _ = Describe("API server rendering tests (Calico Enterprise)", func() {
...
+		policy := testutils.GetAllowTigeraPolicyFromResources(policyName, resources)
+		Expect(policy).ToNot(BeNil())
+		Expect(policy.Spec).ToNot(BeNil())
+		Expect(policy.Spec.Egress).ToNot(BeNil())
+		Expect(policy.Spec.Egress).To(ContainElement(calicov3.Rule{
+			Action:   calicov3.Allow,
+			Protocol: &networkpolicy.TCPProtocol,
+			Destination: calicov3.EntityRule{
+				Ports:   networkpolicy.Ports(1234),
+				Domains: []string{"k8shost"},
```
tomastigera, line 684: The test took me a bit by surprise since the host is (most) often an IP :sweat_smile:

tmjd: When using EKS it is a long AWS DNS name.

pkg/render/apiserver.go
```diff
@@ -499,6 +499,14 @@ func allowTigeraAPIServerPolicy(cfg *APIServerConfiguration) *v3.NetworkPolicy {
 		},
 	}...)
 
+	if r, e := cfg.K8SServiceEndpoint.DestinationEntityRule(); r != nil && e == nil {
```
caseydavenport, line 502: can we just call the error `err`? Made me do a double-take here and I don't think there's a code reason it needs to be named something else.

pkg/render/apiserver_test.go
caseydavenport: I think we probably want a test that verifies the rule generation for when it is an IP as well though right?

caseydavenport (approved):

PR #3488, opened 2024-09-04
[RS-1843] DPI Snort rules configurable via intrusiondetection CR.

bartolini:
Original JIRA ticket: https://tigera.atlassian.net/browse/RS-1843


api/v1/intrusiondetection_types.go
```diff
@@ -36,6 +36,27 @@ type IntrusionDetectionSpec struct {
 	// IntrusionDetectionControllerDeployment configures the IntrusionDetection Controller Deployment.
 	// +optional
 	IntrusionDetectionControllerDeployment *IntrusionDetectionControllerDeployment `json:"intrusionDetectionControllerDeployment,omitempty"`
+
+	// DeepPacketInspectionDaemonset configures the DPI Daemonset
+	// +optional
+	DeepPacketInspectionDaemonset *DeepPacketInspectionDaemonset `json:"deepPacketInspectionDaemonset,omitempty"`
+}
+
+type DeepPacketInspectionDaemonset struct {
```
caseydavenport, line 45: The pattern we want to use here is to create a struct that mimics the behavior and structure of the `v1.DaemonSet` object in upstream kubernetes (with only the fields that we want to expose).

So, we want to include the full resource hierarchy here. Take a look at [IntrusionDetectionControllerDeployment](https://github.com/tigera/operator/blob/dcb21dfb148f47b5dfd39bd136398e04c4c21935/api/v1/intrusiondetection_types.go#L100-L105) for an example.

In this case, something like this:

```go
type DeepPacketInspectionDaemonset struct {
  Spec *DPIDaemonsetSpec
}

type DPIDaemonsetSpec struct {
  Template *DPIDaemonsetTemplate
}

type DPIDaemonsetTemplate struct {
  Spec *DPIDaemonsetTemplateSpec
}

type DPIDaemonsetTemplateSpec struct {
  InitContainers []DPIDaemonsetInitContainer
}

type DPIDaemonsetInitContainer struct {
  Image string
  Resources *corev1.ResourceRequirements
}
```

this is more verbose than what you have written, so might seem like overkill, but allows us to extend in the future if we want to allow further configuration of the DaemonSet (e.g., metadata.labels, metadata.annotations, or other spec fields) and is identitcal to the structure of a built-in k8s resource that users are already very familiar with.

pkg/render/intrusiondetection/dpi/dpi.go
```diff
@@ -180,6 +183,25 @@ func (d *dpiComponent) dpiDaemonset() *appsv1.DaemonSet {
...
+				Image:           snortInitContainer.Image,
+				ImagePullPolicy: corev1.PullAlways,
+				VolumeMounts: []corev1.VolumeMount{
+					{
+						Name:      DeepPacketInspectionSnortRulesVolumeName,
+						MountPath: DeepPacketInspectionSnortRulesVolumePath,
+					},
+				},
+			}
+			if snortInitContainer.Resources != nil {
```
caseydavenport, line 199: Instead of doing this explicitly, we should follow the established "overrides" pattern: https://github.com/tigera/operator/blob/dcb21dfb148f47b5dfd39bd136398e04c4c21935/pkg/render/intrusion_detection.go#L538-L540

caseydavenport: There is an equivalent function for DaemonSet: https://github.com/tigera/operator/blob/dcb21dfb148f47b5dfd39bd136398e04c4c21935/pkg/render/common/components/components.go#L103-L126

You would just need to make your struct implement this interface: https://github.com/tigera/operator/blob/dcb21dfb148f47b5dfd39bd136398e04c4c21935/pkg/components/overrides.go#L24-L60

Most functions will return `nil` to indicate that we don't support overrides on them, but the `GetInitContainers` will merge in the resources appropriately.

pkg/render/intrusiondetection/dpi/dpi.go
```diff
@@ -262,6 +284,17 @@ func (d *dpiComponent) dpiVolumes() []corev1.Volume {
 			})
 	}
 
+	if len(d.cfg.IntrusionDetection.Spec.DeepPacketInspectionDaemonset.SnortInitContainers) > 0 {
```
caseydavenport, line 287: This (and the volume mount equivalent below) need comments explaining what's going on here (basically, a mini description of the design for this featur) - this is an unusual pattern for our code and I don't think it's immediately obvious why the presence of init containers in the IntrusionDetection API means we should include an empty dir volume.

pkg/render/intrusiondetection/dpi/dpi.go
```diff
@@ -180,6 +183,25 @@ func (d *dpiComponent) dpiDaemonset() *appsv1.DaemonSet {
 	if d.cfg.TyphaNodeTLS.NodeSecret.UseCertificateManagement() {
 		initContainers = append(initContainers, d.cfg.TyphaNodeTLS.NodeSecret.InitContainer(DeepPacketInspectionNamespace))
 	}
+	if d.cfg.IntrusionDetection.Spec.DeepPacketInspectionDaemonset.SnortInitContainers != nil {
+		for iter, snortInitContainer := range d.cfg.IntrusionDetection.Spec.DeepPacketInspectionDaemonset.SnortInitContainers {
+			container := corev1.Container{
+				Name:            fmt.Sprintf("%s-%d", DeepPacketInspectionInitContainerName, iter),
```
caseydavenport, line 189: I think we should have a well-known name for the "snort rules init container" (e.g., `Name: "snort-rules"`) and also require that the user input that name into the `Name` field in the `IntrusionDetection.DeepPacketInspectionDaemonSet.Spec.Template.Spec.InitContainers` API.

Given we use validation to ensure there is only a single init container anyway, I don't think this generative approach is correct -- if in the future we want to expose other non-snort-related init containers we will want the ability to name them appropriately and not based on the order they happen to be put into the API.

pkg/render/intrusiondetection/dpi/dpi.go
```diff
@@ -180,6 +183,25 @@ func (d *dpiComponent) dpiDaemonset() *appsv1.DaemonSet {
...
+				Image:           snortInitContainer.Image,
+				ImagePullPolicy: corev1.PullAlways,
```
caseydavenport, line 191: We shouldn't leave the pull policy as always. We have a helper function for this: https://github.com/tigera/operator/blob/master/pkg/render/intrusion_detection.go#L659

pkg/render/intrusiondetection/dpi/dpi.go
```diff
@@ -180,6 +183,25 @@ func (d *dpiComponent) dpiDaemonset() *appsv1.DaemonSet {
 	if d.cfg.TyphaNodeTLS.NodeSecret.UseCertificateManagement() {
 		initContainers = append(initContainers, d.cfg.TyphaNodeTLS.NodeSecret.InitContainer(DeepPacketInspectionNamespace))
 	}
+	if d.cfg.IntrusionDetection.Spec.DeepPacketInspectionDaemonset.SnortInitContainers != nil {
```
caseydavenport, line 186: I think this will cause a panic if `d.cfg.IntrusionDetection.Spec.DeepPacketInspectionDaemonset` is `nil`

caseydavenport (changes requested):

api/v1/intrusiondetection_types.go
bartolini: That's now corrected.

pkg/render/intrusiondetection/dpi/dpi.go
bartolini: We agreed this is OK.

pkg/render/intrusiondetection/dpi/dpi.go
bartolini: Corrected.

pkg/render/intrusiondetection/dpi/dpi.go
bartolini: Corrected as well.

pkg/render/intrusiondetection/dpi/dpi.go
bartolini: Corrected and comments added.

pkg/render/intrusiondetection/dpi/dpi.go
bartolini: Implemented as suggested.

api/v1/intrusiondetection_types.go
```diff
@@ -36,6 +36,46 @@ type IntrusionDetectionSpec struct {
...
+
+type DPIDaemonsetTemplateSpec struct {
+	// List of DPI Daemonset Init containers definitions
+	// +kubebuilder:validation:MaxItems=1
+	InitContainers []DPIDaemonsetInitContainer `json:"initContainers,omitempty"`
+}
+
+type DPIDaemonsetInitContainer struct {
+	// Container name
+	Name string `json:"name"`
```
caseydavenport, line 71:
```suggestion
	Name string `json:"name"`
	
```

api/v1/intrusiondetection_types.go
```diff
@@ -36,6 +36,46 @@ type IntrusionDetectionSpec struct {
...
+	// Image name for the init container
+	Image string `json:"image"`
```
caseydavenport, line 73:
```suggestion
	Image string `json:"image"`
	
```

api/v1/intrusiondetection_types.go
caseydavenport: (newline between fields)

api/v1/intrusiondetection_types.go
```diff
@@ -36,6 +36,46 @@ type IntrusionDetectionSpec struct {
...
+}
+
+type DPIDaemonsetTemplateSpec struct {
+	// List of DPI Daemonset Init containers definitions
+	// +kubebuilder:validation:MaxItems=1
+	InitContainers []DPIDaemonsetInitContainer `json:"initContainers,omitempty"`
+}
+
+type DPIDaemonsetInitContainer struct {
+	// Container name
```
caseydavenport, line 70: We should have validation on this field - we only support a single container here right now (the snort rules init container), so we should reject any thing that doesn't match a pre-established conventional name (e.g., "snort-rules").

For example: 
https://github.com/tigera/operator/blob/dcb21dfb148f47b5dfd39bd136398e04c4c21935/api/v1/calico_node_types.go#L28-L29

pkg/render/intrusiondetection/dpi/dpi_test.go
```diff
@@ -52,6 +52,20 @@ var (
 		TypeMeta:   metav1.TypeMeta{Kind: "IntrusionDetection", APIVersion: "v1"},
 		ObjectMeta: metav1.ObjectMeta{Name: "tigera-secure"},
 		Spec: operatorv1.IntrusionDetectionSpec{
+			DeepPacketInspectionDaemonset: &operatorv1.DeepPacketInspectionDaemonset{
```
caseydavenport, line 55: I think this change (plus the one below) means we are now only testing the codepath where `DeepPacketInspectionDaemonset != nil`

I would like to see test cases that execute this code with `DeepPacketInspectionDaemonset: nil` as well, to ensure that we're not hitting any nil pointer exceptions and to verify that existing behavior is unchanged when the DaemonSet overrides are not specified.

Instead of modifying the existing tests, could you add a new `Context("with snort rules init container specified")` and leave the existing tests in place?

api/v1/intrusiondetection_types.go
```diff
@@ -36,6 +36,46 @@ type IntrusionDetectionSpec struct {
...
+	Name string `json:"name"`
+	// Image name for the init container
+	Image string `json:"image"`
+	// Resources allows customization of limits and requests for compute resources such as cpu and memory.
+	// If specified, this overrides the named SnortInitContainer Daemonset init container's resources.
+	// If omitted, the SnortInitContainer Deployment will use its default value for this init container's resources.
```
caseydavenport, line 76: SnortInitContainer isn't a deployment

api/v1/intrusiondetection_types.go
```diff
@@ -36,6 +36,46 @@ type IntrusionDetectionSpec struct {
...
+	InitContainers []DPIDaemonsetInitContainer `json:"initContainers,omitempty"`
+}
+
+type DPIDaemonsetInitContainer struct {
+	// Container name
+	Name string `json:"name"`
+	// Image name for the init container
+	Image string `json:"image"`
+	// Resources allows customization of limits and requests for compute resources such as cpu and memory.
+	// If specified, this overrides the named SnortInitContainer Daemonset init container's resources.
```
caseydavenport, line 75: SnortInitContainer isn't a DaemonSet

api/v1/intrusiondetection_types.go
bartolini: Corrected.

api/v1/intrusiondetection_types.go
bartolini: Corrected as well.

api/v1/intrusiondetection_types.go
bartolini: Done.

pkg/render/intrusiondetection/dpi/dpi_test.go
bartolini: Done.

caseydavenport (approved):
LGTM - thanks @bartolini 

Once this is merged, please let's put up another PR to adjust how the `Resources` are handled to follow the conventional "Overrides" pattern.

PR #3489, opened 2024-09-10
Use restricted namespace for opensource apiserver

mihivagyok:
## Description

- the opensource calico-apiserver is now running as a non-root container, however the calico-apiserver namespace has privileged labels which is not necessary
- the calico-apiserver pods are happy with the restricted namespace label

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/apiserver.go
+++ b/pkg/render/apiserver.go
@@ -267,9 +267,13 @@ func (c *apiServerComponent) Objects() ([]client.Object, []client.Object) {
 		namespacedEnterpriseObjects = append(namespacedEnterpriseObjects, c.cfg.TrustedBundle.ConfigMap(QueryserverNamespace))
 	}
 
+	podSecurityNamespaceLabel := PodSecurityStandard(PSSRestricted)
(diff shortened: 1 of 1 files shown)


caseydavenport:
@mihivagyok we do actually need privileged PSS in some cases, specifically because in some modes of operation we run using `hostNetwork: true`.

I think we can adjust the PSS based on that though - if we're running with hostNet, we need privileged - otherwise, can run using restricted.

mihivagyok:
@caseydavenport Thanks, you are right, I see it now in the code. I will work on the adjustments. Thanks!

mihivagyok:
Hi @caseydavenport ! Thank you once more. I made the changes, please take a look.
Also, I would like to note that I would like cherry-pick this back to 1.34 and 1.35. Thank you very much!

caseydavenport (approved):

caseydavenport:
LGTM, thanks @mihivagyok

PR #3491, opened 2024-09-12
Fix image component handling, honoring Installation spec

petrutlucian94:
components.GetReference currently ignores the image prefix and path specified in the "Installation" spec.

It makes a copy of the image component, appends the configured image path and prefix but then it uses the original component image when looping over the image set.

The fix is trivial, we just have to use the modified image string.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/components/references.go
+++ b/pkg/components/references.go
@@ -85,12 +85,12 @@ func GetReference(c Component, registry, imagePath, imagePrefix string, is *oper
 	}
 
 	for _, img := range is.Spec.Images {
-		if img.Image == c.Image {
+		if img.Image == image {
 			return fmt.Sprintf("%s%s@%s", registry, image, img.Digest), nil
 		}
 	}
 
-	return "", fmt.Errorf("ImageSet did not contain image %s", c.Image)
+	return "", fmt.Errorf("ImageSet did not contain image %s", image)
 }
 
 func ReplaceImagePath(image, imagePath string) string {


petrutlucian94:
Looks like there is another blocker that prevents us from using custom built images:

https://github.com/tigera/operator/blob/ec635bf82a312806adf9ad210acf5623579e96fa/pkg/controller/utils/imageset/imageset.go#L141

```
"ImageSet calico-v3.28.0: unexpected images: petrutlucian94/calico-node, petrutlucian94/calico-cni, petrutlucian94/calico-typha, petrutlucian94/calico-kube-controllers, petrutlucian94/calico-tigera-operator, petrutlucian94/calico-csi, petrutlucian94/calico-apiserver, petrutlucian94/calico-ctl, petrutlucian94/calico-pod2daemon-flexvol, petrutlucian94/calico-key-cert-provisioner",
```

tmjd:
For an ImageSet we need a reference to match an ImageSet Digest to the image it will be used with. I would suggest we document that the "Image" value in the ImageSet is the original (unmodified) image name and ignores any registry, path, or prefix changes.
This would ensure there is no change in behavior for users that have figured out this is the way the ImageSets work.

tmjd:
Yes that is the documentation to update. You could add a note to that which clarifies the value is the original (unmodified) image name and ignores any registry, path, or prefix changes.
Whatever you think would clarify the existing behavior.
To be clear you will need to drop the existing changes you currently have in this PR, since what you have added would be a change in behavior.

petrutlucian94:
> To be clear you will need to drop the existing changes you currently have in this PR, since what you have added would be a change in behavior.

Please correct me if I'm wrong, but without this change the ``ImagePrefix`` setting of the ``Installation`` spec is unusable:
https://github.com/tigera/operator/blob/2a88b6ba495004c1954c0c29fbccde58e7f954cc/api/v1/installation_types.go#L64-L75

So this behaviour change is actually a bug fix.

To give a little context, we hit this issue while working on Canonical "rock" image replacements:

* https://documentation.ubuntu.com/rockcraft/en/stable/
* https://github.com/canonical/calico-rocks/blob/33f3dff6ed3b9da4aa372869726dc5fa74436aa8/tests/integration/test_calico.py#L55-L57

For now we're just applying this patch but thought that others may benefit from it as well so I opened a PR.

tmjd:
Without your modification it should work fine I believe.

Here I'm including the important bits of the configurations that I think illustrate what I believe a configuration using the current behavior and what I understand you are configuring would do.
It does require understanding the link between the original calico image name and your custom image name to generate the ImageSet.

You could have the image example.com/custompath/customprefix-calico:v3.17.1 loaded into your custom registry and I'm assuming the digest for that image is `sha256:0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef`.

Have Installation of
```
kind: Installation
metadata:
  name: default
spec:
  # Image format:<registry>/<imagePath>/<imagePrefix><imageName>:<image-tag>
  registry: example.com
  imagePath: "custompath"
  imagePrefix: "customprefix-"
```

and ImageSet of
```
kind: ImageSet
metadata:
  name: calico-v3.17.1
spec:
  images:
    - image: 'calico/node'
      digest: 'sha256:0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef'
   ....
```

and I believe the operator will resolve the calico node image to
`example.com/custompath/customprefix-calico@sha256:0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef`

Please let me know what you think is incorrect about what I'm suggesting.

petrutlucian94:
My bad, thank you so much for the clarification. I've created a separate PR that aims to improve the docstring, please let me know what you think: https://github.com/tigera/operator/pull/3635

PR #3492, opened 2024-09-12
revert dikastes flags rename that featured 'per-host'

electricjesus:
This PR fixes the breaking changes introduced in tigera/operator#3460

- dikastes already has a SUBSCRIPTION_TYPE env var that changes its operation mode from per-host to per-pod
- Operator only deploys per-host mode dikastes (SUBSCRIPTION_TYPE is set to `per-host-policies`)
- the ALP feature switches mode based on aforementioned env var
- the WAF feature doesn't care about SUBSCRIPTION_TYPE, functions the same regardless
- Flag `--waf-enabled`  must be retained because WAF is also used in [Ingress Gateway integration](https://docs.tigera.io/calico-cloud/threat/deploying-waf-ingress-gateway)
- new `--alp-enabled` flag will behave the same for both per-host and per-pod dikastes run mode

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/applicationlayer/applicationlayer.go
+++ b/pkg/render/applicationlayer/applicationlayer.go
@@ -316,7 +316,7 @@ func (c *component) containers() []corev1.Container {
 				"--waf-ruleset-file", filepath.Join(ModSecurityRulesetVolumePath, "tigera.conf"),
 			)
 			if c.config.PerHostWAFEnabled {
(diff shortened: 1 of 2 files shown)


caseydavenport (approved):
LGTM - I still think this flag should be renamed but we can take our time to do so.

PR #3493, opened 2024-09-13
Render NonClusterHost resources

hjiawei:
## Description

This changeset adds a nonclusterhost controller to render NonClusterHost related resources like ServiceAccount, ClusterRole, and ClusterRoleBinding. When a NonClusterHost resource is created in a cluster, SA and token secret will be created automatically inside the `calico-system` namespace. Corresponding ClusterRole and ClusterRoleBindings will also be created for non-cluster host calico-node and calico-fluent-bit agents to talk to the cluster and forward logs to the ingestion endpoint.

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


pkg/controller/nonclusterhost/nonclusterhost_controller.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+
+	return add(mgr, c)
+}
+
+func newReconciler(mgr manager.Manager, opts options.AddOptions) reconcile.Reconciler {
+	r := &ReconcileNonClusterHost{
+		client: mgr.GetClient(),
+		scheme: mgr.GetScheme(),
+		status: status.New(mgr.GetClient(), "nonclusterhost", opts.KubernetesVersion),
```
rene-dekker, line 63: We use kebab-case for this: `non-cluster-hosts` (I think plural would probably be my preference).

pkg/controller/nonclusterhost/nonclusterhost_controller.go
```diff
@@ -0,0 +1,135 @@
...
+
+	instance, err := utils.GetNonClusterHost(ctx, r.client)
+	if err != nil {
+		r.status.SetDegraded(operatorv1.ResourceReadError, "Failed to query NonClusterHost resource", err, logc)
+		return reconcile.Result{}, err
+	}
+
+	ch := utils.NewComponentHandler(logc, r.client, r.scheme, instance)
+
+	if instance == nil {
```
rene-dekker, line 97: This block is not necessary.

If the resources are created previously, they would have also had an ownerref associated to the NonClusterHost CR. So if that CR is deleted, all associated resources would also be deleted automatically.

pkg/render/nonclusterhost/nonclusterhost.go
```diff
@@ -0,0 +1,212 @@
...
+type nonClusterHostComponent struct {
+	cfg *Config
+}
+
+func (c *nonClusterHostComponent) ResolveImages(is *operatorv1.ImageSet) error {
+	return nil
+}
+
+func (c *nonClusterHostComponent) SupportedOSType() rmeta.OSType {
+	return rmeta.OSTypeLinux
```
rene-dekker, line 53: I think this should be OSTypeAny; no pods are deployed as a result of this.

pkg/render/nonclusterhost/nonclusterhost.go
```diff
@@ -0,0 +1,212 @@
...
+				"stagedglobalnetworkpolicies",
+				"stagedkubernetesnetworkpolicies",
+				"stagednetworkpolicies",
+				"tiers",
+			},
+			Verbs: []string{"get", "list", "watch"},
+		},
+	}
+
+	// calico fluent-bit rules
```
rene-dekker, line 162: Thanks for thoroughly commenting this func!

pkg/render/nonclusterhost/nonclusterhost.go
```diff
@@ -0,0 +1,212 @@
...
+			// The annotation below will result in the auto-creation of spec.data.token.
+			Annotations: map[string]string{
+				"kubernetes.io/service-account.name": NonClusterHostObjectName,
+			},
+		},
+		Type: "kubernetes.io/service-account-token",
+	}
+}
+
+func ClusterRole() *rbacv1.ClusterRole {
```
hjiawei, line 101: Has to be a ClusterRole becuase calico node will read/write some cluster wide resources.

pkg/controller/nonclusterhost/nonclusterhost_controller.go
hjiawei: Changed to `non-cluster-hosts`.

pkg/controller/nonclusterhost/nonclusterhost_controller.go
hjiawei: This resource deletion part is removed (leaving it to k8s owner reference management). The status is still set to CRNotFound and returned.

pkg/render/nonclusterhost/nonclusterhost.go
hjiawei: Changed to `OSTypeAny`.

pkg/render/node.go
hjiawei, line 1: Reorder by resource names. No real changes.

rene-dekker (approved):
lgtm

PR #3503, opened 2024-09-18
Reduce the number of restart for tigera-manager

asincu:
## Description

Tigera-manager is restarted by operator in a multi-tenant environment due to the following:
- we are picking up kibana certificates instead of elastic and we are setting the annotations to use kibana certificate. At the next reconciliation, this changes. This is happening because external kibana and external elastic have the value and thus the same hash. Certificate manager only keeps one of them, as it stores as a hash for the certificate. We do not need external kibana certificates and we can stop rendering them. (How we store certs for the trusted bundle: https://github.com/tigera/operator/blob/2f071c36038db04ff46aa0de9b0422efe74daff9/pkg/tls/certificatemanagement/certificatebundle.go#L104)

```
diff tigera-linseed-566bbc65cf-cc-tenant-u45b0nkt tigera-linseed-5db6cd6fcd-cc-tenant-u45b0nkt 
7,10c7,9
<     deployment.kubernetes.io/revision: "3"
<     deployment.kubernetes.io/revision-history: "1"
<   creationTimestamp: "2024-09-16T18:39:49Z"
<   generation: 1
---
>     deployment.kubernetes.io/revision: "2"
>   creationTimestamp: "2024-09-16T21:37:18Z"
>   generation: 2
14,15c13,14
<     pod-template-hash: 566bbc65cf
<   name: tigera-linseed-566bbc65cf
---
>     pod-template-hash: 5db6cd6fcd
>   name: tigera-linseed-5db6cd6fcd
24,25c23,24
<   resourceVersion: "696063"
<   uid: ec4c7d2f-60b0-41ed-a6b7-1f79767c3093
---
>   resourceVersion: "696071"
>   uid: feb1aa56-42f0-447d-9ea9-4dc4e9e5d655
27c26
<   replicas: 1
---
>   replicas: 0
31c30
<       pod-template-hash: 566bbc65cf
---
>       pod-template-hash: 5db6cd6fcd
41c40
<         tigera-operator.hash.operator.tigera.io/tigera-secure-es-http-certs-public: ba4e8391ac56eed2f5bf4458d05f25f92d65c672
---
>         tigera-operator.hash.operator.tigera.io/tigera-secure-kb-http-certs-public: ba4e8391ac56eed2f5bf4458d05f25f92d65c672
46c45
<         pod-template-hash: 566bbc65cf
---
>         pod-template-hash: 5db6cd6fcd
249,253c248,249
<   availableReplicas: 1
<   fullyLabeledReplicas: 1
[...]

PR #3504, opened 2024-09-18
[cherry-pick][v1.35]Reduce the number of restart for tigera-manager (#3503)

asincu:
Tigera-manager is restarted by operator in a multi-tenant environment due to the following:
- we are picking up kibana certificates instead of elastic and we are setting the annotations to use kibana certificate. At the next reconciliation, this changes. This is happening because external kibana and external elastic have the value and thus the same hash. Certificate manager only keeps one of them, as it stores as a hash for the certificate. We do not need external kibana certificates and we can stop rendering them
- voltron routes are configured by different components (cloud rbac, image assurance) and can produce different volume mounts/volumes which causes tigera-manager to create a new replica set.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/secrets/tenant_controller.go
+++ b/pkg/controller/secrets/tenant_controller.go
@@ -227,9 +227,8 @@ func (r *TenantController) upstreamCertificates(cm certificatemanager.Certificat
 	}
 
 	if r.elasticExternal {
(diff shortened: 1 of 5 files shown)


tmjd (approved):
LGTM

PR #3513, opened 2024-09-20
Add EnvoyProxyEnvoy to gen-versions

danudey:
With the new EnvoyProxyEnvoy image added, we also need to add this image to gen-versions so that our generated versions files have the correct component information.


radixo:
LGTM

marvin-tigera:

PR #3518, opened 2024-09-24
Append TIGERA_TPROXY_ENABLED to per host envoy

radixo:
## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/applicationlayer/applicationlayer.go
+++ b/pkg/render/applicationlayer/applicationlayer.go
@@ -361,6 +361,7 @@ func (c *component) proxyEnv() []corev1.EnvVar {
 		// envoy needs to run as root to be able to use transparent flag (for tproxy)
 		{Name: "ENVOY_UID", Value: "0"},
 		{Name: "ENVOY_GID", Value: "0"},
+		{Name: "TIGERA_TPROXY_ENABLED", Value: ""},
 	}
 }
 
--- a/pkg/render/applicationlayer/applicationlayer.go
+++ b/pkg/render/applicationlayer/applicationlayer.go
@@ -361,7 +361,7 @@ func (c *component) proxyEnv() []corev1.EnvVar {
 		// envoy needs to run as root to be able to use transparent flag (for tproxy)
 		{Name: "ENVOY_UID", Value: "0"},
 		{Name: "ENVOY_GID", Value: "0"},
-		{Name: "TIGERA_TPROXY_ENABLED", Value: ""},
+		{Name: "TIGERA_TPROXY", Value: "Enabled"},
 	}
 }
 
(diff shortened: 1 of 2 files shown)


pkg/render/applicationlayer/applicationlayer.go
```diff
@@ -361,6 +361,7 @@ func (c *component) proxyEnv() []corev1.EnvVar {
 		// envoy needs to run as root to be able to use transparent flag (for tproxy)
 		{Name: "ENVOY_UID", Value: "0"},
 		{Name: "ENVOY_GID", Value: "0"},
+		{Name: "TIGERA_TPROXY_ENABLED", Value: ""},
```
tmjd, line 364: I think we should be more explicit and having a value of true or false specified or I'd like to see the name be TIGERA_TPROXY and the values of Enabled or Disabled.

radixo: This one follows as other envoy env vars are used, like "DISABLE_TEST_MERGE", where we just check for the nullptr when the envvar is not declared to go with the default behaviour of the software. In this case the TIGERA_PROXY will be always Disabled except when explicitly enabled.

tmjd: It feels ambiguous to see what is effectively `TIGERA_TPROXY_ENABLED=""`, I tried to take a look at `DISABLE_TEST_MERGE` and couldn't find it. I also tried to find envoy environment variables to see how they use environment variables to see if there was a significant amount of "prior art" with envoy environment variables. Could you share where these environment variables are defined/used?

electricjesus:
```suggestion
		{Name: "TIGERA_TPROXY", Value: "Enabled"},
```

pkg/render/applicationlayer/applicationlayer_test.go
```diff
@@ -175,6 +175,7 @@ var _ = Describe("Tigera Secure Application Layer rendering tests", func() {
 		expectedProxyEnvs := []corev1.EnvVar{
 			{Name: "ENVOY_UID", Value: "0"},
 			{Name: "ENVOY_GID", Value: "0"},
+			{Name: "TIGERA_TPROXY_ENABLED", Value: ""},
```
electricjesus, line 178:
```suggestion
			{Name: "TIGERA_TPROXY", Value: "Enabled"},
```

pkg/render/applicationlayer/applicationlayer_test.go
```diff
@@ -462,6 +463,7 @@ var _ = Describe("Tigera Secure Application Layer rendering tests", func() {
...
```
electricjesus, line 466:
```suggestion
			{Name: "TIGERA_TPROXY", Value: "Enabled"},
```

pkg/render/applicationlayer/applicationlayer_test.go
```diff
@@ -620,6 +622,7 @@ var _ = Describe("Tigera Secure Application Layer rendering tests", func() {
...
```
electricjesus, line 625:
```suggestion
			{Name: "TIGERA_TPROXY", Value: "Enabled"},
```

tmjd (approved):
LGTM

PR #3532, opened 2024-10-07
Order certificates so that we have a predictable output

asincu:
## Description

This is a leftover task from https://github.com/tigera/operator/pull/3503. We would like to order certificates so that we have a predictable output.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/tls/certificatemanagement/certificatebundle.go
+++ b/pkg/tls/certificatemanagement/certificatebundle.go
@@ -19,12 +19,12 @@ import (
 	"fmt"
 	"os"
 	"path"
-	"sort"
-
-	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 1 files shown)


asincu:
@rene-dekker This PR came as a follow-up task on a different PR. We wanted to sort the map of certificates so that we have the same output when generating annotations. When using the namespaced name of the secret as a sorting criteria, we end with some tests failures when checking the content of the bundle. (seems like the add all leaf certificates in the map versus just tigera-ca which is the issuer of the other certificates). Should we be outputting the hash annotations using the issuer instead of the leaf certificate secret name ? Or should we be sorting like we are adding the map (sort based on issuer) ?

PR #3537, opened 2024-10-11
[v1.36] Regenerate CRDs

fasaxc:
## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/crds/calico/policy.networking.k8s.io_adminnetworkpolicies.yaml
+++ b/pkg/crds/calico/policy.networking.k8s.io_adminnetworkpolicies.yaml
@@ -4,7 +4,7 @@ metadata:
   annotations:
     api-approved.kubernetes.io: https://github.com/kubernetes-sigs/network-policy-api/pull/30
     policy.networking.k8s.io/bundle-version: v0.1.1
-    policy.networking.k8s.io/channel: standard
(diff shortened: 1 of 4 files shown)


pkg/crds/calico/policy.networking.k8s.io_adminnetworkpolicies.yaml
```diff
@@ -4,7 +4,7 @@ metadata:
   annotations:
     api-approved.kubernetes.io: https://github.com/kubernetes-sigs/network-policy-api/pull/30
     policy.networking.k8s.io/bundle-version: v0.1.1
-    policy.networking.k8s.io/channel: standard
+    policy.networking.k8s.io/channel: experimental
```
tmjd, line 7: It seems the wrong direction to change this to experimental when standard was here before, though I guess we haven't make AdminNetworkPolicies available before so I guess it is ok for this value to change. I assume this has been reviewed in Calico also.

tmjd (approved):
LGTM

fasaxc:
I think we started off with the conservative "standard" ANP spec and moved to the experimental one because the conservative one was missing some newer additions that we wanted to implement.

PR #3539, opened 2024-10-14
Fix Installation API comment

caseydavenport:
## Description

Whoops

Fixes https://github.com/projectcalico/calico/issues/7590

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/api/v1/installation_types.go
+++ b/api/v1/installation_types.go
@@ -468,8 +468,10 @@ type CalicoNetworkSpec struct {
 	// +kubebuilder:validation:Enum=Enabled;Disabled
 	BGP *BGPOption `json:"bgp,omitempty"`
 
-	// IPPools contains a list of IP pools to create if none exist. At most one IP pool of each
-	// address family may be specified. If omitted, a single pool will be configured if needed.
+	// IPPools contains a list of IP pools to manage. If nil, a single IP pool
+	// will be defaulted. If an empty list is provided, the operator will not create any IP pools and will instead
+	// wait for IP pools to be created out-of-band.
+	// IP pools in this list will be reconciled by the operator and should not be modified out-of-band.
 	// +optional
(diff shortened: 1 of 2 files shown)


Brian-McM (approved):

api/v1/installation_types.go
```diff
@@ -468,8 +468,10 @@ type CalicoNetworkSpec struct {
 	// +kubebuilder:validation:Enum=Enabled;Disabled
 	BGP *BGPOption `json:"bgp,omitempty"`
 
-	// IPPools contains a list of IP pools to create if none exist. At most one IP pool of each
-	// address family may be specified. If omitted, a single pool will be configured if needed.
+	// IPPools contains a list of IP pools to manage. If nil, a single IP pool
+	// will be defaulted. If an empty list is provided, the operator will not create any IP pools and will instead
```
nelljerram, line 472: I'm afraid I don't understand "a single IP pool will be defaulted".  I guess you probably mean "the Tigera operator will create a default IP pool" - but then obvious next questions are (1) IPv4 or IPv6? (or one of each?), (2) what CIDR?

Apart from that, everything else is clear, and I assume also true.

nelljerram (changes requested):

api/v1/installation_types.go
caseydavenport: Thanks @nelljerram - yeah I agree this particular bit is unclear. Will see how I can reword. Thanks for looking!

caseydavenport: I think the main thing is that specifics about defaulting for the "defaulted" IP pool are specified on the IP pool fields themselves, so don't need to be restated here. But worth calling out that the generated pool in this case will be IPv4 at least since CIDR is typically a required field on a user-specified field.

PR #3544, opened 2024-10-16
Add tolerations for arm64 workloads on GKE

hjiawei:
## Description

By default, GKE schedules workloads only to x86-based nodes by by placing a taint (kubernetes.io/arch=arm64:NoSchedule) on all Arm nodes [1]. This taint prevents Calico components from being scheduled on ARM nodes. This change adds toleration to the workload specification as suggested in [2].

[1] https://cloud.google.com/kubernetes-engine/docs/how-to/prepare-arm-workloads-for-deployment#overview
[2] https://cloud.google.com/kubernetes-engine/docs/how-to/prepare-arm-workloads-for-deployment#multi-arch-schedule-any-arch

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/common/meta/meta.go
+++ b/pkg/render/common/meta/meta.go
@@ -60,6 +60,15 @@ var (
 
 	TolerateCriticalAddonsAndControlPlane = append(TolerateControlPlane, TolerateCriticalAddonsOnly)
 
+	// TolerateGKEArm64NoSchedule allows pods to be scheduled on GKE Arm64 nodes.
+	// See https://cloud.google.com/kubernetes-engine/docs/how-to/prepare-arm-workloads-for-deployment#multi-arch-schedule-any-arch
+	TolerateGKEArm64NoSchedule = corev1.Toleration{
+		Key:      "kubernetes.io/arch",
+		Operator: corev1.TolerationOpEqual,
+		Value:    "arm64",
+		Effect:   corev1.TaintEffectNoSchedule,
+	}
+
 	// TolerateAll returns tolerations to tolerate all taints. When used, it is not necessary
 	// to include the user's custom tolerations because we already tolerate everything.
 	TolerateAll = []corev1.Toleration{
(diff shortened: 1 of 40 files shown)


rene-dekker:
The only resource that I don't see in here is egress gateway. I think that also has an arm build.

pkg/render/common/meta/meta.go
```diff
@@ -60,6 +60,15 @@ var (
 
 	TolerateCriticalAddonsAndControlPlane = append(TolerateControlPlane, TolerateCriticalAddonsOnly)
 
+	// TolerateGKEArm64NoSchedule allows pods to be scheduled on GKE Arm64 nodes.
+	// See https://cloud.google.com/kubernetes-engine/docs/how-to/prepare-arm-workloads-for-deployment#multi-arch-schedule-any-arch
+	TolerateGKEArm64NoSchedule = corev1.Toleration{
```
rene-dekker, line 65: In go abreviations should be all caps, so Arm -> ARM.

pkg/render/egressgateway/egressgateway.go
```diff
@@ -153,6 +153,10 @@ func (c *component) deploymentPodTemplate() *corev1.PodTemplateSpec {
 	for _, x := range c.config.PullSecrets {
 		ps = append(ps, corev1.LocalObjectReference{Name: x.Name})
 	}
+	tolerations := []corev1.Toleration{}
```
rene-dekker, line 156: This will leave a warning in goland IDE, best to replace with:
```
var tolerations []corev1.Toleration
```

pkg/render/common/meta/meta.go
hjiawei: Fixed.

pkg/render/egressgateway/egressgateway.go
hjiawei: Fixed.

rene-dekker (approved):
lgtm

PR #3563, opened 2024-10-24
[EV-5299] Disable Policy Recommendation for Windows OS

dimitri-nicolo:
## Description

Disable Image Resolution for Windows
* Policy Recommendations is unsupported until Staged Network Policies are supported in Windows

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/policyrecommendation/policyrecommendation_controller.go
+++ b/pkg/controller/policyrecommendation/policyrecommendation_controller.go
@@ -288,6 +288,17 @@ func (r *ReconcilePolicyRecommendation) Reconcile(ctx context.Context, request r
 		return reconcile.Result{RequeueAfter: utils.StandardRetry}, nil
 	}
 
+	// Policy Recommendation is not supported in combination with Windows nodes.
+	hasWindowsNodes, err := common.HasWindowsNodes(r.client)
+	if err != nil {
+		r.status.SetDegraded(operatorv1.ResourceReadError, "Unable to determine if there are Windows nodes", err, logc)
+		return reconcile.Result{}, err
+	}
+	if hasWindowsNodes {
+		r.status.SetDegraded(operatorv1.ResourceValidationError, "Policy Recommendation is not supported in combination with Windows nodes", err, logc)
+		return reconcile.Result{}, err
+	}
+
 	license, err := utils.FetchLicenseKey(ctx, r.client)
 	if err != nil {
 		if errors.IsNotFound(err) {


pkg/controller/policyrecommendation/policyrecommendation_controller.go
```diff
@@ -430,6 +431,12 @@ func (r *ReconcilePolicyRecommendation) Reconcile(ctx context.Context, request r
 		return reconcile.Result{}, nil
 	}
 
+	// Render a Policy Recommendation component for Windows if the cluster has Windows nodes.
```
rene-dekker, line 434: If I understand it correctly, we cannot have a cluster using policy recommendation and windows nodes at the same time, because windows does not support staged policies.

If that is the case, we should degrade the TigeraStatus immediately if hasWindowsNodes is true. I think we can move this if-statement further up, so we may save some unnecessary calculations, perhaps right underneath the ready checks at line 291.

dimitri-nicolo: Moved the check to below line 291.

It isn't used, but I've kept the check in the render to enforce a defensive pattern

pkg/controller/policyrecommendation/policyrecommendation_controller.go
```diff
@@ -288,6 +288,17 @@ func (r *ReconcilePolicyRecommendation) Reconcile(ctx context.Context, request r
...
+	hasWindowsNodes, err := common.HasWindowsNodes(r.client)
+	if err != nil {
+		r.status.SetDegraded(operatorv1.ResourceNotReady, "OS not found", err, logc)
+		return reconcile.Result{}, err
+	}
+	if hasWindowsNodes {
+		r.status.SetDegraded(operatorv1.ResourceValidationError, "Unsupported OS (Windows)", err, logc)
```
rene-dekker, line 298: I was wondering if we could make it a bit more descriptive, so they understand well what's going on.

> Policy Recommendation is not supported in combination with Windows nodes.

pkg/controller/policyrecommendation/policyrecommendation_controller.go
```diff
@@ -288,6 +288,17 @@ func (r *ReconcilePolicyRecommendation) Reconcile(ctx context.Context, request r
 		return reconcile.Result{RequeueAfter: utils.StandardRetry}, nil
 	}
 
+	// Render a Policy Recommendation component for Windows if the cluster has Windows nodes.
+	hasWindowsNodes, err := common.HasWindowsNodes(r.client)
+	if err != nil {
+		r.status.SetDegraded(operatorv1.ResourceNotReady, "OS not found", err, logc)
```
rene-dekker, line 294: Judging by the implementation of HashWindowsNodes, I think it should be a `ResourceReadError` and perhaps the following message would be better.

> Unable to determine if there are windows nodes.

rene-dekker (approved):
lgtm

marvin-tigera:

marvin-tigera:

marvin-tigera:

marvin-tigera:

PR #3564, opened 2024-10-24
Add VadlidatingAdmissionPolicy RBAC and bump k8s to v1.30 release

hjiawei:
## Description

This changeset adds necessary VadlidatingAdmissionPolicy RBACs for Tigera apiserver and bumps k8s dependency to v1.30 release. This change is tested in a Calico Enterprise cluster.

Related to: https://github.com/projectcalico/calico/pull/9335.

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


api/v1/monitor_types.go
```diff
@@ -90,13 +90,13 @@ type Endpoint struct {
 	HonorTimestamps *bool `json:"honorTimestamps,omitempty"`
 
 	// MetricRelabelConfigs to apply to samples before ingestion.
-	MetricRelabelConfigs []*v1.RelabelConfig `json:"metricRelabelings,omitempty"`
+	MetricRelabelConfigs []v1.RelabelConfig `json:"metricRelabelings,omitempty"`
```
hjiawei, line 93: This is changed in upstream [PR #6479](https://github.com/prometheus-operator/prometheus-operator/pull/6479). @rene-dekker I think this might be the most clean way to adapt to the vendor change. WDYT?

pkg/controller/utils/utils.go
```diff
@@ -165,7 +165,7 @@ func AddDeploymentWatch(c ctrlruntime.Controller, name, namespace string) error
 }
 
 func AddPeriodicReconcile(c ctrlruntime.Controller, period time.Duration, handler handler.EventHandler) error {
-	return c.Watch(&source.Channel{Source: createPeriodicReconcileChannel(period)}, handler)
+	return c.Watch(source.Channel(createPeriodicReconcileChannel(period), handler))
```
hjiawei, line 168: Related to https://github.com/kubernetes-sigs/controller-runtime/pull/2783.

api/v1/monitor_types.go
rene-dekker: We effectively only use this variable in 1 place. I think it would be better to convert the field where it is used, so that we don't need to make breaking changes to our API.

Maybe in pkg/ptr/conversion.go we can add a utility function converting pointer slices to slices using generic?

rene-dekker: Jiawei reminded me that this is not really a breaking change. Everything that is not an array containing nil values should deserialize properly. That paired with the knowledge that this feature is quite new and not used so much yet, makes me think that it is probably ok.

WDYT: @tmjd @caseydavenport

caseydavenport: I think it is OK, but I do also think it is a breaking API change.

For one, anyone importing our code using this field (doubt that's anyone!) will fail to compile when they bump the version.

From an API only perspective, it also means that `metricRelabelConfigs: [null]` is no longer valid? But I suspect that's not a big deal because we probably didn't handle that gracefully anyway :sweat_smile:

All this to say that I'm OK with making this change to make our lives easier, even though technically it's a breaking change.

tmjd: Seems ok to me

rene-dekker (approved):
lgtm

PR #3587, opened 2024-11-05
[For Reference] Example of adding a new API and Controller

caseydavenport:
## Description

This is an example PR to show rougly the changes necessary to add the
skeleton for a new API and Controller.

The commits are broken down and show precisely which commands were run.
Specifically:

1. Generate the new Resource and controller skeleton 

   ```
   operator-sdk create api --group=operator --version=v1 --kind=DemoResource --resource --controller 
   ```

1. Generate manifests

   ```
   make manifests 
   ```

1. Add a new Reconciler to `pkg/controller`, and point the `internal/controller` implementation at it.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


electricjesus:
Love it.. what version of operator-sdk should I use for this? If we have a specific version that works with our `PROJECT` file maybe we could consider a dockerized version as a makefile target perhaps?

PR #3597, opened 2024-11-15
Add egress rule to kubernetes service when configured from configmap

rene-dekker:
Looks like kops does not come up with BPF. Some thoughts for this draft:

Some notes:
- Cluster is configured with configmap to set the kubernetes service, which is then propagated to tigera-apiserver and kube-controller
- Pod networked containers cannot resolve this domain. Host networked pods can.
- Network policy (egress) is missing from ckc to the service, this may be a bug?

Note to self: Is this a provisioning bug or a product bug? Should we do something in order to make pods resolve this service name?


tmjd (approved):
LGTM

PR #3598, opened 2024-11-16
upgrade es and kb to version 8

vara2504:
https://github.com/tigera/calico-private/pull/8119

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/logstorage.go
+++ b/pkg/render/logstorage.go
@@ -40,6 +40,7 @@ import (
 	"github.com/tigera/operator/pkg/common"
 	"github.com/tigera/operator/pkg/components"
 	"github.com/tigera/operator/pkg/dns"
+	"github.com/tigera/operator/pkg/ptr"
 	relasticsearch "github.com/tigera/operator/pkg/render/common/elasticsearch"
 	rmeta "github.com/tigera/operator/pkg/render/common/meta"
 	"github.com/tigera/operator/pkg/render/common/networkpolicy"
@@ -358,15 +359,10 @@ func (es *elasticsearchComponent) podTemplate() corev1.PodTemplateSpec {
 		},
 	}
 
-	sc := securitycontext.NewRootContext(false)
-	// These capabilities are required for docker-entrypoint.sh.
-	// See: https://github.com/elastic/elasticsearch/blob/7.17/distribution/docker/src/docker/bin/docker-entrypoint.sh.
-	// TODO Consider removing for Elasticsearch v8+.
-	sc.Capabilities.Add = []corev1.Capability{
-		"SETGID",
-		"SETUID",
-		"SYS_CHROOT",
-	}
+	sc := securitycontext.NewNonRootContext()
+	// Set the user and group to be the default elasticsearch ID
+	sc.RunAsUser = ptr.Int64ToPtr(1000)
+	sc.RunAsGroup = ptr.Int64ToPtr(1000)
 
 	esContainer := corev1.Container{
 		Name: "elasticsearch",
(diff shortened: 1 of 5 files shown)


pkg/render/logstorage.go
```diff
@@ -359,14 +360,9 @@ func (es *elasticsearchComponent) podTemplate() corev1.PodTemplateSpec {
...
-	// These capabilities are required for docker-entrypoint.sh.
-	// See: https://github.com/elastic/elasticsearch/blob/7.17/distribution/docker/src/docker/bin/docker-entrypoint.sh.
-	// TODO Consider removing for Elasticsearch v8+.
-	sc.Capabilities.Add = []corev1.Capability{
-		"SETGID",
-		"SETUID",
-		"SYS_CHROOT",
-	}
+	// Set the user and group to be the default elasticsearch ID
+	sc.RunAsUser = ptr.Int64ToPtr(1000)
```
rene-dekker, line 364: IIUC, the init containers still need to run as root to:
- set os settings
- chown the file system
and so the namespace still needs to have the privileged context. The main container now runs as 1000:1000, which is the es user. This is also the user the prepareFS script will chown to.

Is my understanding correct? Have we considered changing the user to 10000:10000 (our default user)?

pkg/render/logstorage.go
```diff
@@ -717,9 +713,12 @@ func (es *elasticsearchComponent) nodeSets() []esv1.NodeSet {
...
 func (es *elasticsearchComponent) nodeSetTemplate(pvcTemplate corev1.PersistentVolumeClaim) esv1.NodeSet {
 	config := map[string]interface{}{
-		"node.master":                 "true",
-		"node.data":                   "true",
-		"node.ingest":                 "true",
+		"node.roles": []string{
+			"data",
+			"ingest",
+			"master",
+			"remote_cluster_client",
```
rene-dekker, line 720: Should we remove this role?

pkg/render/logstorage.go
```diff
@@ -359,14 +360,9 @@ func (es *elasticsearchComponent) podTemplate() corev1.PodTemplateSpec {
 	}
 
 	sc := securitycontext.NewRootContext(false)
```
hjiawei, line 362: I think this securitycontext is for the main elasticsearch container. It should use `securitycontext.NewNonRootContext()`.

pkg/render/logstorage.go
hjiawei: I believe this SecurityContext is for the main ElasticSearch container. The SecurityContext for each init container still uses the root context.

vara2504: Init container "elastic-internal-init-filesystem" still run as root because we need to set the vm max map count "echo 262144 > /proc/sys/vm/max_map_count". This change is only for elasticsearch container.
Currently, Elasticsearch 7, we override the security context of the Elasticsearch container to run as the root user and group. The [docker-entrypoint.sh](https://github.com/elastic/elasticsearch/blob/7.17/distribution/docker/src/docker/bin/docker-entrypoint.sh) script in Elasticsearch contained logic to switch to user 1000 when it encountered a root user. However, this logic has been removed in Elasticsearch 8. Since the ownership of the /data folder is set to the default Elasticsearch UID, we are encountering an unwritable error when running as the root user.

pkg/render/logstorage.go
vara2504: We might need these role - data, ingest, master same as the previous version.
Reg:remote_cluster_client :  if we are setting node.roles, then [remote_cluster_client](https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-node.html#node-roles) role must be explicitly included to enable cross-cluster search. Otherwise, this could result in an access denied issue when loading the Stack Monitoring page in Kibana.

pkg/render/logstorage.go
vara2504: Done

pkg/render/logstorage.go
rene-dekker: :/ so be it...

pkg/render/logstorage.go
rene-dekker: We spoke offline about a potential follow-up to explore using uid:gid of 10001:10001, but continue with the PR in its current form.

rene-dekker (approved):
lgtm

PR #3599, opened 2024-11-19
Add oidc certificate and configs for queryserver

ti-afra:
## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/apiserver/apiserver_controller.go
+++ b/pkg/controller/apiserver/apiserver_controller.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/tigera/operator/pkg/render/common/authentication"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
@@ -292,8 +293,10 @@ func (r *ReconcileAPIServer) Reconcile(ctx context.Context, request reconcile.Re
 	var applicationLayer *operatorv1.ApplicationLayer
 	var managementCluster *operatorv1.ManagementCluster
 	var managementClusterConnection *operatorv1.ManagementClusterConnection
+	var keyValidatorConfig authentication.KeyValidatorConfig
 	includeV3NetworkPolicy := false
 	if installationSpec.Variant == operatorv1.TigeraSecureEnterprise {
+		trustedBundle = certificateManager.CreateTrustedBundle()
 		applicationLayer, err = utils.GetApplicationLayer(ctx, r.client)
 		if err != nil {
 			r.status.SetDegraded(operatorv1.ResourceReadError, "Error reading ApplicationLayer", err, reqLogger)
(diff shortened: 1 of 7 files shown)


pkg/controller/apiserver/apiserver_controller.go
```diff
@@ -382,6 +383,43 @@ func (r *ReconcileAPIServer) Reconcile(ctx context.Context, request reconcile.Re
 	// Render the desired objects from the CRD and create or update them.
 	reqLogger.V(3).Info("rendering components")
 
+	var authenticationCR *operatorv1.Authentication
+	// Fetch the Authentication spec. If present, we use to configure user authentication.
```
rene-dekker, line 387:
```suggestion
	// Fetch the Authentication spec. If present, we use it to configure user authentication.
```

pkg/controller/apiserver/apiserver_controller.go
```diff
@@ -382,6 +383,43 @@ func (r *ReconcileAPIServer) Reconcile(ctx context.Context, request reconcile.Re
...
+				err, reqLogger)
+			return reconcile.Result{}, err
+		} else if certificate == nil {
+			reqLogger.Info(fmt.Sprintf("Waiting for secret '%s' to become available", secret))
+			r.status.SetDegraded(operatorv1.ResourceNotReady,
+				fmt.Sprintf("Waiting for secret '%s' to become available", secret),
+				nil, reqLogger)
+			return reconcile.Result{}, nil
+		}
+		trustedBundle.AddCertificates(certificate)
```
rene-dekker, line 417: Theoretically speaking, trustedBundle can be nil here. I'd suggest creating the bundle after this line:

```
if installationSpec.Variant == operatorv1.TigeraSecureEnterprise {
trustedBundle = certificateManager.CreateTrustedBundle()
```

Then change the following line:

```
- trustedBundle = certificateManager.CreateTrustedBundle()
+ trustedBundle.AddCertificates(prometheusCertificate)
```

ti-afra: thanks for catching this :+1:

pkg/render/common/networkpolicy/networkpolicy.go
```diff
@@ -92,6 +93,49 @@ func CreateSourceEntityRule(namespace string, deploymentName string) v3.EntityRu
...
+
+	oidcEgressRules = append(oidcEgressRules, v3.Rule{
+		Action:      v3.Allow,
+		Protocol:    &TCPProtocol,
+		Destination: OIDCEntityRuleExternal,
+	})
+
+	hostnameParts := strings.Split(hostname, ".")
+
+	if len(hostnameParts) == 2 {
```
rene-dekker, line 122: This will turn an external host "idp.org" into a service of ns=idp and name=org. Is that really desirable?

ti-afra: it will, but there is not rule saying that your namespace cannot be called "org" :grin:
I thought this is not a bad compromise unless we want to query the cluster and check if namespace `org` already exists or not.

rene-dekker: Given the time sensitivity of the issue I propose we assume that the idp is external and create egress according to that assumption. As a follow-up we can explore adding the right egress rule in the case that it isn't. OIDC type tigera is something that we don't expect anyone uses in the wild, and in the case of cloud, they won't be needing this egress rule as the queryserver won't get any traffic from idp issued bearer tokens.

pkg/render/common/networkpolicy/networkpolicy.go
```diff
@@ -92,6 +93,27 @@ func CreateSourceEntityRule(namespace string, deploymentName string) v3.EntityRu
 	}
 }
 
+// GetOIDCEgressRule creates egress rule for oidc connection.
+// the result will include an egress rules with the urlString passed in:
+//  1. egress rule: egress rule assuming the oidc is external to the cluster
+func GetOIDCEgressRule(urlString string) v3.Rule {
+	parsedURL, err := url.Parse(urlString)
```
rene-dekker, line 100: Do we check already somewhere that this can be parsed? We don't want this to ever happen.

ti-afra: I think this has to be check when initializing the authentication cr values. Let me check it.

pkg/controller/apiserver/apiserver_controller.go
```diff
@@ -360,7 +364,51 @@ func (r *ReconcileAPIServer) Reconcile(ctx context.Context, request reconcile.Re
...
+		}
+
+		keyValidatorConfig, err = utils.GetKeyValidatorConfig(ctx, r.client, authenticationCR, r.clusterDomain)
+		if err != nil {
+			r.status.SetDegraded(operatorv1.ResourceReadError, "Failed to get KeyValidator Config", err, reqLogger)
+			return reconcile.Result{}, err
+		}
+
+		if keyValidatorConfig != nil {
+			if _, err := url.Parse(keyValidatorConfig.Issuer()); err != nil {
```
rene-dekker, line 408: I see now that this check is not necessary.
The constructor for KeyValidatorConfig will create a well known config, which is obtained through making a request to the URL.
This means that we can remove this check again.

pkg/controller/compliance/compliance_controller.go
```diff
@@ -442,6 +443,13 @@ func (r *ReconcileCompliance) Reconcile(ctx context.Context, request reconcile.R
 		return reconcile.Result{}, err
 	}
 
+	if keyValidatorConfig != nil {
+		if _, err := url.Parse(keyValidatorConfig.Issuer()); err != nil {
```
rene-dekker, line 447: I see now that this check is not necessary.
The constructor for KeyValidatorConfig will create a well known config, which is obtained through making a request to the URL.
This means that we can remove this check again.

rene-dekker (approved):
lgtm

PR #3601, opened 2024-11-19
Adjust connection controller and how it handles the license and network policy

tmjd:
## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/clusterconnection/clusterconnection_controller.go
+++ b/pkg/controller/clusterconnection/clusterconnection_controller.go
@@ -17,15 +17,11 @@ package clusterconnection
 import (
 	"context"
 	"fmt"
-	"net"
 
-	"github.com/tigera/operator/pkg/url"
 	"golang.org/x/net/http/httpproxy"
 	v1 "k8s.io/api/apps/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
-
-	"github.com/go-logr/logr"
 
 	corev1 "k8s.io/api/core/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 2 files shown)


pkg/controller/clusterconnection/clusterconnection_controller.go
```diff
@@ -410,8 +410,8 @@ func (r *ReconcileConnection) Reconcile(ctx context.Context, request reconcile.R
 		return reconcile.Result{RequeueAfter: utils.StandardRetry}, nil
 	}
 
-	// Ensure the allow-tigera tier exists, before rendering any network policies within it.
-	includeV3NetworkPolicy := false
+	// Ensure the allow-tigera tier exists and can support enterprise policy, before rendering any network policies within it.
+	includeEgressNetworkPolicy := false
```
pasanw, line 414: The current implementation gates whether or not we try to render network policy based on whether the tier is present. I'm wondering if we should just change this so that it gates based on whether the license is present. 

I think this achieves the same result as what you've suggested, but the code would read a bit simpler. We would switch the tier check and the license check, e.g.
```
	includeV3NetworkPolicy := false
	if license, err := utils.FetchLicenseKey(ctx, r.Client); err != nil {
		if !k8serrors.IsNotFound(err) {
                   // error
		}	
	} else {
		includeV3NetworkPolicy = true


		if err := r.Client.Get(ctx, client.ObjectKey{Name: networkpolicy.TigeraComponentTierName}, &v3.Tier{}); err != nil {
			// error
		}

		if egressAccessControlRequired(managementClusterConnection.Spec.ManagementClusterAddr, r.resolvedPodProxies, log) {
			if !utils.IsFeatureActive(license, common.EgressAccessControlFeature) {
				// error
			}
		}
	}
```

pasanw: The only time it's not required is if the management cluster host is a domain rather than an IP. I want to say that we should not expect to see a user who uses an IP for their mgmt cluster host and also has a license that does not include EgressAccessControl... but I can't guarantee it?

Nitpicking here, but if we want to proceed without network policy if either license or tier is not found, maybe it could be structured like this?
```
	license, err := utils.FetchLicenseKey(ctx, r.Client)
	if err != nil {
		if !k8serrors.IsNotFound(err) {
			// error
		}
	}
	licenseFound := err != nil
	
	if err := r.Client.Get(ctx, client.ObjectKey{Name: networkpolicy.TigeraComponentTierName}, &v3.Tier{}); err != nil {
		if !k8serrors.IsNotFound(err) {
			// error
		}
	}
	tierFound := err != nil


	
	includeV3NetworkPolicy := false
	if licenseFound && tierFound {
		// check license feature
		
		includeV3NetworkPolicy = true
	}

```

tmjd: I've asked about licenses with limited feature sets and it sounds like we've never given anything like that out to customers, so I think we can safely not worry about the edge case of a managed cluster without the EgressAccessControl feature.

pasanw: Got it - I'm good if we remove the extra logic. I think some of the structure nits I spoke about above could still apply regardless, but I'll leave it up to you to decide on the final structure

pasanw (approved):

rene-dekker (approved):
lgtm

PR #3610, opened 2024-11-22
Ensure CRDs are created on boot

caseydavenport:
## Description

Make sure CRDs exist early in bootup, as we require these CRDs to be present in order to launch controllers.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -104,31 +104,34 @@ func main() {
 	var sgSetup bool
 	var manageCRDs bool
 	var preDelete bool
+	var variant string
 
 	// bootstrapCRDs is a flag that can be used to install the CRDs and exit. This is useful for
 	// workflows that use an init container to install CustomResources prior to the operator starting.
 	var bootstrapCRDs bool
 
-	flag.BoolVar(&enableLeaderElection, "enable-leader-election", true,
-		"Enable leader election for controller manager. "+
-			"Enabling this will ensure there is only one active controller manager.")
-	flag.StringVar(&urlOnlyKubeconfig, "url-only-kubeconfig", "",
-		"Path to a kubeconfig, but only for the apiserver url.")
-	flag.BoolVar(&showVersion, "version", false,
-		"Show version information")
-	flag.StringVar(&printImages, "print-images", "",
-		"Print the default images the operator could deploy and exit. Possible values: list")
-	flag.StringVar(&printCalicoCRDs, "print-calico-crds", "",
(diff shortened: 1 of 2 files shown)


cmd/main.go
```diff
@@ -345,6 +345,15 @@ func main() {
 	}
 	setupLog.WithValues("provider", provider).Info("Checking type of cluster")
 
+	// If configured to managed CRDs, do a preliminary install of them here. The Installation controller
```
coutinhop, line 348:
```suggestion
	// If configured to manage CRDs, do a preliminary install of them here. The Installation controller
```

cmd/main.go
```diff
@@ -269,6 +274,20 @@ func main() {
 		os.Exit(1)
 	}
 
+	// If configured to managed CRDs, do a preliminary install of them here. The Installation controller
```
coutinhop, line 277: typo: managed -> manage?

coutinhop (approved):

cmd/main.go
```diff
@@ -269,6 +274,20 @@ func main() {
...
+	// will reconcile them as well, but we need to make sure they are installed before we start the rest of the controllers.
+	if bootstrapCRDs || manageCRDs {
```
rene-dekker, line 279: If I understand it correctly both  bootstrapCRDs and manageCRDs will install the OSS CRDs on startup. When manageCRDs is specified, the core controller will also apply the CRDs based on the variant.
What I don't fully understand at this point is that if my memory serves me right, the controller runtime requires a full restart in order to use new CRDs. In the case of enterprise, how can we use these flags to allow us to stop applying manifests manually?

caseydavenport: > What I don't fully understand at this point is that if my memory serves me right, the controller runtime requires a full restart in order to use new CRDs.

I'll try this out.. I didn't notice this problem when testing my changes with OSS Calico.

caseydavenport: I sort of assume we would have handled that already with the existing `manageCRDs` logic though.

rene-dekker: I checked and didn't see a restart in the core controller.

rene-dekker: There is this logic later in the reconcile loop that may solve the issue I brought up: https://github.com/tigera/operator/blob/master/pkg/controller/installation/core_controller.go#L968

rene-dekker (approved):
lgtm

PR #3618, opened 2024-11-26
Adding AzurePolicyMode to Installation

vikastigera:
Changes done to add Azure config to installation spec which has option to provide PolicyMode. This helps clients to have control over "control-plane" labels for namespaces.

## Description
https://tigera.atlassian.net/browse/EV-5377

For context in design doc : https://docs.google.com/document/d/1sttHDBs18_JoTCd-f2snZBn5kCWwqIst1mCseWu1hec/edit?usp=sharing

## For PR author

- [ ✓] Tests for change.
- [ ✓] If changing pkg/apis/, run `make gen-files`

## For PR reviewers

- [ ] Appropriate labels:


api/v1/installation_types.go
```diff
@@ -198,8 +198,30 @@ type InstallationSpec struct {
...
+
+type Azure struct {
+	// PolicyMode determines whether the "control-plane" label is applied to namespaces. It offers two options: Default and Manual.
+	// The Default option adds the "control-plane" label to namespaces that have the PodSecurityStandard set to privileged.
+	// The Manual option does not apply the "control-plane" label to any namespace.
+	// Default: Default
+	// +optional
+	// +kubebuilder:validation:Enum=Default;Manual
+	// +kubebuilder:default:=Default
+	PolicyMode *PolicyMode `json:"policyMode,omitempty"`
```
rene-dekker, line 215: Just wondering why this field is a pointer exactly. The validation and the controller will fill this field if the struct isn't nil. I think it is never possible to be nil in practice.

caseydavenport: Any optional field in the API should be a pointer - it is possible to send a yaml like this:

```
spec:
  azure: {}
```

In which case it will be `nil` in the API server and defaulted using the kubebuilder tag.

This is important for the forward looking chance that we add other fields to this sub-struct as well, where someone may want to set another field in the Azure section and our code expects this to be set.

rene-dekker: > may want to set another field in the Azure section and our code expects this to be set.

If we introduce field azure.foo and you apply `azure: {foo: bar}`, kube builder, the policyMode field will be set once our controllers are retrieving the CR.

Other than that I guess my question is answered; we should keep the pointer.

api/v1/installation_types.go
```diff
@@ -198,8 +198,30 @@ type InstallationSpec struct {
...
 	ServiceCIDRs []string `json:"serviceCIDRs,omitempty"`
+
+	// Azure is used to configure azure provider specific options.
+	// +optional
+	Azure *Azure `json:"azure,omitempty"`
+}
+
+type Azure struct {
+	// PolicyMode determines whether the "control-plane" label is applied to namespaces. It offers two options: Default and Manual.
+	// The Default option adds the "control-plane" label to namespaces that have the PodSecurityStandard set to privileged.
```
tmjd, line 209: Why are we making this dependent on the PodSecurityStandard too? Did I miss something in the design ore previous discussions that PodSecurityStandard was going to be a condition on setting the label?

pkg/render/namespaces.go
```diff
@@ -108,7 +108,16 @@ func CreateNamespace(name string, provider operatorv1.Provider, pss PodSecurityS
...
-		ns.Labels["control-plane"] = "true"
+		if applyAzurePolicy(azure, pss) {
+			ns.Labels["control-plane"] = "true"
+		}
 	}
 	return ns
 }
+
+func applyAzurePolicy(azure *operatorv1.Azure, pss PodSecurityStandard) bool {
+	if azure == nil || azure.PolicyMode == nil || *azure.PolicyMode == operatorv1.Default {
```
tmjd, line 119: I think we should only possibly return true if the PolicyMode is Default.

api/v1/installation_types.go
caseydavenport: Yeah, these shouldn't be directly tied to each other. We just brought that up as another similar configuration.

vikastigera: Are you suggesting to add label to all the namespaces in Default option ?
In the meeting we decided to add it only to the privileged namespaces

pkg/render/namespaces.go
vikastigera: Is it related to the previous comment, return "true" instead of  "PSSPrivileged == pss" or not to return true for nil cases ?

api/v1/installation_types.go
caseydavenport: Nope, not all namespaces. 

We should only add it to namespaces that need this policy configured - that may or may not be the same set of namespaces as those that require PSSPrivileged, but the point is that doesn't mean hard equality between `control-plane` and `PSSPrivileged`, and it definitely isn't something the user should know about / be able to rely on.

vikastigera: Got it, I'll update the documentation.

pkg/render/namespaces.go
vikastigera: I believe not having this config set (nil cases) should apply default behavior.

caseydavenport: It should, but it should also always default `nil` -> `Default`

rene-dekker: I think the tests in the core_controller and the renderer reflect the desired outcomes for each permutation and the tests pass.
So I think it is ok to proceed with merging this PR.

rene-dekker (approved):
lgtm

PR #3626, opened 2024-12-03
UT to check for unhandled override fields

nelljerram:
Follow-up to #3613; in particular please see https://github.com/tigera/operator/pull/3613#pullrequestreview-2473623372

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/common/components/components.go
+++ b/pkg/render/common/components/components.go
@@ -17,6 +17,7 @@ package components
 import (
 	"fmt"
 	"reflect"
+	"strings"
 
 	kbv1 "github.com/elastic/cloud-on-k8s/v2/pkg/apis/kibana/v1"
 	monitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
@@ -164,8 +165,6 @@ func GetPriorityClassName(overrides any) string {
 	return ""
 }
 
-var getFieldHookUT = func(fieldNames []string) {}
-
 func getField(overrides any, fieldNames ...string) (value reflect.Value) {
 	// SPECIAL CASE: ComplianceReporterPodTemplate doesn't follow the Spec, Template, Spec, ...
 	// pattern that all our other override structures follow.  Instead it skips the top-level
@@ -176,9 +175,9 @@ func getField(overrides any, fieldNames ...string) (value reflect.Value) {
 		}
 	}
 
-	// Call a hook that UT can use to check for any fields that are (accidentally) present in a
-	// customization structure, but which we don't handle.
-	getFieldHookUT(fieldNames)
+	// Record that we're handling `fieldNames`.  See `overrideFieldsHandledInLastApplyCall` for
+	// why.
+	recordHandledField(fieldNames)
 
 	typ := reflect.TypeOf(overrides)
 	for _, fieldName := range fieldNames {
@@ -206,6 +205,8 @@ func getField(overrides any, fieldNames ...string) (value reflect.Value) {
 
 // applyReplicatedPodResourceOverrides takes the given replicated pod resource data and applies the overrides.
 func applyReplicatedPodResourceOverrides(r *replicatedPodResource, overrides any) *replicatedPodResource {
+	resetHandledFields()
+
 	// If `overrides` has a Metadata field, and it's non-nil, non-clashing labels and annotations from that
 	// metadata are added into `r.labels` and `r.annotations`.
 	if metadata := GetMetadata(overrides); metadata != nil {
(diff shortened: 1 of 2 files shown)


pkg/render/common/components/components_test.go
```diff
@@ -93,6 +96,111 @@ var _ = Describe("Common components render tests", func() {
...
+		Entry("ManagerDeployment", &v1.ManagerDeployment{}, false),
+		Entry("PacketCaptureAPIDeployment", &v1.PacketCaptureAPIDeployment{}, false),
+		Entry("PolicyRecommendationDeployment", &v1.PolicyRecommendationDeployment{}, false),
+		Entry("TyphaDeployment", &v1.TyphaDeployment{}, false),
+
+		// This last entry checks that the code above really does identify when a
+		// structure has unhandled fields.  To do this we can use any available structure
+		// with some field names that don't overlap with those that are handled by our
+		// override code.
+		Entry("Installation", &v1.Installation{}, true),
```
Brian-McM, line 201: Is this a test that we need to keep, i.e. a test that the test is doing what it should? 

I would personally be fine with you testing this once manually to ensure it. If that's not quite enough for you, then would you be willing to pull the findUnhandled function out of the test and put it at a higher level and write a separate block for this? We'll never want expectUnhandled to be true for any case other than testing the test function.

nelljerram: I'm ok with moving `findUnhandled` to a higher level - although even that is a touch fiddly because it needs changes to how `unhandledFields` is collected.  I wouldn't like to remove the Installation case or make separate blocks for the positive and negative tests - either of those facilitates someone accidentally breaking the test and then not realizing.

pkg/render/common/components/components.go
```diff
@@ -158,9 +158,14 @@ func GetPriorityClassName(overrides any) string {
...
 		return ""
 	}
-	return value.String()
+	if value.Kind() == reflect.String {
+		return value.Interface().(string)
+	}
+	return ""
 }
 
+var getFieldHookUT = func(fieldNames []string) {}
```
Brian-McM, line 167: I know this doesn't affect the actual functionality of the code, but this sort of global variable / function setting, even just for tests, can lead to unexpected behaviour. This function will still be set when a test that sets it ends, and it could block (in the case that channels were used) or panic depending on how it's defined and used outside of the intended scope.

nelljerram: I suppose I could just hardcode the handled fields in components_test.go, instead of "discovering" them dynamically in this way.  WDYT?

Brian-McM: I suppose we could, but that sort leaves a different problem of the tests not aligning with code changes implicitly (which I guess would be worse?).

pkg/render/common/components/components_test.go
Brian-McM: Would you ever need more than one test for the negative case? Isn't this really just a test for the `findUnhandledFunction` to ensure it's working correctly, as opposed to testing that the overrides only have the supported fields?

If this is the case, it seems like this is merging two somewhat disjointed test cases into one.

pkg/render/common/components/components.go
```diff
@@ -294,6 +302,20 @@ func applyReplicatedPodResourceOverrides(r *replicatedPodResource, overrides any
 	return r
 }
 
+// For UT purposes, only, this variable stores the override fields that were handled in that most
+// recent `applyReplicatedPodResourceOverrides` call.  UT code then checks that the structures we
+// use for `overrides` do not have any _other_ fields than those (except for known special cases).
+var overrideFieldsHandledInLastApplyCall []string
```
Brian-McM, line 308: I'm fine for leaving this, but one problem with this approach is that if we do change the tests to be the component_test package then this becomes a little more difficult to test.

For the most part I like using the _test package because it forces the uts to only interact with the interface that is accessible outside of the package and therefore test the same code paths that are used by the users of the package.

The operator project doesn't really do this in general though, and I think addressing this change would be the least of the issues we'd have to deal with if we were to start doing this.

Ultimately, seems like there's no awesome way to deal with this without possibly changing the interfaces we're testing, and probably not worth it to spend any more time iterating on this.

Brian-McM (approved):

pkg/render/common/components/components.go
nelljerram: Thanks, Brian, I agree in principle with the concerns here.  But as you say, I don't currently see any better solution.  I could add an optional reporter arg to `applyReplicatedPodResourceOverrides`, but then I'd also need to pass that to all of the methods like `GetMetadata` - which feels like a bit too invasive, e.g. there would be an inscrutable `, nil` added to all of the calls to those methods from the validation code.  Or we could make `applyReplicatedPodResourceOverrides` etc. methods of some override management component - but that would be a bigger change.  I think what we have is a decent balance for now.

pkg/render/common/components/components_test.go
nelljerram: Maybe I'm not following your point here, but for me it's essential for the positive and negative cases to be using the same code.  I think I'd need to see a precise proposed code change to understand precisely.  Anyway I'll merge this PR now, but happy to discuss possible improvements further if you feel like proposing those.

PR #3631, opened 2024-12-04
Disable the ValidatingAdmissionPolicy API on k8s v1.28 and older

rene-dekker:
The Validating Admission Policy went GA with k8s 1.30 and is now enabled by default. Because of this, the apiserver project will automatically establish watches for this feature gate and so does our aggregated ApiServer.

However, for k8s 1.28 and older establishing these watches will result in crash-loops.

### DIFF
--- a/pkg/render/apiserver.go
+++ b/pkg/render/apiserver.go
@@ -21,7 +21,6 @@ import (
 
 	v3 "github.com/tigera/api/pkg/apis/projectcalico/v3"
 	"github.com/tigera/api/pkg/lib/numorstring"
-	"github.com/tigera/operator/pkg/render/common/authentication"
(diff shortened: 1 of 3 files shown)


coutinhop (approved):

pkg/render/apiserver.go
```diff
@@ -33,6 +32,8 @@ import (
 	apiregv1 "k8s.io/kube-aggregator/pkg/apis/apiregistration/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
+	"github.com/tigera/operator/pkg/render/common/authentication"
+
```
tmjd, line 36: Please remove this newline so this import can be groups with the other tigera/operator imports.

Brian-McM (approved):

PR #3636, opened 2024-12-06
Expand proxy detection support to Dex and support proxy configuration

pasanw:
## Description
This PR expands the support for proxy detection to Dex, and adds support for users to specify a proxy that the Operator will set on Guardian and Dex containers.

The proxy detection logic will always be used to render policy, in case the user has used both the proxy configuration and a mutating admission webhook to set the proxy environment variables.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


pkg/controller/authentication/authentication_controller.go
```diff
@@ -306,6 +311,78 @@ func (r *ReconcileAuthentication) Reconcile(ctx context.Context, request reconci
...
+
+					podProxies = append(podProxies, podProxyConfig)
+				}
+			}
+		}
+
+		r.resolvedPodProxies = podProxies
+	}
+	r.lastAvailabilityTransition = currentAvailabilityTransition
+
```
pasanw, line 385: For context, this logic is a mirror of what is found in the cluster connection controller

pkg/render/dex.go
```diff
@@ -398,25 +408,10 @@ func (c *dexComponent) allowTigeraNetworkPolicy(installationVariant operatorv1.P
...
-		{
-			Action:   v3.Allow,
-			Protocol: &networkpolicy.TCPProtocol,
-			Destination: v3.EntityRule{
-				Nets:  []string{"::/0"},
-				Ports: networkpolicy.Ports(443, 6443, 389, 636),
-			},
-		},
 	}...)
+	for _, egressRule := range c.resolveEgressRulesByDestination() {
```
rene-dekker, line 412: I know that policy rules are already tested in the controller test code as well, what I am missing is a test showing a practical example of proxy settings being set in the Installation Spec and the resulting network policy, so the two can be compared side-by-side to see if it the resulting policy makes sense. I feel that would help maintaining the code if we make any changes in the future.

rene-dekker:
Two questions around validation:

- Since it is our expectation that the user sets no_proxy for in-cluster destinations, do we consider it a misconfiguration if they don't? In other words, should we have validation for it?
- It looks like the env variables on the pod are set based on the proxy settings in the Installation spec and that the egress rules are based on the actual env variables that are on the pod. That's very clever, that means that whether the variables are set by the installation CR or by a mutating webhook, you end up allowing the traffic. I was wondering if we wanted to have any validation in the case that the user uses both methods.

api/v1/installation_types.go
```diff
@@ -202,6 +202,12 @@ type InstallationSpec struct {
 	// Azure is used to configure azure provider specific options.
 	// +optional
 	Azure *Azure `json:"azure,omitempty"`
+
+	// Proxy is used to configure the HTTP(S) proxy settings that will be applied to Tigera containers that connect
+	// to destinations outside the cluster. It is expected that NO_PROXY is configured such that destinations within
+	// the cluster (including the API server) are exempt from proxying.
+	// +optional
+	Proxy *Proxy `json:"proxy,omitempty"`
```
rene-dekker, line 210: Two questions around validation:

- Since it is our expectation that the user sets no_proxy for in-cluster destinations, do we consider it a misconfiguration if they don't? In other words, should we have validation for it?
- It looks like the env variables on the pod are set based on the proxy settings in the Installation spec and that the egress rules are based on the actual env variables that are on the pod. That's very clever, that means that whether the variables are set by the installation CR or by a mutating webhook, you end up allowing the traffic. I was wondering if we wanted to have any validation in the case that the user uses both methods.

rene-dekker (approved):
lgtm

api/v1/installation_types.go
pasanw: Discussed on a call: we decided that it's best to not get in the way of what the user configures, it's more likely that we'll cause issues than do anything meaningful to prevent or mitigate them

api/v1/installation_types.go
```diff
@@ -1048,3 +1054,68 @@ type WindowsNodeSpec struct {
...
+type Proxy struct {
+	// HTTPProxy defines the value of the HTTP_PROXY environment variable that will be set on Tigera containers that connect to
+	// destinations outside the cluster.
+	// +optional
+	HTTPProxy string `json:"httpProxy,omitempty"`
+
+	// HTTPSProxy defines the value of the HTTPS_PROXY environment variable that will be set on Tigera containers that connect to
+	// destinations outside the cluster.
+	// +optional
+	HTTPSProxy string `json:"httpsProxy,omitempty"`
```
caseydavenport, line 1067: Not to throw a wrench into things, but since these vars can take a separated list of items should we make them lists in our API?

```
httpsProxy:
- foo.com
- bar.org
```

and combine them? Or nah?

pasanw: `httpProxy` and `httpsProxy` should always be a single value, not a list AFAIK. `noProxy` can be a list of exempted destinations, but it's expected that this is a comma-separated list. If we ask for an array, it will likely be an inconvenience for the user IMO - they've likely been passing a comma-separated list elsewhere

caseydavenport: ah right, it was noProxy I was thinking about. I'm fine with the stance that it should just be the same as the env var.

PR #3638, opened 2024-12-09
Add Gateway API and Envoy Gateway install to Calico Enterprise

nelljerram:
## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


api/v1/gatewayapi_types.go
```diff
@@ -0,0 +1,129 @@
...
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []GatewayAPI `json:"items"`
+}
+
+func init() {
+	SchemeBuilder.Register(&GatewayAPI{}, &GatewayAPIList{})
+}
+
+// Optional customization of the gateway controller deployment.
```
rene-dekker, line 51: nit: While the documentation in this file is well written, it violates the linter in Goland and https://go.dev/doc/comment for not starting the comments of exported variables/funcs with the name of the exported variables/funcs.

pkg/controller/gatewayapi/gatewayapi_controller.go
```diff
@@ -0,0 +1,182 @@
...
+const ResourceName string = "gatewayapi"
+
+var log = logf.Log.WithName("controller_gatewayapi")
+
+// Add creates a new GatewayAPI Controller and adds it to the Manager. The Manager will set fields on the Controller
+// and Start it when the Manager is Started.
+//
+// Start Watches within the Add function for any resources that this controller creates or monitors. This will trigger
+// calls to Reconcile() when an instance of one of the watched resources is modified.
+func Add(mgr manager.Manager, opts options.AddOptions) error {
```
rene-dekker, line 49: If enterprise CRDs do not exist, this should probably exit, otherwise it will try to establish watches for unregistered kinds and crash.

nelljerram: Thanks, I added that gate.

api/v1/gatewayapi_types.go
nelljerram: I'm not sure about this.  It seems to me like an unnatural restriction.  Can you configure Goland not to expect this?

Brian-McM: We certainly could but goland is following golang comment standards where every exported function comment should start with the function name. I think this has to do with generating go docs.

nelljerram: Can you provide a reference?

Brian-McM: [This](https://go.dev/blog/godoc) blog post states the convention:

```
Notice this comment is a complete sentence that begins with the name of the element it describes. This important convention allows us to generate documentation in a variety of formats, from plain text to HTML to UNIX man pages, and makes it read better when tools truncate it for brevity, such as when they extract the first line or sentence.
```

Looking at the updated documentation though [here](https://go.dev/doc/comment) that the older blog post references I don't see any mention of that, however, all the examples of the exported functions use this convention.

Also to clarify, this isn't goland exactly, this is the golang linter that's catching these (which goland itself uses).

I think there's lots of posts about this if you google it, maybe there's a reason why it's not needed anymore but I find comments supporting that it helps doc generation.

Personally I think it would be best to just stick with this convention unless there's some compelling reason not to, I had thought that this was already an adopted convention at tigera.

pkg/render/gateway_api.go
```diff
@@ -0,0 +1,526 @@
...
+const (
+	GatewayAPIName                      = "gateway-api"
+	GatewayControllerLabel              = GatewayAPIName + "-controller"
+	EnvoyGatewayConfigName              = "envoy-gateway-config"
+	EnvoyGatewayConfigKey               = "envoy-gateway.yaml"
+	EnvoyGatewayDeploymentContainerName = "envoy-gateway"
+	EnvoyGatewayJobContainerName        = "envoy-gateway-certgen"
+)
+
+func GatewayAPIResourcesGetter() func() *gatewayAPIResources {
```
Brian-McM, line 82: This doesn't seem like something we need to have happen at run time since it will never change but it's redone multiple times when reconciliation happens. This seems to provide an unneeded source of panics that could be caught at build time.

Instead, could we just generate the structures here from the yaml at build time and save them to a file, similar to how we generate the version structures https://github.com/tigera/operator/tree/master/hack/gen-versions using go template?

api/v1/gatewayapi_types.go
```diff
@@ -0,0 +1,296 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// GatewayAPISpec has fields that can be used to customize our GatewayAPI support.
+type GatewayAPISpec struct {
+	// Allow optional customization of the gateway controller deployment.
+	GatewayControllerDeployment *GatewayControllerDeployment `json:"gatewayControllerDeployment,omitempty"`
+
+	// Allow optional customization of the gateway certgen job.
+	GatewayCertgenJob *GatewayCertgenJob `json:"gatewayCertgenJob,omitempty"`
```
rene-dekker, line 31: I think we generally write out api field names. I am assuming Certgen stands for Certificate Generation, meaning it is two words and both words should be capitalized in the field name as well.

pkg/render/gateway_api.go
```diff
@@ -0,0 +1,507 @@
...
+		objs[i] = objs[i].DeepCopyObject().(client.Object)
+	}
+
+	// Substitute possibly modified image names into the envoy-gateway-config.
+	envoyGatewayConfig := resources.envoyGatewayConfig.DeepCopyObject().(*envoyapi.EnvoyGateway)
+	envoyGatewayConfig.Provider.Kubernetes.RateLimitDeployment.Container.Image = &pr.envoyRatelimitImage
+	envoyGatewayConfig.Provider.Kubernetes.ShutdownManager.Image = &pr.envoyGatewayImage
+
+	// Add pull secrets into the envoy-gateway-config, insofar as the API allows this.  (In
+	// other words, currently only for the ratelimit image.  So effectively the gateway image
```
rene-dekker, line 423: Is this comment correct? It looks like you are able to set the pull secret on the deployment  in line 447.

pkg/render/gateway_api.go
```diff
@@ -0,0 +1,507 @@
...
+	certgenJob.Spec.Template.Spec.ImagePullSecrets = append(
+		certgenJob.Spec.Template.Spec.ImagePullSecrets,
+		secret.GetReferenceList(pr.cfg.PullSecrets)...)
+
+	// Apply customizations from the GatewayCertgenJob field of the GatewayAPI CR.
+	rcomp.ApplyJobOverrides(certgenJob, pr.cfg.GatewayAPI.Spec.GatewayCertgenJob)
+
+	objs = append(objs, certgenJob)
+
+	objs = append(objs, secret.ToRuntimeObjects(secret.CopyToNamespace(resources.namespace.Name, pr.cfg.PullSecrets...)...)...)
```
rene-dekker, line 475: I was wondering if it would be better to render the pull secrets before you render the deployments to avoid k8s v1/Events, like `FailedToRetrieveImagePullSecret`.

pkg/controller/gatewayapi/gatewayapi_controller.go
```diff
@@ -0,0 +1,187 @@
...
+	operatorv1 "github.com/tigera/operator/api/v1"
+	"github.com/tigera/operator/pkg/controller/options"
+	"github.com/tigera/operator/pkg/controller/status"
+	"github.com/tigera/operator/pkg/controller/utils"
+	"github.com/tigera/operator/pkg/controller/utils/imageset"
+	"github.com/tigera/operator/pkg/ctrlruntime"
+	"github.com/tigera/operator/pkg/render"
+)
+
+const ResourceName string = "gatewayapi"
```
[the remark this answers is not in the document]
nelljerram: Thanks, removed.

pkg/controller/gatewayapi/gatewayapi_controller.go
```diff
@@ -0,0 +1,187 @@
...
+	}
+
+	pullSecrets, err := utils.GetNetworkingPullSecrets(installation, r.client)
+	if err != nil {
+		r.status.SetDegraded(operatorv1.ResourceReadError, "Error retrieving pull secrets", err, reqLogger)
+		return reconcile.Result{}, err
+	}
+
+	// Get the GatewayAPI CR.
+	gatewayAPI := &operatorv1.GatewayAPI{}
```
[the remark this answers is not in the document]
nelljerram: Makes sense, thanks!

pkg/controller/gatewayapi/gatewayapi_controller.go
```diff
@@ -0,0 +1,187 @@
...
+		r.status.SetDegraded(operatorv1.ResourceReadError, "Error querying for GatewayAPI CR", err, reqLogger)
+		return reconcile.Result{}, err
+	}
+	r.status.OnCRFound()
+
+	// Render CRDs.  For these we specify nil for the owning CR - i.e. no ownership - so that
+	// the CRDs are left in place even if the GatewayAPI CR is removed again.  This is in case
+	// the customer uses a second (or more) implementation of the Gateway API in addition to the
+	// one that we are providing here.
+	crdComponent := render.NewPassthrough(render.GatewayAPICRDs(log)...)
```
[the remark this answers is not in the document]
nelljerram: I've invented a concept of a component handler having "create only" operation.  PTAL https://github.com/tigera/operator/pull/3638/commits/ecb23158ff65eb40cd0c5bb682ed1a2708ebbe03

hack/gen-versions/enterprise.go.tpl
```diff
@@ -357,6 +357,25 @@ var (
 		Registry: "{{ .Registry }}",
 	}
 {{- end }}
+
+	ComponentGatewayAPIEnvoyGateway = Component{
```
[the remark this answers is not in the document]
nelljerram: Thanks, done; good catch.

pkg/render/gateway_api_resources.yaml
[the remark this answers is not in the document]
nelljerram: Nice, thanks.

pkg/render/gateway_api.go
nelljerram: Thanks.  I wasn't sure if I could fully rely on the object ordering, and hence if the ordering really mattered.  But it makes sense that it would do.

pkg/render/gateway_api.go
nelljerram: Ah yes, you are right.  I thought the envoy coding and config was still missing a way to set pull secrets for the "ShutdownManager" image - because the config only allows setting an image - but in fact pull secrets are set at the pod level and so the handling at
```
	envoyGatewayConfig.Provider.Kubernetes.RateLimitDeployment.Pod.ImagePullSecrets = secret.GetReferenceList(pr.cfg.PullSecrets)
```
actually covers the "ShutdownManager" image as well as the "RateLimit" image.

pkg/controller/gatewayapi/gatewayapi_controller.go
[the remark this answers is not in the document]
Brian-McM: How is this going to work when we've updated our CRDs and need to apply the changes?

I think we need to instead know if we're managing the installation or the customer is doing it (as @rene-dekker points allude to).

@rene-dekker, how are we doing this for prometheus and eck? I think prometheus requires a different insallation step, but I can't remember what we do for eck.

pkg/render/gateway_api.go
nelljerram: I think it's already doing what you think it should do.  In other words, it only reads and parses the YAML once. The only reason for the slightly odd (to you) form is to have the `resources` cache and associated lock internal to the function, instead of being globals that other code might accidentally misuse.

pkg/controller/gatewayapi/gatewayapi_controller.go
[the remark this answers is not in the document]
rene-dekker: We discussed this offline and decided to follow up with the right approach for CRDs.

rene-dekker (approved):
lgtm

PR #3643, opened 2024-12-10
Add waf default core ruleset as configmap

LorcanMcVeigh:
## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/applicationlayer/applicationlayer_controller.go
+++ b/pkg/controller/applicationlayer/applicationlayer_controller.go
@@ -45,6 +45,8 @@ import (
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	coreruleset "github.com/corazawaf/coraza-coreruleset/v4"
 )
 
 const ResourceName = "applicationlayer"
@@ -124,6 +126,14 @@ func add(mgr manager.Manager, c ctrlruntime.Controller) error {
 		)
 	}
 
+	err = utils.AddConfigMapWatch(c, applicationlayer.DefaultCoreRuleset, common.OperatorNamespace(), &handler.EnqueueRequestForObject{})
+	if err != nil {
+		return fmt.Errorf(
+			"applicationlayer-controller failed to watch ConfigMap %s: %v",
+			applicationlayer.DefaultCoreRuleset, err,
+		)
+	}
+
 	// Watch mutatingwebhookconfiguration responsible for sidecar injetion
 	err = c.WatchObject(&admregv1.MutatingWebhookConfiguration{ObjectMeta: metav1.ObjectMeta{Name: common.SidecarMutatingWebhookConfigName}},
 		&handler.EnqueueRequestForObject{})
@@ -135,6 +145,7 @@ func add(mgr manager.Manager, c ctrlruntime.Controller) error {
 	maps := []string{
 		applicationlayer.EnvoyConfigMapName,
 		applicationlayer.ModSecurityRulesetConfigMapName,
+		applicationlayer.DefaultCoreRuleset,
 	}
 	for _, configMapName := range maps {
 		if err = utils.AddConfigMapWatch(c, configMapName, common.CalicoNamespace, &handler.EnqueueRequestForObject{}); err != nil {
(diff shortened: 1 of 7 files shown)


pkg/controller/applicationlayer/applicationlayer_controller.go
```diff
@@ -450,6 +467,30 @@ func getDefaultCoreRuleset(ctx context.Context) (*corev1.ConfigMap, error) {
...
+		return nil, err
+	}
+	return ruleset, nil
+}
+
+func getOSWAPCoreRuleSet(ctx context.Context) (*corev1.ConfigMap, error) {
+	ruleset, err := embed.AsConfigMap(
+		applicationlayer.DefaultCoreRuleset,
+		common.OperatorNamespace(),
+		coreruleset.FS,
```
gantony, line 493: The naming is confusing here (`getOSWAPCoreRuleSet()` and `getCorazaRuleSet()`). `coreruleset.FS` refers to https://github.com/corazawaf/coraza-coreruleset/tree/main/rules, which I would refer to as the `coraza-ruleset`. Within this `coraza-ruleset`, there is the `@owasp_crs` folder which contains the `coreruleset` taken from https://github.com/coreruleset/coreruleset/tree/main/rules. 

I think the `default-coreruleset` config-map should contain the content of `@owasp_crs`. The rest of `coraza-ruleset` (as I call it :sweat_smile:) is basically what we have in `modsecurity-ruleset` and `tigera.conf`...

pkg/render/applicationlayer/applicationlayer.go
```diff
@@ -56,7 +56,10 @@ const (
 	DikastesContainerName            = "dikastes"
 	ModSecurityRulesetVolumeName     = "modsecurity-ruleset"
 	ModSecurityRulesetVolumePath     = "/etc/modsecurity-ruleset"
+	DefaultCoreRulesetVolumeName     = "coreruleset-default"
+	DefaultCoreRulesetVolumePath     = "/etc/coreruleset-default"
```
gantony, line 60: Please change to `DefaultCoreRulesetVolumePath = "/etc/modsecurity-ruleset/coreruleset-default"` (adapt if you rename `modsecurity-ruleset`)

pkg/render/applicationlayer/applicationlayer.go
```diff
@@ -330,6 +335,11 @@ func (c *component) containers() []corev1.Container {
 						MountPath: ModSecurityRulesetVolumePath,
 						ReadOnly:  true,
 					},
+					{
```
gantony, line 338: The order matters a lot here. Maybe a comment would be good or a test about the order. Comment in the test maybe?

gantony (changes requested):
Beside the comment, need to update `tigera.conf` too to make this work together:
replace `Include @owasp_crs/*.conf` with `Include /etc/modsecurity-ruleset/coreruleset-default/*.conf`.

pkg/render/applicationlayer/embed/embed.go
```diff
@@ -37,7 +37,7 @@ func init() {
 	}
 }
 
-func AsMap() (map[string]string, error) {
+func AsMap(fileSystem fs.FS) (map[string]string, error) {
```
electricjesus, line 40: I think you can just add corerulesetFS as this package's global variable (like FS fs.FS above) so you don't need to have to specify this as a parameter.


You can also merge both FS'es together e.g.:

```golang
import (
	"embed"
	"fmt"
	"io/fs"
	
	"github.com/jcchavezs/mergefs"
	
	coreruleset "github.com/corazawaf/coraza-coreruleset/v4"
	
	corev1 "k8s.io/api/core/v1"
)

var (
	FS fs.FS

        coreruleset.FS
        
	//go:embed coreruleset
	crsFS embed.FS
)

func init() {
	crsSubFS, err := fs.Sub(crsFS, "coreruleset")
	if err != nil {
		panic(err)
 	}
 	
 	FS = mergefs.Merge(coreruleset.FS, crsSubFS)
 }
```

doing it this way you can make new file entry assertions in embed_test.go where you expect a bunch of files that should appear on in the FS object if you walk it

pkg/render/applicationlayer/embed/embed.go
```diff
@@ -48,23 +48,27 @@ func AsMap() (map[string]string, error) {
 			return err
 		}
 
-		if b, err := fs.ReadFile(FS, path); err != nil {
+		if d.Name()[0] == '@' {
```
electricjesus, line 51: not sure why this is done? are we skipping any directories prefixed with '@'? I thought rules were in a directory named '@owasp-' something

electricjesus (commented):
overall looks good, just wanted to know about a few things first

pkg/render/applicationlayer/embed/embed.go
gantony: Yes that works by accident, need to make the logic clearer.

LorcanMcVeigh: "by accident" i think you mean "very skilfully"

pkg/render/applicationlayer/embed/embed.go
LorcanMcVeigh: im not sure I follow. The idea behind add the function parameter was so that I could generate a different configmap based on the coraza coreruleset as well as the existing configmap that contained `tigera.conf`. 
Would doing it this way not make it 1 configmap altogether?

electricjesus: Got on a zoom with Lorcan, and he helped me undestand this better -- resolved

pkg/controller/applicationlayer/applicationlayer_controller.go
electricjesus: there was an agreement to do this in a separate PR -- resolved conversation

pkg/render/applicationlayer/embed/embed.go
gantony: Changing this turned out to be easier than anticipated :smile: 
image

gantony:
Replaced by #3649.

PR #3652, opened 2024-12-16
expose logLevel for apiserver containers

ti-afra:
## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


api/v1/apiserver_types.go
```diff
@@ -86,6 +86,11 @@ type APIServerDeploymentContainer struct {
 	// If used in conjunction with the deprecated ComponentResources, then this value takes precedence.
 	// +optional
 	Resources *v1.ResourceRequirements `json:"resources,omitempty"`
+
+	// +kubebuilder:validation:Enum=fatal;error;warn;info;debug;trace
```
caseydavenport, line 90: These should be capitalized - e.g., `Fatal`, `Error`, etc. 

This just matches k8s API conventions for enum types

pkg/render/apiserver.go
```diff
@@ -1284,6 +1300,24 @@ func (c *apiServerComponent) queryServerContainer() corev1.Container {
...
+		for _, con := range containers {
+			if strings.Contains(con.Name, "queryserver") {
+				if logLevel := con.LogLevel; logLevel != nil {
+					env = append(env, corev1.EnvVar{Name: "LOGLEVEL", Value: *logLevel})
+				}
+			}
+		}
+	} else {
+		// set default LOG_LEVEL to info when not set by the user
+		env = append(env, corev1.EnvVar{Name: "LOGLEVEL", Value: "info"})
```
caseydavenport, line 1318: LOG_LEVEL?

ti-afra: in queryserver, it's checking for `LOGLEVEL` and in apiserver it's checking for `LOG_LEVEL` and I just found out about it. Since we already cut the release in calico-private, I thought I keep it consistent with the variables already defined in these two services and maybe change the one in queryserver master. WDYT?
Is the operator version to calico-private release version 1-1? or do we need to be backward compatible to some versions in the operator?

caseydavenport: A given operator always installs a specific version of Calico, so don't need to worry about that. But if they use different vars then there's nothing to do here in this PR - just kind of annoying :sweat_smile: 

Wanted to make sure it wasn't a typo

ti-afra: I should be updating the comment though!

caseydavenport (approved):

PR #3667, opened 2024-12-19
expose logLevel for apiserver containers (#3652)

ti-afra:
* expose logLevel in apiserver manifest for its containers

* fix unstable gomega match for large nested dex policies

gomega Expect was not able to compare correctly the dex policies and would fail when running in "Run" mode but pass when running in "Debug" mode.
It was also passing when we print the values before the Expect().To() call.

To fix this, we convert the larger and nested struct to String and compare the string values.

* update crd: securityeventwebhooks

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


rene-dekker (approved):
lgtm

PR #3630, opened 2024-12-03
Restrict tigera-operator secret access to namespace only

vara2504:
Restrict tigera-operator secret access to namespace only, retain get/list/watch cluster-wide 
https://tigera.atlassian.net/browse/EV-5410

Tested external prometheus ,
bz addons run external-prometheus:setup
bz addons run wireguard:setup
image

External-prometheus E2E  passed

## For PR author

- [ ] Tests for change.
- [ ] If changing pkg/apis/, run `make gen-files`
- [ ] If changing versions, run `make gen-versions`

## For PR reviewers

A note for code reviewers - all pull requests must have the following:

- [ ] Milestone set according to targeted release.
- [ ] Appropriate labels:
  - `kind/bug` if this is a bugfix.
  - `kind/enhancement` if this is a a new feature.
  - `enterprise` if this PR applies to Calico Enterprise only.

PR #3668, opened 2024-12-19
Allow operator to create secrets in tigera-gateway namespace

nelljerram:
(Per pattern of #3630)

Also fix a log typo


rene-dekker (approved):
lgtm

PR #3670, opened 2024-12-20
Include dex secrets  only when oidc is non tigera or nil

vara2504:
Include dex secrets  only when oidc is non tigera or nil 

```
vara@vara:~/bzprofiles/Clusters/tc_dex$ ts authentication -oyaml
kind: TigeraStatus
metadata:
  creationTimestamp: "2024-12-20T17:47:23Z"
  generation: 1
  managedFields:
    fieldsType: FieldsV1
    fieldsV1:
      f:spec: {}
    manager: operator
    operation: Update
    time: "2024-12-20T17:47:23Z"
    fieldsType: FieldsV1
    fieldsV1:
      f:status:
        .: {}
        f:conditions: {}
    manager: operator
    operation: Update
    subresource: status
    time: "2024-12-20T17:47:48Z"
  name: authentication
  resourceVersion: "19993"
  uid: 4af74605-7363-4a5e-bf9a-2cb6ffb37ccf
spec: {}
status:
  conditions:
  - lastTransitionTime: "2024-12-20T17:47:48Z"
    message: All objects available
    observedGeneration: 2
    reason: AllObjectsAvailable
    status: "True"
    type: Available
  - lastTransitionTime: "2024-12-20T17:47:28Z"
    message: All Objects Available
    observedGeneration: 2
    reason: AllObjectsAvailable
    status: "False"
    type: Progressing
  - lastTransitionTime: "2024-12-20T17:47:48Z"
    message: All Objects Available
    observedGeneration: 2
    reason: AllObjectsAvailable
    status: "False"
    type: Degraded

```

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/dex.go
+++ b/pkg/render/dex.go
@@ -138,10 +138,11 @@ func (c *dexComponent) Objects() ([]client.Object, []client.Object) {
 	// TODO the RequiredSecrets in the dex condig to not pass back secrets of this type.
 	if !c.cfg.DeleteDex {
 		objs = append(objs, secret.ToRuntimeObjects(c.cfg.DexConfig.RequiredSecrets(common.OperatorNamespace())...)...)
-	}
 
-	objs = append(objs, secret.ToRuntimeObjects(c.cfg.DexConfig.RequiredSecrets(DexNamespace)...)...)
-	objs = append(objs, secret.ToRuntimeObjects(secret.CopyToNamespace(DexNamespace, c.cfg.PullSecrets...)...)...)
+		// The Dex namespace exists only for non-Tigera OIDC types to create secrets within the namespace.
+		objs = append(objs, secret.ToRuntimeObjects(c.cfg.DexConfig.RequiredSecrets(DexNamespace)...)...)
+		objs = append(objs, secret.ToRuntimeObjects(secret.CopyToNamespace(DexNamespace, c.cfg.PullSecrets...)...)...)
+	}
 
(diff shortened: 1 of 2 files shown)


pkg/render/dex.go
```diff
@@ -138,11 +138,10 @@ func (c *dexComponent) Objects() ([]client.Object, []client.Object) {
 	// TODO the RequiredSecrets in the dex condig to not pass back secrets of this type.
 	if !c.cfg.DeleteDex {
```
vara2504, line 139: @rene-dekker DeleteDex is set to true only when (authentication.Spec.OIDC != nil && authentication.Spec.OIDC.Type == operatorv1.OIDCTypeTiger) [here](https://github.com/tigera/operator/blob/598cdf09604312d2916c917a5d7d6eef75c0c8a3/pkg/controller/utils/utils.go#L901), which is logically equivalent to !(authentication.Spec.OIDC == nil || authentication.Spec.OIDC.Type != operatorv1.OIDCTypeTiger). Therefore, it should be ok to included under the same if condition.

rene-dekker (approved):
The PR is imperfect in some cases that I believe are not worth solving:
- If someone first does regular OIDC, then OIDC type=tigera, it won't properly clean everything.

I think this is a use case that shouldn't/won't happen in the wild, since oidc type tigera should only be used in cloud. The worst case is that a user needs to delete some resources by kubectl if it affects them.

Therefore, LGTM.

PR #3675, opened 2024-12-30
Update the upstream versions of alertmanager and prometheus

rene-dekker:
image

Officially this operator version combined with this prometheus version is not supported, but it works fine.
As for alertmanager it matches the compatibility matrix.


vikastigera:
LGTM
Approving

vikastigera (approved):

PR #3676, opened 2025-01-01
[release-v1.36] Auto pick #3631: Disable the ValidatingAdmissionPolicy API on k8s v1.28 and

coutinhop:
Cherry pick of #3631 on release-v1.36.

#3631: Disable the ValidatingAdmissionPolicy API on k8s v1.28 and

# Original PR Body below

The Validating Admission Policy went GA with k8s 1.30 and is now enabled by default. Because of this, the apiserver project will automatically establish watches for this feature gate and so does our aggregated ApiServer.

However, for k8s 1.28 and older establishing these watches will result in crash-loops.


rene-dekker (approved):
lgtm

PR #3689, opened 2025-01-09
[EV-5553] Pick #3688 apiserver: move loglevel under spec

ti-afra:
## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


api/v1/apiserver_types.go
```diff
@@ -25,6 +25,16 @@ import (
 
 // APIServerSpec defines the desired state of Tigera API server.
 type APIServerSpec struct {
+	// +kubebuilder:validation:Enum=Fatal;Error;Warn;Info;Debug;Trace
+	// +kubebuilder:default=Info
+	// +optional
+	ApiServerLogLevel *string `json:"apiServerLogLevel,omitempty"`
+
+	// +kubebuilder:validation:Enum=Fatal;Error;Warn;Info;Debug;Trace
```
caseydavenport, line 33: Do we want to define a `type LogLevel string` that can carry these validations / defaults rather than replicate them in both places? Not super important but might be a bit cleaner.

tmjd: I had the same thought that we should have a LogLevel type.

rene-dekker (approved):

PR #3690, opened 2025-01-10
Override ECK mount path such that our plugin continues to work

rene-dekker:
We need to override this mount and change the mountPath. Otherwise, ECK will mount an emptyDir in the original location (/usr/share/kibana/plugins), which would remove our custom theme resulting in crash-looping pods, because our custom config is unexpected.


vara2504 (approved):
lgtm

PR #3691, opened 2025-01-10
Automated cherry pick of #3690: Override ECK mount path such that our plugin continues to

rene-dekker:
Cherry pick of #3690 on release-v1.37.

Original branch name ev-5445-2

#3690: Override ECK mount path such that our plugin continues to


vara2504 (approved):
lgtm

PR #3699, opened 2025-01-14
allow creating a new operator based off a previous operator

radTuti:
## Description

Allow building an operator image using a previous hashrelease/release as base and changing a few components.

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


hack/release-from/action.go
```diff
@@ -0,0 +1,263 @@
...
+	}
+
+	for _, file := range []string{calicoConfig, enterpriseConfig} {
+		// open file locally
+		localFile := fmt.Sprintf("%s/%s", repoRootDir, file)
+		out, err := os.OpenFile(localFile, os.O_CREATE|os.O_WRONLY|os.O_TRUNC, 0o644)
+		if err != nil {
+			return fmt.Errorf("Error opening local file %s: %s", file, err)
+		}
+		defer out.Close()
```
danudey, line 220: Suggestion: use more descriptive variable names so that it's easier to know at a glance what is being passed around. For example:

```suggestion
	for _, configFileName := range []string{calicoConfig, enterpriseConfig} {
		// open file locally
		localConfigFilePath := fmt.Sprintf("%s/%s", repoRootDir, configFileName)
		outputConfigFile, err := os.OpenFile(localConfigFilePath, os.O_CREATE|os.O_WRONLY|os.O_TRUNC, 0o644)
		if err != nil {
			return fmt.Errorf("Error opening local file %s: %s", configFilePath, err)
		}
		defer outputConfigFile.Close()
```

Clarifying variable names makes it a lot easier to tell at a glance what something is. In this case 'file' is a string, 'localFile' is a string, and 'out' is a file, meaning none of these variable identifiers are correct.

This is true in multiple other functions, but I just happened to call it out here.

hack/release-from/action.go
```diff
@@ -0,0 +1,263 @@
...
+	// overwrite local file with updated config
+	if err := os.WriteFile(localFile, []byte(fmt.Sprintf("%s\n", config)), 0o644); err != nil {
+		return fmt.Errorf("Error overwriting local file %s: %s", configFile, err)
+	}
+	return nil
+}
+
+// retrieveBaseVersionConfig gets the config based to use
+func retrieveBaseVersionConfig(baseVersion, repoRootDir string) error {
+	url, err := generateBaseConfigDownloadURL(baseVersion)
```
danudey, line 208: `url` should be something more descriptive, like `configDownloadURL`, for clarity later on.

hack/release-from/action.go
```diff
@@ -0,0 +1,263 @@
...
+		logrus.WithFields(logrus.Fields{
+			"file":         file,
+			"localPath":    localFile,
+			"downloadPath": url,
+		}).Debug("Overwrote local file with downloaded file")
+	}
+	return nil
+}
+
+// generateBaseConfigDownloadURL returns the url to download base configs from
```
danudey, line 245: Returns it from where? What does it generate it from? The comment doesn't really tell us more than the function name (and since the function name says 'generate' and not 'return', I think the function name tells us more).

hack/release-from/action.go
```diff
@@ -0,0 +1,263 @@
...
+		return fmt.Errorf("Error determining if version is a release: %s", err)
+	} else if release {
+		return newOperator(repoRootDir, version, c.String(remoteFlag.Name))
+	}
+
+	return newHashreleaseOperator(repoRootDir, version, c.StringSlice(archFlag.Name))
+}
+
+// isRelease checks if the version is a release version.
+// A release version is in the format vX.Y.Z.
```
danudey, line 81: Useful comment, but could probably be more succinct:

```suggestion
// isRelease checks if the version is in the format vX.Y.Z.
```

hack/release-from/action.go
```diff
@@ -0,0 +1,263 @@
...
+	Version  string `yaml:"version"`
+	Registry string `yaml:"registry"`
+}
+
+type productConfig struct {
+	Title      string               `yaml:"title"`
+	Components map[string]component `yaml:"components"`
+}
+
+func releaseFrom(ctx context.Context, c *cli.Command) error {
```
danudey, line 42: Missing a comment that should probably exist to provide context and scope for what this function does.

hack/release-from/action.go
```diff
@@ -0,0 +1,263 @@
...
+// A release version is in the format vX.Y.Z.
+func isRelease(version string) (bool, error) {
+	releaseRegex, err := regexp.Compile(releaseFormat)
+	if err != nil {
+		return false, fmt.Errorf("Error compiling release regex: %s", err)
+	}
+	return releaseRegex.MatchString(version), nil
+}
+
+// newOperator creates a new operator release.
```
danudey, line 90: Function comments should contain information about the intended scope of the function. Creates how? Defines? Defines and builds? Defines, builds, and publishes?

hack/release-from/action.go
```diff
@@ -0,0 +1,263 @@
...
+		fmt.Sprintf("%s-init:latest", imageName),
+		fmt.Sprintf("%s/%s-init:%s", quayRegistry, imageName, version),
+	}, env); err != nil {
+		return err
+	}
+	logrus.WithField("tag", initTag).Debug("Built init image")
+	return publishHashreleaseOperator(version, archs)
+}
+
+// publishHashreleaseOperator publishes the hashrelease operator to quay.
```
danudey, line 140: Does it make sense that we're hard-coding quay.io into this? It's obviously what we're actually doing at the moment, but being able to specify another registry might provide more flexibility down the road (e.g. for testing the release process, for customers building their own images, etc).

hack/release-from/action.go
```diff
@@ -0,0 +1,263 @@
...
+		}
+	}
+	enterpriseOverrides := c.StringSlice(exceptEnterpriseFlag.Name)
+	if len(enterpriseOverrides) > 0 {
+		if err := updateConfig(repoRootDir, enterpriseConfig, enterpriseOverrides); err != nil {
+			return fmt.Errorf("Error overriding calico config: %s", err)
+		}
+	}
+
+	// Either build a new release or a new hashrelease operator
```
danudey, line 68: I absolutely hate the english language for its ambiguity. I read this as:

Either:
1. Build a new release; or
2. A new hashrelease operator (will do something?)

Maybe for clarity rephrase this as "Build either a new release or a new hashrelease operator". Or don't, I doubt it matters.

hack/release-from/action.go
```diff
@@ -0,0 +1,263 @@
...
+	version := c.String(versionFlag.Name)
+	release, err := isRelease(version)
```
danudey, line 70: This code is really confusing to me.

We have a version (from `versionFlag.Name`). We pass it to `isRelease(version)`, which sounds like it should return a bool, but we call it `release`. So is it a release? Or a bool?

The `isRelease` function returns `releaseRegex.MatchString(version)`, but I don't know off the top of my head what that is. Is it a bool? It sounds like you're getting a `release` back, but I don't know what type that would be.

I would name this variable `versionIsRelease`, or name it `isRelease` and rename `isRelease()` to something else, like `checkIsRelease()`, for clarity.

hack/release-from/action.go
```diff
@@ -0,0 +1,263 @@
...
+	initTag := fmt.Sprintf("%s/%s-init:%s", quayRegistry, imageName, version)
+	if _, err := runCommand("docker", []string{
+		"tag",
+		fmt.Sprintf("%s-init:latest", imageName),
+		fmt.Sprintf("%s/%s-init:%s", quayRegistry, imageName, version),
+	}, env); err != nil {
+		return err
+	}
+	logrus.WithField("tag", initTag).Debug("Built init image")
+	return publishHashreleaseOperator(version, archs)
```
danudey, line 137: As above, this should probably be called from outside of this function to decouple building and publishing. Side effects are bad, but when the side effect is publishing an image to a public registry that's potentially even worse (especially since we don't seem to have any way of preventing clobbering existing images).

hack/release-from/action.go
```diff
@@ -0,0 +1,263 @@
...
+
+	initImage := fmt.Sprintf("%s/%s-init:%s", quayRegistry, imageName, version)
+	if _, err := runCommand("docker", []string{"push", initImage}, nil); err != nil {
+		return err
+	}
+	logrus.WithField("image", initImage).Debug("Pushed init image")
+	return nil
+}
+
+// updateConfig updates the version of image(s) in a config file
```
danudey, line 172: ...based on what information?

hack/release-from/action.go
```diff
@@ -0,0 +1,263 @@
...
+		return "", fmt.Errorf("Error determining if version is a release: %s", err)
+	}
+	if release {
+		return fmt.Sprintf("%s/refs/tags/%s", baseDownloadURL, baseVersion), nil
+	}
+	gitHashRegex, err := regexp.Compile(`^g([a-f0-9]{12})`)
+	if err != nil {
+		return "", fmt.Errorf("Error compiling git hash regex: %s", err)
+	}
+	matches := gitHashRegex.FindStringSubmatch(baseVersion)
```
danudey, line 258: I'm operating under the assumption that this code works, but I'm surprised that this code correctly finds what it's looking for given that the regex starts with `^`, which should mean "start of string". I would think this would only work if the string was only the `g1ac2b3a4` component of the version and not the entire version string.

hack/release-from/utils.go
```diff
@@ -0,0 +1,86 @@
...
+func contains(haystack []string, needle string) bool {
+	for _, item := range haystack {
+		if item == needle {
+			return true
+		}
+	}
+	return false
+}
+
+func gitVersion() (string, error) {
```
danudey, line 53: Should this maybe be more accurately named `gitDescribe()`? Calling it `gitVersion()` makes it sound like it's getting the version of Git that we're using.

radTuti: this is the standard established across our repos

hack/release-from/action.go
radTuti: i was not sure if i should expand the scope

danudey (approved):
lgtm, no other feedback

hack/release-from/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+To create a new release
+
+```sh
+release-from --base-version <previous operator version> --version <version to release> \
+  [--except-calico | --except-calico-enterprise] <image>:<image version>
+```
+
+> [!IMPORTANT]
+> To publish the newly created operator, use the `--publish` flag
```
rene-dekker, line 32: Maybe you can clarify what publish means, from what I understand it is pushing the tag, pushing images + image manifest, but not creating a release in https://github.com/tigera/operator/releases

hack/release-from/action.go
```diff
@@ -0,0 +1,305 @@
...
+	"gopkg.in/yaml.v3"
+)
+
+// releaseFrom is the main action for the command.
+// It builds (and publishes) a new operator based on the base version specified.
+func releaseFrom(ctx context.Context, c *cli.Command) error {
+	// get root directory of operator git repo
+	repoRootDir, err := runCommand("git", []string{"rev-parse", "--show-toplevel"}, nil)
+	if err != nil {
+		return fmt.Errorf("Error getting git root directory: %s", err)
```
rene-dekker, line 36: For golang users such as myself, there are a lot of warnings. Maybe nice to clean up.
https://go.dev/wiki/CodeReviewComments#error-strings

> Error Strings[¶](https://go.dev/wiki/CodeReviewComments#error-strings)
> Error strings should not be capitalized (unless beginning with proper nouns or acronyms) or end with > punctuation, since they are usually printed following other context. That is, use fmt.Errorf("something bad") not fmt.Errorf("Something bad"), so that log.Printf("Reading %s: %v", filename, err) formats without a spurious capital letter mid-message. This does not apply to logging, which is implicitly line-oriented and not combined inside other messages.

hack/release-from/action.go
```diff
@@ -0,0 +1,305 @@
...
+	}
+
+	// overwrite local file with updated config
+	var buf bytes.Buffer
+	encoder := yaml.NewEncoder(&buf)
+	encoder.SetIndent(2)
+	if err := encoder.Encode(&root); err != nil {
+		return fmt.Errorf("Error encoding updated config: %s", err)
+	}
+	encoder.Close()
```
rene-dekker, line 218: You are not handling the potential error

hack/release-from/flags.go
```diff
@@ -0,0 +1,163 @@
...
+	Usage: "A list of Calico images and the version to use for them. " +
+		"This should use the format based on the config/calico_versions.yaml file. " +
+		"e.g. --except-calico calico/cni:vX.Y.Z --except-calico csi-node-driver-registrar:vA.B.C-n-g<git-hash>",
+	Sources: cli.EnvVars("OS_IMAGES_VERSIONS"),
+	Action:  validateOverrides,
+}
+
+var exceptEnterpriseFlag = &cli.StringSliceFlag{
+	Name:    "except-calico-enterprise",
+	Aliases: []string{"except-enterprise", "except-calient"},
```
rene-dekker, line 100: The help for this cli is hard to read. I think the number of aliases is part of the reason why. Maybe keep it to 0 or 1 alias?

hack/release-from/flags.go
```diff
@@ -0,0 +1,163 @@
...
+	Usage: "A list of Enterprise images and the versions to use for them. " +
```
rene-dekker, line 101: I think this cli tool is not made to handle long usage strings, as a result the help output is hard to read.

rene-dekker (approved):

PR #3703, opened 2025-01-15
Allow kube-controllers to read its own configuration

nelljerram:
Previously we had optimisations in libcalico-go that translated a list of one item into a get.  But that wasn't sound and was removed in [#8469](https://github.com/tigera/calico-private/pull/8469).  So there can be places where we were previously OK with a "get" permission but now need "list" as well.

### DIFF
--- a/pkg/render/intrusion_detection.go
+++ b/pkg/render/intrusion_detection.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2019-2025 Tigera, Inc. All rights reserved.
+// Copyright (c) 2019-2024 Tigera, Inc. All rights reserved.
 
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -274,7 +274,7 @@ func (c *intrusionDetectionComponent) intrusionDetectionClusterRole() *rbacv1.Cl
 				"licensekeys",
 			},
 			Verbs: []string{
-				"get", "list", "watch",
+				"get", "watch",
 			},
 		},
 		{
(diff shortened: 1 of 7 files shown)


caseydavenport (approved):
LGTM - although I do wonder why we're doing a List() in kube controllers instead of a Get() - AFAIK we only support a single KubeControllersConfiguration!

caseydavenport:
@nelljerram oh, I approved before I saw the additional "list" added to other roles. Some of those seem unrelated to the libcalico change? e.g., storageclasses?

nelljerram:
> LGTM - although I do wonder why we're doing a List() in kube controllers instead of a Get() - AFAIK we only support a single KubeControllersConfiguration!

Yes, good question.  I haven't dug into that one yet.

nelljerram:
> @nelljerram oh, I approved before I saw the additional "list" added to other roles. Some of those seem unrelated to the libcalico change? e.g., storageclasses?
> 
> Those changes seem fishy - are we actually seeing failures / errors for those? Maybe the correct fix is to remove "watch" rather than add "list"?

No, we have not seen failures / errors for any other cases yet, so they are "prophylactic", or perhaps paranoid.

nelljerram:
(Also sorry, I was just running UT locally before pushing the additional commit...)

nelljerram:
Would you prefer me to remove the second commit for now?

caseydavenport:
@nelljerram yeah I think that we should address that in a second PR - especially with RBAC I don't want to add things that might not be necessary, as they become difficult to spot / remove in the future.

PR #3704, opened 2025-01-15
Allow kube-controllers to read its own configuration

nelljerram:
Previously we had optimisations in libcalico-go that translated a list of one item into a get.  But that wasn't sound and was removed in [#8469](https://github.com/tigera/calico-private/pull/8469).  So there can be places where we were previously OK with a "get" permission but now need "list" as well.


nelljerram:
This PR is the same as #3703 but for the master branch.

caseydavenport (approved):

PR #3706, opened 2025-01-16
Update libcalico-go version to the branch, rather than the last relea…

rene-dekker:
Update libcalico-go version to the branch, rather than the last released tag

As a consequence of fixing libcalico-go being pinned to the latest tag, we do not update CRD changes between that tag and the branch, which is problematic for the new feature where the operator deploys CRDs itself. ECK therefore, was still pinned to ECK 2.6.1, whereas we are currently deploying v2.16.0, due to CVE fixes, etc.

### DIFF
--- a/config/enterprise_versions.yml
+++ b/config/enterprise_versions.yml
@@ -2,7 +2,7 @@
 title: v3.20.0-2.2
 components:
   libcalico-go:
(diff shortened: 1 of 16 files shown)


tmjd (approved):
LGTM

PR #3727, opened 2025-01-29
Improve certificate management

rene-dekker:
- Rotate certificates 1 month before expiry (if issued by operator) to minimize potential for down time. (Currently certificates are valid for 825 days.)
- If a certificate is expired, a controller that is fetching a cert for the purpose of adding it to a trusted bundle will skip it. This is similar to the behaviour if a certificate does not exist at all.
- If a certificate is expired, a controller that is responsible for creating it will replace it. (This is already the current behaviour, just mentioning it here for completeness.)
- Improve the test suite, such that cert creation for some certs is done once, and not BeforeEach, cutting down run time of the suite from 40s -> 10s.

### DIFF
--- a/pkg/controller/certificatemanager/certificatemanager.go
+++ b/pkg/controller/certificatemanager/certificatemanager.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2022-2024 Tigera, Inc. All rights reserved.
+// Copyright (c) 2025 Tigera, Inc. All rights reserved.
 
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


pkg/controller/certificatemanager/certificatemanager.go
```diff
@@ -468,11 +468,33 @@ func (cm *certificateManager) getKeyPair(cli client.Client, secretName, secretNa
...
+			if readCertOnly {
+				cm.log.Info("secret %s/%s is not valid at this date, so we return nil and continue as though it does not exist", secretNamespace, secretName)
+				return nil, nil, nil
+			}
+			return nil, nil, fmt.Errorf("secret %s/%s is not valid at this date for a certificate not created by the operator, user action required", secretNamespace, secretName)
 		}
 		return nil, nil, newCertExtKeyUsageError(secretName, secretNamespace, requiredKeyUsages)
 	}
 
+	if !readCertOnly && x509Cert.NotAfter.Before(time.Now().Add(30*24*time.Hour)) {
```
Brian-McM, line 480: Might be nice to have the one month duration as a constant.

pkg/controller/certificatemanager/certificatemanager.go
```diff
@@ -468,11 +468,33 @@ func (cm *certificateManager) getKeyPair(cli client.Client, secretName, secretNa
...
+		// trusting the issuer, so there will be no disruption.
+		if !strings.HasPrefix(x509Cert.Issuer.CommonName, rmeta.TigeraOperatorCAIssuerPrefix) {
+			cm.log.Info("this certificate will soon expire and is not managed by the operator, user action required", "name", secretName)
+		} else {
+			if cm.keyPair.CertificateManagement != nil {
+				// When certificate management is enabled, we can simply return a certificate management key pair;
+				// the old secret will be deleted automatically.
+				return certificateManagementKeyPair(cm, secretName, secretNamespace, dnsNames), nil, nil
+			}
+			cm.log.Info("this certificate will soon expire, so we rotate it now", "name", secretName)
```
Brian-McM, line 492: Nit, this seems a little word, could just say "Rotating soon to expire certificate <name>."

pkg/controller/certificatemanager/certificatemanager.go
```diff
@@ -468,11 +468,33 @@ func (cm *certificateManager) getKeyPair(cli client.Client, secretName, secretNa
...
 		}
 		return nil, nil, newCertExtKeyUsageError(secretName, secretNamespace, requiredKeyUsages)
 	}
 
+	if !readCertOnly && x509Cert.NotAfter.Before(time.Now().Add(30*24*time.Hour)) {
+		// The certificate is not valid one month from now. Let's start the rotation process, so there will be plenty of time
+		// to roll out the changes without disruption. All components that need to trust this certificate are already
+		// trusting the issuer, so there will be no disruption.
+		if !strings.HasPrefix(x509Cert.Issuer.CommonName, rmeta.TigeraOperatorCAIssuerPrefix) {
+			cm.log.Info("this certificate will soon expire and is not managed by the operator, user action required", "name", secretName)
```
Brian-McM, line 485: This should probably be a warning log, and might be nice to update the wording so it's more clear to the user "Certificate <name> in secret <secret name and namespace> will expire soon and yadda yadda".

Brian-McM (approved):

PR #3753, opened 2025-02-03
[Cherry pick 1.36] Allow kube-controllers to read its own configuration

matthewdupre:
Previously we had optimisations in libcalico-go that translated a list of one item into a get.  But that wasn't sound and was removed in #8469.  So there can be places where we were previously OK with a "get" permission but now need "list" as well.


coutinhop (approved):

marvin-tigera:

marvin-tigera:

PR #3759, opened 2025-02-05
[RS-2297] Run `dashboard-installer` on operator update

Dean-Coakley:
## Description

https://tigera.atlassian.net/browse/RS-2297 - Honeypods dashboard remains in Kibana on upgrade from `v3.19`

* Make the operator generate a new job name per operator version. This will ensure a new `dashboard-installer` Job runs so that new Kibana dashboards are installed and deprecated ones are removed. Works in combination with https://github.com/tigera/calico-private/pull/8613

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/logstorage/dashboards/dashboards.go
+++ b/pkg/render/logstorage/dashboards/dashboards.go
@@ -48,6 +48,7 @@ var (
 	ServiceAccountName       = "dashboards-installer"
 	ElasticCredentialsSecret = "tigera-ee-dashboards-installer-elasticsearch-user-secret"
 	PolicyName               = networkpolicy.TigeraComponentPolicyPrefix + GetJobName()
+	SelectorContainsName     = "dashboards-installer"
 )
 
 // GetJobName makes a unique job name per operator version.
@@ -182,7 +183,7 @@ func (d *dashboards) AllowTigeraPolicy() *v3.NetworkPolicy {
 		Spec: v3.NetworkPolicySpec{
 			Order:    &networkpolicy.HighPrecedenceOrder,
 			Tier:     networkpolicy.TigeraComponentTierName,
-			Selector: fmt.Sprintf("job-name == '%s'", GetJobName()),
+			Selector: fmt.Sprintf("job-name contains '%s'", SelectorContainsName),
 			Types:    []v3.PolicyType{v3.PolicyTypeEgress},
 			Egress:   egressRules,
 		},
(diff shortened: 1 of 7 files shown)


gantony (approved):
LGTM

pkg/render/testutils/expected_policies/dashboards.json
```diff
@@ -8,7 +8,7 @@
   "spec": {
     "order": 1,
     "tier": "allow-tigera",
-    "selector": "job-name == 'dashboards-installer'",
+    "selector": "job-name contains 'dashboards-installer'",
```
caseydavenport, line 11: I think we should be using a different rather than the contains operator here. There's no reason we need to use the `job-name` label AFAIK - we could have `role: dashboard-installer` applied regardless of the actual name, for example. 

My concern is just that `contains` is a pretty unusual operation, and given there's no concrete _need_ for it here it makes this policy less intuitive.

pkg/render/logstorage/dashboards/dashboards.go
```diff
@@ -49,6 +50,18 @@ var (
 	PolicyName               = networkpolicy.TigeraComponentPolicyPrefix + Name
 )
 
+// GetJobName makes a unique job name per operator version.
+// For dev images it takes the first 32 chars.
+func GetJobName() string {
```
caseydavenport, line 55: If we're switching to this, we should stop exporting the `Name` variable so that nobody accidentally uses the wrong value.

caseydavenport: If `Name` isn't actually used except for within this function, we should probably just scope it to this function to be extra sure (i.e., remove the package level variable)

pkg/render/logstorage/dashboards/dashboards.go
```diff
@@ -49,6 +50,18 @@ var (
...
+	operatorVersion := strings.ReplaceAll(version.VERSION, ".", "-")
+
+	if len(operatorVersion) >= 32 {
```
caseydavenport, line 58: What is "32" in this context? Why do we need to truncate? I think we should have a comment explaining why we do this - it's not obvious!

rene-dekker (changes requested):
I thought changing the job would mean that the operator would delete the job object and recreate it, resulting in a new run. Am I missing something?
See https://github.com/tigera/operator/blob/f4b6cf444529e1f2ba7299db00a878261b83319d/pkg/controller/utils/component.go#L167

I think we should find out why that logic was not sufficient in re-running the job instead.

Dean-Coakley:
Closing.

Opening: https://github.com/tigera/operator/pull/3768 instead

PR #3768, opened 2025-02-06
[RS-2297]  Recreate dashboard-installer job on image mismatch

Dean-Coakley:
## Description

https://tigera.atlassian.net/browse/RS-2297 - Honeypods dashboard remains in Kibana on upgrade from v3.19

* The operator needs to delete and recreate the `dashboard-installer` resource with an updated image to ensure the installer runs on cluster upgrade

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/utils/component.go
+++ b/pkg/controller/utils/component.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2020-2024 Tigera, Inc. All rights reserved.
+// Copyright (c) 2020-2025 Tigera, Inc. All rights reserved.
 
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -446,9 +446,10 @@ func mergeState(desired client.Object, current runtime.Object) client.Object {
 		cj := current.(*batchv1.Job)
 		dj := desired.(*batchv1.Job)
 
-		// We're only comparing jobs based off of annotations for now so we can send a signal to recreate a job. Later
-		// we might want to have some better comparison of jobs so that a changed in the container spec would trigger
-		// a recreation of the job
+		if cj.Spec.Template.Spec.Containers[0].Image != dj.Spec.Template.Spec.Containers[0].Image {
+			return dj
+		}
+
 		if reflect.DeepEqual(cj.Spec.Template.Annotations, dj.Spec.Template.Annotations) {
 			return nil
 		}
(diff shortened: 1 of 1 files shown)


pkg/controller/utils/component.go
```diff
@@ -446,9 +446,10 @@ func mergeState(desired client.Object, current runtime.Object) client.Object {
 		cj := current.(*batchv1.Job)
 		dj := desired.(*batchv1.Job)
 
-		// We're only comparing jobs based off of annotations for now so we can send a signal to recreate a job. Later
-		// we might want to have some better comparison of jobs so that a changed in the container spec would trigger
-		// a recreation of the job
+		if cj.Spec.Template.Spec.Containers[0].Image != dj.Spec.Template.Spec.Containers[0].Image {
```
Dean-Coakley, line 449: @rene-dekker Is it a safe assumption here that there will always be one image in the spec?

rene-dekker: I think you should check for length and iterate over the containers to avoid a potential bug in the future. Otherwise, I think this fix will suffice.

rene-dekker (commented):
lgtm

rene-dekker (approved):

pkg/controller/utils/component.go
```diff
@@ -446,9 +446,16 @@ func mergeState(desired client.Object, current runtime.Object) client.Object {
 		cj := current.(*batchv1.Job)
 		dj := desired.(*batchv1.Job)
 
-		// We're only comparing jobs based off of annotations for now so we can send a signal to recreate a job. Later
-		// we might want to have some better comparison of jobs so that a changed in the container spec would trigger
-		// a recreation of the job
```
caseydavenport, line 451: For what it's worth, I would have left a comment of some sort in place here - having this called out is really useful for anyone coming in the future who may be wondering why the Job isn't being recreated. We didn't "solve" the problem outlined here - we just changed it from "annotations" to "annotations + images + #containers" 

There are a bunch of other fields that we still don't handle.

Dean-Coakley: Yes I didn't fully think this through. Comment is still definitely useful for people less familiar with the operator. (like me :sweat_smile: )

PR to restore comment with minor edit: https://github.com/tigera/operator/pull/3770

PR #3769, opened 2025-02-07
[RS-2260] Add GatewayAPI field to configure CRD management

nelljerram:
The default behaviour - which is used when the new field is not set, or is set to "PreferExisting" - is that the Tigera operator will create the Gateway API CRDs if they do not already exist, but will not overwrite any existing Gateway API CRDs.  This setting may be preferable if the customer is using other implementations of the Gateway API concurrently with the Gateway API support in Calico Enterprise.  It is then the customer's responsibility to ensure that CRDs are installed that meet the needs of all the Gateway API implementations in their cluster.

Alternatively, if this field is set to "Reconcile", the Tigera operator will keep the cluster's Gateway API CRDs aligned with those that it would install on a cluster that does not yet have any version of those CRDs.


api/v1/gatewayapi_types.go
```diff
@@ -22,6 +22,14 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+// +kubebuilder:validation:Enum=Reconcile;PreferExisting
+type GatewayCRDManagement string
+
+const (
+	GatewayCRDManagementReconcile      GatewayCRDManagement = "Reconcile"
```
Brian-McM, line 29: We could probably use this terminology / setting for any third part crd management, i.e. take away the `Gateway` prefix and put this in a common folder.

Brian-McM (approved):

marvin-tigera:
OK, I will merge the pull request when it's ready, squash the commits when I merge it, and delete the branch after I've merged it.

marvin-tigera:

marvin-tigera:

PR #3775, opened 2025-02-14
Deploy OS UI (Whisker) with Goldmane

Brian-McM:
## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


pkg/k8s/container.go
```diff
@@ -0,0 +1,120 @@
...
+	if err != nil {
+		return err
+	}
+
+	ctr.ctr.Image = image
+	return nil
+}
+
+func (ctr *Container) AddEnv(env ...corev1.EnvVar) *Container {
+	ctr.ctr.Env = append(ctr.ctr.Env, env...)
```
caseydavenport, line 50: I think it would be worth a check that we don't have duplicate env vars, and a panic if we do.

pkg/k8s/container.go
```diff
@@ -0,0 +1,120 @@
...
+	return ctr
+}
+
+func (ctr *Container) MountSecret(path string, scrt *corev1.Secret, opts ...VolumeMountOption) *Container {
+	if ctr.SecretMounted(scrt.Name) {
```
caseydavenport, line 55: Is there a mainline case where MountSecret would be called twice for the same secret? What if they are called with different options? Seems like a bug to me?

If so, probably also want to error / panic here if we hit this branch.

pkg/k8s/container.go
```diff
@@ -0,0 +1,120 @@
...
+	for _, scrt := range ctr.mountedSecrets {
+		if scrt.Name == name {
+			return true
+		}
+	}
+	return false
+}
+
+func (ctr *Container) MountConfigMap(path string, cm *corev1.ConfigMap, opts ...VolumeMountOption) *Container {
+	if ctr.ConfigMapMounted(cm.Name) {
```
caseydavenport, line 74: Same here

pkg/k8s/container.go
```diff
@@ -0,0 +1,120 @@
...
+	operatorv1 "github.com/tigera/operator/api/v1"
+	"github.com/tigera/operator/pkg/components"
+	"github.com/tigera/operator/pkg/render"
+	corev1 "k8s.io/api/core/v1"
+)
+
+type Container struct {
+	ctr corev1.Container
+
+	srvc *corev1.Service
```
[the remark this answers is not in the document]
caseydavenport: Are we confident that Service belongs at the container level? A Service selects a Pod, not a container.

Is this forcing our hand to have multiple services for a pod that exposes multiple ports, instead of a Single service with multiple ports?

pkg/k8s/deployment.go
```diff
@@ -0,0 +1,207 @@
...
+			cm.SetNamespace(d.dep.ObjectMeta.Namespace)
+			cms = append(cms, cm)
+		}
+	}
+
+	return cms
+}
+
+func (d *Deployment) Key() string {
+	key := d.dep.ObjectMeta.Name
```
caseydavenport, line 79: ISTR there is already a library function that does this?

https://pkg.go.dev/k8s.io/client-go/tools/cache#MetaNamespaceKeyFunc

Not really a big deal though.

pkg/render/guardian/guardian.go
```diff
@@ -0,0 +1,98 @@
...
+const (
+	GuardianContainerName              = "guardian"
+	GuardianServiceName                = "tigera-guardian"
+	ManagedClusterConnectionSecretName = "tigera-managed-cluster-connection"
+)
+
+func Container(tunnelCAType operatorv1.CAType, voltronURL string,
+	bundle certificatemanagement.TrustedBundle, tunnelSecret *corev1.Secret, proxyEnvVars []corev1.EnvVar) *_k8s.Container {
+
+	return _k8s.NewContainer(GuardianContainerName, components.Component{
```
caseydavenport, line 37: Why the underscore? I don't think thats a pattern we use commonly. Probably should just be `k8s`?

```suggestion
	return k8s.NewContainer(GuardianContainerName, components.Component{
```

pkg/render/guardian/guardian.go
```diff
@@ -0,0 +1,98 @@
...
+		Image:    "calico/guardian",
+		Version:  "bmv1.10",
```
caseydavenport, line 39: hardcoded versions here

pkg/render/guardian/guardian.go
```diff
@@ -0,0 +1,98 @@
...
+				{Name: "GUARDIAN_LOGLEVEL", Value: "DEBUG"},
+				{Name: "GUARDIAN_VOLTRON_URL", Value: voltronURL},
+				{Name: "GUARDIAN_VOLTRON_CA_TYPE", Value: string(tunnelCAType)},
+				{Name: "GUARDIAN_PACKET_CAPTURE_CA_BUNDLE_PATH", Value: bundle.MountPath()},
+				{Name: "GUARDIAN_PROMETHEUS_CA_BUNDLE_PATH", Value: bundle.MountPath()},
+				{Name: "GUARDIAN_QUERYSERVER_CA_BUNDLE_PATH", Value: bundle.MountPath()},
+			}...).AddEnv(proxyEnvVars...).
+		MountConfigMap(bundle.VolumeMountPath(rmeta.OSTypeLinux), bundle.ConfigMap("")).
+		MountSecret("/certs", copySecret(tunnelSecret)).
+		AddService(&corev1.Service{
```
caseydavenport, line 54: Personally I find this much harder to read than if we didn't inline this - the indentation is all over the place.

e.g., 

```go
svc := corev1.Service{ . . .}
component := components.Component(...)
env := []corev1.EnvVar{...}

container := k8s.NewContainer(GuardianContainerName, component).AddEnv(env).AddService(svc)
container.MountConfigMap(...)
container.MountSecret(...)
return container
```

pkg/render/namespaces.go
```diff
@@ -100,8 +100,8 @@ func CreateNamespace(name string, provider operatorv1.Provider, pss PodSecurityS
 
 	// Add in labels for configuring pod security standards.
 	// https://kubernetes.io/docs/concepts/security/pod-security-standards/
-	ns.Labels["pod-security.kubernetes.io/enforce"] = string(pss)
-	ns.Labels["pod-security.kubernetes.io/enforce-version"] = "latest"
+	//ns.Labels["pod-security.kubernetes.io/enforce"] = string(pss)
```
caseydavenport, line 103: Need to undo this?

pkg/render/node.go
```diff
@@ -985,6 +985,7 @@ func (c *nodeComponent) nodeDaemonset(cniCfgMap *corev1.ConfigMap) *appsv1.Daemo
 					ServiceAccountName:            CalicoNodeObjectName,
 					TerminationGracePeriodSeconds: &terminationGracePeriod,
 					HostNetwork:                   true,
+					DNSPolicy:                     corev1.DNSClusterFirstWithHostNet,
```
caseydavenport, line 988: Eek. This one is sorta scary... maybe we need to do it but I am pretty confident something is going to break. I just don't know what!

pkg/render/node.go
```diff
@@ -1294,8 +1295,8 @@ func (c *nodeComponent) nodeContainer() corev1.Container {
 
 	return corev1.Container{
 		Name:            CalicoNodeObjectName,
-		Image:           c.nodeImage,
-		ImagePullPolicy: ImagePullPolicy(),
+		Image:           "gcr.io/unique-caldron-775/brianmcmahon/calico/node:bmv1.10",
```
caseydavenport, line 1298: Can revert

pkg/render/whisker/component.go
```diff
@@ -0,0 +1,259 @@
...
+					Tolerations:        tolerations,
+					ImagePullSecrets:   secret.GetReferenceList(c.cfg.PullSecrets),
+				},
+			},
+		},
+	})
+
+	dep.AddContainers(c.whisker(), c.whiskerBackend(), c.goldmane())
+	if c.cfg.ManagementClusterConnection != nil {
+		guardianCtr := guardian.Container(cfg.ManagementClusterConnection.Spec.TLS.CA, c.cfg.ManagementClusterConnection.Spec.ManagementClusterAddr, c.cfg.TrustedCertBundle, c.cfg.TunnelSecret, c.cfg.Installation.Proxy.EnvVars())
```
caseydavenport, line 81: probably worth splitting this one, line is pretty long.

pkg/render/whisker/component.go
```diff
@@ -0,0 +1,259 @@
...
+
+func (c *WhiskerComponent) serviceAccount() *corev1.ServiceAccount {
+	return &corev1.ServiceAccount{
+		TypeMeta:   metav1.TypeMeta{Kind: "ServiceAccount", APIVersion: "v1"},
+		ObjectMeta: metav1.ObjectMeta{Name: WhiskerServiceAccountName, Namespace: WhiskerNamespace},
+	}
+}
+
+func (c *WhiskerComponent) whisker() *_k8s.Container {
+	return _k8s.NewContainer("whisker", components.Component{
```
caseydavenport, line 148: Same readability comments throughout here

pkg/render/whisker/component.go
```diff
@@ -0,0 +1,259 @@
...
+		})
+}
+
+func (c *WhiskerComponent) whiskerBackend() *_k8s.Container {
+	return _k8s.NewContainer("whisker-backend", components.Component{
+		Image:    "calico/whisker-backend",
+		Version:  "bmv1.10",
+		Registry: "gcr.io/unique-caldron-775/brianmcmahon/",
+	}).AddEnv([]corev1.EnvVar{
+		{Name: "LOG_LEVEL", Value: "DEBUG"},
```
caseydavenport, line 169: probably shouldn't default to DEBUG in production - do we need a config knob for this?

pkg/render/whisker/component.go
```diff
@@ -0,0 +1,259 @@
...
+		Image:    "calico/whisker",
+		Version:  "bmv1.10",
+		Registry: "gcr.io/unique-caldron-775/brianmcmahon/",
+	}).AddEnv(
+		[]corev1.EnvVar{
+			{Name: "LOG_LEVEL", Value: "DEBUG"},
+			{Name: "CA_CERT_PATH", Value: c.cfg.TrustedCertBundle.MountPath()},
+		}...).MountConfigMap(c.cfg.TrustedCertBundle.VolumeMountPath(rmeta.OSTypeLinux), c.cfg.TrustedCertBundle.ConfigMap("")).
+		AddService(&corev1.Service{
+			ObjectMeta: metav1.ObjectMeta{Name: "whisker"},
```
caseydavenport, line 158: We have a bunch of constants for "whisker" - maybe one of them? :grin:

pkg/render/whisker/component.go
```diff
@@ -0,0 +1,259 @@
...
+func (c *WhiskerComponent) whiskerBackend() *_k8s.Container {
+	return _k8s.NewContainer("whisker-backend", components.Component{
+		Image:    "calico/whisker-backend",
+		Version:  "bmv1.10",
+		Registry: "gcr.io/unique-caldron-775/brianmcmahon/",
+	}).AddEnv([]corev1.EnvVar{
+		{Name: "LOG_LEVEL", Value: "DEBUG"},
+		{Name: "CA_CERT_PATH", Value: c.cfg.TrustedCertBundle.MountPath()},
+		{Name: "PORT", Value: "3002"},
+		{Name: "GOLDMANE_HOST", Value: "localhost:7443"},
```
caseydavenport, line 172: I had sort of toyed with this idea of building containers / services on top of each other in other prototypes. Might be something to think about, rather than requiring a bunch of constants / keeping fields in sync across objects.

e.g., some pseudo-code

```go
container := NewContainer()
dep := NewDeployment(container)

// Build the service matching this deployment, and the container's port.
svc := NewService(dep.Labels, container.Port)
```

Commenting on this line because it would allow us to just pass in the goldmane container and get the port from that, rather than requiring two places to agree on the value. 

I don't think this is a comment worth actioning in this PR, and maybe not ever, but wanted to get it on your radar as an idea since you're looking at building tooling around this. 

Same sort of idea for RBAC:

```go
sa := NewServiceAccount()
role := NewRole()
binding := NewBinding(sa, role)
```

Rather than having them all agree on object names.

caseydavenport (commented):
Very first pass. I think building tooling to more consistently, easily, and reliably construct common Objects is probably a good idea, but I do worry about the work required to roll this out across the board and the likelihood of us ending up needing to maintain multiple different ways of constructing objects as a result. So I'm on the fence about shipping it as part of this PR - let's see what @tmjd and @rene-dekker have to say.

caseydavenport:
@Brian-McM I think you'll need some logic like this as well: https://github.com/caseydavenport/operator/pull/6/files#diff-ffe7b4311a6538484cedc5e685c3bde3e6ce1a95f86828103fb12992e8158842

pkg/controller/whisker/whisker_controller.go
```diff
@@ -0,0 +1,301 @@
...
+		}
+		tunnelSecret = nil
+	}
+
+	linseedCASecret, err := getResource[corev1.Secret](ctx, r.Client, newNSObjectKey(render.VoltronLinseedPublicCert, common.OperatorNamespace()))
+	if err != nil {
+		return result, err
+	}
+
+	var trustedCertBundle certificatemanagement.TrustedBundle
```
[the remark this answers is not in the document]
caseydavenport: I had started on this in my other prototype: https://github.com/caseydavenport/operator/pull/6/files#diff-32435fea1e63f120378e1249eb913a3aeaf8cc7611a62195a51c7df194aff76a

Unfortunately, the TyphaNodeTLS bundle is sort of an outlier for some reason (I think it predates our other cert management logic?)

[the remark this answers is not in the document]
caseydavenport: ^ Actually looks like most of that file is copy/paste from the core_controller that I didnt' clean up yet, so mostly not useful.

pkg/render/whisker/component.go
Brian-McM: Definitely worth thinking about, I like the constructor approach to force the things we need to be in the deployment / services.

As discussed out of band, I'll be moving these changes to a different PR (but I'll add something like this).

api/v1/whisker_types.go
```diff
@@ -0,0 +1,53 @@
...
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
+// +kubebuilder:resource:scope=Cluster
+
+type Whisker struct {
```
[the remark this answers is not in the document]
Brian-McM: I was going to leave that for follow up.

pkg/controller/utils/utils.go
```diff
@@ -455,6 +455,25 @@ func GetManagementClusterConnection(ctx context.Context, c client.Client) (*oper
...
+	client.Object
+}
+
+func Get[E any, ClientObj ClientObjType[E]](ctx context.Context, key client.ObjectKey, c client.Client) (*E, error) {
+	obj := new(E)
+
+	err := c.Get(ctx, key, ClientObj(obj))
+	if err != nil {
+		if errors.IsNotFound(err) {
+			return nil, nil
```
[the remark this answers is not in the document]
Brian-McM: I think it would be worth it to change this to `GetIfExists` to be clear on this (or just return the error). I don't think I've had issues in this repo with this pattern, but I have been tripped up in cases where I do actually need to propagate the not found error in certain cases.

Probably leaning towards returning the error...

config/enterprise_versions.yml
```diff
@@ -15,6 +15,9 @@ components:
   cnx-queryserver:
     image: tigera/cnx-queryserver
     version: master
+  goldmane:
```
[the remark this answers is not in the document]
caseydavenport: We probably will? I don't think we'd disable this capability there, right?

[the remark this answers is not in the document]
Brian-McM: I think we'd need to sort out how enterprise will use this before we allow it to be launched, wouldn't we? Would goldmane export in EE as well?

pkg/controller/whisker/whisker_controller.go
```diff
@@ -0,0 +1,301 @@
...
+	r.status.ClearDegraded()
+
+	return result, nil
+}
+
+func newNSObjectKey(name, namespace string) client.ObjectKey {
+	return types.NamespacedName{Name: name, Namespace: namespace}
+}
+
+func getResource[R any, T interface {
```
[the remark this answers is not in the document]
Brian-McM: Nope!

pkg/controller/utils/utils.go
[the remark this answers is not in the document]
Brian-McM: I just changed this to GetIfExists, are you OK with that?

pkg/render/node.go
Brian-McM: Do we have another way to get around not being able to resolve the cluster service address for this?

pkg/controller/whisker/whisker_controller.go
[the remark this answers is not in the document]
Brian-McM: removed

pkg/controller/whisker/whisker_controller.go
```diff
@@ -0,0 +1,301 @@
...
+		}
+	}
+
+	err = c.WatchObject(&operatorv1.Whisker{}, &handler.EnqueueRequestForObject{})
+	if err != nil {
+		return fmt.Errorf("%s failed to watch primary resource: %w", controllerName, err)
+	}
+
+	// Watch for changes to the secrets associated with the ManagementClusterConnection.
+	if err = utils.AddSecretsWatch(c, guardian.ManagedClusterConnectionSecretName, common.OperatorNamespace()); err != nil {
```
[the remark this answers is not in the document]
Brian-McM: done

pkg/controller/whisker/whisker_controller.go
```diff
@@ -0,0 +1,301 @@
...
+// Reconcile reads that state of the cluster for a ManagementClusterConnection object and makes changes based on the
+// state read and what is in the ManagementClusterConnection.Spec. The Controller will requeue the Request to be
+// processed again if the returned error is non-nil or Result.Requeue is true, otherwise upon completion it will
+// remove the work from the queue.
+func (r *Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
+	reqLogger := log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
+	reqLogger.Info("Reconciling Whisker")
+	result := reconcile.Result{}
+
+	variant, instl, err := utils.GetInstallation(ctx, r.Client)
```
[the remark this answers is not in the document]
Brian-McM: changed.

pkg/controller/whisker/whisker_controller.go
```diff
@@ -0,0 +1,301 @@
...
+	c.status.Run(opts.ShutdownContext)
+	return c
+}
+
+// blank assignment to verify that ReconcileConnection implements reconcile.Reconciler
+var _ reconcile.Reconciler = &Reconciler{}
+
+// Reconciler reconciles a ManagementClusterConnection object
+type Reconciler struct {
+	Client                     client.Client
```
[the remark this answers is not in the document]
Brian-McM: updated.

pkg/controller/whisker/whisker_controller.go
```diff
@@ -0,0 +1,301 @@
...
+	tunnelSecret := &corev1.Secret{}
+	err = r.Client.Get(ctx, types.NamespacedName{Name: render.GuardianSecretName, Namespace: common.OperatorNamespace()}, tunnelSecret)
+	if err != nil {
+		if !k8serrors.IsNotFound(err) {
+			return result, err
+		}
+		tunnelSecret = nil
+	}
+
+	linseedCASecret, err := getResource[corev1.Secret](ctx, r.Client, newNSObjectKey(render.VoltronLinseedPublicCert, common.OperatorNamespace()))
```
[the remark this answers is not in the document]
Brian-McM: done.

pkg/render/node.go
caseydavenport: The only other way would be for the operator to query the Service, extract the IP, and put that in directly. Which honestly isn't terrible as a solution. 

But I'm down to try this and see what happens, at least to start...

api/v1/whisker_types.go
```diff
@@ -0,0 +1,53 @@
...
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Status WhiskerStatus `json:"status,omitempty"`
```
caseydavenport, line 31: We'll need to add container overrides to everything this deploys. I'm OK for that to be a follow-up, but I also think it should be rather trivial to add now.

config/enterprise_versions.yml
caseydavenport, line 18: Should we just remove this from enterprise then?

pkg/components/calico.go
```diff
@@ -131,11 +131,28 @@ var (
 	}
 
 	ComponentCalicoGoldmane = Component{
-		Version:  "master",
+		Version:  "bmv1.10",
 		Image:    "calico/goldmane",
-		Registry: "",
+		Registry: "gcr.io/unique-caldron-775/brianmcmahon/",
```
caseydavenport, line 136: Need to regenerate - using your versions still

pkg/controller/whisker/whisker_controller.go
```diff
@@ -0,0 +1,279 @@
...
+			return fmt.Errorf("failed to add watch for secret %s/%s: %w", common.OperatorNamespace(), secretName, err)
+		}
+	}
+
+	err = c.WatchObject(&operatorv1.Whisker{}, &handler.EnqueueRequestForObject{})
+	if err != nil {
+		return fmt.Errorf("%s failed to watch primary resource: %w", controllerName, err)
+	}
+
+	if err = utils.AddSecretsWatch(c, certificatemanagement.CASecretName, common.OperatorNamespace()); err != nil {
```
caseydavenport, line 86: Belongs in the loop above?

pkg/controller/whisker/whisker_controller.go
```diff
@@ -0,0 +1,279 @@
...
+
+	certificateManager, err := certificatemanager.Create(r.cli, installation, r.clusterDomain, common.OperatorNamespace())
+	if err != nil {
+		r.status.SetDegraded(operatorv1.ResourceCreateError, "Unable to create the certificate manager", err, reqLogger)
+		return reconcile.Result{}, err
+	}
+
+	// Copy the secret from the operator namespace to the guardian namespace if it is present.
+	tunnelSecret := &corev1.Secret{}
+	err = r.cli.Get(ctx, types.NamespacedName{Name: render.GuardianSecretName, Namespace: common.OperatorNamespace()}, tunnelSecret)
```
caseydavenport, line 186: Probably only need to do this if there's a management cluster connection present?

pkg/controller/whisker/whisker_controller.go
```diff
@@ -0,0 +1,279 @@
...
+			return result, err
+		}
+	}
+
+	r.status.ClearDegraded()
+
+	return result, nil
+}
+
+func newNSObjectKey(name, namespace string) client.ObjectKey {
```
caseydavenport, line 277: Do we really need this function? Seems more or less just as verbose as inlineing `types.NamespacedName{Name: name, Namespace: ns}` but a lot less readable.

pkg/render/render.go
```diff
@@ -47,5 +47,5 @@ func SetClusterCriticalPod(t *corev1.PodTemplateSpec) {
 
 // ImagePullPolicy returns the image pull policy to use for all components.
 func ImagePullPolicy() corev1.PullPolicy {
-	return corev1.PullIfNotPresent
+	return corev1.PullAlways
```
caseydavenport, line 50: Should revert this change

pkg/render/whisker/component.go
```diff
@@ -0,0 +1,427 @@
...
+	operatorv1 "github.com/tigera/operator/api/v1"
+	rmeta "github.com/tigera/operator/pkg/render/common/meta"
+	"github.com/tigera/operator/pkg/render/common/secret"
+	"github.com/tigera/operator/pkg/tls/certificatemanagement"
+)
+
+// The names of the components related to the Guardian related rendered objects.
+const (
+	WhiskerName               = "whisker"
+	WhiskerNamespace          = common.CalicoNamespace
```
caseydavenport, line 43: I think it's probably better to have the namespace be a field on the configuration object so that controllers can configure the namespace, rather than hard coding it here.

pkg/render/whisker/component.go
```diff
@@ -0,0 +1,427 @@
...
+		c.clusterRoleBinding(),
+		c.goldmaneService(),
+		c.whiskerService(),
+		c.cfg.TrustedCertBundle.ConfigMap(WhiskerNamespace))
+
+	if c.cfg.ManagementClusterConnection != nil {
+		objs = append(objs,
+			secret.CopyToNamespace(WhiskerNamespace, c.cfg.TunnelSecret)[0],
+			// TODO maybe the controller needs to create this?
+			c.cfg.TrustedCertBundle.ConfigMap(WhiskerNamespace),
```
caseydavenport, line 139: We need to write another controller to handle this, per my other comment. Right now, both core_controller and whisker controller are going to be writing the exact same configmap but with different contents.

pkg/render/whisker/component.go
```diff
@@ -0,0 +1,427 @@
...
+	}
+}
+
+func (c *Component) whiskerContainer() corev1.Container {
+	return corev1.Container{
+		Name:            WhiskerContainerName,
+		Image:           c.whiskerImage,
+		ImagePullPolicy: render.ImagePullPolicy(),
+		Env: []corev1.EnvVar{
+			{Name: "LOG_LEVEL", Value: "DEBUG"},
```
caseydavenport, line 171: Still think DEBUG is the wrong default - we typically use INFO

pkg/render/whisker/component.go
```diff
@@ -0,0 +1,427 @@
...
+	}
+}
+
+func (c *Component) whiskerBackendContainer() corev1.Container {
+	return corev1.Container{
+		Name:            WhiskerBackendContainerName,
+		Image:           c.whiskerBackendImage,
+		ImagePullPolicy: render.ImagePullPolicy(),
+		Env: []corev1.EnvVar{
+			{Name: "LOG_LEVEL", Value: "DEBUG"},
```
caseydavenport, line 200: same here

pkg/render/whisker/component.go
```diff
@@ -0,0 +1,427 @@
...
+	env := []corev1.EnvVar{
+		{Name: "LOG_LEVEL", Value: "INFO"},
+		{Name: "CA_CERT_PATH", Value: "/certs/tls.crt"},
+		{Name: "PORT", Value: "7443"},
+	}
+	var volumeMounts []corev1.VolumeMount
+	if c.cfg.LinseedPublicCASecret != nil {
+		env = append(env, corev1.EnvVar{
+			Name:  "PUSH_URL",
+			Value: fmt.Sprintf("https://%s.%s.svc/api/v1/flows/bulk", "tigera-guardian", WhiskerNamespace)})
```
caseydavenport, line 220: I don't think we need to string-format "tigera-guardian" sicne it's just a constant string. 

Also, can't we use localhost since these are in the same pod?

pkg/render/whisker/component.go
```diff
@@ -0,0 +1,427 @@
...
+					Volumes:            volumes,
+				},
+			},
+		},
+	}
+}
+
+func secretMount(path string, scrt *corev1.Secret) corev1.VolumeMount {
+	return corev1.VolumeMount{
+		Name:      fmt.Sprintf("%s-%s", scrt.Name, "scrt"),
```
caseydavenport, line 323:
```suggestion
		Name:      fmt.Sprintf("%s-%s", scrt.Name, "secret"),
```

the characters saved aren't worth the added difficulty reading this :grinning:

pkg/render/whisker/component.go
```diff
@@ -0,0 +1,427 @@
...
+					Name: "linseed",
+					Port: 443,
+					TargetPort: intstr.IntOrString{
+						Type:   intstr.Int,
+						IntVal: 8080,
+					},
+					Protocol: corev1.ProtocolTCP,
+				},
+				{
+					Name: "elasticsearch",
```
caseydavenport, line 356: I don't think we need this port in OSS

pkg/render/whisker/component.go
```diff
@@ -0,0 +1,427 @@
...
+func (c *Component) clusterRole() *rbacv1.ClusterRole {
+	policyRules := []rbacv1.PolicyRule{
+		{
+			APIGroups: []string{"*"},
+			Resources: []string{"*"},
+			Verbs:     []string{"get", "list", "watch"},
+		},
+		{
+			APIGroups: []string{"projectcalico.org"},
+			Resources: []string{"felixconfigurations"},
```
caseydavenport, line 408: Hm, what accesses felixconfigurations from this deployment?

Brian-McM: UI APIs seems to need it when it's connected.

pkg/controller/whisker/whisker_controller.go
```diff
@@ -0,0 +1,231 @@
...
+	// Create a new controller
+	c, err := ctrlruntime.NewController(controllerName, mgr, controller.Options{Reconciler: reconciler})
+	if err != nil {
+		return fmt.Errorf("failed to create %s: %w", controllerName, err)
+	}
+
+	for _, secretName := range []string{
+		monitor.PrometheusServerTLSSecretName,
+		whisker.ManagedClusterConnectionSecretName,
+		"whisker-trusted-bundle",
```
rene-dekker, line 77: This is not a secret.

pkg/tls/certificatemanagement/certificatebundle.go
```diff
@@ -99,6 +99,10 @@ func createTrustedBundle(includeSystemBundle bool, name string, ca CertificateIn
 	return bundle, err
 }
 
+func (t *trustedBundle) SetName(name string) {
```
rene-dekker, line 102: I'd rather that we follow the established pattern and have the existing controller (core_controller in this case) add the certificates.

pkg/render/whisker/component.go
```diff
@@ -0,0 +1,318 @@
...
+		{Name: "CA_CERT_PATH", Value: "/certs/tls.crt"},
+		{Name: "PORT", Value: "7443"},
+	}
+	var volumeMounts []corev1.VolumeMount
+	if c.cfg.LinseedPublicCASecret != nil {
+		env = append(env, corev1.EnvVar{
+			Name:  "PUSH_URL",
+			Value: "https://localhost:8080/api/v1/flows/bulk"})
+		volumeMounts = []corev1.VolumeMount{
+			secretMount("/certs", c.cfg.LinseedPublicCASecret),
```
rene-dekker, line 193: How does this secret get into the namespace? I don't see it rendered in the PR, unless I'm missing something.

pkg/render/whisker/component.go
```diff
@@ -0,0 +1,318 @@
...
+	}
+
+	objs = append(objs,
+		render.CreateOperatorSecretsRoleBinding(WhiskerNamespace),
+		c.serviceAccount(),
+		c.clusterRole(),
+		c.clusterRoleBinding(),
+		c.goldmaneService(),
+		c.whiskerService(),
+		c.cfg.TrustedCertBundle.ConfigMap(WhiskerNamespace))
```
rene-dekker, line 116: This is not how we render the bundle elsewhere. I'd rather have us use the established pattern.

internal/controller/whisker_controller.go
```diff
@@ -0,0 +1,41 @@
...
+)
+
+// WhiskerReconciler reconciles a ManagementClusterConnection object
+type WhiskerReconciler struct {
+	client.Client
+	Log    logr.Logger
+	Scheme *runtime.Scheme
+}
+
+// +kubebuilder:rbac:groups=operator.tigera.io,resources=managementclusterconnections,verbs=get;list;watch;create;update;patch;delete
```
rene-dekker, line 36: The comments in this file are incorrect

pkg/controller/whisker/whisker_controller.go
```diff
@@ -0,0 +1,231 @@
...
+func Add(mgr manager.Manager, opts options.AddOptions) error {
+	enabled, err := utils.WhiskerEnabled(mgr.GetConfig())
+	if err != nil {
+		return fmt.Errorf("failed to check if whisker was enabled: %w", err)
+	}
+	if !enabled {
+		return nil
+	}
+
+	statusManager := status.New(mgr.GetClient(), "gold-rush", opts.KubernetesVersion)
```
rene-dekker, line 62: What is the reason and meaning for this name? Wouldn't it be better if we stick with whisker?

config/enterprise_versions.yml
Brian-McM: done.

pkg/components/calico.go
Brian-McM: done

pkg/controller/whisker/whisker_controller.go
Brian-McM: done

pkg/controller/whisker/whisker_controller.go
Brian-McM: Removed for now.

pkg/controller/whisker/whisker_controller.go
Brian-McM: removed

pkg/render/whisker/component.go
Brian-McM: done

pkg/render/whisker/component.go
Brian-McM: Done.

pkg/render/whisker/component.go
Brian-McM: Removed it.

pkg/render/whisker/component.go
Brian-McM: It's via the install manifest retrieved from CC.

pkg/controller/whisker/whisker_controller.go
Brian-McM: This is left over from some previous naming

api/v1/whisker_types.go
Brian-McM: If you're cool with it I'm going to defer it, I tracked it in a ticket. Not sure if goldmane is going to stay here so kind of want to leave overrides until we sort that out.

pkg/controller/utils/discovery.go
```diff
@@ -62,14 +62,32 @@ func RequiresTigeraSecure(cfg *rest.Config) (bool, error) {
...
+	if err != nil {
+		return false, err
+	}
+	for _, res := range resources.APIResources {
+		if strings.EqualFold(res.Kind, "Whisker") {
+			return true, nil
+		}
+	}
+
+	// Default to single-tenant.
```
caseydavenport, line 87: Comment here is wrong

pkg/controller/whisker/whisker_controller.go
```diff
@@ -0,0 +1,192 @@
...
+		return fmt.Errorf("%s failed to watch ImageSet: %w", controllerName, err)
+	}
+
+	if err := utils.AddDeploymentWatch(c, whisker.WhiskerDeploymentName, whisker.WhiskerNamespace); err != nil {
+		return fmt.Errorf("%s failed to watch Whisker deployment: %w", controllerName, err)
+	}
+
+	// Watch for changes to TigeraStatus.
+	if err = utils.AddTigeraStatusWatch(c, ResourceName); err != nil {
+		return fmt.Errorf("clusterconnection-controller failed to watch management-cluster-connection Tigerastatus: %w", err)
```
caseydavenport, line 97:
```suggestion
		return fmt.Errorf("whisker-controller failed to watch Tigerastatus: %w", err)
```

pkg/controller/whisker/whisker_controller.go
```diff
@@ -0,0 +1,192 @@
...
+// Reconcile reads that state of the cluster for a Whisker object and makes changes based on the
+// state read and what is in the Whisker.Spec. The Controller will requeue the Request to be
+// processed again if the returned error is non-nil or Result.Requeue is true, otherwise upon completion it will
+// remove the work from the queue.
+func (r *Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
+	reqLogger := log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
+	reqLogger.Info("Reconciling Whisker")
+	result := reconcile.Result{}
+
+	variant, installation, err := utils.GetInstallation(ctx, r.cli)
```
caseydavenport, line 146: I think we said we want to uninstall all of this if the Variant is enterprise since it's not supported there yet, right?

pkg/controller/whisker/whisker_controller.go
```diff
@@ -0,0 +1,192 @@
...
+	controllerName = "whisker-controller"
+	ResourceName   = "whisker"
+)
+
+var log = logf.Log.WithName(controllerName)
+
+// Add creates a new Reconciler Controller and adds it to the Manager. The Manager will set fields on the Controller
+// and start it when the Manager is started.
+func Add(mgr manager.Manager, opts options.AddOptions) error {
+	if !opts.WhiskerEnabled {
```
caseydavenport, line 51: This will probably prevent us from uninstalling Whisker in EE, since presumably the CRD won't exist there?

pkg/controller/options/options.go
```diff
@@ -26,6 +26,7 @@ import (
 // use to determine if they should run at all, or store them and influence their
 // reconciliation loops.
 type AddOptions struct {
+	WhiskerEnabled      bool
```
caseydavenport, line 29:
```suggestion
	WhiskerCRDExists      bool
```

"Enabled" seems to suggest that the CR is created, not simply that the API is present.

pkg/render/whisker/component.go
```diff
@@ -0,0 +1,279 @@
...
+			Verbs:     []string{"get", "list", "watch", "create", "update", "patch"},
+		},
+		{
+			APIGroups: []string{""},
+			Resources: []string{"secrets"},
+			Verbs:     []string{"get", "list", "watch"},
+		},
+	}
+
+	return &rbacv1.ClusterRole{
```
caseydavenport, line 272: This should be a namespaced Role instead of a ClusterRole

pkg/render/whisker/component.go
```diff
@@ -0,0 +1,279 @@
...
+			{
+				Kind:      "ServiceAccount",
+				Name:      WhiskerClusterRoleName,
+				Namespace: WhiskerNamespace,
+			},
+		},
+	}
+}
+
+func (c *Component) clusterRole() *rbacv1.ClusterRole {
```
caseydavenport, line 258:
```suggestion
func (c *Component) role() *rbacv1.Role {
```

pkg/tls/certificatemanagement/certificatebundle.go
caseydavenport: Yeah, let's remove this function for now

caseydavenport (changes requested):

pkg/render/whisker/component.go
Brian-McM: Arg darn copy and paste!

pkg/controller/whisker/whisker_controller.go
Brian-McM: Yes for right now, I wanted to follow up with uninstalling in a follow up PR. Right now nobody can install this, so I think follow up should be fine (if you agree).

Brian-McM: Forget what I said, I did the thing.

pkg/render/whisker/component.go
Brian-McM: This is removed for now and I'll add it back when guardian is added.

pkg/tls/certificatemanagement/certificatebundle.go
Brian-McM: All gone!

pkg/controller/whisker/whisker_controller.go
Brian-McM: removed.

pkg/render/whisker/component.go
```diff
@@ -0,0 +1,427 @@
...
+	policyRules := []rbacv1.PolicyRule{
+		{
+			APIGroups: []string{"*"},
+			Resources: []string{"*"},
+			Verbs:     []string{"get", "list", "watch"},
```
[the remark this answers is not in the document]
Brian-McM: All gone.

pkg/render/whisker/component.go
Brian-McM: Removed!

pkg/render/whisker/component.go
Brian-McM: bundle is removed from this PR.

pkg/controller/whisker/whisker_controller.go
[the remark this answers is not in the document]
Brian-McM: Deferred since we don't have the trusted bundle here anymore.

pkg/render/whisker/component.go
Brian-McM: Doesn't exist anymore

pkg/render/whisker/component.go
```diff
@@ -0,0 +1,427 @@
...
+func secretMount(path string, scrt *corev1.Secret) corev1.VolumeMount {
+	return corev1.VolumeMount{
+		Name:      fmt.Sprintf("%s-%s", scrt.Name, "scrt"),
+		MountPath: path,
+	}
+}
+
+func secretVolume(scrt *corev1.Secret) corev1.Volume {
+	return corev1.Volume{
+		Name:         fmt.Sprintf("%s-%s", scrt.Name, "scrt"),
```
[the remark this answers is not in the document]
Brian-McM: Doesn't exist anymore

pkg/render/whisker/component.go
Brian-McM: Removed.

cmd/main.go
```diff
@@ -367,6 +367,12 @@ If a value other than 'all' is specified, the first CRD with a prefix of the spe
 	}
 	setupLog.WithValues("provider", provider).Info("Checking type of cluster")
 
+	whiskerEnabled, err := utils.WhiskerEnabled(mgr.GetConfig())
```
caseydavenport, line 370:
```suggestion
	whiskerCRDExists, err := utils.WhiskerCRDExists(mgr.GetConfig())
```

cmd/main.go
```diff
@@ -367,6 +367,12 @@ If a value other than 'all' is specified, the first CRD with a prefix of the spe
...
+	if err != nil {
+		log.Error(err, "failed to check if whisker was enabled")
```
caseydavenport, line 372:
```suggestion
		log.Error(err, "failed to check if Whisker CRD exists")
```

cmd/main.go
```diff
@@ -440,6 +446,7 @@ If a value other than 'all' is specified, the first CRD with a prefix of the spe
 		ManageCRDs:          manageCRDs,
 		ShutdownContext:     ctx,
 		MultiTenant:         multiTenant,
+		WhiskerCRDExists:    whiskerEnabled,
```
caseydavenport, line 449:
```suggestion
		WhiskerCRDExists:    whiskerCRDExists,
```

pkg/controller/installation/core_controller.go
```diff
@@ -1360,8 +1368,20 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 		return reconcile.Result{}, err
 	}
 
+	var goldmaneRunning bool
+	// Goldmane can only be running if the variant is Calico and the Whisker CRD exists.
+	if instance.Spec.Variant == operator.Calico && r.whiskerCRDExists {
+		whiskerCR, err := utils.GetIfExists[operatorv1.Whisker](ctx, utils.DefaultInstanceKey, r.client)
+		if err != nil {
+			r.status.SetDegraded(operator.ResourceReadError, "Unable Retrieve Whisker CR", err, reqLogger)
```
caseydavenport, line 1376:
```suggestion
			r.status.SetDegraded(operator.ResourceReadError, "Unable to retrieve Whisker CR", err, reqLogger)
```

pkg/controller/utils/discovery.go
```diff
@@ -62,14 +62,31 @@ func RequiresTigeraSecure(cfg *rest.Config) (bool, error) {
...
 		case "EgressGateway":
-			fallthrough
-		case "ManagementClusterConnection":
 			return true, nil
 		}
 	}
 	return false, nil
 }
 
+func WhiskerEnabled(cfg *rest.Config) (bool, error) {
```
caseydavenport, line 71:
```suggestion
func WhiskerCRDExists(cfg *rest.Config) (bool, error) {
```

caseydavenport (commented):
A few more minor naming cleanups, and as discussed we need some UTs to start us off. Otherwise this LGTM

caseydavenport (approved):

PR #3786, opened 2025-02-24
[RS-23978] Make operator api a go module

LorcanMcVeigh:
## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


pkg/render/monitor/monitor_test.go
```diff
@@ -359,7 +359,7 @@ var _ = Describe("monitor rendering tests", func() {
 		Expect(prometheusObj.Spec.RuleSelector.MatchLabels["role"]).To(Equal("tigera-prometheus-rules"))
 		Expect(prometheusObj.Spec.Alerting.Alertmanagers).To(HaveLen(1))
 		Expect(prometheusObj.Spec.Alerting.Alertmanagers[0].Name).To(Equal("calico-node-alertmanager"))
-		Expect(prometheusObj.Spec.Alerting.Alertmanagers[0].Namespace).To(Equal("tigera-prometheus"))
+		Expect(*prometheusObj.Spec.Alerting.Alertmanagers[0].Namespace).To(Equal("tigera-prometheus"))
```
gantony, line 362: Do you really need this change?

LorcanMcVeigh: apparently yes, the CI was throwing an error saying to couldn't compare a string to a string pointer. There are also a few other `Expect` cases that do this just above this one, so there is a precedence for it in this test

pkg/apis/crd.projectcalico.org/v1/ippool.go
```diff
@@ -15,6 +15,7 @@
 package v1
 
 import (
+	operatorapi "github.com/tigera/operator/api/v1"
```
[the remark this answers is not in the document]
caseydavenport: Yep - conventionally we have called this `operator` in other packages (sometimes `operatorv1` but I think that is also more wordy than it needs to be).

`v1` is slightly less good - there are several `v1` packages used in this codebase.

api/go.mod
```diff
@@ -0,0 +1,29 @@
+module github.com/tigera/operator/api
+
+go 1.23.5
+
+require (
+	github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring v0.80.1
```
caseydavenport, line 6: It would be nice to remove this dependency as well

pkg/apis/crd.projectcalico.org/v1/ippool.go
```diff
@@ -103,11 +104,9 @@ const (
 	IPIPModeCrossSubnet IPIPMode = "CrossSubnet"
 )
 
-type AssignmentMode string
-
 const (
-	Automatic AssignmentMode = "Automatic"
-	Manual    AssignmentMode = "Manual"
+	Automatic operatorapi.AssignmentMode = "Automatic"
```
caseydavenport, line 108: These should be defined in the operator API package rather than here.

api/v1/installation_types.go
```diff
@@ -658,6 +656,8 @@ func (nt NATOutgoingType) String() string {
 
 const NodeSelectorDefault string = "all()"
 
+type AssignmentMode string
```
caseydavenport, line 659: Missing the const definitions for this type?

pkg/controller/ippool/pool_controller.go
```diff
@@ -24,8 +24,7 @@ import (
 	configv1 "github.com/openshift/api/config/v1"
 	v3 "github.com/tigera/api/pkg/apis/projectcalico/v3"
 
-	operator "github.com/tigera/operator/api/v1"
-	v1 "github.com/tigera/operator/api/v1"
+	operatorapi "github.com/tigera/operator/api/v1"
```
caseydavenport, line 27:
```suggestion
	operator "github.com/tigera/operator/api/v1"
```

pkg/render/monitor/monitor_test.go
caseydavenport: @LorcanMcVeigh do you know why? It's not obvious to me why this has changed in this PR. Guessing this is because of an accidental bump in the prometheus dependencies we import?

If so, I recommend reverting that change - it shouldn't be needed for what you want here

api/go.mod
LorcanMcVeigh: This import is used in the `Endpoint` API here https://github.com/tigera/operator/blob/master/api/v1/monitor_types.go#L69 for the `Interval` and `ScrapeTimeout` fields. In order to remove that dependency we would need to change the type of those fields or move the `Endpoint` struct to somewhere else. Moving the struct would require moving the `ServiceMonitor`, ` ExternalPrometheus` & `MonitorSpec` out of the API as well because those structs are all dependent on each other in the `installation_types.go` .

I don't know how/where these definitions are used, so any guidance on this would be much appreciated

pkg/render/monitor/monitor_test.go
LorcanMcVeigh: yeah it was done because of `go mod tidy` it updated prometheus to `v0.80.1`. I've reverted it but whenever i run `go mod tidy` it updates it from `v0.70.0`=>`v0.80.1`

pkg/apis/crd.projectcalico.org/v1/ippool.go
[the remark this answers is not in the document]
LorcanMcVeigh: I actually thought the opposite. Having an import named `operator` in operator might have been confusing so I opted for something more descriptive but happy to change it if that's the consensus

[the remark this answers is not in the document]
caseydavenport: Yeah, it's just what we have used (somewhat inconsistently) in this repo.

api/go.mod
caseydavenport: Sounds like it might be more work than it is worth.

caseydavenport: Let's ignore for now

pkg/apis/crd.projectcalico.org/v1/ippool.go
caseydavenport: We should rename both of these constants - the convention for enums is that they should include the type in the name to avoid this type of conflict.

```go
type AssignmentMode string

const (
  AssignmentModeAutomatic = "Automatic"
  AssignmentModeManual = "Manual"
)
```

pkg/render/monitor/monitor_test.go
LorcanMcVeigh: yeah we need to update the version of promtheus-monitor in `operator`  because it's relation to https://github.com/tigera/operator/pull/3786#discussion_r1973923098

go.mod
```diff
@@ -97,40 +94,44 @@ require (
...
-	k8s.io/utils v0.0.0-20240821151609-f90d01438635 // indirect
-	sigs.k8s.io/json v0.0.0-20221116044647-bc3834ca7abd // indirect
-	sigs.k8s.io/structured-merge-diff/v4 v4.4.1 // indirect
+	k8s.io/kube-openapi v0.0.0-20241105132330-32ad38e42d3f // indirect
+	k8s.io/utils v0.0.0-20241210054802-24370beab758 // indirect
+	sigs.k8s.io/json v0.0.0-20241014173422-cfa47c3a1cc8 // indirect
+	sigs.k8s.io/structured-merge-diff/v4 v4.5.0 // indirect
 )
 
+require golang.org/x/sync v0.11.0 // indirect
```
caseydavenport, line 129: Don't think this needs to be in its own `require` block?

caseydavenport (approved):
One nit but otherwise this LGTM

LorcanMcVeigh:
I'm getting a CI failure for the CRD's with `validate-gen-versions` on this PR CI, this branch locally and my local master branch. I've pulled the latest from master and it started to happen, does anyone have any idea @tigera/tigera-operator-maintainers ?

PR #3791, opened 2025-02-26
Don't create resources in terminating Namespaces

caseydavenport:
## Description

In some cases, the operator can attempt to create new resources within a
terminating Namespace with fianlizers set on objects within. If this happens, an error will occur, which
prevents reaching the NodeComponent who is responsible for removing
those finalizers. As a result, we deadlock. 

We should never be installing resources into Namespaces that are being torn down, since it will never succeed! 

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/utils/component.go
+++ b/pkg/controller/utils/component.go
@@ -132,20 +132,6 @@ func (c *componentHandler) createOrUpdateObject(ctx context.Context, obj client.
 		logCtx.V(2).Info("Failed converting object", "obj", obj)
 		return fmt.Errorf("failed converting object %+v", obj)
 	}
-	// Check to see if the object's Namespace exists, and whether the Namespace
-	// is currently terminating. We cannot create objects in a terminating Namespace.
-	namespaceTerminating := false
-	if ns := cur.GetNamespace(); ns != "" {
-		nsKey := client.ObjectKey{Name: ns}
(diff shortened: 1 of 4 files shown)


pkg/controller/utils/component.go
```diff
@@ -132,14 +132,34 @@ func (c *componentHandler) createOrUpdateObject(ctx context.Context, obj client.
...
+	// Check to see if the object exists or not - this determines whether we should create or update.
 	err := c.client.Get(ctx, key, cur)
 	if err != nil {
 		if !errors.IsNotFound(err) {
 			// Anything other than "Not found" we should retry.
 			return err
 		}
 
+		if namespaceTerminating {
+			logCtx.Info("Object's Namespace is terminating, skipping creation.")
```
tmjd, line 159: Should we log what we're skipping creation of?

tmjd (commented):
For all the changes in the in the component_test I don't see any test for this new behavior. Could you add one?

tmjd (approved):

pkg/controller/utils/component.go
caseydavenport: It's included in the context logger already, so no need to add additional fields.

PR #3805, opened 2025-03-06
Add env vars to the whisker deployment

WilliamTigera:
## Description

Add ENV vars that the whisker UI to retrieve cluster context.

Copied over ClusterInformation CR so we don't need to use the V3 api (and wait for apiserver) 

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


test/whisker_test.go
```diff
@@ -136,6 +136,10 @@ var _ = Describe("Tests for Whisker installation", func() {
 			Expect(c.Delete(shutdownContext, goldmaneCR)).ShouldNot(HaveOccurred())
 		}()
 
+		// API server needs to be installed to use the v3 api
+		createAPIServer(c, mgr, shutdownContext, nil)
```
Brian-McM, line 140: This is supposed to run in an OS cluster, so we can't rely on the api server being installed.

WilliamTigera: Fixed.

WilliamTigera: Fixed htis by copying over the clusterinformation cr reference to the repo.

Brian-McM (dismissed):

pkg/controller/whisker/controller.go
```diff
@@ -200,6 +201,16 @@ func (r *Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
...
 	}
 
+	var calicoVersion string
+	var clusterType string
+	clusterInfo := &crdv1.ClusterInformation{}
+	err = r.cli.Get(ctx, utils.DefaultInstanceKey, clusterInfo)
+	if err != nil {
+		reqLogger.Info("Unable to retrieve cluster context to Whisker. Proceeding without adding cluster context to Whisker.", err)
+	}
+	calicoVersion = clusterInfo.Spec.CalicoVersion
```
Brian-McM, line 211: This will panic if the clusterInfo isn't there, Spec will be nil. I would not set the cluster type or calico versions if this is nil.

The whisker backend should handle that appropriate if the env variables are not set.

Brian-McM (changes requested):

pkg/controller/whisker/controller.go
WilliamTigera: Just a note these values are fed to the whisker front end directly.

pkg/crds/enterprise/crd.projectcalico.org_felixconfigurations.yaml
```diff
@@ -982,6 +982,10 @@ spec:
                   Felix exports flow logs.
                 pattern: ^([0-9]+(\\.[0-9]+)?(ms|s|m|h))*$
                 type: string
+              flowLogsGoldmaneServer:
```
WilliamTigera, line 985: Had to add these cause it was failing dirty check ci stage

Brian-McM (approved):

PR #3806, opened 2025-03-07
[CORE-10623] Make apiserver ports configurable to avoid conflicts on EKS

lucastigera:
## Description

The current implementation includes the following changes:
- Added a new `Ports` attribute to the `APIServerDeploymentContainer`.
- Updated the APIServer CRD to include the new `Ports` attribute.
- Modified the `components.go` file to incorporate logic for considering this attribute in the override logic.
- Replaced hardcoded constants for the Readiness and Liveness ports in the ApiServer, QueryServer, and L7AdmissionController containers with default values or overridden values.
- In the `Services` configurations, use the PortName instead of the number, if specified. 

## Related PR
- **L7 Admission Controller** - [Make the server port configurable](https://github.com/tigera/calico-private/pull/8780)

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


pkg/render/apiserver.go
```diff
@@ -976,7 +1030,7 @@ func (c *apiServerComponent) apiServerService() *corev1.Service {
 					Name:       "apiserver",
 					Port:       443,
 					Protocol:   corev1.ProtocolTCP,
-					TargetPort: intstr.FromInt(APIServerPort),
+					TargetPort: getPortNameOrNumber(apiServerTargetPort),
```
tmjd, line 1033: I'm not a fan of us using named ports sometimes (when overrides are used) and using port numbers others.
I think for the service we should always use named ports or always use the port numbers.

pkg/render/apiserver.go
```diff
@@ -1392,9 +1452,10 @@ func (c *apiServerComponent) tolerations() []corev1.Toleration {
 // being cut off from the main API server. The enterprise equivalent is currently handled in manifests.
 //
 // Calico only.
-func (c *apiServerComponent) networkPolicy() *netv1.NetworkPolicy {
+func (c *apiServerComponent) networkPolicy(cfg *APIServerConfiguration) *netv1.NetworkPolicy {
```
tmjd, line 1455: Why do we need to pass in cfg *APIServerConfiguration when `c.cfg` is already APIServerConfiguration?

tmjd (commented):
Looking good. I think some validation needs to be added to the APIServer controller to validate that the correct port name is used with the correct container.
In other words, checks like that the queryserver isn't be provided for the APIServer container.

pkg/render/apiserver.go
lucastigera: Nice catch! It's definitely an oversight, I'll change it. Thanks.

pkg/render/apiserver.go
lucastigera: The challenge here is that when we are not using the override, the containers don't have the ports explicitly configured, so we can't bind the port name with the container (or can we?).

For example, the following is the current configuration for the `apiServer` container:
```
	apiServer := corev1.Container{
		Name:            string(APIServerContainerName),
		Image:           c.apiServerImage,
		ImagePullPolicy: ImagePullPolicy(),
		Args:            c.startUpArgs(),
		Env:             env,
		VolumeMounts:    volumeMounts,
		ReadinessProbe: &corev1.Probe{
			ProbeHandler: corev1.ProbeHandler{
				HTTPGet: &corev1.HTTPGetAction{
					Path:   "/readyz",
					Port:   intstr.FromInt32(apiServerTargetPort),
					Scheme: corev1.URISchemeHTTPS,
				},
			},
			PeriodSeconds: 60,
		},
	}
```

I can change that to include the default port definition with the name, though. Something like this:

```
apiServer := corev1.Container{
    Name:            string(APIServerContainerName),
    Image:           c.apiServerImage,
    Ports: []corev1.ContainerPort{
        {
            ContainerPort: apiServerTargetPort,
            Name:          "apiserver",
        },
    },
    ImagePullPolicy: ImagePullPolicy(),
    Args:            c.startUpArgs(),
    Env:             env,
    VolumeMounts:    volumeMounts,
    ReadinessProbe: &corev1.Probe{
        ProbeHandler: corev1.ProbeHandler{
            HTTPGet: &corev1.HTTPGetAction{
                Path:   "/readyz",
                Port:   intstr.FromInt32(apiServerTargetPort),
                Scheme: corev1.URISchemeHTTPS,
            },
        },
        PeriodSeconds: 60,
    },
}
```
Would that be your expected approach?

lucastigera: Never mind, I kept the port number for the services. =)

pkg/render/common/components/components.go
```diff
@@ -113,16 +113,39 @@ func valueToContainers(value reflect.Value) []corev1.Container {
 	for _, v := range value.Seq2() {
 		name := v.FieldByName("Name")
 		resources := v.FieldByName("Resources")
-		if !resources.IsNil() {
-			cs = append(cs, corev1.Container{
-				Name:      name.String(),
-				Resources: *(resources.Interface().(*corev1.ResourceRequirements)),
-			})
+		ports := valueToContainerPorts(v)
```
tmjd, line 116: @nelljerram could you take a look at these changes and see if they fit with what you'd expect when adding some port overrides?
One thing that seems odd to me is that we can't handle the ports in the same way Resources are handled.

pkg/render/common/components/components.go
```diff
@@ -490,7 +513,12 @@ func mergeContainers(current []corev1.Container, provided []corev1.Container) {
 
 	for i, c := range current {
 		if override, ok := providedMap[c.Name]; ok {
-			current[i].Resources = override.Resources
+			if !reflect.DeepEqual(override.Resources, corev1.ResourceRequirements{}) {
```
tmjd, line 516: Is this needed? Because I think this would block someone from specifying an override to unset Resources.
I doubt there is a real use case for clearing the Resources, but I guess there could be.

pkg/render/common/components/components.go
tmjd: I've taken a closer look and I understand why the ports can't be handled the same.
I'd still appreciate a look when you have the time.

lucastigera: Just to provide more context: the reason I didn't handle Ports in the same way as Resources is that Ports is a custom type, while Resources is a core Kubernetes type. So I needed to map the Ports fields manually.

pkg/render/common/components/components.go
lucastigera: At that point in the code, the `current[i].Resources` are set to our default values — the ones configured when creating the deployment. So, if the user removes a "custom" Resources block, the system will assume the default values (which are currently empty).

On the other hand, if I allow an "empty" `override.Resources` to replace `current[i].Resources`, we could end up in a situation where the user has set custom ports but left the resources unset. This could unintentionally remove the default values if they are set in the future.

tmjd (approved):
I'm approving this on the condition that you do some manual testing of the new overrides because I'll be OOO the next couple days and I don't want to hold this PR up.
Nell can review after the fact.

lucastigera:
> I'm approving this on the condition that you do some manual testing of the new overrides because I'll be OOO the next couple days and I don't want to hold this PR up. Nell can review after the fact.

Manual tests have been done and run successfully.

pkg/render/common/components/components.go
nelljerram: @tmjd This all looks good to me; nice job @lucastigera .  I guess I might have considered using `ContainerPort` directly instead of defining `APIServerDeploymentContainerPort`, and then `valueToContainerPorts` would be a bit simpler.  But perhaps you considered that and there's a reason why it's better to use a new more specific struct.  Anyway not a big deal either way.

I noticed that the code - prior to your change - doesn't check for `resources.IsValid()` in `valueToContainers`.  This will be OK so long as all of our container override structs have a "Resources" field, which is currently the case.  But now that we have the Ports option as well, it's possible that a future override struct would have "Ports" but not "Resources", and so maybe we should add a `resources.IsValid()` check to allow for that - WDYT?

tmjd: I think we should always have Resources for a container overrides.

lucastigera: Thanks for your input on this matter, @nelljerram. We did consider using `corev1.ContainerPort` for a direct map, but ultimately, we decided to create a specific struct to avoid exposing unnecessary fields. =)

nelljerram: OK, thanks.

PR #3808, opened 2025-03-08
Split out guardian and goldmane into their own deployments

Brian-McM:
## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


pkg/controller/certificatemanager/certificatemanager.go
```diff
@@ -583,6 +585,23 @@ func HasExpectedDNSNames(secretName, secretNamespace string, cert *x509.Certific
...
+// CreateNamedTrustedBundleFromSecrets creates a TrustedBundle, which provides standardized methods for mounting a bundle of certificates to trust.
+// It will include:
+// - A bundle with Calico's root certificates + any user supplied certificates in /etc/pki/tls/certs/tigera-ca-bundle.crt.
+func (cm *certificateManager) CreateNamedTrustedBundleFromSecrets(name string, cli client.Client, namespace string, secretsToTrust ...string) (certificatemanagement.TrustedBundle, error) {
+	trustedBundle := certificatemanagement.CreateNamedTrustedBundle(name, cm.keyPair)
+	for _, secretName := range secretsToTrust {
+		secret, err := cm.GetCertificate(cli, secretName, namespace)
+		if err != nil {
+			return nil, fmt.Errorf("failed to retreive secret %s: %w", secretName, err)
+		} else if secret == nil {
```
rene-dekker, line 597: The else here is redundant, addCertificates can handle nil inputs.

pkg/controller/certificatemanager/certificatemanager.go
```diff
@@ -583,6 +585,23 @@ func HasExpectedDNSNames(secretName, secretNamespace string, cert *x509.Certific
...
 
+// CreateNamedTrustedBundleFromSecrets creates a TrustedBundle, which provides standardized methods for mounting a bundle of certificates to trust.
+// It will include:
+// - A bundle with Calico's root certificates + any user supplied certificates in /etc/pki/tls/certs/tigera-ca-bundle.crt.
+func (cm *certificateManager) CreateNamedTrustedBundleFromSecrets(name string, cli client.Client, namespace string, secretsToTrust ...string) (certificatemanagement.TrustedBundle, error) {
+	trustedBundle := certificatemanagement.CreateNamedTrustedBundle(name, cm.keyPair)
+	for _, secretName := range secretsToTrust {
+		secret, err := cm.GetCertificate(cli, secretName, namespace)
+		if err != nil {
+			return nil, fmt.Errorf("failed to retreive secret %s: %w", secretName, err)
```
rene-dekker, line 596:
```suggestion
			return nil, fmt.Errorf("failed to retrieve secret %s: %w", secretName, err)
```

pkg/render/whisker/component.go
```diff
@@ -183,139 +146,21 @@ func (c *Component) whiskerService() *corev1.Service {
...
-		Spec: corev1.ServiceSpec{
-			Ports: []corev1.ServicePort{{Port: 7443}},
-			Selector: map[string]string{
-				"k8s-app": WhiskerDeploymentName,
-			},
+		Env: []corev1.EnvVar{
+			{Name: "LOG_LEVEL", Value: "INFO"},
+			{Name: "PORT", Value: "3002"},
+			{Name: "GOLDMANE_HOST", Value: "goldmane.calico-system.svc.cluster.local:7443"},
+			{Name: "TLS_CERT_PATH", Value: c.cfg.WhiskerBackendKeyPair.VolumeMountCertificateFilePath()},
```
caseydavenport, line 157: Strictly speaking this is introducing the possibility of a nil pointer dereference - I thought we usually had nil checks for this sort of thing.

Probably not the end of the world - we expect this to always be non-nil - but expectations can change over time so the safest code here would be to keep the nil check that we had before.

pkg/controller/whisker/controller.go
```diff
@@ -223,49 +188,32 @@ func (r *Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
 		return reconcile.Result{}, err
 	}
 
-	// Guardian needs a certificate for mTLS with Goldmane.
-	// TODO: Add this to the trusted bundle. This isn't stritctly needed, since the bundle already includes the operator CA that
-	// signed this certificate. But in order to support custom user-supplied certificates, we will need to do this.
-	guardianCertificateNames := dns.GetServiceDNSNames("guardian", whisker.WhiskerNamespace, r.clusterDomain)
-	guardianCertificateNames = append(guardianCertificateNames, "localhost", "127.0.0.1")
-	guardianKeyPair, err := certificateManager.GetOrCreateKeyPair(r.cli, whisker.GuardianKeyPairSecret, whisker.WhiskerNamespace, guardianCertificateNames)
+	trustedBundle, err := certificateManager.CreateNamedTrustedBundleFromSecrets("goldmane-trusted-certs", r.cli,
```
caseydavenport, line 191: Is this intentionally using Goldmane's trusted bundle and not its own?

caseydavenport (commented):
First pass done

pkg/render/goldmane/component.go
```diff
@@ -0,0 +1,250 @@
...
+	return rmeta.OSTypeLinux
+}
+
+func (c *Component) Objects() ([]client.Object, []client.Object) {
+	objs := []client.Object{
+		c.serviceAccount(),
+		c.role(),
+		c.roleBinding(),
+		c.goldmaneService(),
+		c.deployment(),
```
caseydavenport, line 99: I think we probably want a network policy here as well, and for whisker

pkg/controller/clusterconnection/clusterconnection_controller.go
```diff
@@ -231,111 +224,89 @@ func (r *ReconcileConnection) Reconcile(ctx context.Context, request reconcile.R
...
+		}
+	}
 
-	// Write the discovered configuration back to the API. This is essentially a poor-man's defaulting, and
-	// ensures that we don't surprise anyone by changing defaults in a future version of the operator.
-	if err := r.Client.Patch(ctx, managementClusterConnection, preDefaultPatchFrom); err != nil {
+	if err := utils.ApplyDefaults(ctx, r.cli, managementClusterConnection); err != nil {
 		r.status.SetDegraded(operatorv1.ResourceUpdateError, err.Error(), err, reqLogger)
 		return reconcile.Result{}, err
 	}
```
caseydavenport, line 256: We should have a validation step here that verifies the MCC CA type is set to Tigera and not Public

caseydavenport (commented):
A couple more comments

pkg/render/guardian.go
```diff
@@ -480,54 +540,69 @@ func guardianAllowTigeraPolicy(cfg *GuardianConfiguration) (*v3.NetworkPolicy, e
...
+				Destination: guardianIngressDestinationEntityRule,
+			},
+			{
+				Action:      v3.Allow,
+				Protocol:    &networkpolicy.TCPProtocol,
+				Destination: guardianIngressDestinationEntityRule,
+			},
+		}...)
+	} else {
+		ingressRules = append(ingressRules, []v3.Rule{
```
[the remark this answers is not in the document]
Brian-McM: No we don't, I put this in when I mistakenly thought we did.

pkg/render/whisker/component.go
Brian-McM: I'm of the mind that this should panic if we're require this to be set. If there was an issue creating this then then controller needs to degrade.

pkg/render/goldmane/component.go
Brian-McM: Are we adding net policies? I'm not opposed to it, it's just we don't have any for anything other than the apiserver.

pkg/render/whisker/component.go
caseydavenport: A panic is not going to cause the controller to degrade though.

I think it would be fine if you want to change the Component interface to return an `error` if it is missing required information, and bubble that up to the status manager. But that sounds like out of scope for this project, so I'd say we should follow the pattern we use elsewhere?

pkg/controller/clusterconnection/clusterconnection_controller.go
Brian-McM: Does this comment belong here, for applying the defaults, or down below where I took out the check.

caseydavenport: it may not belong anywhere, I didn't see this logic but if it already exists then we're all good! I just threw it somewhere close to where I thought it should be.

pkg/render/goldmane/component.go
caseydavenport: Yeah, in our discussion with Ben and Anthony we agreed to this.

pkg/render/whisker/component.go
Brian-McM: But this is the pattern we follow, for things that must be configured. This is nothing other than a developer error (and panicing on developer errors is good), since it is created by the controller and if it's not then the user can't do anything about it.

pkg/controller/clusterconnection/clusterconnection_controller.go
Brian-McM: I updated the code to use the create the trusted bundle with the system cert since that's what it was doing before.

pkg/render/whisker/component.go
caseydavenport: Yeah, I guess we are inconsistent in this. Sometimes we check nil, and sometimes we don't.

Panicking on developer errors is good, but it's much better to explicitly do that than to implicitly.

Like, maybe the constructor should have this:

```go
if cfg.WhickerBackendKeyPair == nil {
  logrus.Fatal("BUG: Missing required keypair %s")
}
```

caseydavenport: but fine if you just want to leave it as is, since we don't do this anywhere else.

pkg/controller/clusterconnection/clusterconnection_controller.go
Brian-McM: also added the check to not allow public if this is calico.

pkg/render/guardian.go
```diff
@@ -558,3 +675,16 @@ func ProcessPodProxies(podProxies []*httpproxy.Config) []*httpproxy.Config {
...
 	return podProxies
 }
+
+func GuardianService(clusterDomain string) string {
+	return fmt.Sprintf("https://%s.%s.svc.%s:%d", GuardianServiceName, GuardianNamespace, clusterDomain, 443)
+}
+
+func deprecatedObjects() []client.Object {
+	return []client.Object{
+		&corev1.Namespace{
```
caseydavenport, line 685:
```suggestion
                // All the Guardian objects were moved to "calico-system" circa Calico v3.30, and so the legacy tigera-guardian Namespace and
                // everything within it should be removed.
		&corev1.Namespace{
```

pkg/controller/clusterconnection/clusterconnection_controller.go
```diff
@@ -231,111 +224,95 @@ func (r *ReconcileConnection) Reconcile(ctx context.Context, request reconcile.R
...
-	// Copy the secret from the operator namespace to the guardian namespace if it is present.
-	tunnelSecret := &corev1.Secret{}
-	err = r.Client.Get(ctx, types.NamespacedName{Name: render.GuardianSecretName, Namespace: common.OperatorNamespace()}, tunnelSecret)
-	if err != nil {
-		r.status.SetDegraded(operatorv1.ResourceReadError, "Error retrieving secrets from guardian namespace", err, reqLogger)
-		if !k8serrors.IsNotFound(err) {
-			return result, nil
+	includeSystem := false
+	if managementClusterConnection.Spec.TLS.CA == operatorv1.CATypePublic {
+		if variant == operatorv1.TigeraSecureEnterprise {
```
caseydavenport, line 268:
```suggestion
		if variant == operatorv1.Calico {
```

pkg/controller/clusterconnection/clusterconnection_controller.go
```diff
@@ -231,111 +224,95 @@ func (r *ReconcileConnection) Reconcile(ctx context.Context, request reconcile.R
...
+			r.status.SetDegraded(operatorv1.InvalidConfigurationError, "Guardian CA cannot be public in Calico.", nil, reqLogger)
```
caseydavenport, line 269:
```suggestion
			r.status.SetDegraded(operatorv1.InvalidConfigurationError, "ManagementClusterConnection CA cannot be public in Calico.", nil, reqLogger)
```

pkg/render/goldmane/component.go
```diff
@@ -0,0 +1,320 @@
...
+	GoldmaneServiceAccountName = GoldmaneName
+	GoldmaneDeploymentName     = GoldmaneName
+	GoldmaneRoleName           = GoldmaneName
+	GoldmaneServicePort        = 7443
+	GoldmaneContainerName      = "goldmane"
+
+	GoldmaneKeyPairSecret = "goldmane-key-pair"
+	GoldmaneServiceName   = "goldmane"
+
+	GuardianTargetPort     = 8080
```
caseydavenport, line 52: Do we not want to import these from the Guardian render package?

pkg/render/goldmane/component.go
```diff
@@ -0,0 +1,320 @@
...
+			Namespace: GoldmaneNamespace,
+		},
+		Rules: policyRules,
+	}
+}
+
+func (c *Component) deploymentSelector() *metav1.LabelSelector {
+	return &metav1.LabelSelector{
+		MatchLabels: map[string]string{
+			"app.kubernetes.io/name": GoldmaneDeploymentName,
```
caseydavenport, line 262: Ahh, finally breaking the seal on "k8s-app"

pkg/render/goldmane/component.go
```diff
@@ -0,0 +1,320 @@
...
+						Protocol: ptr.ToPtr(corev1.ProtocolTCP),
+						Port:     ptr.ToPtr(intstr.FromInt32(GoldmaneServicePort)),
+					}},
+				},
+			},
+			Egress: []netv1.NetworkPolicyEgressRule{
+				{
+					To: []netv1.NetworkPolicyPeer{
+						{
+							PodSelector: &metav1.LabelSelector{
```
caseydavenport, line 286: Do we want a common LabelSelector function for this? Looks awfully similar to `deploymentSelector()`

Not required for this PR, just a thought.

pkg/render/goldmane/component.go
```diff
@@ -0,0 +1,320 @@
...
+								MatchLabels: map[string]string{
+									"app.kubernetes.io/name": GuardianDeploymentName,
+								},
+							},
+							NamespaceSelector: &metav1.LabelSelector{
```
caseydavenport, line 291: I don't think we need the NamespaceSelector clause here - these pods are in the same Namespace already.

pkg/render/goldmane/component.go
```diff
@@ -0,0 +1,320 @@
...
+					},
+				},
+				{
+					Ports: []netv1.NetworkPolicyPort{
+						{
+							Protocol: ptr.ToPtr(corev1.ProtocolUDP),
+							Port:     ptr.ToPtr(intstr.FromInt32(53)),
+						},
+						{
+							Protocol: ptr.ToPtr(corev1.ProtocolTCP),
```
caseydavenport, line 312: What is this port for? 

In general I think it's good practice to add a comment to each rule explaining its intended purpose, for posterity.

pkg/render/goldmane/component.go
```diff
@@ -0,0 +1,320 @@
...
+							Port:     ptr.ToPtr(intstr.FromInt32(GuardianTargetPort)),
+						},
+					},
+				},
+				{
+					Ports: []netv1.NetworkPolicyPort{
+						{
+							Protocol: ptr.ToPtr(corev1.ProtocolUDP),
```
caseydavenport, line 308: In theory (but unlikely in practice) we could also see DNS traffic on TCP/53. Worth adding a rule for that

pkg/render/guardian.go
```diff
@@ -207,20 +220,38 @@ func (c *GuardianComponent) serviceAccount() *corev1.ServiceAccount {
...
+				"tiers",
+				"stagednetworkpolicies",
+				"tier.stagednetworkpolicies",
+				"networkpolicies",
+				"tier.networkpolicies",
+				"globalnetworkpolicies",
+				"tier.globalnetworkpolicies",
+				"globalnetworksets",
+			},
+			Verbs: []string{"get", "list", "watch"},
```
caseydavenport, line 254: I think OSS is going to need create / update on network policy if connected to a management cluster.

pkg/render/guardian.go
```diff
@@ -311,20 +342,47 @@ func (c *GuardianComponent) volumes() []corev1.Volume {
...
+		volumes = append(volumes, c.cfg.GuardianClientKeyPair.Volume())
+	}
+	return volumes
 }
 
 func (c *GuardianComponent) container() []corev1.Container {
 	envVars := []corev1.EnvVar{
 		{Name: "GUARDIAN_PORT", Value: "9443"},
-		{Name: "GUARDIAN_LOGLEVEL", Value: "INFO"},
+		{Name: "GUARDIAN_LOGLEVEL", Value: "DEBUG"},
```
caseydavenport, line 354: Can revert this one before merge

pkg/render/guardian.go
```diff
@@ -368,6 +430,58 @@ func (c *GuardianComponent) annotations() map[string]string {
 	return annotations
 }
 
+func (c *GuardianComponent) deploymentSelector() *metav1.LabelSelector {
+	return &metav1.LabelSelector{
+		MatchLabels: map[string]string{
+			"app.kubernetes.io/name": GuardianDeploymentName,
```
caseydavenport, line 436: Maybe a constant for the app.kubernetes.io/name label is appropriate if we're using it in a bunch of places?

pkg/render/guardian.go
```diff
@@ -368,6 +430,58 @@ func (c *GuardianComponent) annotations() map[string]string {
...
+			Ingress: []netv1.NetworkPolicyIngressRule{
+				{
+					From: []netv1.NetworkPolicyPeer{
+						{
+							PodSelector: &metav1.LabelSelector{
+								MatchLabels: map[string]string{
+									"app.kubernetes.io/name": GoldmaneDeploymentName,
+								},
+							},
+							NamespaceSelector: &metav1.LabelSelector{
```
caseydavenport, line 462: Same here for namespace selector - we can get rid of it when they are in the same namespace.

pkg/render/guardian.go
```diff
@@ -368,6 +430,58 @@ func (c *GuardianComponent) annotations() map[string]string {
...
+
+const (
+	GoldmaneDeploymentName = "goldmane"
+	GoldmaneNamespace      = common.CalicoNamespace
+)
+
+func (c *GuardianComponent) networkPolicy() *netv1.NetworkPolicy {
+	return &netv1.NetworkPolicy{
+		TypeMeta:   metav1.TypeMeta{Kind: "NetworkPolicy", APIVersion: "networking.k8s.io/v1"},
+		ObjectMeta: metav1.ObjectMeta{Name: "allow-whisker", Namespace: GuardianNamespace},
```
caseydavenport, line 449:
```suggestion
		ObjectMeta: metav1.ObjectMeta{Name: "guardian", Namespace: GuardianNamespace},
```

pkg/render/goldmane/component.go
```diff
@@ -0,0 +1,320 @@
...
+		MatchLabels: map[string]string{
+			"app.kubernetes.io/name": GoldmaneDeploymentName,
+		},
+	}
+}
+
+func (c *Component) networkPolicy() *netv1.NetworkPolicy {
+	return &netv1.NetworkPolicy{
+		TypeMeta:   metav1.TypeMeta{Kind: "NetworkPolicy", APIVersion: "networking.k8s.io/v1"},
+		ObjectMeta: metav1.ObjectMeta{Name: "allow-goldmane", Namespace: GoldmaneNamespace},
```
caseydavenport, line 270:
```suggestion
		ObjectMeta: metav1.ObjectMeta{Name: "goldmane", Namespace: GoldmaneNamespace},
```

pkg/render/goldmane/component.go
Brian-McM: Oh I guess we could do that... I started with goldmane and whisker and if I was to import the values from each other it would be circular dependency. Forgot guardian is in the render package.

pkg/render/guardian.go
Brian-McM: I'm not sure if we're adding that by default, I had thought the permissions were going to be explicitly by the user.

Let's chat offline with the policy dudes and I'll follow up with that.

pkg/render/goldmane/component.go
Brian-McM: Really? Is this for something other than DNS lookups?

I don't think we add DNS rules for tcp on 53 for other policies, unless I'm mistaken.

pkg/render/goldmane/component.go
Brian-McM: k8s API I'll add a comment.

caseydavenport: I think a port won't be enough in this case necessarily, since the API server can run on different ports?

You may want to use a Calico network policy, which would allow using a service selector like we do elsewhere: https://github.com/tigera/operator/blob/cd83253b701673b4536c0d1190a134593cde67eb/pkg/render/common/networkpolicy/networkpolicy.go#L204-L209

pkg/render/goldmane/component.go
caseydavenport: Yep, we do actually do this for other policies as well.

https://github.com/tigera/operator/blob/cd83253b701673b4536c0d1190a134593cde67eb/pkg/render/common/networkpolicy/networkpolicy.go#L44-L79

caseydavenport: Looks like we only do it for OCP for some reason, but I think it applies to regular k8s DNS as well....

pkg/render/guardian.go
caseydavenport: Yeah I think you're right, I forgot.

pkg/render/goldmane/component.go
Brian-McM: Is this compatible? More wondering why aren't we using a calico net policy for the apiserver?

I suppose I could always watch the k8s service and pull the port from there to add it (might prefer that if you don't mind).

Brian-McM: Chatted offline, we're taking out the egress rules altogether because it's a bit more complex to actually start using v3 policies since we can't rely on the api server being there in calico.

pkg/render/goldmane/component.go
Brian-McM: Done.

pkg/render/goldmane/component.go
Brian-McM: Added it to the function I created for this.

pkg/controller/goldmane/controller.go
```diff
@@ -0,0 +1,233 @@
...
+
+// Reconcile reads that state of the cluster for a Goldmane object and makes changes based on the
+// state read and what is in the Goldmane.Spec. The Controller will requeue the Request to be
+// processed again if the returned error is non-nil or Result.Requeue is true, otherwise upon completion it will
+// remove the work from the queue.
+func (r *Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
+	reqLogger := log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
+	reqLogger.Info("Reconciling Goldmane")
+
+	variant, installation, err := utils.GetInstallation(ctx, r.cli)
```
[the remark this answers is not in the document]
caseydavenport: We'll want to short-circuit as well if the installation is `nil` 

This also applies to the Whisker controller IIRC

pkg/controller/whisker/controller.go
```diff
@@ -223,49 +188,32 @@ func (r *Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
...
 
-	// Guardian needs a certificate for mTLS with Goldmane.
-	// TODO: Add this to the trusted bundle. This isn't stritctly needed, since the bundle already includes the operator CA that
-	// signed this certificate. But in order to support custom user-supplied certificates, we will need to do this.
-	guardianCertificateNames := dns.GetServiceDNSNames("guardian", whisker.WhiskerNamespace, r.clusterDomain)
-	guardianCertificateNames = append(guardianCertificateNames, "localhost", "127.0.0.1")
-	guardianKeyPair, err := certificateManager.GetOrCreateKeyPair(r.cli, whisker.GuardianKeyPairSecret, whisker.WhiskerNamespace, guardianCertificateNames)
+	trustedBundle, err := certificateManager.CreateNamedTrustedBundleFromSecrets(whisker.WhiskerDeploymentName, r.cli,
+		common.OperatorNamespace(), false,
+		goldmane.GoldmaneKeyPairSecret)
```
caseydavenport, line 193:
```suggestion
	trustedBundle, err := certificateManager.CreateNamedTrustedBundleFromSecrets(
	    whisker.WhiskerDeploymentName, 
	    r.cli,
	    common.OperatorNamespace(), 
	    false,
	    goldmane.GoldmaneKeyPairSecret,
	)
```

If we're splitting across multiple lines we gotta go all the wayyy

pkg/render/goldmane/component.go
```diff
@@ -0,0 +1,280 @@
...
+		{Name: "CA_CERT_PATH", Value: c.cfg.TrustedCertBundle.MountPath()},
+	}
+	guardianSvc := render.GuardianService(c.cfg.ClusterDomain)
+
+	volumeMounts := []corev1.VolumeMount{c.cfg.GoldmaneServerKeyPair.VolumeMount(c.SupportedOSType())}
+	volumeMounts = append(volumeMounts, c.cfg.TrustedCertBundle.VolumeMounts(c.SupportedOSType())...)
+
+	if c.cfg.ManagementClusterConnection != nil {
+		env = append(env,
+			corev1.EnvVar{Name: "PUSH_URL", Value: fmt.Sprintf("%s/api/v1/flows/bulk", guardianSvc)},
```
caseydavenport, line 141: Maybe not for this PR, but we need to figure out how we're going to dynamically learn this information.

Should we always set this env var and have Goldmane determine when to push based on some other indicator?  Should we use a ConfigMap to mount in configuration and use a file watcher?

pkg/render/goldmane/component.go
```diff
@@ -0,0 +1,280 @@
...
+	tolerations := append(c.cfg.Installation.ControlPlaneTolerations, rmeta.TolerateCriticalAddonsAndControlPlane...)
+	if c.cfg.Installation.KubernetesProvider.IsGKE() {
+		tolerations = append(tolerations, rmeta.TolerateGKEARM64NoSchedule)
+	}
+
+	ctrs := []corev1.Container{c.goldmaneContainer()}
+	volumes := []corev1.Volume{c.cfg.GoldmaneServerKeyPair.Volume(), c.cfg.TrustedCertBundle.Volume()}
+
+	// Add an annotation for the key pair as it requires a server restart (may as well restart the pod). Don't add an
+	// annotation for the mount CA since it's used for a client that can pick up the changes be recreating the client.
```
caseydavenport, line 180:
```suggestion
	// annotation for the mount CA since it's used for a client that can pick up the changes without a pod restart.
```

pkg/render/goldmane/component_test.go
```diff
@@ -0,0 +1,135 @@
...
+		),
+	)
+
+	DescribeTable("Goldmane Deployment", func(cfg *goldmane.Configuration, expected *appsv1.Deployment) {
+		component := goldmane.Goldmane(cfg)
+		objsToCreate, _ := component.Objects()
+
+		deployment, err := rtest.GetResourceOfType[*appsv1.Deployment](objsToCreate, goldmane.GoldmaneName, goldmane.GoldmaneNamespace)
+		Expect(err).ShouldNot(HaveOccurred())
+		Expect(deployment).To(Equal(expected))
```
caseydavenport, line 77: A bit of a nit, but the existing tests you've written for Whisker for example are exceptionally tricky to debug when they fail because the error message is so huge - do you have some tip for how to do that?

It would be really nice if we could factor these in a way where the precise diff was presented in the error message, and if we could do that before getting too deep into this testing pattern (most of our other tests check fields individually rather than a full equality on two fully specified Deployments)

pkg/render/guardian_test.go
```diff
@@ -49,6 +49,7 @@ var _ = Describe("Rendering tests", func() {
 	var resources []client.Object
 
 	createGuardianConfig := func(i operatorv1.InstallationSpec, addr string, openshift bool) *render.GuardianConfiguration {
+		i.Variant = operatorv1.TigeraSecureEnterprise
```
caseydavenport, line 52: Seems odd that we are arbitrarily settings this to enterprise here - why not in the place that passes the Installation to this function?

pkg/controller/whisker/controller.go
Brian-McM: I personally don't agree with it should be all on one line or every parameter should be on it's own line.

I parse large blocks of code better when it's not somewhat needlessly spread out over more lines (this adds up).

I'll go with your precendence on this, but this does not make it easier for me to read.

pkg/render/goldmane/component_test.go
Brian-McM: Ya the way I do it is add the cmp.Diff function, and add the output to the comment. Maybe I should just keep those cmp.Diff calls in here and commit it.

pkg/controller/whisker/controller.go
caseydavenport: Yeah I think our internal coding standards are not very clear about this. My editor will fix this for you next time I open the file though. I find inconsistent spacing to be very hard to read - two variables on one line and one variable the next and then two again? I'm going to miss something.

pkg/controller/goldmane/controller.go
```diff
@@ -0,0 +1,233 @@
...
+	for _, secretName := range []string{
+		goldmane.GoldmaneKeyPairSecret,
+		certificatemanagement.CASecretName,
+	} {
+		if err = utils.AddSecretsWatch(c, secretName, common.OperatorNamespace()); err != nil {
+			return fmt.Errorf("failed to add watch for secret %s/%s: %w", common.OperatorNamespace(), secretName, err)
+		}
+	}
+
+	if err = utils.AddConfigMapWatch(c, certificatemanagement.TrustedCertConfigMapName, common.OperatorNamespace(), &handler.EnqueueRequestForObject{}); err != nil {
```
rene-dekker, line 78: The name is incorrect, since you deviated from the default.

pkg/render/goldmane/component_test.go
caseydavenport: yeah, it's a no-no if I need to edit the test in order to understand its output. Why not just commit the good stuff?

pkg/render/guardian_test.go
Brian-McM: This is to keep these current tests running since it was expected that all of them are for EE.

caseydavenport: Yeah I know that but why is it jammed in here and not wherever the rest of the `InstallationSpec` that is given is generated? What if we want to run any of these tests against a different Variant now that Guardian is deployed for Calico as well?

pkg/render/goldmane/component_test.go
Brian-McM: I've just been doing it way to long to tests I don't want to commit changes for and it's become a bad habit to not commit it lol.

I'll commit it.

Brian-McM: done

pkg/controller/certificatemanager/certificatemanager.go
```diff
@@ -602,6 +619,10 @@ func (cm *certificateManager) CreateMultiTenantTrustedBundleWithSystemRootCertif
 	return certificatemanagement.CreateMultiTenantTrustedBundleWithSystemRootCertificates(cm.keyPair, certificates...)
 }
 
+func (cm *certificateManager) LoadNamedTrustedBundle(ctx context.Context, client client.Client, name, ns string) (certificatemanagement.TrustedBundleRO, error) {
```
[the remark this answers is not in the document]
Brian-McM: fixed

pkg/render/common/networkpolicy/k8snetworkpolicy.go
```diff
@@ -0,0 +1,83 @@
...
+						},
+						NamespaceSelector: &metav1.LabelSelector{
+							MatchLabels: map[string]string{
+								selector.CalicoPodNamespaceLabel: "kube-system",
+							},
+						},
+					},
+				},
+				Ports: []netv1.NetworkPolicyPort{
+					NewK8sPolicyPort(corev1.ProtocolTCP, 53),
```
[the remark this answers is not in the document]
Brian-McM: done

pkg/controller/clusterconnection/clusterconnection_controller.go
```diff
@@ -99,6 +99,7 @@ func Add(mgr manager.Manager, opts options.AddOptions) error {
 	for _, secretName := range []string{
 		render.PacketCaptureServerCert,
 		monitor.PrometheusServerTLSSecretName,
+		"guardian" + certificatemanagement.TrustedCertConfigMapSuffix,
```
caseydavenport, line 102: It's worth having a helper for this so we can ensure the same value is always used.

```
func CertBundleName("guardian") string {
}
```

would be used here as well as in the certificatemanagement code when generating the bundle.

pkg/controller/goldmane/controller.go
```diff
@@ -75,7 +75,7 @@ func Add(mgr manager.Manager, opts options.AddOptions) error {
 		}
 	}
 
-	if err = utils.AddConfigMapWatch(c, certificatemanagement.TrustedCertConfigMapName, common.OperatorNamespace(), &handler.EnqueueRequestForObject{}); err != nil {
+	if err = utils.AddConfigMapWatch(c, "goldmane"+certificatemanagement.TrustedCertConfigMapSuffix, common.OperatorNamespace(), &handler.EnqueueRequestForObject{}); err != nil {
```
caseydavenport, line 78: Same here

pkg/controller/whisker/controller.go
```diff
@@ -84,7 +82,7 @@ func Add(mgr manager.Manager, opts options.AddOptions) error {
 		}
 	}
 
-	if err = utils.AddConfigMapWatch(c, certificatemanagement.TrustedCertConfigMapName, common.OperatorNamespace(), &handler.EnqueueRequestForObject{}); err != nil {
+	if err = utils.AddConfigMapWatch(c, "whisker"+certificatemanagement.TrustedCertConfigMapSuffix, common.OperatorNamespace(), &handler.EnqueueRequestForObject{}); err != nil {
```
caseydavenport, line 85: and here

api/v1/goldmane_types.go
```diff
@@ -0,0 +1,53 @@
...
+
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
+// +kubebuilder:resource:scope=Cluster
+
+type Goldmane struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Status GoldmaneStatus `json:"status,omitempty"`
```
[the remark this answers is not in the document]
caseydavenport: We agree this would be OK as an immediate follow-up in order to reduce the scope of this PR.

caseydavenport (approved):
A few minor comments about name generation but otherwise this LGTM. Don't need to re-review after those are addressed.

pkg/controller/clusterconnection/clusterconnection_controller.go
Brian-McM: done.

PR #3818, opened 2025-03-18
Add overrides to goldmane and whisker

Brian-McM:
## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


pkg/render/goldmane/overrides_test.go
```diff
@@ -0,0 +1,120 @@
+// Copyright (c) 2025 Tigera, Inc. All rights reserved.
```
caseydavenport, line 1: I don't think we use a separate file for overrides in any of the other packages - I'd prefer to keep these within component_test.go myself.

I understand it's probably because you used code generation to build this, which clearly makes it easier if it's in a separate file, but linkage between X.go -> X_test.go is a nice property to maintain.

caseydavenport (approved):
One comment about test file structure that I'd really prefer to see addressed. But LGTM.

PR #3820, opened 2025-03-18
Add goldmane configmap volume + mount

caseydavenport:
## Description

Goes with https://github.com/projectcalico/calico/pull/9995

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/goldmane/component.go
+++ b/pkg/render/goldmane/component.go
@@ -171,9 +171,6 @@ func (c *Component) goldmaneContainer() corev1.Container {
 		MountPath: GoldmaneConfigFilePath,
 	})
 
-	if c.cfg.ManagementClusterConnection != nil {
-	}
-
 	return corev1.Container{
 		Name:            GoldmaneContainerName,
 		Image:           c.goldmaneImage,
--- a/pkg/render/goldmane/component.go
+++ b/pkg/render/goldmane/component.go
@@ -19,7 +19,6 @@ import (
 	"fmt"
 	"path/filepath"
 
-	"github.com/sirupsen/logrus"
 	"github.com/tigera/operator/pkg/common"
 	"github.com/tigera/operator/pkg/components"
 	"github.com/tigera/operator/pkg/ptr"
@@ -104,7 +103,7 @@ func (c *Component) Objects() ([]client.Object, []client.Object) {
 		c.serviceAccount(),
 		c.role(),
 		c.roleBinding(),
-		c.configMap(),
+		c.hotReloadConfigMap(),
 		c.goldmaneService(),
 		c.deployment(),
 		c.networkPolicy(),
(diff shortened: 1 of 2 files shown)


pkg/render/goldmane/component.go
```diff
@@ -15,8 +15,11 @@
 package goldmane
 
 import (
+	"encoding/json"
 	"fmt"
+	"path/filepath"
 
+	"github.com/sirupsen/logrus"
```
Brian-McM, line 22: Were you intending to introduce logrus to the operator?

pkg/render/goldmane/component.go
```diff
@@ -123,23 +131,48 @@ func (c *Component) serviceAccount() *corev1.ServiceAccount {
 	}
 }
 
+// configMap returns a ConfigMap containing configuration for Goldmane that does not require a restart of the pod.
+// This is mounted as a file within the Pod, which can be dynamically reloaded when changed.
+func (c *Component) configMap() *corev1.ConfigMap {
```
Brian-McM, line 136: Nit, might be nice to name this dynamicConfigConfigMap or something like that (equally fine to leave it as well).

pkg/render/goldmane/component.go
```diff
@@ -123,23 +131,48 @@ func (c *Component) serviceAccount() *corev1.ServiceAccount {
...
+
 func (c *Component) goldmaneContainer() corev1.Container {
+	guardianSvc := render.GuardianService(c.cfg.ClusterDomain)
 	env := []corev1.EnvVar{
 		{Name: "LOG_LEVEL", Value: "INFO"},
 		{Name: "PORT", Value: fmt.Sprintf("%d", GoldmaneServicePort)},
 		{Name: "SERVER_CERT_PATH", Value: c.cfg.GoldmaneServerKeyPair.VolumeMountCertificateFilePath()},
 		{Name: "SERVER_KEY_PATH", Value: c.cfg.GoldmaneServerKeyPair.VolumeMountKeyFilePath()},
 		{Name: "CA_CERT_PATH", Value: c.cfg.TrustedCertBundle.MountPath()},
+		{Name: "PUSH_URL", Value: fmt.Sprintf("%s/api/v1/flows/bulk", guardianSvc)},
```
Brian-McM, line 163: I know this isn't likely to change, but this also isn't something that needs to force a restart of goldmane so we could add it to the config map.

caseydavenport: We don't support changing this without a restart in Goldmane today - we could add that capability, but would require a small refactor and I don't think is something we really need at this point?

pkg/render/goldmane/component.go
caseydavenport: nope

pkg/render/goldmane/component.go
Brian-McM: No we don't need to do it, but this is configuring the client and you basically added the ability to reload the configuration for the client if it changes. So it seems less intuitive to me that some config changes would force a restart for the emitter and some won't.

I've no problem leaving this as is.

caseydavenport: I don't have a reason to implement this - it's not something we need. If it's easy to do, we can implement it if the need ever arises. Let's leave it as is.

PR #3845, opened 2025-03-28
Rename tigera-guardian components to guardian

Brian-McM:
## Description

Since tigera-guardian has been moved to the calico-system namespace we've decided to remove the tigera prefix since it's redundant

The container name needs to stay the same so the overrides API doesn't break.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

## Release Note

```release-note
The tigera-guardian deployment and resources previously in the tigera-guardian namespace have been moved to the calico-system namespace. The `tigera-` prefix has been remove for all namespaced resources and replaced with `calico-` for all cluster resources.

Custom policies or RBAC written for these moved components will need to account for the namespace and name changes.
```

### DIFF
--- a/pkg/render/guardian.go
+++ b/pkg/render/guardian.go
@@ -52,18 +52,22 @@ import (
 
 // The names of the components related to the Guardian related rendered objects.
 const (
-	GuardianName                   = "tigera-guardian"
+	GuardianName                   = "guardian"
 	GuardianNamespace              = common.CalicoNamespace
 	GuardianServiceAccountName     = GuardianName
-	GuardianClusterRoleName        = GuardianName
-	GuardianClusterRoleBindingName = GuardianName
+	GuardianClusterRoleName        = "calico-guardian"
+	GuardianClusterRoleBindingName = "calico-guardian"
 	GuardianDeploymentName         = GuardianName
-	GuardianServiceName            = "tigera-guardian"
-	GuardianVolumeName             = "tigera-guardian-certs"
-	GuardianSecretName             = "tigera-managed-cluster-connection"
-	GuardianTargetPort             = 8080
-	GuardianPolicyName             = networkpolicy.TigeraComponentPolicyPrefix + "guardian-access"
-	GuardianKeyPairSecret          = "guardian-key-pair"
+
+	// GuardianContainerName name is the name of the container running guardian. It's named `tigera-guardian`, instead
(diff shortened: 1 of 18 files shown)


pkg/render/guardian_test.go
```diff
@@ -180,7 +180,7 @@ var _ = Describe("Rendering tests", func() {
 		Expect(component.ResolveImages(nil)).To(BeNil())
 		resources, _ := component.Objects()
 
-		role := rtest.GetResource(resources, "tigera-guardian", "", "rbac.authorization.k8s.io", "v1", "ClusterRole").(*rbacv1.ClusterRole)
+		role := rtest.GetResource(resources, "guardian", "", "rbac.authorization.k8s.io", "v1", "ClusterRole").(*rbacv1.ClusterRole)
```
rene-dekker, line 183: I think global scoped resoures should keep the prefix.

Brian-McM: Thanks for the spot, updated it to be "calico-" instead (as per the conversations outside of the PR).

pkg/render/guardian.go
```diff
@@ -52,18 +52,22 @@ import (
...
-	GuardianTargetPort             = 8080
-	GuardianPolicyName             = networkpolicy.TigeraComponentPolicyPrefix + "guardian-access"
-	GuardianKeyPairSecret          = "guardian-key-pair"
+
+	// GuardianContainerName name is the name of the container running guardian. It's named `tigera-guardian`, instead
+	// of `guardian` so that the API for the container overrides don't have to change (`tigera-guardian` is a legacy name).
+	GuardianContainerName = "tigera-guardian"
+	GuardianServiceName   = "guardian"
+	GuardianVolumeName    = "guardian-certs"
+	GuardianSecretName    = "tigera-managed-cluster-connection"
```
caseydavenport, line 67: Any reason this can't be `guardian-managed-cluster-connection` or `calico-managed-cluster-connection` or just `managed-cluster-connection`?

caseydavenport:
@rene-dekker @Brian-McM is there something currently blocking this one?

rene-dekker (approved):

pkg/render/guardian.go
caseydavenport: Not asking you to do this now, but seems like it should probably be a SecretRef on the ManagedClusterConnection resource rather than hard-coded.

PR #3849, opened 2025-03-31
Remove dependency on authentication_controller while reconciling moni…

rene-dekker:


pkg/controller/monitor/monitor_controller.go
```diff
@@ -326,15 +327,13 @@ func (r *ReconcileMonitor) Reconcile(ctx context.Context, request reconcile.Requ
...
-	if authenticationCR != nil && authenticationCR.Status.State != operatorv1.TigeraStatusReady {
-		r.status.SetDegraded(operatorv1.ResourceNotReady, fmt.Sprintf("Authentication is not ready - authenticationCR status: %s", authenticationCR.Status.State), err, reqLogger)
-		return reconcile.Result{}, nil
-	}
-
-	keyValidatorConfig, err := utils.GetKeyValidatorConfig(ctx, r.client, authenticationCR, r.clusterDomain)
-	if err != nil {
-		r.status.SetDegraded(operatorv1.ResourceUpdateError, "Failed to process the authentication CR.", err, reqLogger)
-		return reconcile.Result{}, err
+	var keyValidatorConfig rauth.KeyValidatorConfig
```
Brian-McM, line 330: Just wanted to double check, if this isn't set it's not going to break anything further on is it (like in render)?

rene-dekker: All usages are behind a nil check. This is also nil when authentication CR is not used at all (like in our quickstart guide).

PR #3852, opened 2025-04-02
Enhance GatewayAPI for LB type and additional configurability

nelljerram:
## Release Note
```release-note
The GatewayAPI CR now allows controlling the type and properties of the external load balancer that is provisioned for each Gateway.  It also has enhanced configurability in other ways, including the number of replicas provisioned for each Gateway, or whether to provision a DaemonSet instead of a Deployment.
```


test/gatewayapi_test.go
```diff
@@ -0,0 +1,371 @@
...
+	})
+
+	It("watches custom EnvoyProxy resources", func() {
+		By("Creating Installation")
+		instance := &operator.Installation{
+			TypeMeta:   metav1.TypeMeta{Kind: "Installation", APIVersion: "operator.tigera.io/v1"},
+			ObjectMeta: metav1.ObjectMeta{Name: "default"},
+			Spec: operator.InstallationSpec{
+				Registry: "myregistry.io/",
+				Variant:  operator.TigeraSecureEnterprise,
```
gantony, line 236: We probably want a corresponding test with `Variant: operator.Calico,` to verify that we implement the expected behavior in OSS (no support for custom EnvoyProxy)

nelljerram: Yes, sorry, I forgot about making sure that these enhancements only work in Enterprise.  What I propose to do is:

1. Add a comment for each of the new fields in gatewayapi_types.go, to say that they are honoured only in Calico Enterprise.
2. Code a firewall near the top of the Reconcile function, to set a degraded status and return error if the variant is OSS and any of the new fields are being used.

api/v1/gatewayapi_types.go
```diff
@@ -24,13 +24,49 @@ import (
...
+	//
+	// 2. It configures the `tigera/envoy-gateway` and `tigera/envoy-ratelimit` images that will
+	// be used (according to the current Calico version, private registry and image set
+	// settings) and any pull secrets that are needed to pull those images.
+	//
+	// 3. It enables use of the Backend API.
+	//
+	// The resulting EnvoyGateway is provisioned as the `envoy-gateway-config` ConfigMap (which
+	// the gateway controller then uses as its config).
+	EnvoyGatewayRef *v1.ObjectReference `json:"envoyGatewayRef,omitempty"`
```
caseydavenport: Maybe `EnvoyGatewayConfigRef`? It's a reference to a configmap, not an EnvoyGateway (which as it seems isn't a real CR)

api/v1/gatewayapi_types.go
```diff
@@ -24,13 +24,49 @@ import (
 
 // GatewayAPISpec has fields that can be used to customize our GatewayAPI support.
 type GatewayAPISpec struct {
+	// Reference to a custom EnvoyGateway YAML to use as the base EnvoyGateway configuration for
+	// the gateway controller.  When specified, only the `name` and `namespace` fields of the
```
caseydavenport: >  When specified, only the `name` and `namespace` fields of the

We shouldn't include any fields in our API that are not respected - it's just confusing and requires us writing extra documentation / API comments to clarify it. Better to define our own ObjectReference for use here than to use an object with extra options in it.

api/v1/gatewayapi_types.go
```diff
@@ -24,13 +24,49 @@ import (
...
+	// ObjectReference are significant, and they must identify an existing ConfigMap resource
```
caseydavenport: > and they must identify an existing ConfigMap resource

What happens if the ConfigMap doesn't exist at the time this resource is created? Presumably it's still OK to create these in any order (e.g., GatewayAPI first and ConfigMap second?)

api/v1/gatewayapi_types.go
```diff
@@ -24,13 +24,49 @@ import (
...
+	// with an "envoy-gateway.yaml" key whose value is the desired EnvoyGateway YAML
+	// (i.e. following the same pattern as the default `envoy-gateway-config` ConfigMap).
+	// Please note, it is not possible to create an EnvoyGateway resource outside of a ConfigMap
```
caseydavenport: Is it important to note this? I'd just remove this bit of the comment - it's long enough as it is, and we don't need to explain all of the design decisions as part of the user-facing API documentation.

api/v1/gatewayapi_types.go
```diff
@@ -48,6 +84,42 @@ type GatewayAPISpec struct {
 	CRDManagement *CRDManagement `json:"crdManagement,omitempty"`
 }
 
+type GatewayClassSpec struct {
+	// The name of this GatewayClass.
+	Name string `json:"name,omitempty"`
+
+	// Reference to a custom EnvoyProxy resource to use as the base EnvoyProxy configuration for
+	// this GatewayClass.  When specified, only the `name` and `namespace` fields of the
```
caseydavenport: Yeah, especially if we're having to write out this qualifier in multiple places, a stripped down ObjectRef with only the fields we care about seems in order.

api/v1/gatewayapi_types.go
```diff
@@ -48,6 +84,42 @@ type GatewayAPISpec struct {
...
+	// a GatewayClass that references it.
+	//
+	// If a custom EnvoyProxy resource is specified and uses `EnvoyDaemonSet` instead of the
+	// default `EnvoyDeployment`, deployment-related customizations will be applied within
+	// `EnvoyDaemonSet` instead of within `EnvoyDeployment`.
+	EnvoyProxyRef *v1.ObjectReference `json:"envoyProxyRef,omitempty"`
+
+	// Allow class-specific customization of gateway deployments (or daemonsets) and services,
+	// for Gateways in this GatewayClass.
+	GatewayDeployment *GatewayDeployment `json:"gatewayDeployment,omitempty"`
```
caseydavenport:
```suggestion
	// Allow class-specific customization of gateway deployments.
	GatewayDeployment *GatewayDeployment `json:"gatewayDeployment,omitempty"`
	
	// Allow class-specific customization of this gateway's service.
	GatewayService *GatewayService `json:"gatewayService,omitempty"`

```

GatewayDeployment should only configure the Gateway deployment - if there's a daemonset or Service those should get their own fields.

The intention of the `XDeployment` pattern is that it exactly models the upstream Deployment specification - something that users already know and understand. We shouldn't be adding any new fields to those structures that would diverge them from upstream.

api/v1/gatewayapi_types.go
```diff
@@ -86,11 +158,17 @@ type GatewayControllerDeployment struct {
 
 // GatewayControllerDeploymentSpec allows customization of the gateway controller deployment spec.
 //
+// If GatewayControllerDeployment.Spec.Replicas is non-nil it customizes the number of replicas for
```
caseydavenport: We don't need to put the per-field documentation at the top-level of the structure - these comments should be attached to the particular fields that they are documenting.

That's how all of the automated tooling for generating CRDs, documentation, etc expect the comments to look.

api/v1/gatewayapi_types.go
```diff
@@ -236,23 +323,46 @@ type GatewayCertgenJobContainer struct {
...
 	// +optional
 	Template *GatewayDeploymentPodTemplate `json:"template,omitempty"`
 
 	// The deployment strategy to use to replace existing pods with new ones.
 	// +optional
 	// +patchStrategy=retainKeys
 	Strategy *GatewayDeploymentStrategy `json:"strategy,omitempty" patchStrategy:"retainKeys" protobuf:"bytes,4,opt,name=strategy"`
+
+	// +optional
+	GatewayKind *GatewayKind `json:"gatewayKind,omitempty"`
```
caseydavenport: Doesn't really make sense for GatewayDeploymentSpec to include a field that allows it to not be a deployment :grin:

pkg/controller/gatewayapi/gatewayapi_controller.go
```diff
@@ -192,20 +194,109 @@ func (r *ReconcileGatewayAPI) Reconcile(ctx context.Context, request reconcile.R
...
+			}
+			if err = r.watchEnvoyProxy(namespacedName); err != nil {
+				r.status.SetDegraded(operatorv1.ResourceReadError, "Error watching EnvoyProxyRef", err, log)
+				return reconcile.Result{}, err
+			}
+			gatewayConfig.CustomEnvoyProxies[gatewayAPI.Spec.GatewayClasses[i].Name] = envoyProxy
+		}
+	}
+
+	// Enumerate existing GatewayClasses, in case some of them will need to be cleaned up.
```
caseydavenport, line 341: Might be worth a small comment explaining why these need explicit cleanup instead of getting GC'd by the kube garbage collector (I think it's because they are not-namespaced?)

caseydavenport (commented):
Mostly just API feedback, the controller / render changes largely look good to me!

pkg/controller/gatewayapi/gatewayapi_controller.go
nelljerram: I am not familiar with the kube garbage collector - what could that do for us here?

But yes, you are right that GatewayClasses are not namespaced.

caseydavenport: https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/#owner-references-in-object-specifications

> Cluster-scoped dependents can only specify cluster-scoped owners. In v1.20+, if a cluster-scoped dependent specifies a namespaced kind as an owner, it is treated as having an unresolvable owner reference, and is not able to be garbage collected.

I think I was thinking of this, which might not apply here?

Within kube-controller-manager there is logic to track owner references and delete children when they reference an owner that no longer exists. We rely on this throughout the operator to clean up objects - for example, even if the tigera/operator has been deleted, owner references can be used to ensure that deleting a CR deletes the objects created for that CR.

api/v1/gatewayapi_types.go
nelljerram: Yes, I agree, this one is a bit weird.  Is it something that we could live with as is, or does this need to change?

If it needs to change, I think there are two sub-questions:
1. Do we define GatewayDaemonSet as a copy of GatewayDeployment, and recursively copy all its substructures?  (Except for Replicas)  That does feel a little tedious to me, but maybe we have to do it?
2. Suppose no customizations are wanted at all _inside_ those structures, but a customer still wants to indicate using a DaemonSet instead of a Deployment.  How should the API allow that?

api/v1/gatewayapi_types.go
nelljerram: The tooling works as is too, and I think [it looks OK](https://docs.tigera.io/calico-enterprise/3.21/reference/installation/api#operator.tigera.io/v1.GatewayDeployment).  However, I'm happy to look at updating this if you strongly prefer.

api/v1/gatewayapi_types.go
nelljerram: Makes sense, will do.

api/v1/gatewayapi_types.go
nelljerram: Fine, will do.  (Sometimes less is more!)

api/v1/gatewayapi_types.go
nelljerram: > What happens if the ConfigMap doesn't exist at the time this resource is created?

As the coding stands, the GatewayAPI gets a degraded status and things don't deploy.

> Presumably it's still OK to create these in any order (e.g., GatewayAPI first and ConfigMap second?)

No, and I'm not sure how to make that work because how could we watch a ConfigMap that doesn't exist yet?  (Or perhaps we can do that, just by its name?)

api/v1/gatewayapi_types.go
nelljerram: Do you happen to know if we already have a suitable struct?

api/v1/gatewayapi_types.go
caseydavenport: I think it's OK if the result is "We temporarily go degraded while we wait for the referenced object to show up" so long as the API doesn't reject the GatewayAPI CR up front - i.e., loose coupling, similar to how you can create a Pod that refers to a ConfigMap but it won't get deployed until the configmap exists.  

I guess I was trying to understand the emphasis on "Existing" configmap here. I think it just threw me off a bit. I'd probably drop that qualification (I think it's evident / implied that an object reference should refer to a real object, and the existing implied a required ordering of operations on the users behalf).

pkg/controller/gatewayapi/gatewayapi_controller.go
nelljerram: That would apply if the GatewayAPI CR is deleted.  But here I'm thinking of a scenario like [this](https://github.com/tigera/operator/pull/3852/files#diff-e861930aea9561dfd81058d294281a85a4a5f3e5a62ff7959f735e110e05ea8aR187-R224) where the content inside Spec.GatewayClasses is being updated.

api/v1/gatewayapi_types.go
caseydavenport: I don't think it works the way I expect it to (and the way we use it for all the other APIs).

image

The description section is blank, with just "(optional)" present. 

Similarly, the CRD is missing a description for those fields: https://github.com/tigera/operator/blob/5124369c2b5fb4a430adee88f1781ffd776c8c74/pkg/crds/operator/operator.tigera.io_gatewayapis.yaml#L3575-L3578

As opposed to topology spread constraints: https://github.com/tigera/operator/blob/5124369c2b5fb4a430adee88f1781ffd776c8c74/pkg/crds/operator/operator.tigera.io_gatewayapis.yaml#L3617-L3621

caseydavenport: And one last point, this shows up in `kubectl explain` output, which is hard to parse:

```
kubectl explain gatewayapis.spec.gatewayDeployment.spec.template.spec
KIND:     GatewayAPI
VERSION:  operator.tigera.io/v1

RESOURCE: spec <Object>

DESCRIPTION:
     GatewayDeploymentPodSpec allows customization of the pod spec of gateway
     deployments. If GatewayDeployment.Spec.Template.Spec.Affinity is non-nil,
     it sets the affinity field of each deployment's pod template. If
     GatewayDeployment.Spec.Template.Spec.NodeSelector is non-nil, it sets a
     node selector for where gateway pods may be scheduled. If
     GatewayDeployment.Spec.Template.Spec.Tolerations is non-nil, it sets the
     tolerations field of each deployment's pod template. If
     GatewayDeployment.Spec.Template.Spec.TopologySpreadConstraints is non-nil,
     it sets the topology spread constraints of each deployment's pod template.
     For customization of container resources see
     GatewayControllerDeploymentContainer.

FIELDS:
   affinity     <Object>
     Affinity is a group of affinity scheduling rules.

   containers   <[]Object>

   nodeSelector <map[string]string>

   tolerations  <[]Object>

   topologySpreadConstraints    <[]Object>
     TopologySpreadConstraints describes how a group of pods ought to spread
     across topology domains. Scheduler will schedule pods in a way which abides
     by the constraints. All topologySpreadConstraints are ANDed.
```

api/v1/gatewayapi_types.go
caseydavenport: I took a look around but didn't find one, so probably need a new one here.

api/v1/gatewayapi_types.go
caseydavenport: I think we do want separate GatewayDeployment and GatewayDaemonset fields (and GatewayService). A bit tedious, but the right thing for the end user.

I would have expected a field within the GatewayAPISpec that configures the mode, rather than within the overrides structures. 

e.g.,

```yaml
spec:
  kind: DaemonSet
  gatewayDaemonSet: { . . .}
```

And probably operator validation to ensure you don't have `kind: DaemonSet` set with `gatewayDeployment != nil` (and vice-versa).

nelljerram:
Docs PR is here: https://github.com/tigera/docs/pull/2102

api/v1/gatewayapi_types.go
nelljerram: OK, I will do these updates, but will maybe defer to a following PR in the hope of completing all the other markups today...

pkg/controller/gatewayapi/gatewayapi_controller.go
caseydavenport: Ah yep, that makes sense!

api/v1/gatewayapi_types.go
```diff
@@ -45,9 +97,63 @@ type GatewayAPISpec struct {
...
+	// a GatewayClass that references it.
+	//
+	// If a custom EnvoyProxy resource is specified and uses `EnvoyDaemonSet` instead of the
+	// default `EnvoyDeployment`, deployment-related customizations will be applied within
+	// `EnvoyDaemonSet` instead of within `EnvoyDeployment`.
+	// +optional
+	EnvoyProxyRef *NamespacedName `json:"envoyProxyRef,omitempty"`
+
+	// Specifies whether Gateways in this class are deployed as Deployments (default) or as
+	// DaemonSets.  Note, GatewayKind is ignored when a custom EnvoyProxy is specified and that
```
caseydavenport: We try quite hard to avoid having fields that are ever simply ignored - I would expect that either:

- We have a single way of specifying this so there is no opportunity for conflicting user configuration.
- We return an error if the user provides configuration that is not self-consistent.

api/v1/gatewayapi_types.go
```diff
@@ -24,15 +24,67 @@ import (
...
+	// 3. It enables use of the Backend API.
+	//
+	// The resulting EnvoyGateway is provisioned as the `envoy-gateway-config` ConfigMap (which
+	// the gateway controller then uses as its config).
+	// +optional
+	EnvoyGatewayConfigRef *NamespacedName `json:"envoyGatewayConfigRef,omitempty"`
+
+	// Allow customization of the GatewayClasses that will be available; please see
+	// GatewayClassSpec for more detail.  If GatewayClasses is nil, the Tigera operator
+	// configures a single GatewayClass named "tigera-gateway-class" without any of the
```
caseydavenport: > If GatewayClasses is nil, the Tigera operator configures a single GatewayClass named "tigera-gateway-class"

Regarding this point, what should the user expect if:

- They create GatewayAPI without any classes
- Add their own classes to this spec
- Remove their ow nclasses from this spec.

Would the `tigera-gateway-class` be created, removed, and then created again? And what would happen to any gateways using the default class when it is removed?

Similarly, for this flow:

- Create a GatewayAPI with user-specified classes.
- Remove those classes
- Add another class

I think the behavior - at least from this comment - is not obvious. Normally the pattern we follow is to fill in default values for fields and write them back to the APi explicitly - so if a user creates a GatewayAPI without any gateway classes specified, we would fill in the field ourselves to make sure the API resource is consistent with what is being programmed. That's probably what we should do here as well.

pkg/render/gateway_api.go
```diff
@@ -494,47 +527,168 @@ func (pr *gatewayAPIImplementationComponent) Objects() ([]client.Object, []clien
...
 
-	// Provision a GatewayClass that references the EnvoyProxy config and the controllerName
-	// that the gateway controller expects.
-	objs = append(objs, pr.gatewayClass(envoyGatewayConfig.Gateway.ControllerName, proxyConfig))
+	currentGCNames := set.New[string]()
+	currentEPNames := set.New[string]()
 
-	return objs, nil
-}
+	// Provision the default "tigera-gateway-class" GatewayClass.
```
caseydavenport: As I touched on in an earlier comment, this logic should probably be in the API defaulting rather than in the render code. i.e., if there are no gateway classes in the API, append this default to spec.GatewayClasses and pass it to the render code. 

This pattern is used extensively in the operator API resources and ensures that the state of the cluster is preserved in the API resource itself, and isn't relying on code level defaults that may change across versions. It also makes sure that the API resource behaves as expected, and consistently with others.

The only quirk is that for user-provided ones, we assert the referenced configmap exists, whereas for our own we provision the configmap ourselves _after_ that validation step. I think this is easily circumvented in a number of possible ways, though.

pkg/render/gateway_api.go
```diff
@@ -427,27 +432,55 @@ func (pr *gatewayAPIImplementationComponent) Objects() ([]client.Object, []clien
...
-	objs = append(objs, proxyConfig)
-
-	// Substitute possibly modified image names into the envoy-gateway-config.
+	// Prepare EnvoyGateway, either from upstream or from a custom EnvoyGatewayConfigRef
+	// provided by the user.
 	envoyGatewayConfig := resources.envoyGatewayConfig.DeepCopyObject().(*envoyapi.EnvoyGateway)
-	envoyGatewayConfig.Provider.Kubernetes.RateLimitDeployment.Container.Image = &pr.envoyRatelimitImage
-	envoyGatewayConfig.Provider.Kubernetes.ShutdownManager.Image = &pr.envoyGatewayImage
+	if pr.cfg.CustomEnvoyGateway != nil {
+		// In this case we don't need to worry about modification because the next Reconcile
```
caseydavenport: I don't quite understand this comment - the render code shouldn't ever need to care about future reconciles - It's job is just to take input and convert it to k8s resources in a deterministic and declarative way.

What is the modification we're (not) worried about here?

api/v1/gatewayapi_types.go
```diff
@@ -24,15 +24,67 @@ import (
...
+	// Allow customization of the gateway certgen job.
+	// +optional
 	GatewayCertgenJob *GatewayCertgenJob `json:"gatewayCertgenJob,omitempty"`
 
-	// Allow optional customization of gateway deployments.
+	// Allow customization of gateways when deployed as Kubernetes Deployments.  These
+	// customizations will apply to all of the GatewayClasses that the Tigera operator
+	// provisions.  GatewayClass-specific customizations can be specified in
+	// `GatewayClasses[*].GatewayDeployment`.
+	// +optional
```
caseydavenport: A bit surprised to see both per-class and global overrides here. Could you talk me through the use-cases for this? 

At a minimum, the API documentation doesn't explain how these interact (from the code it seems like the per-class overrides override the global overrides?) - but is this really necessary?

Seems like a per-class set of overrides (in conjunction with writing back our "default" class to the API as part of defaulting) would cover all the bases? But maybe I am missing something.

api/v1/gatewayapi_types.go
```diff
@@ -192,58 +283,54 @@ type GatewayCertgenJobPodTemplate struct {
...
+	// If non-nil, Resources sets the ResourceRequirements of the job's "envoy-gateway-certgen"
+	// container.
 	// +optional
 	Resources *v1.ResourceRequirements `json:"resources,omitempty"`
 }
 
-// GatewayDeployment allows customization of gateway deployments.
-//
-// For detail see GatewayDeploymentSpec.
+type GatewayKind string
```
caseydavenport, line 291:
```suggestion
// +kubebuilder:validation:Enum=Deployment;DaemonSet
type GatewayKind string
```

pkg/controller/gatewayapi/gatewayapi_controller.go
```diff
@@ -137,18 +178,28 @@ func (r *ReconcileGatewayAPI) Reconcile(ctx context.Context, request reconcile.R
 		return reconcile.Result{}, nil
 	}
 
-	pullSecrets, err := utils.GetNetworkingPullSecrets(installation, r.client)
-	if err != nil {
-		r.status.SetDegraded(operatorv1.ResourceReadError, "Error retrieving pull secrets", err, reqLogger)
-		return reconcile.Result{}, err
+	if variant == operatorv1.Calico {
+		reqLogger.Info("Variant is Calico")
```
caseydavenport: This info log is going to be too spammy - we should remove it or drop its verbosity to V(2) or higher.

caseydavenport: Although I am still counting on removing this branch entirely. None of the fields checked by this branch should be enterprise only.

pkg/render/gateway_api.go
```diff
@@ -548,3 +702,45 @@ func (pr *gatewayAPIImplementationComponent) gatewayClass(controllerName string,
...
+	GetConfig() *GatewayAPIImplementationConfig
+}
+
+func (pr *gatewayAPIImplementationComponent) GetConfig() *GatewayAPIImplementationConfig {
+	return pr.cfg
+}
+
+// ApplyEnvoyProxyServiceOverrides applies the overrides to the given EnvoyProxy.
+// Note: overrides must not be nil pointer.
+func ApplyEnvoyProxyServiceOverrides(ep *envoyapi.EnvoyProxy, overrides *operatorv1.GatewayService) {
```
caseydavenport: Think this should be a private function (seems only used in this file)

api/v1/gatewayapi_types.go
nelljerram: OK, I think it's fine to change this to follow your second bullet here.

api/v1/gatewayapi_types.go
nelljerram: > > If GatewayClasses is nil, the Tigera operator configures a single GatewayClass named "tigera-gateway-class"
> 
> Regarding this point, what should the user expect if:
> 
> * They create GatewayAPI without any classes
> * Add their own classes to this spec
> * Remove their ow nclasses from this spec.
> 
> Would the `tigera-gateway-class` be created, removed, and then created again?

Yes.

> And what would happen to any gateways using the default class when it is removed?

I think this is a matter for the Envoy implementation.  I can see two ways this might work:

1. The class is only queried when a gateway is created, so the ongoing existence of the class does not matter.
2. Envoy places a finalizer on the class to prevent it from completing deletion while gateways in that class still exist.

I'd also say that customers ought to delete associated gateways before they change GatewayClasses to remove a class.  We could document that if you think we need to - but I'm not sure it even needs to be said.

> 
> Similarly, for this flow:
> 
> * Create a GatewayAPI with user-specified classes.
> * Remove those classes
> * Add another class
> 
> I think the behavior - at least from this comment - is not obvious. Normally the pattern we follow is to fill in default values for fields and write them back to the APi explicitly - so if a user creates a GatewayAPI without any gateway classes specified, we would fill in the field ourselves to make sure the API resource is consistent with what is being programmed. That's probably what we should do here as well.

I see your point, but I think we should leave this as is for this PR, given that:

1. We're trying to get this done for 3.22ep1.
2. This is the way it already worked in 3.21, so a change would be incompatible.
3. This is mixed up with the discussion about what ends up being enforced as Enterprise-only, and with the possible idea about moving gateway class provisioning to kube-controllers.  (In case you haven't seen it, this is firstly because that could be a more natural home for add-on function that differs between OSS and Enterprise, and secondly because it synergises with the work to provide a helm install for our basic gateway API support.)

I have a sync with Ash about (3) this afternoon.

pkg/render/gateway_api.go
nelljerram: As above - except that I didn't follow your detail about a referenced configmap.

pkg/render/gateway_api.go
nelljerram: When `pr.cfg.CustomEnvoyGateway` is nil, the base envoy-gateway-config comes from `resources.envoyGatewayConfig`, which is populated at start of day from the embedded YAML.  We don't want to risk changes to `resources.envoyGatewayConfig` persisting incorrectly between calls to `Objects`, hence the `DeepCopyObject()` on the previous line.

But in the `pr.cfg.CustomEnvoyGateway != nil` case that does not apply.

If you think it's helpful I can paste all of that into the comment.

api/v1/gatewayapi_types.go
nelljerram: > A bit surprised to see both per-class and global overrides here. Could you talk me through the use-cases for this?

Isn't it fairly obvious?  There might be some tweaks - say a label? - that a user will want on all their classes?
Also it preserves compatibility with 3.21, in which only the global override setting was available.

> At a minimum, the API documentation doesn't explain how these interact (from the code it seems like the per-class overrides override the global overrides?)

It does - please see https://github.com/tigera/operator/pull/3852/files#diff-3116f79c8aeaa96e71f6368da7ecd3d1f357b2ffa19dbf3f0e99d2d085cc26e7R116-R127

> - but is this really necessary?
> 
> Seems like a per-class set of overrides (in conjunction with writing back our "default" class to the API as part of defaulting) would cover all the bases? But maybe I am missing something.

Technically yes, but at the cost of the user having to duplicate parts of their config.  So this is just a convenience and compatibility thing.

pkg/controller/gatewayapi/gatewayapi_controller.go
nelljerram: Sorry, that log was for debugging only, and I meant to remove it.

api/v1/gatewayapi_types.go
caseydavenport: I am fine to leave this out of this PR - I do think we should still do it for the release though.

I don't think this is incompatible with previous versions - it's just encoding the existing behavior into the API. Prior to this increment, we didn't have an API field with which to represent our default GatewayClass behavior. Now that we do, we should use it.

api/v1/gatewayapi_types.go
caseydavenport: > Isn't it fairly obvious? There might be some tweaks - say a label? - that a user will want on all their classes?

I don't think this is something we need to implement ourselves. The user could do one of the following:
- Add that label to all of their classes. They need to define them each anyway.
- Use a mutating webhook (or one of the many ecosystem projects implemented on top of webhooks, like OPA) to automatically apply the label to classes as they are created.

I don't think we need to handle this with a shortcut in our API - unless we have explicit demands from users with legitimate reasons why the above options aren't acceptable. Otherwise, it's better to have a single way to do this.

> Also it preserves compatibility with 3.21, in which only the global override setting was available.

It does more than that though - it introduces additional fields that were not present in v3.21. I would think that, in conjunction with my comments above re: encoding the default class into the API, we would deprecate the v3.21 field entirely in favor of the new per-class configuration.

> It does - please see https://github.com/tigera/operator/pull/3852/files#diff-3116f79c8aeaa96e71f6368da7ecd3d1f357b2ffa19dbf3f0e99d2d085cc26e7R116-R127

Whoops, sorry. I missed this.

pkg/render/gateway_api.go
caseydavenport: Ah, I understand. I think the missing bit for me was a comment along the lines of "We take a deepcopy of the default to make sure we don't accidentally modify the config loaded at start" on the line above.

api/v1/gatewayapi_types.go
nelljerram: OK, happy to do this, given how sure you feel about it.  We can also benefit here from the fact that Gateway API in 3.21 is described as tech preview, which I think means that some small API changes are still possible before GA.

pkg/render/gateway_api.go
caseydavenport: > I didn't follow your detail about a referenced configmap.

Normally, the operator would do something like this:

1. Update the CR with defaults for unspecified fields.
2. Validate the CR
3. Pass the CR to render/ to generate objects to create in the cluster.

For our case:
- (1) adds the `tigera-gateway-class` to the GatewayAPI, if none are present, referencing `envoy-proxy-config`.
- (2) makes sure any referenced EnvoyGatewayConfigRef configmap exists
- (3) creates GatewayClass, but also the `envoy-proxy-config` ConfigMap

So, I foresee step (2) failing because the required configmap exists only after step (3) has executed, but (3) won't execute if (2) fails. 

We could:

- Continue with (3), even if (2) fails.
- Split the logic from either (2) or (3) into multiple steps such that the ordering works out, ensuring the configmap we create happens before we check for its existence as part of validation.

nelljerram: I think the `tigera-gateway-class` write-back should have `EnvoyProxyRef` nil.  Then I think the above concern evaporates?

(This then triggers a stream of thoughts about whether this is an exception to the philosophy of writing back defaults to the CR.  But I don't think it is, because there isn't a representation of the EnvoyProxy content within the CR.  Also it occurs to me that we *don't* do the default write-back thing within all the override structures...)

pkg/render/gateway_api.go
nelljerram: FYI I've improved the coding and commenting here.

pkg/render/gateway_api.go
caseydavenport: >  Also it occurs to me that we don't do the default write-back thing within all the override structures.

Yeah, this was an intentional exception we made to this rule when we introduce the overrides pattern a while back. I am not sure it was the right one!

caseydavenport (approved):

PR #3866, opened 2025-04-04
Set explicit DNS nameservers for calico/node when needed

caseydavenport:
## Description

In eBPF mode (and theoretically any time KUBERNETES_SERVICE_HOST is a domain name instead of an IP), calico/node needs explicit nameserver configuration in order to resolve the address. Inherit this from the operator pod's configuration itself.

This is needed as a consequence of adjusting the DNS policy on calico/node to ClusterFirstWithHostNet, which was done as part of enabling DNS resolution of `goldmane` running as an in-cluster service.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -393,6 +393,7 @@ If a value other than 'all' is specified, the first CRD with a prefix of the spe
 	if err != nil {
 		log.Error(err, "Couldn't find the nameservers from the resolv.conf")
 	}
+	log.Infof("Found nameservers: %v", nameservers)
 
 	kubernetesVersion, err := common.GetKubernetesVersion(clientset)
 	if err != nil {
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -389,6 +389,11 @@ If a value other than 'all' is specified, the first CRD with a prefix of the spe
 		log.Error(err, fmt.Sprintf("Couldn't find the cluster domain from the resolv.conf, defaulting to %s", clusterDomain))
 	}
 
+	nameservers, err := dns.Nameservers(dns.DefaultResolveConfPath)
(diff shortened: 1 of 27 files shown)


pkg/dns/dns.go
```diff
@@ -59,6 +60,41 @@ func GetClusterDomain(resolvConfPath string) (string, error) {
 	return clusterDomain, nil
 }
 
+// IsDomainName checks if the given name is a domain name and returns truee if it is.
```
lucastigera, line 63: typo: **truee**

cmd/main.go
```diff
@@ -389,6 +389,11 @@ If a value other than 'all' is specified, the first CRD with a prefix of the spe
 		log.Error(err, fmt.Sprintf("Couldn't find the cluster domain from the resolv.conf, defaulting to %s", clusterDomain))
 	}
 
+	nameservers, err := dns.Nameservers(dns.DefaultResolveConfPath)
+	if err != nil {
+		log.Error(err, fmt.Sprintf("Couldn't find the nameservers from the resolv.conf, defaulting to %s", nameservers))
```
lucastigera, line 394: Should we log this as an Error, or would a Warning or Info be more appropriate in this situation?
Additionally, if `dns.Nameservers` returns an error, will `nameservers` be nil, or did I misunderstand the function?
If I understood correctly, does this message make sense? `( ... defaulting to %s", nameservers)`

caseydavenport: Yep, good catch. This is just a copy/paste error - we can't default this to anything reasonable AFAIK

Brian-McM (approved):

Issue #1301, opened 2021-05-13
Fixed CNI bin dir prevents nodes from working on Kubernetes deployment


Omar007:
It's not possible to configure the path to the CNI bin directory for the Calico deployment. Calico puts the files in `/opt/cni/bin` for Kubernetes deployments.  
If the cluster does not use this path, the files end up in the wrong location and as a result the nodes never become ready.

## Expected Behavior

The CNI bin directory can be specified on the operator deployment or Installation spec.

## Current Behavior

CNI bin directory can only ever be `/opt/cni/bin` and if a cluster uses anything else, it will fail.

#### Proper
Update the operator to expose configuration options to set the CNI bin directory path with (and the config directory wouldn't be a bad idea either tbh)

#### Workaround
a. Manually copy over the files to the correct location on each node in the cluster.
b. Patch Calico Node DaemonSet after the operator has deployed it to change the hostPath.

## Steps to Reproduce (for bugs)

1. Have a cluster where Kubelet and CRI-O use a CNI bin dir other than `/opt/cni/bin`
2. Deploy Calico on said Kubernetes cluster
3. See the Calico Node instances place 3 files in `/opt/cni/bin` on the nodes
4. See Kubelet CNI network configuration failure error and nodes never becoming ready as the required binaries aren't placed where they are needed.

## Your Environment

* Operating System and version: Linux LTS 5.10.36 (Arch Linux)
* Kubernetes Version: 1.21.0
* CRI-O Version: 1.21.0
* Calico/Operator Version: 3.19.0

tmjd:
Could you provide details on your configuration and the directories needed? It would be best if this was a detectable Provider and the operator could automatically detect the provider and use the appropriat bin dir.

Omar007:
It's a basic `kubeadm` cluster installation using the provided distro packages. These are configured/compiled to use `/usr/lib/cni` as the standard directory instead of `/opt/cni/bin`.  

As far as I'm aware Kubernetes does not expose this information. Additionally it's a flag/config option you supply to the Kubelet as well as CRI-O so it could even differ per node (but in most cases I don't think it's customizable to that degree on any operator and is assumed to at least be the same across nodes).

ananace:
As a note, I'd like to see this type of configurability too, though in my case being able to set the CNI network config directory is more important than the binary one - as I've got a few bare-metal clusters with multus on them, and therefore need the Calico CNI config to end up in `/etc/cni/multus/net.d`.

[...]

PR #3882, opened 2025-04-16
extends the Installation resource to allow specifying directories for installing CNI binaries and configuration files

a-yohe1:
## Description

resolves https://github.com/tigera/operator/issues/1301, resolves https://github.com/projectcalico/calico/issues/7511

This PR extends the Installation resource to allow specifying directories for installing CNI binaries and configuration files.

Some container runtimes allow customization of CNI binaries and configurations, and we believe it is valuable to allow modification of their installation locations. e.g., [containerd](https://github.com/containerd/containerd/blob/main/docs/cri/config.md)
Please also refer to this comment:
https://github.com/projectcalico/calico/issues/7511#issuecomment-2788240330

In this PR, we have performed code-based E2E tests (ginkgo) and the following local verification.
Anyone can try the local verification by applying the same changes.

<details>

<summary>Local Environment Verification Results</summary>

A k3d cluster was built using the following diff.
Cluster construction follows the steps in the documentation:
https://github.com/tigera/operator/blob/master/docs/common_tasks.md

```diff
diff --git a/deploy/kind-config.yaml b/deploy/kind-config.yaml
index 69ace47d..247d95a1 100644
--- a/deploy/kind-config.yaml
+++ b/deploy/kind-config.yaml
@@ -25,3 +25,8 @@ kubeadmConfigPatches:
   mode: ipvs
   conntrack:
     maxPerCore: 0
+containerdConfigPatches:
+- |-
+  [plugins."io.containerd.grpc.v1.cri".cni]
+    bin_dir = '/opt/cni/custom_dir/bin'
+    conf_dir = '/etc/cni/custom_dir/net.d'
```

After building the cluster and running the operator with `go run`, the Installation resource was applied with the following diff.

```diff
diff --git a/config/samples/operator_v1_installation.yaml b/config/samples/operator_v1_installation.yaml
index 0a5876a7..9efd3ace 100644
--- a/config/samples/operator_v1_installation.yaml
+++ b/config/samples/operator_v1_installation.yaml
@@ -2,3 +2,8 @@ apiVersion: operator.tigera.io/v1
 kind: Installation
 metadata:
   name: default
+spec:
+  cni:
+    type: "Calico"
+    binDir: "/opt/cni/custom_dir/bin"
+    confDir: "/etc/cni/custom_dir/net.d"
```

Confirmed that Nodes and Pods become Ready.

```console
❯ k get nodes
NAME                 STATUS   ROLES           AGE     VERSION
kind-control-plane   Ready    control-plane   9m      v1.30.4
kind-worker          Ready    <none>          8m39s   v1.30.4
kind-worker2         Ready    <none>          8m39s   v1.30.4
kind-worker3         Ready    <none>          8m39s   v1.30.4


❯ k get ds -n calico-system calico-node -o yaml | yq -r '.spec.template.spec.volumes.[].hostPath'
{
  "path": "/opt/cni/custom_dir/bin",
  "type": "DirectoryOrCreate"
}
{
  "path": "/var/log/calico/cni",
  "type": ""
}
{
  "path": "/etc/cni/custom_dir/net.d",
  "type": ""
}
{
  "path": "/usr/libexec/kubernetes/kubelet-plugins/volume/exec/nodeagent~uds",
  "type": "DirectoryOrCreate"
}
{
  "path": "/lib/modules",
  "type": ""
}
null
{
  "path": "/var/run/nodeagent",
  "type": "DirectoryOrCreate"
}
null
{
  "path": "/var/lib/calico",
  "type": "DirectoryOrCreate"
}
{
  "path": "/var/run/calico",
  "type": "DirectoryOrCreate"
}
{
  "path": "/run/xtables.lock",
  "type": "FileOrCreate"
}


❯ kubectl debug  node/kind-worker -it --image=ubuntu
--profile=legacy is deprecated and will be removed in the future. It is recommended to explicitly specify a profile, for example "--profile=general".
Creating debugging pod node-debugger-kind-worker-n28sh with container debugger on node kind-worker.
If you don't see a command prompt, try pressing enter.
root@kind-worker:/# cd /ho
home/ host/
root@kind-worker:/# cd /host/
root@kind-worker:/host# ls
LICENSES  boot  etc   kind  lib64  mnt  proc  run   srv  tmp  var
bin       dev   home  lib   media  opt  root  sbin  sys  usr
root@kind-worker:/host# ls -ltr etc/cni/custom_dir/net.d/
total 8
-rw------- 1 root root  684 Apr 16 13:07 10-calico.conflist
-rw------- 1 root root 2801 Apr 16 13:08 calico-kubeconfig
root@kind-worker:/host# ls -ltr opt/cni/custom_dir/bin/
total 175120
-rwxr-xr-x 1 root root 79475334 Apr 16 13:07 calico
-rwxr-xr-x 1 root root 79475334 Apr 16 13:07 calico-ipam
-rwxr-xr-x 1 root root  2862657 Apr 16 13:07 flannel
-rwxr-xr-x 1 root root  4268262 Apr 16 13:07 host-local
-rwxr-xr-x 1 root root  4105115 Apr 16 13:07 loopback
-rwxr-xr-x 1 root root  4850594 Apr 16 13:07 portmap
-rwxr-xr-x 1 root root  4257111 Apr 16 13:07 tuning

```

</details>

This change affects [calico/node](https://docs.tigera.io/calico/latest/reference/configure-calico-node), which is responsible for installing CNI binaries and related files on nodes when Calico is selected as the plugin type.

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


a-yohe1:
Hi @caseydavenport 

Could you please take a look at this pull request?
Your feedback would be greatly appreciated!

caseydavenport:
@a-yohe1 yes, thank you! I missed this one

api/v1/installation_types.go
```diff
@@ -926,6 +926,24 @@ type CNISpec struct {
...
 	IPAM *IPAMSpec `json:"ipam"`
+
+	// BinDir is the path to the CNI binaries directory.
+	// If you have changed the installation directory for CNI binaries in the container runtime configuration,
+	// please ensure that this field points to the same directory as specified in the container runtime settings.
+	// Default: /opt/cni/bin
+	// +optional
+	// +kubebuilder:default:="/opt/cni/bin"
+	// +kubebuilder:validation:Type=string
+	BinDir string `json:"binDir,omitempty"`
```
caseydavenport, line 937: Optional fields should be pointers in the API per conventions: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required


```suggestion
	BinDir *string `json:"binDir,omitempty"`
```

api/v1/installation_types.go
```diff
@@ -926,6 +926,24 @@ type CNISpec struct {
 	// Calico Enterprise installation.
 	// +optional
 	IPAM *IPAMSpec `json:"ipam"`
+
+	// BinDir is the path to the CNI binaries directory.
+	// If you have changed the installation directory for CNI binaries in the container runtime configuration,
+	// please ensure that this field points to the same directory as specified in the container runtime settings.
+	// Default: /opt/cni/bin
+	// +optional
+	// +kubebuilder:default:="/opt/cni/bin"
```
caseydavenport, line 935: I think setting a default here via kubebuilder is going to be a breaking change, since the default value we have used in the past has varied based on platform. 

I think we need to move `cniDirectoriesForProvider` out of `render` and into the core controller as part of `fillDefaults`

https://github.com/tigera/operator/blob/f370b45e84be9fa8d807f139612fc61e9079f451/pkg/controller/installation/core_controller.go#L440-L728

caseydavenport (changes requested):

api/v1/installation_types.go
a-yohe1: Fixed in the following commit.

https://github.com/tigera/operator/pull/3882/commits/fa7184a673b869063c01fba32e5d42d2614ff532

api/v1/installation_types.go
a-yohe1: Thank you. I unintentionally introduced a breaking change. As you pointed out, setting defaults via kubebuilder was incorrect. I've changed it to set defaults within fillDefaults in the following commit. Since node_test.go also has assertions for rendering results per provider, the default logic remains in node.go. This is configured similarly to the [Windows settings](https://github.com/tigera/operator/blob/master/pkg/controller/installation/core_controller.go#L547).

https://github.com/tigera/operator/pull/3882/commits/3f42e01f958db4384902b968fbab12fe2c2836fc

pkg/controller/installation/core_controller.go
```diff
@@ -632,6 +632,14 @@ func fillDefaults(instance *operator.Installation, currentPools *crdv1.IPPoolLis
 		instance.Spec.CalicoNetwork.LinuxPolicySetupTimeoutSeconds = &delay
 	}
 
+	defaultCniNetDir, defaultCniBinDir := render.DefaultCNIDirectories(instance.Spec.KubernetesProvider)
```
caseydavenport, line 635:
```suggestion
	defaultCNINetDir, defaultCNIBinDir := render.DefaultCNIDirectories(instance.Spec.KubernetesProvider)
```

Just a nit, but acronyms should be same-case in Golang variables.

caseydavenport (approved):
Just one small comment, but otherwise this LGTM! Thanks for doing this!

pkg/controller/installation/core_controller.go
a-yohe1: Thank you for the feedback. 
https://github.com/tigera/operator/pull/3882/commits/b354eb4920df016816579b77498fd86dc9c38946

a-yohe1:
Thank you. I forgot to run `make gen-files` after editing. I've added the generated files in the following commit. CI is pending, so there might still be things to address since I couldn't see the final result.

https://github.com/tigera/operator/pull/3882/commits/3270a6b4aba23117c5358a3c918c8a3a28aa70db

caseydavenport (approved):
LGTM!

PR #3886, opened 2025-04-16
Add whisker and goldmane finalizers

WilliamTigera:
## Description

Add finalizers based on the api server implementation 

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


pkg/controller/clusterconnection/clusterconnection_controller.go
```diff
@@ -461,3 +466,45 @@ func (r *ReconcileConnection) Reconcile(ctx context.Context, request reconcile.R
 	// We should create the Guardian deployment.
 	return result, nil
 }
+
+// maintainInstallationFinalizer manages this controller's finalizer on the Installation resource.
+// We add a finalizer to the Installation when Guardian has been installed, and only remove that finalizer when
+// the Guardian has been deleted and its pods have stopped running. This allows for a graceful cleanup of Guardian resources
+// prior to the CNI plugin being removed.
+func maintainInstallationFinalizer(ctx context.Context, c client.Client, mgmtClusterConnection *operatorv1.ManagementClusterConnection) error {
```
caseydavenport, line 474: Now that we're using this across many controllers, I think it is time we make this a common generic utility function rather than duplicate the code in each controller.

pkg/controller/clusterconnection/clusterconnection_controller.go
```diff
@@ -461,3 +466,45 @@ func (r *ReconcileConnection) Reconcile(ctx context.Context, request reconcile.R
...
+	// a finalizer. Otherwise, if the Whisker namespace doesn't exist we should remove it. This ensures the finalizer
+	// is always present so long as the resources managed by this controller exist in the cluster.
+	if mgmtClusterConnection != nil {
+		// Add a finalizer indicating that the Goldmane is still running.
+		utils.SetInstallationFinalizer(installation, render.GuardianFinalizer)
+	} else {
+		// Check if the Guardian namespace exists, and remove the finalizer if not. Gating this on Namespace removal
+		// in the best way to approximate that all Guardian related resources have been removed.
+		l := &corev1.Namespace{}
+		err := c.Get(ctx, types.NamespacedName{Name: render.GuardianNamespace}, l)
```
caseydavenport, line 497: This isn't going to work because there is no Guardian Namespace - it runs within calico-system, which is going to exist until all of this teardown logic is completed. 

We need to check the Guardian deployment itself.

pkg/controller/goldmane/controller.go
```diff
@@ -236,3 +244,45 @@ func (r *Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
...
+	// a finalizer. Otherwise, if the Whisker namespace doesn't exist we should remove it. This ensures the finalizer
+	// is always present so long as the resources managed by this controller exist in the cluster.
+	if goldmaneCr != nil {
+		// Add a finalizer indicating that the Goldmane is still running.
+		utils.SetInstallationFinalizer(installation, render.GoldmaneFinalizer)
+	} else {
+		// Check if the Goldmane namespace exists, and remove the finalizer if not. Gating this on Namespace removal
+		// in the best way to approximate that all Goldmane related resources have been removed.
+		l := &corev1.Namespace{}
+		err := c.Get(ctx, types.NamespacedName{Name: goldmane.GoldmaneNamespace}, l)
```
caseydavenport, line 275: Same here - there is no Goldmane namespace, it runs within `calico-system` so we can't use the same heuristic here.

pkg/controller/whisker/controller.go
```diff
@@ -268,3 +276,45 @@ func updateWhiskerWithDefaults(instance *operatorv1.Whisker) {
...
+	// a finalizer. Otherwise, if the Whisker namespace doesn't exist we should remove it. This ensures the finalizer
+	// is always present so long as the resources managed by this controller exist in the cluster.
+	if whiskerCr != nil {
+		// Add a finalizer indicating that the Whisker is still running.
+		utils.SetInstallationFinalizer(installation, render.WhiskerFinalizer)
+	} else {
+		// Check if the Whisker namespace exists, and remove the finalizer if not. Gating this on Namespace removal
+		// in the best way to approximate that all Whisker related resources have been removed.
+		l := &corev1.Namespace{}
+		err := c.Get(ctx, types.NamespacedName{Name: whisker.WhiskerNamespace}, l)
```
caseydavenport, line 307: and here

caseydavenport (changes requested):

pkg/controller/utils/utils.go
```diff
@@ -986,3 +987,49 @@ func RemoveInstallationFinalizer(i *operatorv1.Installation, finalizer string) {
 		i.SetFinalizers(stringsutil.RemoveStringInSlice(finalizer, i.GetFinalizers()))
 	}
 }
+
+// maintainInstallationFinalizer manages a controller's finalizer on the Installation resource.
+// We add a finalizer to the Installation when the mainResource has been installed, and only remove that finalizer when
+// the resource has been deleted and its secondary resources have stopped running. This allows for a graceful cleanup of any resources
+// prior to the CNI plugin being removed.
+func MaintainInstallationFinalizer(ctx context.Context, c client.Client, mainResource client.Object,
+	secondaryResource client.Object, finalizer string) error {
```
caseydavenport, line 996:
```suggestion
func MaintainInstallationFinalizer(
  ctx context.Context, 
  c client.Client, 
  mainResource client.Object,
  secondaryResource client.Object, 
  finalizer string,
) error {
```

If splitting across lines, please put each argument on its own line for legibility.

caseydavenport: I would probably suggest for allowing an arbitrary number of secondary resources, though. Some controllers do manage multiple deployments, for example. 

```go
func MaintainInstallationFinalizer(
  ctx context.Context, 
  c client.Client, 
  mainResource client.Object,
  finalizer string,
  objs ...client.Object, 
) error {

File: pkg/controller/goldmane/controller.go [hunk 6]
```diff
@@ -144,13 +146,15 @@ func (r *Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
...
 
+	goldmaneDeployment := &v1.Deployment{ObjectMeta: metav1.ObjectMeta{Namespace: common.CalicoNamespace, Name: goldmane.GoldmaneDeploymentName}}
 	goldmaneCR, err := utils.GetIfExists[operatorv1.Goldmane](ctx, utils.DefaultInstanceKey, r.cli)
 	if err != nil {
 		r.status.SetDegraded(operatorv1.ResourceReadError, "Error querying Goldmane CR", err, reqLogger)
 		return reconcile.Result{}, err
 	} else if goldmaneCR == nil {
 		r.status.OnCRNotFound()
-		return reconcile.Result{}, nil
+		return reconcile.Result{}, utils.MaintainInstallationFinalizer(ctx, r.cli, goldmaneCR,
```
caseydavenport, line 156: No need to split this line here. It just makes the code harder to read.

caseydavenport: I think if you just passed `nil` instead of `goldmaneCR` you wouldn't need the special `isNil` function?

pkg/controller/goldmane/controller.go
```diff
@@ -144,13 +146,15 @@ func (r *Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
 	reqLogger := log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
 	reqLogger.Info("Reconciling Goldmane")
 
+	goldmaneDeployment := &v1.Deployment{ObjectMeta: metav1.ObjectMeta{Namespace: common.CalicoNamespace, Name: goldmane.GoldmaneDeploymentName}}
```
caseydavenport, line 149: It would be nice to encapsulate this to make it a bit more clear.

```go
func (r *Reconciler) maintainFinalizer(ctx context.Context, gm client.Object) error {
  // These objects require graceful termination before the CNI plugin is torn down.
  goldmaneDeployment := &v1.Deployment{ObjectMeta: metav1.ObjectMeta{Namespace: common.CalicoNamespace, Name: goldmane.GoldmaneDeploymentName}}
  return utils.MaintainInstallationFinalizer(ctx, r.cli, gm, goldmaneDeployment, render.GoldmaneFinalizer)
```

This just keeps all the logic related to finalizing tidy and in one place.

pkg/controller/utils/utils.go
```diff
@@ -986,3 +987,49 @@ func RemoveInstallationFinalizer(i *operatorv1.Installation, finalizer string) {
 		i.SetFinalizers(stringsutil.RemoveStringInSlice(finalizer, i.GetFinalizers()))
 	}
 }
+
+// maintainInstallationFinalizer manages a controller's finalizer on the Installation resource.
```
caseydavenport, line 991:
```suggestion
// MaintainInstallationFinalizer manages a controller's finalizer on the Installation resource.
```

pkg/render/finalizers.go
```diff
@@ -27,4 +27,10 @@ const (
 	// InstallationControllerFinalizer is added to the Installation by the core Installation controller when installing Calico
 	// so that Calico CNI resources are not removed until calico-kube-controllers has had time to properly be torn down.
 	InstallationControllerFinalizer = "operator.tigera.io/installation-controller"
+
+	WhiskerFinalizer = "operator.tigera.io/whisker-controller"
```
caseydavenport, line 31: These should have equivalent level of commenting as the ones above.

caseydavenport (changes requested):
A few small suggestions

pkg/render/finalizers.go
```diff
@@ -27,4 +27,14 @@ const (
 	// InstallationControllerFinalizer is added to the Installation by the core Installation controller when installing Calico
 	// so that Calico CNI resources are not removed until calico-kube-controllers has had time to properly be torn down.
 	InstallationControllerFinalizer = "operator.tigera.io/installation-controller"
+
+	// WhiskerFinalizer is added to the Installation by the whisker controller when the whisker CR is created so that
+	// Calico CNI resources are not removed until the whisker controller has had time to properly delete the whisker deployment.
+	WhiskerFinalizer = "operator.tigera.io/whisker-controller"
+	// GuardianFinalizer is added to the Installation by the cluster connection controller when the management cluster connection CR
+	//  is created so that Calico CNI resources are not removed until the controller has had time to properly delete the guardian deployment.
+	GuardianFinalizer = "operator.tigera.io/guardian-controller"
```
caseydavenport, line 36:
```suggestion
	GuardianFinalizer = "operator.tigera.io/guardian-controller"
	
```

pkg/render/finalizers.go
```diff
@@ -27,4 +27,14 @@ const (
 	// InstallationControllerFinalizer is added to the Installation by the core Installation controller when installing Calico
 	// so that Calico CNI resources are not removed until calico-kube-controllers has had time to properly be torn down.
 	InstallationControllerFinalizer = "operator.tigera.io/installation-controller"
+
+	// WhiskerFinalizer is added to the Installation by the whisker controller when the whisker CR is created so that
+	// Calico CNI resources are not removed until the whisker controller has had time to properly delete the whisker deployment.
+	WhiskerFinalizer = "operator.tigera.io/whisker-controller"
```
caseydavenport, line 33:
```suggestion
	WhiskerFinalizer = "operator.tigera.io/whisker-controller"
	
```

pkg/controller/utils/utils.go
```diff
@@ -986,3 +986,53 @@ func RemoveInstallationFinalizer(i *operatorv1.Installation, finalizer string) {
...
+		// Keep track of all the secondary resources that the main resource creates.
+		// Only delete the finalizer if all of the secondary resources are deleted.
+		for _, secondaryResource := range secondaryResources {
+			err := c.Get(ctx, types.NamespacedName{Namespace: secondaryResource.GetNamespace(), Name: secondaryResource.GetName()}, secondaryResource)
+			if err != nil && !errors.IsNotFound(err) {
+				return err
+			} else if errors.IsNotFound(err) {
+				log.Info(fmt.Sprintf("Object %s in Namespace %s no longer exists", secondaryResource.GetName(), secondaryResource.GetNamespace()), "finalizer", finalizer)
+			} else {
+				log.Info(fmt.Sprintf("Object %s in Namespace %s is still present, waiting for termination", secondaryResource.GetName(), secondaryResource.GetNamespace()))
```
caseydavenport, line 1028: Not every object is going to have a namespace - throughout this function, the logging is going to be awkward if the namespace is an empty string.

We should handle that case explicitly - e.g.

```go
name := types.NamespacedName{Namespace: secondaryResource.GetNamespace(), Name: secondaryResource.GetName()}
```

Then you can use it in the logs:

```go
log.Info("%s is still present, waiting for termination", name)
```

pkg/controller/utils/utils.go
```diff
@@ -986,3 +986,53 @@ func RemoveInstallationFinalizer(i *operatorv1.Installation, finalizer string) {
 		i.SetFinalizers(stringsutil.RemoveStringInSlice(finalizer, i.GetFinalizers()))
 	}
 }
+
+// MaintainInstallationFinalizer manages a controller's finalizer on the Installation resource.
+// We add a finalizer to the Installation when the mainResource has been installed, and only remove that finalizer when
+// the resource has been deleted and its secondary resources have stopped running. This allows for a graceful cleanup of any resources
+// prior to the CNI plugin being removed.
+func MaintainInstallationFinalizer(
```
caseydavenport, line 994: We should update the apiserver and installation core controllers to use this helper function instead of the duplicate ones they have there.

caseydavenport (changes requested):

pkg/controller/utils/utils.go
```diff
@@ -986,3 +986,53 @@ func RemoveInstallationFinalizer(i *operatorv1.Installation, finalizer string) {
...
+	} else {
+		// Check if the namespaced secondaryResources are still present.
+		// Keep track of all the secondary resources that the main resource creates.
+		// Only delete the finalizer if all of the secondary resources are deleted.
+		for _, secondaryResource := range secondaryResources {
+			err := c.Get(ctx, types.NamespacedName{Namespace: secondaryResource.GetNamespace(), Name: secondaryResource.GetName()}, secondaryResource)
+			if err != nil && !errors.IsNotFound(err) {
+				return err
+			} else if errors.IsNotFound(err) {
+				log.Info("Object no longer exists.", secondaryResource.GetName(), secondaryResource)
```
caseydavenport, line 1026:
```suggestion
				log.Info("Object no longer exists.", "object", secondaryResource)
```

Intention of these loggers is to produce key=value pairs that can be searched - the name should already be present from `secondaryResource` so I don't think we need it in both the key and the value?

pkg/controller/utils/utils.go
```diff
@@ -986,3 +986,53 @@ func RemoveInstallationFinalizer(i *operatorv1.Installation, finalizer string) {
...
+			} else {
+				log.Info("Object is still present, waiting for termination", secondaryResource.GetName(), secondaryResource)
```
caseydavenport, line 1028:
```suggestion
				log.Info("Object is still present, waiting for termination", "object", secondaryResource)
```

Same here

caseydavenport:
Looks like tests are failing, though:

```
/go/src/github.com/tigera/operator/test/whisker_test.go:120
Expected
    <[]string | len:3, cap:4>: [
        "operator.tigera.io/installation-controller",
        "tigera.io/operator-cleanup",
        "operator.tigera.io/goldmane-controller",
    ]
to contain element matching
    <string>: operator.tigera.io/whisker-controller
/go/src/github.com/tigera/operator/test/whisker_test.go:143
```

cmd/main.go
```diff
@@ -554,7 +554,11 @@ func executePreDeleteHook(ctx context.Context, c client.Client) error {
...
 	apiserver.Name = utils.DefaultInstanceKey.Name
-	for _, o := range []client.Object{installation, apiserver} {
+	whisker := &operatortigeraiov1.Whisker{}
+	whisker.Name = utils.DefaultInstanceKey.Name
+	goldmane := &operatortigeraiov1.Goldmane{}
+	goldmane.Name = utils.DefaultInstanceKey.Name
+	for _, o := range []client.Object{whisker, goldmane, installation, apiserver} {
 		if err := c.Delete(ctx, o); err != nil {
 			if errors.IsNotFound(err) {
 				return nil
```
caseydavenport, line 564: I think this is actually a bug here in the existing code ? We shouldn't return right away if one of these objects is missing.

The only case we want to return is if the Installation itself is already deleted... but I think that will already be handled on line 579...

Maybe this should just be `continue` instead? WDYT?

WilliamTigera: Yeah I think you're right, good catch. It looked weird to me but I didn't question it :cry:

caseydavenport (approved):

PR #3910, opened 2025-04-29
Fix manager RBAC for mixed mode management clusters

caseydavenport:
## Description

We need a ClusterRole/Binding per Tenant class, rather than shared RBAC
for all.

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/utils/namespace_helper.go
+++ b/pkg/controller/utils/namespace_helper.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2023-2024 Tigera, Inc. All rights reserved.
+// Copyright (c) 2023-2025 Tigera, Inc. All rights reserved.
 
 // Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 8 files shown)


pkg/controller/utils/namespace_helper.go
```diff
@@ -52,6 +57,9 @@ type NamespaceHelper interface {
 	// For single-tenant clusters, this simply returns the InstallNamespace.
 	TenantNamespaces(client.Client) ([]string, error)
 
+	// FilteredTenantNamespaces returns all namespaces for all Tenants that have Calico OSS managed clusters.
```
tmjd, line 60: I don't think this comment is accurate.

caseydavenport: Whoops, yeah that's from an earlier iteration.

gcosgrave (approved):
LGTM

PR #3938, opened 2025-05-08
CORE-11022: Implementing TLS Cipher configurable for API Server

lucastigera:
## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/api/v1/installation_types.go
+++ b/api/v1/installation_types.go
@@ -152,6 +152,10 @@ type InstallationSpec struct {
 	// +optional
 	CertificateManagement *CertificateManagement `json:"certificateManagement,omitempty"`
 
+	// TLSCipherSuites defines the cipher suite list that the TLS protocol should use during secure communication.
+	// +optional
+	TLSCipherSuites TLSCipherSuites `json:"tlsCipherSuites,omitempty"`
+
 	// NonPrivileged configures Calico to be run in non-privileged containers as non-root users where possible.
 	// +optional
 	NonPrivileged *NonPrivilegedType `json:"nonPrivileged,omitempty"`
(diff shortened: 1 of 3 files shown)


api/v1/installation_types.go
```diff
@@ -208,6 +212,23 @@ type InstallationSpec struct {
 	Proxy *Proxy `json:"proxy,omitempty"`
 }
 
+type TLSCipherSuite struct {
+	// This should be a valid TLS cipher suite name.
+	// +optional
+	// +kubebuilder:validation:Enum=TLS_AES_256_GCM_SHA384;TLS_CHACHA20_POLY1305_SHA256;TLS_AES_128_GCM_SHA256;TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384;TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384;TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256;TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256;TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256;TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256;TLS_RSA_WITH_AES_256_GCM_SHA384;TLS_RSA_WITH_AES_128_GCM_SHA256;TLS_ECDHE_ECDSA_WITH_AES_256_CBC_SHA;TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA;TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA
+	Name string `json:"name"`
```
caseydavenport, line 219:
```suggestion
	Name TLSCipher  `json:"name"`
```

and then

```go
//+kubebuilder:validation:Enum=TLS_AES_256_GCM_SHA384;TLS_CHACHA20_POLY1305_SHA256;TLS_AES_128_GCM_SHA256;TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384;TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384;TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256;TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256;TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256;TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256;TLS_RSA_WITH_AES_256_GCM_SHA384;TLS_RSA_WITH_AES_128_GCM_SHA256;TLS_ECDHE_ECDSA_WITH_AES_256_CBC_SHA;TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA;TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA
type TLSCipher string

const (
 TLS_AES_256_GCM_SHA384 TLSCipher = "TLS_AES_256_GCM_SHA384"
)
```

This is normally how we do enum types - it provides type checking in the Go code so you can't accidentally set an unsupported "string" value.

lucastigera: Thanks for the comment! 
But what’s the necessity of creating those enums?
At first glance, having the Kubebuilder validation alone should suffice. IMO, the Operator can handle formatting the string and passing it to the components via environment variable.
I might be overlooking something, but so far I haven’t found a reason to define the enums explicitly, given that the Operator doesn’t need to use them directly. :thinking:

caseydavenport: It does a few things:

- Any code in the operator (e.g., unit tests, or switch statements) are required to use the constant, which avoids accidental typos when writing or comparing strings.
- This API is importable and supported for external use in Go code - if someone wants to configure the installation using Go code, having enum types provides a better development experience for them with type checking, and typo protection.
- Putting the kubebuilder validation on the enum type itself instead of the field means that wherever this type is used (e.g., if we use it in multiple places in the future), the same kubebuilder validation will be applied and we'll only need to specify it once.

lucastigera: Okay! I knew that I was overlooking something. :sweat_smile: 
Thanks for the explanation. Will crate the type and enums! :+1:

pkg/controller/utils/component.go
```diff
@@ -663,6 +669,47 @@ func setImagePullPolicy(podSpec *v1.PodSpec) {
...
+	switch obj := obj.(type) {
+	case *apps.Deployment:
+		containers = obj.Spec.Template.Spec.Containers
+	case *apps.DaemonSet:
+		containers = obj.Spec.Template.Spec.Containers
+	default:
+		return nil
+	}
+
+	_, installationSpec, err := GetInstallation(context.Background(), c)
```
Brian-McM, line 684: Could you please pass the context in from the calling createOrUpdateObject function and use that instead.

Brian-McM (changes requested):

Brian-McM (approved):

PR #3922, opened 2025-05-01
Update get-manifests.sh with renamed yamls and CRD location

coutinhop:
## For PR author

- [ ] Tests for change.
- [ ] If changing pkg/apis/, run `make gen-files`
- [ ] If changing versions, run `make gen-versions`

## For PR reviewers

A note for code reviewers - all pull requests must have the following:

- [ ] Milestone set according to targeted release.
- [ ] Appropriate labels:
  - `kind/bug` if this is a bugfix.
  - `kind/enhancement` if this is a a new feature.
  - `enterprise` if this PR applies to Calico Enterprise only.

PR #3944, opened 2025-05-13
Update ocp manifest building master

danudey:
This PR includes a significant amount of changes to the OCP bundle building process. Notable updates are:

1. Update the versions of `operator-sdk` and `kustomize` to the latest versions to be compatible with the v4 operator spec
2. Update the generation workflow for the bundle by changing how we call `operator-sdk`
3. Update the 'base' ClusterServiceVersion in `config/manifests/bases` even though we end up having to re-specify a lot of the values later anyway
4. Incorporate @coutinhop's changes from #3922 

It also includes housekeeping tasks, modernizations, and refactoring, like:
1. Improve output of the scripts being run during the bundle process to more clearly convey what's happening/what's breaking
2. Update the `curl` commands being run so that they fail if they get a 404 (!!)
3. Switch `yq` syntax to work with version 4
4. Compile all `yq` updates for CSV into one large file and then apply them all at once (the file can be inspected later if there are errors)
5. Update how we fetch `kustomize`, `operator-sdk`, and `helm3` while we're at it
6. Create the new Makefile variables `NATIVE_ARCH` and `NATIVE_OS` to represent the OS and architecture we're building _on_ rather than the OS/architecture we're building _for_. Use these variables when we fetch binaries for `kustomize` and `operator-sdk`, but not for `helm3` (yet) because it may need to be run in non-native-arch containers and we don't want to break that if so.


tmjd (approved):
LGTM

PR #3960, opened 2025-05-30
Move apiserver component into calico-system ns for enterprise

vara2504:
## Description

Need to be merged along with the calico-private changes.

Changes to move the API server component into the calico-system namespace for the enterprise variant

Update default deny policy in calico-system to exclude API server pods
Create a separate CA bundle for the API server in calico-system
Clean up the legacy tigera-system namespace
Update all references from tigera-system to calico-system

RELEASE_NOTE_REQUIRED, DOCS_PR_REQUIRED

## Release Note

```release-note
Component Migration: To support a minimal footprint and simplify resource management, the tigera-apiserver component and its associated resources have been moved from the tigera-system namespace to the calico-system namespace
```

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/apiserver.go
+++ b/pkg/render/apiserver.go
@@ -52,7 +52,7 @@ type ContainerName string
 const (
 	APIServerPort       = 5443
 	APIServerPortName   = "apiserver"
-	APIServerPolicyName = networkpolicy.TigeraComponentPolicyPrefix + "cnx-apiserver-access"
+	APIServerPolicyName = networkpolicy.TigeraComponentPolicyPrefix + "apiserver-access"
 
 	auditLogsVolumeName   = "tigera-audit-logs"
 	auditPolicyVolumeName = "tigera-audit-policy"
--- a/pkg/controller/apiserver/apiserver_controller_test.go
+++ b/pkg/controller/apiserver/apiserver_controller_test.go
@@ -325,7 +325,7 @@ var _ = Describe("apiserver controller tests", func() {
 			policies := v3.NetworkPolicyList{}
 			Expect(cli.List(ctx, &policies)).ToNot(HaveOccurred())
 			Expect(policies.Items).To(HaveLen(1))
-			Expect(policies.Items[0].Name).To(Equal("allow-tigera.cnx-apiserver-access"))
+			Expect(policies.Items[0].Name).To(Equal("allow-tigera.apiserver-access"))
 		})
 
 		It("should omit allow-tigera policy and not degrade when tier is not ready", func() {
--- a/pkg/render/apiserver_test.go
+++ b/pkg/render/apiserver_test.go
@@ -753,7 +753,7 @@ var _ = Describe("API server rendering tests (Calico Enterprise)", func() {
 
 		component := render.APIServerPolicy(cfg)
 		resources, _ := component.Objects()
-		policyName := types.NamespacedName{Name: "allow-tigera.cnx-apiserver-access", Namespace: "calico-system"}
+		policyName := types.NamespacedName{Name: "allow-tigera.apiserver-access", Namespace: "calico-system"}
 		policy := testutils.GetAllowTigeraPolicyFromResources(policyName, resources)
 		Expect(policy).ToNot(BeNil())
 		Expect(policy.Spec).ToNot(BeNil())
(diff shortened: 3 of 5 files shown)


pkg/controller/apiserver/apiserver_controller.go
```diff
@@ -429,14 +443,25 @@ func (r *ReconcileAPIServer) Reconcile(ctx context.Context, request reconcile.Re
...
+		CanCleanupOlderResources:    canCleanupOlderResources,
+	}
+
+	var components []render.Component
+
+	// v3 NetworkPolicy will fail to reconcile if the API server deployment is unhealthy. In case the API Server
+	// deployment becomes unhealthy and reconciliation of non-NetworkPolicy resources in the apiserver controller
+	// would resolve it, we render the network policies of components last to prevent a chicken-and-egg scenario.
+	if includeV3NetworkPolicy {
+		components = append(components, render.APIServerPolicy(&apiServerCfg))
```
vara2504, line 451: Moved includeV3NetworkPolicy from later in the render sequence (line 455) to the beginning. This ensures the cnx-apiserver-access NetworkPolicy is created early during upgrades, while API server pods are still running in the tigera-system namespace. This prevents default deny issues in calico-system when new API server pods start there.

Even though the default deny policy in calico-system is updated to exclude API server pods, applying that update still requires a healthy API server — hence the need to ensure access first

pkg/controller/apiserver/apiserver_controller.go
```diff
@@ -408,6 +418,10 @@ func (r *ReconcileAPIServer) Reconcile(ctx context.Context, request reconcile.Re
 		return reconcile.Result{}, err
 	}
 
+	// Check if the legacy namespace 'tigera-system' exists and can be cleaned up
+	canCleanupOlderResources := false
```
rene-dekker: Looks like line 422 can be removed. Optionally, 423 can be removed and called in-line @446.

pkg/controller/installation/core_controller.go
```diff
@@ -2063,3 +2063,20 @@ func crdPoolsToOperator(crds []crdv1.IPPool) []operator.IPPool {
 	}
 	return pools
 }
+
+func allowTigeraDefaultDenyForCalicoSystem(namespace string) *v3.NetworkPolicy {
```
rene-dekker: No need for passing a namespace.

pkg/controller/utils/utils.go
```diff
@@ -1039,3 +1039,15 @@ func MaintainInstallationFinalizer(
 	// Update the installation with any finalizer changes.
 	return c.Patch(ctx, installation, patchFrom)
 }
+
+func IsTigeraStatusReady(ts *operatorv1.TigeraStatus, l logr.Logger) bool {
```
rene-dekker: Should we perhaps make an `available()` method on the object itself? Logging can be removed or handled outside of the func.

pkg/render/apiserver.go
```diff
@@ -2325,6 +2327,16 @@ func (c *apiServerComponent) getDeprecatedResources() []client.Object {
 				Name: "tigera-tier-getter",
 			},
 		})
+
+		// Delete the older namespace for enterprise
+		if c.cfg.CanCleanupOlderResources {
```
rene-dekker, line 2332: I must be missing something, but where are the deployment and service cleaned up?

vara2504: I thought deleting a namespace does trigger garbage collection for all namespaced resources within it.
API Server switching from OSS to Enterprise https://github.com/tigera/operator/blob/b72f4372a9fa1855fc1caa9c0400ed2374675192/pkg/render/apiserver.go#L246
Moving Guardian from `tigera-guardian` to `calico-system`
https://github.com/tigera/operator/blob/b72f4372a9fa1855fc1caa9c0400ed2374675192/pkg/render/guardian.go#L695

I observed the same behavior during my testing, the namespace enters `Terminating` and eventually all resources are cleaned up as expected. I’ve also included this scenario as part of the test case.

WDYT? Do we need to explicitly delete each resource inside `tigera-system` before removing the namespace?

rene-dekker: Since there is precedence and since your testing shows no issues, that means I learned something new. I guess we don't need further changes.

rene-dekker (approved):

pkg/controller/apiserver/apiserver_controller.go
caseydavenport: Can we rename this from `cnx-access` as part of this move? `cnx` is confusing and has no meaning.

caseydavenport: OK if not, but if it's easy that would be nice.

PR #3962, opened 2025-05-30
Use pointer helpers from k8s utils lib

hjiawei:
## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -39,6 +39,7 @@ require (
 	k8s.io/apiserver v0.32.4
 	k8s.io/client-go v0.32.4
 	k8s.io/kube-aggregator v0.30.5
+	k8s.io/utils v0.0.0-20241210054802-24370beab758
 	sigs.k8s.io/controller-runtime v0.20.2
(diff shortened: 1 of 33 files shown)


Brian-McM:
Why does the pointer library need to be changed?

Brian-McM:
> Leveraging the existing library helped reduce our codebase footprint, even if only slightly in this case.

I personally don't agree with that, especially for small utilities. We're importing k8s for k8s utilities, not for general utils.

I've found this leads to using multiple different third parties to import basic utils and then if we stop using those utilities for the reasons we actually needed them, we're stuck with the dependencies / need a bigger refactor. 

It makes it harder to maintain if there's even a slight descrepency between what we need and what they have, leading to our own implementations anyway.

In this case when it's just a single function it doesn't seem useful, especially when the function is already there (just unnecessary churn). Like I'm sure k8s and other libraries have implemented all the simple helper functions we have, I wouldn't advocate for swapping out all our small util functions for third party implementations across all our imports.

PR #3964, opened 2025-06-02
Add Talos kubernetes provider to support Kubeprism endpoint

### LINKED ISSUE #3470: Support KubePrism endpoint on Talos
On Talos, it is recommended to pass the KubePrism endpoint to CNI components etc. https://www.talos.dev/v1.7/kubernetes-guides/configuration/kubeprism/

## Expected Behavior

Setting the `kubernetesServiceEndpoint`  to the KubePrism endpoint (`localhost:7445`) should work

## Current Behavior

Because some components like the kube controllers do not run in host network mode, these containers fail when trying to access that endpoint. There does not appear to be any way to separate out using different endpoints for the host network vs non-host network cases as is.

## Possible Solution

Add a new Talos Provider which can handle the case of using the KubePrism endpoint on Talos, similar to how the Docker EE provider handles using `proxy.local` here https://github.com/tigera/operator/blob/master/pkg/controller/k8sapi/k8s-endpoint.go#L54.

Another possible solution would be to allow configuring a host mode endpoint and a non-host mode endpoint which might be a bit more generic and wouldn't require special casing for Talos.

## Your Environment

* Operating System and version: Talos 1.7.6

Pythoner6:
## Description

Fixes #3470 by adding a new kubenetes provider for Talos, which detects setting the k8s service endpoint up for kubeprism and handling it similarly to the existing DockerEE provider. I couldn't find any other existing way to get the operator to use the kubeprism endpoint for host networked pods, but the k8s service for non-host networked pods.

Added unit tests for the new provider as well as tested deploying on my physical homelab cluster of RPIs.

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/api/v1/installation_types.go
+++ b/api/v1/installation_types.go
@@ -82,7 +82,7 @@ type InstallationSpec struct {
 	// If the specified value is not empty, the Operator will still attempt auto-detection, but
 	// will additionally compare the auto-detected value to the specified value to confirm they match.
 	// +optional
-	// +kubebuilder:validation:Enum="";EKS;GKE;AKS;OpenShift;DockerEnterprise;RKE2;TKG;
+	// +kubebuilder:validation:Enum="";EKS;GKE;AKS;OpenShift;DockerEnterprise;RKE2;TKG;Talos;
 	KubernetesProvider Provider `json:"kubernetesProvider,omitempty"`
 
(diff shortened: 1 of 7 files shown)


caseydavenport:

```

```

PR #3969, opened 2025-06-06
Consolidate RBAC for Guardian to handle backend component requests from the management cluster

vara2504:
## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/common/clusterrole/clusterrolerules.go
+++ b/pkg/render/common/clusterrole/clusterrolerules.go
@@ -1,0 +1,184 @@
+// Copyright (c) 2020-2025 Tigera, Inc. All rights reserved.
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package clusterrole
+
+import (
+	"github.com/tigera/operator/pkg/render/common/securitycontextconstraints"
+	rbacv1 "k8s.io/api/rbac/v1"
+)
+
+// CalicoKubeControllersClusterRoleRules is used to populate rules for calico-kube-controller CR to handle request from calico-kube-controller pods
+// and for guardian CR to handle request kube-controller request from management cluster.
+func CalicoKubeControllersClusterRoleRules(isEnterpriseCluster bool, isOpenShift bool, isManagedCluster bool) []rbacv1.PolicyRule {
+	rules := KubeControllersRoleCommonRules(isOpenShift)
+	if isEnterpriseCluster {
+		rules = append(rules, KubeControllersRoleEnterpriseCommonRules(isManagedCluster)...)
+		rules = append(rules,
+			rbacv1.PolicyRule{
+				APIGroups: []string{"crd.projectcalico.org"},
+				Resources: []string{"remoteclusterconfigurations"},
+				Verbs:     []string{"watch", "list", "get"},
+			},
+			rbacv1.PolicyRule{
+				APIGroups: []string{""},
+				Resources: []string{"endpoints"},
+				Verbs:     []string{"create", "update", "delete"},
+			},
+			rbacv1.PolicyRule{
+				APIGroups: []string{""},
+				Resources: []string{"namespaces"},
+				Verbs:     []string{"get"},
+			},
+			rbacv1.PolicyRule{
+				APIGroups: []string{"usage.tigera.io"},
+				Resources: []string{"licenseusagereports"},
+				Verbs:     []string{"create", "update", "delete", "watch", "list", "get"},
+			},
+		)
+	}
+	return rules
+}
+
+func KubeControllersRoleCommonRules(isOpenShift bool) []rbacv1.PolicyRule {
+	rules := []rbacv1.PolicyRule{
+		{
+			// Nodes are watched to monitor for deletions.
+			APIGroups: []string{""},
+			Resources: []string{"nodes", "endpoints", "services"},
+			Verbs:     []string{"watch", "list", "get"},
+		},
+		{
+			// Pods are watched to check for existence as part of IPAM GC.
+			APIGroups: []string{""},
+			Resources: []string{"pods"},
+			Verbs:     []string{"get", "list", "watch"},
+		},
+		{
+			APIGroups: []string{""},
+			Resources: []string{"services", "services/status"},
+			Verbs:     []string{"get", "list", "update", "watch"},
+		},
+		{
+			// IPAM resources are manipulated in response to node and block updates, as well as periodic triggers.
+			APIGroups: []string{"crd.projectcalico.org"},
+			Resources: []string{"ipreservations"},
+			Verbs:     []string{"list"},
+		},
+		{
+			APIGroups: []string{"crd.projectcalico.org"},
+			Resources: []string{"blockaffinities", "ipamblocks", "ipamhandles", "networksets", "ipamconfigs"},
+			Verbs:     []string{"get", "list", "create", "update", "delete", "watch"},
+		},
+		{
+			// Pools are watched to maintain a mapping of blocks to IP pools.
+			APIGroups: []string{"crd.projectcalico.org"},
+			Resources: []string{"ippools"},
+			Verbs:     []string{"list", "watch"},
+		},
+		{
+			// Needs access to update clusterinformations.
+			APIGroups: []string{"crd.projectcalico.org"},
+			Resources: []string{"clusterinformations"},
+			Verbs:     []string{"get", "create", "update", "list", "watch"},
+		},
+		{
+			// Needs to manage hostendpoints.
+			APIGroups: []string{"crd.projectcalico.org"},
+			Resources: []string{"hostendpoints"},
+			Verbs:     []string{"get", "list", "create", "update", "delete", "watch"},
+		},
+		{
+			// Needs to manipulate kubecontrollersconfiguration, which contains
+			// its config.  It creates a default if none exists, and updates status
+			// as well.
+			APIGroups: []string{"crd.projectcalico.org"},
+			Resources: []string{"kubecontrollersconfigurations"},
+			Verbs:     []string{"get", "create", "list", "update", "watch"},
+		},
+		{
+			// calico-kube-controllers requires tiers create
+			APIGroups: []string{"crd.projectcalico.org"},
+			Resources: []string{"tiers"},
+			Verbs:     []string{"create"},
+		},
+	}
+
+	if isOpenShift {
+		rules = append(rules, rbacv1.PolicyRule{
+			APIGroups:     []string{"security.openshift.io"},
+			Resources:     []string{"securitycontextconstraints"},
+			Verbs:         []string{"use"},
+			ResourceNames: []string{securitycontextconstraints.NonRootV2},
+		})
+	}
+
+	return rules
+}
+
+func KubeControllersRoleEnterpriseCommonRules(isManagedCluster bool) []rbacv1.PolicyRule {
+	rules := []rbacv1.PolicyRule{
+		{
+			APIGroups: []string{""},
+			Resources: []string{"configmaps"},
+			Verbs:     []string{"watch", "list", "get", "update", "create", "delete"},
+		},
+		{
+			// The Federated Services Controller needs access to the remote kubeconfig secret
+			// in order to create a remote syncer.
+			APIGroups: []string{""},
+			Resources: []string{"secrets"},
+			Verbs:     []string{"watch", "list", "get"},
+		},
+		{
+			// Needed to validate the license
+			APIGroups: []string{"projectcalico.org"},
+			Resources: []string{"licensekeys"},
+			Verbs:     []string{"get", "watch", "list"},
+		},
+		{
+			// Needed to validate the license
+			APIGroups: []string{"crd.projectcalico.org"},
+			Resources: []string{"licensekeys"},
+			Verbs:     []string{"get", "watch"},
+		},
+		{
+			APIGroups: []string{"projectcalico.org", "crd.projectcalico.org"},
+			Resources: []string{"deeppacketinspections"},
+			Verbs:     []string{"get", "watch", "list"},
+		},
+		{
+			APIGroups: []string{"crd.projectcalico.org"},
+			Resources: []string{"deeppacketinspections/status"},
+			Verbs:     []string{"update"},
+		},
+		{
+			APIGroups: []string{"crd.projectcalico.org"},
+			Resources: []string{"packetcaptures"},
+			Verbs:     []string{"get", "list", "update"},
+		},
+	}
+
+	if isManagedCluster {
+		rules = append(rules,
+			rbacv1.PolicyRule{
+				APIGroups: []string{"projectcalico.org"},
+				Resources: []string{"licensekeys"},
+				Verbs:     []string{"get", "create", "update", "list", "watch"},
+			},
+		)
+	}
+
+	return rules
+}
--- a/pkg/render/compliance.go
+++ b/pkg/render/compliance.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2019-2024 Tigera, Inc. All rights reserved.
+// Copyright (c) 2019-2025 Tigera, Inc. All rights reserved.
 
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -725,8 +725,8 @@ func (c *complianceComponent) externalLinseedRoleBinding() *rbacv1.RoleBinding {
 		Subjects: []rbacv1.Subject{
 			{
 				Kind:      "ServiceAccount",
-				Name:      linseed,
-				Namespace: ElasticsearchNamespace,
+				Name:      GuardianName,
+				Namespace: GuardianNamespace,
 			},
 		},
 	}
--- a/pkg/render/fluentd.go
+++ b/pkg/render/fluentd.go
@@ -339,8 +339,8 @@ func (c *fluentdComponent) externalLinseedRoleBinding() *rbacv1.RoleBinding {
 		Subjects: []rbacv1.Subject{
 			{
 				Kind:      "ServiceAccount",
-				Name:      "tigera-linseed",
-				Namespace: ElasticsearchNamespace,
+				Name:      GuardianName,
+				Namespace: GuardianNamespace,
 			},
 		},
 	}
--- a/pkg/render/guardian.go
+++ b/pkg/render/guardian.go
@@ -47,6 +47,7 @@ import (
 	"github.com/tigera/operator/pkg/render/common/securitycontext"
 	"github.com/tigera/operator/pkg/render/common/securitycontextconstraints"
 	"github.com/tigera/operator/pkg/render/common/selector"
+	"github.com/tigera/operator/pkg/render/kubecontrollers"
 	"github.com/tigera/operator/pkg/tls/certificatemanagement"
 )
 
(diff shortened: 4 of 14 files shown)


pkg/render/common/clusterrole/clusterrolerules.go
```diff
@@ -0,0 +1,191 @@
...
+
+import (
+	rbacv1 "k8s.io/api/rbac/v1"
+
+	"github.com/tigera/operator/pkg/render/common/securitycontextconstraints"
+)
+
+// CalicoKubeControllersClusterRoleRules defines RBAC rules for calico-kube-controller pods
+// and for guardian to handle kube-controller requests from the management cluster.
+func CalicoKubeControllersClusterRoleRules(isEnterpriseCluster bool, isOpenShift bool, isManagedCluster bool) []rbacv1.PolicyRule {
```
vara2504: In our existing code, we have some common functions within kubecontroller for building the ClusterRole rules used by both calico-kube-controllers and es-calico-kube-controllers, which are passed into the renderer via cfg.
To enable reuse between kube-controllers and Guardian, I’ve moved these functions to a common file under render.

pkg/render/common/clusterrole/clusterrolerules.go
```diff
@@ -0,0 +1,191 @@
...
+	rules := KubeControllersRoleCommonRules(isOpenShift)
+	if isEnterpriseCluster {
+		rules = append(rules, KubeControllersRoleEnterpriseCommonRules(isManagedCluster)...)
+		rules = append(rules,
+			rbacv1.PolicyRule{
```
caseydavenport: It would be nice to see commenting explaining why each set of permissions is needed (like we do for other roles)

pkg/render/compliance.go
```diff
@@ -720,13 +720,13 @@ func (c *complianceComponent) externalLinseedRoleBinding() *rbacv1.RoleBinding {
 		RoleRef: rbacv1.RoleRef{
 			APIGroup: "rbac.authorization.k8s.io",
 			Kind:     "ClusterRole",
-			Name:     TigeraLinseedSecretsClusterRole,
```
caseydavenport: Do we need this binding at all any more? Guardian should already have the guardian cluster role bound to its serviceaccount, right?

pkg/render/fluentd.go
```diff
@@ -334,13 +334,13 @@ func (c *fluentdComponent) externalLinseedRoleBinding() *rbacv1.RoleBinding {
 		RoleRef: rbacv1.RoleRef{
 			APIGroup: "rbac.authorization.k8s.io",
 			Kind:     "ClusterRole",
-			Name:     TigeraLinseedSecretsClusterRole,
+			Name:     GuardianClusterRoleName,
 		},
 		Subjects: []rbacv1.Subject{
 			{
 				Kind:      "ServiceAccount",
-				Name:      "tigera-linseed",
```
caseydavenport, line 385: same question here

pkg/render/guardian.go
```diff
@@ -239,6 +233,21 @@ func (c *GuardianComponent) clusterRole() *rbacv1.ClusterRole {
 			Verbs:     []string{"impersonate"},
 		})
 
+		// Add rules needed for guardian to handle the policy recommendation request in managed clusters.
```
caseydavenport: I have a slight worry about breaking it down this way - I think it will be prone to rule duplication. For example, if two components both need read access to NetworkPolicy, we will duplicate those rules. Or more confusing, if one component needs "read" access and another needs "write" to the same resource, those will not end up in the same block. 

I wonder if we need to be doing something a bit more here - e.g., merging and deduplicating permissions needed by each component?

pkg/render/guardian.go
```diff
@@ -708,5 +717,23 @@ func deprecatedObjects() []client.Object {
 			TypeMeta:   metav1.TypeMeta{Kind: "ClusterRoleBinding", APIVersion: "rbac.authorization.k8s.io/v1"},
 			ObjectMeta: metav1.ObjectMeta{Name: "tigera-guardian"},
 		},
+
+		// Remove manager namespace objects since the guardian identity is responsible for handling manager requests
```
caseydavenport, line 993: Do we need to do the same for tigera-elasticsearch, tigera-intrusion-detection, etc.?

pkg/render/intrusion_detection.go
```diff
@@ -433,13 +440,13 @@ func (c *intrusionDetectionComponent) externalLinseedRoleBinding() *rbacv1.RoleB
 		RoleRef: rbacv1.RoleRef{
 			APIGroup: "rbac.authorization.k8s.io",
 			Kind:     "ClusterRole",
-			Name:     TigeraLinseedSecretsClusterRole,
+			Name:     GuardianClusterRoleName,
```
caseydavenport: And here too - do we need this role at all?

pkg/render/logstorage.go
```diff
@@ -1410,3 +1368,26 @@ func (m managedClusterLogStorage) kubeControllersRolesAndBindings() ([]*rbacv1.R
 	return []*rbacv1.Role{operatorSecretsRole}, []*rbacv1.RoleBinding{operatorSecretBinding}
 
 }
+
+func (m *managedClusterLogStorage) deprecatedObjects() []client.Object {
+	return []client.Object{
+		// In a managed cluster, the guardian identity handles Linseed requests.
+		// Therefore, Linseed's cluster-scoped RBAC objects should be removed in a managed cluster.
```
caseydavenport, line 1380: Do we want to delete the namespace here too?

pkg/render/guardian.go
caseydavenport: Ah, I see some of them are handled in other controllers. Is there a reason to remove the manager resources from the guardian controller, but the other resources from their own controllers?

pkg/render/manager.go
```diff
@@ -1166,3 +993,183 @@ func managerClusterWideDefaultView() *v3.UISettings {
...
+		{
+			APIGroups: []string{""},
+			Resources: []string{"services/proxy"},
+			ResourceNames: []string{
+				"https:tigera-api:8080", "calico-node-prometheus:9090",
+			},
+			Verbs: []string{"get", "create"},
+		},
+		{
+			// Add access to Linseed APIs.
```
caseydavenport: I don't think these are needed for managed clusters? There's no linseed in a managed cluster.

pkg/render/manager.go
```diff
@@ -1166,3 +993,183 @@ func managerClusterWideDefaultView() *v3.UISettings {
...
+		{
+			APIGroups: []string{"authorization.k8s.io"},
+			Resources: []string{"subjectaccessreviews"},
+			Verbs:     []string{"create"},
+		},
+		{
+			APIGroups: []string{"authentication.k8s.io"},
+			Resources: []string{"tokenreviews"},
+			Verbs:     []string{"create"},
+		},
```
caseydavenport: Are you sure these are needed in managed clusters?

pkg/render/common/clusterrole/clusterrolerules.go
caseydavenport: Could guardian not just import these from `pkg/render/kubecontrollers`? 

Guessing there was an import cycle? Best way to fix it would likely be to have `pkg/render/guardian/guardian.go`?

caseydavenport (approved):
Thanks @vara2504 - added some thoughts

pkg/render/compliance.go
vara2504: Ah, my bad - my intention was to:
- Retain the ClusterRole for tigera-linseed-configmap and tigera-linseed-secrets.
- Create a "tigera-linseed" RoleBinding in the calico-system namespace for the tigera-linseed-configmap CR.
- Create "tigera-linseed" RoleBindings in the fluentd, dpi, intrusion-detection, and compliance namespaces for the tigera-linseed-secrets CR, using the Guardian service account.
This way, Guardian has write access to secrets only in those specific namespaces, not cluster-wide — and similarly limited access for the configmap.

I got a bit mixed up and mistakenly moved it to cluster scope during cleanup, but I’ve now reverted those changes

pkg/render/common/clusterrole/clusterrolerules.go
vara2504: Yes, it’s to avoid an import cycle. 
I moved it into a common function to prevent duplicating the same code in both Guardian and KubeController, which would otherwise require updates in two separate places in the future.

pkg/render/common/clusterrole/clusterrolerules.go
vara2504: I’ve just relocated the [existing function code ](https://github.com/tigera/operator/pull/3969/files#diff-0e769f5212593b665ec7a86aef0006a37f70981b42320d0c20864a35fad029aeL113)here to avoid duplication specifically, the rules defined in the ClusterRole. I'm not fully familiar with the original intent or specific use of each rule. :thinking:

pkg/render/fluentd.go
vara2504: As mentioned [here](https://github.com/tigera/operator/pull/3969#discussion_r2138708297), this is intended to restrict write access to secrets within the namespace scope

pkg/render/guardian.go
vara2504: While the current breakdown might lead to duplicate rules, as far as I understand Kubernetes effectively unions them at runtime, so there's no functional issue. The current breakdown was primarily done for clarity and to avoid duplicating rules across components like Guardian and others.

That said, I agree that when reading the rendered ClusterRole, there could be multiple entries for the same resource possibly with duplicate or overlapping verbs.

One possible improvement could be introducing a utility to merge rules by resource, verb, and API group before applying them to the final ClusterRole. WDYT?

pkg/render/guardian.go
vara2504: Because in managed clusters, Guardian was responsible for creating the manager namespace and its associated ClusterRoles (as shown above in line 154).

pkg/render/logstorage.go
vara2504: tigera-elasticsearch namespace currently has two external services. I’ll clean up the namespace as part of the task to remove those external services.

pkg/render/guardian.go
caseydavenport: > Kubernetes effectively unions them at runtime, so there's no functional issue

Yeah, the code will work, although it will be harder to track what permissions are actually granted (I suppose not any harder than it is today!)

I think longer term, I would prefer distinct code for "These are the permissions needed by a management cluster when talking to a managed clsuter" and "These are the permissions needed in a standalone/management cluster".

i.e., a singular ClusterRole rule set for Guardian with each permission set explicitly, and a comment explaining why. This would solve the problem without the need for merging code, and also make it easier to explicitly state the required permissions.

pkg/render/manager.go
vara2504: I’m not certain about that. If we’re confident these aren’t required in managed clusters, I can add a !managedCluster check accordingly

pkg/render/guardian.go
vara2504: Sounds good to me. I’ll make these changes in a follow-up PR, as I might need input from experts to determine exactly which permissions are needed by Guardian (for managed cluster) and why.

pkg/render/common/clusterrole/clusterrolerules.go
caseydavenport: I think these should still be in the kubecontrollers package, and guardian should be a new guardian package in order to fix this - otherwise these are outliers compared to how we do permissions for everything else. 

Happy for that to be another PR though.

pkg/render/manager.go
```diff
@@ -730,180 +730,7 @@ func managerClusterRole(managedCluster bool, kubernetesProvider operatorv1.Provi
...
-	if tenant.MultiTenant() {
-		cr.Rules = append(cr.Rules,
-			rbacv1.PolicyRule{
-				APIGroups: []string{"authorization.k8s.io"},
-				Resources: []string{"localsubjectaccessreviews"},
-				Verbs:     []string{"create"},
-			},
-		)
-
-		if tenant.ManagedClusterIsCalico() {
```
vara2504: @caseydavenport Since the Free tier now includes Guardian, will adding these rules to the Guardian RBAC also apply to the Free tier? For certain components, do we need to add a check for !tenant.ManagedClusterIsCalico() when merging these RBAC rules?

pkg/render/guardian.go
```diff
@@ -239,6 +232,10 @@ func (c *GuardianComponent) clusterRole() *rbacv1.ClusterRole {
 			Verbs:     []string{"impersonate"},
 		})
 
+		// to remove isManagedCluster and tenant param from this func.
```
caseydavenport: Is this comment a TODO?

pkg/render/guardian.go
```diff
@@ -690,6 +687,624 @@ func GuardianService(clusterDomain string) string {
 	return fmt.Sprintf("https://%s.%s.svc.%s:%d", GuardianServiceName, GuardianNamespace, clusterDomain, 443)
 }
 
+func AddClusterRoleRulesForManagementBackendRequest(isManagedCluster, isOpenShift bool, tenant *operatorv1.Tenant) []rbacv1.PolicyRule {
```
caseydavenport:
```suggestion
// rulesForManagementClusterRequests returns the set of RBAC rules needed by Guardian in order to 
// satisfy requests from the management cluster over the tunnel.
func rulesForManagementClusterRequests(isManagedCluster, isOpenShift bool, tenant *operatorv1.Tenant) []rbacv1.PolicyRule {
```

Name change reasoning: I think this can be private? Plus it's not adding rules, just generating rules (which are added in the caller)

pkg/render/guardian.go
```diff
@@ -690,6 +687,624 @@ func GuardianService(clusterDomain string) string {
...
+
+	// Sort ap
```
caseydavenport: Sort ap?

pkg/render/guardian.go
```diff
@@ -690,6 +687,624 @@ func GuardianService(clusterDomain string) string {
...
+	// General rules need for more than one of the following managemen cluster backend components
+	// such as linseed, manager, policy, calico-kube-controller, intrusion detection
+
+	rules := []rbacv1.PolicyRule{}
+
+	// Rules needed  by guardian to handle linseed request
```
caseydavenport:
```suggestion
	// Rules needed  by guardian to handle linseed requests
```

pkg/render/guardian.go
```diff
@@ -690,6 +687,624 @@ func GuardianService(clusterDomain string) string {
...
+
+	// This permission allows Linseed to watch for namespace creation and existence in the managed cluster
+	// before attempting to copy the Linseed token into those namespaces.
+	rules = append(rules, rbacv1.PolicyRule{
+		APIGroups: []string{""},
+		Resources: []string{"namespaces"},
+		Verbs:     []string{"get", "list", "watch"},
+	})
+
+	// Rules needed by guardian to handle manager request
```
caseydavenport:
```suggestion
	// Rules needed by guardian to handle manager requests
```

pkg/render/guardian.go
```diff
@@ -690,6 +687,624 @@ func GuardianService(clusterDomain string) string {
...
+		rbacv1.PolicyRule{
+			APIGroups: []string{"policy.networking.k8s.io"},
+			Resources: []string{
+				"adminnetworkpolicies",
+				"baselineadminnetworkpolicies",
+			},
+			Verbs: []string{"list"},
+		},
+	)
+	/* This logic could be removed for Guardian
```
caseydavenport: Yep, we can remove this

pkg/render/guardian.go
```diff
@@ -690,6 +687,624 @@ func GuardianService(clusterDomain string) string {
...
+			rbacv1.PolicyRule{
+				APIGroups:     []string{"security.openshift.io"},
+				Resources:     []string{"securitycontextconstraints"},
+				Verbs:         []string{"use"},
+				ResourceNames: []string{securitycontextconstraints.NonRootV2},
+			},
+		)
+	}
+
+	/* In a managed cluster, tenant is nil, so this logic can be safely removed.
```
caseydavenport: Same here

pkg/render/guardian.go
```diff
@@ -690,6 +687,624 @@ func GuardianService(clusterDomain string) string {
...
+					Resources: []string{"flowlogs"},
+					Verbs:     []string{"create"},
+				})
+		}
+	}
+	*/
+
+	// End of manager rules
+
+	// Rules needed  by guardian to handle policy recommendation request
```
caseydavenport:
```suggestion
	// Rules needed  by guardian to handle policy recommendation requests
```

pkg/render/guardian.go
```diff
@@ -690,6 +687,624 @@ func GuardianService(clusterDomain string) string {
...
+	}
+	*/
+
+	// Rules needed  by guardian to handle calico-kube-controller request
+
+	// Is it possible to differentiate between the RBAC rules required for calico-kube-controllers
+	// and those needed by Guardian to handle management cluster requests?
+
+	// kc common rules
+	rules = append(rules,
```
caseydavenport: Yeah, I think a bunch of these aren't needed. e.g., management cluster kube-controllers shouldnt do anything with IPAM in the managed cluster. 

I suspect most of these aren't needed.

pkg/render/guardian.go
```diff
@@ -690,6 +687,624 @@ func GuardianService(clusterDomain string) string {
...
+			APIGroups: []string{"projectcalico.org"},
+			Resources: []string{"licensekeys"},
+			Verbs:     []string{"get", "watch", "list"},
+		},
+		rbacv1.PolicyRule{
+			// Needed to validate the license
+			APIGroups: []string{"crd.projectcalico.org"},
+			Resources: []string{"licensekeys"},
+			Verbs:     []string{"get", "watch"},
+		},
```
caseydavenport: These can be combined?

pkg/render/guardian.go
```diff
@@ -690,6 +687,624 @@ func GuardianService(clusterDomain string) string {
...
+			Verbs:     []string{"update"},
+		},
+		rbacv1.PolicyRule{
+			APIGroups: []string{"crd.projectcalico.org"},
+			Resources: []string{"packetcaptures"},
+			Verbs:     []string{"get", "list", "update"},
+		},
+	)
+
+	if isManagedCluster {
```
caseydavenport: Isn't this always true?

pkg/render/guardian.go
```diff
@@ -690,6 +687,624 @@ func GuardianService(clusterDomain string) string {
...
+				"globalalerts/status",
+				"globalthreatfeeds",
+				"globalthreatfeeds/status",
+				"globalnetworksets",
+			},
+			Verbs: []string{
+				"get", "list", "watch", "create", "update", "patch", "delete",
+			},
+		},
+		rbacv1.PolicyRule{
```
caseydavenport: This is also a duplicate of above.

Probably better to have a single "licensekeys" permission, and a comment explaining who uses it.

pkg/render/guardian.go
```diff
@@ -690,6 +687,624 @@ func GuardianService(clusterDomain string) string {
...
+			Resources: []string{"events"},
+			Verbs:     []string{"create"},
+		},
+		rbacv1.PolicyRule{
+			// Add write/read/delete access to Linseed APIs.
+			APIGroups: []string{"linseed.tigera.io"},
+			Resources: []string{"threatfeeds_ipset", "threatfeeds_domainnameset"},
+			Verbs:     []string{"create", "delete", "get"},
+		},
+		rbacv1.PolicyRule{
```
caseydavenport: Shouldn't need any `linseed.tigera.io` permissions in managed clusters.

pkg/render/guardian.go
vara2504: will clean it up - Just a draft note for myself to sort the rules alphabetically under each component in this PR, in case that makes the consolidation easier

pkg/render/guardian.go
vara2504: The API groups appear to be different: projectcalico.org and crd.projectcalico.org.

caseydavenport: Ah yes, interesting. I wonder why (and if both are actually needed).

caseydavenport (approved):

vara2504:
@caseydavenport Since these changes affect only the backend components of the management cluster, no documentation or release notes are required

PR #3970, opened 2025-06-06
Remove bogus mockStatus.On calls for SetDegraded

nelljerram:
Calls like this were bogus:

    mockStatus.On("SetDegraded", "Waiting for LicenseKeyAPI to be ready", "").Return().Maybe()

because all of the actual calls like that have an initial `operatorv1.ResourceNotReady` arg.

(And they were also confusing, because they were the model that I copied when I needed to mock out another actual SetDegraded call.)


caseydavenport (approved):
Good catch!

PR #3977, opened 2025-06-10
Use rolebingins to access managed clusters instead of cluster role bi…

asincu:
…ndings

## Description

Use rolebindings to managed RBAC for managedclusters instead of cluster
role bindings. Create a ClusterRole that allows watch access (allow
get,list,watch). All components will bind those as needed via RoleBindings
in their canonical namespace. Manager will create its additional RBAC via Role/RoleBindings to update managed clusters. KubeControllers RBAC to update managed clusters will be reverted. Leaving this as it is now.

## Release Note

```release-note
TBD
```

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/apiserver.go
+++ b/pkg/render/apiserver.go
@@ -287,10 +287,10 @@ func (c *apiServerComponent) Objects() ([]client.Object, []client.Object) {
 		globalEnterpriseObjects = append(globalEnterpriseObjects,
 			c.tigeraUserClusterRole(),
 			c.tigeraNetworkAdminClusterRole(),
+			c.managedClusterWatchClusterRole(),
 		)
 	}
 
-	globalEnterpriseObjects = append(globalEnterpriseObjects, c.managedClusterClusterRoles()...)
 	if c.cfg.ManagementCluster != nil {
 		if c.cfg.MultiTenant {
 			// Multi-tenant management cluster API servers need access to per-tenant CA secrets in order to sign
@@ -2294,10 +2294,9 @@ func (c *apiServerComponent) multiTenantManagedClusterAccessClusterRoles() []cli
 	return objects
 }
 
-// managedClusterClusterRoles creates ClusterRoles for accessing the ManagedCluster API
-func (c *apiServerComponent) managedClusterClusterRoles() []client.Object {
-	var objects []client.Object
-	objects = append(objects, &rbacv1.ClusterRole{
+// managedClusterWatchClusterRole creates a ClusterRole for watching the ManagedCluster API
+func (c *apiServerComponent) managedClusterWatchClusterRole() client.Object {
+	return &rbacv1.ClusterRole{
 		TypeMeta:   metav1.TypeMeta{Kind: "ClusterRole", APIVersion: "rbac.authorization.k8s.io/v1"},
 		ObjectMeta: metav1.ObjectMeta{Name: ManagedClustersWatchClusterRoleName},
 		Rules: []rbacv1.PolicyRule{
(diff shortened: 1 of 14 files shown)


pkg/render/common/components/components.go
```diff
@@ -548,6 +548,25 @@ func ClusterRoleBinding(name, clusterRole, sa string, namespaces []string) *rbac
 	}
 }
 
+func RoleBinding(name, clusterRole, sa string, namespace string) *rbacv1.RoleBinding {
```
caseydavenport, line 553: Could use a function comment

pkg/render/apiserver.go
```diff
@@ -2290,6 +2293,38 @@ func (c *apiServerComponent) multiTenantManagedClusterAccessClusterRoles() []cli
 	return objects
 }
 
+func (c *apiServerComponent) managedClusterClusterRoles() []client.Object {
```
caseydavenport:
```suggestion
// managedClusterClusterRoles creates ClusterRoles for accessing the ManagedCluster API - one for read access, and one for write access. 
// Individual controllers should use Namespaced RoleBindings (in multi-tenant environments) to grant these permissions within the scope of a single namespace. 
func (c *apiServerComponent) managedClusterClusterRoles() []client.Object {
```

pkg/render/intrusion_detection.go
```diff
@@ -420,6 +419,10 @@ func (c *intrusionDetectionComponent) intrusionDetectionClusterRoleBinding() *rb
 	return rcomponents.ClusterRoleBinding(IntrusionDetectionName, IntrusionDetectionName, IntrusionDetectionName, c.cfg.BindNamespaces)
 }
 
+func (c *intrusionDetectionComponent) managedClustersWatchRoleBinding() client.Object {
```
caseydavenport, line 422: Do we really need a function wrapping this other function? :grin:

pkg/render/apiserver.go
```diff
@@ -73,6 +73,8 @@ const (
 	tigeraAPIServerTLSSecretName                                  = "tigera-apiserver-certs"
 	APIServerSecretsRBACName                                      = "tigera-extension-apiserver-secrets-access"
 	MultiTenantManagedClustersAccessClusterRoleName               = "tigera-managed-cluster-access"
+	ManagedClustersWatchClusterRoleName                           = "tigera-managed-cluster-watch"
```
caseydavenport, line 77: This is defined in a few places - can / should this be commonized?

pkg/render/intrusion_detection.go
asincu: The other function has a long definition and makes the Objects() code hard to read. It takes you a while to unpack cluster role names, namespaces from the intent of adding RBAC to watch managed clusters.

caseydavenport: Hm, ok. I disagree but it's not worth changing if we're not in agreement!

pkg/render/apiserver.go
```diff
@@ -2290,6 +2292,26 @@ func (c *apiServerComponent) multiTenantManagedClusterAccessClusterRoles() []cli
 	return objects
 }
 
+// managedClusterClusterRoles creates ClusterRoles for accessing the ManagedCluster API
+func (c *apiServerComponent) managedClusterClusterRoles() []client.Object {
```
caseydavenport: Does this need to return a slice any more? It's only a single cluster role at this point.

pkg/render/common/components/components.go
```diff
@@ -548,6 +548,27 @@ func ClusterRoleBinding(name, clusterRole, sa string, namespaces []string) *rbac
 	}
 }
 
+// RoleBinding returns a role binding with the given name, that binds the given cluster role
+// to the service account in each of the provided namespaces.
```
caseydavenport: It only accepts a single namespace, so either the comment or the function is wrong :)

caseydavenport (approved):

PR #3979, opened 2025-06-11
Handle replicas override generically instead of specially for EgressGateway

nelljerram:
Replicas override is needed as part of #3852, and that prompted this change, but I think it makes sense to test it for EgressGateway and review that separately in advance.


nelljerram:
This PR is no longer required, because its changes were also included in #3852

PR #3986, opened 2025-06-13
Move EE API server namespace out of global Enterprise object

vara2504:
Avoids accidental cleanup of the namespace when creating OSS variant.
Removed namespace creation from apiserver components; handled by core-controller

## Release Note

```release-note
TBD
```

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/apiserver.go
+++ b/pkg/render/apiserver.go
@@ -269,8 +269,6 @@ func (c *apiServerComponent) Objects() ([]client.Object, []client.Object) {
 
 	// Global enterprise-only objects.
 	globalEnterpriseObjects := []client.Object{
-		CreateNamespace(rmeta.APIServerNamespace(operatorv1.TigeraSecureEnterprise), c.cfg.Installation.KubernetesProvider, PSSPrivileged, c.cfg.Installation.Azure),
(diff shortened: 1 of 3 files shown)


pkg/render/apiserver.go
```diff
@@ -336,6 +334,9 @@ func (c *apiServerComponent) Objects() ([]client.Object, []client.Object) {
 	// Compile the final arrays based on the variant.
 	if c.cfg.Installation.Variant == operatorv1.TigeraSecureEnterprise {
 		// Create any enterprise specific objects.
+		globalObjects = append(globalObjects,
+			CreateNamespace(rmeta.APIServerNamespace(operatorv1.TigeraSecureEnterprise), c.cfg.Installation.KubernetesProvider, PSSPrivileged, c.cfg.Installation.Azure),
```
caseydavenport: I don't think we need create namespace here at all any more, right? For enterprise, this is in calico-system, which will have been created already by the core controller?

vara2504: Yes, the core_controller renders the calico-system namespace for both Calico and EE variants.
https://github.com/tigera/operator/blob/master/pkg/controller/installation/core_controller.go#L1247

caseydavenport (approved):

PR #3988, opened 2025-06-13
Copy pull secret before linseed deployment

asincu:
## Release Note

```release-note
TBD
```

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


tmjd:
Why didn't we need this before? I'm wondering if copying the secret is going to conflict with another controller copying the secret or if maybe it is the same controller and we need to reconsider how the pull secret gets created in the ES namespace.

asincu:
Yup, there is another controller that copies it eventually. From what I can tell, we never put in annotations for the pull secrets. To answer why we did not do it before, I actually have three bugs for this problem. We normally see this when spinning a new node or cc-tigera tenant.

PR #3989, opened 2025-06-14
Move Calico API server from calico-apiserver namespace to calico-system for OSS

vara2504:
## Release Note

```release-note
Component Migration: To support a minimal footprint and simplify resource management, the calico-apiserver component and its associated resources have been moved from the calico-apiserver namespace to the calico-system namespace in Calico OSS .
```

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/apiserver/apiserver_controller.go
+++ b/pkg/controller/apiserver/apiserver_controller.go
@@ -21,6 +21,7 @@ import (
 	"github.com/go-logr/logr"
 	appsv1 "k8s.io/api/apps/v1"
 
+	v1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
@@ -155,9 +156,6 @@ func add(c ctrlruntime.Controller, r *ReconcileAPIServer) error {
 	}
 
 	// Watch for the namespace(s) managed by this controller.
-	if err = c.WatchObject(&corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: render.APIServerNamespace}}, &handler.EnqueueRequestForObject{}); err != nil {
-		return fmt.Errorf("apiserver-controller failed to watch resource: %w", err)
-	}
 	if err = c.WatchObject(&corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: render.APIServerNamespace}}, &handler.EnqueueRequestForObject{}); err != nil {
 		return fmt.Errorf("apiserver-controller failed to watch resource: %w", err)
 	}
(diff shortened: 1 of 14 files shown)


pkg/render/apiserver.go
```diff
@@ -321,35 +322,22 @@ func (c *apiServerComponent) Objects() ([]client.Object, []client.Object) {
...
 		globalObjects = append(globalObjects, globalEnterpriseObjects...)
 		namespacedObjects = append(namespacedObjects, namespacedEnterpriseObjects...)
 
-		// Explicitly delete any global OSS objects.
-		// Namespaced objects will be handled by namespace deletion.
-		objsToDelete = append(objsToDelete, globalCalicoObjects...)
-	} else {
-		// Create any Calico-only objects
-		globalObjects = append(globalObjects, globalCalicoObjects...)
+		objsToDelete = append(objsToDelete, c.cleanupDeployment())
```
vara2504: cleanupDeployment ensures cleanup of the calico-apiserver deployment during OSS to EE upgrades, where the API server is replaced by the tigera-apiserver deployment.

pkg/controller/apiserver/apiserver_controller.go
```diff
@@ -529,12 +521,16 @@ func (r *ReconcileAPIServer) maintainFinalizer(ctx context.Context, apiserver cl
 // - The new API server deployment in "calico-system" exists and is available.
 // - The old API server deployment in "tigera-system" is either removed or inactive.
 // - Both the APIServer custom resource and the TigeraStatus for 'apiserver' are in the Ready state
-func (r *ReconcileAPIServer) canCleanupLegacyNamespace(ctx context.Context, logger logr.Logger) bool {
-	const (
-		newNamespace   = "calico-system"
-		oldNamespace   = "tigera-system"
-		deploymentName = "tigera-apiserver"
-	)
+func (r *ReconcileAPIServer) canCleanupLegacyNamespace(ctx context.Context, variant operatorv1.ProductVariant, logger logr.Logger) bool {
```
rene-dekker, line 524: Will this clean up calico-apiserver if you upgrade from oss(operator v1.n-1)->ee?

Should this method iterate over both namespaces instead to ensure the removal of both?

vara2504: It should handle cleaning up the calico-apiserver deployment, as both the calico-apiserver deployment and the tigera-system namespace are removed once canCleanupLegacyNamespace is true here in getDeprecatedResources. 
https://github.com/tigera/operator/pull/3989/files#diff-88957f5c859486b2cbf260254b87b36b75fadbc99bfd38c01ea29dc0ac22a260R2340

If we're switching from OSS to the enterprise variant, the OSS namespace will need to be cleaned up anyway, and I believe EE doesn't have a dependency on the OSS apiserver component to come up. This means we can directly clean up the calico-apiserver namespace. (correct me if i am missing something).

We only need to iterate over the namespaces when migrating the APIServer between namespaces within the same variant. This is because, in order to run the apiserver inside the calico-system namespace, we need to create some allow-tigera policy rules in calico-system, which requires the apiserver deployment to be up and running in tigera-system.

rene-dekker:
Vara, could you also make it so that we no longer have a different name for the deployment based on the Variant? I.e. remove `func resourceNameBasedOnVariant()`, `APIServerServiceAccountName()`, `ProjectCalicoAPIServerServiceName()`, `ProjectCalicoAPIServerTLSSecretName()` and normalize the name to be calico-apiserver?

PR #3993, opened 2025-06-16
Pull secrets and namspace will be created by a setup component

asincu:
## Description

Pull secrets, namespaces and role bindings for operator secrets will be
created in the controller versus the rendered. A new setup component
will create the namespace (if specified), the rolebindings and pull
secrets. For MT setup, these objects will have the Tenant Owner.

## Release Note

```release-note
TBD
```

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/intrusiondetection/intrusiondetection_controller.go
+++ b/pkg/controller/intrusiondetection/intrusiondetection_controller.go
@@ -472,7 +472,7 @@ func (r *ReconcileIntrusionDetection) Reconcile(ctx context.Context, request rec
 		PullSecrets:     pullSecrets,
 		Namespace:       helper.InstallNamespace(),
 		PSS:             getPSS(lc),
-		CreateNamespace: shouldCreateNamespace(tenant, hasNoLicense),
+		CreateNamespace: !tenant.MultiTenant(),
 	})
 	intrusionDetectionComponent := render.IntrusionDetection(intrusionDetectionCfg)
 
@@ -602,13 +602,6 @@ func (r *ReconcileIntrusionDetection) Reconcile(ctx context.Context, request rec
 		return reconcile.Result{}, err
 	}
 	return reconcile.Result{}, nil
-}
-
-func shouldCreateNamespace(tenant *operatorv1.Tenant, hasNoLicense bool) bool {
-	if hasNoLicense || tenant.MultiTenant() {
-		return false
-	}
-	return true
 }
 
 func getPSS(logCollector *operatorv1.LogCollector) render.PodSecurityStandard {
--- a/pkg/controller/intrusiondetection/intrusiondetection_controller.go
+++ b/pkg/controller/intrusiondetection/intrusiondetection_controller.go
@@ -466,13 +466,13 @@ func (r *ReconcileIntrusionDetection) Reconcile(ctx context.Context, request rec
 		ExternalElastic:              r.elasticExternal,
 		SyslogForwardingIsEnabled:    syslogForwardingIsEnabled(lc),
 	}
-	intrusionDetectionNamespaceComponent := render.IntrusionDetectionNamespaceComponent(&render.IntrusionDetectionNamespaceConfiguration{
-		KubernetesProvider:        network.KubernetesProvider,
-		Namespace:                 helper.InstallNamespace(),
-		Tenant:                    tenant,
-		HasNoLicense:              hasNoLicense,
-		SyslogForwardingIsEnabled: syslogForwardingIsEnabled(lc),
-		Azure:                     network.Azure,
+	setUp := render.NewSetup(&render.SetUpConfiguration{
+		OpenShift:       r.provider.IsOpenShift(),
+		Installation:    network,
+		PullSecrets:     pullSecrets,
+		Namespace:       helper.InstallNamespace(),
+		PSS:             getPSS(lc),
+		CreateNamespace: shouldCreateNamespace(tenant, hasNoLicense),
 	})
 	intrusionDetectionComponent := render.IntrusionDetection(intrusionDetectionCfg)
 
@@ -493,7 +493,6 @@ func (r *ReconcileIntrusionDetection) Reconcile(ctx context.Context, request rec
 	}
 
 	components := []render.Component{
-		intrusionDetectionNamespaceComponent,
 		rcertificatemanagement.CertificateManagement(&rcertificatemanagement.Config{
 			Namespace:       helper.InstallNamespace(),
 			TruthNamespace:  helper.TruthNamespace(),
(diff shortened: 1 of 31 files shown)


pkg/render/setup.go
```diff
@@ -0,0 +1,75 @@
...
+// are returned. It is valid to pass nil for 'is' as GetReference accepts the value.
+// ResolveImages must be called before Objects is called for the component.
+func (p *SetUpComponent) ResolveImages(is *operatorv1.ImageSet) error {
+	return nil
+}
+
+// Objects returns the lists of objects in this component that should be created and/or deleted during
+// rendering.
+func (p *SetUpComponent) Objects() (objsToCreate []client.Object, objsToDelete []client.Object) {
+	if !p.cfg.CreateNamespace {
```
tmjd: Do we need to conditionally create the namespace?
I don't think we do, since all creations of the namespace should be using the same owner reference. We should be able to always return the namespace as something to create for the same reason we can return the pull secret.

pkg/controller/manager/manager_controller.go
```diff
@@ -602,7 +602,12 @@ func (r *ReconcileManager) Reconcile(ctx context.Context, request reconcile.Requ
 	}
 
 	// Create a component handler to manage the rendered component.
-	componentHandler := utils.NewComponentHandler(log, r.client, r.scheme, instance)
+	defaultHandler := utils.NewComponentHandler(log, r.client, r.scheme, instance)
+	tenantHandler := utils.NewComponentHandler(log, r.client, r.scheme, tenant)
```
tmjd: tenantHandler should be the same as defaultHandler unless MultiTenant.

pkg/render/setup.go
asincu: I think we do. The tenant namespaces are created externally.

pkg/controller/manager/manager_controller.go
asincu: PolicyRec deployment should be owned by PolicyRec CR even in a multi-tenant setup. But, the pull secrets should be owned by a Tenant.

pkg/controller/logstorage/initializer/initializing_controller.go
```diff
@@ -238,30 +238,37 @@ func (r *LogStorageInitializer) Reconcile(ctx context.Context, request reconcile
...
-		return reconcile.Result{}, err
-	}
+	// Before we can create secrets, we need to ensure the tigera-elasticsearch namespace exists.
+	hdler := utils.NewComponentHandler(reqLogger, r.client, r.scheme, ls)
+	components := []render.Component{render.NewSetup(&render.SetUpConfiguration{
+		OpenShift:       r.provider.IsOpenShift(),
+		Installation:    install,
+		PullSecrets:     pullSecrets,
+		Namespace:       render.ElasticsearchNamespace,
+		PSS:             render.PSSPrivileged,
```
asincu: Seems like external controller and initializer controller create the namespace with different PSS restrictions.

pkg/controller/intrusiondetection/intrusiondetection_controller.go
```diff
@@ -466,13 +466,13 @@ func (r *ReconcileIntrusionDetection) Reconcile(ctx context.Context, request rec
...
-		HasNoLicense:              hasNoLicense,
-		SyslogForwardingIsEnabled: syslogForwardingIsEnabled(lc),
-		Azure:                     network.Azure,
+	setUp := render.NewSetup(&render.SetUpConfiguration{
+		OpenShift:       r.provider.IsOpenShift(),
+		Installation:    network,
+		PullSecrets:     pullSecrets,
+		Namespace:       helper.InstallNamespace(),
+		PSS:             getPSS(lc),
+		CreateNamespace: shouldCreateNamespace(tenant, hasNoLicense),
```
tmjd: Previously no license would result in both the namespace and the operator secrets binding not being created, so what we have currently is not duplicating what we had before.
I'm not sure if what we had previously makes much sense, because why would not having a license lead to to skipping creating the namespace and operator permissions but allow other resources.

pkg/controller/logstorage/initializer/initializing_controller.go
asincu: @caseydavenport @tmjd Seems like we have two different controllers that set PSS differently. Which one should we use: baseline or privileged ?

pkg/controller/logstorage/elastic/external_elastic_controller.go
```diff
@@ -142,10 +142,20 @@ func (r *ExternalESController) Reconcile(ctx context.Context, request reconcile.
 	clusterConfig := relasticsearch.NewClusterConfig(render.DefaultElasticsearchClusterName, ls.Replicas(), logstoragecommon.DefaultElasticsearchShards, flowShards)
 
 	hdler := utils.NewComponentHandler(reqLogger, r.client, r.scheme, ls)
+	setup := render.NewSetup(&render.SetUpConfiguration{
```
tmjd: We don't need to include NewSetup in this controller do we?
It seems like this controller should just check if the ES namespace exists and set degraded if not and wait for it.

pkg/controller/logstorage/linseed/linseed_controller.go
```diff
@@ -457,9 +462,11 @@ func (r *LinseedSubController) Reconcile(ctx context.Context, request reconcile.
 	} else {
 		hdler = utils.NewComponentHandler(reqLogger, r.client, r.scheme, logStorage)
 	}
-	if err := hdler.CreateOrUpdateOrDelete(ctx, linseedComponent, r.status); err != nil {
-		r.status.SetDegraded(operatorv1.ResourceUpdateError, "Error creating / updating / deleting resource", err, reqLogger)
-		return reconcile.Result{}, err
+	for _, component := range []render.Component{setup, linseedComponent} {
```
tmjd, line 465: I think here we should be using a setup handler because we want the pull secret to be owned by the tenant CR.

asincu: Linseed is always owned by the tenant CR.

pkg/controller/logstorage/elastic/external_elastic_controller.go
asincu: Yeah, we can do that.

pkg/controller/intrusiondetection/intrusiondetection_controller.go
asincu: True, thanks for catching this! I also think we should probably skip all resource creation in case we are missing a license.

asincu: I think previously, we would pass the namespace for deletion, if we had no license.

asincu: Not sure if that was the correct behaviour though ...

asincu: Talked offline: intrusion detection controller should close its own controllers.

pkg/controller/logstorage/elastic/external_elastic_controller.go
```diff
@@ -132,6 +133,15 @@ func (r *ExternalESController) Reconcile(ctx context.Context, request reconcile.
 		return reconcile.Result{}, err
 	}
 
+	esNamespace, err := utils.GetIfExists[v1.Namespace](ctx, client.ObjectKey{Name: render.ElasticsearchNamespace}, r.client)
```
tmjd, line 141: I think we need a watch on the ES namespace creation.

tmjd (approved):

PR #3996, opened 2025-06-17
Increase the lifecycle.poll_interval for Elasticsearch

rene-dekker:
```
$ curl -ks  -u"elastic:$token" https://127.0.0.1:9200/_cluster/settings?include_defaults | jq | grep life -A 3
      "lifecycle": {
        "history_index_enabled": "true",
        "poll_interval": "60m",
        "step": {
```

```release-note
Increase the lifecycle.poll_interval for Elasticsearch. In a case where a cluster has many indices, the default setting can cause ES performance issues.
```

### DIFF
--- a/pkg/render/logstorage.go
+++ b/pkg/render/logstorage.go
@@ -743,6 +743,12 @@ func (es *elasticsearchComponent) nodeSetTemplate(pvcTemplate corev1.PersistentV
 		"cluster.max_shards_per_node": 10000,
 		// Disable geoip downloader. This removes an error from the startup logs, because our network policy blocks it.
 		"ingest.geoip.downloader.enabled": false,
+		// (Dynamic, time unit value) How often index lifecycle management checks for indices that meet policy criteria.
+		// Defaults to 10m.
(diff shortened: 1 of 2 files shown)


pkg/render/logstorage.go
```diff
@@ -743,6 +743,12 @@ func (es *elasticsearchComponent) nodeSetTemplate(pvcTemplate corev1.PersistentV
 		"cluster.max_shards_per_node": 10000,
 		// Disable geoip downloader. This removes an error from the startup logs, because our network policy blocks it.
 		"ingest.geoip.downloader.enabled": false,
+		// (Dynamic, time unit value) How often index lifecycle management checks for indices that meet policy criteria.
+		// Defaults to 10m.
```
matthewdupre, line 747: This comment is maybe slightly confusing (since it will now default to 60m)?  Don't mind particularly though.

matthewdupre: I'd be OK with keeping it matching to prevent a merge conflict

tmjd (approved):

PR #4000, opened 2025-06-17
Deduplicate API writes

caseydavenport:
## Description

This PR attempts to minimize the number of API calls that the tigera/operator makes using caches to determine when an update is appropriate.

Namely, we need to handle two cases:
- The object on the cluster has changed since we wrote it, and it needs to be reverted.
- The operator has rendered a new version of the object.

To do this, we cache two things:
- The desired state of the object as intended by the operator.
- The object Generation.

This allows us to spot when either the desired state changes, or when
another client has modified one of our resources.

Fixes https://github.com/projectcalico/calico/issues/10498

## Release Note

```release-note
Reduce the number of unnecessary API calls made by tigera/operator.
```

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/utils/component_test.go
+++ b/pkg/controller/utils/component_test.go
@@ -2042,10 +2042,6 @@ var _ = Describe("Mocked client Component handler tests", func() {
 				InputMutator: setToDS,
 			})
 			mc.Info = append(mc.Info, mockReturn{
-				Method: "Get",
-				Return: nil,
-			})
-			mc.Info = append(mc.Info, mockReturn{
 				Method: "Update",
 				Return: errors.NewConflict(schema.GroupResource{}, "error name", fmt.Errorf("test error message")),
 			})
@@ -2066,10 +2062,6 @@ var _ = Describe("Mocked client Component handler tests", func() {
 				InputMutator: setToDS,
 			})
 			mc.Info = append(mc.Info, mockReturn{
-				Method: "Get",
-				Return: nil,
-			})
-			mc.Info = append(mc.Info, mockReturn{
 				Method:       "Update",
 				Return:       nil,
 				InputMutator: setToDS,
@@ -2078,7 +2070,7 @@ var _ = Describe("Mocked client Component handler tests", func() {
 			err := handler.CreateOrUpdateOrDelete(ctx, fc, nil)
 			Expect(err).To(BeNil())
 
-			Expect(mc.Index).To(Equal(10))
+			Expect(mc.Index).To(Equal(8))
 		})
 
 		It("if Updating a resource conflicts try the update again (retry fails)", func() {
@@ -2098,10 +2090,6 @@ var _ = Describe("Mocked client Component handler tests", func() {
 				InputMutator: setToDS,
 			})
 			mc.Info = append(mc.Info, mockReturn{
-				Method: "Get",
-				Return: nil,
-			})
-			mc.Info = append(mc.Info, mockReturn{
 				Method: "Update",
 				Return: errors.NewConflict(schema.GroupResource{}, "error name", fmt.Errorf("test error message")),
 			})
@@ -2122,16 +2110,12 @@ var _ = Describe("Mocked client Component handler tests", func() {
 				InputMutator: setToDS,
 			})
 			mc.Info = append(mc.Info, mockReturn{
-				Method: "Get",
-				Return: nil,
-			})
-			mc.Info = append(mc.Info, mockReturn{
 				Method: "Update",
 				Return: errors.NewConflict(schema.GroupResource{}, "error name", fmt.Errorf("test error message 2")),
 			})
 
 			err := handler.CreateOrUpdateOrDelete(ctx, fc, nil)
-			Expect(mc.Index).To(Equal(10))
+			Expect(mc.Index).To(Equal(8))
 			Expect(err).NotTo(BeNil())
 		})
 	})
@@ -2195,11 +2179,6 @@ var _ = Describe("Mocked client Component handler tests", func() {
 			})
 
 			mc.Info = append(mc.Info, mockReturn{
-				Method: "Get",
-				Return: nil,
-			})
-
-			mc.Info = append(mc.Info, mockReturn{
 				Method:       "Update",
 				Return:       nil,
 				InputMutator: setToBaseNP,
@@ -2207,7 +2186,7 @@ var _ = Describe("Mocked client Component handler tests", func() {
 
 			err := handler.CreateOrUpdateOrDelete(ctx, fc, nil)
 			Expect(err).To(BeNil())
-			Expect(mc.Index).To(Equal(3))
+			Expect(mc.Index).To(Equal(2))
 		})
 	})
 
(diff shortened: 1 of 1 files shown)


pkg/controller/utils/component.go
```diff
@@ -85,6 +92,98 @@ func (c *componentHandler) SetCreateOnly() {
 	c.createOnly = true
 }
 
+func (c *componentHandler) create(ctx context.Context, obj client.Object, opts ...client.CreateOption) error {
+	// Make a deep copy of the object, so we can stash away the original object in the cache.
+	cached := obj.DeepCopyObject().(client.Object)
```
Brian-McM: If I may be pedantic, _technically_ speaking this isn't a cached object, it's just the copy at this point.

pkg/controller/utils/component.go
```diff
@@ -85,6 +92,98 @@ func (c *componentHandler) SetCreateOnly() {
...
+	// Update the caches so that we don't try to update the object on subsequent reconciliations.
+	dCache.set(cached, obj.GetGeneration())
+	return nil
+}
+
+func (c *componentHandler) update(ctx context.Context, obj client.Object, opts ...client.UpdateOption) error {
+	logCtx := ContextLoggerForResource(c.log, obj)
+	if !c.needsUpdate(ctx, obj) {
+		// The object does not need to be updated, so we can skip it.
+		logCtx.V(2).Info("Object does not need to be updated, skipping")
```
Brian-McM, line 114: Since this is a V(2) log it would be good to include some object info.

pkg/controller/utils/component.go
```diff
@@ -85,6 +92,98 @@ func (c *componentHandler) SetCreateOnly() {
...
+		return nil
+	}
+	logCtx.Info("Object needs to be updated")
```
Brian-McM: Maybe V(2) log as well? I'd think every log in this component handler that isn't a warning or error should probably be at the V(2) level.

pkg/controller/utils/component.go
```diff
@@ -85,6 +92,98 @@ func (c *componentHandler) SetCreateOnly() {
...
+	return nil
+}
+
+func (c *componentHandler) needsUpdate(ctx context.Context, obj client.Object) bool {
+	key := client.ObjectKeyFromObject(obj)
+	cur := obj.DeepCopyObject().(client.Object)
+	err := c.client.Get(ctx, key, cur)
+	if err != nil {
+		// If we can't get the object, assume it needs to be created or updated.
+		return true
```
Brian-McM: Probably want to log the error at some level if we're not returning it.

pkg/controller/utils/component.go
```diff
@@ -85,6 +92,98 @@ func (c *componentHandler) SetCreateOnly() {
...
+	logCtx := ContextLoggerForResource(c.log, obj)
+
+	// Only update the object if one of the following is true:
+	// - the object is not in the generation cache. This means we have not created this object before.
+	// - the generation on the cluster is newer than the generation we have cached. This means that the object has been updated on the cluster.
+	// - the object differs from the last cached version. This means the operator wants to update the object.
+	// This helps prevent us from updating objects unnecessarily.
+	cachedObj, cachedGen, ok := dCache.get(obj)
+	if !ok {
+		// The object is not in the cache, so we need to update it.
```
Brian-McM, line 173: Maybe a V(2) or even V(3) log?

pkg/controller/utils/component.go
caseydavenport: The `logCtx` includes object info already

pkg/controller/utils/component.go
caseydavenport: yeah, I was unsure about this one - with these new changes, this log is only emitted when we actually update an object in the API so it's a lot less noisy. But it still spits out a lot at startup, for example. So might be worth downgrading anyway.

pkg/controller/utils/component.go
```diff
@@ -142,6 +247,9 @@ func (c *componentHandler) createOrUpdateObject(ctx context.Context, obj client.
 	// Check to see if the object exists or not - this determines whether we should create or update.
 	err := c.client.Get(ctx, key, cur)
 	if err != nil {
+		// Invalidate our cached object.
+		dCache.delete(obj)
```
pasanw, line 254: Wondering why we have this cache invalidation here, rather than after the `errors.IsNotFound` check?

pkg/controller/utils/component.go
```diff
@@ -85,6 +92,104 @@ func (c *componentHandler) SetCreateOnly() {
...
+	}
+	logCtx.V(2).Info("Object needs to be updated")
+
+	// Make a deep copy of the object, so we can stash away the original object in the cache.
+	cp := obj.DeepCopyObject().(client.Object)
+
+	// Pass to the client.
+	err := c.client.Update(ctx, obj, opts...)
+	if err != nil {
+		return err
```
pasanw, line 129: Should we invalidate the cache entry here if the error is "not found"?

pkg/controller/utils/component.go
```diff
@@ -85,6 +92,104 @@ func (c *componentHandler) SetCreateOnly() {
...
+}
+
+func (c *componentHandler) needsUpdate(ctx context.Context, obj client.Object) bool {
+	key := client.ObjectKeyFromObject(obj)
+	cur := obj.DeepCopyObject().(client.Object)
+	err := c.client.Get(ctx, key, cur)
+	if err != nil {
+		// If we can't get the object, assume it needs to be created or updated.
+		if errors.IsNotFound(err) {
+			c.log.WithValues("key", key).V(2).Info("Object does not exist, we should create it")
```
pasanw: Should we be invalidating the associated cache entry here?

pasanw (commented):
Just some questions to better understand the cache invalidation strategy

pkg/controller/utils/component.go
caseydavenport: I thought doing this in the face of any error would be more fool proof, but I can't think of an exact scenario where we'd need to do this here. Thoughts?

caseydavenport:
Thanks @pasanw - I think it's a good idea to invalidate in the locations you mentioned! Pushed a new commit.

pkg/controller/utils/component.go
pasanw: My main thought was for scenarios where there is some kind of transient error that resolves on retry. In that case, we would be creating a potential cache miss that seems unnecessary. Happy to leave this choice up to you

pkg/controller/utils/component.go
```diff
@@ -85,6 +92,111 @@ func (c *componentHandler) SetCreateOnly() {
...
+
+	// Invalidate our cached object.
+	dCache.delete(obj)
+	return nil
+}
+
+func (c *componentHandler) needsUpdate(ctx context.Context, obj client.Object) bool {
+	key := client.ObjectKeyFromObject(obj)
+	cur := obj.DeepCopyObject().(client.Object)
+	err := c.client.Get(ctx, key, cur)
```
pasanw: Wondering if you considered removing this `Get`? From my understanding, when we reach this point in the code, we already know that the object exists, otherwise we would have hit the create branch in `createOrUpdateObject`. We could also utilize the generation from the obj passed in. Perhaps you've implemented it this way to ensure we have the freshest information on the object?

caseydavenport: I hadn't considered it - I think it should be possible? It maybe widens the possibility of using out-of-date data, but shouldn't by much. 

That said, having the get here should be more or less free since the client is served from a cache.

I think we'd just need to ensure the generation is set as part of `mergeState` and then this function could rely on that field. Would clean things up here a fair bit!

PR #4004, opened 2025-06-18
Redirect Elasticsearch requests from managed cluster to Guardian inst…

vara2504:
…ead of Linseed external svc

## Release Note

```release-note
Not Applicable
```

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/common/elasticsearch/service.go
+++ b/pkg/render/common/elasticsearch/service.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2021-2024 Tigera, Inc. All rights reserved.
+// Copyright (c) 2021-2025 Tigera, Inc. All rights reserved.
 
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -21,17 +21,30 @@ import (
 )
 
 const (
-	httpsEndpoint       = "https://tigera-secure-es-gateway-http.%s.svc:9200"
-	linseedEndpoint     = "https://tigera-linseed.%s.svc"
-	httpsFQDNEndpoint   = "https://tigera-secure-es-gateway-http.%s.svc.%s:9200"
-	linseedFQDNEndpoint = "https://tigera-linseed.%s.svc.%s"
+	httpsEndpoint     = "https://tigera-secure-es-gateway-http.%s.svc:9200"
+	httpsFQDNEndpoint = "https://tigera-secure-es-gateway-http.%s.svc.%s:9200"
 )
 
-func LinseedEndpoint(osType rmeta.OSType, clusterDomain, namespace string) string {
-	if osType == rmeta.OSTypeWindows {
-		return fmt.Sprintf(linseedFQDNEndpoint, namespace, clusterDomain)
+func LinseedEndpoint(osType rmeta.OSType, clusterDomain, namespace string, isManagedCluster bool, isFluentd bool) string {
+	switch {
+	// In a managed cluster, all Elasticsearch requests to Linseed are redirected via the Guardian service.
+	// Clients using the Linseed client are automatically configured with the correct SNI for certificate validation.
+	// Since Fluentd doesn't use the Linseed client, we expose an external service named "tigera-linseed" that redirects to Guardian.
+	// The Linseed certificate is already configured to accept connections with SNI set to "tigera-linseed".
+	case isManagedCluster && isFluentd:
+		return "https://tigera-linseed"
+
+	// Non-Fluentd components in the managed cluster forward traffic to Guardian
+	case isManagedCluster && osType == rmeta.OSTypeWindows:
+		return fmt.Sprintf("https://guardian.calico-system.svc.%s", clusterDomain)
+	case isManagedCluster:
+		return "https://guardian.calico-system.svc"
+	// Linseed URL used by components in standalone and management cluster.
+	case osType == rmeta.OSTypeWindows:
+		return fmt.Sprintf("https://tigera-linseed.%s.svc.%s", namespace, clusterDomain)
+	default:
+		return fmt.Sprintf("https://tigera-linseed.%s.svc", namespace)
 	}
-	return fmt.Sprintf(linseedEndpoint, namespace)
 }
 
 // GatewayEndpoint returns the endpoint for the Elasticsearch service. For
--- a/pkg/controller/logstorage/managedcluster/managed_cluster_controller_test.go
+++ b/pkg/controller/logstorage/managedcluster/managed_cluster_controller_test.go
@@ -18,13 +18,11 @@ import (
 	"context"
 
 	. "github.com/onsi/ginkgo"
-	. "github.com/onsi/ginkgo/extensions/table"
 	. "github.com/onsi/gomega"
 
 	admissionv1beta1 "k8s.io/api/admissionregistration/v1beta1"
 	appsv1 "k8s.io/api/apps/v1"
 	batchv1 "k8s.io/api/batch/v1"
-	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	storagev1 "k8s.io/api/storage/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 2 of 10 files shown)


pkg/render/common/elasticsearch/service.go
```diff
@@ -25,13 +25,26 @@ const (
 	linseedEndpoint     = "https://tigera-linseed.%s.svc"
 	httpsFQDNEndpoint   = "https://tigera-secure-es-gateway-http.%s.svc.%s:9200"
 	linseedFQDNEndpoint = "https://tigera-linseed.%s.svc.%s"
+
+	// Handle for multi tenant
+	guardianEndpoint       = "https://guardian.calico-system.svc"
+	guardinaFQDNEndpoint   = "https://guardian.calico-system.svc.%s"
+	fluentdLinseedEndpoint = "https://tigera-linseed"
```
vara2504: @caseydavenport Just wondering — how is this handled in a multi-tenant scenario? Does Voltron know which tenant namespace it should route to?
Correct me if I'm wrong, but from what I see, the LinseedEndpoint resolves to `tigera-linseed.tenant-namespace.svc` for requests within the multi-tenant management cluster. However, any request coming from a managed cluster would still resolve to `tigera-linseed.tigera-elasticsearch.svc`

caseydavenport: The certificate that Voltron presents to clients in managed clusters is always valid for the following:

- `tigera-linseed`
- `tigera-linseed.tigera-elasticsearch`
- `tigera-linseed.tigera-elasticsearch.svc`
- `tigera-linseed.tigera-elasticsearch.svc.<clusterdomain>`

It's generated here: https://github.com/caseydavenport/operator/blob/8a04459ed8b533eaad0d3d8ea0d2a2f2de0e00c7/pkg/controller/manager/manager_controller.go#L533-L537

So the managed cluster doesn't need to be aware of the namespace used within the management cluster. Voltron establishes its own mTLS to Linseed within the management cluster, which uses a certificate with SANs that include the actual namespace. 

This mean that clients in managed clusters should be able to simply set their TLS host header to `tigera-linseed` for TLS verification.

pkg/render/common/elasticsearch/service.go
```diff
@@ -25,13 +25,26 @@ const (
 	linseedEndpoint     = "https://tigera-linseed.%s.svc"
 	httpsFQDNEndpoint   = "https://tigera-secure-es-gateway-http.%s.svc.%s:9200"
 	linseedFQDNEndpoint = "https://tigera-linseed.%s.svc.%s"
+
+	// Handle for multi tenant
+	guardianEndpoint       = "https://guardian.calico-system.svc"
+	guardinaFQDNEndpoint   = "https://guardian.calico-system.svc.%s"
```
caseydavenport:
```suggestion
	guardianFQDNEndpoint   = "https://guardian.calico-system.svc.%s"
```

caseydavenport: I think we do have a helper for this already, though, so we probably don't need these variables: https://github.com/tigera/operator/blob/e072c811dd0e6c1111add2787a6c8c98ee42801d/pkg/render/guardian.go#L684-L686

pkg/render/common/elasticsearch/service.go
```diff
@@ -25,13 +25,26 @@ const (
...
-func LinseedEndpoint(osType rmeta.OSType, clusterDomain, namespace string) string {
-	if osType == rmeta.OSTypeWindows {
+func LinseedEndpoint(osType rmeta.OSType, clusterDomain, namespace string, isManagedCluster bool, isFluentd bool) string {
+	switch {
+	case isManagedCluster && isFluentd:
+		return fluentdLinseedEndpoint
+	case isManagedCluster && osType == rmeta.OSTypeWindows:
+		return fmt.Sprintf(guardinaFQDNEndpoint, clusterDomain)
+	case isManagedCluster:
+		return guardianEndpoint
```
caseydavenport: Do these need to be different? Won't the FQDN work on both Windows and Linux?

pkg/render/common/elasticsearch/service.go
```diff
@@ -25,13 +25,26 @@ const (
...
+	guardinaFQDNEndpoint   = "https://guardian.calico-system.svc.%s"
+	fluentdLinseedEndpoint = "https://tigera-linseed"
 )
 
-func LinseedEndpoint(osType rmeta.OSType, clusterDomain, namespace string) string {
-	if osType == rmeta.OSTypeWindows {
+func LinseedEndpoint(osType rmeta.OSType, clusterDomain, namespace string, isManagedCluster bool, isFluentd bool) string {
+	switch {
+	case isManagedCluster && isFluentd:
+		return fluentdLinseedEndpoint
```
caseydavenport: We should have a comment explaining why this is an exception, otherwise this code is pretty opaque to anyone trying to reverse engineer it.

pkg/render/common/elasticsearch/service.go
vara2504: I believe it should work for both Windows and Linux. 

We’ve generally used FQDN only for Windows, and I found this commit where the change was introduced:
https://github.com/tigera/operator/pull/1151 - which you reviewed.

it says:
`For components running on Linux, FQDN endpoints for ES and Kibana aren’t required so revert that change. This smooths out the upgrade for users of custom TLS certs.`

Do you recall the reason behind this decision?

caseydavenport: Hmmm, guessing that our original custom TLS certs stuff didn't include the FQDN in the certificate SANs? not sure to be honest, but probably good enough to leave it as is to avoid unnecessary change.

pkg/render/common/elasticsearch/service.go
```diff
@@ -25,13 +25,30 @@ const (
...
+func LinseedEndpoint(osType rmeta.OSType, clusterDomain, namespace string, isManagedCluster bool, isFluentd bool) string {
+	switch {
+	// In a managed cluster, all Elasticsearch requests to Linseed are redirected via the Guardian service.
+	// Clients using the Linseed client are automatically configured with the correct SNI for certificate validation.
+	// Since Fluentd doesn't use the Linseed client, we expose an external service named "tigera-linseed" that redirects to Guardian.
+	// The Linseed certificate is already configured to accept connections with SNI set to "tigera-linseed".
+	case isManagedCluster && isFluentd:
+		return fluentdLinseedEndpoint
+	case isManagedCluster && osType == rmeta.OSTypeWindows:
+		return fmt.Sprintf(guardianFQDNEndpoint, clusterDomain)
```
rene-dekker: Why do we include cluster domain in the endpoint? Is there a situation where this adds a working configuration that would not be possible without? Could we omit `clusterDomain` from all endpoints in `service.go`?
cc: @caseydavenport 

I'm finding the config a bit confusing, especially in the example where mgmt and mgd have a different cluster domain set.

pkg/render/common/elasticsearch/service.go
```diff
@@ -25,13 +25,30 @@ const (
 	linseedEndpoint     = "https://tigera-linseed.%s.svc"
 	httpsFQDNEndpoint   = "https://tigera-secure-es-gateway-http.%s.svc.%s:9200"
 	linseedFQDNEndpoint = "https://tigera-linseed.%s.svc.%s"
+
+	// Handle for multi tenant
+	guardianEndpoint       = "https://guardian.calico-system.svc"
```
rene-dekker: nit: all the constants that are declared are used exactly one time. The result is that the switch statement is actually very hard to read; if they didn't exist at all, I'd be able to see what the URLs look like as I am reading the code.

rene-dekker (approved):

rene-dekker (approved):

pkg/render/common/elasticsearch/service.go
caseydavenport: @rene-dekker it's because Windows clients require the FQDN for DNS resolution in some capacity (I don't recall the specifics) so the host header is always set to the FQDN when the client is a fluentd pod on a Windows node.

PR #4005, opened 2025-06-19
Align API server resources by renaming tigera prefix to calico

vara2504:
## Release Note

```release-note
TBD
```

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/api/v1/apiserver_types.go
+++ b/api/v1/apiserver_types.go
@@ -28,7 +28,7 @@ type APIServerSpec struct {
 	// +optional
 	Logging *APIServerPodLogging `json:"logging,omitempty"`
 
-	// APIServerDeployment configures the calico-apiserver (or tigera-apiserver in Enterprise) Deployment. If
+	// APIServerDeployment configures the calico-apiserver Deployment. If
 	// used in conjunction with ControlPlaneNodeSelector or ControlPlaneTolerations, then these overrides
 	// take precedence.
 	APIServerDeployment *APIServerDeployment `json:"apiServerDeployment,omitempty"`
--- a/pkg/controller/apiserver/apiserver_controller.go
+++ b/pkg/controller/apiserver/apiserver_controller.go
@@ -161,7 +161,7 @@ func add(c ctrlruntime.Controller, r *ReconcileAPIServer) error {
 	}
 
 	for _, secretName := range []string{
-		"calico-apiserver-certs", "tigera-apiserver-certs",
+		"calico-apiserver-certs",
 		certificatemanagement.CASecretName, render.DexTLSSecretName, monitor.PrometheusClientTLSSecretName,
 	} {
 		if err = utils.AddSecretsWatch(c, secretName, common.OperatorNamespace()); err != nil {
--- a/pkg/render/apiserver.go
+++ b/pkg/render/apiserver.go
@@ -72,7 +72,6 @@ const (
 	TigeraAPIServerQueryServerContainerName ContainerName = "tigera-queryserver"
 
 	calicoAPIServerTLSSecretName                                  = "calico-apiserver-certs"
-	tigeraAPIServerTLSSecretName                                  = "tigera-apiserver-certs"
 	APIServerSecretsRBACName                                      = "tigera-extension-apiserver-secrets-access"
 	MultiTenantManagedClustersAccessClusterRoleName               = "tigera-managed-cluster-access"
 	ManagedClustersWatchClusterRoleName                           = "tigera-managed-cluster-watch"
@@ -106,24 +105,15 @@ var (
 // The following functions are helpers for determining resource names based on
 // the configured product variant.
 func ProjectCalicoAPIServerTLSSecretName(v operatorv1.ProductVariant) string {
-	if v == operatorv1.Calico {
-		return calicoAPIServerTLSSecretName
-	}
-	return tigeraAPIServerTLSSecretName
+	return calicoAPIServerTLSSecretName
 }
 
 func ProjectCalicoAPIServerServiceName(v operatorv1.ProductVariant) string {
-	if v == operatorv1.Calico {
-		return "calico-api"
-	}
-	return "tigera-api"
+	return "calico-api"
 }
 
 func APIServerServiceAccountName(v operatorv1.ProductVariant) string {
-	if v == operatorv1.Calico {
-		return "calico-apiserver"
-	}
-	return "tigera-apiserver"
+	return "calico-apiserver"
 }
 
 func APIServer(cfg *APIServerConfiguration) (Component, error) {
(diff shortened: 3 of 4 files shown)


pkg/controller/apiserver/apiserver_controller.go
```diff
@@ -284,8 +283,8 @@ func (r *ReconcileAPIServer) Reconcile(ctx context.Context, request reconcile.Re
 	}
 
 	// We need separate certificates for OSS vs Enterprise.
```
caseydavenport: Comment is no longer correct

pkg/controller/apiserver/apiserver_controller_test.go
```diff
@@ -114,10 +114,10 @@ var _ = Describe("apiserver controller tests", func() {
 		Expect(cli.Create(ctx, &v3.Tier{ObjectMeta: metav1.ObjectMeta{Name: "allow-tigera"}})).NotTo(HaveOccurred())
 		cryptoCA, err := tls.MakeCA("byo-ca")
 		Expect(err).NotTo(HaveOccurred())
-		apiSecret, err = secret.CreateTLSSecret(cryptoCA, "tigera-apiserver-certs", common.OperatorNamespace(), "key.key", "cert.crt", time.Hour, nil, dns.GetServiceDNSNames(render.ProjectCalicoAPIServerServiceName(operatorv1.TigeraSecureEnterprise), "calico-system", dns.DefaultClusterDomain)...)
+		apiSecret, err = secret.CreateTLSSecret(cryptoCA, "calico-apiserver-certs", common.OperatorNamespace(), "key.key", "cert.crt", time.Hour, nil, dns.GetServiceDNSNames(render.APIServerServiceName, "calico-system", dns.DefaultClusterDomain)...)
 		Expect(err).NotTo(HaveOccurred())
 		Expect(cli.Create(ctx, &operatorv1.Authentication{
-			ObjectMeta: metav1.ObjectMeta{Name: "tigera-secure"},
+			ObjectMeta: metav1.ObjectMeta{Name: "calico-secure"},
```
caseydavenport: this change from "tigera-secure" to "calico-secure" is wrong - this is the name of the API resource that we respect, and isn't the same thing as the name of the objects that are rendered.

Changing this would break all enterprise users, as their existing configuration won't be respected by the operator any more.

pkg/controller/apiserver/apiserver_controller_test.go
```diff
@@ -133,7 +133,7 @@ var _ = Describe("apiserver controller tests", func() {
 		Expect(err).NotTo(HaveOccurred())
 		Expect(cli.Create(ctx, dexSecret)).ToNot(HaveOccurred())
 		Expect(cli.Create(ctx, &corev1.Secret{
-			ObjectMeta: metav1.ObjectMeta{Name: "tigera-oidc-credentials", Namespace: common.OperatorNamespace()},
+			ObjectMeta: metav1.ObjectMeta{Name: "calico-oidc-credentials", Namespace: common.OperatorNamespace()},
```
caseydavenport: We need to be careful not to modify the names of any user inputs - just the objects created by the operator itself. @rene-dekker can confirm but I think this is user input?

User input needs to be backwards compatible - if we do change the name, it needs to be an additional supported name rather than a replacement.

api/v1/apiserver_types.go
```diff
@@ -28,7 +28,7 @@ type APIServerSpec struct {
 	// +optional
 	Logging *APIServerPodLogging `json:"logging,omitempty"`
 
-	// APIServerDeployment configures the calico-apiserver (or tigera-apiserver in Enterprise) Deployment. If
+	// APIServerDeployment configures the calico-apiserver Deployment. If
```
caseydavenport, line 31: See this line: https://github.com/tigera/operator/pull/4005/files#diff-27662d60dc2ac690d14c7bd8dec8c031e702cf9f96e5405bc780de7b29118349R83

We need to be careful about the API inputs here - we can't break existing user configuration / APIServer CRs.

For example, we're changing the name of the queryserver container from `tigera-queryserver` to `calico-queryserver` - that means we likely need to:

- update the API to accept both modes
- treat `tigera-queryserver` as if it were `calico-queryserver` internally
- add a deprecation notice on the old name.

pkg/controller/installation/core_controller.go
```diff
@@ -2069,7 +2069,7 @@ func allowTigeraDefaultDenyForCalicoSystem() *v3.NetworkPolicy {
 			Tier: networkpolicy.TigeraComponentTierName,
 			// Default deny policy should exclude pods with label k8s-app=tigera-apiserver
 			// so the API server remains accessible within the calico-system namespace.
-			Selector: "k8s-app != 'tigera-apiserver'",
+			Selector: "k8s-app != 'tigera-apiserver' && k8s-app != 'calico-apiserver'",
```
caseydavenport: Do we need the `k8s-app != tigera-apiserver` any more?

pkg/render/common/meta/meta.go
```diff
@@ -112,10 +112,7 @@ func SecretsAnnotationHash(secrets ...*corev1.Secret) string {
 
 // APIServerDeploymentName returns the deployment to use for the API server component.
 func APIServerDeploymentName(v operatorv1.ProductVariant) string {
```
caseydavenport: I think this can just be turned into a constant instead of using a function to return a string.

pkg/render/apiserver.go
```diff
@@ -240,12 +212,12 @@ func (c *apiServerComponent) Objects() ([]client.Object, []client.Object) {
 	// These objects are global, and have different names based on Calico or Calico Enterprise.
```
caseydavenport: Comment is no longer correct - I think we can simply move all of these into the instantiation of `globalObjects` above rather than using `append` here?

pkg/render/apiserver.go
```diff
@@ -364,28 +335,12 @@ func (c *apiServerComponent) deploymentSelector() *metav1.LabelSelector {
...
-}
-
 func (c *apiServerComponent) apiServerPodDisruptionBudget() *policyv1.PodDisruptionBudget {
 	maxUnavailable := intstr.FromInt(1)
-	name, _ := c.resourceNameBasedOnVariant("tigera-apiserver", "calico-apiserver")
 	return &policyv1.PodDisruptionBudget{
 		TypeMeta: metav1.TypeMeta{Kind: "PodDisruptionBudget", APIVersion: "policy/v1"},
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      name,
+			Name:      "calico-apiserver",
```
caseydavenport: Do we have a constant for this?

pkg/render/apiserver.go
```diff
@@ -1115,7 +1025,7 @@ func (c *apiServerComponent) apiServerDeployment() *appsv1.Deployment {
 			Selector: c.deploymentSelector(),
 			Template: corev1.PodTemplateSpec{
 				ObjectMeta: metav1.ObjectMeta{
-					Name:      name,
+					Name:      "calico-apiserver",
```
caseydavenport: Probably should have a constant for all of these instances.

api/v1/apiserver_types.go
vara2504: Since queryserver seems to be used only in the Enterprise variant, I'll leave the container name as tigera-queryserver for now to avoid additional changes.

caseydavenport: SGTM!

That said, I think we may have the same issue with `tigera-apiserver`?

vara2504: Nope. Only the deployment names differed between variants; the container name was always calico-apiserver across both EE and OSS, even before.
https://github.com/tigera/operator/blob/master/api/v1/apiserver_types.go#L83

PR #4008, opened 2025-06-20
[RS-2546] Update operator to deploy waf-http-filter in Enterprise

gantony:
## Description

This PR updates the operator to deploy `waf-http-filter` in Enterprise when enabling the Gateway API. 

This will setup the required `waf-http-filter` sidecar container in the envoy-proxy deployment, but envoy won't be configured to use it by default. To do so, the user will need to create a separate `EnvoyExtensionPolicy` to enable the filter and protect the traffic with WAF. 

## Release Note

```release-note
TBD
```

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/gateway_api.go
+++ b/pkg/render/gateway_api.go
@@ -25,9 +25,11 @@ import (
 	operatorv1 "github.com/tigera/operator/api/v1"
 	"github.com/tigera/operator/pkg/common"
 	"github.com/tigera/operator/pkg/components"
+	"github.com/tigera/operator/pkg/ptr"
 	rcomp "github.com/tigera/operator/pkg/render/common/components"
 	rmeta "github.com/tigera/operator/pkg/render/common/meta"
 	"github.com/tigera/operator/pkg/render/common/secret"
+	"github.com/tigera/operator/pkg/render/common/securitycontext"
 	appsv1 "k8s.io/api/apps/v1"
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
@@ -341,6 +343,7 @@ type gatewayAPIImplementationComponent struct {
 	envoyGatewayImage   string
 	envoyProxyImage     string
 	envoyRatelimitImage string
+	wafHTTPFilterImage  string
 }
 
 func GatewayAPIImplementationComponent(cfg *GatewayAPIImplementationConfig) Component {
@@ -365,6 +368,10 @@ func (pr *gatewayAPIImplementationComponent) ResolveImages(is *operatorv1.ImageS
 			return err
 		}
 		pr.envoyRatelimitImage, err = components.GetReference(components.ComponentGatewayAPIEnvoyRatelimit, reg, path, prefix, is)
+		if err != nil {
+			return err
+		}
+		pr.wafHTTPFilterImage, err = components.GetReference(components.ComponentWafHTTPFilter, reg, path, prefix, is)
 		if err != nil {
 			return err
 		}
@@ -635,6 +642,9 @@ func (pr *gatewayAPIImplementationComponent) envoyProxyConfig(defaultName string
 			envoyProxy.Spec.Provider.Kubernetes.EnvoyDaemonSet.Container = &envoyapi.KubernetesContainerSpec{}
 		}
 		envoyProxy.Spec.Provider.Kubernetes.EnvoyDaemonSet.Container.Image = &pr.envoyProxyImage
+
+		// The WAF HTTP filter is not supported when the envoy proxy is deployed as a DaemonSet
+		// as there is no support for init containers in a DaemonSet.
 	} else {
 		// Deployment as a Deployment.
 		if envoyProxy.Spec.Provider.Kubernetes.EnvoyDeployment == nil {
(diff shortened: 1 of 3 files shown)


pkg/render/gateway_api.go
```diff
@@ -402,7 +409,7 @@ func (pr *gatewayAPIImplementationComponent) Objects() ([]client.Object, []clien
 		CreateNamespace(
 			resources.namespace.Name,
 			pr.cfg.Installation.KubernetesProvider,
-			PSSBaseline,
+			PSSPrivileged,
```
caseydavenport: We should comment why we need Privileged here.

pkg/render/gateway_api.go
```diff
@@ -647,6 +658,71 @@ func (pr *gatewayAPIImplementationComponent) envoyProxyConfig(defaultName string
...
+								Name:      "var-log-calico",
+								MountPath: "/var/log/calico",
+							},
+						},
+						SecurityContext: securitycontext.NewRootContext(true),
+					},
+				}
+			}
+
+			if envoyProxy.Spec.Provider.Kubernetes.EnvoyDeployment.Container.VolumeMounts == nil {
```
caseydavenport: What if the mounts are non-nil? How/when can that happen, and do we need to append this volume mount?

pkg/render/gateway_api.go
```diff
@@ -647,6 +658,71 @@ func (pr *gatewayAPIImplementationComponent) envoyProxyConfig(defaultName string
...
+						VolumeSource: corev1.VolumeSource{
+							EmptyDir: &corev1.EmptyDirVolumeSource{},
+						},
+						Name: "waf-http-filter",
+					},
+				}
+			}
+
+			// Add the init container to the deployment
+			if envoyProxy.Spec.Provider.Kubernetes.EnvoyDeployment.InitContainers == nil {
```
caseydavenport: What scenarios can the InitContainers be non-nil? And if they are non-nil, do we not want to insert this init container into them?

pkg/render/gateway_api.go
```diff
@@ -647,6 +658,71 @@ func (pr *gatewayAPIImplementationComponent) envoyProxyConfig(defaultName string
 			envoyProxy.Spec.Provider.Kubernetes.EnvoyDeployment.Container = &envoyapi.KubernetesContainerSpec{}
 		}
 		envoyProxy.Spec.Provider.Kubernetes.EnvoyDeployment.Container.Image = &pr.envoyProxyImage
+
+		if pr.cfg.Installation.Variant == operatorv1.TigeraSecureEnterprise {
+			// Update the envoy proxy deployment to include the WAF HTTP filter
+
+			// Update Pod volumes
+			if envoyProxy.Spec.Provider.Kubernetes.EnvoyDeployment.Pod.Volumes == nil {
```
caseydavenport: Same question here regarding nill-ness

pkg/render/gateway_api.go
gantony: Added a comment. Was discussed here: https://tigera.slack.com/archives/GS3MM6YR0/p1747350565419999?thread_ts=1747349921.824689&cid=GS3MM6YR0

pkg/render/gateway_api.go
gantony: It can be non-nil if we use a custom envoy proxy (functionality recently added by Nell). It's best to handle it so I've updated it with changes to do so (and tests).

pkg/render/gateway_api.go
```diff
@@ -654,6 +662,100 @@ func (pr *gatewayAPIImplementationComponent) envoyProxyConfig(className string,
...
+
+			// Add or update Container volume mount
+			socketVolumeMount := corev1.VolumeMount{
+				Name:      "waf-http-filter",
+				MountPath: "/var/run/waf-http-filter",
+			}
+			hasSocketVolumeMount := false
+			for _, volumeMount := range envoyProxy.Spec.Provider.Kubernetes.EnvoyDeployment.Container.VolumeMounts {
+				if volumeMount.Name == socketVolumeMount.Name {
+					hasSocketVolumeMount = true
```
tmjd, line 717: For the initContainer we ensured the init container wafHTTPFilter was set as we expect. Why don't we do the same here?

pkg/render/gateway_api.go
```diff
@@ -654,6 +662,100 @@ func (pr *gatewayAPIImplementationComponent) envoyProxyConfig(className string,
...
+				VolumeSource: corev1.VolumeSource{
+					EmptyDir: &corev1.EmptyDirVolumeSource{},
+				},
+				Name: "waf-http-filter",
+			}
+			hasLogsVolume := false
+			hasSocketVolume := false
+			for _, volume := range envoyProxy.Spec.Provider.Kubernetes.EnvoyDeployment.Pod.Volumes {
+				if volume.Name == logsVolume.Name {
+					hasLogsVolume = true
```
tmjd, line 744: Same question

pkg/render/gateway_api.go
```diff
@@ -654,6 +662,100 @@ func (pr *gatewayAPIImplementationComponent) envoyProxyConfig(className string,
...
+				}
+				if volume.Name == socketVolume.Name {
+					hasSocketVolume = true
```
tmjd, line 747: Same question

LorcanMcVeigh:
@tmjd I had to create a new PR with these changes because Antony created this on his fork and Although I had permissions to pull i couldn't push. So I created a branch off his on my own fork and re-created the PR with the changes your suggested. Sorry for any confusion : new PR https://github.com/tigera/operator/pull/4032

PR #4015, opened 2025-06-25
Annotate guardian with cluster version info

vara2504:
## Release Note

```release-note
Operator now annotates Guardian pods with cluster version information
```

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/clusterconnection/clusterconnection_controller.go
+++ b/pkg/controller/clusterconnection/clusterconnection_controller.go
@@ -70,7 +70,8 @@ func Add(mgr manager.Manager, opts options.AddOptions) error {
 
 	// Create the reconciler
 	tierWatchReady := &utils.ReadyFlag{}
-	reconciler := newReconciler(mgr.GetClient(), mgr.GetScheme(), statusManager, opts.DetectedProvider, tierWatchReady, opts)
+	clusterInfoWatchReady := &utils.ReadyFlag{}
+	reconciler := newReconciler(mgr.GetClient(), mgr.GetScheme(), statusManager, opts.DetectedProvider, tierWatchReady, clusterInfoWatchReady, opts)
 
 	// Create a new controller
 	c, err := ctrlruntime.NewController(controllerName, mgr, controller.Options{Reconciler: reconciler})
@@ -82,6 +83,10 @@ func Add(mgr manager.Manager, opts options.AddOptions) error {
 		// Watch for changes to License and Tier, as their status is used as input to determine whether network policy should be reconciled by this controller.
 		go utils.WaitToAddLicenseKeyWatch(c, opts.K8sClientset, log, nil)
 		go utils.WaitToAddTierWatch(networkpolicy.TigeraComponentTierName, c, opts.K8sClientset, log, tierWatchReady)
+
+		// Watch for changes to ClusterInformation, as Guardian needs to restart the tunnel
+		// if the cluster's CNX version changes.
+		go utils.WaitToAddClusterInformationWatch(c, opts.K8sClientset, log, clusterInfoWatchReady)
 
 		go utils.WaitToAddNetworkPolicyWatches(c, opts.K8sClientset, log, []types.NamespacedName{
 			{Name: render.GuardianPolicyName, Namespace: render.GuardianNamespace},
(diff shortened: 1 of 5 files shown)


pkg/controller/clusterconnection/clusterconnection_controller.go
```diff
@@ -408,6 +423,15 @@ func (r *ReconcileConnection) Reconcile(ctx context.Context, request reconcile.R
 		// License becomes available. Therefore, if we fail to query the Tier, we exclude NetworkPolicy from reconciliation
 		// and tolerate errors arising from the Tier not being created.
 		includeEgressNetworkPolicy = tierAvailable && licenseActive
+
+		clusterInformation, err := utils.FetchClusterInformation(ctx, r.cli)
+		if err != nil {
+			r.status.SetDegraded(operatorv1.ResourceReadError, "Error querying clusterInformation", err, reqLogger)
+			return reconcile.Result{}, err
+		}
+		if len(clusterInformation.Spec.CNXVersion) != 0 {
```
caseydavenport: nit, but I don't think we need this branch check - if the length is 0, then we'll get `cnxVersion = ""` which is the same as if we did nothing.

pkg/controller/clusterconnection/clusterconnection_controller.go
```diff
@@ -422,6 +446,7 @@ func (r *ReconcileConnection) Reconcile(ctx context.Context, request reconcile.R
 		TrustedCertBundle:           trustedBundle,
 		ManagementClusterConnection: managementClusterConnection,
 		GuardianClientKeyPair:       guardianKeyPair,
+		CnxVersion:                  cnxVersion,
```
caseydavenport:
```suggestion
		Version:                  cnxVersion,
```

`CNX` isn't really a thing. Best to avoid using it where we can. I know it's still in the ClusterInformation API (ideally we'd get rid of / rename that field at some point).

pkg/render/guardian.go
```diff
@@ -454,6 +457,10 @@ func (c *GuardianComponent) volumeMounts() []corev1.VolumeMount {
 func (c *GuardianComponent) annotations() map[string]string {
 	annotations := c.cfg.TrustedCertBundle.HashAnnotations()
 	annotations["hash.operator.tigera.io/tigera-managed-cluster-connection"] = rmeta.AnnotationHash(c.cfg.TunnelSecret.Data)
+
+	if len(c.cfg.CnxVersion) != 0 {
+		annotations["hash.operator.tigera.io/cnx-version"] = rmeta.AnnotationHash([]byte(c.cfg.CnxVersion))
```
caseydavenport: Do we need to hash the version? why not just use the version string as-is?

caseydavenport:
```suggestion
		annotations["hash.operator.tigera.io/version"] = rmeta.AnnotationHash([]byte(c.cfg.CnxVersion))
```

pkg/render/guardian.go
```diff
@@ -113,6 +113,9 @@ type GuardianConfiguration struct {
 	PodProxies []*httpproxy.Config
 
 	GuardianClientKeyPair certificatemanagement.KeyPairInterface
+
+	//CnxVersion stores the manged cluster's cnx Version information
```
caseydavenport:
```suggestion
	// Version stores the version of the cluster, as reported by the ClusterInformation object. It is used to restart 
	// guardian when the version changes, which triggers the management cluster to re-check for version skew. 
```

pkg/controller/clusterconnection/clusterconnection_controller.go
caseydavenport: What about OSS clusters - they don't have this field. Should we use the OSS version for them?

pkg/controller/clusterconnection/clusterconnection_controller.go
```diff
@@ -89,6 +90,10 @@ func Add(mgr manager.Manager, opts options.AddOptions) error {
 		})
 	}
 
+	// Watch for changes to ClusterInformation, as Guardian needs to restart the tunnel
+	// if the cluster's CNX version changes.
```
caseydavenport:
```suggestion
	// if the cluster's version changes.
```

caseydavenport (approved):
LGTM, one minor comment.

PR #4019, opened 2025-06-26
Bugfix: Provision EnvoyProxy with the gateway class name

nelljerram:
There was a bug in the logic of cleaning up unwanted GatewayClasses and EnvoyProxies: if a custom EnvoyProxy had a name different from that of the GatewayClass using it, the renderer would accidentally delete the EnvoyProxy again from the tigera-gateway namespace, almost immediately after provisioning it, making the GatewayClass inoperable.

As part of this fix, we now create EnvoyProxies in the tigera-gateway namespace with the same name as that of the GatewayClass.  In other words:

    custom EnvoyProxy exists with <user-name> in <user-namespace>

    GatewayClasses[*].Name = <custom-class-name>
    GatewayClasses[*].EnvoyProxyRef = {<user-name>, <user-namespace>}

    =>

    EnvoyProxy provisioned with <custom-class-name> in tigera-gateway
    GatewayClass provisioned with <custom-class-name> in tigera-gateway

whereas previously the penultimate line would have been

    EnvoyProxy provisioned with <user-name> in tigera-gateway

My previous thinking was "if the user has chosen a name, let's respect that." My thinking now is:

1. That leads to bugs like the one here.

2. If we respect `<user-name>`, it's possible that `<user-name>` for the EnvoyProxy of a GatewayClass with EnvoyProxyRef will clash with `<custom-class-name>` for another GatewayClass without EnvoyProxyRef.  Let's just use `<custom-class-name>` in all cases.

### DIFF
--- a/pkg/controller/gatewayapi/gatewayapi_controller.go
+++ b/pkg/controller/gatewayapi/gatewayapi_controller.go
@@ -23,6 +23,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/utils/set"
 	gapi "sigs.k8s.io/gateway-api/apis/v1"
 	"sigs.k8s.io/yaml" // gopkg.in/yaml.v2 didn't parse all the fields but this package did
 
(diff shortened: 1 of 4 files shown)


nelljerram:
> Looks good to me! although I'm torn between suggesting changing the log level to V(2) from V(1) just because I think it will be spurious enough..

Thanks. I think that makes sense for the "GatewayAPI rendering" one, because that one gets emitted every time the renderer is called.  Perhaps should be even more than 2?  I'm not really familiar with our levels in this code - WDYT?

But all the others are 1:1 with explicit user actions, and should not repeat once that action has been effectuated.  (For example, an outdated GatewayClass only needs deleting once.). So I think best to leave those others as V(1).

caseydavenport (approved):

PR #4025, opened 2025-06-27
Ev 5885 fix security context

rene-dekker:
Fix security context for csr init container

The CSR init container should have the same scc as the container using
the private key, cert and ca files that are created. Otherwise, 
permission errors will prevent the main container from starting.

### DIFF
--- a/pkg/render/apiserver.go
+++ b/pkg/render/apiserver.go
@@ -32,6 +32,7 @@ import (
 
 	v3 "github.com/tigera/api/pkg/apis/projectcalico/v3"
 	"github.com/tigera/api/pkg/lib/numorstring"
+
 	operatorv1 "github.com/tigera/operator/api/v1"
 	"github.com/tigera/operator/pkg/common"
 	"github.com/tigera/operator/pkg/components"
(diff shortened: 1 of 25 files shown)


pkg/render/apiserver.go
```diff
@@ -985,16 +990,20 @@ func (c *apiServerComponent) apiServerDeployment() *appsv1.Deployment {
...
-		initContainer := c.cfg.TLSKeyPair.InitContainer(APIServerNamespace)
-		initContainer.Name = fmt.Sprintf("%s-%s", CalicoAPIServerTLSSecretName, certificatemanagement.CSRInitContainerName)
-		initContainers = append(initContainers, initContainer)
-	}
+		initContainerAS := c.cfg.TLSKeyPair.InitContainer(APIServerNamespace, c.apiServerContainer().SecurityContext)
+		initContainerAS.Name = fmt.Sprintf("%s-%s", CalicoAPIServerTLSSecretName, certificatemanagement.CSRInitContainerName)
 
-	annotations := map[string]string{
-		c.cfg.TLSKeyPair.HashAnnotationKey(): c.cfg.TLSKeyPair.HashAnnotationValue(),
+		initContainerQS := c.cfg.QueryServerTLSKeyPairCertificateManagementOnly.InitContainer(APIServerNamespace, c.queryServerContainer().SecurityContext)
```
hjiawei: nit. The AS and QS maybe unclear to readers without proper context. It is recommended to use the full terms when naming these init container variables.

hjiawei (approved):
One small nitpicking regarding the variable naming. Otherwise, LGTM.

PR #4046, opened 2025-07-16
add environment variables for logger gateway name and namespace

electricjesus:
## Description

the gateway waf http filter depends on these env vars. the content of these env vars will be used in the filter's log entries for providing context on what the gateway pod name and gateway pod namespace it is currently attached to. 

if these env vars are left out, gateway waf filter logs will just have blank field entries for gateway pod name and namespace.

## Release Note

```release-note
TBD
```

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/crds/calico/crd.projectcalico.org_bgppeers.yaml
+++ b/pkg/crds/calico/crd.projectcalico.org_bgppeers.yaml
@@ -35,6 +35,9 @@ spec:
                   type: array
                 keepOriginalNextHop:
                   type: boolean
+                localASNumber:
+                  format: int32
+                  type: integer
                 localWorkloadSelector:
                   type: string
                 maxRestartTime:


LorcanMcVeigh (approved):
LGTM. 

Nit: 
I think we could probably just ignore the if statement where it checks the values of the fields. We would just set the value of those fields to be whatever we want them to be. But it's done through the file so good to just stick with convention for now.

pkg/render/gateway_api.go
```diff
@@ -724,6 +748,29 @@ func (pr *gatewayAPIImplementationComponent) envoyProxyConfig(className string,
 				envoyProxy.Spec.Provider.Kubernetes.EnvoyDeployment.Container.VolumeMounts = append(envoyProxy.Spec.Provider.Kubernetes.EnvoyDeployment.Container.VolumeMounts, socketVolumeMount)
 			}
 
+			hasGatewayNameEnv := false
```
caseydavenport: Could you add a comment explaining what this logic is meant to do? I am perhaps I am being a bit dense but I can't quite follow the intention here.

electricjesus: thanks @caseydavenport ! I took a look at this again and apparently I did too much here. This bit just distributes the env vars much like how the proxy overrides work. 

turned out i only needed it in the waf http filter initContainer instead. I have updated the code and tests accordingly

caseydavenport: Excellent, thank you!

caseydavenport:
@electricjesus I think you just need to do a `make gen-versions` and then CI will pass

PR #4063, opened 2025-07-25
Gateway API: allow for OpenShift 4.19 CRD lockdown

nelljerram:
OpenShift 4.19+ pre-installs some of the Gateway CRDs, but not all of them, and has a webhook that prevents us from installing the ones that OpenShift is missing.  To work with this we distinguish between "essential" and "optional" CRDs.  The "essential" set must be a subset of those that OpenShift installs and/or allows to be installed, and must also suffice for all of the Gateway-related feature that we consider important as part of Calico; and this controller will report an error and degraded status if any of those do not already exist and cannot be installed.  The "optional" set is everything else that we would ideally install, to provide more options to our users; but this controller will only warn if any of those cannot be installed (and do not already exist).

Fixes https://tigera.atlassian.net/browse/RS-2689
Fixes https://tigera.atlassian.net/browse/RS-2690

## Release Note

```release-note
Accommodate OpenShift 4.19, which only provides some of the Gateway API CRDs itself and prevents our operator from installing the others.  It means that our Gateway API support is less rich on OpenShift 4.19 than on other platforms, but that is better than not working at all.
```

### DIFF
--- a/pkg/controller/gatewayapi/gatewayapi_controller.go
+++ b/pkg/controller/gatewayapi/gatewayapi_controller.go
@@ -192,12 +192,22 @@ func (r *ReconcileGatewayAPI) Reconcile(ctx context.Context, request reconcile.R
 	// so that the CRDs are left in place even if the GatewayAPI CR is removed again.  This is
 	// in case the customer uses a second (or more) implementation of the Gateway API in
 	// addition to the one that we are providing here.
-	crdComponent := render.NewPassthrough(render.GatewayAPICRDs(log)...)
+	//
+	// OpenShift 4.19+ pre-installs some of the Gateway CRDs, but not all of them, and has a
+	// webhook that prevents us from installing the ones that OpenShift is missing.  To work
+	// with this we distinguish between "essential" and "optional" CRDs.  The "essential" set
+	// must be a subset of those that OpenShift installs and/or allows to be installed, and must
+	// also suffice for all of the Gateway-related feature that we consider important as part of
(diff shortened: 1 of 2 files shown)


pkg/controller/gatewayapi/gatewayapi_controller.go
```diff
@@ -233,9 +243,13 @@ func (r *ReconcileGatewayAPI) Reconcile(ctx context.Context, request reconcile.R
 		}
 	}
 	if err != nil && !errors.IsAlreadyExists(err) {
-		r.status.SetDegraded(operatorv1.ResourceCreateError, "Error rendering GatewayAPI CRDs", err, log)
+		r.status.SetDegraded(operatorv1.ResourceCreateError, "Error rendering essential GatewayAPI CRDs", err, log)
 		return reconcile.Result{}, err
 	}
+	err = handler.CreateOrUpdateOrDelete(ctx, render.NewPassthrough(optionalCRDs...), nil)
+	if err != nil && !errors.IsAlreadyExists(err) {
```
tmjd, line 250: Does CreateOrUpdateOrDelete ever return an error AlreadyExists? That doesn't seem like an error it would ever return.

nelljerram: It can do with a "create-only" handler; please see https://github.com/tigera/operator/blob/master/pkg/controller/utils/component.go#L61-L69

nelljerram:
Closing because we decided not to include this in Enterprise 3.22 EP1.  A similar fix for master is in #4068

PR #4064, opened 2025-07-25
[CORE-11605] Rename init container `mount-bpffs` to `ebpf-bootstrap`

lucastigera:
## Description

The init container `mount-bpffs` has been renamed to `ebpf-bootstrap` to better reflect its purpose, as it will handle more than just mounting the BPF filesystem. 
 - The old name is still supported in the `CalicoNodeDaemonSet` CRD to maintain backward compatibility.
 -  A validation check has been added to prevent both `mount-bpffs` and `ebpf-bootstrap` from being specified at the same time.

**Note:**
The `mount-bpffs` value was removed from the `CalicoNodeWindowsDaemonSet` CRD. Since eBPF is not supported on Windows and this init container was not created by the Operator, it's more appropriate to remove it entirely rather than keep the old, the new, or even both values.

## Release Note

```release-note
The init container `mount-bpffs` is now named `ebpf-bootstrap` to reflect its broader responsibilities. No impact on functionality.
```

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/api/v1/calico_node_types.go
+++ b/api/v1/calico_node_types.go
@@ -38,8 +38,8 @@ type CalicoNodeDaemonSetContainer struct {
 // CalicoNodeDaemonSetInitContainer is a calico-node DaemonSet init container.
 type CalicoNodeDaemonSetInitContainer struct {
 	// Name is an enum which identifies the calico-node DaemonSet init container by name.
-	// Supported values are: install-cni, hostpath-init, flexvol-driver, ebpf-bootstrap, node-certs-key-cert-provisioner, calico-node-prometheus-server-tls-key-cert-provisioner
-	// +kubebuilder:validation:Enum=install-cni;hostpath-init;flexvol-driver;ebpf-bootstrap;node-certs-key-cert-provisioner;calico-node-prometheus-server-tls-key-cert-provisioner
+	// Supported values are: install-cni, hostpath-init, flexvol-driver, mount-bpffs (legacy value, replaced by ebpf-bootstrap), ebpf-bootstrap, node-certs-key-cert-provisioner, calico-node-prometheus-server-tls-key-cert-provisioner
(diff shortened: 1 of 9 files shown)


api/v1/calico_node_types.go
```diff
@@ -38,8 +38,8 @@ type CalicoNodeDaemonSetContainer struct {
 // CalicoNodeDaemonSetInitContainer is a calico-node DaemonSet init container.
 type CalicoNodeDaemonSetInitContainer struct {
 	// Name is an enum which identifies the calico-node DaemonSet init container by name.
-	// Supported values are: install-cni, hostpath-init, flexvol-driver, mount-bpffs, node-certs-key-cert-provisioner, calico-node-prometheus-server-tls-key-cert-provisioner
-	// +kubebuilder:validation:Enum=install-cni;hostpath-init;flexvol-driver;mount-bpffs;node-certs-key-cert-provisioner;calico-node-prometheus-server-tls-key-cert-provisioner
+	// Supported values are: install-cni, hostpath-init, flexvol-driver, ebpf-bootstrap, node-certs-key-cert-provisioner, calico-node-prometheus-server-tls-key-cert-provisioner
+	// +kubebuilder:validation:Enum=install-cni;hostpath-init;flexvol-driver;ebpf-bootstrap;node-certs-key-cert-provisioner;calico-node-prometheus-server-tls-key-cert-provisioner
```
rene-dekker: We need to support both for these overrides now, removing a previously supported enum value is dangerous.

In the controller I think we should check for both. If both are set, we should degrade. If one is set, then we use that one for propagating values to the pod spec.

lucastigera: Good point!
Updated to maintain backward compatibility with `mount-bpffs` name.

PR #4068, opened 2025-07-31
Gateway API: allow for OpenShift 4.19 CRD lockdown

nelljerram:
OpenShift 4.19+ pre-installs some of the Gateway CRDs, but not all of them, and has a webhook that prevents us from installing the ones that OpenShift is missing.  To work with this we distinguish between "essential" and "optional" CRDs.  The "essential" set must be a subset of those that OpenShift installs and/or allows to be installed, and must also suffice for all of the Gateway-related feature that we consider important as part of Calico; and this controller will report an error and degraded status if any of those do not already exist and cannot be installed.  The "optional" set is everything else that we would ideally install, to provide more options to our users; but this controller will only warn if any of those cannot be installed (and do not already exist).

Fixes https://tigera.atlassian.net/browse/RS-2689
Fixes https://tigera.atlassian.net/browse/RS-2690

## Release Note

```release-note
Accommodate OpenShift 4.19, which only provides some of the Gateway API CRDs itself and prevents our operator from installing the others.  It means that our Gateway API support is less rich on OpenShift 4.19 than on other platforms, but that is better than not working at all.
```

### DIFF
--- a/pkg/render/gateway_api.go
+++ b/pkg/render/gateway_api.go
@@ -21,7 +21,6 @@ import (
 	"sync"
 
 	envoyapi "github.com/envoyproxy/gateway/api/v1alpha1"
-	"github.com/go-logr/logr"
 	operatorv1 "github.com/tigera/operator/api/v1"
 	"github.com/tigera/operator/pkg/common"
 	"github.com/tigera/operator/pkg/components"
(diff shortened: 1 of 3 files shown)


nelljerram:
Docs PR is here: https://github.com/tigera/docs/pull/2205

pkg/render/gateway_api.go
```diff
@@ -336,17 +335,31 @@ func GatewayAPIResourcesGetter() func() *gatewayAPIResources {
...
-	gatewayAPICRDs := make([]client.Object, 0, len(resources.k8sCRDs)+len(resources.envoyCRDs))
+	essentialCRDs := make([]client.Object, 0, len(resources.k8sCRDs)+len(resources.envoyCRDs))
+	optionalCRDs := make([]client.Object, 0, len(resources.k8sCRDs)+len(resources.envoyCRDs))
 	for _, crd := range resources.k8sCRDs {
-		gatewayAPICRDs = append(gatewayAPICRDs, crd.DeepCopyObject().(client.Object))
+		if provider.IsOpenShift() {
+			// OpenShift 4.19+ restricts the Gateway CRDs that we can install, so report
+			// that only some of them are essential.
+			switch strings.TrimSuffix(crd.Name, ".gateway.networking.k8s.io") {
+			case "gatewayclasses", "gateways", "httproutes", "referencegrants":
```
Brian-McM, line 347: This is a pretty good way to check a value is in a slice of values, not sure why I never though of that before.

Brian-McM (approved):

electricjesus (approved):

PR #4085, opened 2025-08-07
Allow the user to configure impersonation permissions

rene-dekker:
Example spec required for an ldap user called test-user@example.com in group readers

```
kind: ManagementClusterConnection
metadata:
  name: tigera-secure
spec:
  impersonation:
    groups:
    - readers
    users:
    - test-user@example.com
```

Example spec for a service account called jane in the default namespace:

```
kind: ManagementClusterConnection
metadata:
  name: tigera-secure
spec:
  impersonation:
    groups:
    - system:serviceaccounts
    - system:serviceaccounts:default
    - system:authenticated
    serviceAccounts:
    - jane
```
Since service accounts have a number of standard groups in their tokens, you quickly need more groups permissions that you'd initially think of. Perhaps a further iteration could allow for certain groups to be excluded from the tunnel. However, that would not affect this repo.

```release-note
The impersonation permissions on guardian are made configurable through the ManagementClusterConnection resource.
```


pkg/render/guardian.go
```diff
@@ -238,13 +239,38 @@ func (c *GuardianComponent) serviceAccount() *corev1.ServiceAccount {
...
 func (c *GuardianComponent) clusterRole() *rbacv1.ClusterRole {
 	var policyRules []rbacv1.PolicyRule
-
 	if c.cfg.Installation.Variant == operatorv1.TigeraSecureEnterprise {
-		policyRules = append(policyRules, rbacv1.PolicyRule{
-			APIGroups: []string{""},
-			Resources: []string{"users", "groups", "serviceaccounts"},
-			Verbs:     []string{"impersonate"},
-		})
+		impersonation := c.cfg.ManagementClusterConnection.Spec.Impersonation
```
caseydavenport, line 243: Looks like this field only takes effect for enterprise - but we don't have validation preventing this being set for OSS Calico. 

Should we apply it to OSS as well? Or add the validation?

rene-dekker: Great catch, I hadn't considered that. Let me update the CRD text and validation in the controller.

rene-dekker: I have changed the explanation in the CRD. Rather than degrading I set it to nil for Variant=Calico and explained why in the code comment.

pkg/controller/utils/utils.go
```diff
@@ -501,10 +501,13 @@ type Defaultable[E any, O ClientObjType[E]] interface {
 }
 
 // ApplyDefaults sets any defaults that haven't been set on the given object and writes it to the k8s server.
-func ApplyDefaults[E any, O ClientObjType[E], D Defaultable[E, O]](ctx context.Context, c client.Client, obj D) error {
+// funcs can be passed in for filling defaults that cannot be derived from D alone.
+func ApplyDefaults[E any, O ClientObjType[E], D Defaultable[E, O]](ctx context.Context, c client.Client, obj D, funcs ...func()) error {
 	preDefaultPatchFrom := client.MergeFrom(obj.DeepCopy())
 	obj.FillDefaults()
-
+	for _, fn := range funcs {
```
caseydavenport: I don't quite follow this change - I thought the point of this function was to enable a `FillDefaults` bound function on the object to keep defaulting in one place. Doesn't adding an optional set of defaulting functions defeat the purpose of this? https://github.com/tigera/operator/blob/fa6fc3c7c39f36c7995b34d6024e6eabfaaebdd6/api/v1/managementclusterconnection_types.go#L97-L104

I suspect the problem is that `FillDefaults` is missing context necessary for conditional defautling, but in that case I think the fix should be to provide that context as input to the defaulting function and not to split defaulting into multiple places.

api/v1/managementclusterconnection_types.go
```diff
@@ -33,6 +33,32 @@ type ManagementClusterConnectionSpec struct {
 
 	// GuardianDeployment configures the guardian Deployment.
 	GuardianDeployment *GuardianDeployment `json:"guardianDeployment,omitempty"`
+
+	// Impersonation configures the RBAC impersonation permissions for the guardian deployment. This field is not
+	// applicable to installation variant Calico as no impersonation is ever used. Otherwise, if this field is left nil,
+	// a default set of permissions will be applied.
+	//
+	// WARNING: If this field is specified, it completely replaces the default permissions.
+	// For example, providing an empty `impersonation: {}` block will result in the guardian
```
caseydavenport:
```suggestion
	// For example, providing an empty `impersonation: {}` block will result in guardian
```

api/v1/managementclusterconnection_types.go
```diff
@@ -33,6 +33,32 @@ type ManagementClusterConnectionSpec struct {
...
+	// having NO impersonation permissions. Similarly, if you specify `users` but omit `groups`,
+	// the guardian will lose its default permissions to impersonate groups.
```
caseydavenport:
```suggestion
	// guardian will lose its default permissions to impersonate groups.
```

caseydavenport (approved):
Two typos but otherwise LGTM

PR #4089, opened 2025-08-12
Update apiserver anti-affinity; remove old pods to prevent hostNetwork port conflicts

vara2504:
## Description

Update apiserver anti-affinity to exclude new/legacy namespaces; clean up old pods to prevent hostNetwork port conflicts during upgrade

## Release Note

```release-note
Not applicable
```

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/apiserver.go
+++ b/pkg/render/apiserver.go
@@ -132,7 +132,6 @@ type APIServerConfiguration struct {
 	MultiTenant                 bool
 	KeyValidatorConfig          authentication.KeyValidatorConfig
 	KubernetesVersion           *common.VersionInfo
-	CanCleanupOlderResources    bool
 
 	// When certificate management is enabled, we need a separate init container to create a cert, running
 	// with the same permissions as query server.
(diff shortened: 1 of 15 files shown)


pkg/render/apiserver.go
```diff
@@ -1054,7 +1053,7 @@ func (c *apiServerComponent) apiServerDeployment() *appsv1.Deployment {
 	}
 
 	if c.cfg.Installation.ControlPlaneReplicas != nil && *c.cfg.Installation.ControlPlaneReplicas > 1 {
-		d.Spec.Template.Spec.Affinity = podaffinity.NewPodAntiAffinity(APIServerName, APIServerNamespace)
+		d.Spec.Template.Spec.Affinity = podaffinity.NewPodAntiAffinity(APIServerName, []string{APIServerNamespace, "tigera-system", "calico-apiserver"})
```
tmjd, line 1056: Didn't the APIServerName change also?

vara2504: Good catch. 
Yes , APIServer name changes when upgrading from enterprise (with the apiserver named tigera-apiserver) to enterprise (with the new apiserver named calico-apiserver). However, it should not be a problem when upgrading from OSS (calico-apiserver) to enterprise (calico-apiserver)

tmjd:
In general this is quite a bit of changes for, what I am thinking of as, a hotfix release. I'd prefer if we could have a small change to get a hotfix quickly out and save the more extensive change for a follow on.

pkg/render/apiserver.go
vara2504: As discussed with @tmjd , we might not need this change since we are removing the older apiserver anyways.

PR #4091, opened 2025-08-14
[RS-2539] Add L7 to envoy proxy resource

LorcanMcVeigh:
## Description

Add l7 log collector to the ingress gateway API deployment for enterprise. 

## Release Note

```release-note
Export l7 access logs from Ingress Gateway 
```

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/gateway_api.go
+++ b/pkg/render/gateway_api.go
@@ -743,7 +743,7 @@ func (pr *gatewayAPIImplementationComponent) envoyProxyConfig(className string, 
 				Image: pr.L7LogCollectorImage,
 				Env: []corev1.EnvVar{
 					{
-						Name:  "READ_FILES",
+						Name:  "ENABLE_LOG_TAIL",
 						Value: "true",
 					},
 					{


pkg/render/gateway_api.go
```diff
@@ -738,27 +780,60 @@ func (pr *gatewayAPIImplementationComponent) envoyProxyConfig(className string,
...
 			}
 			if !hasWAFHTTPFilter {
 				envoyProxy.Spec.Provider.Kubernetes.EnvoyDeployment.InitContainers = append(envoyProxy.Spec.Provider.Kubernetes.EnvoyDeployment.InitContainers, wafHTTPFilter)
 			}
 
+			if !hasL7LogCollector {
+				envoyProxy.Spec.Provider.Kubernetes.EnvoyDeployment.InitContainers = append(envoyProxy.Spec.Provider.Kubernetes.EnvoyDeployment.InitContainers, l7LogCollector)
+			}
+
+			wafSocketFilterName := "waf-http-filter"
```
gantony: I would leave that as it was. You only use it twice and it's somewhat misleading/confusing as it does not reference the same thing. If you want a var I would tweak the name to remove `Socket` from it.

pkg/render/gateway_api.go
```diff
@@ -377,6 +377,7 @@ type gatewayAPIImplementationComponent struct {
 	envoyProxyImage     string
 	envoyRatelimitImage string
 	wafHTTPFilterImage  string
+	L7LogCollectorImage string
```
rene-dekker, line 383: We require now that every container can have its resource requirements and limits set. Could you add that to your PR?

LorcanMcVeigh: yeah no problem. Is there an example for what the limits are for the containers?

rene-dekker: This CRD has an example: https://github.com/tigera/operator/blob/master/api/v1/packetcaptureapi_types.go

If you don't know the right values it's ok, as long as a user is able to set/configure them.

LorcanMcVeigh: I changed it so we don't overwrite the `Resource` field of the container if the user has set it. Hope that's acceptable

rene-dekker: I meant that we should add fields to the CRD that allow you to set resources and limits.

See also this page: https://docs.tigera.io/calico-enterprise/latest/reference/component-resources/configure-resources#configure-resources

Our cloud solution and also many users require all pods to be configurable in their resource usage.

LorcanMcVeigh: Oh, This is isn't our CRD. This is Envoy Proxy CR https://gateway.envoyproxy.io/docs/api/extension_types/#envoyproxy

rene-dekker (approved):

pkg/render/gateway_api.go
rene-dekker: It looks like none of the images have configurable resource requirements and limits, so I'll resolve this for now.

pkg/render/gateway_api.go
```diff
@@ -738,27 +783,63 @@ func (pr *gatewayAPIImplementationComponent) envoyProxyConfig(className string,
...
+
+			l7SocketVolumeMount := corev1.VolumeMount{
+				Name:      accessLogsName,
+				MountPath: "/access_logs",
+			}
+
+			hasWAFFilterSocketVolumeMount := false
+			hasAccessLogsVolumeMount := false
+
 			for i, volumeMount := range envoyProxy.Spec.Provider.Kubernetes.EnvoyDeployment.Container.VolumeMounts {
```
rene-dekker, line 826: It is generally not recommended to modify elements of an array as you are iterating over it.

PR #4099, opened 2025-08-19
Add dashboard sidecar for Enterprise

xiumozhan:
## Description

Added support for embedding a dashboard query API sidecar in the manager deployment when running the enterprise variant, wiring in a `TLSTerminatedRoute` and local forwarding to `8444`

Introduced a dedicated container configuration with required environment variables, readiness/liveness probes, and mounts for Linseed credentials and certificates

Broadened manager RBAC to include `managedclusters` access, and multi-cluster log permissions for Linseed APIs

Relaxed Voltron route builder to allow `TLSTerminatedRoute` entries without a CA bundle, enabling plain HTTP destinations

## Release Note
```release-note
The operator is now responsible for installing custom dashboard in the Enterprise environment.
```

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/manager.go
+++ b/pkg/render/manager.go
@@ -89,10 +89,15 @@ const (
 
 // ManagementClusterConnection configuration constants
 const (
+	ManagerName              = "tigera-manager"
+	UIAPIsName               = "tigera-ui-apis"
 	VoltronName              = "tigera-voltron"
 	VoltronTunnelSecretName  = "tigera-management-cluster-connection"
 	defaultVoltronPort       = "9443"
 	defaultTunnelVoltronPort = "9449"
+	DashboardAPIPort         = "8444"
+	DashboardAPIHealthPort   = "8090"
+	DashboardAPIName         = "tigera-dashboard-api"
 )
 
 // Manager returns a component for rendering namespaced manager resources.
@@ -300,7 +305,7 @@ func (c *managerComponent) managerDeployment() *appsv1.Deployment {
 
 	managerPodContainers := []corev1.Container{c.managerUIAPIsContainer(), c.voltronContainer()}
 	if c.cfg.Tenant == nil {
-		managerPodContainers = append(managerPodContainers, c.managerContainer())
+		managerPodContainers = append(managerPodContainers, c.dashboardContainer(), c.managerContainer())
 	}
 	annotations := c.tlsAnnotations
 	if c.cfg.VoltronRouteConfig != nil {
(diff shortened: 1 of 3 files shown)


pkg/render/manager.go
```diff
@@ -434,6 +439,7 @@ func (c *managerComponent) managerEnvVars() []corev1.EnvVar {
 		{Name: "CNX_PROMETHEUS_API_URL", Value: fmt.Sprintf("/api/v1/namespaces/%s/services/calico-node-prometheus:9090/proxy/api/v1", common.TigeraPrometheusNamespace)},
 		{Name: "CNX_COMPLIANCE_REPORTS_API_URL", Value: "/compliance/reports"},
 		{Name: "CNX_QUERY_API_URL", Value: "/api/v1/namespaces/calico-system/services/https:calico-api:8080/proxy"},
+		{Name: "CNX_DASHBOARD_API_URL", Value: "/dashboard"},
```
rene-dekker: The best practice for RESTful APIs is to use plural nouns for resource names in the URL path, so I'd propose `/dashboards`.

pkg/render/manager.go
```diff
@@ -600,6 +606,63 @@ func (c *managerComponent) voltronContainer() corev1.Container {
...
+		Env:             env,
+		VolumeMounts:    mounts,
+		SecurityContext: securitycontext.NewNonRootContext(),
+		ReadinessProbe: &corev1.Probe{
+			ProbeHandler: corev1.ProbeHandler{
+				Exec: &corev1.ExecAction{
+					Command: []string{"/usr/bin/dashboard-api", "-ready"},
+				},
+			},
+			FailureThreshold:    3,
```
rene-dekker, line 646: We have code in components.go that sents defaults for our probes `setProbeTimeouts`. I was just wondering if these settings you hard-coded are copied over from kubernetes defaults or if they need to be set specifically, like they have here.

pkg/render/manager.go
```diff
@@ -850,6 +913,11 @@ func managerClusterRole(managedCluster bool, kubernetesProvider operatorv1.Provi
 				},
 				Verbs: []string{"list"},
 			},
+			{
```
rene-dekker, line 917: Could you add a comment in which contexts these permissions are used? Just so we keep better track as to why permissions are introduced.

pkg/render/manager.go
xiumozhan: They are copied over from `cc-mgmt-core` that installs custom dashboards in CC.

PR #4104, opened 2025-08-21
use a host alias for calico/node to access Goldmane

caseydavenport:
## Description

Use a host alias for Felix to talk to Goldmane, avoiding the need for Felix to use in-cluster DNS for any purpose.

## Release Note

```release-note
TBD
```

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/installation/core_controller.go
+++ b/pkg/controller/installation/core_controller.go
@@ -80,6 +80,7 @@ import (
 	relasticsearch "github.com/tigera/operator/pkg/render/common/elasticsearch"
 	"github.com/tigera/operator/pkg/render/common/networkpolicy"
 	"github.com/tigera/operator/pkg/render/common/resourcequota"
+	"github.com/tigera/operator/pkg/render/goldmane"
 	"github.com/tigera/operator/pkg/render/kubecontrollers"
 	"github.com/tigera/operator/pkg/render/monitor"
 	"github.com/tigera/operator/pkg/tls/certificatemanagement"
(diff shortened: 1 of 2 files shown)


matthewdupre (approved):

pkg/render/node.go
```diff
@@ -1431,8 +1445,9 @@ func (c *nodeComponent) nodeEnvVars() []corev1.EnvVar {
 		{Name: "NO_DEFAULT_POOLS", Value: "true"},
 	}
 
-	// Only append goldmane variables if goldmane is running.
-	if c.cfg.GoldmaneRunning {
+	// Only append goldmane variables if goldmane is running and we have a valid IP address,
+	// as we rely on an explicitly configured host alias to resolve the goldmane service.
+	if c.cfg.GoldmaneRunning && c.cfg.GoldmaneIP != "" {
```
matthewdupre, line 1452: I don't think this is a change so not a comment, but it would be preferable if we could somehow not restart Felix when Goldmane starts.  It's early in the day (so to speak), so not a huge deal, but ideally we would restart calico-node as little as possible.  It's not completely graceful (some flow logging data is lost).

pkg/controller/installation/core_controller.go
```diff
@@ -1386,6 +1383,19 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 		}
 	}
 
+	// Create a component handler to create or update the rendered components.
```
caseydavenport: I moved namespace creation up here in order to unblock the Goldmane controller from deploying the Service.

This allows us to - on initial install - wait to deploy calico/node until Kubernetes has allocated a ClusterIP for goldmane, thus removing the need to perform a second rollout of the daemonset. 

In the case Goldmane is NOT enabled, we simply carry on as normal.

PR #4115, opened 2025-08-26
[CORE-11751] Enable BPF dataplane auto-install by default for kube-proxy clusters

lucastigera:
## Description

This PR enables BPF dataplane auto-install by default for clusters using kube-proxy, advancing Calico's transition toward eBPF as the default dataplane. It follows the initial implementation of BPF auto-install and focuses on providing a seamless, low-friction experience for users deploying Calico in BPF mode.
- **Kind clusters as auto-detected providers:** Implemented a new logic to detect KinD clusters and changed the FVs to consider this new provider, as we use KinD to run them.
- **Auto-detection:** New logic to detect clusters where BPF dataplane can be enabled by default.
- **Default configuration:** For targeted clusters, install Calico with the eBPF dataplane, without requiring any user configuration or manual intervention.
- **Refactors:**
  - Wrapped the logic for detecting whether BPF networking can be bootstrapped, so it can be reused in the new auto-detection flow.
  - Wrapped the logic for determining whether the Operator can manage kube-proxy, also for reuse in the new auto-detection flow.
  - Removed the MergeAndFillDefaults function and incorporated its logic directly into updateInstallationWithDefaults.

## Release Note

```release-note
- eBPF dataplane is now enabled by default for clusters using kube-proxy. Calico auto-detects eligible clusters and installs with eBPF mode without user configuration.
- Kind was added as a new option for `KubernetesProvider` in Installation CRD.
```

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/installation/core_controller.go
+++ b/pkg/controller/installation/core_controller.go
@@ -419,11 +419,6 @@ func autoDetectDefaultBPF(ctx context.Context, client client.Client, instance *o
 		return false
 	}
 
-	// At the time of this implementation, we don't support BPF on KinD clusters.
-	if isKind, err := utils.IsKindCluster(ctx, client); err != nil || isKind {
-		return false
-	}
-
 	// Check if the cluster has the required resources.
 	if _, err := utils.CheckBPFClusterResourcesReq(ctx, client); err != nil {
 		return false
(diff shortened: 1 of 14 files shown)


pkg/controller/installation/core_controller.go
```diff
@@ -396,47 +396,102 @@ func getActivePools(ctx context.Context, client client.Client) (*crdv1.IPPoolLis
...
-		return err
+// autoDetectDefaultBPF checks whether the Operator can default to the BPF dataplane and install it automatically without user intervention.
+func autoDetectDefaultBPF(ctx context.Context, client client.Client, instance *operator.Installation) bool {
+	// Check if it's a new installation. `instance.Status.Computed` is set at the end of a successful reconcile. Thus, if it's
+	// nil, it's likely a new installation.
+	if instance == nil || instance.Status.Computed != nil {
+		return false
+	}
+
+	// linuxDataplane, bpfNetworkingBootstrap, and kubeProxyManagement shouldn't be set, or should be set to their default values.
```
tmjd: > their default values

Please update like you did in the doc to something like `the desired values`

pkg/controller/installation/core_controller_test.go
```diff
@@ -572,7 +572,7 @@ var _ = Describe("Testing core-controller installation", func() {
 					VXLANMode:    crdv1.VXLANModeAlways,
 				},
 			})
-			Expect(MergeAndFillDefaults(installation, nil, &currentPools)).To(BeNil())
+			Expect(fillDefaults(installation, &currentPools)).To(BeNil())
```
tmjd, line 575: I think updateInstallationWithDefaults replaced MergeAndFillDefaults so I think it is more appropriate to use the new function here rather than fillDefaults.

pkg/controller/installation/core_controller_test.go
```diff
@@ -585,7 +585,7 @@ var _ = Describe("Testing core-controller installation", func() {
 					KubernetesProvider: provider,
 				},
 			}
-			Expect(MergeAndFillDefaults(installation, nil, nil)).To(BeNil())
+			Expect(fillDefaults(installation, nil)).To(BeNil())
```
tmjd, line 588: ditto

pkg/controller/utils/kubeproxy.go
```diff
@@ -0,0 +1,65 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var (
+	KubeProxyInstanceKey = client.ObjectKey{Name: "kube-proxy", Namespace: "kube-system"}
+)
+
+// ValidateDaemonSetManaged checks whether the DaemonSet is managed by an external tool,
+// based on common labels or annotations typically added by automation.
+func ValidateDaemonSetManaged(ds *appsv1.DaemonSet) error {
```
tmjd: I think this name should be changed to `ValidateDaemonSetUnmanaged` because no error is the positive result for the function.

pkg/controller/installation/core_controller_test.go
lucastigera: That makes sense when you consider the intended scope of both `updateInstallationWithDefaults` and `MergeAndFillDefaults`.

But looking at the unit tests, both are passing `nil` for the `awsNode` (second) parameter. That effectively reduces the logic in the original `MergeAndFillDefaults` to just a call to `fillDefaults`:
```
// MergeAndFillDefaults merges in configuration from the Kubernetes provider, if applicable, and then
// populates defaults in the Installation instance.
func MergeAndFillDefaults(i *operator.Installation, awsNode *appsv1.DaemonSet, currentPools *crdv1.IPPoolList) error {
	if awsNode != nil {
		if err := updateInstallationForAWSNode(i, awsNode); err != nil {
			return fmt.Errorf("could not resolve AWS node configuration: %s", err.Error())
		}
	}

	return fillDefaults(i, currentPools)
}
```

So if the tests are always passing `nil` here, they're not actually verifying the merge logic - just the defaulting behavior.

Given that, I think it would make more sense to just test `fillDefaults` directly.

tmjd: Ok that is a good point. Are there any UTs that test the AWSNode functionality? If not could you please add some to make sure that hasn't been broken.

lucastigera: There are few testes in the `defaults_test.go` file to test:
- [updateInstallationForAWSNode](https://github.com/tigera/operator/blob/master/pkg/controller/installation/defaults_test.go#L544); and
- [fillDefaults with AmazonVPC CNI](https://github.com/tigera/operator/blob/master/pkg/controller/installation/defaults_test.go#L557)

tmjd (approved):

PR #4032, opened 2025-07-09
[RS-2546] Update operator to deploy waf-http-filter in Enterprise

LorcanMcVeigh:
## Description

This PR updates the operator to deploy waf-http-filter in Enterprise when enabling the Gateway API.

This will setup the required waf-http-filter sidecar container in the envoy-proxy deployment, but envoy won't be configured to use it by default. To do so, the user will need to create a separate EnvoyExtensionPolicy to enable the filter and protect the traffic with WAF.

## Release Note

```release-note
Deploy the WAF HTTP Filter alongside the Envoy Proxy when running in Calico Enterprise.
```

## For PR author

- [ ] Tests for change.
- [ ] If changing pkg/apis/, run `make gen-files`
- [ ] If changing versions, run `make gen-versions`

## For PR reviewers

A note for code reviewers - all pull requests must have the following:

- [ ] Milestone set according to targeted release.
- [ ] Appropriate labels:
  - `kind/bug` if this is a bugfix.
  - `kind/enhancement` if this is a a new feature.
  - `enterprise` if this PR applies to Calico Enterprise only.

PR #4119, opened 2025-08-28
[RS-2546] Update operator to deploy waf-http-filter in Enterprise (#4032)

LorcanMcVeigh:
* Start a test for new behaviour

* Use privileged PSS for tigera-gateway ns

* Enable EnvoyPatchPolicy

* Deploy and configure waf-http-filter

* Fix codegen

* Fix capitalisation

* Support custom envoy proxy

* Add comment for PSSPrivileged

* Check volumes are set as expected

---------

## Release Note

```release-note
Deploy the WAF HTTP Filter alongside the Envoy Proxy when running in Calico Enterprise.
```

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/gateway_api.go
+++ b/pkg/render/gateway_api.go
@@ -25,9 +25,11 @@ import (
 	operatorv1 "github.com/tigera/operator/api/v1"
 	"github.com/tigera/operator/pkg/common"
 	"github.com/tigera/operator/pkg/components"
+	"github.com/tigera/operator/pkg/ptr"
 	rcomp "github.com/tigera/operator/pkg/render/common/components"
 	rmeta "github.com/tigera/operator/pkg/render/common/meta"
 	"github.com/tigera/operator/pkg/render/common/secret"
+	"github.com/tigera/operator/pkg/render/common/securitycontext"
 	appsv1 "k8s.io/api/apps/v1"
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
@@ -341,6 +343,7 @@ type gatewayAPIImplementationComponent struct {
 	envoyGatewayImage   string
 	envoyProxyImage     string
 	envoyRatelimitImage string
+	wafHTTPFilterImage  string
 }
 
 func GatewayAPIImplementationComponent(cfg *GatewayAPIImplementationConfig) Component {
@@ -365,6 +368,10 @@ func (pr *gatewayAPIImplementationComponent) ResolveImages(is *operatorv1.ImageS
 			return err
 		}
 		pr.envoyRatelimitImage, err = components.GetReference(components.ComponentGatewayAPIEnvoyRatelimit, reg, path, prefix, is)
+		if err != nil {
+			return err
+		}
+		pr.wafHTTPFilterImage, err = components.GetReference(components.ComponentWAFHTTPFilter, reg, path, prefix, is)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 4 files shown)


pkg/render/gateway_api.go
```diff
@@ -654,6 +662,110 @@ func (pr *gatewayAPIImplementationComponent) envoyProxyConfig(className string,
 	}
 	applyEnvoyProxyServiceOverrides(envoyProxy, classSpec.GatewayService)
 
+	// Setup WAF HTTP Filter on Enterprise.
+	if pr.cfg.Installation.Variant == operatorv1.TigeraSecureEnterprise {
+		// The WAF HTTP filter is not supported when the envoy proxy is deployed as a DaemonSet
```
nelljerram, line 667: This restriction needs to be documented.
- One place, I think, would be under `GatewayKind` here: https://docs.tigera.io/calico-enterprise/3.22/reference/installation/api#gatewayclassspec
- Will there also be a dedicated doc page for GatewayWAF?  If so, I guess it should be mentioned there too.

@LorcanMcVeigh Do you have tickets tracking this?

pkg/render/gateway_api.go
```diff
@@ -654,6 +662,110 @@ func (pr *gatewayAPIImplementationComponent) envoyProxyConfig(className string,
...
+					},
+				},
+				SecurityContext: securitycontext.NewRootContext(true),
+			}
+			hasWAFHTTPFilter := false
+			for i, initContainer := range envoyProxy.Spec.Provider.Kubernetes.EnvoyDeployment.InitContainers {
+				if initContainer.Name == wafHTTPFilter.Name {
+					hasWAFHTTPFilter = true
+					// Handle update
+					if initContainer.Image != wafHTTPFilter.Image {
```
nelljerram, line 700: Why only check the image?  We may in future want to update the WAF sidecar in other ways, so it might be better for this code to support that.

pkg/render/gateway_api.go
```diff
@@ -654,6 +662,110 @@ func (pr *gatewayAPIImplementationComponent) envoyProxyConfig(className string,
...
+				},
+				Name: "waf-http-filter",
+			}
+			hasLogsVolume := false
+			hasSocketVolume := false
+			for i, volume := range envoyProxy.Spec.Provider.Kubernetes.EnvoyDeployment.Pod.Volumes {
+				if volume.Name == logsVolume.Name {
+					hasLogsVolume = true
+					// Handle update
+					if volume.VolumeSource.HostPath.Path != logsVolume.VolumeSource.HostPath.Path || volume.VolumeSource.HostPath.Type != logsVolume.VolumeSource.HostPath.Type {
```
nelljerram, line 749: As above, I don't think we need this check.  Usual operator behaviour is just to blat the complete thing that it wants.

pkg/render/gateway_api.go
nelljerram: I think the operator should just update the whole container spec that it wants, without checking for image change.

nelljerram (changes requested):
I've commented a few things.  If you agree, they'll need to be made in master first and then those changes added to this PR.

LorcanMcVeigh:
@nelljerram At this point, this PR is simply to get the feature into the release it was meant to be apart of. These changes don't need to be in this release. We can make the changes to master and they can be picked up in the next release but I don't see the need for them to be in this cherry-pick or any follow-up cherry-pick.

pkg/render/gateway_api.go
gantony: Gateway WAF is currently documented in https://docs.tigera.io/calico-enterprise/3.22/threat/deploying-waf-ingress-gateway. Agree we need to add a note under `GatewayKind`.

pkg/render/gateway_api.go
gantony: Let's address this in a follow up PR to keep changes minimal for the cherry-pick.

pkg/render/gateway_api.go
gantony: Let's address this in a follow up PR to keep changes minimal for the cherry-pick.

PR #4130, opened 2025-09-06
chore: update Envoy Gateway to v1.5.0

electricjesus:
Update Envoy Gateway from the previous version to v1.5.0 following the [provided upgrade instructions](docs/common_tasks.md).

- **Dependencies**: Updated `go.mod` to use Envoy Gateway v1.5.0
- **Configuration**: Updated `Makefile` to use `envoy-gateway/gateway-helm` chart reference
- **Resources**: Regenerated `pkg/render/gateway_api_resources.yaml` with updated resources (40,569 lines)
- **Tests**: Fixed CRD count expectation in `pkg/render/gateway_api.go` from 11 to 10 k8s CRDs (SecurityPolicy is now an envoy CRD)

- ✅ All Gateway tests passing (11/11)
- ✅ All unit tests passing (66 suites)
- ✅ No regressions introduced

This update includes new Envoy Gateway v1.5.0 features such as the SecurityPolicy CRD and other enhancements.

- Ran focused Gateway API tests with `go test ./pkg/render -v -ginkgo.focus ".*Gateway.*"`
- Executed full unit test suite with `make ut`
- All tests pass successfully

## Release Note

```release-note
TBD
```

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


pkg/controller/gatewayapi/gatewayapi_controller.go
```diff
@@ -44,6 +44,7 @@ import (
 	"github.com/tigera/operator/pkg/controller/utils/imageset"
 	"github.com/tigera/operator/pkg/ctrlruntime"
 	"github.com/tigera/operator/pkg/render"
+	"github.com/tigera/operator/pkg/render/gatewayapi"
```
nelljerram, line 48: Moving the gateway API rendering into its own package seems like a nice move to me, but I'd appreciate an operator-dev OK'ing this as well, in case there's some gotcha with not having all of our rendering in a single package.

pkg/controller/compliance/compliance_controller_test.go
```diff
@@ -385,7 +385,7 @@ var _ = Describe("Compliance controller tests", func() {
 	It("create namespace, operator secrets role and pull secrets", func() {
 		result, err := r.Reconcile(ctx, reconcile.Request{})
 		Expect(err).NotTo(HaveOccurred())
-		Expect(result.RequeueAfter).To(Equal(0 * time.Second))
+		Expect(result.RequeueAfterAfter).To(Equal(0 * time.Second))
```
nelljerram: Really?  That's such an obviously terrible name.  But I presume this isn't your choice at all.  Out of interest do you have a pointer to the deprecation notice?

pkg/render/gatewayapi/gateway_api.go
```diff
@@ -447,25 +455,38 @@ func (pr *gatewayAPIImplementationComponent) Objects() ([]client.Object, []clien
...
+	for _, cr := range resources.clusterRoles {
+		objs = append(objs, cr.DeepCopyObject().(client.Object))
+	}
+	for _, crb := range resources.clusterRoleBindings {
+		objs = append(objs, crb.DeepCopyObject().(client.Object))
+	}
+	for _, mwc := range resources.mutatingWebhookConfigurations {
+		objs = append(objs, mwc.DeepCopyObject().(client.Object))
+	}
+	for _, resource := range []client.Object{
```
nelljerram, line 490: Please note that this function is quite careful about the order of objsToCreate - so that when object A depends on other objects {B}, {B} are emitted first, then A.  Considering the insertions here:
- I think the extra cluster role (binding) for certgen job is fine, because the job itself is created later.
- Not sure about the mutating web hook.  Does this depend on the envoy-gateway controller already running, since that what it points to?  Perhaps it's fine because of its policy to ignore failures.  I'm curious what happens later when the envoy-gateway controller is actually deployed... then the webhook fires and tries to call the controller while it's still being created?

pkg/controller/compliance/compliance_controller_test.go
nelljerram: Oh I see, it was a typo.

electricjesus: sorry that was a search-and-replace fail! 

here's the specific release we've upgrade to, take a look at Breaking Changes https://github.com/kubernetes-sigs/controller-runtime/releases/tag/v0.21.0

pkg/render/gatewayapi/gateway_api.go
electricjesus: @nelljerram apparently the webhook's only job is to label gateway proxy pods with node topology labels. these seem to be mostly informational doesn't seem to affect operation. perhaps it's used for something else? taxonomy purposes?

pkg/controller/gatewayapi/gatewayapi_controller.go
electricjesus: we do already have existing component-specific subdirectories under pkg/render.. so I think that's okay? (cc @tigera/tigera-operator-maintainers maybe i'm not seeing something here)

rene-dekker (approved):

PR #4137, opened 2025-09-09
feat(dex): Enable PKCE and silent token renewal for SPA

rene-dekker:
### feat(dex): Enable PKCE and silent token renewal for SPA

This change updates the Dex client configuration to support Single Page Applications (SPAs).

The Dex client is configured as public (`public: true`), which disables the use of a client secret and enables the more secure Proof Key for Code Exchange (PKCE) flow. This is the recommended standard for public
clients like browser-based applications.

Additionally, the 'X-Frame-Options' header is changed from 'DENY' to 'SAMEORIGIN'. This allows the SPA to use a hidden iframe to perform silent OIDC callbacks, enabling seamless token renewal without requiring a full-
page redirect.

Together, these changes improve security and user experience for the Calico Enterprise Manager when used as an SPA.

```release-note
This change updates the Dex client configuration to support Single Page Applications (SPAs) code flow with PKCE.
The 'X-Frame-Options' header was changed from 'DENY' to 'SAMEORIGIN'.
```

### DIFF
--- a/pkg/render/dex.go
+++ b/pkg/render/dex.go
@@ -132,6 +132,11 @@ func (c *dexComponent) Objects() ([]client.Object, []client.Object) {
 		c.clusterRoleBinding(),
 		c.configMap(),
 	}
+	objectsToDelete := []client.Object{
+		// Delete the secret called tigera-dex which in the past was used to store a client secret.
+		&corev1.Secret{ObjectMeta: metav1.ObjectMeta{Namespace: DexObjectName, Name: DexObjectName}},
+		&corev1.Secret{ObjectMeta: metav1.ObjectMeta{Namespace: common.OperatorNamespace(), Name: DexObjectName}},
+	}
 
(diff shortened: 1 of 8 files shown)


anthonytwh (approved):

pkg/render/dex.go
```diff
@@ -379,6 +381,9 @@ func (c *dexComponent) configMap() *corev1.ConfigMap {
 				"redirectURIs": c.cfg.DexConfig.RedirectURIs(),
 				"name":         "Calico Enterprise Manager",
 				"secretEnv":    dexSecretEnv,
```
anthonytwh: Is `secretEnv` still needed when `public: true`?

rene-dekker: I pushed a commit to delete that secret as well, great catch!

anthonytwh (approved):

PR #4144, opened 2025-09-11
Set vxlanPort to 8472 and turn off nftables for MKE in ebpf mode.

sridhartigera:
## Description

This PR has the below changes

1. Turn off NFTables When installing calico in eBPF mode - MKE installs some iptable rules as part of docker swarm which don't interop with the nft rules when in BPF mode. Hence it is better to turn off nft and use iptables for the few rules that calico installs for eBPF dp.
2. Change the VXLAN port for ebpf installs in MKE - MKE already has a vxlan device at port `4789` and vni `4096`. Until v3.30, we have been using FDB based vxlan in calico across all dataplanes. Hence we just moved our `vxlanVNI` to `10000`. Now with 3.31, Calico creates Flow based vxlan for ebpf dataplane. When creating a flow based vxlan, we don't specify the VNI. This results in 2 devices at the same vxlan port causing conflicts. Hence when installing Calico in eBPF dataplane, the default vxlanPort is `8472`. 

## Release Note

```release-note
Changed default vxlanPort to 8472 for MKE when installing in ebpf mode. MKE already has a vxlan device at port 4789 and vni 4096. Until v3.30, we have been using FDB based vxlan in calico across all dataplanes. Now with 3.31, Calico creates Flow based vxlan for ebpf dataplane. When creating a flow based vxlan, we don't specify the VNI. This results in 2 devices at the same vxlan port causing conflicts. 8472 is/was the port allocated to vxlan before IANA standardized on 4789.
```

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/api/v1/installation_types.go
+++ b/api/v1/installation_types.go
@@ -998,7 +998,9 @@ func (s *InstallationSpec) BPFEnabled() bool {
 // IsNftables is an extension method that returns true if the Installation resource
 // has Calico Network Linux Dataplane set and equal to value "Nftables" or "BPF", otherwise false.
 //
-// BPF is included here as it uses nftables to program some rules, except when on DockerEE.
+// BPF is included here as it uses nftables to program some rules, except when on DockerEE,
+// since docker-ee programs some rules in iptables. These rules does not interact well with the
+// nftable rules that calico programs, so we exclude BPF when on DockerEE.
 func (s *InstallationSpec) IsNftables() bool {
 	if s.CalicoNetwork != nil && s.CalicoNetwork.LinuxDataplane != nil {
 		if s.KubernetesProvider.IsDockerEE() &&


api/v1/installation_types.go
```diff
@@ -998,12 +998,17 @@ func (s *InstallationSpec) BPFEnabled() bool {
 // IsNftables is an extension method that returns true if the Installation resource
 // has Calico Network Linux Dataplane set and equal to value "Nftables" or "BPF", otherwise false.
 //
-// BPF is included here as it uses nftables to program some rules.
+// BPF is included here as it uses nftables to program some rules, except when on DockerEE.
```
caseydavenport: I think we should add more context here explaining why we exclude "docker EE" (we also really need to update this field to be MKE)

pkg/controller/installation/core_controller.go
```diff
@@ -1914,19 +1914,31 @@ func (r *ReconcileInstallation) setDefaultsOnFelixConfiguration(ctx context.Cont
...
 	}
 	vxlanVNI := 4096
+	vxlanPort := 4789
 	// MKE uses a vxlanVNI:4096 and vxlanPort:4789 for its docker swarm vxlan.
 	// This results in a conflict with calico's VXLAN and the vxlan.calico interface
 	// gets deleted. To fix this we change the vxlanVNI to 10000 as recommended by
 	// MKE docs (https://docs.mirantis.com/mke/3.7/cli-ref/mke-cli-install.html).
 	if install.Spec.KubernetesProvider == operator.ProviderDockerEE {
 		vxlanVNI = 10000
+		// Starting Calico v3.31, we are using a flow based VXLAN device for
```
caseydavenport:
```suggestion
		// We are using a flow based VXLAN device for
```

a bit of a nit, but no need to include version context in the comments - we have git history to tell us when we made this change.

PR #4146, opened 2025-09-12
[CORE-11824] BPF auto-config - Warning message for auto-detected providers

lucastigera:
## Description

- Log a warning message if we're auto-configuring BPF (bootstrapping network or managing `kube-proxy`) on a cluster running in a provider that may have unknown incompatibilities.
- Add KinD as a auto-detected provider.
- Included a few renamings to improve readability.

## Release Note

```release-note
TBD
```

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


pkg/controller/installation/core_controller.go
```diff
@@ -1493,6 +1493,10 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 		// Extract k8s service and endpoints to push them to ebpf-bootstrap init container.
 		nodeCfg.K8sServiceAddrs = serviceIPsAndPorts(bpfBootstrapReq.K8sService)
 		nodeCfg.K8sEndpointSlice = serviceEndpointSlice(bpfBootstrapReq.K8sServiceEndpoints)
+
+		if !instance.Spec.KubernetesProvider.IsNone() {
+			reqLogger.Info(fmt.Sprintf("[WARNING] Auto bootstrapping BPF network may result in unexpected behavior in %s.", instance.Spec.KubernetesProvider))
```
caseydavenport: I think we'll want to use a rate limited logger of some sort of this message - otherwise this could get logged out repeatedly and get pretty spammy?

caseydavenport: I think you can use `Infof` to avoid calling `fmt.Sprintf` also

caseydavenport: I also feel like we should be directing the user in some way - like, this message is a good warning, but it doesn't say much about what the next action should be. Is there any instruction we could provide to help a user determine if they need to do anything?

lucastigera: `logr.Logger` doesn't support `Infof`.
Would it make sense to switch to `logrus.Warnf` here, or should we keep it and just refactor the message to use structured logging instead?

**1. logr.Logger**

```go
reqLogger.Info("[WARNING] Auto bootstrapping BPF network may result in unexpected behavior with the current provider.", "provider", instance.Spec.KubernetesProvider)
```
Output:
```
{"level":"info","ts":"2025-09-19T00:25:42Z","logger":"controller_installation","msg":"[WARNING] Auto bootstrapping BPF network may result in unexpected behavior with the current provider.","Request.Namespace":"tigera-operator","Request.Name":"tigera-ca-private","provider":"Kind"}
```

**2. logrus.Infof**

```go
logrus.Warnf("Auto bootstrapping BPF network may result in unexpected behavior in %s.", instance.Spec.KubernetesProvider)
```
Output:
```
time="2025-09-19T00:25:42Z" level=warning msg="Auto bootstrapping BPF network may result in unexpected behavior in Kind."  
```

lucastigera: > Is there any instruction we could provide to help a user determine if they need to do anything?

Could we guide the user to disable it and install eBPF with the manual configuration?

_[WARNING] Auto bootstrapping BPF network may result in unexpected behavior in Kind. If you experience API server communication issues, disable "bpfBootstrapNetworking" in the Installation CR and follow the manual eBPF installation guide at docs.tigera.io._

lucastigera: > I think we'll want to use a rate limited logger of some sort of this message

What do you think about comparing `instance.*Spec*.CalicoNetwork.bpfNetworkBootstrap` with `instance.*Status.Computed*.CalicoNetwork.bpfNetworkBootstrap`, and only logging the message if they differ **and** the spec value is enabled?

Would that be a safe way to ensure the warning is only logged when necessary?

caseydavenport: @lucastigera yeah that probably works!

> logr.Logger doesn't support Infof.

Bizarre, I thought we used that elsewhere. We shouldn't introduce another logging library here - probably either leave it as is, or use `keysAndValues` for context logging.

pkg/controller/installation/core_controller_test.go
```diff
@@ -1002,7 +1002,7 @@ var _ = Describe("Testing core-controller installation", func() {
 			})
 		})
 
-		Context("with LinuxDataplane=BPF and BPFNetworkBootstrap=Enabled", func() {
+		Context("streamline BPF installation", func() {
```
caseydavenport: I actually think the previous description of this context was more clear / accurate!

caseydavenport (approved):

PR #4150, opened 2025-09-16
Fix finalizer removal to account for lingering Pods in teardown

electricjesus:
## Description

This is an attempt at fixing a flake.

**Problem**

FV tests are flaking with "Calico namespace still exists" timeouts during teardown. The root cause is that the [MaintainInstallationFinalizer](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) logic only checks for the existence of secondary resources (e.g., Deployments) before removing the Installation finalizer. However, Pods from these resources tend to linger after the Deployment/ReplicaSet is deleted (due to Kubernetes cleanup delays or other issues), preventing namespace deletion. Once the finalizer is removed prematurely, CNI permissions are revoked, leaving lingering Pods unable to terminate and blocking the calico-system namespace from being cleaned up.

**Solution**

Enhanced the MaintainInstallationFinalizer to also check for lingering Pods before removing the finalizer. After confirming a secondary resource no longer exists, the code now lists Pods in the same namespace with the label `k8s-app=<secondaryResource.GetName()>`. If any Pods exist, finalizer removal is delayed. Only when both the resource and its associated Pods are gone does the finalizer get removed.

## Release Note

```release-note
None
```

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


pkg/controller/gatewayapi/gatewayapi_controller.go
```diff
@@ -474,3 +489,24 @@ func (r *ReconcileGatewayAPI) getPolicySyncPathPrefix(fcSpec *crdv1.FelixConfigu
 
 	return DefaultPolicySyncPrefix
 }
+
+func (r *ReconcileGatewayAPI) handleDeletion(ctx context.Context, gatewayAPI *operatorv1.GatewayAPI, reqLogger logr.Logger) (reconcile.Result, error) {
```
caseydavenport: I don't think we should need this logic explicitly? The namespace should be automatically removed due to owner references should it not?

pkg/controller/gatewayapi/gatewayapi_controller.go
```diff
@@ -474,3 +489,24 @@ func (r *ReconcileGatewayAPI) getPolicySyncPathPrefix(fcSpec *crdv1.FelixConfigu
...
+	ns := &corev1.Namespace{
+		ObjectMeta: metav1.ObjectMeta{Name: "tigera-gateway"},
+	}
+	if err := r.client.Delete(ctx, ns); err != nil && !errors.IsNotFound(err) {
+		r.status.SetDegraded(operatorv1.ResourceUpdateError, "Failed to delete tigera-gateway namespace", err, reqLogger)
+		return reconcile.Result{}, err
+	}
+
+	// Remove the finalizer.
+	controllerutil.RemoveFinalizer(gatewayAPI, render.GatewayAPIFinalizer)
```
caseydavenport: Do we really need this finalizer? The namespace is owned by the GatewayAPI isn't it?

pkg/controller/utils/utils.go
```diff
@@ -1082,19 +1082,39 @@ func MaintainInstallationFinalizer(
...
+			// Only applicable for workload types that create Pods.
+			switch secondaryResource.(type) {
+			case *appsv1.Deployment, *appsv1.DaemonSet, *appsv1.StatefulSet:
+				// We key off the standard k8s-app label which we set to the resource name for our Deployments/DaemonSets.
+				podList := &corev1.PodList{}
+				if listErr := c.List(
+					ctx,
+					podList,
+					client.InNamespace(secondaryResource.GetNamespace()),
+					client.MatchingLabels{"k8s-app": secondaryResource.GetName()},
```
caseydavenport: We should be able to get the `Selector` from the owning object rather than assuming it is using this pattern (can't be sure that all future objects might use this selector)

pkg/controller/utils/utils.go
```diff
@@ -1082,19 +1082,39 @@ func MaintainInstallationFinalizer(
...
 				return nil
 			}
+
+			// If the secondary resource itself is gone, ensure there are no Pods left over that would still rely on CNI.
+			// Only applicable for workload types that create Pods.
+			switch secondaryResource.(type) {
+			case *appsv1.Deployment, *appsv1.DaemonSet, *appsv1.StatefulSet:
+				// We key off the standard k8s-app label which we set to the resource name for our Deployments/DaemonSets.
+				podList := &corev1.PodList{}
+				if listErr := c.List(
```
caseydavenport: No need to define a new err name? I think we can reuse `err` here.

pkg/controller/utils/utils.go
```diff
@@ -1082,19 +1082,39 @@ func MaintainInstallationFinalizer(
...
 		// Check if the namespaced secondaryResources are still present.
 		// Keep track of all the secondary resources that the main resource creates.
-		// Only delete the finalizer if all of the secondary resources are deleted.
+		// Only delete the finalizer if all of the secondary resources are deleted and there are no lingering Pods.
 		for _, secondaryResource := range secondaryResources {
 			err := c.Get(ctx, types.NamespacedName{Namespace: secondaryResource.GetNamespace(), Name: secondaryResource.GetName()}, secondaryResource)
 			if err != nil && !errors.IsNotFound(err) {
 				return err
-			} else if errors.IsNotFound(err) {
-				log.Info("Object no longer exists.", "object", secondaryResource)
```
caseydavenport, line 1074: I think having this log was useful?

pkg/controller/gatewayapi/gatewayapi_controller.go
electricjesus: actually let me remove the gateway api finalizer stuff as i'm not super sure about it for now

pkg/controller/gatewayapi/gatewayapi_controller.go
electricjesus: as per my other comment, i removed pkg/controller/gatewayapi related finalizer work so we can focus on the main one that's flaking

pkg/controller/utils/utils.go
electricjesus: agreed.. restored this one

pkg/controller/utils/utils.go
electricjesus: I moved it out to its own function

pkg/controller/utils/utils.go
electricjesus: agreed - i took a stab at it, lemme know if it's decent!

pkg/controller/utils/utils.go
```diff
@@ -1082,22 +1082,64 @@ func MaintainInstallationFinalizer(
...
+				} else if exists {
+					log.V(2).Info("Lingering Pods still exist for StatefulSet", "statefulset", obj.Name, "count", count)
+					return nil
+				}
+			default:
+				// Non-workload resource; no Pod listing needed.
+			}
 		}
-		log.Info("All objects no longer exist. Removing finalizer", "finalizer", finalizer)
+		log.V(2).Info("All objects and their Pods no longer exist. Removing finalizer", "finalizer", finalizer)
```
electricjesus, line 1096: ah maybe i shouldn't've .V(2)'d this one

caseydavenport: Yeah, I'd prefer we leave all of these logs at regular verbosity - they are rare and only occur during teardown, which is unfortunately an error prone process and having the logs at all verbosity levels has proven helpful!

pkg/controller/utils/utils.go
```diff
@@ -1065,22 +1065,83 @@ func MaintainInstallationFinalizer(
...
+				log.V(2).Info("Object is still present, waiting for termination", "object", secondaryResource)
 				return nil
 			}
+
+			// If the secondary resource itself is gone, ensure there are no Pods left over that would still rely on CNI.
+			// Only applicable for workload types that create Pods.
+			switch obj := secondaryResource.(type) {
+			case *appsv1.Deployment:
+				matchLabels := getWorkloadMatchLabels(obj)
+				return checkLingeringPodsForWorkload(ctx, c, obj, matchLabels, "Deployment", "deployment")
```
caseydavenport, line 1085: The last param here is always just a lowercase version of the previous one - can we omit it?

pkg/controller/utils/utils.go
```diff
@@ -1065,22 +1065,83 @@ func MaintainInstallationFinalizer(
...
+			case *appsv1.DaemonSet:
+				matchLabels := getWorkloadMatchLabels(obj)
+				return checkLingeringPodsForWorkload(ctx, c, obj, matchLabels, "DaemonSet", "daemonset")
+			case *appsv1.StatefulSet:
+				matchLabels := getWorkloadMatchLabels(obj)
+				return checkLingeringPodsForWorkload(ctx, c, obj, matchLabels, "StatefulSet", "statefulset")
```
caseydavenport, line 1091: If we always return here, we'll never fall through to the logic below to remove the installation finalizer.

pkg/controller/utils/utils.go
```diff
@@ -1065,22 +1065,83 @@ func MaintainInstallationFinalizer(
...
 		RemoveInstallationFinalizer(installation, finalizer)
 	}
 
 	// Update the installation with any finalizer changes.
 	return c.Patch(ctx, installation, patchFrom)
 }
+
+func hasLingeringPods(ctx context.Context, c client.Client, namespace string, matchLabels map[string]string) (bool, int, error) {
+	if matchLabels == nil {
+		return false, 0, nil
```
caseydavenport, line 1106: This feels like a coding bug if we ever fall into this branch, right? Probably worth a warning log if we end up here. 

```
log.Info("BUG: missing expected label selector")
```

 or something like this

pkg/controller/utils/utils.go
```diff
@@ -1065,22 +1065,83 @@ func MaintainInstallationFinalizer(
...
+	}
+	podList := &corev1.PodList{}
+	if err := c.List(
+		ctx,
+		podList,
+		client.InNamespace(namespace),
+		client.MatchingLabels(matchLabels),
```
caseydavenport, line 1113:
```suggestion
	if err := c.List(ctx, podList, client.InNamespace(namespace), client.MatchingLabels(matchLabels),
```

Just a nit, but IMO this line isn't long enough to warrant being split. Up to you though.

pkg/controller/utils/utils.go
```diff
@@ -1065,22 +1065,83 @@ func MaintainInstallationFinalizer(
...
+	); err != nil {
+		return false, 0, err
+	}
+	count := len(podList.Items)
+	return count > 0, count, nil
+}
+
+func getWorkloadMatchLabels(obj client.Object) map[string]string {
```
caseydavenport, line 1121: We could simplify the logic in `MaintainInstallationFinalizer` if we moved the select statement inside of this function. Then, MaintainInstallationFinalizer would look more like this:

```go
sel, ok := getWorkloadMatchLabels(secondaryResource)
if ok {
  checkLingeringPodsForWorkload(ctx, c, secondaryResource, sel)
}
```

Or even just 

```go
  checkLingeringPodsForWorkload(ctx, c, secondaryResource, sel)
```

And have `checkLingeringPodsForWorkload` responsible for returning `nil` if not an applicable type, and parsing the selector if it is.

caseydavenport:
Took over this one here: https://github.com/tigera/operator/pull/4179

PR #4153, opened 2025-09-17
Minimal footprint - consolidate tigera-manager namespace and contents into calico-system on standalone and management clusters

Josh-L:
## Release Note

```release-note
Contents of tigera-manager namespace have been moved to calico-system namespace on standalone and management clusters (managed clusters were moved in a previous release).
```

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


caseydavenport:
Not sure why Semaphore isn't recognizing you @Josh-L ...

Josh-L:
> Not sure why Semaphore isn't recognizing you @Josh-L ...

@caseydavenport now that I think about it I did change my github username and there was some warning about things that might break, but I don't remember group membership being one of them.

pkg/controller/installation/core_controller_test.go
```diff
@@ -1255,6 +1260,46 @@ var _ = Describe("Testing core-controller installation", func() {
 			Expect(*fc.Spec.BPFEnabled).To(BeFalse())
 		})
 
+		It("should reconcile namespace, role binding and pull secrts", func() {
```
rene-dekker:
```suggestion
		It("should reconcile namespace, role binding and pull secrets", func() {
```

pkg/render/manager.go
```diff
@@ -49,11 +49,11 @@ import (
 )
 
 const (
-	managerPort           = 9443
+	ManagerPort           = 9443
 	managerTargetPort     = 9443
 	ManagerServiceName    = "tigera-manager"
 	ManagerDeploymentName = "tigera-manager"
```
rene-dekker: I'm wondering if we want to change the prefix of all resources to "calico-". We've been doing that for other components and it would be nice over time to end up in a fully consistent state.

pkg/render/manager.go
```diff
@@ -231,57 +231,60 @@ func (c *managerComponent) SupportedOSType() rmeta.OSType {
...
 		// when there is no tenant.
-		objs = append(objs,
+		objsToCreate = append(objsToCreate,
 			managerClusterWideSettingsGroup(),
 			managerUserSpecificSettingsGroup(),
 			managerClusterWideTigeraLayer(),
 			managerClusterWideDefaultView(),
 		)
+
+		objsToDelete = c.deprecatedResources()
```
rene-dekker: Why do we only clean up resources if it is not a multi-tenant setup?

Josh-L: In a multi-tenant setup the `tigera-manager` namespace doesn't exist, all of the manager components are already in the tenant namespace like `cc-tenant-<tenant-ID>`

pkg/render/manager.go
Josh-L: That's a good point, I'll start doing that now.

pkg/render/manager.go
Josh-L: I suppose this logic will change now that we've decided to rename `tigera-*` resources to `calico-*` resources since I'll need to clean up all of the `tigera-*` resources even inside of the `cc-tenant-*` namespace and replace them with `calico-*` equivalents.

api/v1/managementcluster_types.go
```diff
@@ -46,11 +46,11 @@ type TLS struct {
...
 	// To reconnect existing managed clusters, change the tls.ca of the  managed clusters' ManagementClusterConnection resource.
 	//
-	// One of: tigera-management-cluster-connection, manager-tls
+	// One of: calico-management-cluster-connection, manager-tls
 	//
-	// Default: tigera-management-cluster-connection
+	// Default: calico-management-cluster-connection
 	//
-	// +kubebuilder:validation:Enum=tigera-management-cluster-connection;manager-tls
+	// +kubebuilder:validation:Enum=calico-management-cluster-connection;manager-tls
```
rene-dekker: For all of these kubebuilder validations I think we must support the union of the old + the new values. Otherwise previously valid and stored database objects would not be able to be persisted again.

I think it would be nice if we supported both the old and the new names for these, so user will not run into surprised during/after upgrades.

pkg/controller/logstorage/kubecontrollers/es_kube_controllers.go
```diff
@@ -330,12 +330,6 @@ func (r *ESKubeControllersController) Reconcile(ctx context.Context, request rec
 		return reconcile.Result{}, err
 	}
 
-	setup := render.NewSetup(&render.SetUpConfiguration{
```
rene-dekker, line 333: I do think these changes are probably correct, but I do wonder why they are in this PR. What prompted you to do this?

Josh-L: This was due to some thrashing of the tigerastatus I was seeing when multiple components were all trying to create and own the same resources. Basically when these things all existed in their own namespace it was correct for each component to own these. Once these components were consolidated into the same namespace having multiple entities try to create/own/update them was problematic. I think it was you I discussed this with about moving the ownership of these things to the core controller since it's the owner of the calico-system namespace

api/v1/managementcluster_types.go
Josh-L: That's a good point, I suppose we'll also need some code to intercept the old values and convert them to the new ones to fully cover the upgrade scenario.

api/v1/managementcluster_types.go
```diff
@@ -46,11 +46,11 @@ type TLS struct {
 	// If changed on a running cluster with connected managed clusters, all managed clusters will disconnect as they will no longer be able to verify Voltron's identity.
 	// To reconnect existing managed clusters, change the tls.ca of the  managed clusters' ManagementClusterConnection resource.
 	//
-	// One of: tigera-management-cluster-connection, manager-tls
+	// One of: calico-management-cluster-connection, manager-tls, tigera-management-cluster-connection (deprecated)
```
rene-dekker, line 49: @tmjd 
I am not so sure about this change, at least in its current form. It will break MCM setups on upgrade (including cloud) think we should either:
- Revert.
- Fetch both calico-management-cluster-connection and tigera-management-cluster-connection, so we can fall back on the old secret if the former does not exist.

api/v1/manager_types.go
```diff
@@ -67,8 +67,8 @@ type ManagerDeploymentPodSpec struct {
 // ManagerDeploymentContainer is a Manager Deployment container.
 type ManagerDeploymentContainer struct {
 	// Name is an enum which identifies the Manager Deployment container by name.
-	// Supported values are: tigera-voltron, tigera-manager, tigera-ui-apis, and tigera-es-proxy (deprecated).
-	// +kubebuilder:validation:Enum=tigera-voltron;tigera-manager;tigera-es-proxy;tigera-ui-apis
+	// Supported values are: calico-voltron, calico-manager, calico-ui-apis, tigera-voltron (deprecated), tigera-manager (deprecated), tigera-ui-apis (deprecated)
```
rene-dekker: We should keep tigera-es-proxy and tigera-voltron for backwards compatibility.

pkg/render/manager.go
```diff
@@ -352,6 +355,21 @@ func (c *managerComponent) managerDeployment() *appsv1.Deployment {
 
 	if c.cfg.Manager != nil {
 		if overrides := c.cfg.Manager.Spec.ManagerDeployment; overrides != nil {
+			// Due to the change of prefix for some components (eg. tigera-manager -> calico-manager) we need to support
+			// the upgrade scenario in a graceful way that doesn't force the user to update the new names. To accomplish
+			// this we will simply perform the substitution from old name to new name before applying the settings to
+			// the deployment
+
+			if overrides.Spec != nil && overrides.Spec.Template != nil && overrides.Spec.Template.Spec != nil {
```
rene-dekker, line 379: Clever.

It looks like we never migrated es-proxy to ui-apis here. I think it is better to add a substitution for that as well.

api/v1/managementcluster_types.go
Josh-L: Yes I've discovered the breakage in my testing. To handle that I'm adding logic to fetch the `ClusterInformation` to check the version and conditionally copy to the old name if it detects that the managed cluster is an older version. It's not in this PR yet because I don't have it fully squared away yet. I got the idea and am borrowing logic from voltron checks that determine impersonation logic based on whether the managed cluster is classifed as "older" or not.

api/v1/manager_types.go
Josh-L: tigera-voltron was already here and marked as deprecated but I'll add tigera-es-proxy back in.

pkg/render/manager.go
Josh-L: I've added similar substitution for tigera-es-proxy -> *-ui-apis here as well.

api/v1/managementcluster_types.go
Josh-L: I misspoke the logic for that is in the token-controller of linseed and therefore can be seen in my calico-private companion commit to this PR.

PR #4163, opened 2025-09-22
separate image path from image name to allow custom defaults

radTuti:
## Description

This update the `Component` struct to have a separation of the image name and image path.

Additionally updated gen-version to allow generating the components file based on the new changes and cleaned it up.

Updated tests based on changes

## Release Note

```release-note
add support for developmental builds of operator with custom image paths for components
```

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/apiserver/apiserver_controller.go
+++ b/pkg/controller/apiserver/apiserver_controller.go
@@ -227,7 +227,13 @@ func (r *ReconcileAPIServer) Reconcile(ctx context.Context, request reconcile.Re
 		if errors.IsNotFound(err) {
 			reqLogger.Info("APIServer config not found")
 			r.status.OnCRNotFound()
-			return reconcile.Result{}, r.maintainFinalizer(ctx, nil)
+			f, err := r.maintainFinalizer(ctx, nil)
+			// If the finalizer is still set, then requeue so we aren't dependent on the periodic reconcile to check and remove the finalizer
+			if f {
+				return reconcile.Result{RequeueAfter: utils.FinalizerRemovalRetry}, nil
+			} else {
+				return reconcile.Result{}, err
+			}
 		}
 		r.status.SetDegraded(operatorv1.ResourceReadError, fmt.Sprintf("An error occurred when querying the APIServer resource: %s", msg), err, reqLogger)
 		return reconcile.Result{}, err
@@ -432,7 +438,7 @@ func (r *ReconcileAPIServer) Reconcile(ctx context.Context, request reconcile.Re
 	}
 
 	// API server exists and configuration is valid - maintain a Finalizer on the installation.
-	if err := r.maintainFinalizer(ctx, instance); err != nil {
+	if _, err := r.maintainFinalizer(ctx, instance); err != nil {
 		r.status.SetDegraded(operatorv1.ResourceReadError, "Error setting finalizer on Installation", err, reqLogger)
 		return reconcile.Result{}, err
 	}
@@ -532,7 +538,8 @@ func validateAPIServerResource(instance *operatorv1.APIServer) error {
 // We add a finalizer to the Installation when the API server has been installed, and only remove that finalizer when
 // the API server has been deleted and its pods have stopped running. This allows for a graceful cleanup of API server resources
 // prior to the CNI plugin being removed.
-func (r *ReconcileAPIServer) maintainFinalizer(ctx context.Context, apiserver client.Object) error {
+// The bool return value indicates if the finalizer is Set
+func (r *ReconcileAPIServer) maintainFinalizer(ctx context.Context, apiserver client.Object) (bool, error) {
 	// These objects require graceful termination before the CNI plugin is torn down.
 	apiServerDeployment := v1.Deployment{ObjectMeta: metav1.ObjectMeta{Name: "calico-apiserver", Namespace: render.APIServerNamespace}}
 	return utils.MaintainInstallationFinalizer(ctx, r.client, apiserver, render.APIServerFinalizer, &apiServerDeployment)
--- a/pkg/controller/clusterconnection/clusterconnection_controller.go
+++ b/pkg/controller/clusterconnection/clusterconnection_controller.go
@@ -218,7 +218,13 @@ func (r *ReconcileConnection) Reconcile(ctx context.Context, request reconcile.R
 		return result, err
 	} else if managementClusterConnection == nil {
 		r.status.OnCRNotFound()
-		return result, r.maintainFinalizer(ctx, nil)
+		f, err := r.maintainFinalizer(ctx, nil)
+		// If the finalizer is still set, then requeue so we aren't dependent on the periodic reconcile to check and remove the finalizer
+		if f {
+			return reconcile.Result{RequeueAfter: utils.FinalizerRemovalRetry}, nil
+		} else {
+			return reconcile.Result{}, err
+		}
 	}
 	r.status.OnCRFound()
 	// SetMetaData in the TigeraStatus such as observedGenerations.
@@ -270,7 +276,7 @@ func (r *ReconcileConnection) Reconcile(ctx context.Context, request reconcile.R
 		r.status.SetDegraded(operatorv1.ResourceUpdateError, err.Error(), err, reqLogger)
 	}
 
-	if err = r.maintainFinalizer(ctx, managementClusterConnection); err != nil {
+	if _, err = r.maintainFinalizer(ctx, managementClusterConnection); err != nil {
 		r.status.SetDegraded(operatorv1.ResourceReadError, "Error setting finalizer on Installation", err, reqLogger)
 		return reconcile.Result{}, err
 	}
@@ -501,7 +507,8 @@ func (r *ReconcileConnection) Reconcile(ctx context.Context, request reconcile.R
 	return result, nil
 }
 
-func (r *ReconcileConnection) maintainFinalizer(ctx context.Context, managementClusterConnection client.Object) error {
+// The bool return value indicates if the finalizer is Set
+func (r *ReconcileConnection) maintainFinalizer(ctx context.Context, managementClusterConnection client.Object) (bool, error) {
 	// These objects require graceful termination before the CNI plugin is torn down.
 	guardianDeployment := v1.Deployment{ObjectMeta: metav1.ObjectMeta{Name: render.GuardianDeploymentName, Namespace: render.GuardianNamespace}}
 	return utils.MaintainInstallationFinalizer(ctx, r.cli, managementClusterConnection, render.GuardianFinalizer, &guardianDeployment)
@@ -513,6 +520,7 @@ func validate(cr *operatorv1.ManagementClusterConnection, variant operatorv1.Pro
 	}
 	return nil
 }
+
 func fillDefaults(cr *operatorv1.ManagementClusterConnection, variant operatorv1.ProductVariant) {
 	if cr.Spec.TLS == nil {
 		cr.Spec.TLS = &operatorv1.ManagementClusterTLS{}
--- a/pkg/controller/gatewayapi/gatewayapi_controller.go
+++ b/pkg/controller/gatewayapi/gatewayapi_controller.go
@@ -165,7 +165,13 @@ func (r *ReconcileGatewayAPI) Reconcile(ctx context.Context, request reconcile.R
 		if errors.IsNotFound(err) {
 			reqLogger.V(2).Info("GatewayAPI object not found")
 			r.status.OnCRNotFound()
-			return reconcile.Result{}, r.maintainFinalizer(ctx, nil)
+			f, err := r.maintainFinalizer(ctx, nil)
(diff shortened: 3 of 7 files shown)


pkg/components/images_test.go
```diff
@@ -85,27 +85,27 @@ var _ = Describe("test GetReference", func() {
...
-				Expect(GetReference(c, "", "userpath", "", nil)).To(Equal(fmt.Sprintf("%s%s:%s", registry, image, c.Version)))
+				Expect(GetReference(c, "", "userpath", "", nil)).To(Equal(fmt.Sprintf("%suserpath/%s:%s", registry, image, c.Version)))
 			},
-			Entry("a calico image correctly", ComponentCalicoNode, CalicoRegistry, "userpath/node"),
-			Entry("a tigera image correctly", ComponentTigeraNode, TigeraRegistry, "userpath/node"),
-			Entry("an ECK image correctly", ComponentElasticsearchOperator, TigeraRegistry, "userpath/eck-operator"),
-			Entry("an operator init image correctly", ComponentOperatorInit, InitRegistry, "userpath/operator"),
-			Entry("a CSR init image correctly", ComponentCalicoCSRInitContainer, CalicoRegistry, "userpath/key-cert-provisioner"),
-			Entry("a CSR init image correctly", ComponentTigeraCSRInitContainer, TigeraRegistry, "userpath/key-cert-provisioner"),
+			Entry("a calico image correctly", ComponentCalicoNode, CalicoRegistry, "node"),
```
tmjd: You should also remove the last argument to each of the Entry lines because that image name is part of the Component that is passed to the Entry, so the entry can use `c.ImageName` instead of needing the last argument.

tmjd: Same for the other changes at 102 and 127

pkg/components/operator.go
```diff
@@ -0,0 +1,23 @@
...
+// limitations under the License.
+
+package components
+
+import "github.com/tigera/operator/version"
+
+var ComponentOperatorInit = Component{
+	Version:   version.VERSION,
+	ImageName: "operator",
+	ImagePath: InitImagePath,
```
tmjd: I'm not sure why but we set `Registry: ""` for the enterprise and calico components so I guess we should do the same here.

pkg/components/references.go
```diff
@@ -32,42 +37,22 @@ type Component struct {
...
-			ComponentCalicoCSIRegistrarFIPS,
-			ComponentCalicoEnvoyGateway,
-			ComponentCalicoEnvoyProxy,
-			ComponentCalicoEnvoyRatelimit,
-			ComponentCalicoGoldmane,
-			ComponentCalicoWhisker,
-			ComponentCalicoWhiskerBackend,
-			ComponentCalicoGuardian:
-
+		switch {
```
tmjd: Instead of adding hard coded ImagePaths to pkg/components/enterprise.go, pkg/components/calico.go, hack/gen-versions/enterprise.go.tpl, and hack/gen-versions/calico.go.tpl I think we should just select the imagePath that is used for an image as part of this switch.

pkg/components/references.go
```diff
@@ -85,41 +70,29 @@ func GetReference(c Component, registry, imagePath, imagePrefix string, is *oper
...
+		if !strings.HasSuffix(imagePath, "/") {
+			imagePath = fmt.Sprintf("%s/", imagePath)
+		}
+		image = fmt.Sprintf("%s%s", imagePath, image)
+	} else {
+		image = fmt.Sprintf("%s%s", c.ImagePath, image)
 	}
 
 	if is == nil {
 		return fmt.Sprintf("%s%s:%s", registry, image, c.Version), nil
```
tmjd: I think this line should include the imagePath and be changed to
`return fmt.Sprintf("%s%s%s:%s", registry, imagePath, image, c.Version), nil`.
Along with that it would mean removing merging the image and imagePath above.

pkg/components/references.go
```diff
@@ -85,41 +70,29 @@ func GetReference(c Component, registry, imagePath, imagePrefix string, is *oper
...
 	}
 
 	for _, img := range is.Spec.Images {
-		if img.Image == c.Image {
+		if img.Image == c.Image() {
 			return fmt.Sprintf("%s%s@%s", registry, image, img.Digest), nil
```
tmjd: If my previous suggestion to merge the registry, imagePath, and image on the return line is done, then this line needs imagePath added also.

hack/gen-versions/components.go
```diff
@@ -17,48 +17,51 @@ package main
...
+		"calico/cni":                  "cni",
+		"calico/cni-windows":          "cni-windows",
+		"calico/dikastes":             "dikastes",
+		"calico/kube-controllers":     "kube-controllers",
+		"calico/node":                 "node",
+		"calico/node-windows":         "node-windows",
+		"calico/goldmane":             "goldmane",
+		"calico/guardian":             "guardian",
+		"calico/whisker":              "whisker",
+		"calico/whisker-backend":      "whisker-backend",
```
tmjd: Is the  calico_versions.yml going to have the "imagePath" portion removed? If so we should probably duplicate all these lines (with a path) so we have for example:
```
"calico/cni":    "cni",
"cni":                "cni",
```
so that the map works with and without it.

tmjd: I just noticed you updated config/enterprise_versions.yml in this PR, shouldn't config/calico_versions.yml have the same update?

pkg/components/references.go
```diff
@@ -85,41 +70,29 @@ func GetReference(c Component, registry, imagePath, imagePrefix string, is *oper
...
+		image = fmt.Sprintf("%s%s", c.ImagePath, image)
 	}
 
 	if is == nil {
 		return fmt.Sprintf("%s%s:%s", registry, image, c.Version), nil
 	}
 
 	for _, img := range is.Spec.Images {
-		if img.Image == c.Image {
+		if img.Image == c.Image() {
```
tmjd: With my other suggestion to remove ImagePath from the Component, that would mean this would need to change to something like
` if img.Image == fmt.Sprintf("%s/%s", defaultImagePath, c.ImageName) {`

hack/gen-versions/components.go
radTuti: `calico_versions.yml` does not have `image` specified for its components.
I will add the duplication as it will be nice to remove the "calico/" in the versions.yml here and in monorepo. Once changes have been merged to the monorepo, I will create a separate PR to remvoe the duplications.

pkg/components/operator.go
radTuti: the `Registry` is not set for `ComponentOperatorInit`. 
For calico & enterprise components, it done for gen-versions when the versions.yml file passed sets registry for the calico & enterprise components respectively explictly.

radTuti:
> I think we should remove the Registry and ImagePath from the Component struct. The registry is currently unused AFAICT and the ImagePath is effectively hardcoded for each Component with all Components for Calico having one ImagePath and all for Enterrpise having another. We have the switch case where we already switch for Calico components, Enterprise components, or the Operator init. So I think we should use that switch case to get both the registry and the imagePath.

Since some components have the same image name across both calico and enterprise, I have left the `ImagePath` (as `imagePath`) to allow differentiating those components when trying to get their default registry & imagePath.

Also, there is still some value in having the Registry for developer workflow e.g. specific components in a different registry. A recent use of the `Registry` was in [early implementation of gateway API envoy](https://github.com/tigera/operator/pull/3638/files#diff-17f0af61a90c84b252a49225929a83114196ef14ee0a3b7493f6d508d142708d) where it needed to specify a different registry for the envoy components.

tmjd:
> Since some components have the same image name across both calico and enterprise, I have left the `ImagePath` (as `imagePath`) to allow differentiating those components when trying to get their default registry & imagePath.

When do we need the registry and imagePath to differentiate those components? In the operator there is a different Component for calico and enterprise components so I don't see how they would ever need to be differentiated by the registry and path.

> Also, there is still some value in having the Registry for developer workflow e.g. specific components in a different registry. A recent use of the `Registry` was in [early implementation of gateway API envoy](https://github.com/tigera/operator/pull/3638/files#diff-17f0af61a90c84b252a49225929a83114196ef14ee0a3b7493f6d508d142708d) where it needed to specify a different registry for the envoy components.

It seems like it is easy enough to add custom code to components.GetReference() to set custom registries for particular images. I'm not sure this fairly rare case of adding new images from different registries is a good reason to keep all that.
Something like this would be enough
```
switch c {
case ComponentNew1, ComponentNew2:
  registry = "temporary.io"
}
```

tmjd:
>When do we need the registry and imagePath to differentiate those components? In the operator there is a different Component for calico and enterprise components so I don't see how they would ever need to be differentiated by the registry and path.

To answer my own question a bit, I guess this is needed with the detection done in components.GetReferences. Hmmm. I wonder if we should be more explicit and have a "variant" field or something like that in the Components that is Calico or Enterprise, to ensure they can't be mixed up.

hack/gen-versions/calico.go.tpl
```diff
@@ -27,8 +25,20 @@ var (
...
-{{- end }}
-{{ with index .Components "calico/cni" }}
+
+	ComponentCalicoCNIFIPS = Component{
+		Version:  "{{ .Version }}-fips",
+		Image:    "{{ .Image }}",
+		Registry: "{{ .Registry }}",
+	}
+{{- end -}}
+{{ with index .Components.cni }}
```
tmjd, line 22: I don't think it would be necessary to put duplicates in this template file if the hack/gen-versions/components.go was updated to strip any path off the `key` before the line
```
cv.Components[key] = component
```

tmjd: I'm not super confident on this but I think it would work.

hack/gen-versions/components.go
tmjd: If it works to do https://github.com/tigera/operator/pull/4163/files#r2402990554 then I don't think it would be necessary to add those duplicates to defaultImages like I originally suggested in the first place here.

hack/gen-versions/calico.go.tpl
tmjd: Can you try removing all the sections with `{{ with index .Components "calico/<image>" }}` since you've stripped the `calico` from all the keys?

hack/gen-versions/enterprise.go.tpl
```diff
@@ -77,14 +77,12 @@ var (
 {{- end }}
 {{ with index .Components "eck-elasticsearch" }}
 	ComponentEckElasticsearch = Component{
-		Version:  "{{ .Version }}",
-		Registry: "{{ .Registry }}",
+		Version: "{{ .Version }}",
```
tmjd, line 88: I think you should add the variant here

pkg/components/operator.go
```diff
@@ -0,0 +1,22 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package components
+
+import "github.com/tigera/operator/version"
+
+var ComponentOperatorInit = Component{
+	Version: version.VERSION,
+	Image:   "operator",
```
tmjd, line 21: I think you should add the variant here too

pkg/components/references.go
```diff
@@ -21,58 +21,58 @@ import (
...
 }
 
 const UseDefault = "UseDefault"
 
+// getDefaults returns the default registry and imagePath for a given component.
+// This is used when no registry is explicitly defined by the component
+// and user does not explicitly specify a registry or imagePath.
+func getDefaults(c Component) (registry string, imagePath string) {
+	switch {
+	// If the component does not specify a variant, it is assumed to be an operator component.
```
tmjd: We're not really using the specific fact that the variant isn't set on the operator component here.

hack/gen-versions/enterprise.go.tpl
radTuti: will add variant for all enterpise components.

pkg/components/operator.go
radTuti: i think operator should not specify one as it does not belong to either product. 
I have added a variant for enterprise instead and updated the checks based on variants.

pkg/components/references.go
radTuti: will fix

hack/gen-versions/calico.go.tpl
```diff
@@ -17,189 +17,348 @@
...
+	ComponentCalicoCNI = Component{
+		Version:  "{{ .Version }}",
+		Image:    "{{ .Image }}",
+		Registry: "{{ .Registry }}",
+		variant:  calicoVariant,
+	}
+
```
tmjd, line 29: You've still got duplicates of all the components in this file that I think it should be possible to remove since you added this https://github.com/tigera/operator/pull/4163/files#diff-4b96c2c907b7124cb5bef1f77e2e981ccf2b6c1f6bd2a1aa63561ce31f9c7716R113

Maybe you missed my comment here https://github.com/tigera/operator/pull/4163#discussion_r2406042810 because it was in a Resolved comment, maybe i should have Unresolved that.

PR #4065, opened 2025-07-30
Migrate policy recommendation components into calico-system

vara2504:
## Release Note

```release-note
TBD
```

## For PR author

- [ ] Tests for change.
- [ ] If changing pkg/apis/, run `make gen-files`
- [ ] If changing versions, run `make gen-versions`

## For PR reviewers

A note for code reviewers - all pull requests must have the following:

- [ ] Milestone set according to targeted release.
- [ ] Appropriate labels:
  - `kind/bug` if this is a bugfix.
  - `kind/enhancement` if this is a a new feature.
  - `enterprise` if this PR applies to Calico Enterprise only.

PR #4172, opened 2025-09-24
Fix PSS conflict in PolicyRecommendation controller

gantony:
### Issue Summary
The `calico-system` namespace is experiencing Pod Security Standard (PSS) conflicts, causing CSI driver pods to fail startup with security policy violations.

### Root Cause
**Commit:** 27214c44 - "Migrate policy recommendation components into calico-system (#4065)"  
**Date:** August 15, 2025

This commit introduced a critical bug by moving PolicyRecommendation components from their dedicated namespace (`tigera-policy-recommendation`) into the `calico-system` namespace without properly handling Pod Security Standards.

#### Before the Migration (Correct State)
```go
// PolicyRecommendation had its own namespace
PolicyRecommendationNamespace = PolicyRecommendationName  // "tigera-policy-recommendation"

// Controllers managed different namespaces with appropriate PSS:
// - Installation controller: calico-system → PSS privileged ✅
// - PolicyRecommendation controller: tigera-policy-recommendation → PSS restricted ✅
```

#### After the Migration (Broken State)
```go
// PolicyRecommendation moved to calico-system
PolicyRecommendationNamespace = common.CalicoNamespace  // "calico-system"

// Controllers now conflict on the same namespace:
// - Installation controller: calico-system → PSS privileged ✅
// - PolicyRecommendation controller: calico-system → PSS restricted ❌
```

### Evidence of the Bug

**Code Location:** `/pkg/controller/policyrecommendation/policyrecommendation_controller.go:479`
```go
setUp := render.NewSetup(&render.SetUpConfiguration{
    Namespace: helper.InstallNamespace(),  // "calico-system"
    PSS:       render.PSSRestricted,       // ❌ CONFLICT!
    // ...
})
```

**Live Confirmation:** Triggering PolicyRecommendation reconciliation produces:
```
{"level":"info","ts":"2025-09-24T11:23:28Z","logger":"controller_policy_recommendation","msg":"Reconciling PolicyRecommendation"}
{"level":"info","ts":"2025-09-24T11:23:28Z","logger":"KubeAPIWarningLogger","msg":"existing pods in namespace \"calico-system\" violate the new PodSecurity enforce level \"restricted:latest\""}
```

### Impact
- CSI driver pods fail to start due to PSS violations
- Core Calico components generate security policy warnings
- Kubernetes rejects the `restricted` PSS change, keeping namespace at `privileged`
- System remains functional but with degraded security posture warnings

## Release Note

```release-note
Fix calico-system Namespace PSS Conflict where, under certain conditions, the calico-system would end up with a PSS value of `restricted` instead of `privileged`. This started happening on August 15, 2025 (so we may not have released an Enterprise version since).
```

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


pkg/controller/policyrecommendation/policyrecommendation_controller_test.go
```diff
@@ -180,7 +180,7 @@ var _ = Describe("PolicyRecommendation controller tests", func() {
 		Expect(c.Get(ctx, client.ObjectKey{
 			Name: render.PolicyRecommendationNamespace,
```
gantony: This is confusing! But looking at the code we can find:
```go
PolicyRecommendationNamespace = common.CalicoNamespace
```

caseydavenport: Yeah! I think I had wanted to get rid of this assignment in the original PR but we ended up not doing it... still think we probably should!

pkg/controller/policyrecommendation/policyrecommendation_controller.go
```diff
@@ -471,12 +471,24 @@ func (r *ReconcilePolicyRecommendation) Reconcile(ctx context.Context, request r
 		return reconcile.Result{}, err
 	}
 
+	// Determine the appropriate PSS based on the namespace
+	// calico-system requires privileged access, so don't override its PSS
+	targetNamespace := helper.InstallNamespace()
+	var pss render.PodSecurityStandard
+	if targetNamespace == common.CalicoNamespace {
+		// Don't apply PSS restrictions to calico-system as it needs privileged access
+		pss = render.PSSPrivileged
```
caseydavenport: Should we be setting this at all? Why does this controller even touch the namespace in calico-system mode?

It should probably be entirely hands off?

gantony: I think being totally hand off could probably work. My understanding is that the only use case where `targetNamespace != common.CalicoNamespace` is when the tenant namespace is used. Should it touch it there? Do you know if there is anything else needing `render.PSSPrivileged`?

gantony: I asked Claude for his opinion :sweat_smile::
> Is there any controller that sets a namespace to render.PSSPrivileged where the namespace could be a tenant namespace in multi-tenant mode?

The response confirms what I was afraid of:

> Based on my investigation, I found that YES, there are controllers that set namespaces to render.PSSPrivileged where the namespace could be a tenant namespace in multi-tenant mode:
> 
> Controllers that could affect tenant namespaces with PSSPrivileged:
> 1. Compliance Controller (compliance_controller.go):
>   - Sets PSS: render.PSSPrivileged
>   - In multi-tenant mode: helper.InstallNamespace() returns request.Namespace (tenant namespace)
>   - In single-tenant mode: uses render.ComplianceNamespace ("tigera-compliance")
> 2. Intrusion Detection Controller (intrusiondetection_controller.go):
>   - Sets PSS: getPSS(lc) which can return render.PSSPrivileged when syslog forwarding is enabled
>   - In multi-tenant mode: helper.InstallNamespace() returns request.Namespace (tenant namespace)
>   - In single-tenant mode: uses render.IntrusionDetectionNamespace ("tigera-intrusion-detection")
> 
> Controllers that set PSSPrivileged but NOT in tenant namespaces:
> 1. LogCollector Controller: Always uses hardcoded render.LogCollectorNamespace ("tigera-fluentd"), not tenant-aware for namespace placement
> 1. Gateway API: Uses "tigera-gateway" namespace from external manifests, not tenant-related
> 1. Various render components: Set PSS for system namespaces like "calico-system", "tigera-elasticsearch", etc.
> 
> The Critical Issue:
> In multi-tenant mode, when the Compliance and Intrusion Detection controllers run, they can potentially set tenant namespaces to PSSPrivileged, which would conflict with the expected PSSRestricted that tenant namespaces should have. This could create the same type of race condition you were investigating with the PolicyRecommendation controller!

So it seems there is a bigger issue at play here. I feel we have 2 options:
 - Use `PSSRestricted` everywhere in multi-tenant namespace (not really an option I would imagine)
 - In multi-tenant mode, manage the `pss` value outside of individual controller (set it to `PSSPrivileged` when creating the tenant's namespace)

rene-dekker: To me this seems off:

```
	setUp := render.NewSetup(&render.SetUpConfiguration{
		OpenShift:       r.provider.IsOpenShift(),
		Installation:    installation,
		PullSecrets:     pullSecrets,
		Namespace:       helper.InstallNamespace(),
		PSS:             render.PSSRestricted,
		CreateNamespace: !tenant.MultiTenant(), <====
	})
```
So if multitenant is false, we create the namespace. That is wrong, this controller has no business creating calico-system. Or am I missing something here?

caseydavenport: Frankly I don't trust Claude very much for this type of thing :sweat_smile: - Claude only knows what the current code does, and the current code is bugged.

I think it's pretty clear that this controller should never modify the calico-system namespace, on principle.

> 		CreateNamespace: !tenant.MultiTenant(), <====

Yeah, seems like create namespace should always be false now? In single-tenant, this goes in calico-system which is managed by another controller. In multi-tenant this goes into the tenant namespace, which is also created elsewhere?

gantony: Yes to set it to false here.

I think I also need to set pss to privileged when creating the tenant namespace, and not set it anywhere else in multi tenant mode. WDYT?

caseydavenport: > Yeah, seems like create namespace should always be false now?

If it's always false, it should just be removed so we aren't carrying that dead logic forward :+1: 

> I think I also need to set pss to privileged when creating the tenant namespace, and not set it anywhere else in multi tenant mode. WDYT?

Yeah, I think this is already the case because of this: `CreateNamespace: !tenant.MultiTenant()` - the issue is that since we've consolidated namespaces, we no longer need to do this even in single tenant mode. So I think this only applies to components that have been migrated to calico-system.

pkg/controller/policyrecommendation/policyrecommendation_controller.go
```diff
@@ -471,13 +471,14 @@ func (r *ReconcilePolicyRecommendation) Reconcile(ctx context.Context, request r
 		return reconcile.Result{}, err
 	}
 
+	targetNamespace := helper.InstallNamespace()
+
 	setUp := render.NewSetup(&render.SetUpConfiguration{
```
caseydavenport: @gantony do we need to call `NewSetup` at all here? 

`NewSetup` seems to do this:

- Create a namespace (if configured)
- Copy pull secrets
- Create a secrets binding for RBAC

When running in calico-system, these should all already be done for us, right?

Same would go when deploying into an existing tenant namespace?

caseydavenport (approved):
LGTM, think it's just file generation that needs updating.

PR #4178, opened 2025-09-29
Monitor prometheus-operator deployment

alexh-tigera:
## Description

A failed deploy of the prometheus-operator wasn't detected as unhealthy, which caused the cluster installation to proceed, and fail at a later point. We should monitor the status of the prometheus-operator deployment to be sure the cluster is ready.

## Release Note

```release-note
Monitor the status of the `prometheus-operator` deployment. This allows `TigeraStatus` to be unhealthy if the prometheus-operator deploy itself fails.
```

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/monitor/monitor_controller.go
+++ b/pkg/controller/monitor/monitor_controller.go
@@ -108,11 +108,6 @@ func newReconciler(mgr manager.Manager, opts options.AddOptions, prometheusReady
 		multiTenant:     opts.MultiTenant,
 	}
 
-	r.status.AddStatefulSets([]types.NamespacedName{
-		{Namespace: common.TigeraPrometheusNamespace, Name: fmt.Sprintf("alertmanager-%s", monitor.CalicoNodeAlertmanager)},
-		{Namespace: common.TigeraPrometheusNamespace, Name: fmt.Sprintf("prometheus-%s", monitor.CalicoNodePrometheus)},
-	})
-
 	r.status.Run(opts.ShutdownContext)
 	return r
 }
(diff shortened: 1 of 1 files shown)


pkg/controller/monitor/monitor_controller.go
```diff
@@ -113,6 +114,10 @@ func newReconciler(mgr manager.Manager, opts options.AddOptions, prometheusReady
 		{Namespace: common.TigeraPrometheusNamespace, Name: fmt.Sprintf("prometheus-%s", monitor.CalicoNodePrometheus)},
 	})
 
+	r.status.AddDeployments([]types.NamespacedName{
```
caseydavenport: I think this (and the existing AddStatefulSets) calls are actually bugged.

If you look at the status manager code, the way it expects clients to use these functions is to set the objects to monitor as part of reconcile - otherwise, a user who creates / deletes / creates the CR for this controller will clear out the values set here during initialization / construction:

https://github.com/tigera/operator/blob/dbcc686666711987b96718ee66ee77c2a2c062b0/pkg/controller/status/status.go#L262-L265

caseydavenport: I am a bit surprised though - I would expect AddDeployments to already be called from here:

https://github.com/tigera/operator/blob/dbcc686666711987b96718ee66ee77c2a2c062b0/pkg/controller/utils/component.go#L484-L498

alexh-tigera: Ah, I see how the other code is using the Add* functions. I'll see if I can follow that pattern instead.

alexh-tigera: > I am a bit surprised though - I would expect AddDeployments to already be called from here:

I'm still learning about how the operator works. It seems like `CreateOrUpdateOrDelete()` is only called on resources that are being created, but the prometheus resources are being created by the prometheus-operator. Is that a special case?

caseydavenport: Ah yes, I missed that nuance - the objects created by the Prometheus operator itself won't get automatically added to our status monitoring :+1:

PR #4214, opened 2025-10-15
Include prometheus and alertmanager status in monitor status

alexh-tigera:
## Description

The monitor should be unavailable if there is a prometheus or alertmanager instance that is unavailable. This handles the case where the prometheus install failed, and the statefulsets haven't been created yet.

## Release Note

```release-note
Fixed a race condition in tigerastatus monitor where the alertmanager and prometheus statefulsets hadn't been created yet, but the monitor was marked as Available.
```

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/monitor/monitor_controller.go
+++ b/pkg/controller/monitor/monitor_controller.go
@@ -364,15 +364,17 @@ func (r *ReconcileMonitor) Reconcile(ctx context.Context, request reconcile.Requ
 	}
 
 	if p != nil {
+		available := monitoringv1.ConditionFalse
+
 		for _, cond := range p.Status.Conditions {
-			if cond.Type == monitoringv1.Available && cond.Status != monitoringv1.ConditionTrue {
-				r.status.SetDegraded(operatorv1.ResourceNotReady, "Prometheus component is not available", err, reqLogger)
-				return reconcile.Result{}, err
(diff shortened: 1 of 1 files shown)


pkg/controller/monitor/monitor_controller.go
```diff
@@ -356,6 +357,44 @@ func (r *ReconcileMonitor) Reconcile(ctx context.Context, request reconcile.Requ
...
+		return reconcile.Result{}, err
+	}
+
+	if p != nil {
+		for _, cond := range p.Status.Conditions {
+			if cond.Type == monitoringv1.Available && cond.Status != monitoringv1.ConditionTrue {
+				r.status.SetDegraded(operatorv1.ResourceNotReady, "Prometheus component is not available", err, reqLogger)
+				return reconcile.Result{}, err
+			}
+			if cond.Type == monitoringv1.Reconciled && cond.Status != monitoringv1.ConditionTrue {
```
rene-dekker: Just curious what made you decide to add this condition check as well? Is there a situation where this could be false, while the Available condition is not?

alexh-tigera: Good point. We don't want to check Reconciled as it can be False during an upgrade, but still available. 

Fixed in 7aec2000

PR #4280, opened 2025-12-03
Add CEL validation for object names

caseydavenport:
## Description

We can use CEL validation to enforce the singleton nature of many of our CRs.

This is hopefully the first of many CEL improvements we can make to our
API validation.

xref: https://kubernetes.io/docs/reference/using-api/cel/

## Release Note

```release-note
Use CEL validation for CR names. 
```

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/test/mainline_test.go
+++ b/test/mainline_test.go
@@ -230,53 +230,6 @@ var _ = Describe("Mainline component function tests", func() {
 	})
 })
 
-var _ = Describe("Mainline component function tests with ignored resource", func() {
-	var c client.Client
-	var mgr manager.Manager
-	var shutdownContext context.Context
-	var cancel context.CancelFunc
-
-	BeforeEach(func() {
-		c, shutdownContext, cancel, mgr = setupManager(ManageCRDsDisable, SingleTenant, EnterpriseCRDsExist)
-		verifyCRDsExist(c, operator.TigeraSecureEnterprise)
-	})
-
-	AfterEach(func() {
-		removeInstallation(context.Background(), c, "not-default")
-	})
-
-	It("Should ignore a CRD resource not named 'default'", func() {
-		By("Creating a CRD resource not named default")
-		instance := &operator.Installation{
-			TypeMeta:   metav1.TypeMeta{Kind: "Installation", APIVersion: "operator.tigera.io/v1"},
-			ObjectMeta: metav1.ObjectMeta{Name: "not-default"},
-			Spec:       operator.InstallationSpec{},
-		}
-		err := c.Create(context.Background(), instance)
-		Expect(err).NotTo(HaveOccurred())
-
-		By("Running the operator")
-		done := RunOperator(mgr, shutdownContext)
-		defer func() {
-			cancel()
-			Eventually(func() error {
-				select {
-				case <-done:
-					return nil
-				default:
-					return fmt.Errorf("operator did not shutdown")
-				}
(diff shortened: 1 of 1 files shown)


api/v1/installation_types.go
```diff
@@ -977,6 +977,8 @@ type InstallationStatus struct {
 // Installation configures an installation of Calico or Calico Enterprise. At most one instance
 // of this resource is supported. It must be named "default". The Installation API installs core networking
 // and network policy components, and provides general install-time configuration.
+//
+// +kubebuilder:validation:XValidation:rule="self.metadata.name == 'default' || self.metadata.name == 'tigera-secure'", message="resource name must be 'default' or 'tigera-secure'"
```
tmjd: This needs to include `overlay` as an allowed value

api/v1/gatewayapi_types.go
```diff
@@ -132,6 +132,7 @@ type GatewayClassSpec struct {
 //+kubebuilder:object:root=true
 //+kubebuilder:resource:scope=Cluster
 
+// +kubebuilder:validation:XValidation:rule="self.metadata.name == 'default'", message="resource name must be 'default'"
```
tmjd: The code seems to allow for both default and tigera-secure

api/v1/authentication_types.go
```diff
@@ -234,6 +234,7 @@ type UserMatch struct {
 // +kubebuilder:resource:path=authentications,scope=Cluster
 
 // Authentication is the Schema for the authentications API
+// +kubebuilder:validation:XValidation:rule="self.metadata.name == 'default'", message="resource name must be 'default'"
```
tmjd: The code only uses `tigera-secure`

api/v1/gatewayapi_types.go
caseydavenport: Weill update validation but leave the warning message as "default" for now since the code suggests "tigera-secure" is legacy.

api/v1/installation_types.go
```diff
@@ -26,6 +26,25 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+// +kubebuilder:object:root=true
```
tmjd, line 29: I think unless we're moving all the other top level types to the start of the file I think this should go back to where it was.

test/validation_test.go
```diff
@@ -0,0 +1,120 @@
...
+		// build the test name.
+		name := fmt.Sprintf("%T with name %q", tc.cr, tc.cr.GetName())
+		if tc.err != "" {
+			name += " should fail"
+		} else {
+			name += " should pass"
+		}
+
+		t.Run(name, func(t *testing.T) {
+			defer testSetup(t)()
```
tmjd, line 105: Can you explain for me why this needs to be defer'ed? I don't see why.

api/v1/installation_types.go
caseydavenport: Seems overly rigid but sure, I'll move them back.

test/validation_test.go
caseydavenport: It's a common pattern for go test setup, where the equivalent of the "AfterEach" is returned by the setup func and deferred. 

There is no deferred afterach logic yet so if you'd like me to remove it I can, but this is just a common pattern we use in many of our go tests.

PR #4284, opened 2025-12-08
feat(gatewayapi): expose owning gateway info to l7-log-collector

electricjesus:
Adds OWNING_GATEWAY_NAME and OWNING_GATEWAY_NAMESPACE environment variables to the l7-log-collector container. These are populated via the Kubernetes downward API from pod labels set by EnvoyProxy:
- gateway.envoyproxy.io/owning-gateway-name
- gateway.envoyproxy.io/owning-gateway-namespace

This allows the collector to know which Gateway it is processing logs for

## Release Note

```release-note
TBD
```

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/crds/enterprise/crd.projectcalico.org_felixconfigurations.yaml
+++ b/pkg/crds/enterprise/crd.projectcalico.org_felixconfigurations.yaml
@@ -326,12 +326,6 @@ spec:
                     if it detects the current value is 2 (strict mode that hurts performance). When set to "Strict",
                     Felix will not modify the JIT hardening setting. [Default: Auto]
                   type: string
-                bpfKubeProxyEndpointSlicesEnabled:
-                  description: |-
-                    BPFKubeProxyEndpointSlicesEnabled is deprecated and has no effect. BPF
-                    kube-proxy always accepts endpoint slices. This option will be removed in
-                    the next release.
-                  type: boolean
                 bpfKubeProxyHealthzPort:
                   description: |-
                     BPFKubeProxyHealthzPort, in BPF mode, controls the port that Felix's embedded kube-proxy health check server binds to.


pkg/render/gatewayapi/gateway_api.go
```diff
@@ -111,6 +111,26 @@ var (
...
+			FieldRef: &corev1.ObjectFieldSelector{
+				FieldPath: "metadata.labels['gateway.envoyproxy.io/owning-gateway-name']",
+			},
+		},
+	}
+	OwningGatewayNamespaceEnvVar = corev1.EnvVar{
+		Name: "OWNING_GATEWAY_NAMESPACE",
+		ValueFrom: &corev1.EnvVarSource{
+			FieldRef: &corev1.ObjectFieldSelector{
+				FieldPath: "metadata.labels['gateway.envoyproxy.io/owning-gateway-namespace']",
```
LorcanMcVeigh, line 130: @electricjesus have you tested this and seen this working correctly? 
I was going to try do something like this for the WAF policy work but saw that only `metadata.name` and `metadata.namespace` were supported, not `metadata.labels` 
doc : https://kubernetes.io/docs/concepts/overview/working-with-objects/field-selectors/

tmjd: Could a test be added to `test/gatewayapi_test.go` for this?

electricjesus: @LorcanMcVeigh  I'm not certain why field selectors is relevant. But metadata.labels is documented in the DownwardAPI documentation: https://kubernetes.io/docs/concepts/workloads/pods/downward-api/#downwardapi-fieldRef

Yes i've a working, functional test collector because of it.. I'll look into testing this API in this repo itself

LorcanMcVeigh: oh okay. Whenever i looked into it before I never came across the downward api. I only ever got suggested the field selectors.

test/gatewayapi_test.go
```diff
@@ -314,6 +314,101 @@ var _ = Describe("GatewayAPI tests", func() {
...
+					break
+				}
+			}
+			if l7LogCollector == nil {
+				return errors.New("l7-log-collector init container not found")
+			}
+
+			// Verify owning gateway env vars are present
+			var foundName, foundNamespace bool
+			for _, env := range l7LogCollector.Env {
```
tmjd, line 380: We're basically checking the same thing as we are in the render UT. Is there something we can check that shows OWNING_GATEWAY_NAME has the value expected and is being used? I guess another option would be like an exec in the pod and dump the env vars to see that it is the expected value.

tmjd: It is ok to keep for now since in the future when the image logs the env vars at the start we can check the value is as expected.

electricjesus: I've no image that I can use to exec into the pods unless we swap the collector image here with something that we could exec into.. is that acceptable? in the final gateway collector image binary it should be able to spit out the env vars at the start of runtime

*resolved in slack*: we'll keep it this way because it might be too involved to swap images just to dump env vars 

I'll keep this test here

PR #4298, opened 2025-12-10
fix(istio): add NativeNftables field to GlobalConfig

electricjesus:
## Description

set to true by default because we currently don't have iptables-legacy

## Release Note

```release-note
TBD
```

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/istio/istio.go
+++ b/pkg/render/istio/istio.go
@@ -29,6 +29,7 @@ import (
 	operatorv1 "github.com/tigera/operator/api/v1"
 	"github.com/tigera/operator/pkg/common"
 	"github.com/tigera/operator/pkg/components"
+	"github.com/tigera/operator/pkg/ptr"
 	"github.com/tigera/operator/pkg/render"
 	rcomp "github.com/tigera/operator/pkg/render/common/components"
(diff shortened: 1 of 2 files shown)


pkg/render/istio/istio.go
```diff
@@ -109,6 +110,7 @@ func Istio(cfg *Configuration) (*IstioComponentCRDs, *IstioComponent, error) {
 		IstioCNIOpts: IstioCNIOpts{
 			Global: &GlobalConfig{
 				IstioNamespace: IstioNamespace,
+				NativeNftables: ptr.BoolToPtr(true),
```
caseydavenport, line 113: Just to I understand, this configuration only applies to the rules programmed within the application pod (to send traffic to ztunnel) and not host rules, right? 

If so I think it's OK to set this to true unconditionally, otherwise we'll at least need to verify compatibility with the mode Felix is using.

electricjesus: sorry i should've opened this as Draft, used as a discussion item for something @hjiawei and I are looking at right now


but yes this is the rules in the application pod only so it's quite safe

electricjesus:
@caseydavenport update: looks like we don't need this anymore, will close

PR #4306, opened 2025-12-16
Fix Annotation Removal when PatchinFelixConfiguration

radixo:
It was found a problem removing some annotations from felixconfig.

This code does not fix it for Istio controller, that was using `utils.PatchFelixConfiguration`, but fixes the function and introduces the func `utils.RestoreV3Metadata` for when to restor v3 metadata is necessary.

The removal was not happening because v1 FelixConfigurations sotores v3 metadata in annotations, so it was created a controller util function to restore v3 metadata before patching.

The annotations that needed to be removed were:
* `operator.tigera.io/istioAmbientMode`
* `operator.tigera.io/istioDSCPMark`

Both used by the operator to check if it's consistent with the operator resources. They are included when Istio is deployed and removed when Istio resource is removed.

## Release Note

```release-note-not-required
Fix Annotation Removal when patching FelixConfiguration
```

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/crds/calico/crd.projectcalico.org_felixconfigurations.yaml
+++ b/pkg/crds/calico/crd.projectcalico.org_felixconfigurations.yaml
@@ -893,9 +893,14 @@ spec:
                     none to disable file logging. [Default: /var/log/calico/felix.log]"
                   type: string
                 logPrefix:
-                  description:
-                    "LogPrefix is the log prefix that Felix uses when rendering
-                    LOG rules. [Default: calico-packet]"
+                  description: |-
+                    LogPrefix is the log prefix that Felix uses when rendering LOG rules. It is possible to use the following specifiers
+                    to include extra information in the log prefix.
+                    - %t: Tier name.
+                    - %k: Kind (short names).
+                    - %n: Policy or profile name.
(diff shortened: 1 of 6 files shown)


pkg/controller/utils/utils.go
```diff
@@ -1119,6 +1120,22 @@ func AllPodsTerminated(ctx context.Context, c client.Client, obj client.Object)
...
+		v3meta := metav1.ObjectMeta{}
+		err := json.Unmarshal([]byte(v3metaJSON), &v3meta)
+		if err != nil {
+			return err
+		}
+
+		// Restore the v3 metadata we care about.
+		obj.SetLabels(v3meta.Labels)
+		obj.SetAnnotations(v3meta.Annotations)
+		log.V(1).Info("Restored v3 resource metadata", "labels", v3meta.Labels, "annotations", v3meta.Annotations)
```
tmjd, line 1134: How often is going to be logged? Do we need it logged at all? I'd probably suggest this should go to Debug level.

radixo: I believed 1 was Debug, no? I didn't get the log on my tests now, that is the default log level.

PR #4308, opened 2025-12-16
feat(release): add build & publish to release tool

radTuti:
## Description

This adds creating and publishing an operator version. It aligns the process for both a hashrelease operator version and a release operator version. For release operator version, it creates a github release if indicated. 

For hashreleases, there is a new enhancement for creating a hashrelease by providing the calico and/or enterprise version in place of the versions file.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/crds/calico/v3.projectcalico.org/projectcalico.org_bgpconfigurations.yaml
+++ b/pkg/crds/calico/v3.projectcalico.org/projectcalico.org_bgpconfigurations.yaml
@@ -10,6 +10,9 @@ spec:
     kind: BGPConfiguration
     listKind: BGPConfigurationList
     plural: bgpconfigurations
+    shortNames:
+      - bgpconfig
+      - bgpconfigs
     singular: bgpconfiguration
   preserveUnknownFields: false
   scope: Cluster
--- a/pkg/crds/calico/v3.projectcalico.org/projectcalico.org_bgppeers.yaml
+++ b/pkg/crds/calico/v3.projectcalico.org/projectcalico.org_bgppeers.yaml
@@ -14,7 +14,38 @@ spec:
   preserveUnknownFields: false
   scope: Cluster
   versions:
-    - name: v3
+    - additionalPrinterColumns:
+        - description: Selector for the nodes that should have this peering
+          jsonPath: .spec.nodeSelector
+          name: Node Selector
+          type: string
+        - description:
+            The node name identifying the Calico node instance that is targeted
+            by this peer
+          jsonPath: .spec.node
+          name: Node
+          type: string
+        - description:
+            The optional Local AS Number to use when peering with this remote
+            peer
+          jsonPath: .spec.localASNumber
+          name: Local ASN
+          type: string
+        - description: Selector for the remote nodes to peer with
+          jsonPath: .spec.peerSelector
+          name: Peer Selector
+          type: string
+        - description:
+            The IP address of the peer followed by an optional port number
+            to peer with
+          jsonPath: .spec.peerIP
+          name: Peer IP
+          type: string
+        - description: The AS Number of the peer
+          jsonPath: .spec.asNumber
+          name: Peer ASN
+          type: string
+      name: v3
       schema:
         openAPIV3Schema:
           properties:
@@ -102,3 +133,4 @@ spec:
           type: object
       served: true
       storage: true
+      subresources: {}
--- a/pkg/crds/calico/v3.projectcalico.org/projectcalico.org_blockaffinities.yaml
+++ b/pkg/crds/calico/v3.projectcalico.org/projectcalico.org_blockaffinities.yaml
@@ -26,6 +26,10 @@ spec:
           jsonPath: .spec.node
           name: Node
           type: string
+        - description: The type of block affinity
+          jsonPath: .spec.type
+          name: Type
+          type: string
       name: v3
       schema:
         openAPIV3Schema:
--- a/pkg/crds/calico/v3.projectcalico.org/projectcalico.org_caliconodestatuses.yaml
+++ b/pkg/crds/calico/v3.projectcalico.org/projectcalico.org_caliconodestatuses.yaml
@@ -14,7 +14,16 @@ spec:
   preserveUnknownFields: false
   scope: Cluster
   versions:
-    - name: v3
+    - additionalPrinterColumns:
+        - description: The name of the node
+          jsonPath: .spec.node
+          name: Node
+          type: string
+        - description: The types of information to monitor for this calico/node
+          jsonPath: .spec.classes
+          name: Classes
+          type: string
+      name: v3
       schema:
         openAPIV3Schema:
           properties:
@@ -215,3 +224,4 @@ spec:
           type: object
       served: true
       storage: true
+      subresources: {}
--- a/pkg/crds/calico/v3.projectcalico.org/projectcalico.org_clusterinformations.yaml
+++ b/pkg/crds/calico/v3.projectcalico.org/projectcalico.org_clusterinformations.yaml
@@ -17,7 +17,20 @@ spec:
   preserveUnknownFields: false
   scope: Cluster
   versions:
-    - name: v3
+    - additionalPrinterColumns:
+        - description: The version of Calico that the cluster is running
+          jsonPath: .spec.calicoVersion
+          name: Version
+          type: string
+        - description: Whether the datastore is ready for use
+          jsonPath: .spec.datastoreReady
+          name: Ready
+          type: boolean
+        - description: The types associated with the cluster
+          jsonPath: .spec.clusterType
+          name: Types
+          type: string
+      name: v3
       schema:
         openAPIV3Schema:
           properties:
@@ -46,3 +59,4 @@ spec:
           type: object
       served: true
       storage: true
+      subresources: {}
--- a/pkg/crds/calico/v3.projectcalico.org/projectcalico.org_globalnetworkpolicies.yaml
+++ b/pkg/crds/calico/v3.projectcalico.org/projectcalico.org_globalnetworkpolicies.yaml
@@ -10,11 +10,24 @@ spec:
     kind: GlobalNetworkPolicy
     listKind: GlobalNetworkPolicyList
     plural: globalnetworkpolicies
+    shortNames:
+      - gnp
+      - cgnp
     singular: globalnetworkpolicy
   preserveUnknownFields: false
   scope: Cluster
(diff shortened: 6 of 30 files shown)


hack/release/build.go
```diff
@@ -0,0 +1,350 @@
...
+			fmt.Sprintf("PUSH_IMAGE_PREFIXES=%s", addTrailingSlash(registry)))
+	}
+	if c.Bool(hashreleaseFlag.Name) {
+		buildLog = buildLog.WithField("hashrelease", true)
+		buildEnv = append(buildEnv, fmt.Sprintf("GIT_VERSION=%s", c.String(versionFlag.Name)))
+		resetFn, err := hashreleaseBuildConfig(ctx, c, repoRootDir)
+		defer resetFn()
+		if err != nil {
+			return fmt.Errorf("preparing hashrelease build environment: %w", err)
+		}
```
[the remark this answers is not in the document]
radTuti: we always want the resetFn to run in case there were changes before returning the error.

Makefile
```diff
@@ -553,54 +553,29 @@ endif
...
+## Build a release from start to finish.
+release: clean hack/bin/release var-require-all-VERSION
+	hack/bin/release build
 
 ## Produces a clean build of release artifacts at the specified version.
-release-build: release-prereqs clean
+release-build: release-prereqs var-require-all-VERSION-GIT_VERSION
 # Check that the correct code is checked out.
 ifneq ($(VERSION), $(GIT_VERSION))
 	$(error Attempt to build $(VERSION) from $(GIT_VERSION))
```
danudey: So I get this error:

```
Makefile:573: *** Attempt to build v1.42.0 from v1.42.0-0.dev-27-g90245a141b4c.  Stop.
```

It would be nice to add in some more detail to this error message; what do I do to solve this problem? Is this meant to be running on an existing tag? If so, why do I have to specify the `VERSION`? If not, what do I do to tell it to use the `VERSION` I gave it?

hack/release/main.go
```diff
@@ -43,7 +43,10 @@ func app(version string) *cli.Command {
 		Usage:   "CLI tool for releasing operator",
 		Version: version,
 		Commands: []*cli.Command{
```
danudey, line 45:
```suggestion
		Version: version,
		EnableShellCompletion: true,
		Commands: []*cli.Command{
```

hack/release/build.go
```diff
@@ -0,0 +1,361 @@
...
+	} else {
+		buildLog = buildLog.WithField("release", true)
+		buildEnv = append(buildEnv, "RELEASE=true")
+	}
+
+	// Build the Operator and verify the build
+	buildLog.Info("Building Operator")
+	if out, err := makeInDir(repoRootDir, "release-build", buildEnv...); err != nil {
```
danudey, line 214: I think it would be beneficial to add slightly more detail to the process, even without debug. For example, it might be a good idea to have `makeInDir` log what it's running even if it's as simple as "Running `make release-build`...".

My thinking is that a lot of these targets can take a long time to run, so knowing what step it's on (or what command to run to duplicate the step) could be very beneficial). I know that if debug logging is enabled then commands get logged in `runCommandInDir`, but a lot of other output gets logged also, and we don't necessarily want to log every command we run; just adding unconditional logging of args to `makeInDir` could ease debugging in future.

danudey:
> Also did you follow the [readme](https://github.com/radTuti/tigera-operator/blob/de-3868-build-op/hack/release/README.md?rgh-link-date=2025-12-29T19%3A12%3A35Z)?

Nope, I definitely did not. That said, the readme tells you what commands there are, but it would be nice if it also had sections for "how to publish a hashrelease" (e.g. the steps to get all the correct information, check out calico/calient, etc), "how to publish a release" (what parameters need to be provided, etc.). As an example, it doesn't seem to say anywhere that a tag needs to exist ahead of time, nor does the error message indicate that. I know that that problem is in the `Makefile` right now, but if we're updating how the release process works then at least updating that error message would be a nice QoL fix - either be more specific about the error or direct the user to the release tool readme.

hack/release/README.md
```diff
@@ -34,6 +40,96 @@ release --help
...
+```
+
+For hashrelease, use the `--hashrelease` flag and provide either the Calico or Calico Enterprise version or versions file.
+
+```sh
+release build --version <operator version> --hashrelease \
+  [--calico-version <calico version> | --calico-versions <path to calico version file> |--enterprise-version <enterprise version> | --enterprise-versions <path to enterprise version file>]
+```
+
+#### Examples
```
danudey, line 154: The readme has examples that show creating a build/hashrelease/etc with a Calico version or a Calico Enterprise version. Can both be used? e.g. if I wanted to create a hashrelease which included both the latest Calico and Calient? I'm not sure if this is something we'd ever want, but would it work?

hack/release/README.md
```diff
@@ -46,8 +142,8 @@ which involves creating a new milestone for the next patch version and closing t
 for the release version. All open issues and pull requests associated with the current milestone
 are moved to the new milestone.
 
-  > [!NOTE]
-  > At least one of Calico or Calico Enterprise version must be specified.
+  > [!IMPORTANT]
+  > One of Calico or Calico Enterprise version must be specified.
```
danudey: Only one? Or at least one? The phrasing here implies one xor the other, but we've had enough double-releases recently that I would *assume* I can specify both?

hack/release/github.go
```diff
@@ -158,7 +169,7 @@ func (r *GithubRelease) GenerateNotes(ctx context.Context, outputDir string, use
 	}
 	noteData, err := r.collectReleaseNotes(ctx, useLocal)
 	if err != nil {
-		return fmt.Errorf("error collecting release notes: %s", err)
+		return fmt.Errorf("collecting release notes: %s", err)
```
danudey, line 172: Will this process still error if we successfully fetch from Github but there are no PRs with `release-note-required` (i.e. there are no release notes to include)? If not, do we include a note like that in our output?

Also while I'm thinking about it, I have no idea if it's feasible but I wish our release note generation would highlight which new version an Operator release is supporting, e.g. an automatic "This version of Operator is being released to support Calico v3.31.4". Wishlist.

hack/release/github.go
```diff
@@ -339,6 +348,118 @@ func (r *GithubRelease) collectReleaseNotes(ctx context.Context, local bool) (*R
...
+
+// Create a new GitHub release.
+func (r *GithubRelease) Create(ctx context.Context, isDraft, isPrerelease bool) error {
+	// Check if release already exists
+	if got, err := r.repoRelease(ctx); err != nil {
+		return fmt.Errorf("checking if release exists: %w", err)
+	} else if got != nil {
+		logrus.Warnf("GitHub release already exists, update manually: %s", got.GetHTMLURL())
+		return ErrGitHubReleaseExists
+	}
```
danudey, line 386: Is it possible to include some kind of tags in the release notes so that we can find and update an existing release's (or a draft release's) release notes?

```
<!-- start release notes -->
## Bug fixes

* foo
* bar

<!-- end release notes -->
```

The example I'm thinking is we generate a release with its release notes and someone wants to fix/update the release note in their PR; it would be nice to be able to kick off the release tool again to say "Go update the release notes for this release and/or show me the diff".

Not necessary for now, but a thought.

hack/release/README.md
radTuti: It is likely not going to be done but it is possible. Will add an example for that.

RELEASING.md
```diff
@@ -110,11 +100,9 @@ Go to the PR created and:
     git push <remote> <tag> # e.g git push origin v1.30.2
     ```
 
-1. Log in to semaphore and find the new build for the release branch commit, and
-   click 'Rerun'. When Semaphore starts the rebuild, it will notice the new tag and
-   build and publish an operator release.
+1. Log in to semaphore and run the [release task](https://tigera.semaphoreci.com/projects/operator/schedulers/fcc7fc6c-fb81-4a07-b312-138befbeb111).
```
tmjd: Why do we need to manually do this now? Previously pushing the tag would trigger the release process.

tmjd: Or didn't the tigera/operator build off of a tag? I know we updated operator-cloud to build off the tag.

radTuti: I think it got removed. I have added it back

hack/release/github.go
radTuti: will tackle release notes changes in a separate PR

hack/release/github.go
radTuti: will tackle release notes changes in a separate PR

Makefile
```diff
@@ -559,54 +559,29 @@ endif
...
-	$(MAKE) release-build
-	$(MAKE) release-verify
-
-	@echo ""
-	@echo "Release build complete. Next, push the produced images."
-	@echo ""
-	@echo "  make VERSION=$(VERSION) release-publish"
-	@echo ""
+## Build a release from start to finish.
+release: clean hack/bin/release var-require-all-VERSION
```
tmjd: I don't think it is required in this PR but I think the checks `var-require-all-VERSION` should be moved into  hack/bin/release

hack/release/github.go
```diff
@@ -339,6 +348,118 @@ func (r *GithubRelease) collectReleaseNotes(ctx context.Context, local bool) (*R
...
+		return nil
+	}
+	log.Info("GitHub release created")
+	return nil
+}
+
+// Helper function to determine the value for MakeLatest option for GitHub releases.
+// Ideally, we want GitHub to determine if this is the latest release only if it is not a draft or prerelease
+// using "legacy" option. See https://docs.github.com/en/rest/releases/releases#create-a-release
+func makeLatest(isDraft, isPrerelease bool) *string {
```
tmjd: Is this really all that goes into setting latest or not? I would think there also needs to be a check if there is a "higher" minor release.
Like if 1.40.0 exists and the release being created is 1.39.3, then 1.39.3 shouldn't be made the "latest".

hack/release/publish.go
```diff
@@ -0,0 +1,214 @@
...
+	"github.com/google/go-containerregistry/pkg/name"
+	"github.com/google/go-containerregistry/pkg/v1/remote"
+	"github.com/sirupsen/logrus"
+	"github.com/urfave/cli/v3"
+)
+
+// Command to publish release to remote.
+var publishCommand = &cli.Command{
+	Name:  "publish",
+	Usage: "Publish release to remote",
```
tmjd: I think the Usage should include that it pushes images.

PR #4358, opened 2026-01-14
fix(tunnel-secret): Two owners for one resource

rene-dekker:
```
Every 2.0s: kubectl get secret calico-management-cluster-connection -o yaml -n calico-system                                                                                                                                      
...
metadata:
  creationTimestamp: "2026-01-13T23:59:25Z"
  name: calico-management-cluster-connection
  namespace: calico-system
  ownerReferences:
    blockOwnerDeletion: true
    controller: true
    kind: APIServer
    name: tigera-secure
    uid: 88bcd287-6671-437b-9ef1-228659e66904
  resourceVersion: "182749"
  uid: ea542965-a729-4e32-945e-48614ac19ab0
type: Opaque

...

kind: Secret
metadata:
  creationTimestamp: "2026-01-13T23:59:25Z"
  name: calico-management-cluster-connection
  namespace: calico-system
  ownerReferences:
    blockOwnerDeletion: true
    controller: true
    kind: Manager
    name: tigera-secure
    uid: 313ddf8c-1d7f-4abd-99ce-c826b154d88a
  resourceVersion: "185440"
  uid: ea542965-a729-4e32-945e-48614ac19ab0
...
```

The primary controller responsible for this secret is the manager_controller, so we removed the creation logic from the apiserver_controller.

```release-note
Fixed an issue caused by manager_controller and apiserver_controller both writing the calico-management-cluster-connection secret to calico-system causing constant reconciliations.
```

### DIFF
--- a/pkg/controller/apiserver/apiserver_controller.go
+++ b/pkg/controller/apiserver/apiserver_controller.go
@@ -310,7 +310,6 @@ func (r *ReconcileAPIServer) Reconcile(ctx context.Context, request reconcile.Re
 	}
 
 	// Query enterprise-only data.
-	var tunnelCAKeyPair certificatemanagement.KeyPairInterface
 	var trustedBundle certificatemanagement.TrustedBundle
 	var applicationLayer *operatorv1.ApplicationLayer
 	var managementCluster *operatorv1.ManagementCluster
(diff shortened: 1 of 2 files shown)


Josh-L (approved):

pkg/controller/apiserver/apiserver_controller.go
tmjd: I think this comment could use an update because it is still assuming the different namespaces.

pkg/controller/apiserver/apiserver_controller.go
```diff
@@ -350,24 +349,17 @@ func (r *ReconcileAPIServer) Reconcile(ctx context.Context, request reconcile.Re
 			return reconcile.Result{}, err
 		}
 
-		// This block depends on the Manager controller having defaulted the ManagementCluster CR and having created the tunnel CA secret.
-		// If these conditions are not met, this controller does not degrade as the Manager controller needs API server to be ready to accomplish the above.
+		// This block depends on the Manager controller having defaulted the ManagementCluster CR and having created the
+		// tunnel CA secret in the tigera-operator and calico-system namespaces. If these conditions are not met, this
+		// controller does not degrade as the Manager controller needs API server to be ready to accomplish the above.
```
tmjd: I think you updated the comment according to the changes you made but I think it might have originally been incorrect. It sounds like if the secret is not present in the calico-system namespace that this controller should not degrade, but that isn't the case.

tmjd (approved):

PR #4378, opened 2026-01-23
Update nodeSet naming

pasanw:
### Context
tigera-operator configures the eck-operator to deploy a StatefulSet by specifying a NodeSet in the Elasticsearch CR. The name of the resulting StatefulSet is the name of the NodeSet.

tigera-operator sets the NodeSet name by hashing the PersistentVolumeClaim template of the NodeSet. 

This ensures that if the PVC template changes as a result of LogStorage changes, a new name for the NodeSet is generated, which creates a new StatefulSet with the new PVC template (rather than updating the existing StatefulSet).

A new StatefulSet must be created, since a StatefulSet's PVC template can not be updated after the StatefulSet has been created.

### Issue
The upgrade of k8s.io/apimachinery to v0.34.3 has changed how k8s objects serialize. Specifically, `creationTimestamp` now has an `omitzero` json tag. This change has resulted in the same PersistentVolumeClaim resulting in different hashes depending on the version of tigera-operator (and what version of apimachinery it uses)

The primary impact here is that when users upgrade Calico Enteprise to a newer version, which has an operator with apimachinery >= v0.34.3, tigera-operator computes a different name for the NodeSet (due to the serialization change), and this triggers a new StatefulSet and PVC to be created, even though no LogStorage properties changed.

This recreation could lead to one of two things:
- A slow upgrade on clusters with dynamic provisioning, as new PVs are provisioned and filled with the data from the old PVs
- A stalled upgrade on clusters without dynamic provisioning/available PVs, as the new ES pods wait for PVs to be provisioned

### Proposal
This PR changes the behaviour of NodeSet name generation as follows:
* We change the name of a NodeSet when we know that the already existing PVC template of the NodeSet does not match what it should be based on the LogStorage configuration. Otherwise, we keep the existing name.
* When we need to change (or create) a name for a NodeSet, we choose a random string

This has the key property of retaining the existing NodeSet name, regardless of how it was generated, if the existing PVC in the NodeSet matches the configuration found in LogStorage.

### Alternative approaches
It was attempted to simply augment the serialization to retain the `creationTimestamp: null` that previous versions of apimachinery would create. Unfortunately all attempted approaches resulted in field orders changing, which also meant a different hash.

## Release Note

```release-note
Updated Elasticsearch NodeSet name generation to prevent unnecessary recreations of the Elasticsearch StatefulSet.
```

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/logstorage.go
+++ b/pkg/render/logstorage.go
@@ -796,15 +796,16 @@ func nodeSetName(pvcTemplate corev1.PersistentVolumeClaim, currentES *esv1.Elast
 	if currentES == nil {
 		nameMustBeGenerated = true
 	} else {
+		// Check if the existing PVC template matches our LogStorage CR. If it does not, we need to force a rename.
+		// We assume that the VolumeClaimTemplate is the same across all NodeSets (this is how we configure it).
 		currentPVCTemplate := currentES.Spec.NodeSets[0].VolumeClaimTemplates[0]
-
 		storageClassNamesMatch := *pvcTemplate.Spec.StorageClassName == *currentPVCTemplate.Spec.StorageClassName
 		resourceRequirementsMatch := reflect.DeepEqual(currentPVCTemplate.Spec.Resources, pvcTemplate.Spec.Resources)
-
 		if !storageClassNamesMatch || !resourceRequirementsMatch {
 			nameMustBeGenerated = true
 		}
 
+		// Capture the existing name. If there are multiple NodeSets, there will be a suffix, so we strip it.
 		splitExistingName := strings.Split(currentES.Spec.NodeSets[0].Name, "-")[0]
 		existingName = &splitExistingName
 	}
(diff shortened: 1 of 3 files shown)


pkg/render/logstorage.go
```diff
@@ -769,24 +770,59 @@ func (es *elasticsearchComponent) nodeSetTemplate(pvcTemplate corev1.PersistentV
...
+// NodeSet does not match the storage requirements outlined by LogStorage. This rename triggers the eck-operator
+// to create a new NodeSet, which is required because VolumeClaimTemplates on a StatefulSet are immutable.
+// https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-orchestration.html
+func nodeSetName(pvcTemplate corev1.PersistentVolumeClaim, currentES *esv1.Elasticsearch) string {
+	// We expect the following preconditions to be met, but we check for them explicitly to indicate developer error.
+	if currentES != nil && len(currentES.Spec.NodeSets) == 0 {
+		panic("ElasticSearch CR has no NodeSets")
+	}
+	if currentES != nil && len(currentES.Spec.NodeSets[0].VolumeClaimTemplates) != 1 {
+		panic(fmt.Sprintf("ElasticSearch CR has %d VolumeClaimTemplates in it's NodeSets, expected 1", len(currentES.Spec.NodeSets[0].VolumeClaimTemplates)))
```
rene-dekker:
```suggestion
		panic(fmt.Sprintf("ElasticSearch CR has %d VolumeClaimTemplates in its NodeSets, expected 1", len(currentES.Spec.NodeSets[0].VolumeClaimTemplates)))
```

rene-dekker (approved):

PR #4382, opened 2026-01-24
Support for custom CNI file name

frozenprocess:
## Description
This PR allows users to change the filename via installation manfiest.

example:
```
kind: Installation
metadata:
  name: default
spec:
  cni:
    type: Calico
    ipam:
      type: Calico
    confName: "00-calico.conflist"
```
## Release Note

```release-note
Installation manifest now allows users to change the CNI file name using the `confName` value under the subtree of `cni`.
```

## For PR author

- [ ] Tests for change.
- [ x ] If changing pkg/apis/, run `make gen-files`
- [ x ] If changing versions, run `make gen-versions`

## For PR reviewers

- [ ] Appropriate labels:


api/v1/installation_types.go
```diff
@@ -957,6 +957,14 @@ type CNISpec struct {
 	// +optional
 	// +kubebuilder:validation:Type=string
 	ConfDir *string `json:"confDir,omitempty"`
+
+	// ConfName is the name of the CNI config file.
+	// If you have changed the name of the CNI configuration file in the container runtime configuration,
+	// please ensure that this field matches the same name as specified in the container runtime settings.
+	// Default: "10-calico.conflist"
+	// +optional
+	// +kubebuilder:validation:Type=string
```
caseydavenport, line 966: We should add a kubebuilder default here as well 

```suggestion
	// +kubebuilder:validation:Type=string
	// +kubebuilder:default=10-calico.conflist
```

api/v1/installation_types.go
```diff
@@ -957,6 +957,14 @@ type CNISpec struct {
 	// +optional
 	// +kubebuilder:validation:Type=string
 	ConfDir *string `json:"confDir,omitempty"`
+
+	// ConfName is the name of the CNI config file.
+	// If you have changed the name of the CNI configuration file in the container runtime configuration,
+	// please ensure that this field matches the same name as specified in the container runtime settings.
```
caseydavenport, line 963: I don't think this field needs to match anything in the container runtime settings (looks like copy/paste from the ConfDir param)

pkg/controller/installation/core_controller.go
```diff
@@ -636,6 +636,12 @@ func fillDefaults(instance *operatorv1.Installation, currentPools *crdv1.IPPoolL
 		instance.Spec.CNI.BinDir = &defaultCNIBinDir
 	}
 
+	// For compatability reasons, default the CNI config name if not specified.
+	if instance.Spec.CNI.ConfName == nil {
+		defaultCNIConfName := "10-calico.conflist"
+		instance.Spec.CNI.ConfName = &defaultCNIConfName
```
caseydavenport, line 642:
```suggestion
		instance.Spec.CNI.ConfName = ptr.To("10-calico.conflist")
```

pkg/controller/migration/convert/core.go
```diff
@@ -184,7 +184,7 @@ func handleCore(c *components, install *operatorv1.Installation) error {
 			}
 		}
 
-		if err := c.node.assertEnv(ctx, c.client, containerInstallCNI, "CNI_CONF_NAME", "10-calico.conflist"); err != nil {
+		if err := c.node.assertEnv(ctx, c.client, containerInstallCNI, "CNI_CONF_NAME", *install.Spec.CNI.ConfName); err != nil {
```
caseydavenport, line 187: Instead of asserting the env var matches, we should get the CNI_CONF_NAME env var and update the `install.Spec.CNI.ConfName` field to match.

This means we can now upgrade manifest -> operator even if a custom CNI config name has been used.

pkg/render/node.go
```diff
@@ -1299,7 +1299,7 @@ func (c *nodeComponent) cniEnvvars() []corev1.EnvVar {
 	}
 
 	envVars := []corev1.EnvVar{
-		{Name: "CNI_CONF_NAME", Value: "10-calico.conflist"},
+		{Name: "CNI_CONF_NAME", Value: *c.cfg.Installation.CNI.ConfName},
```
caseydavenport, line 1302: Would like to see a unit test for this field

api/v1/installation_types.go
```diff
@@ -957,6 +957,14 @@ type CNISpec struct {
...
+	// Default: "10-calico.conflist"
+	// +optional
+	// +kubebuilder:validation:Type=string
+	ConfName *string `json:"confName,omitempty"`
```
caseydavenport, line 967: One thing we are going to have to solve is what to do if this value is changed to something with _less_ priority.

e.g., if you switch from `20-calico.conflist` to `30-calico.conflist`, the new configuration file will be ignored by the container runtime because the old configuration sorts higher.

We have some very rudimentary support for removing old config files here: https://github.com/projectcalico/calico/blob/master/cni-plugin/pkg/install/install.go#L413-L420

But it requires knowing the name of the file you want to remove!

caseydavenport (changes requested):
A few thoughts

PR #4418, opened 2026-02-10
Add tiered RBAC webhook

caseydavenport:
## Description

This PR deploys the new calico/webhooks deployment into the cluster, which is used for validating webhooks such as tier based RBAC: https://github.com/projectcalico/calico/pull/11803

- [ ] Tests
- [ ] Overrides support
- [ ] Other configuration options?

## Release Note

```release-note
TBD
```

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


pkg/controller/apiserver/apiserver_controller.go
```diff
@@ -467,18 +512,7 @@ func (r *ReconcileAPIServer) Reconcile(ctx context.Context, request reconcile.Re
 		r.status.SetDegraded(operatorv1.ResourceRenderingError, "Error rendering APIServer", err, reqLogger)
 		return reconcile.Result{}, err
 	}
-
-	components = append(components,
-		component,
-		rcertificatemanagement.CertificateManagement(&rcertificatemanagement.Config{
```
tmjd, line 473: I've got a little concern that this is being moved to before the component. There were no comments before about why this should come after the component so hopefully there wasn't a good reason for the order before.
What is the reason for moving it earlier?

Is there any reason that it needs to come after the APIServerPolicy is added?

pkg/render/webhooks/render.go
```diff
@@ -0,0 +1,449 @@
...
+			Name:      WebhooksName,
+			Namespace: common.CalicoNamespace,
+		},
+	}
+
+	// We need a network policy that selects the deployment to ensure bidirectional traffic is allowed between the API server and the webhook.
+	np := &v3.NetworkPolicy{
+		TypeMeta: metav1.TypeMeta{Kind: "NetworkPolicy", APIVersion: "projectcalico.org/v3"},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "allow-tigera." + WebhooksName, // TODO
```
tmjd: What is the TODO here?

pkg/render/webhooks/render.go
```diff
@@ -0,0 +1,449 @@
...
+			Namespace: common.CalicoNamespace,
+		},
+		Spec: v3.NetworkPolicySpec{
+			Tier:     "allow-tigera", // TODO:
```
tmjd: TODO

pkg/render/webhooks/render.go
```diff
@@ -0,0 +1,449 @@
...
+			Selector: fmt.Sprintf("k8s-app == '%s'", WebhooksName),
+			Ingress: []v3.Rule{
+				{
+					Action: v3.Allow,
+					// Protocol: &networkpolicy.TCPProtocol,
+					// Source:   networkpolicy.KubeAPIServerEntityRule,
```
tmjd: I think these should be added before this gets merged.

pkg/render/webhooks/render.go
```diff
@@ -0,0 +1,449 @@
...
+			Namespace: common.CalicoNamespace,
+		},
+		Spec: appsv1.DeploymentSpec{
+			Replicas: ptr.To[int32](1),
+			Strategy: appsv1.DeploymentStrategy{
+				Type: appsv1.RecreateDeploymentStrategyType,
+			},
+			Selector: &metav1.LabelSelector{
+				MatchLabels: map[string]string{
+					"k8s-app": WebhooksName,
```
tmjd: I don't think adding this label is needed, it should be getting added by our CreateOrUpdateOrDelete machinery.

pkg/render/webhooks/render.go
```diff
@@ -0,0 +1,449 @@
...
+	dep := &appsv1.Deployment{
+		TypeMeta: metav1.TypeMeta{Kind: "Deployment", APIVersion: "apps/v1"},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      WebhooksName,
+			Namespace: common.CalicoNamespace,
+		},
+		Spec: appsv1.DeploymentSpec{
+			Replicas: ptr.To[int32](1),
+			Strategy: appsv1.DeploymentStrategy{
+				Type: appsv1.RecreateDeploymentStrategyType,
```
tmjd: Do we need any special considerations since this is a webhook and a missing webhook handler can cause problems?

tmjd: If you've already consider those please include a comment on why RecreateDeployment is the appropriate strategy for this webhook pod.

pkg/render/webhooks/render.go
```diff
@@ -0,0 +1,449 @@
...
+			Selector: &metav1.LabelSelector{
+				MatchLabels: map[string]string{
+					"k8s-app": WebhooksName,
+				},
+			},
+			Template: corev1.PodTemplateSpec{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: WebhooksName,
+					Labels: map[string]string{
+						"k8s-app": WebhooksName,
```
tmjd: I also think this should be handled by our existing machinery.

pkg/render/webhooks/render.go
```diff
@@ -0,0 +1,449 @@
...
+				AdmissionReviewVersions: []string{"v1"},
+				SideEffects:             ptr.To(admissionregistrationv1.SideEffectClassNone),
+				TimeoutSeconds:          ptr.To[int32](5),
+				FailurePolicy:           ptr.To(admissionregistrationv1.Fail),
+				MatchPolicy:             ptr.To(admissionregistrationv1.Exact),
+			},
+		},
+	}
+
+	// Create a Cluster role and binding with access to all projectcalico.org resources.
```
tmjd: This comment is incorrect.

pkg/render/webhooks/render.go
```diff
@@ -0,0 +1,449 @@
...
+	calicoRules := []rbacv1.PolicyRule{
```
tmjd: I think you should just do `rules := ...` here and skip the additional assignment below. That will also better represent that we are going to use these rules for both Calico and Enterprise.

pkg/render/webhooks/render.go
```diff
@@ -0,0 +1,449 @@
...
+			},
+			Template: corev1.PodTemplateSpec{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: WebhooksName,
+					Labels: map[string]string{
+						"k8s-app": WebhooksName,
+					},
+				},
+				Spec: corev1.PodSpec{
+					HostNetwork:        render.HostNetworkRequired(c.cfg.Installation),
```
tmjd, line 167: Why shouldn't we honor the ForceHostNetwork configuration here?

pkg/controller/apiserver/apiserver_controller.go
caseydavenport: Yeah, I will add some commenting. There is an order of operations requirement here due to some dependencies.

- Webhook requires policy to work.
- Webhook requires TLS certs to be deployed
- Once the ValidatingWebhookConfig is created, policies require the webhook to be functional.

pkg/render/webhooks/render.go
caseydavenport: Whoops, forgot I left that in. I was waiting for changes from @radixo to go in to make this "calico-system", but I suppose I can do that now anyway.

pkg/render/webhooks/render.go
caseydavenport: I don't think we need any special considerations for the pod itself.

pkg/render/webhooks/render.go
caseydavenport: Hm, I''m not familiar with ForceHostNetwork and it looked like it was hard coded to `false` in the API server code, so I ignored it here.

pkg/render/webhooks/render.go
```diff
@@ -0,0 +1,452 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/utils/ptr"
+)
+
+const (
+	WebhooksTLSSecretName = "calico-webhooks-tls"
+
+	WebhooksName       = "calico-webhooks"
+	WebhooksPolicyName = "calico-system.calico-webhooks"
```
tmjd: I guess `calico-system` is the new tier name we're using. Is there a define for that?
It feels like bad practice if we're defining that everywhere.

tmjd (approved):

tmjd (approved):

PR #4419, opened 2026-02-10
Don't wait for API server if using v3 CRDs

caseydavenport:
## Description

We wait for the API server to be ready, but if we're using v3 CRDs there
will be no API server!

In addition, this starts to clean up how we handle options (storing the
whole opts instead of breaking down individual fields) in the affected
controllers.

## Release Note

```release-note
TBD
```

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/compliance/compliance_controller_test.go
+++ b/pkg/controller/compliance/compliance_controller_test.go
@@ -20,6 +20,7 @@ import (
 	"time"
 
 	"github.com/tigera/operator/pkg/controller/certificatemanager"
+	"github.com/tigera/operator/pkg/controller/options"
 	ctrlrfake "github.com/tigera/operator/pkg/ctrlruntime/client/fake"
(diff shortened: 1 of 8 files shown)


pkg/controller/utils/utils.go
```diff
@@ -309,7 +310,12 @@ func AddClusterWatch(c ctrlruntime.Controller, obj client.Object, h handler.Even
 	return AddNamespacedWatch(c, obj, h)
 }
 
-func IsAPIServerReady(client client.Client, l logr.Logger) bool {
+// IsProjectCalicoV3Available checks if projectcalico.org/v3 APIs are available. If the v3 parameter is true, it will skip the check and return true.
```
caseydavenport, line 313: For reviewers, this is really the only functional change in the PR (the rest is minor refactorings / renames)

tmjd (approved):

PR #4420, opened 2026-02-10
ev-6393

rene-dekker:
Openshift uses older CRDs than us. The Scheme field is incompatible. Luckily, there is an alternative way for setting the scheme as per their documentation:
```
  // Supported values are `HTTP` and `HTTPS`. You can also rewrite the
  // `__scheme__` label via relabeling configuration.
  // If empty, the value defaults to `HTTP`.
  // +kubebuilder:validation:Enum=http;https;HTTP;HTTPS
```

```release-note
Use backwards compatible schema configuration for prometheus endpoints on Openshift.
```

### DIFF
--- a/pkg/render/monitor/monitor.go
+++ b/pkg/render/monitor/monitor.go
@@ -607,7 +607,12 @@ func (mc *monitorComponent) prometheus() *monitoringv1.Prometheus {
 						Name:      CalicoNodeAlertmanager,
 						Namespace: &promNamespace,
 						Port:      intstr.FromString("web"),
-						Scheme:    ptr.To(monitoringv1.SchemeHTTP),
+						RelabelConfigs: []monitoringv1.RelabelConfig{
+							{
+								TargetLabel: "__scheme__",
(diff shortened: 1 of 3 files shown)


pkg/render/monitor/monitor.go
```diff
@@ -1473,10 +1513,18 @@ func (mc *monitorComponent) externalServiceMonitor() (client.Object, bool) {
 	var needsRBAC bool
 	endpoints := make([]monitoringv1.Endpoint, len(mc.cfg.Monitor.ExternalPrometheus.ServiceMonitor.Endpoints))
 	for i, ep := range mc.cfg.Monitor.ExternalPrometheus.ServiceMonitor.Endpoints {
+		relabelConfigs := ep.RelabelConfigs
+		if len(relabelConfigs) == 0 {
+			relabelConfigs = []monitoringv1.RelabelConfig{
```
hjiawei, line 1518: > ineffectual assignment to relabelConfigs (ineffassign)

PR #4427, opened 2026-02-11
Render resources specified in the EGW CR

sridhartigera:
## Description

Resource limits and requests specified in the EGW custom resource must be rendered in the EGW deployment.

## Release Note

```release-note
Fixed rendering resource limits and requests for Egress Gateway.
```

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/egressgateway/egressgateway.go
+++ b/pkg/render/egressgateway/egressgateway.go
@@ -155,6 +155,8 @@ func (c *component) egwDeployment() *appsv1.Deployment {
 		rcomp.ApplyDeploymentOverrides(&d, overrides)
 	}
 
+	// Add AWS specific resource requests/limits after applying overrides to avoid being overwritten
+	// if the user has specified their own.
 	c.addAWSResources(&d)
 	return &d
 }


pkg/render/egressgateway/egressgateway.go
```diff
@@ -414,15 +414,22 @@ func (c *component) getHTTPProbe() (string, string, string) {
 }
 
 func (c *component) getResources() corev1.ResourceRequirements {
+	resourceRequirements := corev1.ResourceRequirements{
```
caseydavenport: I don't think we should need to do this explicitly here. We have an "Overrides" mechanism that is meant to handle this automatically - https://github.com/tigera/operator/blob/c3cb855ddcafcb097b98c4edf2ac37b94b6221da/pkg/render/egressgateway/egressgateway.go#L154-L156

pkg/crds/operator/operator.tigera.io_egressgateways.yaml
```diff
@@ -1199,9 +1199,10 @@ spec:
                               name:
                                 description: |-
                                   Name is an enum which identifies the EGW Deployment container by name.
-                                  Supported values are: calico-egw
```
caseydavenport, line 1202: I think we should indicate that "egress-gateway" is the correct value and that "calico-egw" is deprecated and will be removed.

caseydavenport (changes requested):

pkg/render/egressgateway/egressgateway.go
```diff
@@ -155,6 +155,7 @@ func (c *component) egwDeployment() *appsv1.Deployment {
 		rcomp.ApplyDeploymentOverrides(&d, overrides)
 	}
 
+	c.addAWSResources(&d)
```
caseydavenport, line 160: Worth a comment explaining we do this here after applying overrides to ensure they don't get overwritten if a user provides their own values.

caseydavenport (approved):
One minor comment about a comment, but otherwise LGTM

PR #4438, opened 2026-02-13
Rename allow-tigera Tier to calico-system

radixo:
## Description

This PR renames allow-tigera Tier to calico-system with the purpose to enable this Tier on Calico OSS.
All the files were changed only applying regex replaces on a few allow-tigera patterns.
Specially for the tigera-operator a cleanup routine was appended to tier's controller and render.

## Release Note

```release-note
The Tier allow-tigera has been renamed to calico-system.

If your Calico installation does not use the Tigera Operator, or if you have created custom Network Policies within this Tier, you must manually update your resources to reference the new Tier name.

Please review and adjust any affected policies to ensure continued correct behavior.
```

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/dex.go
+++ b/pkg/render/dex.go
@@ -137,7 +137,7 @@ func (c *dexComponent) Objects() ([]client.Object, []client.Object) {
 		&corev1.Secret{ObjectMeta: metav1.ObjectMeta{Namespace: DexObjectName, Name: DexObjectName}},
 		&corev1.Secret{ObjectMeta: metav1.ObjectMeta{Namespace: common.OperatorNamespace(), Name: DexObjectName}},
 		// allow-tigera Tier was renamed to calico-system
-		networkpolicy.DeprecatedAllowTigeraNetworkPolicyObject("calico-system-dex", DexNamespace),
+		networkpolicy.DeprecatedAllowTigeraNetworkPolicyObject("allow-tigera-dex", DexNamespace),
 		networkpolicy.DeprecatedAllowTigeraNetworkPolicyObject("default-deny", DexNamespace),
 	}
 


pkg/controller/tiers/tiers_controller.go
```diff
@@ -232,3 +238,33 @@ func (r *ReconcileTiers) prepareTiersConfig(ctx context.Context, reqLogger logr.
 
 	return &tiersConfig, nil
 }
+
+func (r *ReconcileTiers) fetchDeprecatedObjs(ctx context.Context, tiersConfig *tiers.Config) (err error) {
```
caseydavenport: Generally easier to read if the code doesn't use side-effects like this and just returns the objects instead of modifying the input.

That said, we should know exactly which objects we need to delete without querying the cluster, right? 

What happens if a user has created their own policy within the allow-tigera tier? What will our code do? What _should_ our code do?

I think it probably needs to spot that and abort, warning the user? WDYT?

pkg/render/tiers/tiers.go
```diff
@@ -62,18 +64,24 @@ func (t tiersComponent) ResolveImages(is *operatorv1.ImageSet) error {
...
 	if len(t.cfg.DNSEgressCIDRs.IPV4) > 0 || len(t.cfg.DNSEgressCIDRs.IPV6) > 0 {
-		objsToCreate = append(objsToCreate, t.allowTigeraNodeLocalDNSPolicy())
+		objsToCreate = append(objsToCreate, t.calicoSystemNodeLocalDNSPolicy())
 	} else {
-		objsToDelete = append(objsToDelete, t.allowTigeraNodeLocalDNSPolicy())
+		objsToDelete = append(objsToDelete, t.calicoSystemNodeLocalDNSPolicy())
 	}
 
+	// Deprecate allow-tigera Tier
+	allowTigeraClusterDNSPolicy := t.calicoSystemClusterDNSPolicy()
```
caseydavenport: Why do we need this explicitly here? Shouldn't it already be part of `DeprecatedObjs`?

pkg/controller/tiers/tiers_controller.go
radixo: Totally, I gave each render the responsibility to remove what they created.
If the user tries to create we will have the error on tiers render, trying to delete the old Tier and not being able because it is not empty.

pkg/controller/apiserver/apiserver_controller.go
```diff
@@ -507,7 +507,13 @@ func (r *ReconcileAPIServer) Reconcile(ctx context.Context, request reconcile.Re
 	// deployment becomes unhealthy and reconciliation of non-NetworkPolicy resources in the apiserver controller
 	// would resolve it, we render the network policies of components last to prevent a chicken-and-egg scenario.
 	if !r.opts.UseV3CRDs && includeV3NetworkPolicy {
-		components = append(components, render.APIServerPolicy(&apiServerCfg))
+		components = append(components,
+			render.APIServerPolicy(&apiServerCfg),
```
caseydavenport: Sorry I didn't spot this before, but I think the `render.APIServerPolicy` component can / should just return the old NP as part of its `objectsToDelete` instead of defining a deletion passthru component just for this.

pkg/controller/clusterconnection/clusterconnection_controller.go
```diff
@@ -485,7 +485,14 @@ func (r *ReconcileConnection) Reconcile(ctx context.Context, request reconcile.R
 		if err != nil {
 			log.Error(err, "Failed to create NetworkPolicy component for Guardian, policy will be omitted")
 		} else {
-			components = append(components, policyComponent)
+			components = append(components,
+				policyComponent,
+				// allow-tigera Tier was renamed to calico-system
+				render.NewDeletionPassthrough(
```
caseydavenport: Same here and throughout the PR - we should just use the native `objectsTodelete` support from the policy component

pkg/render/common/networkpolicy/networkpolicy.go
```diff
@@ -318,6 +319,21 @@ func (h *NetworkPolicyHelper) IntrusionDetectionSourceEntityRule() v3.EntityRule
 	return CreateSourceEntityRule(h.namespace("tigera-intrusion-detection"), "intrusion-detection-controller")
 }
 
+// DeprecatedAllowTigeraNetworkPolicyObject returns a CNP object with the
+// allow-tigera tier on the name as helper for deprecating this old Tier.
+func DeprecatedAllowTigeraNetworkPolicyObject(name, namespace string) client.Object {
```
caseydavenport, line 324: Once we move within the component, these functions need not be exposed publicly.

pkg/controller/clusterconnection/clusterconnection_controller.go
radixo: No worries... I will move them to the renders.

pkg/render/common/networkpolicy/networkpolicy.go
radixo: Indeed

radixo: Oh sorry, it is being consumed from pkg/render and some other subcomponents. It needs to be public. Or I misunderstood what you pointed out.

pkg/render/dex.go
```diff
@@ -136,6 +136,9 @@ func (c *dexComponent) Objects() ([]client.Object, []client.Object) {
 		// Delete the secret called tigera-dex which in the past was used to store a client secret.
 		&corev1.Secret{ObjectMeta: metav1.ObjectMeta{Namespace: DexObjectName, Name: DexObjectName}},
 		&corev1.Secret{ObjectMeta: metav1.ObjectMeta{Namespace: common.OperatorNamespace(), Name: DexObjectName}},
+		// allow-tigera Tier was renamed to calico-system
+		networkpolicy.DeprecatedAllowTigeraNetworkPolicyObject("calico-system-dex", DexNamespace),
```
caseydavenport: Looks like a bug in the name here - should be `allow-tigera-dex`

pkg/render/manager.go
```diff
@@ -273,14 +273,20 @@ func (c *managerComponent) Objects() ([]client.Object, []client.Object) {
...
+		c.managerCalicoSystemNetworkPolicy(),
+		networkpolicy.CalicoSystemDefaultDeny(c.cfg.Namespace),
 		managerServiceAccount(c.cfg.Namespace),
 	)
 	objsToCreate = append(objsToCreate, c.getTLSObjects()...)
 	objsToCreate = append(objsToCreate, c.managerService())
 	objsToCreate = append(objsToCreate, c.managerExternalNameService())
 
+	// allow-tigera Tier was renamed to calico-system
+	objsToDelete = append(objsToDelete,
```
caseydavenport: This looks like a duplicate with `deprecatedResources` - we can remove this I think?

pkg/render/policyrecommendation.go
```diff
@@ -424,8 +424,8 @@ func (pr *policyRecommendationComponent) serviceAccount() client.Object {
 	}
 }
 
-// allowTigeraPolicyForPolicyRecommendation defines an allow-tigera policy for policy recommendation.
-func (pr *policyRecommendationComponent) allowTigeraPolicyForPolicyRecommendation() *v3.NetworkPolicy {
+// calicoSystemPolicyForPolicyRecommendation defines an calico-system policy for policy recommendation.
```
caseydavenport:
```suggestion
// calicoSystemPolicyForPolicyRecommendation defines a calico-system policy for policy recommendation.
```

pkg/render/typha.go
```diff
@@ -137,7 +137,7 @@ func (c *typhaComponent) Objects() ([]client.Object, []client.Object) {
 }
 
 func NewTyphaNonClusterHostPolicy(cfg *TyphaConfiguration) Component {
-	return NewPassthrough(typhaNonClusterHostAllowTigeraPolicy(cfg))
+	return NewPassthrough(typhaNonClusterHostCalicoSystemPolicy(cfg))
```
caseydavenport: Will this not clean up the old typha policies?

pkg/render/packet_capture_api.go
```diff
@@ -83,8 +83,23 @@ func PacketCaptureAPI(cfg *PacketCaptureApiConfiguration) Component {
 	}
 }
 
+type packetCaptureAPIPolicyComponent struct {
+	PassthroughComponent
```
caseydavenport: This looks good, although I think I would have just introduced a new constructor that allows passing both objects to create and to delete in a single passthrough component.

pkg/render/dex.go
radixo: Good catch... thank you!

pkg/render/manager.go
radixo: Wow... thanks again.

pkg/render/typha.go
radixo: Good catch!

pkg/render/packet_capture_api.go
radixo: Makes sense, I will improve it.

pkg/render/dex.go
radixo: Oh no... actually the old name was `calico-system-dex` and after Claude sugestion, to not be repeated. I renamed it yesterday to `dex` only. So `calico-system-dex` was right. Fixing. Thanks FVs :raised_hands:

PR #4457, opened 2026-02-19
Import MutatingAdmissionPolicies from Calico

caseydavenport:
## Description

When using v3 CRDs, we need to manage MutatingAdmissionPolicies. These files live in the Calico repository alongside CRDs, so we should import them in much the same way.

This refactors the codebase to include a new `pkg/imports/crds` and
`pkg/imports/admission` instead of a singular `pkg/crds`, and adds the
necessary logic to the operator to manage these new objects.

## Release Note

```release-note
TBD
```

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/imports/crds/calico/v1.crd.projectcalico.org/crd.projectcalico.org_bgpconfigurations.yaml
+++ b/pkg/imports/crds/calico/v1.crd.projectcalico.org/crd.projectcalico.org_bgpconfigurations.yaml
@@ -97,6 +97,11 @@ spec:
                     x-kubernetes-map-type: atomic
                   type: array
                   x-kubernetes-list-type: set
+                programClusterRoutes:
+                  enum:
+                    - Enabled
+                    - Disabled
+                  type: string
                 serviceClusterIPs:
                   items:
                     properties:
--- a/pkg/imports/crds/calico/v3.projectcalico.org/projectcalico.org_bgpconfigurations.yaml
+++ b/pkg/imports/crds/calico/v3.projectcalico.org/projectcalico.org_bgpconfigurations.yaml
@@ -100,6 +100,11 @@ spec:
                     x-kubernetes-map-type: atomic
                   type: array
                   x-kubernetes-list-type: set
+                programClusterRoutes:
+                  enum:
+                    - Enabled
+                    - Disabled
+                  type: string
                 serviceClusterIPs:
                   items:
                     properties:


cmd/main.go
```diff
@@ -295,6 +296,11 @@ If a value other than 'all' is specified, the first CRD with a prefix of the spe
 			os.Exit(1)
 		}
 
+		if err := admission.Ensure(mgr.GetClient(), variant, v3CRDs, setupLog); err != nil {
```
tmjd, line 299: Wouldn't it be more obvious if we didn't pass the v3CRDs into this function and the one above and just call the appropriate one depending on its value?

pkg/common/kubernetes_version.go
```diff
@@ -59,3 +59,9 @@ func (v *VersionInfo) ProvidesCertV1API() bool {
 	}
 	return false
 }
+
+// ProvidesMutatingAdmissionPolicyV1Beta1 returns if admissionregistration.k8s.io/v1beta1 MutatingAdmissionPolicy
```
tmjd, line 63: Should there be a call to this in main also?

cmd/main.go
tmjd: But maybe not including that condition at this level pushes the context down a level, maybe that is better?

pkg/imports/admission/admission.go
```diff
@@ -0,0 +1,163 @@
...
+		return nil
+	}
+
+	objs := GetMutatingAdmissionPolicies(opv1.ProductVariant(variant), v3)
+
+	for _, obj := range objs {
+		log.Info("ensuring MutatingAdmissionPolicy resource exists", "name", obj.GetName(), "kind", obj.GetObjectKind().GroupVersionKind().Kind)
+		ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+		if err := c.Create(ctx, obj); err != nil {
+			cancel()
```
tmjd, line 122: AI suggested to do a `defer cancel()` immediately after the context.WithTimeout as a more idiomatic Go pattern.

cmd/main.go
caseydavenport: Yeah, I had the same back and forth. I think I like it better pushing the context inside the Ensure function to keep that complexity out of main?

pkg/common/kubernetes_version.go
caseydavenport: Yeah, perhaps - right now, if the cluster doesn't support it, the operator will skip MAPs and then get to the core controller where it will set degraded.

If we did this check in main, what would the desired behavior be in the case of failure? Just logging an error? Or exiting?

tmjd: I'm fine with logging an error. I was thinking it might cause problems if the operator was configured to use MAPs but they weren't supported.

pkg/controller/installation/core_controller.go
```diff
@@ -2131,6 +2139,64 @@ func (r *ReconcileInstallation) updateCRDs(ctx context.Context, variant operator
...
+		return err
+	}
+	for i := range existingMAPBs.Items {
+		if !desiredMAPBs[existingMAPBs.Items[i].Name] {
+			toDelete = append(toDelete, &existingMAPBs.Items[i])
+		}
+	}
+
+	// Create or update desired MAPs/MAPBs and delete any stale ones in a single pass.
+	handler := r.newComponentHandler(log, r.client, r.scheme, nil)
```
tmjd: Should we put the OwnerRefernce on these?
If not, then I think you should include a comment in the code saying why.

tmjd: I think you suggested it is ok and we should add the OwnerReference so I think that means the nil argument here needs to be the controlling CR. Which I think we said would be the APIServer CR, right? Because we put the deployment overrides for it under the APIServer CR, IIRC.

PR #4473, opened 2026-02-25
feat(release): Enhance build and publish workflows

radTuti:
## Description

This adds the ability to extend the build and publish workflows:

- switch `setupHashreleaseBuild` and `publishImages` to wrapable function vars and add `isValidReleaseVersion` to allow overriding
- add `buildAfter` to allow multiple cleanups added at various points that get done regardless of build status.
- the builds for hashrelease uses the git hash for the corresponding product to generate CRDs when a path to the product directory locally is not set. 
- for release, the publish creates a github release by default unless user overrides it. Github release is still created in draft mode by default.

## Release Note

```release-note
TBD
```

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -518,9 +518,9 @@ format-check:
 dirty-check:
 	@if [ "$$(git diff --stat)" != "" ]; then \
 	echo "The following files are dirty"; git diff --stat; exit 1; fi
-	@# Check that no new CRDs needed to be committed
-	@if [ "$$(git status --porcelain pkg/imports/crds)" != "" ]; then \
-	echo "The following CRD files need to be added"; git status --porcelain pkg/imports/crds; exit 1; fi
+	@# Check that no new CRDs or admission policies needed to be committed
+	@if [ "$$(git status --porcelain pkg/imports)" != "" ]; then \
+	echo "The following imported files need to be added"; git status --porcelain pkg/imports; exit 1; fi
 
 foss-checks:
 	@echo Running $@...
@@ -680,9 +680,11 @@ $(BINDIR)/gen-versions: $(shell find ./hack/gen-versions -type f)
 define prep_local_crds
     $(eval product := $(1))
 	rm -rf pkg/imports/crds/$(product)
+	rm -rf pkg/imports/admission/$(product)
 	rm -rf .crds/$(product)
 	mkdir -p pkg/imports/crds/$(product)/v1.crd.projectcalico.org/
 	mkdir -p pkg/imports/crds/$(product)/v3.projectcalico.org/
+	mkdir -p pkg/imports/admission/$(product)
 	mkdir -p .crds/$(product)
 endef
 
@@ -711,6 +713,11 @@ define copy_eck_crds
 		$(eval product := $(2))
 	@cp $(dir)/charts/crd.projectcalico.org.v1/templates/eck/* pkg/imports/crds/$(product)/ && echo "Copied $(product) ECK CRDs"
 endef
+define copy_admission_policies
+    $(eval dir := $(1))
+		$(eval product := $(2))
+	@cp $(dir)/api/admission/* pkg/imports/admission/$(product)/ && echo "Copied $(product) admission policies"
+endef
 
 .PHONY: read-libcalico-version read-libcalico-enterprise-version
 .PHONY: update-calico-crds update-enterprise-crds
@@ -729,6 +736,7 @@ read-libcalico-calico-version:
 update-calico-crds: fetch-calico-crds
 	$(call copy_v1_crds, $(CALICO_CRDS_DIR),"calico")
 	$(call copy_v3_crds, $(CALICO_CRDS_DIR),"calico")
+	$(call copy_admission_policies, $(CALICO_CRDS_DIR),"calico")
 
 prepare-for-calico-crds:
 	$(call prep_local_crds,"calico")
@@ -749,6 +757,7 @@ update-enterprise-crds: fetch-enterprise-crds
 	$(call copy_v1_crds,$(ENTERPRISE_CRDS_DIR),"enterprise")
 	$(call copy_v3_crds, $(ENTERPRISE_CRDS_DIR),"enterprise")
 	$(call copy_eck_crds,$(ENTERPRISE_CRDS_DIR),"enterprise")
+	$(call copy_admission_policies,$(ENTERPRISE_CRDS_DIR),"enterprise")
 
 prepare-for-enterprise-crds:
 	$(call prep_local_crds,"enterprise")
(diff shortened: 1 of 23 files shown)


hack/release/build.go
```diff
@@ -257,45 +321,55 @@ func assertOperatorImageVersion(registry, image, expectedVersion string) error {
...
+// and adds a build after hook to reset any changes to git state after the build completes.
+func setupHashreleaseBuild(ctx context.Context, c *cli.Command, repoRootDir string) error {
+	buildAfterHooks.Add("Reset repo git state", func(ctx context.Context, c *cli.Command) error {
+		if out, err := gitInDirContext(ctx, repoRootDir, append([]string{"checkout", "-f"}, changedFiles...)...); err != nil {
+			logrus.Error(out)
+			return fmt.Errorf("resetting git state in repo after hashrelease build: %w", err)
 		}
+		return nil
+	})
+
```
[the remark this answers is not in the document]
radTuti: it is acceptable.

hack/release/flags.go
```diff
@@ -73,7 +80,7 @@ var (
 		Category: githubFlagCategory,
 		Usage:    "Create a GitHub release",
 		Sources:  cli.EnvVars("CREATE_GITHUB_RELEASE"),
-		Value:    false,
+		Value:    true,
```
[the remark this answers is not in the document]
radTuti: not necessary

hack/release/publish.go
```diff
@@ -71,12 +89,7 @@ var publishBefore = cli.BeforeFunc(func(ctx context.Context, c *cli.Command) (co
 		return ctx, nil
 	}
 
-	// If building a hashrelease, publishGithubRelease must be false
-	if c.Bool(hashreleaseFlag.Name) && c.Bool(createGithubReleaseFlag.Name) {
-		return ctx, fmt.Errorf("cannot publish GitHub release for hashrelease builds")
-	}
-
-	if !c.Bool(createGithubReleaseFlag.Name) {
+	if c.Bool(hashreleaseFlag.Name) || !c.Bool(createGithubReleaseFlag.Name) {
```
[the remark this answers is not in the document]
radTuti: this is expected

Makefile
```diff
@@ -559,7 +559,7 @@ endif
 ## Create a release for the specified RELEASE_TAG.
 release-tag: var-require-all-RELEASE_TAG-GITHUB_TOKEN
 	$(MAKE) release VERSION=$(RELEASE_TAG)
-	REPO=$(REPO) CREATE_GITHUB_RELEASE=true $(MAKE) release-publish VERSION=$(RELEASE_TAG)
+	REPO=$(REPO) $(MAKE) release-publish VERSION=$(RELEASE_TAG)
```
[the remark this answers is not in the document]
radTuti: expected

hack/release/build.go
```diff
@@ -78,22 +80,50 @@ var buildCommand = &cli.Command{
...
+// It can be set via init() in separate files to extend the build command behavior.
+var buildBeforeHook cliBeforeHookFunc
+
+// setupHashreleasePreHook is an optional hook called at the start of setupHashreleaseBuild
+// for additional hashrelease setup (e.g., copying image config files).
+var setupHashreleasePreHook cliHookWithRepoDirFunc
+
+// buildAfterHook is an optional hook called in buildAfter for additional post-build tasks.
+// This runs regardless of build success or failure, so it can be used for cleanup tasks.
+var buildAfterHooks multiHook
```
danudey: Should this be a list of hooks for each stage, in case we need to add multiple steps at some point? Can it be?

hack/release/build.go
```diff
@@ -324,32 +399,161 @@ func setupHashreleaseBuild(ctx context.Context, c *cli.Command, repoRootDir stri
...
+		logrus.WithField("depth", depth).Infof("Deepening git fetch for %s repo", r.Product)
+		if _, err := git("-C", repoTmpDir, "fetch", "--deepen", strconv.Itoa(deepen), "origin", r.branch); err != nil {
+			return "", fmt.Errorf("deepening fetch for branch %s in %s repo: %w", r.branch, r.Product, err)
+		}
+	}
+
+	// Checkout the specific commit
+	if _, err := git("-C", repoTmpDir, "checkout", gitHash); err != nil {
+		return "", fmt.Errorf("checking out git hash %s in %s repo: %w", gitHash, r.Product, err)
+	}
```
danudey: Is this premature optimization? A full clone of the operator repo takes 4.8 seconds on my laptop.

If this is for cloning other repos as well, then I think the biggest repo that I've tested that Operator might want to clone only takes 8 seconds if cloned with `git clone --filter=blob:none`, which might clean up a lot of this code.


```suggestion
	// Clone repo without fetching blobs; don't check out the code yet
	if _, err := git( "clone", "-b", r,branch, fmt.Sprintf("git@github.com:%s.git", r.repo), repoTmpDir); err != nil {
		return "", fmt.Errorf("cloning git repo into %s temp dir: %w", r.Product, err)
	}
	// Detached checkout of the commit we want; this will automatically fetch whatever blobs we need
	if _, err := git("-C", repoTmpDir, "switch "--detach", gitHash); err != nil {
		return "", fmt.Errorf("switching to detached commit %s: %w", commit, r.Product, err)
	}
```

hack/release/utils.go
```diff
@@ -132,6 +133,48 @@ func runCommandInDir(dir, name string, args, env []string) (string, error) {
 	return strings.TrimSpace(outb.String()), err
 }
 
+// runCommandInDirContext is a context-aware variant of runCommandInDir.
+// When ctx is cancelled, the entire process group is killed to ensure child processes are cleaned up.
+func runCommandInDirContext(ctx context.Context, dir, name string, args, env []string) (string, error) {
```
danudey: We're starting to have so many `runCommandX` type functions that I wonder if we should just have one command that we pass `Option()`s to, like

```go
runCommand(
    cmd.WithCommand("blah"),
    cmd.InDir("dir"),
    cmd.WithContext(ctx),
    cmd.WithEnv(env)
)
```

Otherwise, it might make sense to have one `runCommandInDirWithContextAndOutputAndTimeout(...)` command and some simplified wrappers that just pass defaults for whichever fields they don't need (e.g. `runCommandInDir()` passes a default ctx and timeout and discards output).

Probably not worth it for this PR but I'm not sure how difficult that would be to implement.

hack/release/build.go
danudey: Note: I timed it locally; this takes about 8 seconds on my laptop for our largest repo, compared to 43 seconds to clone the entire repo.

hack/release/build.go
radTuti: I do not think it really works for the hooks in Before but I will switch them all to lists for consistency

hack/release/utils.go
radTuti: Likely for a different PR to limit the scope for this PR.

danudey (approved):

PR #4474, opened 2026-02-25
[CORE-12314] Use two ServiceEndpoint instances for host-network and pod-network

coutinhop:
## Description

Use two separate ServiceEndpoint instances (Endpoint and PodNetworkEndpoint)
instead of a single struct, to distinguish between host-network and pod-network
K8s API server endpoints. Simplify EnvVars() by removing hostNetworked/provider
params — callers pick the appropriate instance. Remove Docker EE proxy.local
special-case, which is now handled implicitly.

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

## Release Note

```release-note
Split kubernetes-services-endpoint configmap into KUBERNETES_SERVICE_HOST/PORT for host-networked pods (previous behaviour) and KUBERNETES_SERVICE_HOST_POD_NETWORK and KUBERNETES_SERVICE_PORT_POD_NETWORK for pod-networked pods.
```

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


pkg/controller/k8sapi/k8s-endpoint.go
```diff
@@ -26,56 +26,60 @@ import (
...
-		Port: os.Getenv("KUBERNETES_SERVICE_PORT"),
+		HostNetworkHost: os.Getenv("KUBERNETES_SERVICE_HOST"),
+		HostNetworkPort: os.Getenv("KUBERNETES_SERVICE_PORT"),
 	}
 }
 
 // ServiceEndpoint is the Host/Port of the K8s endpoint.
+// HostNetworkHost/HostNetworkPort are used for host-networked pods, while
+// PodNetworkHost/PodNetworkPort are used for pod-networked pods.
 type ServiceEndpoint struct {
```
caseydavenport, line 46: Does it make sense to extend this struct? Or just to use two instances?

```
ep := ServiceEndpoint{}
hostNetEP := ServiceEndpoint{}
```

^ This feels a bit conceptually cleaner to me? I think this will make the code a lot clearer.

pkg/controller/k8sapi/k8s-endpoint.go
```diff
@@ -26,56 +26,60 @@ import (
...
+		if k8s.PodNetworkHost == "" || k8s.PodNetworkPort == "" {
+			return nil
+		}
+		return []v1.EnvVar{
+			{Name: "KUBERNETES_SERVICE_HOST", Value: k8s.PodNetworkHost},
+			{Name: "KUBERNETES_SERVICE_PORT", Value: k8s.PodNetworkPort},
+		}
 	}
 
-	if provider == operator.ProviderDockerEE && !hostNetworked && k8s.Host == dockerEEProxyLocal {
```
caseydavenport, line 57: Is this change intentional? Would this be a breaking change for MKE?

pkg/controller/utils/utils.go
```diff
@@ -449,8 +449,10 @@ func PopulateK8sServiceEndPoint(client client.Client) error {
 			return fmt.Errorf("failed to read ConfigMap %q: %s", render.K8sSvcEndpointConfigMapName, err)
 		}
 	} else {
-		k8sapi.Endpoint.Host = cm.Data["KUBERNETES_SERVICE_HOST"]
-		k8sapi.Endpoint.Port = cm.Data["KUBERNETES_SERVICE_PORT"]
+		k8sapi.Endpoint.HostNetworkHost = cm.Data["KUBERNETES_SERVICE_HOST"]
+		k8sapi.Endpoint.HostNetworkPort = cm.Data["KUBERNETES_SERVICE_PORT"]
+		k8sapi.Endpoint.PodNetworkHost = cm.Data["KUBERNETES_SERVICE_HOST_POD_NETWORK"]
+		k8sapi.Endpoint.PodNetworkPort = cm.Data["KUBERNETES_SERVICE_PORT_POD_NETWORK"]
```
caseydavenport: Hm, I think this is a breaking change - today, both pods share the same configuraiton. 

I think we want to keep that behavior if the `POD_NETWORK` fields aren't present (i.e., use the host network ones). 

At a minimum, we need a release note. Ideally we wouldn't break anyone on upgrade, though!

pkg/controller/k8sapi/k8s-endpoint.go
coutinhop: it was intentional, my reasoning was that if we are making a change for `!hostNetworked` then this condition would be redundant... but if we do change the default behaviour in this PR then this might need to be restored...

pkg/controller/utils/utils.go
coutinhop: In theory, it shouldn't break things, since the pod-networked pods should work fine with the api server service, but by adding the `*_POD_NETWORK` fields to the configmap, we have a quick workaround in case things don't work so well in practice. Would that be OK?

PR #4480, opened 2026-02-26
fix(tier cleanup): Deprecated tier cannot be deleted

rene-dekker:
The manager controller did not clean up deprecated policies, due to trying to delete them from the wrong namespace. As a result, the old tier (allow-tigera) can never be deleted.


pkg/render/manager.go
radixo, line 1494: Should we keep this one? Did I do right using `installNS`?

rene-dekker: Due to manager being moved to another namespace in the previous release, there can be a network policy for 2 namespaces in the old tier.

PR #4482, opened 2026-02-27
Handle enterprise license expiry with grace period

hjiawei:
## Description

When an enterprise license expires (after the grace period elapses),
disable logging (fluentd DaemonSet) and metrics scraping (ServiceMonitors)
while keeping the dataplane running. During the grace period, all
components continue running normally with only a log warning. This aligns
with the enterprise changes that relax license validation so expired
licenses are accepted, with individual components handling degraded behavior.

Changes:
- Add license utility functions in pkg/controller/utils/license.go:
  - ParseGracePeriod: parses the "90d" format grace period string from
    the typed LicenseKey status.gracePeriod field
  - GetLicenseStatus: point-in-time check using a single time.Now() call
    returning Valid, InGracePeriod, or Expired status
  - Move existing license helpers (FetchLicenseKey, IsFeatureActive,
    WaitToAddLicenseKeyWatch) out of utils.go into license.go

- Monitor controller:
  - Add license watch and fetch with grace period support
  - When expired: set degraded status, move ServiceMonitors to toDelete
  - When in grace period: log warning, keep everything running normally

- LogCollector controller:
  - Add grace period support to license expiry check
  - When expired: set degraded status, move fluentd DaemonSet to toDelete
  - When in grace period: log warning, keep everything running normally

- Render layer:
  - Add LicenseExpired flag to monitor Config and FluentdConfiguration
  - Conditionally move ServiceMonitors/DaemonSet to toDelete when expired

- Bump tigera/api to v0.0.0-20260227222130-df0b9e289a34 which includes
  the GracePeriod field in LicenseKeyStatus

## Release Note

```release-note
Operator now disables log forwarding and metrics scraping on enterprise license expiry while keeping the dataplane running, and reports license status in TigeraStatus.
```

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/utils/license.go
+++ b/pkg/controller/utils/license.go
@@ -70,13 +70,18 @@ func IsFeatureActive(license v3.LicenseKey, featureName string) bool {
 	return false
 }
 
-// ParseGracePeriod parses a grace period string (e.g. "90d") and returns the
-// corresponding duration. Returns 0 if the string is empty or cannot be parsed.
+// ParseGracePeriod parses a grace period string in the format "Nd" (e.g. "90d")
+// where N is a non-negative number of days. This is the format produced by the
+// license controller. Returns 0 if the string is empty, cannot be parsed, or
+// represents a negative duration.
 func ParseGracePeriod(gracePeriod string) time.Duration {
 	if gracePeriod == "" {
 		return 0
 	}
-	s := strings.TrimSuffix(gracePeriod, "d")
+	s, ok := strings.CutSuffix(gracePeriod, "d")
+	if !ok {
+		return 0
+	}
 	days, err := strconv.Atoi(s)
 	if err != nil || days < 0 {
 		return 0
--- a/pkg/controller/utils/license_test.go
+++ b/pkg/controller/utils/license_test.go
@@ -45,7 +45,7 @@ var _ = Describe("License helpers", func() {
 			Entry("empty string", "", time.Duration(0)),
 			Entry("invalid string", "abc", time.Duration(0)),
 			Entry("negative days", "-5d", time.Duration(0)),
-			Entry("number without d suffix", "90", 90*24*time.Hour),
+			Entry("bare number without unit", "90", time.Duration(0)),
 		)
 	})
 


pkg/controller/logcollector/logcollector_controller_test.go
```diff
@@ -844,4 +845,82 @@ var _ = Describe("LogCollector controller tests", func() {
...
+					Features:    []string{"export-logs"},
+				},
+			})).NotTo(HaveOccurred())
+
+			result, err := r.Reconcile(ctx, reconcile.Request{})
+			Expect(err).ShouldNot(HaveOccurred())
+
+			// Should requeue to re-reconcile when the grace period expires.
+			Expect(result.RequeueAfter).To(BeNumerically(">", 0))
+			Expect(result.RequeueAfter).To(BeNumerically("~", 89*24*time.Hour, 1*time.Hour))
```
rene-dekker, line 913: TIL "~" comparator can be used here.

pkg/controller/utils/license.go
```diff
@@ -0,0 +1,102 @@
...
+	return false
+}
+
+// ParseGracePeriod parses a grace period string (e.g. "90d") and returns the
+// corresponding duration. Returns 0 if the string is empty or cannot be parsed.
+func ParseGracePeriod(gracePeriod string) time.Duration {
+	if gracePeriod == "" {
+		return 0
+	}
+	s := strings.TrimSuffix(gracePeriod, "d")
```
rene-dekker: This won't be able to deal with durations that don't use d as a suffix, I think it is better to solve for that as well

pkg/controller/utils/license_test.go
```diff
@@ -0,0 +1,81 @@
...
+			func(input string, expected time.Duration) {
+				Expect(ParseGracePeriod(input)).To(Equal(expected))
+			},
+			Entry("90 days", "90d", 90*24*time.Hour),
+			Entry("30 days", "30d", 30*24*time.Hour),
+			Entry("0 days", "0d", time.Duration(0)),
+			Entry("empty string", "", time.Duration(0)),
+			Entry("invalid string", "abc", time.Duration(0)),
+			Entry("negative days", "-5d", time.Duration(0)),
+			Entry("number without d suffix", "90", 90*24*time.Hour),
```
rene-dekker: All of these seem invalid to me.

pkg/controller/utils/license.go
hjiawei: Bare number without unit `d` is now treated as invalid grace period.

pkg/controller/utils/license_test.go
hjiawei: This test case has been updated to require the `d` unit suffix and no longer accepts bare numbers without a unit.

PR #4483, opened 2026-02-27
EV-6336: feat(istio): waypoint pull secret support for private registries

electricjesus:
## Description

**Jira:** [EV-6336](https://tigera.atlassian.net/browse/EV-6336)

When users create an Istio waypoint Gateway (`gatewayClassName: istio-waypoint`) in their namespace, istiod automatically creates a waypoint Deployment there. On clusters with private registries (like AKS with `gcr.io/unique-caldron-775`), the waypoint pod fails with `ImagePullBackOff` because the pull secret only exists in the operator namespace and istiod doesn't inject `imagePullSecrets` into waypoint pod specs.

This PR fixes the issue with two changes:

**Part 1: Pass imagePullSecrets to istiod via Helm values**
- Adds `ImagePullSecrets` field to `GlobalConfig` in the Istio render package
- Populates `global.imagePullSecrets` in istiod Helm values from Installation pull secrets
- This makes istiod inject `imagePullSecrets` references into waypoint pod specs it creates

**Part 2: New waypoint sub-controller**
- Watches for `istio-waypoint` Gateway resources across all namespaces
- Copies pull Secret objects from the operator namespace to waypoint namespaces so the references resolve
- Tracks copied secrets with `operator.tigera.io/istio-waypoint-pull-secret` label for cleanup
- Cleans up stale secrets when gateways are removed or Istio CR is deleted

Both parts are no-ops when no pull secrets are configured (`omitempty` / early return).

**Testing:**
- Unit tests for Helm values serialization (render package)
- 8 controller tests covering: no secrets, single/multiple gateways, same/different namespaces, gateway deletion cleanup, non-matching gatewayClassName, Istio CR deletion, and missing Installation
- All existing istio controller and render tests continue to pass

**Components affected:** `pkg/render/istio`, `pkg/controller/istio`, `internal/controller`

## Release Note

```release-note
Istio waypoint Gateways now work on clusters with private registries. The operator passes imagePullSecrets to istiod via Helm values and copies pull secrets to namespaces containing waypoint Gateways.
```

## For PR author

- [x] Tests for change.

[EV-6336]: https://tigera.atlassian.net/browse/EV-6336?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/controller/istio/waypoint/waypoint_secrets_controller.go
+++ b/pkg/controller/istio/waypoint/waypoint_secrets_controller.go
@@ -21,6 +21,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/event"
@@ -32,6 +33,7 @@ import (
 	gapi "sigs.k8s.io/gateway-api/apis/v1"
 
 	operatorv1 "github.com/tigera/operator/api/v1"
+	"github.com/tigera/operator/pkg/common"
 	"github.com/tigera/operator/pkg/controller/options"
 	"github.com/tigera/operator/pkg/controller/utils"
 	"github.com/tigera/operator/pkg/ctrlruntime"
@@ -139,7 +141,7 @@ func (r *ReconcileWaypointSecrets) Reconcile(ctx context.Context, request reconc
 	// Build the desired set of secrets if Istio is active.
 	targetNamespaces := map[string]bool{}
 	if istioActive {
-		_, installation, err := utils.GetInstallation(ctx, r)
+		_, installationSpec, err := utils.GetInstallationSpec(ctx, r)
 		if err != nil {
 			if errors.IsNotFound(err) {
 				reqLogger.V(1).Info("Installation not found")
@@ -148,7 +150,7 @@ func (r *ReconcileWaypointSecrets) Reconcile(ctx context.Context, request reconc
 			return reconcile.Result{}, err
 		}
 
-		pullSecrets, err := utils.GetNetworkingPullSecrets(installation, r)
+		pullSecrets, err := utils.GetInstallationPullSecrets(installationSpec, r)
 		if err != nil {
 			return reconcile.Result{}, err
 		}
@@ -161,7 +163,8 @@ func (r *ReconcileWaypointSecrets) Reconcile(ctx context.Context, request reconc
 
 		for i := range gatewayList.Items {
 			gw := &gatewayList.Items[i]
-			if string(gw.Spec.GatewayClassName) == IstioWaypointClassName {
+			if string(gw.Spec.GatewayClassName) == IstioWaypointClassName &&
+				gw.Namespace != common.OperatorNamespace() {
 				targetNamespaces[gw.Namespace] = true
 			}
 		}
@@ -179,6 +182,13 @@ func (r *ReconcileWaypointSecrets) Reconcile(ctx context.Context, request reconc
 		}
 	}
 
+	// Build the desired set keyed on (namespace, name) so that renamed or removed
+	// secrets are correctly detected as stale.
+	desiredSecrets := map[types.NamespacedName]bool{}
+	for _, obj := range toCreate {
+		desiredSecrets[types.NamespacedName{Namespace: obj.GetNamespace(), Name: obj.GetName()}] = true
+	}
+
 	// List all existing secrets managed by this controller and mark stale ones for deletion.
 	existingSecrets := &corev1.SecretList{}
 	if err := r.List(ctx, existingSecrets, client.MatchingLabels{WaypointPullSecretLabel: "true"}); err != nil {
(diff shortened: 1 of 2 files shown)


pkg/controller/istio/waypoint/waypoint_secrets_controller.go
```diff
@@ -0,0 +1,222 @@
...
+	"github.com/tigera/operator/pkg/render/common/secret"
+)
+
+const (
+	// IstioWaypointClassName is the GatewayClass name used by Istio waypoints.
+	IstioWaypointClassName = "istio-waypoint"
+
+	// WaypointPullSecretLabel is the label applied to secrets copied by this controller
+	// for tracking and cleanup purposes.
+	WaypointPullSecretLabel = "operator.tigera.io/istio-waypoint-pull-secret"
```
caseydavenport, line 54: Curious why a label is the right call here instead of using owner references? If there's a reason it would be good to expand on that in the comment.

pkg/controller/istio/waypoint/waypoint_controller.go
```diff
@@ -0,0 +1,222 @@
...
+		return reconcile.Result{}, err
+	}
+
+	pullSecrets, err := utils.GetNetworkingPullSecrets(installation, r)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	// If no pull secrets configured, clean up any previously copied secrets.
+	if len(pullSecrets) == 0 {
```
caseydavenport: Instead of an explicit branch for this, I'd expect that our reconcile logic would be able to handle this naturally? i.e., if a secret is removed, we need to remove it from namespaces anyway, so if ALL secrets are removed, I'd expect that mainline reconcile logic to already handle this.

pkg/controller/istio/waypoint/waypoint_controller.go
```diff
@@ -0,0 +1,222 @@
...
+	// Clean up stale secrets from namespaces that no longer have waypoints.
+	if err := r.cleanupStaleSecrets(ctx, targetNamespaces); err != nil {
+		return reconcile.Result{}, err
+	}
+
+	return reconcile.Result{}, nil
+}
+
+// cleanupAllSecrets removes all secrets created by this controller.
+func (r *ReconcileWaypointSecrets) cleanupAllSecrets(ctx context.Context) error {
```
caseydavenport: I think we should use the NewPassthrough `objectsToDelete` logic to handle this. WDYT?

pkg/controller/istio/istio_controller.go
```diff
@@ -95,6 +96,10 @@ func Add(mgr manager.Manager, opts options.ControllerOptions) error {
 		return fmt.Errorf("istio-controller failed to create periodic reconcile watch: %w", err)
 	}
 
+	if err := waypoint.Add(mgr, opts); err != nil {
```
caseydavenport, line 94: Hm, a bit of a new pattern here (I think all the other "sub" controllers are just added in the same place as main controllers).

Not necessarily against this, though.

electricjesus: it was just easier for me to understand this format, but also happy to go with the established pattern. Let me know what your preference would be as a maintainer 😅

pkg/controller/istio/waypoint/waypoint_secrets_controller.go
electricjesus: Went with labels here since we need to find and clean up these secrets across namespaces; not just on Gateway deletion, but also when pull secrets get removed from Installation or the Istio CR gets deleted. `MatchingLabels` makes that easy. Owner refs would only cover the Gateway-deletion case via GC, and we'd still need some way to track secrets for the other cleanup paths. Similar to how `ESGatewaySelectorLabel` works in the logstorage controller.

pkg/controller/istio/waypoint/waypoint_controller.go
```diff
@@ -0,0 +1,202 @@
...
+	}
+
+	// List all existing secrets managed by this controller and mark stale ones for deletion.
+	existingSecrets := &corev1.SecretList{}
+	if err := r.List(ctx, existingSecrets, client.MatchingLabels{WaypointPullSecretLabel: "true"}); err != nil {
+		return reconcile.Result{}, fmt.Errorf("failed to list waypoint pull secrets: %w", err)
+	}
+	for i := range existingSecrets.Items {
+		s := &existingSecrets.Items[i]
+		if !targetNamespaces[s.Namespace] {
```
caseydavenport: This checks staleness by namespace only, not by secret name. If the Installation's `imagePullSecrets` list changes (e.g., `secret-a` is renamed to `secret-b`), the old secret won't be cleaned up because the namespace is still a target.

I think you need to key on `(namespace, name)` pairs instead using, e.g.,

```go
types.NamespacedName{...}
```

pkg/controller/istio/waypoint/waypoint_secrets_controller_test.go
```diff
@@ -0,0 +1,293 @@
...
+
+			// Reconcile again.
+			_, err = doReconcile()
+			Expect(err).ShouldNot(HaveOccurred())
+
+			// Secrets should be cleaned up.
+			Expect(listTrackedSecrets()).To(BeEmpty())
+		})
+
+		It("should not take action for non-matching gatewayClassName", func() {
```
caseydavenport, line 248: Good coverage overall. A few cases I think would be worth adding:

1. **Pull secret rename/change while gateways exist** — change `imagePullSecrets` from `[secret-a]` to `[secret-b]` and verify `secret-a` is cleaned up. This would exercise the stale cleanup bug I flagged in the controller.
2. **Multiple pull secrets** — verify both get copied, not just the first.
3. **Gateway in the operator namespace** — edge case: should the controller copy a secret to the namespace it already lives in?

pkg/controller/istio/waypoint/waypoint_secrets_controller.go
```diff
@@ -0,0 +1,202 @@
+// Copyright (c) 2026 Tigera, Inc. All rights reserved.
```
caseydavenport, line 1: Should probably name this "waypoint_secrets_controller.go"

PR #4489, opened 2026-03-03
Enable tiers oss

radixo:
## Description

* Enable calico-system tier for OSS
* Translate k8s NPs used on OSS renders to CNPs

## Release Note

```release-note-not-required
TBD
```

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/guardian_test.go
+++ b/pkg/render/guardian_test.go
@@ -334,11 +334,11 @@ var _ = Describe("Rendering tests", func() {
 			It("should render OSS network policy regardless of IncludeEgressNetworkPolicy flag", func() {
 				// OSS variant should always render a network policy, even when IncludeEgressNetworkPolicy is false
 				renderGuardianPolicy("127.0.0.1:1234", false, operatorv1.Calico, false)
-				
+
 				policyName := types.NamespacedName{Name: "calico-system.guardian-access", Namespace: "calico-system"}
 				policy := testutils.GetCalicoSystemPolicyFromResources(policyName, resources)
 				Expect(policy).NotTo(BeNil(), "OSS variant should always render a network policy")
-				
+
 				// Verify it's the OSS policy (should only have Ingress type, no Egress)
 				Expect(policy.Spec.Types).To(ConsistOf(v3.PolicyTypeIngress))
 				Expect(policy.Spec.Egress).To(BeEmpty())
@@ -347,7 +347,7 @@ var _ = Describe("Rendering tests", func() {
 			It("should not render Enterprise network policy when IncludeEgressNetworkPolicy is false", func() {
 				// Enterprise variant with IncludeEgressNetworkPolicy=false should not render any policy
 				renderGuardianPolicy("127.0.0.1:1234", false, operatorv1.TigeraSecureEnterprise, false)
-				
+
 				policyName := types.NamespacedName{Name: "calico-system.guardian-access", Namespace: "calico-system"}
 				policy := testutils.GetCalicoSystemPolicyFromResources(policyName, resources)
 				Expect(policy).To(BeNil(), "Enterprise variant with IncludeEgressNetworkPolicy=false should not render a network policy")
@@ -356,11 +356,11 @@ var _ = Describe("Rendering tests", func() {
 			It("should render Enterprise network policy when IncludeEgressNetworkPolicy is true", func() {
 				// Enterprise variant with IncludeEgressNetworkPolicy=true should render the full policy
 				renderGuardianPolicy("127.0.0.1:1234", false, operatorv1.TigeraSecureEnterprise, true)
-				
+
 				policyName := types.NamespacedName{Name: "calico-system.guardian-access", Namespace: "calico-system"}
 				policy := testutils.GetCalicoSystemPolicyFromResources(policyName, resources)
 				Expect(policy).NotTo(BeNil(), "Enterprise variant with IncludeEgressNetworkPolicy=true should render a network policy")
-				
+
 				// Verify it's the Enterprise policy (should have both Ingress and Egress types)
 				Expect(policy.Spec.Types).To(ConsistOf(v3.PolicyTypeIngress, v3.PolicyTypeEgress))
 				Expect(policy.Spec.Egress).NotTo(BeEmpty())


pkg/controller/apiserver/apiserver_controller.go
```diff
@@ -79,14 +79,12 @@ func Add(mgr manager.Manager, opts options.ControllerOptions) error {
 	}
 
 	// Established deferred watches against the v3 API that should succeed after the Enterprise API Server becomes available.
-	if opts.EnterpriseCRDExists {
-		// Watch for changes to Tier, as its status is used as input to determine whether network policy should be reconciled by this controller.
-		go utils.WaitToAddTierWatch(networkpolicy.TigeraComponentTierName, c, opts.K8sClientset, log, r.tierWatchReady)
+	// Watch for changes to Tier, as its status is used as input to determine whether network policy should be reconciled by this controller.
+	go utils.WaitToAddTierWatch(networkpolicy.TigeraComponentTierName, c, opts.K8sClientset, log, r.tierWatchReady)
```
caseydavenport:
```suggestion
	go utils.WaitToAddTierWatch(networkpolicy.CalicoTier, c, opts.K8sClientset, log, r.tierWatchReady)
```

^ We should rename this now that it's not Tigera specific.

pkg/controller/apiserver/apiserver_controller.go
```diff
@@ -425,6 +405,24 @@ func (r *ReconcileAPIServer) Reconcile(ctx context.Context, request reconcile.Re
 		}
 	}
 
+	includeV3NetworkPolicy := false
```
caseydavenport: Why did this move?

pkg/controller/clusterconnection/clusterconnection_controller.go
```diff
@@ -83,13 +83,13 @@ func Add(mgr manager.Manager, opts options.ControllerOptions) error {
...
-		go utils.WaitToAddNetworkPolicyWatches(c, opts.K8sClientset, log, []types.NamespacedName{
-			{Name: render.GuardianPolicyName, Namespace: render.GuardianNamespace},
-			{Name: networkpolicy.TigeraComponentDefaultDenyPolicyName, Namespace: render.GuardianNamespace},
-		})
 	}
+	go utils.WaitToAddTierWatch(networkpolicy.TigeraComponentTierName, c, opts.K8sClientset, log, tierWatchReady)
+
+	go utils.WaitToAddNetworkPolicyWatches(c, opts.K8sClientset, log, []types.NamespacedName{
+		{Name: render.GuardianPolicyName, Namespace: render.GuardianNamespace},
+		{Name: networkpolicy.TigeraComponentDefaultDenyPolicyName, Namespace: render.GuardianNamespace},
```
caseydavenport: We should rename all of these as well. 
```suggestion
		{Name: networkpolicy.DefaultDenyPolicyName, Namespace: render.GuardianNamespace},
```

pkg/render/goldmane/component.go
```diff
@@ -100,7 +102,7 @@ func (c *Component) SupportedOSType() rmeta.OSType {
 	return rmeta.OSTypeLinux
 }
 
-func (c *Component) Objects() ([]client.Object, []client.Object) {
+func (c *Component) Objects() (toCreate, toDelete []client.Object) {
```
caseydavenport: Prefer not to use named return values unless it's really necessary.

pkg/render/goldmane/component.go
```diff
@@ -312,29 +318,32 @@ func (c *Component) role() *rbacv1.Role {
...
 			},
 		},
 	}
 }
+
+func (c *Component) deprecatedObjects() []client.Object {
+	return []client.Object{
+		&netv1.NetworkPolicy{
+			TypeMeta:   metav1.TypeMeta{Kind: "NetworkPolicy", APIVersion: "networking.k8s.io/v1"},
+			ObjectMeta: metav1.ObjectMeta{Name: "goldmane", Namespace: GoldmaneNamespace},
```
caseydavenport, line 410: Worth a comment on this one explaining.

pkg/render/whisker/component.go
```diff
@@ -121,7 +121,7 @@ func (c *Component) SupportedOSType() rmeta.OSType {
 	return rmeta.OSTypeLinux
 }
 
-func (c *Component) Objects() ([]client.Object, []client.Object) {
+func (c *Component) Objects() (toCreate, toDelete []client.Object) {
```
caseydavenport: Same here

pkg/render/whisker/component_test.go
```diff
@@ -70,7 +70,7 @@ var _ = Describe("ComponentRendering", func() {
 				WhiskerBackendKeyPair: defaultTLSKeyPair,
 				Whisker:               &operatorv1.Whisker{Spec: operatorv1.WhiskerSpec{Notifications: ptr.ToPtr(operatorv1.Enabled)}},
 			},
-			0, numExpectedObjects,
+			0, numExpectedObjects+1,
```
caseydavenport: Seems odd - why a +1 here instead of just updating "numExpectedObjects"?

pkg/render/guardian.go
```diff
@@ -88,16 +86,24 @@ func Guardian(cfg *GuardianConfiguration) Component {
...
 func GuardianPolicy(cfg *GuardianConfiguration) (Component, error) {
-	guardianAccessPolicy, err := guardianCalicoSystemPolicy(cfg)
-	if err != nil {
-		return nil, err
-	}
+	var policies []client.Object
 
-	return NewPassthrough(
-		[]client.Object{
+	// XXX TODO: To implement different policies when is Enterprise and
```
caseydavenport: Do we have an issue tracking this? Best not to leave it just as a TODO unless you're planning on addressing in this PR.

pkg/controller/apiserver/apiserver_controller.go
radixo: I'm thinking to change this one to CalicoTierName, to be more descritive, once it stores a the Tier name and not a Tier component struct, WDYT?

caseydavenport: sounds good!

pkg/controller/apiserver/apiserver_controller.go
radixo: No reason, probably during a test I moved it and didn't realise. My bad, fixing it.

pkg/render/guardian.go
radixo: @Brian-McM was any Issue created for this one? I'm removing this comment.

pkg/render/common/networkpolicy/networkpolicy.go
```diff
@@ -184,25 +184,18 @@ func CalicoSystemDefaultDeny(namespace string) *v3.NetworkPolicy {
...
 
 // Entity rules not belonging to Calico/Tigera components.
-var KubeAPIServerEntityRule = v3.EntityRule{
-	Services: &v3.ServiceMatch{
-		Name:      "kubernetes",
-		Namespace: "default",
-	},
-}
-
 var KubeAPIServerServiceSelectorEntityRule = v3.EntityRule{
```
caseydavenport: I think we should keep the other one :p 

They're identical, but the other one has a much nicer name.

pkg/render/guardian.go
```diff
@@ -88,16 +86,21 @@ func Guardian(cfg *GuardianConfiguration) Component {
 }
 
 func GuardianPolicy(cfg *GuardianConfiguration) (Component, error) {
-	guardianAccessPolicy, err := guardianCalicoSystemPolicy(cfg)
-	if err != nil {
-		return nil, err
-	}
+	var policies []client.Object
```
caseydavenport, line 89: I think there's a bug here. For OSS, `IncludeEgressNetworkPolicy` will always be false (it's gated on the enterprise license having `EgressAccessControlFeature`), so `policies` will be empty and no CNP gets rendered for Guardian at all.

Meanwhile, the old k8s NetworkPolicy was removed from the Guardian component and added to `deprecatedObjects()`. So OSS Guardian ends up with zero network policy — the old one is deleted and the new one is never created.

It looks like `ossNetworkPolicy()` inside `guardianCalicoSystemPolicy` was intended to cover the OSS case, but it's dead code because it's behind this gate.

I think the fix is to always call `guardianCalicoSystemPolicy` when the tier is available, and use `IncludeEgressNetworkPolicy` (or variant) to control just the egress rules within the policy, rather than gating the entire policy creation.

pkg/render/goldmane/component.go
```diff
@@ -312,29 +318,35 @@ func (c *Component) role() *rbacv1.Role {
...
-			"app.kubernetes.io/name": GoldmaneDeploymentName,
-		},
-	}
-}
-
-func (c *Component) networkPolicy() *netv1.NetworkPolicy {
-	return &netv1.NetworkPolicy{
+func (c *Component) networkPolicy() *v3.NetworkPolicy {
+	return &v3.NetworkPolicy{
 		TypeMeta:   metav1.TypeMeta{Kind: "NetworkPolicy", APIVersion: "networking.k8s.io/v1"},
```
caseydavenport: This is a `*v3.NetworkPolicy` but the TypeMeta says `networking.k8s.io/v1`. Should be `projectcalico.org/v3`.

pkg/render/whisker/component.go
```diff
@@ -267,27 +271,27 @@ func (c *Component) deployment() *appsv1.Deployment {
...
-				Protocol: ptr.ToPtr(corev1.ProtocolTCP),
-				Port:     ptr.ToPtr(intstr.FromInt32(GoldmaneServicePort)),
-			}},
 		},
 	}
-	return &netv1.NetworkPolicy{
+	egressRules = networkpolicy.AppendDNSEgressRules(egressRules, c.cfg.OpenShift)
+
+	return &v3.NetworkPolicy{
 		TypeMeta:   metav1.TypeMeta{Kind: "NetworkPolicy", APIVersion: "networking.k8s.io/v1"},
```
caseydavenport: Same TypeMeta bug — `*v3.NetworkPolicy` with `APIVersion: "networking.k8s.io/v1"`. Should be `projectcalico.org/v3`.

pkg/render/whisker/component.go
```diff
@@ -267,27 +271,27 @@ func (c *Component) deployment() *appsv1.Deployment {
...
-		ObjectMeta: metav1.ObjectMeta{Name: WhiskerName, Namespace: WhiskerNamespace},
-		Spec: netv1.NetworkPolicySpec{
-			PolicyTypes: []netv1.PolicyType{netv1.PolicyTypeIngress, netv1.PolicyTypeEgress},
-			PodSelector: *selector.PodLabelSelector(WhiskerDeploymentName),
-			Egress:      append(egressRules, networkpolicy.K8sDNSEgressRules(c.cfg.OpenShift)...),
+		ObjectMeta: metav1.ObjectMeta{Name: WhiskerPolicyName, Namespace: WhiskerNamespace},
+		Spec: v3.NetworkPolicySpec{
+			Tier:     networkpolicy.CalicoTierName,
+			Types:    []v3.PolicyType{v3.PolicyTypeEgress},
```
caseydavenport: The old policy had both ingress / egress defined here - is this intentional?

pkg/render/guardian.go
```diff
@@ -547,37 +549,43 @@ func (c *GuardianComponent) annotations() map[string]string {
 	return annotations
 }
 
-func (c *GuardianComponent) networkPolicy() *netv1.NetworkPolicy {
-	return &netv1.NetworkPolicy{
+func ossNetworkPolicy() *v3.NetworkPolicy {
+	return &v3.NetworkPolicy{
 		TypeMeta:   metav1.TypeMeta{Kind: "NetworkPolicy", APIVersion: "networking.k8s.io/v1"},
```
caseydavenport: Same TypeMeta bug here — this is a `*v3.NetworkPolicy` but has `APIVersion: "networking.k8s.io/v1"`. Should be `projectcalico.org/v3`.

pkg/render/goldmane/component.go
radixo: :shaking_face: sorry

pkg/render/guardian.go
radixo: Actually that it is how it was before, this code it was the most confusing one, you can see that I treated on a way to leave the Entreprise how it was, and created a v3.NetworkPolicy for OSS. About the way that the Enterprise was creating the rule I warned Operator team.

pkg/render/whisker/component.go
radixo: No, I really missed that, good catch!

pkg/render/guardian.go
caseydavenport: @radixo `ossNetworkPolicy` is called from `guardianCalicoSystemPolicy` which is only ever called when ` cfg.IncludeEgressNetworkPolicy` is true, which only ever happens on enterprise. Right?

PR #4493, opened 2026-03-03
Fix legacy CA-signed certs not reissued on SAN mismatch

caseydavenport:
Fixes projectcalico/calico#11363

After the calico-apiserver namespace migration from `calico-apiserver` to
`calico-system` in v3.31, long-lived clusters with older operator-signed certs
hit TLS errors because the cert SANs still referenced the old namespace.

The root cause is in `getKeyPair`: the issuer identity check at line 504 used
exact match (`==`) against `TigeraOperatorCAIssuerPrefix`, but legacy operator
CAs use the format `tigera-operator-signer@<timestamp>`. This caused valid
legacy certs to be misidentified as BYO (Issuer=nil, so `BYO()` returns true),
which skips the SAN revalidation in `GetOrCreateKeyPair`.

The other issuer checks in the same function (lines 460, 489) already use
`strings.HasPrefix` correctly. This just makes the last check consistent.

With the fix, old certs enter the `HasPrefix` block, hit the authority key ID
mismatch (old CA key != current CA key), and `getKeyPair` returns nil —
triggering `GetOrCreateKeyPair` to issue a new cert with correct SANs.

```release-note
Fix calico-apiserver TLS errors on upgrade to v3.31 for long-lived clusters.
The operator now correctly reissues certificates with updated SANs when the
apiserver namespace changes, instead of treating legacy operator-signed certs
as user-provided.
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -518,9 +518,9 @@ format-check:
 dirty-check:
 	@if [ "$$(git diff --stat)" != "" ]; then \
 	echo "The following files are dirty"; git diff --stat; exit 1; fi
-	@# Check that no new CRDs needed to be committed
-	@if [ "$$(git status --porcelain pkg/imports/crds)" != "" ]; then \
-	echo "The following CRD files need to be added"; git status --porcelain pkg/imports/crds; exit 1; fi
+	@# Check that no new CRDs or admission policies needed to be committed
+	@if [ "$$(git status --porcelain pkg/imports)" != "" ]; then \
+	echo "The following imported files need to be added"; git status --porcelain pkg/imports; exit 1; fi
 
 foss-checks:
 	@echo Running $@...
@@ -680,9 +680,11 @@ $(BINDIR)/gen-versions: $(shell find ./hack/gen-versions -type f)
 define prep_local_crds
     $(eval product := $(1))
 	rm -rf pkg/imports/crds/$(product)
+	rm -rf pkg/imports/admission/$(product)
 	rm -rf .crds/$(product)
 	mkdir -p pkg/imports/crds/$(product)/v1.crd.projectcalico.org/
 	mkdir -p pkg/imports/crds/$(product)/v3.projectcalico.org/
+	mkdir -p pkg/imports/admission/$(product)
 	mkdir -p .crds/$(product)
 endef
 
@@ -711,6 +713,11 @@ define copy_eck_crds
 		$(eval product := $(2))
 	@cp $(dir)/charts/crd.projectcalico.org.v1/templates/eck/* pkg/imports/crds/$(product)/ && echo "Copied $(product) ECK CRDs"
 endef
+define copy_admission_policies
+    $(eval dir := $(1))
+		$(eval product := $(2))
+	@cp $(dir)/api/admission/* pkg/imports/admission/$(product)/ && echo "Copied $(product) admission policies"
+endef
 
 .PHONY: read-libcalico-version read-libcalico-enterprise-version
 .PHONY: update-calico-crds update-enterprise-crds
@@ -729,6 +736,7 @@ read-libcalico-calico-version:
 update-calico-crds: fetch-calico-crds
 	$(call copy_v1_crds, $(CALICO_CRDS_DIR),"calico")
 	$(call copy_v3_crds, $(CALICO_CRDS_DIR),"calico")
+	$(call copy_admission_policies, $(CALICO_CRDS_DIR),"calico")
 
 prepare-for-calico-crds:
 	$(call prep_local_crds,"calico")
@@ -749,6 +757,7 @@ update-enterprise-crds: fetch-enterprise-crds
 	$(call copy_v1_crds,$(ENTERPRISE_CRDS_DIR),"enterprise")
 	$(call copy_v3_crds, $(ENTERPRISE_CRDS_DIR),"enterprise")
 	$(call copy_eck_crds,$(ENTERPRISE_CRDS_DIR),"enterprise")
+	$(call copy_admission_policies,$(ENTERPRISE_CRDS_DIR),"enterprise")
 
 prepare-for-enterprise-crds:
 	$(call prep_local_crds,"enterprise")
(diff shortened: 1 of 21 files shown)


pkg/controller/certificatemanager/certificatemanager_test.go
```diff
@@ -315,6 +322,21 @@ var _ = Describe("Test CertificateManagement suite", func() {
 				Expect(kp.GetCertificatePEM()).NotTo(Equal(secret.Data[corev1.TLSCertKey]))
 			})
 
+			It("should reissue a valid legacy CA-signed cert when DNS names don't match", func() {
```
rene-dekker: This test will never hit line certificatemanager.go:504

rene-dekker: It looks like the more recent changes to renew secrets prior to expiration, make it such that some tests are not proper now. Test secrets such as legacySecret, legacyWithClientKeyUsage, byoSecret, legacyBYOSecret, validLegacyCASecret are created with 1 Hour of validity. We should modify these so that they are `365*24*time.Hour` or something like that.

rene-dekker: The reason that test "verifying it does replace a legacy secret when dns names are missing" passes is because the legacy secret is missing extended key usages. 
So I think your fix is correct, but maybe this test is not necessary, as long as we fix the duration in the above post and as long as we fix the existing test, such that `legacyWithClientKeyUsage` is used (For that it needs to have the legacy secret name though). (introduced in https://github.com/tigera/operator/pull/4396/changes)

Something like the following maybe?

```
diff --git a/pkg/controller/certificatemanager/certificatemanager_test.go b/pkg/controller/certificatemanager/certificatemanager_test.go
index 9e9f80531..c198af231 100644
--- a/pkg/controller/certificatemanager/certificatemanager_test.go
+++ b/pkg/controller/certificatemanager/certificatemanager_test.go
@@ -84,19 +84,19 @@ var _ = BeforeSuite(func() {
        certkeyusage.SetCertKeyUsage(legacySecretName, []x509.ExtKeyUsage{x509.ExtKeyUsageClientAuth, x509.ExtKeyUsageServerAuth})
        // Create a legacy secret (how certs were before v1.24) with non-standardized legacy key and cert name, and no CA.
        // Use a secret
-       legacySecret, err = secret.CreateTLSSecret(nil, legacySecretName, appNs, legacyKeyFieldName, legacyCertFieldName, time.Hour, legacyOpts, legacySecretName)
+       legacySecret, err = secret.CreateTLSSecret(nil, legacySecretName, appNs, legacyKeyFieldName, legacyCertFieldName, 365*24*time.Hour, legacyOpts, legacySecretName)
 
        // This is a special case, which may or may not exist in the wild. It's a legacy-style certificate signed by tigera-operator but also with client usage.
-       legacyWithClientKeyUsage, err = secret.CreateTLSSecret(nil, appSecretName, appNs, legacyKeyFieldName, legacyCertFieldName, time.Hour, modernOpts, appSecretName)
+       legacyWithClientKeyUsage, err = secret.CreateTLSSecret(nil, legacySecretName, appNs, legacyKeyFieldName, legacyCertFieldName, 365*24*time.Hour, modernOpts, appSecretName)
 
        // Create a byo secret with non-standardized legacy key and cert name (like our docs for felix/typha).
        cryptoCA, err := tls.MakeCA("byo-ca")
-       byoSecret, err = secret.CreateTLSSecret(cryptoCA, appSecretName, appNs, "key.key", "cert.crt", time.Hour, modernOpts, appSecretName)
+       byoSecret, err = secret.CreateTLSSecret(cryptoCA, appSecretName, appNs, "key.key", "cert.crt", 365*24*time.Hour, modernOpts, appSecretName)
-       legacyBYOSecret, err = secret.CreateTLSSecret(cryptoCA, legacySecretName, appNs, "key.key", "cert.crt", time.Hour, legacyOpts, legacySecretName)
+       legacyBYOSecret, err = secret.CreateTLSSecret(cryptoCA, legacySecretName, appNs, "key.key", "cert.crt", 365*24*time.Hour, legacyOpts, legacySecretName)
        expiredBYOSecret, err = secret.CreateTLSSecret(cryptoCA, appSecretName, appNs, "key.key", "cert.crt", -time.Hour, modernOpts, appSecretName)
@@ -110,7 +110,7 @@ var _ = BeforeSuite(func() {
        // Create a valid (non-expired) cert signed by the legacy CA with modern key usage. This simulates a
        // long-lived cluster where the cert is still valid but has stale DNS SANs (e.g., after the
        // calico-apiserver namespace migration from calico-apiserver to calico-system).
-       validLegacyCASecret, err = secret.CreateTLSSecret(legacyCryptoCA, appSecretName, appNs, corev1.TLSPrivateKeyKey, corev1.TLSCertKey, time.Hour, modernOpts, "old-dns-name")
+       validLegacyCASecret, err = secret.CreateTLSSecret(legacyCryptoCA, appSecretName, appNs, corev1.TLSPrivateKeyKey, corev1.TLSCertKey, 365*24*time.Hour, modernOpts, "old-dns-name")
 })
 
@@ -536,7 +536,7 @@ var _ = Describe("Test CertificateManagement suite", func() {
                        test.VerifyCertSANs(keyPair.GetCertificatePEM(), missingDNSNames...)
 
                        By("verifying it does replace a legacy secret when dns names are missing")
-                       Expect(cli.Create(ctx, legacySecret)).NotTo(HaveOccurred())
+                       Expect(cli.Create(ctx, legacyWithClientKeyUsage)).NotTo(HaveOccurred())
                        keyPair, err = certificateManager.GetOrCreateKeyPair(cli, legacySecretName, appNs, appDNSNames)
                        test.VerifyCertSANs(keyPair.GetCertificatePEM(), appDNSNames...)

```

caseydavenport: wow, good catch. I've updated the test code and confirmed that the tests now fail without the fix, and pass with it.

PR #4496, opened 2026-03-05
Add docs/principles.md for operator design principles

caseydavenport:
Adds a new `docs/principles.md` that codifies the key architecture and design principles for developing in this repo:

- **API design** — declarative APIs, minimal surface, overrides for every container, kubebuilder-first validation
- **Respecting user input** — never overwrite user fields, never delete user resources, track field ownership on shared resources, copy (don't modify) user-provided secrets/configmaps downstream
- **Component isolation** — new components default to `calico-system` (separate namespaces only when specifically needed)
- **Controller design** — watch/reconcile/render/apply pattern, pure render packages, user-facing status messages
- **Security** — mTLS for all component-to-component communication
- **Certificates and secrets** — OwnerReference conventions, naming conventions

Also refactors `docs/dev_guidelines.md` to focus on workflow (code structure, code generation, cherry-picks, common tasks) and reference `principles.md` for design guidance. Moves the code generation section from principles into dev_guidelines where it fits better.

Adds a `.claude/CLAUDE.md` with project-specific context for AI-assisted development.

### DIFF
--- a/docs/principles.md
+++ b/docs/principles.md
@@ -39,7 +39,7 @@ Code architecture and design principles for the tigera/operator repository. For 
 
 - **Component-to-component communication must be authenticated and encrypted.** Use mTLS or TLS + token-based authentication for all internal communication between operator-managed components.
 
-## Certificates and Secrets
+## Resource Ownership
 
-- **Operator-managed secrets use OwnerReferences.** Secrets created by the operator should have OwnerReferences so they get cleaned up automatically.
-- **User-provided secrets must NOT have OwnerReferences.** This is how the operator distinguishes user-provided from operator-managed secrets. They may be copied to other namespaces (with OwnerReferences on the copies), but the originals must not be claimed.
+- **Operator-created resources use OwnerReferences.** Objects created by the operator should have OwnerReferences so they get cleaned up automatically. Exceptions exist for resources where cascading deletion would be catastrophic (e.g., CRDs).
+- **User-provided resources must NOT have OwnerReferences.** This is how the operator distinguishes user-provided from operator-managed resources. They may be copied to other namespaces (with OwnerReferences on the copies), but the originals must not be claimed.


docs/principles.md
```diff
@@ -0,0 +1,49 @@
...
+  2. Read current state from the Kubernetes API.
+  3. Call into `pkg/render` to generate desired resources.
+  4. Apply via `CreateOrUpdateOrDelete`.
+  5. Report status via the TigeraStatus API.
+- **Render packages are pure.** The `pkg/render` package generates Kubernetes manifests from inputs. It should not make API calls or have side effects — that's the controller's job.
+- **Status messages are for users, not developers.** TigeraStatus conditions should be actionable and user-facing. Don't surface internal error strings or stack traces.
+
+## Security
+
+- **Component-to-component communication must use mTLS.** All internal communication between operator-managed components must be secured with mutual TLS.
```
rene-dekker: We use a combination of mTLS and TLS + token auth.

caseydavenport: @rene-dekker Where is TLS + token authz used? And what would be the guiding principle on when / where to use one vs. the other?

docs/principles.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Certificates and Secrets
+
+- **Operator-managed secrets use OwnerReferences** and are rendered via the Passthrough component.
```
rene-dekker: > and are rendered via the Passthrough component.

Is this true?

docs/principles.md
```diff
@@ -0,0 +1,49 @@
...
+- **User-provided secrets must NOT have OwnerReferences.** This is how the operator distinguishes user-provided from operator-managed secrets.
+- **Secret naming conventions:**
+  - `PrivateCertSecret` — operator-managed cert + key
```
rene-dekker: PrivateCertSecret and CustomizablePrivateCertSecret have no occurrences in this repo.

> `CertSecret` — cert only, no key

Is not true, see ComplianceServerCertSecret, IntrusionDetectionCertSecret, DPICertSecret, more

caseydavenport: Yeah, looks like it cribbed this from an outdated .md file in here. I'll update.

docs/principles.md
rene-dekker: I would think mTLS is preferred, but both are secure. 
If components would talk to components outside of the cluster then mTLS has the advantage that we can have identity being managed at a higher level than the cluster signer of JWTs. I'm thinking of technologies like Spiffe/Spire.

There are some places where service accounts talk directly to servers using their own JWT, like Manager -> Queryserver, prometheus server.

PR #4502, opened 2026-03-05
Support Canonical k8s DNS pod labels in network policies

rene-dekker:
## Summary
- Canonical Kubernetes labels DNS pods as `k8s-app=coredns` instead of the standard `k8s-app=kube-dns`. This causes operator-managed network policies to block DNS traffic on Canonical clusters.
- Updated all Calico network policy selectors to use an OR expression: `k8s-app == 'kube-dns' || k8s-app == 'coredns'`
- Updated the k8s NetworkPolicy DNS rule to use `MatchExpressions` with `In` operator for both values
- Added a second service selector rule for the `coredns` service name in `AppendServiceSelectorDNSEgressRules`
- Updated the tiers cluster DNS policy selector to match both labels

## Test plan
- [x] All `pkg/render/...` unit test suites pass
- [x] Verify on a Canonical k8s cluster that DNS egress is allowed
- [ ] Verify on a standard k8s cluster that DNS egress is still allowed

```release-note
Added label selector for networkpolicies selecting coredns on Canonical Kubernetes clusters.
```

### DIFF
--- a/pkg/render/common/networkpolicy/networkpolicy.go
+++ b/pkg/render/common/networkpolicy/networkpolicy.go
@@ -71,7 +71,7 @@ func AppendDNSEgressRules(egressRules []v3.Rule, openShift bool) []v3.Rule {
 			Destination: v3.EntityRule{
 				NamespaceSelector: "projectcalico.org/name == 'kube-system'",
 				// In most Kubernetes distros the label is for kube-dns, but in Canonical it is for coredns.
-				Selector: "k8s-app == 'kube-dns' || k8s-app == 'coredns'",
+				Selector: "k8s-app in { 'kube-dns', 'coredns' }",
 				Ports:    Ports(53),
 			},
 		})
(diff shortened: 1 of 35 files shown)


pkg/render/common/networkpolicy/networkpolicy.go
```diff
@@ -70,8 +70,9 @@ func AppendDNSEgressRules(egressRules []v3.Rule, openShift bool) []v3.Rule {
 			Protocol: &UDPProtocol,
 			Destination: v3.EntityRule{
 				NamespaceSelector: "projectcalico.org/name == 'kube-system'",
-				Selector:          "k8s-app == 'kube-dns'",
-				Ports:             Ports(53),
+				// In most Kubernetes distros the label is for kube-dns, but in Canonical it is for coredns.
+				Selector: "k8s-app == 'kube-dns' || k8s-app == 'coredns'",
```
caseydavenport: I think we can use `k8s-app in [kube-dns, coredns]` instead? A bit more clear?

caseydavenport: TBH, should this just be a service selector too?

PR #4511, opened 2026-03-10
cluster routing mode

mazdakn:
## Description

Currently, configuring cluster routing mode is exposed to users through `programClusterRoutes` in both [FelixConfig](https://github.com/projectcalico/calico/blob/2ef4ccf00ba37563d57f288955d57f950397a342/api/pkg/apis/projectcalico/v3/felixconfig.go#L576) and [BGPConfig](https://github.com/projectcalico/calico/blob/2ef4ccf00ba37563d57f288955d57f950397a342/api/pkg/apis/projectcalico/v3/bgpconfig.go#L156).

This PR introduces a new config option to the installation resource, named `clusterRoutingMode` to provide better UX for users to set cluster routing mode. Supported values are `BIRD` (default) and `Felix`.

When set to :
- `BIRD`, Calico route engine, i.e. BIRD, is responsible for programming IPIP and no-encap routes. This effectively updates `programClusterRoutes` in `FelixConfig` to `Disabled` and in `BGPConfig` to `Enabled`
- When set to `Felix`, Felix will program those routes instead. As such `programClusterRoutes` is updated in `FelixConfig` to `Disabled` and in `BGPConfig` to `Enabled`

This PR assumes `BGPConfiguration` resources is always available and only performs patch operation. As such The following PR is need which created the default `BGPConfiguration` resource on startup:
https://github.com/projectcalico/calico/pull/12045

Operator now needs to create `bgpconfiguration` resources which is handed in this PR: https://github.com/projectcalico/calico/pull/12080

## Release Note

```release-note
Config option to control whether BIRD or Felix manages intra-cluster routing. 
```

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/imports/crds/enterprise/v3.projectcalico.org/projectcalico.org_globalnetworkpolicies.yaml
+++ b/pkg/imports/crds/enterprise/v3.projectcalico.org/projectcalico.org_globalnetworkpolicies.yaml
@@ -483,6 +483,8 @@ spec:
             - metadata
             - spec
           type: object
+      selectableFields:
+        - jsonPath: .spec.tier
       served: true
       storage: true
       subresources: {}
--- a/pkg/imports/crds/enterprise/v3.projectcalico.org/projectcalico.org_tiers.yaml
+++ b/pkg/imports/crds/enterprise/v3.projectcalico.org/projectcalico.org_tiers.yaml
@@ -23,6 +23,10 @@ spec:
           jsonPath: .spec.defaultAction
           name: DefaultAction
           type: string
+        - description: Current status of the tier
+          jsonPath: .status.conditions[?(@.type=='Ready')].reason
+          name: Status
+          type: string
       name: v3
       schema:
         openAPIV3Schema:
@@ -49,6 +53,45 @@ spec:
                 order:
                   type: number
               type: object
+            status:
+              properties:
+                conditions:
+                  items:
+                    properties:
+                      lastTransitionTime:
+                        format: date-time
+                        type: string
+                      message:
+                        maxLength: 32768
+                        type: string
+                      observedGeneration:
+                        format: int64
+                        minimum: 0
+                        type: integer
+                      reason:
+                        maxLength: 1024
+                        minLength: 1
+                        pattern: ^[A-Za-z]([A-Za-z0-9_,:]*[A-Za-z0-9_])?$
+                        type: string
+                      status:
+                        enum:
+                          - "True"
+                          - "False"
+                          - Unknown
+                        type: string
+                      type:
+                        maxLength: 316
+                        pattern: ^([a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*/)?(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])$
+                        type: string
+                    required:
+                      - lastTransitionTime
+                      - message
+                      - reason
+                      - status
+                      - type
+                    type: object
+                  type: array
+              type: object
           required:
             - metadata
             - spec
@@ -68,4 +111,5 @@ spec:
                 : true"
       served: true
       storage: true
-      subresources: {}
+      subresources:
+        status: {}


api/v1/installation_types.go
```diff
@@ -614,6 +624,14 @@ type CalicoNetworkSpec struct {
 	// +kubebuilder:validation:Enum=Enabled;Disabled
 	BGP *BGPOption `json:"bgp,omitempty"`
 
+	// ClusterRoutingMode controls how nodes gets a route to a workload on another node,
+	// when that workload's IP comes from an IP Pool with vxlanMode: Never. When ClusterRoutingMode is BIRD,
+	// confd and BIRD program that route. When ClusterRoutingMode is Felix, it is expected that Felix will program that route.
+	// Felix always programs such routes for IP Pools with vxlanMode: Always or vxlanMode: CrossSubnet. [Default: BIRD]
+	// +optional
+	// +kubebuilder:validation:Enum=BIRD;Felix
```
caseydavenport: I think we usually put the Enum validation on the enum type so it can be re-used wherever the enum is used.

pkg/controller/installation/core_controller.go
```diff
@@ -2014,6 +2036,61 @@ func (r *ReconcileInstallation) setDefaultsOnFelixConfiguration(ctx context.Cont
...
+// setDefaultOnBGPConfiguration will take the passed in bgpConfig and add any defaulting needed
+// based on the install config.
+func (r *ReconcileInstallation) setClusterRoutingOnBGPConfiguration(
+	ctx context.Context,
+	install *operatorv1.Installation,
+	bgpConfig *v3.BGPConfiguration,
+	reqLogger logr.Logger,
+) (bool, error) {
+	updated := false
+	desiredValue := "Enabled"
```
caseydavenport, line 2069: Are there constants we should be using for this?

pkg/controller/installation/core_controller.go
```diff
@@ -2014,6 +2036,61 @@ func (r *ReconcileInstallation) setDefaultsOnFelixConfiguration(ctx context.Cont
...
+
+	if felixProgramsClusterRoutes(install) {
+		desiredValue = "Disabled"
+	}
+
+	if bgpConfig.Spec.ProgramClusterRoutes == nil || *bgpConfig.Spec.ProgramClusterRoutes != desiredValue {
```
caseydavenport, line 2075: Hm, we usually try to avoid overwriting end-user modifications of FelixConfig, etc using an annotation. Technically I suppose we should do it here... but we're actually pretty inconsistent about it.

I've started a prototype of how to make this more easily implemented: https://github.com/tigera/operator/pull/4523

But I don't want to block this PR on it. So I'll just make sure to incorporate this into my changes.

pkg/controller/installation/core_controller.go
```diff
@@ -1147,7 +1147,29 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
...
+		// Configure defaults.
+		u, err := r.setClusterRoutingOnBGPConfiguration(ctx, instance, bgpConfig, reqLogger)
+		if err != nil {
+			return false, err
+		}
+
+		return u, nil
 	})
 	if err != nil {
 		return reconcile.Result{}, err
```
caseydavenport, line 1175: I don't think we can make the Felix + BGP patches atomic, but it's worth noting that if the FelixConfiguration patch above succeeds and this BGPConfiguration patch fails, we'll have an inconsistent state until the next reconcile converges them. Might be worth a comment here calling that out.

pkg/controller/installation/core_controller.go
```diff
@@ -2014,6 +2036,61 @@ func (r *ReconcileInstallation) setDefaultsOnFelixConfiguration(ctx context.Cont
...
+// setDefaultOnBGPConfiguration will take the passed in bgpConfig and add any defaulting needed
+// based on the install config.
+func (r *ReconcileInstallation) setClusterRoutingOnFelixConfiguration(
+	ctx context.Context,
+	install *operatorv1.Installation,
+	fc *v3.FelixConfiguration,
+	reqLogger logr.Logger,
+) (bool, error) {
+	updated := false
+	desiredValue := "Disabled"
```
caseydavenport: `ctx` is unused in this function — can be removed.

pkg/controller/installation/core_controller.go
caseydavenport: `ctx` is unused here too.

Also, both this and `setClusterRoutingOnFelixConfiguration` above have stale doc comments — they both say `setDefaultOnBGPConfiguration` which looks like a copy-paste leftover. Should match the actual function names.

pkg/controller/installation/validation.go
```diff
@@ -167,14 +168,16 @@ func validateCustomResource(instance *operatorv1.Installation) error {
 					// Verify the specified encapsulation type is valid.
 					switch pool.Encapsulation {
 					case operatorv1.EncapsulationIPIP, operatorv1.EncapsulationIPIPCrossSubnet:
-						// IPIP currently requires BGP to be running in order to program routes.
-						if instance.Spec.CalicoNetwork.BGP == nil || *instance.Spec.CalicoNetwork.BGP == operatorv1.BGPDisabled {
+						// In BIRD cluster routing mode, IPIP currently requires BGP to be running in order to program routes.
+						if birdClusterRoutingMode &&
+							(instance.Spec.CalicoNetwork.BGP == nil || *instance.Spec.CalicoNetwork.BGP == operatorv1.BGPDisabled) {
 							return fmt.Errorf("IPIP encapsulation requires that BGP is enabled")
```
caseydavenport: The error message still just says "IPIP encapsulation requires that BGP is enabled" — now that there's a way to fix this without BGP (by setting `ClusterRoutingMode: Felix`), should the error message mention that option? Same for the unencapsulated case below. Otherwise users won't know about the escape hatch.

pkg/controller/installation/core_controller_test.go
```diff
@@ -1427,6 +1427,65 @@ var _ = Describe("Testing core-controller installation", func() {
...
+			Expect(*fc.Spec.ProgramClusterRoutes).To(Equal("Enabled"))
+
+			bgpConfig := &v3.BGPConfiguration{}
+			err = c.Get(ctx, types.NamespacedName{Name: "default"}, bgpConfig)
+			Expect(err).ShouldNot(HaveOccurred())
+			Expect(bgpConfig.Spec.ProgramClusterRoutes).NotTo(BeNil())
+			Expect(*bgpConfig.Spec.ProgramClusterRoutes).To(Equal("Disabled"))
+		})
+
 		It("should set vxlanVNI to 10000 when provider is DockerEE", func() {
```
caseydavenport, line 1516: Good coverage for the happy paths. Might be worth adding a test for the case where the default BGPConfiguration doesn't already exist — to verify the create-on-patch path in `PatchBGPConfiguration` works correctly (or to verify we get an appropriate error if the prerequisite calico PR hasn't landed).

pkg/controller/installation/validation.go
```diff
@@ -139,6 +139,7 @@ func validateCustomResource(instance *operatorv1.Installation) error {
 	// Verify Calico settings, if specified.
 	if instance.Spec.CalicoNetwork != nil {
 		bpfDataplane := instance.Spec.CalicoNetwork.LinuxDataplane != nil && *instance.Spec.CalicoNetwork.LinuxDataplane == operatorv1.LinuxDataplaneBPF
+		birdClusterRoutingMode := !felixProgramsClusterRoutes(instance)
```
caseydavenport: nit: I find this a bit confusing with the negation.


```suggestion
		felixClusterRouting := felixProgramsClusterRoutes(instance)
```

mazdakn: fixed.

pkg/controller/installation/validation.go
mazdakn: Updated.

api/v1/installation_types.go
mazdakn: Fixed.

pkg/controller/installation/core_controller.go
mazdakn: Thanks for the input and including the case in your PR.

pkg/controller/installation/core_controller.go
mazdakn: No, for these options in the BGPConfig and FelixConfig we use string type, and we rely on kubebuilder to validate inputs.
We use both approach though. But it's cleaner to declare constants. I'll address it separately since needs merging a few PRs.

mazdakn: Fixed.

pkg/controller/installation/core_controller.go
mazdakn: Added.

pkg/controller/installation/core_controller_test.go
mazdakn: Interesting. Is create-on-patch a feature of Operator? I was trying to figure out how felixconfiguration is created.

mazdakn: I see! it's part of the patching logic.

caseydavenport (approved):

PR #4529, opened 2026-03-11
feat(logstorage): validate replicas < node count to prevent ILM stall

tianfeng92:
## Description

On single-node ES clusters with replicas: 1, replica shards can never be allocated. This causes the ILM warm phase migrate action to wait indefinitely for shard copies to become active, blocking progression to the delete phase and causing indices to accumulate beyond retention.

This PR adds two levels of validation in the LogStorage initializer:

1. **Error** (`nodeCount <= replicas`): Rejects configurations where replica shards cannot be allocated at all, with a clear error message guiding users to set replicas to 0 for single-node deployments.
2. **Warning** (`nodeCount == replicas + 1`): Logs a warning when the node count only exceeds replicas by 1, since the Elasticsearch PodDisruptionBudget will prevent voluntary pod evictions (e.g., node repaving). This is not treated as an error because it is a valid and common configuration (e.g., Calico Cloud 2-node clusters with 1 replica).

For CE master
Jira ticket: CI-1940

## Release Note

```release-note
Add validation for logstorage node count and replicas setting.
```

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/logstorage/initializer/initializing_controller.go
+++ b/pkg/controller/logstorage/initializer/initializing_controller.go
@@ -168,33 +168,33 @@ func FillDefaults(opr *operatorv1.LogStorage) {
 	}
 }
 
-func validateLogStorage(spec *operatorv1.LogStorageSpec) (error, string) {
-	err, warning := validateReplicasForNodeCount(spec)
+func validateLogStorage(spec *operatorv1.LogStorageSpec) (string, error) {
+	warning, err := validateReplicasForNodeCount(spec)
 	if err != nil {
-		return err, ""
+		return "", err
 	}
 	if err := validateComponentResources(spec); err != nil {
-		return err, ""
+		return "", err
 	}
-	return nil, warning
+	return warning, nil
 }
 
-func validateReplicasForNodeCount(spec *operatorv1.LogStorageSpec) (error, string) {
+func validateReplicasForNodeCount(spec *operatorv1.LogStorageSpec) (string, error) {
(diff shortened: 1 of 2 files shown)


pkg/controller/logstorage/initializer/initializing_controller.go
```diff
@@ -168,6 +168,27 @@ func FillDefaults(opr *operatorv1.LogStorage) {
...
+	return validateComponentResources(spec)
+}
+
+func validateReplicasForNodeCount(spec *operatorv1.LogStorageSpec) error {
+	if spec.Nodes == nil || spec.Indices == nil || spec.Indices.Replicas == nil {
+		return nil
+	}
+	replicas := int(*spec.Indices.Replicas)
+	nodeCount := int(spec.Nodes.Count)
+	if replicas > 0 && nodeCount <= replicas {
```
rene-dekker, line 189: Node count should be at least 2 higher than replica count. Otherwise pods cannot be restarted / upgraded.

rene-dekker: Potentially we want to show a warning if nodes only exceed replicas by 1, but I think we need more than just the current change

tmjd: > Otherwise pods cannot be restarted / upgraded.

I don't think that is true. ES doesn't do any type of bringing up new pod before taking down an old one.

PR #4581, opened 2026-03-19
Add Calico Ingress Gateway to calico system networkpolicy

alexh-tigera:
## Description

Fixed an issue where the Calico Ingress Gateway failed to deploy when the [recommended default deny policy](https://docs.tigera.io/calico-enterprise/latest/network-policy/default-deny) was applied.

This adds: 

* DNS and api-server egress rules for the certgen job in order to apply the generated certificates.
* DNS and api-server egress rules for envoy-gateway to manage envoy proxies.

## Release Note

```release-note
* Fixed an issue where Calico Ingress Gateway failed to deploy when default-deny rules were applied.
```

## For PR author

- [x] Tests for change.
- [ ] ~~ If changing pkg/apis/, run `make gen-files`~~
- [ ] ~~ If changing versions, run `make gen-versions`~~

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/gatewayapi/gatewayapi_controller.go
+++ b/pkg/controller/gatewayapi/gatewayapi_controller.go
@@ -40,6 +40,7 @@ import (
 	envoyapi "github.com/envoyproxy/gateway/api/v1alpha1"
 	"github.com/go-logr/logr"
 	operatorv1 "github.com/tigera/operator/api/v1"
+	"github.com/tigera/operator/pkg/common"
 	"github.com/tigera/operator/pkg/controller/options"
 	"github.com/tigera/operator/pkg/controller/status"
 	"github.com/tigera/operator/pkg/controller/utils"
@@ -496,6 +497,6 @@ func (r *ReconcileGatewayAPI) getPolicySyncPathPrefix(fcSpec *v3.FelixConfigurat
 // The bool return value indicates if the finalizer is Set
 func (r *ReconcileGatewayAPI) maintainFinalizer(ctx context.Context, gatewayAPI client.Object) (bool, error) {
 	// These objects require graceful termination before the CNI plugin is torn down.
-	gatewayAPIDeployment := v1.Deployment{ObjectMeta: metav1.ObjectMeta{Name: "envoy-gateway", Namespace: "tigera-gateway"}}
+	gatewayAPIDeployment := v1.Deployment{ObjectMeta: metav1.ObjectMeta{Name: "envoy-gateway", Namespace: common.TigeraGatewayNamespace}}
 	return utils.MaintainInstallationFinalizer(ctx, r.client, gatewayAPI, render.GatewayAPIFinalizer, &gatewayAPIDeployment)
 }
--- a/pkg/controller/gatewayapi/gatewayapi_controller.go
+++ b/pkg/controller/gatewayapi/gatewayapi_controller.go
@@ -22,6 +22,7 @@ import (
 	v1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
@@ -47,6 +48,7 @@ import (
 	"github.com/tigera/operator/pkg/controller/utils/imageset"
 	"github.com/tigera/operator/pkg/ctrlruntime"
 	"github.com/tigera/operator/pkg/render"
+	"github.com/tigera/operator/pkg/render/common/networkpolicy"
 	"github.com/tigera/operator/pkg/render/gatewayapi"
 )
 
@@ -62,6 +64,8 @@ var log = logf.Log.WithName("controller_gatewayapi")
 // Start Watches within the Add function for any resources that this controller creates or monitors. This will trigger
 // calls to Reconcile() when an instance of one of the watched resources is modified.
 func Add(mgr manager.Manager, opts options.ControllerOptions) error {
+	tierWatchReady := &utils.ReadyFlag{}
+
 	r := &ReconcileGatewayAPI{
 		client:              mgr.GetClient(),
 		scheme:              mgr.GetScheme(),
@@ -89,6 +93,12 @@ func Add(mgr manager.Manager, opts options.ControllerOptions) error {
 		log.V(5).Info("Failed to create Installation watch", "err", err)
 		return fmt.Errorf("gatewayapi-controller failed to watch Installation resource: %w", err)
 	}
+
+	go utils.WaitToAddTierWatch(networkpolicy.CalicoTierName, c, opts.K8sClientset, log, tierWatchReady)
+	go utils.WaitToAddNetworkPolicyWatches(c, opts.K8sClientset, log, []types.NamespacedName{
+		{Name: gatewayapi.GatewayControllerPolicyName, Namespace: common.TigeraGatewayNamespace},
+		{Name: gatewayapi.GatewayCertgenPolicyName, Namespace: common.TigeraGatewayNamespace},
+	})
 
 	// Perform periodic reconciliation. This acts as a backstop to catch reconcile issues,
 	// and also makes sure we spot when things change that might not trigger a reconciliation.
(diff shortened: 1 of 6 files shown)


pkg/render/gatewayapi/gateway_api.go
```diff
@@ -635,6 +646,14 @@ func (pr *gatewayAPIImplementationComponent) Objects() ([]client.Object, []clien
 
 	objs = append(objs, certgenJob)
 
+	// Add network policies to allow gateway components to function under a
+	// default-deny policy.
+	objs = append(objs,
+		pr.gatewayCertgenAllowTigeraPolicy(),
+		pr.gatewayControllerAllowTigeraPolicy(),
+		networkpolicy.CalicoSystemDefaultDeny(common.TigeraGatewayNamespace),
```
nelljerram: I don't think the operator should unconditionally install this default deny, right?  Were you only doing this for testing?  Instead I think we should have an FV test that applies the default-deny policy and then verifies that CIG can still be successfully deployed.

nelljerram (changes requested):
Looking good, except
- I don't think operator should unconditionally install default-deny policy for tigera-gateway namespace
- I do think we should have an FV test that does that, and then checks CIG can be successfully deployed.

pkg/render/gatewayapi/gateway_api.go
alexh-tigera: Yes, I missed that after removing the other changes I made. While doing that I also realized that it needs a watch for the network policies, so I've fix the default deny and added the watches in 48a0ab83c.

~~Where do FV tests like that live? I see some similar things in `felix/fv/` ?~~  Nevermind, I thought that the `test/` dir was unit tests in this repo.

nelljerram: Yes, that's the one, `test/*.go` in this repo.  Also see the `fv` target in the Makefile.

nelljerram: Please do LMK when the PR is ready for another review.

alexh-tigera: Ok, it took me a while to understand how to implement the FV, but I'm happy with it now. Please have a look.

pkg/controller/gatewayapi/gatewayapi_controller.go
```diff
@@ -89,6 +94,12 @@ func Add(mgr manager.Manager, opts options.ControllerOptions) error {
 		return fmt.Errorf("gatewayapi-controller failed to watch Installation resource: %w", err)
 	}
 
+	go utils.WaitToAddTierWatch(networkpolicy.CalicoTierName, c, opts.K8sClientset, log, tierWatchReady)
```
nelljerram, line 97: How does this work, given that we don't pass tierWatchReady outside of this function - so no other code can depend on whether that tier is ready yet.

pkg/controller/gatewayapi/gatewayapi_controller.go
```diff
@@ -89,6 +94,12 @@ func Add(mgr manager.Manager, opts options.ControllerOptions) error {
...
+	go utils.WaitToAddNetworkPolicyWatches(c, opts.K8sClientset, log, []types.NamespacedName{
+		{Name: gatewayapi.GatewayControllerPolicyName, Namespace: common.TigeraGatewayNamespace},
+		{Name: gatewayapi.GatewayCertgenPolicyName, Namespace: common.TigeraGatewayNamespace},
```
nelljerram, line 100: Could you comment what these watches are for?  IIUC they mean that Reconcile will be called once these policies exist - but why, aka what will then happen next?

pkg/controller/gatewayapi/gatewayapi_controller.go
```diff
@@ -416,12 +427,34 @@ func (r *ReconcileGatewayAPI) Reconcile(ctx context.Context, request reconcile.R
...
+	// rest of the gateway resources.
+	includeV3NetworkPolicy := false
+	if err := r.client.Get(ctx, client.ObjectKey{Name: networkpolicy.CalicoTierName}, &v3.Tier{}); err != nil {
+		if !errors.IsNotFound(err) && !meta.IsNoMatchError(err) {
+			r.status.SetDegraded(operatorv1.ResourceReadError, "Error querying calico-system tier", err, reqLogger)
+			return reconcile.Result{}, err
+		}
+	} else {
+		includeV3NetworkPolicy = true
+	}
```
nelljerram, line 448: So, IIUC, instead of this check, you could add tierWatchReady into the ReconcileGatewayAPI struct and then use r.tierWatchReady.IsReady() - see policyrecommendation_controller.go for an example.

pkg/controller/gatewayapi/gatewayapi_controller.go
```diff
@@ -416,12 +427,34 @@ func (r *ReconcileGatewayAPI) Reconcile(ctx context.Context, request reconcile.R
...
+
+	if includeV3NetworkPolicy {
+		policyComponent := gatewayapi.GatewayPolicy(gatewayConfig)
```
nelljerram, line 451: Does this need to be a separate component, instead of being added into the Objects of GatewayAPIImplementationComponent ?

pkg/render/gatewayapi/gateway_api.go
```diff
@@ -424,6 +435,23 @@ func GatewayAPIImplementationComponent(cfg *GatewayAPIImplementationConfig) rend
 	}
 }
 
+// GatewayPolicy returns a Component that renders the calico-system tier network policies
+// for the gateway control plane. This is rendered separately from the main gateway component
+// so that a failure to create v3 NetworkPolicy resources does not block the rest of the
+// gateway resources. We intentionally do not include a namespace-scoped default deny here
```
nelljerram, line 441: Why should it not block?  Is there a common scenario where we expect to fail to create the network policies?

pkg/render/gatewayapi/gateway_api_test.go
```diff
@@ -1295,4 +1295,53 @@ var _ = Describe("Gateway API rendering tests", func() {
...
+			installation := &operatorv1.InstallationSpec{
+				KubernetesProvider: operatorv1.ProviderOpenShift,
+			}
+			policyComp := GatewayPolicy(&GatewayAPIImplementationConfig{
+				Installation: installation,
+				GatewayAPI:   &operatorv1.GatewayAPI{},
+			})
+
+			objsToCreate, _ := policyComp.Objects()
+			Expect(objsToCreate).To(HaveLen(2))
```
nelljerram, line 1344: It sounds like you intended to add some specific OpenShift testing here.

test/gatewayapi_test.go
```diff
@@ -515,6 +519,220 @@ var _ = Describe("GatewayAPI tests", func() {
 	})
 })
 
+// This Describe exercises the gateway controller's v3 NetworkPolicy rendering path. Unlike the
+// suite above (which runs only the GatewayAPI controller against a cluster without the v3 API
+// available), this one stands up the full Calico + APIServer stack so that the projectcalico.org/v3
+// aggregated API becomes reachable. With v3 present, the controller should render the two
+// calico-system tier allow policies (GatewayPolicy component) and keep them in place via the
+// NetworkPolicy watch.
```
nelljerram, line 527: What does "keep them in place via the NetworkPolicy watch" mean?

test/gatewayapi_test.go
```diff
@@ -515,6 +519,220 @@ var _ = Describe("GatewayAPI tests", func() {
...
+	It("comes up under default-deny and re-creates allow policies on deletion", func() {
+		By("Installing Calico (OSS variant — the FV kind cluster only has OSS images loaded)")
+		// We intentionally leave Variant unset so the installation controller uses the
+		// Calico variant images that are pre-loaded into the kind cluster. Tiers and v3
+		// NetworkPolicies are part of OSS, so the gateway controller's v3 NetworkPolicy
+		// rendering path is fully exercised regardless of variant.
+		operatorDone = createInstallation(c, mgr, shutdownContext, nil)
+		verifyCalicoHasDeployed(c)
+
+		By("Installing the APIServer so the projectcalico.org/v3 aggregated API becomes available")
```
nelljerram, line 626: Is this still necessary?  I thought Casey had made the API server optional now?

test/gatewayapi_test.go
```diff
@@ -515,6 +519,220 @@ var _ = Describe("GatewayAPI tests", func() {
...
+				if np.Spec.Tier != "calico-system" {
+					return fmt.Errorf("NetworkPolicy %s/%s has unexpected tier %q", key.Namespace, key.Name, np.Spec.Tier)
+				}
+			}
+			return nil
+		}, 120*time.Second, 2*time.Second).ShouldNot(HaveOccurred())
+
+		By("Deleting each NetworkPolicy and verifying the watch triggers re-creation")
+		// Both policies are registered in the same WaitToAddNetworkPolicyWatches call; exercising
+		// each one guards against a future refactor accidentally dropping an entry from that list.
```
nelljerram, line 721: OK, this answers one of my earlier answers.  However, what does "against a future refactor" mean?  How could we distinguish between an accidental drop and an intended drop?  And won't it be the future version of the code running, not this version, anyway?

Once this is clarified, please also explain next to the WaitToAdd ... calls.

nelljerram (commented):
A few detailed questions.

Also, after seeing this in detail, I'm now a bit worried about the big picture.

1. Prior to this PR, I believe we do have an overall consistent position on NetworkPolicy - namely:
   - except for tigera-gateway, the operator configures default-deny for Calico namespaces, and exceptions to that that are required for Calico's own operation
   - for tigera-gateway we do not do that because Gateway deployments are also created in the tigera-namespace and we can't predict what NetworkPolicy they would need, if there was a default-deny
   - our users are encouraged to configure default-deny for their own namespace - but that is up to them.

2. However I'm not sure that is adequately documented anywhere.

3. This PR arguably makes it look more like a bug that we don't implement NetworkPolicy to allow specific Gateway flows.

I'm sorry I didn't spot this before, but now I'm wondering it it would be better just to add "tigera-gateway" to the set of excepted namespaces at https://docs.tigera.io/calico-enterprise/latest/network-policy/default-deny

test/gatewayapi_test.go
alexh-tigera: I'll look into this, I haven't rebased since his changes.

alexh-tigera:
> I'm sorry I didn't spot this before, but now I'm wondering it it would be better just to add "tigera-gateway" to the set of excepted namespaces at https://docs.tigera.io/calico-enterprise/latest/network-policy/default-deny

Would that be too permissive to exclude the tigera-gateway ns from default deny? If the operator controls the permissions, we can limit it to DNS + kube-apiserver access. IIRC the concern was that excluding tigera-gateway was too open?

pkg/controller/gatewayapi/gatewayapi_controller.go
alexh-tigera: :+1:   I started by using `apiserver_controller.go` as an example but sounds like that's the wrong approach - if we decide to make the fix in operator rather than excluding tigera-gateway from the default-deny rules, I'll look at `policyrecommendation_controller.go`.

alexh-tigera:
> I'm sorry I didn't spot this before, but now I'm wondering it it would be better just to add "tigera-gateway" to the set of excepted namespaces at https://docs.tigera.io/calico-enterprise/latest/network-policy/default-deny

I talked this over with Seth as well, and he agreed that adding it to the default deny would work. Especially with Walter's work around namespaced gateways, which should also fix the problem because it would push the network policy responsibility to the user.

I'm going to make a docs PR instead of these operator changes.

PR #4608, opened 2026-03-28
Support combined calico/calico image for OSS deployments

caseydavenport:
Teaches the operator to deploy the combined `calico/calico` image for OSS installations. Each component gets dispatched to the right subcommand via `/usr/bin/calico component <name>`, and health probes use the generic `/usr/bin/calico health` binary.

When the combined image is in use, the operator sets:
- `image: calico/calico:$VERSION` for typha, kube-controllers, apiserver, webhooks, goldmane, whisker-backend, csi, key-cert-provisioner, guardian
- `command: ["/usr/bin/calico", "component", "<name>"]` to dispatch to the right component
- Exec probes using `/usr/bin/calico health --port=<port> --type=readiness|liveness` for components with a HealthAggregator

FIPS variants are also supported via `calico/calico-fips`. Enterprise still uses per-component images for now.

Health probe changes:
- Goldmane: exec probe via generic `calico health` command (port 8080)
- Kube-controllers: exec probe via generic `calico health` command (port 9440), replacing the legacy file-based `check-status`

Depends on: https://github.com/projectcalico/calico/pull/12225

```release-note
The operator now deploys OSS Calico components using the combined `calico/calico` image, reducing the number of per-component images required.
```


pkg/components/combined.go
```diff
@@ -0,0 +1,43 @@
...
+		return false
+	}
+	return !installation.Variant.IsEnterprise()
+}
+
+// CombinedCalicoImage returns the combined calico/calico Component to deploy for the given
+// installation, picking the FIPS-tagged variant when FIPS mode is enabled. Callers should only
+// invoke this when UsesCombinedCalicoImage returns true.
+func CombinedCalicoImage(installation *operatorv1.InstallationSpec) Component {
+	if operatorv1.IsFIPSModeEnabled(installation.FIPSMode) {
```
jaderhs, line 43: Could `CombinedCalicoImage` call `UsesCombinedCalicoImage` and either return an error or panic if `UsesCombinedCalicoImage` return false given the requirement of it returning `true`?

Otoh it looks like it gets called on a number of places to set struct field values without checking the result of `UsesCombinedCalicoImage` so maybe this doc comment needs updating unless it refers to non-operator callers.

pkg/render/goldmane/component.go
```diff
@@ -236,25 +239,50 @@ func (c *Component) goldmaneContainer() corev1.Container {
...
 
+	readinessProbe := &corev1.Probe{
+		ProbeHandler: corev1.ProbeHandler{Exec: &corev1.ExecAction{
+			Command: []string{"/health", "-ready"},
+		}},
+		PeriodSeconds: 10,
+	}
+	livenessProbe := &corev1.Probe{
+		ProbeHandler: corev1.ProbeHandler{Exec: &corev1.ExecAction{
+			Command: []string{"/health", "-live"},
```
jaderhs: Could these be removed in favour of always using `CalicoBinaryPath` given that the Goldmane controller reconciler checks for `nil` InstallationSpec and returns before rendering the component and that enterprise does not make use of Goldmane ?

jaderhs (approved):

pkg/components/combined.go
caseydavenport: Good catch — the doc comment was lying about the actual call sites. Tightened the contract in 68bebebeb: `CombinedCalicoImage` now returns `(Component, bool)` so the paired `UsesCombinedCalicoImage` + `CombinedCalicoImage` call pattern collapses into a single call, and callers can't accidentally use the combined image for an installation that shouldn't have it.

pkg/render/goldmane/component.go
caseydavenport: Agreed — Goldmane (and Whisker) are OSS-only so the fallback branch was dead code. Simplified in 68bebebeb: both components now unconditionally use the combined calico binary for their command and probes, and the `useCombinedImage` field is gone. `ResolveImages` returns an error up front if the installation somehow isn't OSS, to make the invariant explicit.

PR #4645, opened 2026-04-07
Improve status manager pod diagnostics

caseydavenport:
## Description

This refactors the status manager's pod health checking to give users more actionable information when things go wrong. Previously, the status manager would only report the first pod error it found, silently skip workloads it couldn't find, and had no way to tell you whether a failing pod was from the current rollout or a previous one.

Changes:

- **Report all pod issues, deduplicated and capped.** Instead of stopping at the first broken pod, the status manager now collects all issues, deduplicates them by root cause, and reports up to 3 distinct reasons per workload (with a count of how many pods share each issue).
- **Detect not-found workloads.** If the status manager is told to watch a DaemonSet/Deployment/StatefulSet/CronJob that doesn't exist, it now reports that as degraded instead of silently skipping it.
- **Surface pending/unschedulable pod reasons.** Pods stuck in Pending now show the scheduler's reason (e.g., "0/3 nodes are available: 3 Insufficient memory") instead of just a generic "not yet scheduled" message at the workload level.
- **Distinguish old vs new revision pods during rollouts.** During a rolling update, failures in new-revision pods are prioritized and old-revision pod failures are annotated with "(old revision)" so you can tell whether the new version is the problem or old pods are just winding down.

Builds on #4644, which added readiness probe detection and crash loop termination context.

```release-note
None
```


pkg/controller/status/status.go
```diff
@@ -53,6 +53,101 @@ const (
...
+
+	// isOldRevision is true if the pod belongs to an old revision during a rollout.
+	isOldRevision bool
+
+	// terminationReason and exitCode provide dedup granularity within an issue type.
+	terminationReason string
+	exitCode          int32
+}
+
+func (p podIssue) key() string {
```
caseydavenport, line 115: Add a comment explaining the purpose of this.

pkg/controller/status/status.go
```diff
@@ -53,6 +53,101 @@ const (
...
+	return fmt.Sprintf("%d:%s:%d", p.issueType, p.terminationReason, p.exitCode)
+}
+
+const maxIssuesPerWorkload = 3
```
caseydavenport: comment on this too.

pkg/controller/status/status.go
```diff
@@ -676,72 +774,207 @@ func (m *statusManager) removeTigeraStatus() {
...
-				} else {
-					msg += fmt.Sprintf(" (%s, exit code %d)", lt.Reason, lt.ExitCode)
+// diagnoseContainers checks a list of container statuses for known error states and
+// returns a podIssue for each one found.
+func diagnoseContainers(p corev1.Pod, statuses []corev1.ContainerStatus, oldRevision bool) []podIssue {
+	var issues []podIssue
+	for _, c := range statuses {
+		if c.State.Waiting != nil {
+			switch c.State.Waiting.Reason {
+			case "CrashLoopBackOff":
```
caseydavenport: There must be constants for these, surely. If not we should define them.

pkg/controller/status/status.go
```diff
@@ -676,72 +774,207 @@ func (m *statusManager) removeTigeraStatus() {
...
-	if c.State.Terminated != nil {
-		if c.State.Terminated.Reason == terminationReasonError {
-			return fmt.Sprintf("Pod %s/%s has terminated container: %s", p.Namespace, p.Name, c.Name)
+	return issues
+}
+
+// podMatchesRevision checks whether a pod belongs to the given revision.
+// Works for DaemonSets and StatefulSets (controller-revision-hash label)
+// and Deployments (pod-template-hash label).
+func podMatchesRevision(p corev1.Pod, currentRevision string) bool {
```
caseydavenport, line 963: Do we have a good way to test this is working correctly?

pkg/controller/status/status.go
```diff
@@ -676,72 +774,207 @@ func (m *statusManager) removeTigeraStatus() {
...
+// currentDeploymentRevision returns the pod-template-hash of the active ReplicaSet
+// for the given Deployment. Returns empty string if it cannot be determined.
+func (m *statusManager) currentDeploymentRevision(dep *appsv1.Deployment) string {
+	rsList := &appsv1.ReplicaSetList{}
+	s, err := metav1.LabelSelectorAsMap(dep.Spec.Selector)
+	if err != nil {
+		log.WithValues("reason", err).Info("Failed to parse deployment selector for revision lookup")
+		return ""
+	}
+	err = m.client.List(context.TODO(), rsList, client.MatchingLabels(s), client.InNamespace(dep.Namespace))
```
caseydavenport: Is this going to generate a List() call against the API server every iteration? That feels like too much load.

pkg/controller/status/status.go
```diff
@@ -676,72 +774,207 @@ func (m *statusManager) removeTigeraStatus() {
...
 		}
 	}
 	return ""
 }
 
+// currentDaemonSetRevision returns the controller-revision-hash of the most recent
+// ControllerRevision for the given DaemonSet. Returns empty string if it cannot be determined.
+func (m *statusManager) currentDaemonSetRevision(ds *appsv1.DaemonSet) string {
+	revList := &appsv1.ControllerRevisionList{}
+	err := m.client.List(context.TODO(), revList, client.InNamespace(ds.Namespace))
```
caseydavenport: Same here - we probably need to drive API calls made here off of an informer cache.

pkg/controller/status/status.go
```diff
@@ -676,72 +774,207 @@ func (m *statusManager) removeTigeraStatus() {
...
+		}
+		if rev.Revision > maxRevision {
+			maxRevision = rev.Revision
+			currentHash = rev.Labels[appsv1.ControllerRevisionHashLabelKey]
+		}
+	}
+	return currentHash
+}
+
+func isOwnedBy(refs []metav1.OwnerReference, uid types.UID) bool {
```
caseydavenport: is there not a common utility already for this?

pkg/controller/status/status.go
caseydavenport: Done.

pkg/controller/status/status.go
caseydavenport: Done.

pkg/controller/status/status.go
caseydavenport: Defined local constants - waitingCrashLoopBackOff, waitingImagePullBackOff, waitingErrImagePull. These are free-form strings from the container runtime, no k8s constants exist for them.

pkg/controller/status/status.go
caseydavenport: The rollout integration test exercises this indirectly, but it uses a fake client so it doesn't validate what real Kubernetes actually sets on pods. I'll verify against a live cluster before merging - bringing one up now.

pkg/controller/status/status.go
caseydavenport: This goes through the controller-runtime informer cache (mgr.GetClient() is cache-backed for reads), not directly to the API server. Also only called when the Deployment is unhealthy (the early-continue check gates it). Added a comment noting both.

pkg/controller/status/status.go
caseydavenport: Same as above - cache-backed, only on unhealthy path. Added comment.

pkg/controller/status/status.go
caseydavenport: Switched to metav1.GetControllerOf from the apimachinery package. Removed the custom isOwnedBy helper.

pkg/controller/status/status.go
caseydavenport: Verified on a live kind cluster. Deployment ReplicaSets have `pod-template-hash` labels and `controller: true` owner references as expected. DaemonSet pods have `controller-revision-hash` labels matching ControllerRevision objects. The code correctly reads these.

pkg/controller/status/status.go
```diff
@@ -51,8 +51,114 @@ const (
...
+		if idx, ok := seen[k]; ok {
+			groups[idx].count++
+			continue
+		}
+		seen[k] = len(groups)
+		groups = append(groups, issueGroup{issue: iss, count: 1})
+	}
+
+	// Cap at maxIssuesPerWorkload unique reasons.
+	if len(groups) > maxIssuesPerWorkload {
```
caseydavenport:
```suggestion
	if len(groups) > maxIssuesToReport {
```

pkg/controller/status/status.go
```diff
@@ -599,21 +709,20 @@ func (m *statusManager) syncState() {
...
-				failing = append(failing, f)
-			}
-		} else {
-			log.WithValues("reason", err, "statefuleset", depnn).Info("Failed to check for failing pods")
-			continue
-		}
+		issues := m.diagnosePods(ss.Spec.Selector, ss.Namespace, ss.Status.UpdateRevision)
+		f, p := summarizeIssues(issues)
+		failing = append(failing, f...)
+		progressing = append(progressing, p...)
```
caseydavenport: This pattern is repeated a bit and I think we can probably cut down on the repetition and also make this harder to mess up in the future if we introduce a better interface for tracking this.

Would probably be better if we just maintained a `[]issue` and then summarized once before returning?

pkg/controller/status/status.go
```diff
@@ -676,72 +785,201 @@ func (m *statusManager) removeTigeraStatus() {
...
 	err = m.client.List(context.TODO(), &l, client.MatchingLabels(s), client.InNamespace(namespace))
 	if err != nil {
-		return "", err
+		log.WithValues("reason", err).Info("Failed to list pods for diagnosis")
+		return nil
 	}
+
+	var issues []podIssue
 	for _, p := range l.Items {
+		oldRevision := currentRevision != "" && !podMatchesRevision(p, currentRevision)
```
caseydavenport:
```suggestion
		isOldRevision := currentRevision != "" && !podMatchesRevision(p, currentRevision)
```

pkg/controller/status/status.go
```diff
@@ -676,72 +785,201 @@ func (m *statusManager) removeTigeraStatus() {
...
-				return msg, nil
-			}
+		if iss := diagnoseContainers(p, p.Status.ContainerStatuses, oldRevision); len(iss) > 0 {
+			issues = append(issues, iss...)
+			continue
 		}
 
-		// If none of the container-level checks matched, check if the pod is running but
-		// not passing readiness checks.
+		// Running but not passing readiness checks.
```
caseydavenport:
```suggestion
		// Running - check if passing readiness checks.
```

pkg/controller/status/status.go
```diff
@@ -676,72 +785,201 @@ func (m *statusManager) removeTigeraStatus() {
...
+						isOldRevision: oldRevision,
+					})
+					break
+				}
+			}
+			continue
+		}
+
+		// Pending pod - check for scheduling failures.
+		if p.Status.Phase == corev1.PodPending {
```
caseydavenport, line 860: nit - for clarity we should do this in chronological order - i.e., pending comes before running, so let's put this above.

pkg/controller/status/status.go
```diff
@@ -676,72 +785,201 @@ func (m *statusManager) removeTigeraStatus() {
...
+					message:           msg,
+					isOldRevision:     oldRevision,
+					terminationReason: termReason,
+					exitCode:          exitCode,
+				})
+			case waitingImagePullBackOff, waitingErrImagePull:
+				issues = append(issues, podIssue{
+					severity:      severityFailing,
+					issueType:     issueImagePull,
+					message:       fmt.Sprintf("Pod %s/%s failed to pull container image for: %s", p.Namespace, p.Name, c.Name),
```
caseydavenport: can we extract the actual image here to display?

pkg/controller/status/status.go
```diff
@@ -676,72 +785,201 @@ func (m *statusManager) removeTigeraStatus() {
...
+// Only called when the Deployment is unhealthy. The List call goes through
+// the controller-runtime informer cache, not directly to the API server.
+func (m *statusManager) currentDeploymentRevision(dep *appsv1.Deployment) string {
+	rsList := &appsv1.ReplicaSetList{}
+	s, err := metav1.LabelSelectorAsMap(dep.Spec.Selector)
+	if err != nil {
+		log.WithValues("reason", err).Info("Failed to parse deployment selector for revision lookup")
+		return ""
+	}
+	err = m.client.List(context.TODO(), rsList, client.MatchingLabels(s), client.InNamespace(dep.Namespace))
```
caseydavenport: Can use context.WithTimeout here.

pkg/controller/status/status.go
```diff
@@ -676,72 +785,201 @@ func (m *statusManager) removeTigeraStatus() {
...
+	if err != nil {
+		log.WithValues("reason", err).Info("Failed to list ReplicaSets for revision lookup")
+		return ""
+	}
+
+	for _, rs := range rsList.Items {
+		if ref := metav1.GetControllerOf(&rs); ref == nil || ref.UID != dep.UID {
```
caseydavenport, line 1004: Add a brief comment explaining this branch.

pkg/controller/status/status.go
```diff
@@ -676,72 +785,201 @@ func (m *statusManager) removeTigeraStatus() {
...
 		}
 	}
 	return ""
 }
 
+// currentDaemonSetRevision returns the controller-revision-hash of the most recent
+// ControllerRevision for the given DaemonSet. Returns empty string if it cannot be determined.
+// Only called when the DaemonSet is unhealthy. The List call goes through
+// the controller-runtime informer cache, not directly to the API server.
+func (m *statusManager) currentDaemonSetRevision(ds *appsv1.DaemonSet) string {
```
caseydavenport, line 1023: Same comments here.

pkg/controller/status/status.go
caseydavenport: Done.

pkg/controller/status/status.go
caseydavenport: Refactored. `diagnosePods` now stamps each issue with a workload identifier; `syncState` collects all issues into a single slice and calls `summarizeIssues` once at the end. `summarizeIssues` groups by workload so the per-workload cap is preserved.

pkg/controller/status/status.go
caseydavenport: Done.

pkg/controller/status/status.go
caseydavenport: Done.

pkg/controller/status/status.go
caseydavenport: Done - moved the pending block above the running block.

pkg/controller/status/status.go
caseydavenport: Done - now includes `c.Image` in the message.

pkg/controller/status/status.go
caseydavenport: Done - 10s timeout via `context.WithTimeout`.

pkg/controller/status/status.go
caseydavenport: Done.

pkg/controller/status/status.go
caseydavenport: Done - same treatment (`context.WithTimeout` + filter comment).

pkg/controller/status/status.go
```diff
@@ -676,72 +798,223 @@ func (m *statusManager) removeTigeraStatus() {
...
+				issueType:     issuePending,
+				message:       msg,
+				isOldRevision: isOldRevision,
+			})
+			continue
 		}
 
-		// If none of the container-level checks matched, check if the pod is running but
-		// not passing readiness checks.
+		// Running - check if passing readiness checks.
```
tmjd: Do we need some kind of default for checking p.Status.Phase != PodRunning? That way we'd catch any other Phases we haven't handled above (maybe because new ones are added in the future).

caseydavenport: Yeah, I don't think it's worth it. The phases haven't changed in a decade, and a `!= PodRunning` default would actually mis-fire on `PodSucceeded` (e.g. CronJob pods that completed normally).

rene-dekker:
**`currentDeploymentRevision` likely returns the wrong revision mid-rollout**

```go
for _, rs := range rsList.Items {
    if ref := metav1.GetControllerOf(&rs); ref == nil || ref.UID != dep.UID {
        continue
    }
    if rs.Status.Replicas > 0 {
        return rs.Labels[appsv1.DefaultDeploymentUniqueLabelKey]
    }
}
```

This returns the **first** owned ReplicaSet (in arbitrary informer-cache list order) that has `Replicas > 0`. During a rolling update **both** the new and old ReplicaSets have `Replicas > 0`, so this can return the *old* hash — which inverts the whole feature: new-revision pods get tagged `(old revision)` and old-revision pods are treated as current. List order from the cache isn't guaranteed, so it's also non-deterministic.

The canonical "current" ReplicaSet is the one with the highest `deployment.kubernetes.io/revision` annotation, not just any RS with replicas. Suggest selecting by max revision annotation — mirroring how `currentDaemonSetRevision` correctly picks the max `rev.Revision`.

Note the unit test passes only because the old RS is set to `Replicas: 0`, which is exactly the case that doesn't exercise this path. Worth adding a test where the old RS still has `Replicas > 0` during the rollout.

tmjd:
I'm ready to approve if you can address Rene's comment.

pkg/controller/status/status.go
```diff
@@ -57,6 +58,16 @@ const (
...
+
+	// deploymentRevisionAnnotation records the revision number of a ReplicaSet
+	// within its owning Deployment. Set by the deployment controller; Kubernetes
+	// does not export a constant for it.
+	deploymentRevisionAnnotation = "deployment.kubernetes.io/revision"
+
+	// readinessGracePeriod is how long a running pod may stay unready before we
+	// consider it failing. This avoids tripping Degraded for pods that are only
+	// briefly unready during a rolling update.
+	readinessGracePeriod = 60 * time.Second
```
tmjd, line 70: Shouldn't we come up with the grace period based on the readinessProbe delay/timeout/thresholds?

tmjd (approved):

PR #4651, opened 2026-04-07
Show most relevant message in TigeraStatus kubectl output

caseydavenport:
Adds a top-level message to TigeraStatus and points the `kubectl get tigerastatus` MESSAGE column at it instead of the Degraded condition message. Previously the column only showed content when the component was degraded, so warnings and rollout detail needed `-o yaml` to see.

The message picks the most actionable thing available, in order:

1. the degraded message, if degraded
2. the progressing message, if progressing
3. any warnings
4. empty

The degraded message keeps a column of its own under `-o wide`, so it stays available on a version skew in either direction.

Example output when a component has a warning but is otherwise healthy:

```
NAME     AVAILABLE   PROGRESSING   DEGRADED   SINCE   MESSAGE
calico   True        False         False      5m      DaemonSet "calico-system/calico-node" has the unsupported ignore annotation; the operator is not managing this resource
```

Example output during rollout:

```
NAME     AVAILABLE   PROGRESSING   DEGRADED   SINCE   MESSAGE
calico   False       True          False      30s     DaemonSet "calico-system/calico-node" update is rolling out (2 out of 3 updated)
```

```release-note
None
```


pkg/controller/status/status.go
```diff
@@ -761,8 +1055,11 @@ func (m *statusManager) set(retry bool, conditions ...operator.TigeraStatusCondi
 		}
 	}
 
+	// Set the top-level Message field based on the most actionable condition.
+	ts.Status.Message = m.statusMessage(ts.Status.Conditions)
+
 	// If nothing has changed, we don't need to update in the API.
-	if reflect.DeepEqual(ts.Status.Conditions, old.Status.Conditions) {
+	if reflect.DeepEqual(ts.Status.Conditions, old.Status.Conditions) && ts.Status.Message == old.Status.Message {
```
electricjesus: this might loop if the CRD is older than the operator. `.status.message` is new. The schema is structural and does not keep unknown fields, so an old CRD drops it on write. `old.Status.Message` then reads back empty each time. The guard never holds, so we `Status().Update` every 5s tick, per component. Nothing logs it.

Bites `manage-crds=false`, and anyone who applies CRDs out of band.

Could we keep the last message we wrote on the manager, and check that instead of the read-back?

pkg/controller/status/status.go
```diff
@@ -870,6 +1167,25 @@ func (m *statusManager) degradedMessage() string {
 	return strings.Join(msgs, "\n")
 }
 
+// statusMessage returns the most actionable message for the top-level Message field,
+// based on the current conditions. Priority: degraded > progressing > warnings > empty.
+func (m *statusManager) statusMessage(conditions []operator.TigeraStatusCondition) string {
```
electricjesus, line 1317: nothing covers this yet, as far as I can tell. The precedence order is the whole point of the field, so a small table would earn its keep. The fake client in `status_test.go` already does the set-then-read-back dance, so it's cheap:

```go
sm.SetWarning("w", "ignore annotation set")
sm.updateStatus()
Expect(readStatus(client).Status.Message).To(Equal("ignore annotation set"))
```

The pruning case in my other comment needs a real apiserver. So envtest, pointed at a CRD with no `message` field. Same shape as `pkg/controller/installation/cel_validation_test.go`, and `make ut` already exports `KUBEBUILDER_ASSETS`, so it runs in the normal lane.

api/v1/tigerastatus_types.go
```diff
@@ -45,7 +51,7 @@ type TigeraStatusStatus struct {
 // +kubebuilder:printcolumn:name="Progressing",type="string",JSONPath=".status.conditions[?(@.type=='Progressing')].status",description="Whether the component is processing changes."
 // +kubebuilder:printcolumn:name="Degraded",type="string",JSONPath=".status.conditions[?(@.type=='Degraded')].status",description="Whether the component is degraded."
 // +kubebuilder:printcolumn:name="Since",type="date",JSONPath=".status.conditions[?(@.type=='Available')].lastTransitionTime",description="The time the component's Available status last changed."
-// +kubebuilder:printcolumn:name="Message",type="string",JSONPath=".status.conditions[?(@.type=='Degraded')].message",description="Error message when the component is degraded.",priority=0
+// +kubebuilder:printcolumn:name="Message",type="string",JSONPath=".status.message",description="Human-readable summary of the component's current state."
```
electricjesus, line 54: tiny: this column goes blank on a version skew, both ways. An old CRD drops the field. An old operator never sets it. Today the degraded message always shows up.

Worth keeping the old jsonPath as a `priority=1` column, under a new name? Then `-o wide` still has it.

pkg/controller/status/status.go
```diff
@@ -870,6 +1167,25 @@ func (m *statusManager) degradedMessage() string {
...
+	for _, c := range conditions {
+		if c.Type == operator.ComponentDegraded && c.Status == operator.ConditionTrue {
+			return c.Message
+		}
+	}
+	for _, c := range conditions {
+		if c.Type == operator.ComponentProgressing && c.Status == operator.ConditionTrue {
+			return c.Message
```
electricjesus: nit, not blocking: progressing messages get joined with `\n`. So the column picks up line breaks on a plain rollout. Before, that only hit when degraded, which is rarer.

```suggestion
			return strings.ReplaceAll(c.Message, "\n", "; ")
```

Applied it here and `go test ./pkg/controller/status/...` is still green. Same for the degraded branch above, if you agree.

electricjesus (commented):
looks pretty solid to me! a few comments inline.

Two process bits. It needs a rebase, it's conflicting on master now, and `createOrUpdateObject` grew an `installationSpec` arg since you branched. CI has also never run here, only the CLA check, so none of this has a green signal yet.

Also worth saying out loud for anyone else reading: GitHub shows 8 files because #4649 was squash merged, so the merge base predates it. The change here is really the one commit.

pkg/controller/status/status.go
caseydavenport: Yep, and the read-back can never tell us. Switched the guard to compare against the last message we wrote, tracked on the manager and only updated after a successful write. Also cleared on delete so a recreated status gets the message again.

pkg/controller/status/status.go
caseydavenport: Added a table test for the precedence order, including a degraded condition that's present but False so the positive cases can't pass for the wrong reason.

For the pruning case I went with the fake client instead of envtest - after the first write I clear the message on the stored object and assert the resource version doesn't move on the next tick. That fails against the old guard and passes with the new one, so it pins the actual mechanism without needing a real apiserver.

api/v1/tigerastatus_types.go
caseydavenport: Yeah, the skew cuts both ways. Kept the degraded jsonPath as an `Error` column at `priority=1`, so `-o wide` still has it either way.

pkg/controller/status/status.go
caseydavenport: Yep, took this. Applied it to the progressing and degraded branches through a small helper so both get it.

pkg/controller/status/status.go
electricjesus: that does it. i went looking for a hole on restart. `lastMessage` starts empty, but the object can already carry one. couldn't make it bite though. the warning text rides along in the condition messages too, so `DeepEqual` catches it and you write anyway. the clear on delete is a nice touch, i'd have missed that.

pkg/controller/status/status.go
electricjesus: fake client plus a resource version check beats what i suggested. no envtest assets to keep alive. the degraded entry that's present but False is the part i like, that's the one that would have hidden a bad pass..

api/v1/tigerastatus_types.go
electricjesus: yep, that covers it. `-o wide` keeps the message on both sides of the skew.

pkg/controller/status/status.go
electricjesus: the helper reads better than my inline version. i see `warningMessageLocked()` stayed out of it and i think that's right, it already joins with `; `.

electricjesus (approved):
LGTM!

PR #4675, opened 2026-04-11
Delay marking degraded due to readiness for 60s

caseydavenport:
## Description

So that rolling updates don't trigger this accidentally.

## Release Note

```release-note
TBD
```

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


pkg/controller/status/status.go
```diff
@@ -704,10 +704,12 @@ func (m *statusManager) podsFailing(selector *metav1.LabelSelector, namespace st
 		}
 
 		// If none of the container-level checks matched, check if the pod is running but
-		// not passing readiness checks.
+		// not passing readiness checks. We only note a pod as failing if it has been
+		// unready for over a minute, to minimize false positives.
 		if p.Status.Phase == corev1.PodRunning {
 			for _, cond := range p.Status.Conditions {
-				if cond.Type == corev1.ContainersReady && cond.Status == corev1.ConditionFalse {
+				old := time.Now().Sub(cond.LastTransitionTime.Time) > 60*time.Second
```
electricjesus: A few nits on this line, none blocking:

nit (worth doing, may red golangci-lint): `time.Now().Sub(x)` trips gosimple S1012. Use `time.Since(cond.LastTransitionTime.Time)`.

nit: `old` is computed for every condition in the loop even though it's only used when the type and status match. Moving it inside the `if` reads cleaner and skips the clock read on conditions that don't match.

nit: the `60 * time.Second` is a magic number. A named const like `readinessGracePeriod` would make the intent obvious and easier to tune later.

nit: if a `ContainersReady=False` condition ever carries an unset `LastTransitionTime`, `time.Now().Sub(zero)` is enormous and the pod gets flagged immediately, skipping the grace window. Kubelet sets it in practice so this is mostly theoretical, but a zero-check would close it.

electricjesus (approved):
LGTM. Correct and minimal, and it does what the title says: a pod that's only briefly unready during a rollout won't trip Degraded, but one that stays unready past 60s still will. The status manager's 5s ticker re-runs the check, so nothing gets stuck waiting on a requeue once a pod crosses the threshold.

Approving. A few nits inline, none blocking.

suggestion: the author checklist has "Tests for change" unchecked, and this is a behavioral change that's easy to regress. A small `podsFailing` unit test (one condition unready for ~10s, one for ~90s) would lock the 60s boundary in. Not blocking.

PR #4676, opened 2026-04-11
Grant calico-manager read access to UISettings for ui-apis

caseydavenport:
ui-apis now serves UISettings and UISettingsGroups reads on behalf of users via Voltron. The calico-manager SA (which runs the ui-apis sidecar) needs broad read access to these resources so it can fetch them, then enforce per-group RBAC via SubjectAccessReviews before returning results to the user.

This adds `get`/`list`/`watch` on `uisettings`, `uisettingsgroups`, and `uisettingsgroups/data` to the calico-manager ClusterRole. The duplicate `uisettingsgroups` list-only rule is folded into the new rule.

Related: https://github.com/tigera/calico-private/pull/11457

```release-note
None
```

### DIFF
--- a/pkg/render/manager.go
+++ b/pkg/render/manager.go
@@ -939,10 +939,20 @@ func managerClusterRole(managedCluster bool, kubernetesProvider operatorv1.Provi
 					"stagednetworkpolicies",
 					"tier.stagednetworkpolicies",
 					"stagedkubernetesnetworkpolicies",
-					"uisettings",
-					"uisettingsgroups",
 				},
 				Verbs: []string{"list"},
+			},
+			{
+				// ui-apis needs broad read access to UISettings and UISettingsGroups to serve
+				// requests on behalf of users. It performs SubjectAccessReviews to enforce
+				// per-group RBAC before returning results.
+				APIGroups: []string{"projectcalico.org"},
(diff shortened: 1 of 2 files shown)


pkg/render/manager.go
```diff
@@ -947,10 +939,20 @@ func managerClusterRole(managedCluster bool, kubernetesProvider operatorv1.Provi
...
+				},
+				Verbs: []string{"list"},
+			},
+			{
+				// ui-apis needs broad read access to UISettings and UISettingsGroups to serve
+				// requests on behalf of users. It performs SubjectAccessReviews to enforce
+				// per-group RBAC before returning results.
+				APIGroups: []string{"projectcalico.org"},
+				Resources: []string{
 					"uisettings",
```
pasanw, line 951: Is this necessary? Based on our discussions I thought that we don't expect any `uisettings` RBAC checks to be performed by ui-apis

caseydavenport: ui-apis DOES need to list uisettings, in order to return them to the user!

But the RBAC provided to users never includes "uisettings" directly.

pasanw: Ah, of course :facepalm:

pasanw (approved):

PR #4690, opened 2026-04-14
Gatewayapi Namespaced Mode

radixo:
## Description

Replace the previous Gateway-API install — which ran an Envoy Gateway controller in `tigera-gateway` and deployed all proxy workloads in that same namespace — with a single envoy-gateway controller in `calico-system` running with `deploy.type=GatewayNamespace`, so proxy workloads land in each Gateway's own namespace.

This is a **breaking change** for clusters running the legacy install. Existing `Gateway` CRs do not need edits — `tigera-gateway-class` and its controllerName are preserved — but proxy Pods, their `Service`s and LoadBalancer addresses are recreated in each Gateway's own namespace on first reconcile after upgrade. Anything pinned to `tigera-gateway` (NetworkPolicies, monitoring, RBAC, external DNS) must follow.

- **Single controller in `calico-system`** with `controllerName=gateway.envoyproxy.io/gatewayclass-controller` (chart default) and `deploy.type=GatewayNamespace`. ControllerName + GatewayClass name are deliberately reused from the legacy install so existing `Gateway` CRs continue to be claimed.
- **Auto-provisioned `tigera-gateway-class`** GatewayClass + EnvoyProxy. Users can declare additional classes via `GatewayAPI.Spec.GatewayClasses`; all classes target the single controller.
- **Embed `gateway-helm.tgz`** and render at runtime via the Helm SDK; result is cached per process via `sync.Once`. Replaces the previous pre-rendered YAML.
- **Per-namespace Enterprise resources** for each Gateway-hosting namespace: `waf-http-filter` SA + per-namespace `waf-http-filter-gateway-resources` RoleBinding (least-privilege Gateway-API reads), `tigera-operator-secrets` RoleBinding, `tigera-pull-secret` copy. Cluster-scoped perms (licensekeys, tokenreviews) go through a single shared `waf-http-filter-gateway-namespaces` ClusterRoleBinding whose Subjects list is recomputed each reconcile.
- **Reserved-namespace guard** (`calico-system`, `tigera-operator`): the operator does not create or delete the shared `tigera-operator-secrets` RoleBinding or `tigera-pull-secret` copy in those namespaces — the core Installation controller owns them.
- **Per-namespace cleanup** when a namespace no longer hosts a Gateway: pull Secret → SA + RoleBinding → `tigera-operator-secrets` RoleBinding. That RoleBinding is what grants the operator Secret-delete perms, so reversing the order yields a 403 and aborts the reconcile.
- **Explicit upgrade-cleanup queue** for the legacy `tigera-gateway` install: controller Deployment/Service/SAs/ConfigMap, certgen Job + RBAC, namespaced Role/RoleBinding, copied pull Secrets, `tigera-operator-secrets` RoleBinding, `envoy-gateway-topology-injector.tigera-gateway` MWC, the orphaned `waf-http-filter-cluster-scoped` and `waf-http-filter-gateway-resources` ClusterRoleBindings, and the deprecated combined `waf-http-filter` ClusterRole/ClusterRoleBinding. Pull Secrets are queued before `tigera-operator-secrets`. **The `tigera-gateway` Namespace itself is intentionally not queued** — users may have placed their own resources in it.
- **v3 NetworkPolicy** (`calico-system.envoy-gateway`) under the `calico-system` tier to keep the controller + certgen Job working under default-deny. Selector covers both `app.kubernetes.io/name=gateway-helm` and `app=certgen` (the chart applies different labels to the certgen Job vs its pod template). Egress: DNS + kube-apiserver, then Pass. Ingress: 9443 (topology-injector webhook), 18000-18002 (xDS), 19001 (metrics).
- **Tests:** UT coverage for OSS + Enterprise renders, GatewayClass + EnvoyProxy customisation, per-namespace lifecycle (create + cleanup), reserved-namespace guards, Secret-before-RoleBinding ordering, and the explicit legacy-cleanup queue (with the `tigera-gateway` Namespace asserted *not* in the delete list). FV coverage: deploys the controller in `calico-system` and asserts nothing lands in `tigera-gateway`, provisions and cleans up per-namespace resources, GatewayClass + EnvoyProxy cleanup, custom EnvoyProxy watch, l7-log-collector owning-gateway env wiring, custom EnvoyGateway ConfigMap.

### Security

The Enterprise per-namespace render copies `tigera-pull-secret` into every namespace that hosts a Gateway, so permissive RBAC on those namespaces can expose the pull secret. Reserved namespaces are excluded from create and delete of the shared resources, so the operator does not clobber core-Installation-owned secrets.

### Upgrade / compatibility

In-place upgrade. The single controller in `calico-system` claims all `tigera-gateway-class` Gateways unchanged. Proxy Pods, their Services, and LoadBalancer addresses are recreated in each Gateway's own namespace on first reconcile, so any cluster setting pinned to `tigera-gateway` (NetworkPolicies, monitoring, RBAC, external DNS) must be repointed to the Gateway's own namespace. The legacy controller install is removed automatically; the `tigera-gateway` Namespace itself is preserved in case it holds user resources.

Calico-private operator RBAC update required: `envoy-gateway-topology-injector.calico-system` added to the `mutatingwebhookconfigurations` `resourceNames` list with `update` and `delete` verbs (the legacy `.tigera-gateway` entry is retained with `delete` so the upgrade-cleanup can reap it).

## Release Note

```release-note
**Breaking change.** The Calico Ingress Gateway controller now runs in
`calico-system` and provisions proxy workloads in the Gateway's own namespace
(`deploy.type=GatewayNamespace`). Existing `tigera-gateway-class` Gateways
keep working without CR edits, but their proxy Pods are relaunched in the
Gateway's own namespace; any NetworkPolicy, monitoring, RBAC or external DNS
previously pinned to `tigera-gateway` must be repointed to the Gateway's own
namespace. The legacy controller install is removed automatically on upgrade;
the `tigera-gateway` Namespace itself is left in place in case it holds user
resources.
```

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


electricjesus (commented):
Good work on the runtime Helm rendering migration — dropping 53K lines of pre-rendered YAML is a huge win. The GatewayNamespace mode looks solid with good test coverage. A few observations below.

pkg/render/gatewayapi/gateway_api.go
```diff
@@ -477,31 +494,48 @@ func (pr *gatewayAPIImplementationComponent) Ready() bool {
...
 
-	// First create the namespace.  We take the name from the read resources, but otherwise
-	// follow our own pattern for namespace creation.
+	// Render the opposite mode to find resources that only exist in that mode,
+	// so we can clean them up when switching between deployment modes.
+	oppositeMode := operatorv1.GatewayDeploymentModeGatewayNamespace
+	if deploymentMode == operatorv1.GatewayDeploymentModeGatewayNamespace {
+		oppositeMode = operatorv1.GatewayDeploymentModeControllerNamespace
+	}
+	oppositeResources, err := renderChart(pr.cfg.Scheme, oppositeMode)
```
electricjesus: **Perf concern:** `renderChart()` is called twice here in `Objects()` (current mode + opposite mode), and also once each in `K8SGatewayAPICRDs()` and `GatewayAPICRDs()` which run earlier in the same reconcile. That's 4 Helm template renders per reconcile — each one loads the .tgz, runs the Helm SDK, and parses the output.

The old code parsed static YAML once and cached it. Consider caching the render results (keyed by deployment mode) so you're not re-rendering the same chart multiple times per reconcile cycle. A simple `sync.Map` or even two package-level vars (one per mode) would work.

pkg/render/gatewayapi/gateway_api.go
```diff
@@ -477,31 +494,48 @@ func (pr *gatewayAPIImplementationComponent) Ready() bool {
...
 
 func (pr *gatewayAPIImplementationComponent) Objects() ([]client.Object, []client.Object) {
-	resources := GatewayAPIResources()
+	deploymentMode := operatorv1.GatewayDeploymentModeControllerNamespace
+	if pr.cfg.GatewayAPI.Spec.GatewayDeploymentMode != nil {
+		deploymentMode = *pr.cfg.GatewayAPI.Spec.GatewayDeploymentMode
+	}
+	resources, err := renderChart(pr.cfg.Scheme, deploymentMode)
+	if err != nil {
+		panic(fmt.Sprintf("failed to render gateway-helm chart: %v", err))
```
electricjesus: This `panic()` was fine when parsing static embedded YAML (couldn't realistically fail), but now it's running the Helm SDK at runtime — chart loading, value merging, template rendering. These can fail for reasons beyond your control (e.g., a future chart version introduces an incompatible template). Consider returning an error from `Objects()` or at least logging before panicking, so the failure is diagnosable in production.

pkg/render/gatewayapi/gateway_api.go
```diff
@@ -677,6 +706,33 @@ func (pr *gatewayAPIImplementationComponent) Objects() ([]client.Object, []clien
...
+			objs = append(objs, secret.ToRuntimeObjects(secret.CopyToNamespace(ns, pr.cfg.PullSecrets...)...)...)
+		}
+
+		// Clean up resources for namespaces that no longer have Gateway resources.
+		if pr.cfg.CurrentGatewayNamespaces != nil {
+			currentNS := set.New(pr.cfg.GatewayNamespaces...)
+			for _, ns := range pr.cfg.CurrentGatewayNamespaces.UnsortedList() {
+				if !currentNS.Has(ns) {
+					objsToDelete = append(objsToDelete,
+						pr.gatewayNamespaceSA(ns),
```
electricjesus, line 758: The stale namespace cleanup deletes SA and CRB but not the pull secrets that were copied to that namespace (line 718 copies them on creation). These will be orphaned when a Gateway is removed from a namespace. Should this also append the pull secrets to `objsToDelete`?

pkg/controller/gatewayapi/gatewayapi_controller.go
```diff
@@ -275,13 +298,18 @@ func (r *ReconcileGatewayAPI) Reconcile(ctx context.Context, request reconcile.R
...
 	gatewayConfig := &gatewayapi.GatewayAPIImplementationConfig{
+		Scheme:                r.scheme,
 		Installation:          installationSpec,
 		PullSecrets:           pullSecrets,
 		GatewayAPI:            gatewayAPI,
 		CustomEnvoyProxies:    make(map[string]*envoyapi.EnvoyProxy),
 		CurrentGatewayClasses: set.New[string](),
 	}
 
+	if gatewayAPI.Spec.GatewayDeploymentMode == nil {
```
electricjesus, line 310: Nit: The CRD already has `+kubebuilder:default=ControllerNamespace`, so any persisted GatewayAPI resource will have this field populated by the API server. This runtime defaulting only matters for in-memory objects that were never persisted (tests?). Not a problem, just noting the redundancy — if the CRD default is the source of truth, a comment here explaining *why* you also default in code would help future readers.

radixo:
> Good work on the runtime Helm rendering migration — dropping 53K lines of pre-rendered YAML is a huge win. The GatewayNamespace mode looks solid with good test coverage. A few observations below.

All good catches man, all sorted

pkg/render/gatewayapi/gateway_api.go
```diff
@@ -677,6 +736,33 @@ func (pr *gatewayAPIImplementationComponent) Objects() ([]client.Object, []clien
...
+						pr.gatewayNamespaceCRB(ns),
+					)
+				}
```
electricjesus, line 761: Pull secrets are copied to each gateway namespace on creation (line 749) but aren't cleaned up here when a gateway is removed. This will leave orphaned secrets accumulating over time.

```suggestion
						pr.gatewayNamespaceSA(ns),
						pr.gatewayNamespaceCRB(ns),
					)
					objsToDelete = append(objsToDelete, secret.ToRuntimeObjects(secret.CopyToNamespace(ns, pr.cfg.PullSecrets...)...)...)
```

The matching test ("should clean up stale per-namespace resources when Gateways are removed") should also add a pull secret to the config and assert it appears in `objsToDelete`:

```go
		// Add pull secret to config.
		pullSecret := &corev1.Secret{
			ObjectMeta: metav1.ObjectMeta{Name: "tigera-pull-secret", Namespace: "tigera-operator"},
			Data:       map[string][]byte{".dockerconfigjson": []byte("{}")},
		}
		gatewayComp, gatewayCompErr := GatewayAPIImplementationComponent(&GatewayAPIImplementationConfig{
			Scheme:                   testScheme(),
			Installation:             installation,
			GatewayAPI:               gatewayAPI,
			PullSecrets:              []*corev1.Secret{pullSecret},
			GatewayNamespaces:        []string{"default"},
			CurrentGatewayNamespaces: set.New("default", "removed-ns"),
		})

		// ... existing SA/CRB assertions, then add:
		foundSecret := false
		for _, obj := range objsToDelete {
			if s, ok := obj.(*corev1.Secret); ok && s.Name == "tigera-pull-secret" && s.Namespace == "removed-ns" {
				foundSecret = true
			}
		}
		Expect(foundSecret).To(BeTrue(), "expected stale pull secret in objsToDelete")
```

radixo: Good catch, thank you! :rocket:

pkg/controller/gatewayapi/gatewayapi_controller.go
radixo: It is used by the tests

electricjesus (approved):
LGTM!

pkg/controller/gatewayapi/gatewayapi_controller.go
```diff
@@ -401,10 +451,66 @@ func (r *ReconcileGatewayAPI) Reconcile(ctx context.Context, request reconcile.R
 		}
 	}
 
+	// When in GatewayNamespace mode and Enterprise, discover namespaces that contain
+	// Gateway resources using operator-managed GatewayClasses. These namespaces need
+	// per-namespace Enterprise resources (SA, CRB, pull secrets).
+	if *gatewayAPI.Spec.GatewayDeploymentMode == operatorv1.GatewayDeploymentModeGatewayNamespace &&
+		variant.IsEnterprise() {
```
caseydavenport: Why does the Variant matter here at all?

radixo: For resources that are rendered only on EE license, like WAF.

pkg/render/gatewayapi/gateway_api.go
```diff
@@ -133,242 +147,220 @@ var (
...
+// sync.Once. Callers must deep-copy any object they intend to mutate.
+var (
+	chartOnce    sync.Once
+	chartResults *gatewayAPIResources
+	chartErr     error
+)
+
+func chartResourcesFor(scheme *runtime.Scheme) (*gatewayAPIResources, error) {
+	chartOnce.Do(func() {
+		chartResults, chartErr = renderChart(scheme, ReleaseName, common.CalicoNamespace, "", "GatewayNamespace")
```
rene-dekker: It is the only called of renderChart and passes "" for controller name, this effectively means that controllername is not a used param in renderChart.

pkg/render/gatewayapi/gateway_api.go
```diff
@@ -133,242 +147,220 @@ var (
...
+		case *appsv1.Deployment:
+			resources.controllerDeployment = typedObj
+		case *batchv1.Job:
+			resources.certgenJob = typedObj
+		case *admissionregv1.MutatingWebhookConfiguration:
+			resources.mutatingWebhookConfigurations = append(resources.mutatingWebhookConfigurations, typedObj)
+		case *corev1.Namespace:
+			// The chart may render a namespace; we create our own in Objects(), so skip it.
+		default:
+			// Skip unknown types.
```
rene-dekker: Should we make a warning log for this? Do we have a way to catch it in a test if there are new types we can't render?

pkg/render/gatewayapi/gateway_api.go
radixo: Good one, it remained from the old structure for both installs, cleaning it up.

pkg/render/gatewayapi/gateway_api.go
radixo: Good catch — the tests didn't cover it. Switched `default:` to return an error so any unhandled kind from a future chart bump fails the render tests in CI.

pkg/render/gatewayapi/gateway_api.go
```diff
@@ -1207,24 +1325,106 @@ func (pr *gatewayAPIImplementationComponent) wafHttpFilterClusterRole() *rbacv1.
...
+			Order:    &networkpolicy.HighPrecedenceOrder,
+			Tier:     networkpolicy.CalicoTierName,
+			Selector: EnvoyGatewayPolicySelector,
+			Types:    []v3.PolicyType{v3.PolicyTypeIngress, v3.PolicyTypeEgress},
+			// 9443: webhook. 18000-18002: xDS. 19001: metrics.
+			Ingress: []v3.Rule{
+				{
+					Action:   v3.Allow,
+					Protocol: &networkpolicy.TCPProtocol,
+					Source:   v3.EntityRule{Nets: []string{"0.0.0.0/0"}},
```
rene-dekker: This wouldn't work on ipv6, should we add "::/0"?

pkg/render/gatewayapi/gateway_api.go
```diff
@@ -1207,24 +1325,106 @@ func (pr *gatewayAPIImplementationComponent) wafHttpFilterClusterRole() *rbacv1.
...
 				Name:      wafFilterName,
-				Namespace: "tigera-gateway",
+				Namespace: namespace,
+			},
+		},
+	}
+}
+
+// gatewayAPIControllerPolicy allows the controller + certgen to reach kube-apiserver and DNS.
+func gatewayAPIControllerPolicy(namespace string, openShift bool) *v3.NetworkPolicy {
```
rene-dekker, line 1414: The labels on the deployment are taken from helm. And so if they decide to change something upstream, our network policy may start to block traffic unnoticed by us. There also seems to be a mismatch between the Deploy label and the pod label. Is it possible that we add our own labels to the deploy and pod, and  use these for the network policy as well? Maybe `calico-gateway-controller`?

pkg/controller/gatewayapi/gatewayapi_controller.go
```diff
@@ -419,10 +472,64 @@ func (r *ReconcileGatewayAPI) Reconcile(ctx context.Context, request reconcile.R
 		}
 	}
 
+	// Start watching Gateway resources now that the CRDs are in place, so future
```
rene-dekker, line 475: Should we add a check here to disallow gateways with namespace tigera-gateway? If they specify this namespace, will there be issues with the clean up and creation of resources?

pkg/render/gatewayapi/gateway_api.go
```diff
@@ -102,6 +102,7 @@ type gatewayAPIResources struct {
 const (
 	GatewayAPIName                      = "gateway-api"
```
rene-dekker: Could we prepend with calico- ? `calico-gateway-api`

radixo: Yeah.. good, avoids collisions.

rene-dekker (approved):

PR #4702, opened 2026-04-16
Defer Gateway API watch in waypoint secrets controller

caseydavenport:
The waypoint secrets controller (#4483) registers a `WatchObject` for Gateway resources at startup. On clusters without Gateway API CRDs installed, the informer cache sync times out after ~2 minutes and crashes the entire operator manager, putting it into CrashLoopBackOff. This breaks basic operator functionality (e.g., IP pool creation) on clusters that don't have Gateway API CRDs.

Switch to the existing `WaitToAddResourceWatch` pattern used elsewhere in the operator, which retries the watch registration in a background goroutine with exponential backoff until the CRD appears. Gate the `Reconcile()` Gateway listing behind a `ReadyFlag` to avoid unnecessary `List()` calls before the watch is established.

```release-note
Fix operator crash on clusters without Gateway API CRDs installed.
```

### DIFF
--- a/pkg/controller/istio/waypoint/waypoint_secrets_controller.go
+++ b/pkg/controller/istio/waypoint/waypoint_secrets_controller.go
@@ -20,6 +20,7 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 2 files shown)


electricjesus (approved):
LGTM!

PR #4704, opened 2026-04-16
Stop mutating AddToSchemes on every AddToScheme call

alexh-tigera:
## Description

`AddToScheme` appended a new `calicoSchemeBuilder` closure to the package-level AddToSchemes slice on every call, so the Nth call ran N accumulated closures against the target scheme. 

In test binaries this caused dozens of "Registering Calico CRD types" log lines and left a latent bug where mixing `v3=true` and `v3=false` within one process would double-register the variable CRD types under both API groups.

Invoke `calicoSchemeBuilder(v3)` directly on the provided scheme instead, leaving AddToSchemes as the init()-populated set of builders.

## Release Note

```release-note
Fixed a bug in `AddToScheme` that was causing superfluous log lines in FV tests.
```

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/apis/register.go
+++ b/pkg/apis/register.go
@@ -51,8 +51,10 @@ var (
 
 // AddToScheme adds all Resources to the Scheme
 func AddToScheme(s *runtime.Scheme, v3 bool) error {
-	AddToSchemes = append(AddToSchemes, calicoSchemeBuilder(v3))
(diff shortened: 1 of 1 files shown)


caseydavenport (approved):
Thank you! This has been bugging me but hadn't gotten around to a fix yet!

PR #4742, opened 2026-04-23
Fix concurrent-append panic in override-field recording (#4703)

### LINKED ISSUE #4703: operator - showing continuous panic errors related to memory and appears to be causing pods to loose network access until restarted
We have been using `calico` very successfully for 5+ years now however in the last couple weeks we started to notice that some of our Kubernetes pods were getting network failures, they would effectively be offline for ingress traffic for 2-3 mins then start working for 1-2 mins then down again. 

After a lot of various troubleshooting we found that the `tigera-operator` was continuously spamming:
`"panic\":\"runtime error: invalid memory address or nil pointer dereference\"`

However immediately after that message is another error:
`"error\":\"panic: runtime error: invalid memory address or nil pointer dereference [recovered]\"`

We aren't sure if this is indicating there is no issue, but if so why is it continuously spamming these messages until we restart, then it seems to work ok for 1-3 days and then starts panic'ing again.

What we did find is that if we `restart` our impacted pods they continue to experience network issues. However if we FIRST restart the `tigera-operator` all the `panic` messages stop and if we then restart the impacted pods they start working normally until the panics start again. 

At this point we are pretty stumped we upgraded to the latest versions:

- image: quay.io/tigera/operator:v1.40.8
- image: quay.io/calico/apiserver:v3.31.5
- image: quay.io/calico/kube-controllers:v3.31.5
- image: quay.io/calico/node:v3.31.5
- image: quay.io/calico/pod2daemon-flexvol:v3.31.5
- image: quay.io/calico/typha:v3.31.5
- AWS EKS 1.35.0 / BottleRocket 1.57.0 / containerd 2.1.6+bottlerocket 
- |  vpc-cni |  ACTIVE |  v1.21.1-eksbuild.7  |
- |  coredns |  ACTIVE |  v1.13.2-eksbuild.4  |
- |  kube-proxy |  ACTIVE |  v1.35.3-eksbuild.2  |
- |  aws-ebs-csi-driver |  ACTIVE |  v1.58.0-eksbuild.1  |

I did see another [post](https://github.com/tigera/operator/issues/4222) from Oct 2025 that looks very similar to the panic I'm getting but the ticket doesn't appear to have been closed without resolution?

At this point this is have a serious impact on our clusters.. we have 6 separate EKS clusters all similar versions and 4 out of 5 are showing the issue. And the same `restart operator, restart pod` seems to fix things for a few days. We are pretty desperate to find a solution. So appreciate any help you can offer.

One other oddity is that our production clusters.. has never experienced the same network failures to date, all clusters run the same versions. However even though Prod never has the issues.. I can see in the logs this panic error going back to Jan 2026 .. which makes me second guess the cause, however the restarts do appear to solve it.

Thanks for any help.

[operator-panic-errors.txt](https://github.com/user-attachments/files/26788548/operator-panic-errors.txt)

tmjd:
## Description

Bug fix for a panic in the tigera-operator.

The package-level `overrideFieldsHandledInLastApplyCall` slice in `pkg/render/common/components` is documented as UT-only, but `getField` appends to it on every call — including from production code paths such as `ValidateReplicatedPodResourceOverrides` → `GetMetadata`. Multiple controllers reconcile concurrently, so the unsynchronized appends race; when two goroutines trigger `growslice` at once the slice header tears and `memmove` dereferences a bogus pointer, producing `runtime error: invalid memory address or nil pointer dereference` panics from `recordHandledField`.

This PR gates the recording (and its `resetHandledFields` companion) on a new package-level `recordHandledFields` flag. Production reconcilers never touch the shared slice. The one unit test that reads the slice (`check for unhandled fields`) flips the flag on for the duration of the test.

**Components affected:** `pkg/render/common/components`. The fix is entirely internal to that package — all production callers (across `pkg/render/*` and `pkg/common/validation`) go through the same `GetMetadata` / `GetContainers` / `GetNodeSelector` / etc. accessors and get the new no-op behavior for free.

**Observed panic (v1.40.7):**

```
panic: runtime error: invalid memory address or nil pointer dereference
...
runtime.memmove
runtime.growslice
github.com/tigera/operator/pkg/render/common/components.recordHandledField
    /go/src/github.com/tigera/operator/pkg/render/common/components/components.go:384
github.com/tigera/operator/pkg/render/common/components.getField
    /go/src/github.com/tigera/operator/pkg/render/common/components/components.go:230
github.com/tigera/operator/pkg/render/common/components.GetMetadata
    /go/src/github.com/tigera/operator/pkg/render/common/components/components.go:50
github.com/tigera/operator/pkg/common/validation.ValidateReplicatedPodResourceOverrides
    /go/src/github.com/tigera/operator/pkg/common/validation/overrides.go:39
github.com/tigera/operator/pkg/controller/installation.validateCustomResource
    /go/src/github.com/tigera/operator/pkg/controller/installation/validation.go:351
```

The `growslice` → `memmove` signature is the textbook concurrent-`append` race.

**Testing:**
- `go build ./...`
- `go test -race -count=1 ./pkg/render/common/components/...` (the existing `check for unhandled fields` table test still passes, now with the flag toggled per-case)

Fixes #4703.

## Release Note

```release-note
Fixed a panic in the tigera-operator caused by concurrent reconcilers racing on an unsynchronized slice used to track handled override fields.
```

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/common/components/components.go
+++ b/pkg/render/common/components/components.go
@@ -453,13 +453,25 @@ func applyReplicatedPodResourceOverrides(r *replicatedPodResource, overrides any
 // For UT purposes, only, this variable stores the override fields that were handled in that most
 // recent `applyReplicatedPodResourceOverrides` call.  UT code then checks that the structures we
 // use for `overrides` do not have any _other_ fields than those (except for known special cases).
-var overrideFieldsHandledInLastApplyCall []string
+// Writes to it are gated on recordHandledFields: production reconcilers run many Apply*Overrides
+// calls concurrently, and an unsynchronized append on a shared slice races during growslice
+// (see issue #4703). Tests that inspect the slice must set recordHandledFields = true.
+var (
+	overrideFieldsHandledInLastApplyCall []string
+	recordHandledFields                  bool
+)
 
 func resetHandledFields() {
+	if !recordHandledFields {
+		return
+	}
 	overrideFieldsHandledInLastApplyCall = nil
 }
 
 func recordHandledField(fieldNames []string) {
+	if !recordHandledFields {
+		return
+	}
 	dottedName := strings.Join(fieldNames, ".")
 	overrideFieldsHandledInLastApplyCall = append(overrideFieldsHandledInLastApplyCall, dottedName)
 }
--- a/pkg/render/common/components/components_test.go
+++ b/pkg/render/common/components/components_test.go
@@ -135,6 +135,8 @@ var _ = Describe("Common components render tests", func() {
 
 	DescribeTable("check for unhandled fields",
 		func(overrides any, expectUnhandled bool, allowedExtraFields ...string) {
+			recordHandledFields = true
+			defer func() { recordHandledFields = false }()
 			// Call applyReplicatedPodResourceOverrides to discover all the fields that
 			// we handle.
 			r := &replicatedPodResource{}


nelljerram (approved):

tmjd:
Replaced by #4755

PR #4756, opened 2026-04-28
gatewayapi: render namespace before EnvoyProxyRef resolution

electricjesus:
## Description

Bug fix in the gatewayapi controller. Components affected: `pkg/controller/gatewayapi`, `pkg/render/gatewayapi`.

When a `GatewayAPI` CR is created on a fresh install with `envoyProxyRef` set on a `GatewayClass` pointing at an `EnvoyProxy` in the operator-managed `tigera-gateway` namespace, reconcile early-returns at `pkg/controller/gatewayapi/gatewayapi_controller.go:350` on the missing `EnvoyProxy` before reaching the non-CRD render at `gatewayapi_controller.go:419` that creates the namespace. Users can't create the `EnvoyProxy` because the namespace doesn't exist, so the controller deadlocks in `Degraded` state with:

```
gatewayapi  Degraded=True  "Error reading EnvoyProxyRef: EnvoyProxy.gateway.envoyproxy.io 'gateway-conformance-base' not found"
```

This PR splits the namespace render out of the non-CRD component into a new `GatewayAPINamespaceComponent` and applies it early, alongside CRDs and before `EnvoyProxyRef` resolution. The `tigera-gateway` namespace is part of the operator's contract for any `GatewayAPI` CR, so it should always exist; otherwise users can't pre-create custom `EnvoyProxy` resources in it.

The new component is owned by the `GatewayAPI` CR (same as before), so the namespace is still cleaned up when the CR is removed. `CreateOrUpdateOrDelete` is idempotent on the existing namespace, so existing clusters are unaffected.

### Testing

- `make ut UT_DIR=./pkg/controller/gatewayapi` — 14/14 pass; added 1 regression test covering the bug case (`EnvoyProxyRef` configured but `EnvoyProxy` missing → namespace still rendered)
- `make ut UT_DIR=./pkg/render/gatewayapi`
- Manual verification on a single-node Calico OSS kind cluster (master): apply `GatewayAPI` with `envoyProxyRef` pointing at a not-yet-existent `EnvoyProxy` in `tigera-gateway` → `tigera-gateway` namespace created within seconds, `tigerastatus gatewayapi` Degraded with the expected `Error reading EnvoyProxyRef`; then apply the `EnvoyProxy` → reconcile completes, gatewayapi `Available=True`, `envoy-gateway` pod Running, `GatewayClass tigera-gateway-class` Accepted

## Release Note

```release-note
Fix gatewayapi controller deadlock on fresh install when a GatewayClass envoyProxyRef points at an EnvoyProxy in tigera-gateway: the tigera-gateway namespace is now rendered before EnvoyProxyRef resolution, so users can pre-create custom EnvoyProxy resources without the controller stalling.
```

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:
  - [x] `kind/bug` if this is a bugfix.
  - [ ] `kind/enhancement` if this is a a new feature.
  - [ ] `enterprise` if this PR applies to Calico Enterprise only.

### DIFF
--- a/pkg/controller/gatewayapi/gatewayapi_controller.go
+++ b/pkg/controller/gatewayapi/gatewayapi_controller.go
@@ -262,6 +262,19 @@ func (r *ReconcileGatewayAPI) Reconcile(ctx context.Context, request reconcile.R
 		reqLogger.Info("Could not render all optional GatewayAPI CRDs", "err", err)
 	}
 
+	// Render the tigera-gateway Namespace early — before any of the steps below that may
+	// early-return (e.g. EnvoyProxyRef resolution). The namespace is part of the operator's
+	// contract for any GatewayAPI CR, so it must always exist; otherwise users referencing
+	// custom EnvoyProxy resources in tigera-gateway on a fresh install hit a deadlock where
+	// reconcile fails on the missing EnvoyProxy and never reaches the non-CRD render that
+	// would have created the namespace.
+	namespaceComponent := gatewayapi.GatewayAPINamespaceComponent(installationSpec)
+	err = r.newComponentHandler(log, r.client, r.scheme, gatewayAPI).CreateOrUpdateOrDelete(ctx, namespaceComponent, nil)
+	if err != nil {
+		r.status.SetDegraded(operatorv1.ResourceCreateError, "Error rendering tigera-gateway namespace", err, log)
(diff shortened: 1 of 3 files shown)


caseydavenport:

```

```

PR #4765, opened 2026-05-01
Only render IDS managed-cluster watch binding on management clusters

caseydavenport:
The intrusion-detection-controller binding to `calico-managed-cluster-watch` is rendered for any non-managed cluster, but the `ClusterRole` it binds to is only rendered when `ManagementCluster` is set. On standalone clusters that produces a dangling binding referring to a ClusterRole that doesn't exist, and the IDS pod logs a constant stream of `cannot list managedclusters ... RBAC: clusterrole.rbac.authorization.k8s.io "calico-managed-cluster-watch" not found` errors.

This PR gates the binding on `ManagementCluster` so it matches the ClusterRole's gating, and explicitly cleans up both the `RoleBinding` and `ClusterRoleBinding` variants on non-management clusters in case a cluster transitions from management to standalone (or had the binding leftover from an older version). It also passes a `MANAGEMENT_CLUSTER` env var through to the IDS container so the controller itself can skip the `ManagedCluster` watch on standalone clusters - that side lives in https://github.com/tigera/calico-private.

```release-note
Stops the intrusion-detection-controller from logging repeated RBAC errors about ManagedClusters on standalone clusters.
```

### DIFF
--- a/pkg/render/intrusion_detection.go
+++ b/pkg/render/intrusion_detection.go
@@ -17,6 +17,7 @@ package render
 import (
 	"crypto/x509"
 	"fmt"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


dimitri-nicolo (approved):
lgtm

PR #4767, opened 2026-05-01
Add LinuxPodInterfaceType to select netkit for pod CNI device

tomastigera:
## Description

Adds `Installation.spec.calicoNetwork.linuxPodInterfaceType` so users can opt the Calico CNI plugin into creating netkit L2 pairs (Linux 6.7+) instead of veth pairs for the pod-side interface. The operator plumbs the field through to the rendered CNI ConfigMap as `"device_type": "netkit"`.

**Depends on [projectcalico/calico#12619](https://github.com/projectcalico/calico/pull/12619)**, which adds the `device_type` knob to the Calico CNI plugin itself. Without that change merged & shipped, setting this field has no effect (the CNI plugin will ignore the unknown key).

The CNI plugin silently falls back to veth on kernels older than 6.7, so the field is safe to set on heterogeneous clusters. For non-BPF dataplanes netkit L2 is functionally equivalent to veth (TC/TCX still attaches); the performance benefit (per ByteDance: ~10% throughput, big P99 improvement under contention) comes when the BPF dataplane uses `BPF_NETKIT_PRIMARY`.

Validation rejects the field for non-Calico CNI plugins, matching `ContainerIPForwarding` / `HostPorts` / `MultiInterfaceMode`. `device_type` is only emitted in the rendered CNI config when the field is explicitly set to `Netkit`, so existing installations are unchanged.

Related: [CORE-10672](https://tigera.atlassian.net/browse/CORE-10672)

### Testing
- `make ut UT_DIR=./pkg/render` — added two render tests covering Netkit (emits `device_type`) and Veth-default (no key emitted).
- `make ut UT_DIR=./pkg/controller/installation` — added validation test entry covering the Calico-CNI-only restriction.
- `make static-checks`, `make format-check`, `make test-crds` — clean.

## Release Note

```release-note
Added `Installation.spec.calicoNetwork.linuxPodInterfaceType` to select between `Veth` (default) and `Netkit` for the Calico CNI pod interface. Requires Calico CNI plugin support for `device_type` (Linux 6.7+ for netkit; older kernels fall back to veth).
```

## For PR author

- [x] Tests for change.
- [ ] If changing versions, run `make gen-versions` (n/a)

[CORE-10672]: https://tigera.atlassian.net/browse/CORE-10672?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/v1/installation_types.go
+++ b/api/v1/installation_types.go
@@ -496,6 +496,17 @@ const (
 	ContainerIPForwardingDisabled ContainerIPForwardingType = "Disabled"
 )
 
+// LinuxPodInterfaceType specifies the type of virtual device the Calico CNI plugin
+// creates for the pod-side interface on Linux nodes.
+//
+// One of: Veth, Netkit
+type LinuxPodInterfaceType string
+
+const (
+	LinuxPodInterfaceVeth   LinuxPodInterfaceType = "Veth"
+	LinuxPodInterfaceNetkit LinuxPodInterfaceType = "Netkit"
+)
+
 // HostPortsType specifies host port support.
 //
 // One of: Enabled, Disabled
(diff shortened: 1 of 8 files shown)


api/v1/installation_types.go
```diff
@@ -496,6 +496,17 @@ const (
 	ContainerIPForwardingDisabled ContainerIPForwardingType = "Disabled"
 )
 
+// LinuxPodInterfaceType specifies the type of virtual device the Calico CNI plugin
+// creates for the pod-side interface on Linux nodes.
+//
+// One of: Veth, Netkit
+type LinuxPodInterfaceType string
```
caseydavenport, line 504: We should stick the kubebuilder tag on the type rather than the var.

caseydavenport (approved):
One nit but otherwise LGTM.

PR #4768, opened 2026-05-01
Default BPFKubeProxyHealthzPort to 0 when kube-proxy is not managed

tomastigera:
## Description

When BPF dataplane is enabled but the operator is not managing the kube-proxy DaemonSet (e.g. on AKS, where the platform owns kube-proxy), kube-proxy keeps running alongside calico-node and holds the default healthz port (`10256`). Felix's BPF kube-proxy healthz server then fails to bind and `calico-node` logs:

```
failed to start proxy healthz on :10256: listen tcp :10256: bind: address already in use
```

Until now the documented workaround was for users to manually set `FelixConfiguration.BPFKubeProxyHealthzPort` to `0` (or to an unused port).

This change has the operator default `BPFKubeProxyHealthzPort` to `0` (disabled) in `setDefaultsOnFelixConfiguration` when:

1. BPF is enabled (`Installation.Spec.BPFEnabled()`),
2. the operator is not managing kube-proxy (`!Installation.Spec.KubeProxyManagementEnabled()`), and
3. the field is currently unset on `FelixConfiguration`.

When the operator manages kube-proxy (i.e. it disables the platform's kube-proxy DaemonSet on BPF nodes), Felix's BPF kube-proxy healthz server keeps the upstream default port and continues to act as the kube-proxy healthz replacement — no behavior change for that path. User-set values on `FelixConfiguration.BPFKubeProxyHealthzPort` are always preserved.

The new helper `disableBPFKubeProxyHealthz` mirrors the existing `disableBPFHostConntrackBypass` pattern.

## Testing

- New unit tests for `disableBPFKubeProxyHealthz` in `pkg/controller/installation/bpf_test.go` (sets to 0 from nil; overwrites existing value).
- New reconcile-driven tests in `pkg/controller/installation/core_controller_test.go` covering all four gating cases:
  - BPF on + KubeProxyManagement disabled → port set to 0 (the AKS case).
  - BPF off → port unset.
  - BPF on + KubeProxyManagement = Enabled → port unset.
  - User-set value → preserved.
- `make ut UT_DIR=./pkg/controller/installation` — 298 of 298 specs pass.

## Related

- Internal: CORE-12755
- OSS PRs that introduced and made the port configurable in Felix: projectcalico/calico#10947, projectcalico/calico#11011

## Release Note

```release-note
On BPF clusters where the operator is not managing kube-proxy (e.g. AKS), the operator now defaults FelixConfiguration.BPFKubeProxyHealthzPort to 0 to avoid a port conflict with the platform's kube-proxy. Users can still override the port explicitly.
```

## For PR author

- [x] Tests for change.

### DIFF
--- a/pkg/controller/installation/bpf.go
+++ b/pkg/controller/installation/bpf.go
@@ -129,3 +129,11 @@ func disableBPFHostConntrackBypass(fc *v3.FelixConfiguration) {
 	hostConntrackBypassDisabled := false
 	fc.Spec.BPFHostConntrackBypass = &hostConntrackBypassDisabled
 }
+
+// disableBPFKubeProxyHealthz disables Felix's BPF kube-proxy healthz server by setting
+// BPFKubeProxyHealthzPort to 0. Use when Calico runs in BPF mode but the platform's
+// kube-proxy is still running (e.g. AKS) and holds the default port (10256).
+func disableBPFKubeProxyHealthz(fc *v3.FelixConfiguration) {
+	port := 0
+	fc.Spec.BPFKubeProxyHealthzPort = &port
+}
--- a/pkg/controller/installation/bpf.go
+++ b/pkg/controller/installation/bpf.go
@@ -26,6 +26,7 @@ import (
 	"github.com/tigera/operator/pkg/render"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/utils/ptr"
 )
 
 // bpfValidateAnnotations validate Felix Configuration annotations match BPF Enabled spec for all scenarios.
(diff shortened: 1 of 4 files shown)


pkg/controller/installation/bpf.go
```diff
@@ -129,3 +129,11 @@ func disableBPFHostConntrackBypass(fc *v3.FelixConfiguration) {
 	hostConntrackBypassDisabled := false
 	fc.Spec.BPFHostConntrackBypass = &hostConntrackBypassDisabled
 }
+
+// disableBPFKubeProxyHealthz disables Felix's BPF kube-proxy healthz server by setting
+// BPFKubeProxyHealthzPort to 0. Use when Calico runs in BPF mode but the platform's
+// kube-proxy is still running (e.g. AKS) and holds the default port (10256).
+func disableBPFKubeProxyHealthz(fc *v3.FelixConfiguration) {
+	port := 0
+	fc.Spec.BPFKubeProxyHealthzPort = &port
```
caseydavenport: Can use `ptr.To` here

caseydavenport (approved):
LGTM - one nit.

pkg/controller/installation/bpf.go
tomastigera: yeah,that is what I was looking for! I wanted to use it for the first time! ;-)

Issue #4718, opened 2026-04-20
Allow setting healthCheckNodePort in GatewayAPI


sebhoss:
In our setup, we do not rely upon node ports. We can already set `gatewayService.spec.allocateLoadBalancerNodePorts: false` to close one node port, however the healthCheckNodePort is still opened. Since it [cannot be closed](https://github.com/kubernetes/kubernetes/issues/130776), how about allowing setting the HCNP port manually instead of letting Kubernetes pick a random port number. The manually selected port number is easier to use in network policies. Something like this would be perfect for us:

```
apiVersion: operator.tigera.io/v1
kind: GatewayAPI
metadata:
  name: default
spec:
  gatewayClasses:
    - name: calico
      gatewayService:
        spec:
          healthCheckNodePort: 12345
```

electricjesus:
Thanks for filing this.

The Tigera operator renders an Envoy Gateway `EnvoyProxy` resource, and the fields we expose under `gatewayService.spec` map directly to upstream `KubernetesServiceSpec`. As of envoy-gateway main, that struct exposes `type`, `loadBalancerClass`, `allocateLoadBalancerNodePorts`, `loadBalancerSourceRanges`, `loadBalancerIP`, and `externalTrafficPolicy` but not `healthCheckNodePort`. So even if we added the field to our CRD today, there's nothing to pipe it through; the envoy-gateway service translator never sets it, and kube-apiserver picks next available from cluster nodeport range.

I couldn't find any existing issue or PR on envoyproxy/gateway tracking this.

Path forward:

1. An issue filed or a PR against envoyproxy/gateway to add `HealthCheckNodePort *int32` to `KubernetesServiceSpec`, with CEL validation restricting it to `type: LoadBalancer` + `externalTrafficPolicy: Local` (matching k8s rules).
2. Once that's merged and a release is cut, we'll bump the envoy-gateway dependency here and expose the new field on `GatewayServiceSpec`.

Change seems small; I'll look into driving it and will keep this issue open for tracking. If you'd like to help on the upstream PR, happy to coordinate.

EDIT: Please see the issue (+ PR) I opened below

sebhoss:
Thanks @electricjesus, this is awesome!

electricjesus:
@sebhoss looks like @nelljerram added a way for generally patching the gateway service - you should check it out!

sebhoss:
[...]

Issue #4717, opened 2026-04-20
Support DualStack load balancer in GatewayAPI service


sebhoss:
We have a dual-stack cluster and want to set the `ipFamilyPolicy` of the generated `Service` to `RequireDualStack`. Currently that is not possible and the operator always sets this to `SingleStack` which allocates only an IPv4 address in our stack. Something like the following would be perfect:

```yaml
apiVersion: operator.tigera.io/v1
kind: GatewayAPI
metadata:
  name: default
spec:
  gatewayClasses:
    - name: calico
      gatewayService:
        spec:
          ipFamilyPolicy: RequireDualStack
```

nelljerram:
@sebhoss Can I just check which `Service` you mean?  The `Service` for the gateway controller, or the `Service` that gets provisioned for each `Gateway`?

This kind of request makes sense to me, but I just want to be sure I understand it in detail.

nelljerram:
D'oh, I realized after reading #4718 as well that you obviously mean the per-Gateway Service.

nelljerram:
So the situation is the same as @electricjesus said [here](https://github.com/tigera/operator/issues/4718#issuecomment-4311974848).  Envoy's API doesn't currently expose a direct way to set `ipFamilyPolicy` in the Services that it creates - i.e. in https://gateway.envoyproxy.io/latest/api/extension_types/#kubernetesservicespec - so it isn't straightforward for us to do this, unless Envoy is enhanced first.

Except ... if it's possible using the `patch` field in KubernetesServiceSpec.  If you're interested in pursuing this immediately, you could try:
1. Configure a Gateway in Calico as things work today.
2. Record the resulting EnvoyProxy YAML (named after the gateway class, in namespace "tigera-gateway").
3. Create your own EnvoyProxy based on that YAML, but with `KubernetesServiceSpec.patch` crafted to set `ipFamilyPolicy` in the way you want.
4. Configure a GatewayClass that uses the `envoyProxyRef` field to reference your EnvoyProxy.
5. Configure a Gateway with that class, and check the results.
6. Iterate if needed from (3).

If you're able to craft a `patch` that works, we could then look at adding a field in our `GatewayServiceSpec` struct and plumbing it through to your patch.

nelljerram:
@sebhoss I had a similar kind of conversation back in February where a user reported success with a `patch` like this:
```
 apiVersion: gateway.envoyproxy.io/v1alpha1
 kind: EnvoyProxy
 metadata:
[...]

PR #4773, opened 2026-05-05
Support general patching of Gateway Service

nelljerram:
Adds a Patch field to GatewayServiceSpec that passes through to the EnvoyProxy's KubernetesServiceSpec.Patch, allowing arbitrary JSON-merge or strategic-merge patches to the rendered Service. This enables use cases such as ipFamilyPolicy: RequireDualStack (#4717) and a user-specified healthCheckNodePort (#4718).

Includes unit tests covering generic passthrough and both of those specific use cases.

## Release Note

```release-note
The `GatewayAPI` resource now permits general patching of the Services that are provisioned for gateways within a custom gateway class, by adding a `Patch` field in `GatewayServiceSpec` that is passed through to the corresponding field in `EnvoyProxy`.
```


api/v1/gatewayapi_types.go
```diff
@@ -508,4 +509,7 @@ type GatewayServiceSpec struct {
 
 	// +optional
 	LoadBalancerIP *string `json:"loadBalancerIP,omitempty"`
+
+	// +optional
```
caseydavenport, line 523: The rest of this struct is also under documented, but let's add a comment to this new field at least - it's rendered into user-facing documentation (e.g., via `kubectl explain`) in the CRD.

caseydavenport (approved):
LGTM - just one nit

electricjesus (approved):

api/v1/gatewayapi_types.go
nelljerram: https://github.com/tigera/operator/pull/4773/commits/5ada9bbcb7af04d583a58096070ae78149bc77c3 - WDYT?

PR #4780, opened 2026-05-06
fix(apiserver): wire queryserver to Linseed via tigera-linseed service in MCM

tianfeng92:
## What

On managed clusters, render the queryserver to talk to Linseed through the same `tigera-linseed` Service / signed-token / Linseed CA path as fluentd:

- Add `VoltronLinseedPublicCert` to `apiserver-ca-bundle` and watch it.
- Render `Service/tigera-linseed` (ExternalName → guardian) and `RoleBinding/tigera-linseed` (guardian SA, `tigera-linseed-secrets`) in `calico-system`.
- On the queryserver: `LINSEED_URL=https://tigera-linseed.calico-system.svc.<domain>`, `CLUSTER_ID=""`, mount `calico-apiserver-tigera-linseed-token`, `LINSEED_TOKEN=<mounted path>`.

Pairs with https://github.com/tigera/calico-private/pull/11839.

## Why

Manager UI policy list returns 500 — queryserver's `enrichPoliciesWithActivity` fails. Three gaps on the path queryserver → guardian → voltron tunnel → Linseed:

1. **Trust bundle** missed the management cluster's `tigera-operator-signer`; both signers share Subject DN, so x509 matched the wrong CA (`crypto/rsa: verification error`).
2. **URL** pointed at `guardian.calico-system.svc` directly, diverging from fluentd's `tigera-linseed` Service pattern that aligns SNI/Host with the cert voltron actually serves for Linseed.
3. **Auth** — no `LINSEED_TOKEN` was set, so the queryserver sent its managed-cluster SA token which Linseed cannot validate → 401. And `CLUSTER_ID` defaulted to the literal `"cluster"`, which voltron's `inner_handler` rejected.

## Behavior changes

Managed clusters only — every change gated on `ManagementClusterConnection != nil`. New `Service` + `RoleBinding` in `calico-system`, new env vars + token volume on the queryserver, and `apiserver-ca-bundle` extended with the management cluster's CA. No-op on management/standalone.

## Release Note

```release-note
Fix 500 errors on the policy list page for managed clusters by routing the calico-apiserver queryserver to Linseed through the tigera-linseed service with the correct trusted CA, the Linseed-issued bearer token, and an empty x-cluster-id so voltron rewrites it.
```

### DIFF
--- a/pkg/render/apiserver.go
+++ b/pkg/render/apiserver.go
@@ -140,6 +140,7 @@ type APIServerConfiguration struct {
 	MultiTenant                  bool
 	KeyValidatorConfig           authentication.KeyValidatorConfig
 	KubernetesVersion            *common.VersionInfo
+	ClusterDomain                string
 
 	// Whether or not we should run the aggregation API server for projectcalico.org/v3 APIs
 	// as part of this component.
@@ -327,6 +328,12 @@ func (c *apiServerComponent) Objects() ([]client.Object, []client.Object) {
 		namespacedEnterpriseObjects = append(namespacedEnterpriseObjects, c.sidecarMutatingWebhookConfig())
 	} else {
 		objsToDelete = append(objsToDelete, &admregv1.MutatingWebhookConfiguration{ObjectMeta: metav1.ObjectMeta{Name: common.SidecarMutatingWebhookConfigName}})
+	}
+	if c.cfg.ManagementClusterConnection != nil {
+		namespacedEnterpriseObjects = append(namespacedEnterpriseObjects,
+			c.externalLinseedService(),
+			c.externalLinseedRoleBinding(),
+		)
 	}
 
 	// Compile the final arrays based on the variant.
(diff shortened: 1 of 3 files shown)


pkg/render/apiserver.go
```diff
@@ -1280,13 +1287,19 @@ func (c *apiServerComponent) queryServerContainer() corev1.Container {
 	// Linseed client configuration for policy activity enrichment.
 	linseedURL := fmt.Sprintf("https://tigera-linseed.%s.svc", ElasticsearchNamespace)
 	if c.cfg.ManagementClusterConnection != nil {
-		linseedURL = "https://guardian.calico-system.svc"
+		linseedURL = fmt.Sprintf("https://tigera-linseed.%s.svc.%s", APIServerNamespace, c.cfg.ClusterDomain)
```
caseydavenport: Think we want to use this: https://github.com/tigera/operator/blob/662ba6fb628fa374366d7c492b13758cd8b8225a/pkg/render/common/elasticsearch/service.go#L28-L48

caseydavenport (commented):
Just one comment!

have you tried this out to make sure it works?

caseydavenport (approved):

PR #4784, opened 2026-05-07
[CORE-12452] feat: auto-recover host-networked pods when node IP changes

coutinhop:
## Description

Detect Calico's host-networked pods (calico-typha, calico-node, calico-node-windows) whose status.podIPs no longer matches the node's current InternalIP, and delete them so the Deployment / DaemonSet controller recreates them with the correct IP.

This works around an upstream Kubernetes limitation [1] where status.podIPs is immutable for hostNetwork pods once set: when a node's IP changes (e.g. KubeVirt VM reboot pulls a new DHCP lease), existing hostNetwork pods keep their old IP. The kube EndpointSlice controller reads from status.podIPs, so the calico-typha EndpointSlice ends up advertising stale IPs and Felix times out connecting to Typha. Restarting the container does not help — only deleting and recreating the pod itself causes the kubelet to repopulate status.podIPs from the current node IP.

Implementation lives in the existing Typha autoscaler tick (every 10s, already has a Node informer cache):

  - Compare each pod's status.podIPs to its node's status.InternalIP (which the kubelet does update promptly via heartbeat).
  - Delete stale pods, paced one per workload-batch per tick. Batch size is read from each workload's existing rolling-update setting: the Typha PDB's maxUnavailable, or the DaemonSet's updateStrategy.rollingUpdate.maxUnavailable. Falls back to 1 if not set or if the resolved value is < 1 (minimum-progress guarantee).
  - Order: Typha first; if any Typha was deleted this cycle, skip the calico-node deletions until the next tick to give the new Typha pod a clean window to come up. Linux and Windows DaemonSets are paced independently of each other.
  - Skipped entirely on the non-cluster-host autoscaler instance.

Tested by ODCN on KubeVirt: 3-node cluster with all node IPs changed, all calico-node and Typha pods recovered automatically without manual intervention.

[1] https://github.com/kubernetes/kubernetes/issues/93897

Jira: CI-1951, CORE-12452

## Release Note

```release-note
Automatically recover Calico pods stranded with stale pod IPs after a node IP change (e.g. KubeVirt node reboot).
```

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


pkg/controller/podiprecovery/podiprecovery_controller.go
```diff
@@ -0,0 +1,273 @@
...
+	"github.com/tigera/operator/pkg/render/webhooks"
+)
+
+var log = logf.Log.WithName("controller_podiprecovery")
+
+// targetLabelSelectors is the set of label selectors identifying
+// operator-managed pods that are (or may be) host-networked. The controller
+// applies a per-pod hostNetwork check before deleting, so non-hostNetwork
+// pods that happen to match are left alone.
+var targetLabelSelectors = []labels.Selector{
```
caseydavenport: I do wonder if we could instead apply a label to ALL pods created by the operator, so we can do a single query for all pods and then just skip pods that don't have hostNet set? 

That would mean we're resilient to any future host net pods we might introduce? 

We could even do something like automatically label the pods we create with hostNet status.

```yaml
labels:
  operator.tigera.io/hostNetworked: true
```

pkg/controller/podiprecovery/podiprecovery_controller.go
```diff
@@ -0,0 +1,273 @@
...
+	labels.SelectorFromSet(labels.Set{render.AppLabelName: render.TyphaK8sAppName}),
+	labels.SelectorFromSet(labels.Set{render.AppLabelName: render.CalicoNodeObjectName}),
+	labels.SelectorFromSet(labels.Set{render.AppLabelName: render.WindowsNodeObjectName}),
+	labels.SelectorFromSet(labels.Set{render.AppLabelName: dpi.DeepPacketInspectionName}),
```
caseydavenport: Some of these are enterprise only so might want to separate them out.

pkg/controller/podiprecovery/podiprecovery_controller.go
```diff
@@ -0,0 +1,273 @@
...
+	scheme *runtime.Scheme
+}
+
+var _ reconcile.Reconciler = &Reconciler{}
+
+// Reconcile is called for a Node when its InternalIPs change (or on initial
+// creation). It lists operator-managed pods on the node and deletes any
+// host-networked pod whose status.podIPs doesn't include any of the node's
+// current InternalIPs.
+func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
```
caseydavenport, line 155: I think we'll probably want a Gate on Installation here? To prevent this controller running until we've actually deployed Calico?

pkg/controller/podiprecovery/podiprecovery_controller.go
```diff
@@ -0,0 +1,273 @@
...
+	nodeIPs := internalIPSet(node.Status.Addresses)
+	if len(nodeIPs) == 0 {
+		// Nothing to compare against; bail out to avoid deleting pods
+		// based on a transient empty status.
+		logger.V(1).Info("Node has no InternalIPs reported; skipping pod IP check")
+		return ctrl.Result{}, nil
+	}
+
+	// List operator-managed pods running on this node. We list once per
+	// label selector and filter by spec.nodeName on the client side. The
```
caseydavenport: I think we might be able to use a field selector on `spec.nodeName` to avoid the client-side filter?

pkg/controller/podiprecovery/podiprecovery_controller.go
```diff
@@ -0,0 +1,273 @@
...
+	// pod list is small enough that this is cheap.
+	pods, err := r.listOperatorManagedPodsOnNode(ctx, node.Name)
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to list pods on node %q: %w", node.Name, err)
+	}
+
+	var firstErr error
+	deleted := 0
+	for i := range pods {
```
caseydavenport: why not `_, pod := range pods`?

pkg/controller/podiprecovery/podiprecovery_controller.go
```diff
@@ -0,0 +1,273 @@
...
+		}
+		if podIPMatchesNode(pod, nodeIPs) {
+			continue
+		}
+
+		podIPs := make([]string, 0, len(pod.Status.PodIPs))
+		for _, pip := range pod.Status.PodIPs {
+			podIPs = append(podIPs, pip.IP)
+		}
+		logger.Info("Deleting pod with stale IP after node IP change so its controller can recreate it with the current IP",
```
caseydavenport:
```suggestion
		logger.Info("Host networked Pod has stale IP, recreate",
```

^ log is a bit of a mouthful, maybe this?

pkg/controller/podiprecovery/podiprecovery_controller.go
```diff
@@ -0,0 +1,273 @@
...
+		},
+		GenericFunc: func(e event.GenericEvent) bool {
+			return false
+		},
+	}
+}
+
+// sameInternalIPs returns true when both slices contain the same set of
+// NodeInternalIP addresses (order-independent).
+func sameInternalIPs(a, b []corev1.NodeAddress) bool {
```
caseydavenport: We can drop `sameInternalIPs` entirely and just call `maps.Equal` on the two `internalIPSet` results. Or, if we switch `internalIPSet` to return a `sets.Set[string]` (from `k8s.io/apimachinery/pkg/util/sets`), it is `.Equal()` and the helper goes away.

pkg/controller/podiprecovery/podiprecovery_controller.go
```diff
@@ -0,0 +1,273 @@
...
+func podIPMatchesNode(pod *corev1.Pod, nodeIPs map[string]bool) bool {
+	for _, pip := range pod.Status.PodIPs {
+		if nodeIPs[pip.IP] {
+			return true
+		}
+	}
+	return false
+}
+
+func keys(m map[string]bool) []string {
```
caseydavenport: Same deal as the `sameInternalIPs` comment - if we make `nodeIPs` a `sets.Set[string]` we get `.UnsortedList()` for free and can drop this helper. Otherwise `slices.Collect(maps.Keys(m))` from the stdlib works.

pkg/controller/podiprecovery/podiprecovery_controller.go
```diff
@@ -0,0 +1,273 @@
...
+	deleted := 0
+	for i := range pods {
+		pod := &pods[i]
+		if !pod.Spec.HostNetwork {
+			// Safety check: only delete hostNetwork pods. A non-hostNetwork
+			// pod that happens to match our labels has a CNI-assigned IP
+			// that legitimately differs from the node's IP.
+			continue
+		}
+		if len(pod.Status.PodIPs) == 0 && pod.Status.PodIP == "" {
```
caseydavenport: `PodIP` is just `PodIPs[0]` by contract, so `len(pod.Status.PodIPs) == 0` on its own is sufficient here.

pkg/controller/podiprecovery/podiprecovery_controller.go
```diff
@@ -0,0 +1,273 @@
...
+
+	if deleted > 0 {
+		logger.Info("Deleted stale-IP pods on node", "count", deleted)
+	}
+	return ctrl.Result{}, firstErr
+}
+
+// listOperatorManagedPodsOnNode lists pods on the given node that match any
+// of the operator's host-networked-workload label selectors.
+func (r *Reconciler) listOperatorManagedPodsOnNode(ctx context.Context, nodeName string) ([]corev1.Pod, error) {
```
caseydavenport: Related to my other comment about the field selector - we can also collapse all the `k8s-app` selectors into one List using `labels.NewRequirement(render.AppLabelName, selection.In, []string{...})`, plus a second List for the `apiserver=true` selector. Combined with a `spec.nodeName` field selector that is two Lists total instead of seven, and the `seen` dedup map drops out too.

pkg/common/common.go
```diff
@@ -42,6 +42,14 @@ const (
 	// the controller in the owner reference will not be set.
 	MultipleOwnersLabel = "operator.tigera.io/multipleOwners"
 
+	// HostNetworkedPodLabel marks pods that run with hostNetwork=true and are
+	// managed by the operator. The podiprecovery controller uses this label
+	// to identify pods that need their IPs re-checked after a node IP change
+	// (Kubernetes does not refresh status.podIPs for existing hostNetwork pods;
+	// see https://github.com/kubernetes/kubernetes/issues/93897). All render
+	// packages that produce hostNetwork pod templates must apply this label.
+	HostNetworkedPodLabel = "operator.tigera.io/hostNetworked"
```
caseydavenport:
```suggestion
	HostNetworkedPodLabel = "operator.tigera.io/host-networked"
```

pkg/controller/podiprecovery/podiprecovery_controller.go
```diff
@@ -0,0 +1,228 @@
...
+		},
+	}
+}
+
+// internalIPSet returns the set of NodeInternalIP addresses from a Node's
+// status.
+func internalIPSet(addrs []corev1.NodeAddress) sets.Set[string] {
+	out := sets.New[string]()
+	for _, a := range addrs {
+		if a.Type == corev1.NodeInternalIP {
```
caseydavenport: Is there ever a chance that we'll want to compare against an externalIP?

pkg/render/typha.go
```diff
@@ -446,6 +446,9 @@ func (c *typhaComponent) typhaDeployment() []client.Object {
 			Template: corev1.PodTemplateSpec{
 				ObjectMeta: metav1.ObjectMeta{
 					Annotations: annotations,
+					Labels: map[string]string{
```
caseydavenport: I was imagining we'd be able to insert this label in the common helper that we run all pods through: https://github.com/tigera/operator/blob/master/pkg/controller/utils/component.go#L1017

pkg/controller/podiprecovery/podiprecovery_controller.go
coutinhop: yes, thanks for catching that. I changed it to mirror kubelet's behaviour, falling back to externalIP if there is no internal IP...

pkg/render/typha.go
coutinhop: right, that is much better... done, thanks!

caseydavenport (approved):

PR #4785, opened 2026-05-07
EV-6586: reload metrics ClientCAs per connection

rene-dekker:
## Summary
- The operator metrics server loaded its mTLS client-CA pool once at startup. The CA Secret is created by the operator *after* the pod starts, so `loadClientCAFromFile` read a non-existent file, silently returned an empty `CertPool`, and the trust pool stayed empty for the life of the process — every Prometheus scrape then failed `tls: unknown certificate authority` until the operator was restarted.
- Move `ClientCAs` into `tls.Config.GetConfigForClient` so it is re-loaded per connection, symmetric with how `GetCertificate` already reloads the serving keypair. Volume updates and CA rotations are now picked up without a restart.
- Log the read / parse failure in `loadClientCAFromFile` so a missing or malformed CA file is no longer silent.

Bug: https://tigera.atlassian.net/browse/EV-6586
Regressed in: #4558

## Test plan
- [ ] Install Calico Enterprise master-stream on a fresh cluster; confirm `kubectl logs -n tigera-operator deploy/tigera-operator` shows no `TLS handshake error` entries after install completes.
- [ ] `kubectl exec -n tigera-prometheus prometheus-calico-node-prometheus-0 -- ...` shows the `tigera-operator-metrics` target as `health: up`.
- [ ] Delete and recreate `tigera-ca-private` to simulate rotation; confirm scrapes recover without restarting the operator pod.

```release-note
Fix operator metrics server starting with an empty client-CA trust pool when the CA Secret is created after the operator pod, which previously broke Prometheus scraping until a restart.
```

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -289,9 +289,13 @@ If a value other than 'all' is specified, the first CRD with a prefix of the spe
 				// is created by the operator after pod start, so an eager one-shot
 				// load would leave the trust pool permanently empty.
 				cfg.GetConfigForClient = func(*tls.ClientHelloInfo) (*tls.Config, error) {
+					pool, err := loadClientCAFromFile(metricsTLSCAFile())
+					if err != nil {
+						return nil, err
+					}
 					c := cfg.Clone()
 					c.ClientAuth = clientAuth
-					c.ClientCAs = loadClientCAFromFile(metricsTLSCAFile())
+					c.ClientCAs = pool
 					return c, nil
 				}
 			},
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -283,8 +283,17 @@ If a value other than 'all' is specified, the first CRD with a prefix of the spe
 			func(cfg *tls.Config) {
 				cfg.GetCertificate = getCert
 				cfg.ClientAuth = clientAuth
-				cfg.ClientCAs = loadClientCAFromFile(metricsTLSCAFile())
 				cfg.MinVersion = minVersion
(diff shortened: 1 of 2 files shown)


cmd/main.go
```diff
@@ -283,8 +283,17 @@ If a value other than 'all' is specified, the first CRD with a prefix of the spe
...
-				cfg.ClientCAs = loadClientCAFromFile(metricsTLSCAFile())
 				cfg.MinVersion = minVersion
+				// Re-read ClientCAs per connection so kubelet volume updates and CA
+				// rotations are picked up without an operator restart. The CA Secret
+				// is created by the operator after pod start, so an eager one-shot
+				// load would leave the trust pool permanently empty.
+				cfg.GetConfigForClient = func(*tls.ClientHelloInfo) (*tls.Config, error) {
+					c := cfg.Clone()
+					c.ClientAuth = clientAuth
+					c.ClientCAs = loadClientCAFromFile(metricsTLSCAFile())
```
coutinhop: maybe besides the point for this PR, but shouldn't `loadClientCAFromFile()` return errors? seems weird to have an error in the function singnature but we always return `nil`

coutinhop (approved):

cmd/main.go
rene-dekker: Good catch — fair point. Pushed 426413772: loadClientCAFromFile now returns (*x509.CertPool, error), and GetConfigForClient propagates the error. The Go TLS server surfaces returned errors via the standard TLS handshake error path, so the failure is still visible without a separate logger.

PR #4797, opened 2026-05-11
Add Installation.Spec.ImagePullPolicy

### LINKED ISSUE #7107: Allow specifying the Helm Chart's imagePullPolicy in values.yaml
Hello,

We deploy Kubernetes clusters programmatically, and part of the use case that we support is air-gapped (that is, without Internet access). The Helm Chart deploys several pods and daemonsets, among which are `calico-node` and `csi-node-driver`. Would it be possible to specify the pull policy of the aforementioned pods/daemonsets? Otherwise, as soon as we add a second node to the cluster, the daemonsets are brought up, and since we still haven't had time to push the images (we deploy a private Docker Registry within the cluster), they are downloaded from the Internet.

## Current Behavior
You can't change the image pull policy unless you modify the yamls, and even then you can only override `tigera-operator`.

## Possible Solution
Provide the following option:
```
- name: tigera-operator
   imagePullPolicy: IfNotPresent
```
and similar for other pods/daemonsets, or potentially one option for all of them.

## Your Environment

* Calico version: 3.24.5
* Orchestrator version (e.g. kubernetes, mesos, rkt): K3s
* Operating System and version: Red Hat 8.6

caseydavenport:
Adds an `imagePullPolicy` field to the Installation spec. When set, the operator applies it to every container in every pod it renders, overriding any default the renderers set themselves. This is for air-gapped clusters where images are pre-loaded onto nodes and shouldn't get pulled from a remote registry.

The pull policy is also now applied to init containers, which the central pass was previously missing.

Fixes https://github.com/projectcalico/calico/issues/7107.

```release-note
Added `Installation.Spec.ImagePullPolicy` for overriding the pull policy of all operator-managed pods, useful in air-gapped clusters.
```

### DIFF
--- a/pkg/controller/utils/component.go
+++ b/pkg/controller/utils/component.go
@@ -233,18 +233,12 @@ func (c *componentHandler) createOrUpdateObject(ctx context.Context, obj client.
 
 	// Look up the InstallationSpec once and reuse it for the passes that need it
 	// (image pull policy and TLS ciphers), so we don't pay for the same Get twice.
-	// Objects without a pod spec (e.g. NetworkPolicy, Tier) don't need it.
 	var installationSpec *operatorv1.InstallationSpec
-	if hasPodSpec(obj) {
-		if _, spec, err := GetInstallationSpec(ctx, c.client); err == nil {
-			installationSpec = spec
-		}
+	if _, spec, err := GetInstallationSpec(ctx, c.client); err == nil {
+		installationSpec = spec
 	}
 
-	// Make sure any objects with images also have an image pull policy. If the user has
-	// configured a pull policy on the Installation, override every container's policy with
-	// it — the user's choice (e.g. for air-gapped clusters) must win over operator defaults.
-	// Otherwise, fill in IfNotPresent for any container that does not already set a policy.
+	// Set image pull policy based on user input, if specified.
 	var configuredPolicy *v1.PullPolicy
 	if installationSpec != nil {
 		configuredPolicy = installationSpec.ImagePullPolicy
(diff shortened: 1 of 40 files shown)


pkg/controller/utils/component.go
```diff
@@ -230,8 +231,27 @@ func (c *componentHandler) createOrUpdateObject(ctx context.Context, obj client.
 	// system as specified by the osType.
 	ensureOSSchedulingRestrictions(obj, osType)
 
-	// Make sure any objects with images also have an image pull policy.
-	modifyPodSpec(obj, setImagePullPolicy)
+	// Look up the InstallationSpec once and reuse it for the passes that need it
+	// (image pull policy and TLS ciphers), so we don't pay for the same Get twice.
+	// Objects without a pod spec (e.g. NetworkPolicy, Tier) don't need it.
+	var installationSpec *operatorv1.InstallationSpec
+	if hasPodSpec(obj) {
```
caseydavenport: I think his hasPodSpec is probably an over-optimization - this API call is fed from the cache, so it's really cheap. Better to avoid the special case.

pkg/controller/utils/component.go
```diff
@@ -230,8 +231,27 @@ func (c *componentHandler) createOrUpdateObject(ctx context.Context, obj client.
...
+		if _, spec, err := GetInstallationSpec(ctx, c.client); err == nil {
+			installationSpec = spec
+		}
+	}
+
+	// Make sure any objects with images also have an image pull policy. If the user has
+	// configured a pull policy on the Installation, override every container's policy with
+	// it — the user's choice (e.g. for air-gapped clusters) must win over operator defaults.
+	// Otherwise, fill in IfNotPresent for any container that does not already set a policy.
```
caseydavenport:
```suggestion
  // Set image pull policy based on user input, if specified. 
```

rene-dekker (approved):

PR #4798, opened 2026-05-11
Add RBAC for the new ui-apis ManagedCluster / ClusterInformation handlers

Brian-McM:
## Description

The cloud security fix in calico-private (PR #11546) moves UISettings delete, ManagedCluster CRUD, and ClusterInformation reads off the user-impersonated /apis/projectcalico.org/v3 proxy path and onto ui-apis handlers that call the Calico API with their own service-account token. The tigera-manager pod SA therefore needs new grants on these resources.

Changes to managerClusterRole (pkg/render/manager.go):
- uisettings / uisettingsgroups / uisettingsgroups/data — add "delete" (previously only get/list/watch).
- managedclusters — add "create" and "delete" verbs (previously only get/list/watch). "update" continues to live in managedClustersUpdateRBAC().
- clusterinformations — new rule with "get" and "list" so the UI can surface the management cluster version through ui-apis.

manager_test.go updated at both assertion sites to match.

## Release Note

```release-note
TBD
```

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/manager.go
+++ b/pkg/render/manager.go
@@ -994,7 +994,7 @@ func managerClusterRole(managedCluster bool, kubernetesProvider operatorv1.Provi
 				Verbs: []string{"list"},
 			},
 			{
-				// ui-apis needs broad read access to UISettings and UISettingsGroups to serve
+				// ui-apis needs read access to UISettings and UISettingsGroups to serve
 				// requests on behalf of users. It performs SubjectAccessReviews to enforce
 				// per-group RBAC before returning results.
 				APIGroups: []string{"projectcalico.org"},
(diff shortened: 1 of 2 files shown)


rene-dekker (approved):

pkg/render/manager.go
```diff
@@ -999,14 +999,23 @@ func managerClusterRole(managedCluster bool, kubernetesProvider operatorv1.Provi
...
 				APIGroups: []string{"projectcalico.org"},
 				Resources: []string{
 					"uisettings",
 					"uisettingsgroups",
 					"uisettingsgroups/data",
 				},
-				Verbs: []string{"get", "list", "watch"},
+				Verbs: []string{"get", "list", "watch", "delete"},
+			},
+			{
```
caseydavenport, line 1019: In the PR implementing this, you have a toggle for whether ui-apis needs these permissions. Should we respect that toggle here as well? 

Or, should we unconditionally switch over to the new mode?

Brian-McM: I think it's safe to add these permissions in general for the ui-apis so I'd rather not add an extra flag here.

We just added that other flag because we didn't want to break auditing for enterprise clusters by fully going the route of no impersonation here.

PR #4805, opened 2026-05-13
Adopt the ko-built calico image

caseydavenport:
Two changes needed in the operator before the calico/calico image can ship from ko (see projectcalico/calico#12710):

- The combined-image command path switches from `/usr/bin/calico` to a bare `calico` so it resolves via PATH. The legacy Dockerfile build puts the binary at `/usr/bin/calico`; the ko-built image puts it at `/ko-app/calico`. Both directories are on PATH, so manifests, probes, init containers, and lifecycle hooks all work either way. Same treatment for the csi-node-driver-registrar command.
- The CSI driver now picks up csi-node-driver-registrar from a standalone OSS image (`calico/node-driver-registrar`), since the ko-built calico/calico no longer carries the binary. Enterprise keeps using the combined `tigera/calico` image (registrar is still bundled there).

```release-note
None
```


caseydavenport:
Closing - depends on https://github.com/projectcalico/calico/pull/12710 which was closed (benchmark numbers didn't justify the blast radius).

PR #4819, opened 2026-05-18
Add cni-plugins init container to calico-node DaemonSet

caseydavenport:
Pair with the new `calico/third-party-cni-plugins` image from projectcalico/calico (calico#12794). After the mono-image consolidation, `calico/calico` no longer ships the upstream CNI plugin binaries baked into `/opt/cni/bin`. On platforms that don't pre-populate the host CNI bin dir, `install-cni` finds nothing to copy and pods depending on `host-local`, `portmap`, `loopback`, `tuning`, or `flannel` break.

This change adds a `cni-plugins` init container ahead of `install-cni` that runs the new image. Both share a `cni-plugins-stage` emptyDir; the cni-plugins entrypoint copies the binaries into it, and install-cni mounts the same emptyDir at `/opt/cni/bin` so its existing walk picks them up unchanged. No change to install-cni's behavior.

## Description

- `pkg/components`: `ComponentCalicoCNIPlugins` (OSS, `calico/third-party-cni-plugins`) and `ComponentTigeraCNIPlugins` (enterprise, `tigera/third-party-cni-plugins`), with the version templates and config yamls updated to match.
- `pkg/render/node.go`: resolves the cni-plugins image based on variant, adds the `cni-plugins-stage` emptyDir, renders the init container ordered before install-cni, and adds the staging mount at `/opt/cni/bin` on install-cni.
- `api/v1`: adds `CNI.InstallMode`, which defaults to `All` (run the cni-plugins init container). `CalicoOnly` skips it and installs only Calico's own binaries, for hosts that already provide the upstream plugins (kind, some managed node images).
- `pkg/render/node_test.go`: covers the new container, mount, ordering, InstallMode gating, and OSS / Enterprise image selection.

calico#12794 has merged. The init container image is published to docker.io but still needs promoting to quay.io before the operator FV (and default installs) can pull it; until then the init container's image pull fails.

```release-note
NONE
```


tmjd (approved):

pkg/render/node.go
```diff
@@ -184,6 +185,13 @@ func (c *nodeComponent) ResolveImages(is *operatorv1.ImageSet) error {
 	}
 
 	c.calicoImage = appendIfErr(components.GetReference(components.CombinedCalicoImage(c.cfg.Installation), reg, path, prefix, is))
+	if c.installUpstreamPlugins() {
+		if c.cfg.Installation.Variant.IsEnterprise() {
+			c.cniPluginsImage = appendIfErr(components.GetReference(components.ComponentTigeraCNIPlugins, reg, path, prefix, is))
+		} else {
+			c.cniPluginsImage = appendIfErr(components.GetReference(components.ComponentCalicoCNIPlugins, reg, path, prefix, is))
+		}
+	}
```
electricjesus, line 194: This hard-fails the reconcile for anyone pinning an ImageSet that doesn't list `third-party-cni-plugins`: `GetReference` returns `ImageSet did not contain image ...` and `ResolveImages` bails. Release-built ImageSets pick it up automatically, so it's only the hand-maintained ones, but that's the air-gapped crowd and they'll hit it with no obvious cause. Worth calling out as an upgrade note in the release note so it isn't invisible.

pkg/render/node.go
```diff
@@ -1215,6 +1238,35 @@ func (c *nodeComponent) cniContainer() corev1.Container {
...
+}
+
+// cniPluginsContainer creates the init container that stages upstream CNI
+// plugin binaries (host-local, portmap, loopback, tuning, flannel) into a
+// shared volume read by the install-cni init container. The plugins ship as
+// a separate image rather than baked into the combined calico image so the
+// main image stays small.
+func (c *nodeComponent) cniPluginsContainer() corev1.Container {
+	return corev1.Container{
+		Name:            "cni-plugins",
```
electricjesus, line 1261: nit: this container sets no Command/args, so the `/stage` mount only works because it matches the image's default `dst` (the calico install binary defaults `src=/plugins dst=/stage`). If that default ever moves on the calico side this breaks silently: no error, the plugins just don't get staged. calico#12794 has a UT pinning the default so it's covered today, but passing `STAGE_DIR=/stage` as env here would make the contract explicit at the call site. Optional.

electricjesus (commented):
release note suggestion, since this adds a new image pull on every default install + the new `CNI.InstallMode` API field:

```release-note
Calico now stages the upstream CNI plugins (host-local, portmap, loopback, tuning, flannel) from a separate third-party-cni-plugins image instead of baking them into the main image. Set CNI.InstallMode: CalicoOnly to skip it on hosts that already provide them. ImageSet users must add the new image.
```

PR #4821, opened 2026-05-19
feat(applicationlayer): WAF v3 (Coraza WASM) render — kube-controllers plumbing + admission webhook

electricjesus:
## Summary

Operator-side render for WAF v3 (Coraza WASM) on `calico-kube-controllers`. Pairs with the merged reconcilers (`tigera/calico-private#11834`) and the in-process SecLang validating admission webhook (`tigera/calico-private#12141`, EV-6657). Design: `tigera/designs#25` (PMREQ-384).

**Review scope — 3 commits, ~470 lines:**

| Commit | What |
|---|---|
| `feat(api): add GatewayAPI WAF extension gating field` | `GatewayAPI.spec.extensions.waf.state` (`Enabled`/`Disabled`, default off) + deepcopy + CRD |
| `feat(applicationlayer): render WAF v3 + in-process admission webhook` | `gateway_waf.go` (webhook Service + VWC), kube-controllers env/RBAC/cert/port, `coraza-wasm` component |
| `feat(applicationlayer): wire WAF v3 render + webhook into installation controller` | cert issuance + render wiring in core controller |

**Everything is gated** on `GatewayAPI.spec.extensions.waf.state == Enabled` — non-WAF / OSS / WAF-disabled installs render a byte-identical kube-controllers Deployment.

> **Merge order (why `hold merge`):** the webhook render is `failurePolicy: Fail` against the in-process server on `:9443`, which ships in `tigera/calico-private#12141` (open). Merging this before cp#12141 is in the hashrelease kube-controllers image would brick `wafplugins`/`wafpolicies` writes once WAF is enabled. **Review now; merge lands right after cp#12141.** The plumbing half pairs with already-merged cp#11834.

### WAF SecLang admission webhook — in-process (calico-kube-controllers)

The webhook runs **in-process** inside the existing `calico-kube-controllers` Pod (the applicationlayer manager's webhook server) — **not** a standalone Deployment. `pkg/render/applicationlayer/gateway_waf.go` renders only:

- a `Service` (`tigera-waf-webhook`) fronting the kube-controllers Pod (`:443 → :9443`), and
- a `ValidatingWebhookConfiguration` intercepting CREATE/UPDATE on `wafplugins` + `wafpolicies` at `/validate-waf`, `failurePolicy: Fail` (the reconciler backstop is status-only, so the webhook is the hard admission gate), caBundle = the operator CA.

No dedicated ServiceAccount/ClusterRole/ClusterRoleBinding/Deployment — the webhook reuses the kube-controllers ServiceAccount + ClusterRole.

### kube-controllers plumbing (gated on `GatewayAPI.spec.extensions.waf.state == Enabled`)

- `WASM_IMAGE` / `WASM_PULL_SECRET` / `WASM_CA_CERT` env + `ENABLED_CONTROLLERS=applicationlayer` (names verified against the merged reconcilers' `manager.go`).
- WAF reconciler RBAC: the 6 `applicationlayer.projectcalico.org` resources + `/status` + `/finalizers`, EnvoyExtensionPolicy CRUD, events (core + `events.k8s.io`), secret/ConfigMap replication, Gateway/HTTPRoute reads for targetRef validation, `namespaces` patch/update for the upcoming per-namespace `waf-id-range` annotation.
- The webhook serving cert (`WAFWebhookServerTLS`, issued for `tigera-waf-webhook.calico-system.svc`) mounted into the Pod; `WAF_WEBHOOK_CERT_DIR` env; container port `9443`.

### Controller wiring (installation core controller)

Issues the serving cert via CertificateManager, materializes it into `calico-system` through the existing CertificateManagement render, threads it into the kube-controllers config, and renders the webhook Service + ValidatingWebhookConfiguration — all behind the WAF-enabled gate.

## Test plan

- [x] `go test ./pkg/render/applicationlayer/... ./pkg/render/kubecontrollers/...` — webhook contract (resources/path/failurePolicy/caBundle), kube-controllers cert mount + env + port, RBAC.
- [x] `go build ./...`, `go vet`, `gofmt` clean; full Operator CI green (2424 tests).
- [x] RBAC audited against the merged reconcilers' kubebuilder markers + client-call sweep — no missing grants (incl. the controller-runtime cached-client list/watch for Gateway/HTTPRoute Gets, and `*/finalizers` for OCP `OwnerReferencesPermissionEnforcement`).
- [ ] End-to-end admission round-trip on a cluster (apiserver → in-process webhook TLS handshake) once cp#12141 is in a hashrelease image.

## Release Note

```release-note
Add operator render for the WAF v3 (Coraza WASM) SecLang validating admission
webhook — served in-process by calico-kube-controllers — plus the WASM_* env
vars and serving-cert / RBAC plumbing (paired with tigera/calico-private#11834
and #12141). The existing WAF v1 (sidecar / ModSecurity) render path is untouched.
```

## Linked

- Jira: EV-6462 (operator integration) + EV-6657 (admission webhook)
- Design: `tigera/designs#25` (PMREQ-384)
- calico-private reconcilers: `tigera/calico-private#11834` (**merged**)
- SecLang admission webhook: `tigera/calico-private#12141` (EV-6657, open — **must merge before this PR**)
- Hashrelease CRD-update RBAC hotfix (operator install manifest side): `tigera/calico-private#12215` (**merged**)
- Split out of this PR: #4895 (EV-6650 observability, stacked on this), #4894 (envoy-gateway NetworkPolicy dual-stack fix)
- Builds on: #4690 (merged)


api/v1/gatewayapi_types.go
```diff
@@ -79,6 +79,51 @@ type GatewayAPISpec struct {
...
+// WAFExtensionState is the on/off enum for the WAF Gateway API add-on.
+// +kubebuilder:validation:Enum=Enabled;Disabled
+type WAFExtensionState string
+
+const (
+	WAFExtensionStateEnabled  WAFExtensionState = "Enabled"
+	WAFExtensionStateDisabled WAFExtensionState = "Disabled"
+)
+
+// IsWAFGatewayExtensionEnabled returns true iff spec.extensions.waf.state == Enabled.
```
rene-dekker:
```suggestion
// IsWAFGatewayExtensionEnabled returns true if spec.extensions.waf.state == Enabled.
```

hack/gen-versions/enterprise.go.tpl
```diff
@@ -180,6 +180,15 @@ var (
 		variant:   enterpriseVariant,
 	}
 {{- end }}
+{{ with index .Components "coraza-wasm" }}
```
rene-dekker: Can you remind me if we could put this in an existing image and then change the run cmd when it is used. What was the motivation for an extra image again?

pkg/common/common.go
```diff
@@ -36,6 +36,10 @@ const (
 	EgressAccessControlFeature = "egress-access-control"
 	// PolicyRecommendation feature name
 	PolicyRecommendationFeature = "policy-recommendation"
+	// GatewayAddonsFeature gates Tigera-built add-ons that layer on top of an
+	// ingress gateway (currently the WAF v2/v3 admission webhook). The bare
+	// ingress gateway data path is NOT licensed by this feature.
+	GatewayAddonsFeature = "ingress-gateway-addons"
```
rene-dekker: I believe this is currently unsused in the code (that's what my IDE tells me).

pkg/controller/installation/core_controller.go
```diff
@@ -1361,18 +1360,56 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
...
+		r.status.SetDegraded(operatorv1.ResourceReadError, "Error reading GatewayAPI", err, reqLogger)
+		return reconcile.Result{}, err
+	}
+
+	// When the WAF v3 surface is enabled, issue the serving cert for the
+	// in-process WAF admission webhook (hosted by calico-kube-controllers,
+	// fronted by the tigera-waf-webhook Service). It is materialized into
+	// calico-system alongside the other kube-controllers certs below and mounted
+	// into the Pod by the kube-controllers render.
+	var wafWebhookTLS certificatemanagement.KeyPairInterface
```
rene-dekker, line 1394: Could we just reuse the kubecontrollers certificate and add append the SANs required for the webhook? Is there a compelling reason why we shouldn't? It would allow us to make the PR significantly smaller.

pkg/controller/installation/core_controller.go
```diff
@@ -1361,18 +1360,56 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
...
+
+	keyPairOptions := []rcertificatemanagement.KeyPairOption{
+		rcertificatemanagement.NewKeyPairOption(typhaNodeTLS.NodeSecret, true, true),
+		rcertificatemanagement.NewKeyPairOption(nodePrometheusTLS, true, true),
+		rcertificatemanagement.NewKeyPairOption(typhaNodeTLS.TyphaSecret, true, true),
+		rcertificatemanagement.NewKeyPairOption(typhaNodeTLS.TyphaSecretNonClusterHost, true, true),
+		rcertificatemanagement.NewKeyPairOption(kubeControllerTLS, true, true),
+	}
+	if wafWebhookTLS != nil {
+		keyPairOptions = append(keyPairOptions, rcertificatemanagement.NewKeyPairOption(wafWebhookTLS, true, true))
```
rene-dekker: If you decide to keep this TLS and not reuse the existing cert, then you can still make the code simpler. If you pass a nil cert into NewKeyPairOption, it will just skip it. No need for the if-statement.

pkg/controller/installation/core_controller.go
```diff
@@ -1619,9 +1656,19 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
...
+		WAFGatewayExtensionEnabled:  wafGatewayExtensionEnabled,
+		WAFWebhookServerTLS:         wafWebhookTLS,
 	}
 	components = append(components, kubecontrollers.NewCalicoKubeControllers(&kubeControllersCfg))
 
+	// Render the in-process WAF admission webhook Service + ValidatingWebhookConfiguration.
+	// The webhook is served by calico-kube-controllers; the caBundle is the
+	// operator CA that issued the serving cert above.
+	if wafGatewayExtensionEnabled {
+		components = append(components, render.NewPassthrough(
```
rene-dekker: What is the reason that this isn't in the render code for the kube controllers deployment? With this pattern these objects will never go away even if the feature is disabled or the GatewayAPI is removed.

It would be good to add test coverage for deletion.

pkg/render/kubecontrollers/kube-controllers.go
```diff
@@ -55,6 +56,12 @@ const (
 	KubeControllerMetrics           = "calico-kube-controllers-metrics"
 	KubeControllerNetworkPolicyName = networkpolicy.CalicoComponentPolicyPrefix + "kube-controller-access"
 
+	// wafWebhookContainerPort is the in-process WAF admission-webhook server
+	// port on calico-kube-controllers. Must match the TargetPort of the
+	// tigera-waf-webhook Service (see pkg/render/applicationlayer) and the port
+	// the calico-private applicationlayer manager's webhook server listens on.
+	wafWebhookContainerPort = int32(9443)
```
rene-dekker: I feel like this should be removed and we should use an exported variable both in service creation and in kube controllers. (Or move all render code to this file.)

pkg/render/kubecontrollers/kube-controllers.go
```diff
@@ -155,6 +179,9 @@ func NewCalicoKubeControllers(cfg *KubeControllersConfiguration) *kubeController
 			},
 		)
 		enabledControllers = append(enabledControllers, "service", "federatedservices", "usage")
+		if cfg.WAFGatewayExtensionEnabled {
+			enabledControllers = append(enabledControllers, "applicationlayer")
```
rene-dekker, line 201: Maybe it is me, but I expected a different name for this new controller, like gateway-waf or something like that.

pkg/render/applicationlayer/gateway_waf.go
```diff
@@ -0,0 +1,155 @@
...
+}
+
+// wafValidatingWebhookConfiguration rejects unsafe AO-supplied SecLang at
+// admission. It intercepts CREATE/UPDATE on WAFPlugin and WAFPolicy (the
+// resources that carry AO SecLang) and fails closed: FailurePolicy=Fail so an
+// unavailable webhook blocks the (infrequent) WAF resource writes rather than
+// admitting unvalidated directives. The in-cluster reconciler backstop is
+// status-only, so the webhook is the hard admission gate.
+func wafValidatingWebhookConfiguration(caBundle []byte) *admissionregistrationv1.ValidatingWebhookConfiguration {
+	failPolicy := admissionregistrationv1.Fail
```
rene-dekker, line 111: Just a thought:
- It looks like the kube controllers will be able to still reconcile a missed webhook call, avoiding bad config to be deployed on the WAF.
- It would allow us to use fail policy Ignore.

The down-side is that you may want to immediately have feedback written to the status. 

How do you view this trade-off? I'm fine with the current approach, because it is unlikely that this webhook will brick a cluster when unavailable.

rene-dekker:
Nit: It would be possible to cut down on the code comments. Especially when the comments don't add any information you couldn't get directly from reading the code.
Everyone has AI agents now, so the reader could even consult these to help them read/navigate the code.

pkg/render/kubecontrollers/kube-controllers.go
```diff
@@ -571,6 +699,39 @@ func (c *kubeControllersComponent) controllersDeployment() *appsv1.Deployment {
...
+			}
+
+			// WASM_PULL_SECRET names the imagePullSecret the reconciler replicates
+			// from the kube-controllers namespace into a WAFPolicy's namespace so
+			// the rendered EnvoyExtensionPolicy can pull the wasm OCI artifact from
+			// a private Tigera registry. Source the name from the first
+			// Installation.ImagePullSecrets entry so multi-tenant / BYO-registry
+			// installs reuse whatever pull secret operator already attaches here.
+			if len(c.cfg.Installation.ImagePullSecrets) > 0 {
+				env = append(env, corev1.EnvVar{Name: "WASM_PULL_SECRET", Value: c.cfg.Installation.ImagePullSecrets[0].Name})
```
rene-dekker: The `[0]` is a bit unfortunate, but I suppose it cannot be helped. I wonder if it merits a warning log / info log in the core controller or a doc note when this feature is enabled and more than 1 pull secret is set to the installation spec.

pkg/controller/installation/core_controller.go
```diff
@@ -1361,18 +1360,56 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
...
 
+	// Read the GatewayAPI CR (if present) to decide whether to render the WAF
+	// v3 (Gateway API add-on) surface — env vars, RBAC, applicationlayer
+	// reconciler, and the in-process admission webhook — on
+	// calico-kube-controllers. Default-off: if no GatewayAPI CR exists or
+	// spec.extensions.waf.state != Enabled, the WAF surface is not rendered.
+	// See design tigera/designs#25 (PMREQ-384) §Gating.
+	wafGatewayExtensionEnabled := false
+	gatewayAPI := &operatorv1.GatewayAPI{}
+	if err := r.client.Get(ctx, utils.DefaultInstanceKey, gatewayAPI); err == nil {
```
rene-dekker: Please use gatewayapi.GetGatewayAPI, which will check for both default and tigera-secure as the name.

pkg/render/applicationlayer/gateway_waf.go
```diff
@@ -0,0 +1,155 @@
...
+					{
+						Operations: []admissionregistrationv1.OperationType{
+							admissionregistrationv1.Create,
+							admissionregistrationv1.Update,
+						},
+						Rule: admissionregistrationv1.Rule{
+							APIGroups:   []string{"applicationlayer.projectcalico.org"},
+							APIVersions: []string{"v3"},
+							Resources:   []string{"wafplugins", "wafpolicies"},
+							Scope:       ptr.To(admissionregistrationv1.NamespacedScope),
```
rene-dekker, line 137: Is namespace scope correct here?

hack/gen-versions/enterprise.go.tpl
electricjesus: we needed this extra image because this is an oci image that the envoy gateway controller consumes and distributes it out to gateway proxies. it's  a scratch oci image with just 1 file, a .wasm file which contains coraza engine + rules. needs to be small so it gets distributed fast, i think we set our benchmark to be roughly 30mb, we're well under that i think

api/v1/gatewayapi_types.go
electricjesus: took it.

pkg/common/common.go
electricjesus: good catch, that's a leftover. the plan originally was to gate this render on the `ingress-gateway-addons` license feature, but we moved the calico-private side to a binary license-validity check (Valid or InGracePeriod) instead of feature lookups, and i forgot to remove the constant when that changed. dropped now.

pkg/controller/installation/core_controller.go
electricjesus: i kept it separate on purpose. the VWC caBundle has to match whatever CA issued the serving cert, and the kube-controllers metrics cert can be BYO (user-supplied secret). if the webhook piggybacked on that cert, the apiserver would be checking a user-CA cert against the operator-CA bundle we stamp into the VWC and the handshake fails. a dedicated operator-issued cert keeps the issuer pinned to the operator CA. side benefit: toggling WAF doesn't add/remove SANs on the metrics cert and force a rotation.

pkg/controller/installation/core_controller.go
electricjesus: fixed, it's passed unconditionally now. didn't realize the cert-management render skips nil keypairs.

pkg/controller/installation/core_controller.go
electricjesus: moved the Service + VWC into the kube-controllers render, they go out on the delete side when the WAF extension is off or the GatewayAPI CR is gone. added a test for the disabled → delete path.

pkg/render/kubecontrollers/kube-controllers.go
electricjesus: done, exported `WAFWebhookContainerPort` from pkg/render/applicationlayer and the kube-controllers side (container port + the networkpolicy ingress rule) uses it now.

pkg/render/kubecontrollers/kube-controllers.go
electricjesus: the name comes from the merged calico-private side (calico-private#11834), `ENABLED_CONTROLLERS=applicationlayer` maps to `v3.ApplicationLayerControllerConfig` on KubeControllersConfiguration. it also manages the whole applicationlayer.projectcalico.org family (WAFPolicy/WAFPlugin/GlobalWAF*/GlobalWAFValidationPolicy), not just the gateway WAF bits, so gateway-waf would be too narrow if more applicationlayer resources land in that group. renaming at this point would be a v3 API change in calico-private.

pkg/render/applicationlayer/gateway_waf.go
electricjesus: agree on the data plane, the reconciler re-validates and never attaches invalid directives (deterministic `Accepted=False, reason=DirectivesInvalid`, no retry), so Ignore wouldn't deploy bad config either. what Fail buys us is feedback at apply time (CI/gitops fails at kubectl apply instead of polling status) and invalid seclang never sitting in etcd. and since the webhook only covers the two WAF kinds, a dead kube-controllers blocks WAF writes only, nothing cluster-critical, so i went with the stricter default. happy to flip to Ignore if availability turns out to bite.

pkg/render/kubecontrollers/kube-controllers.go
electricjesus: WASM_PULL_SECRET is single-valued on the kube-controllers side, it's one secret name the reconciler replicates into WAFPolicy namespaces. and a heads up from cluster testing (EV-6386): the raw [0] reference clashed with the operator-managed tigera-pull-secret copy in tenant namespaces (ReplicaUnmanaged), so the operator now provisions a dedicated `tigera-waf-pull-secret`, a renamed copy of the first install secret. also added a log line when more than one pull secret is configured.

pkg/controller/installation/core_controller.go
electricjesus: done. a default/tigera-secure duplicate now degrades with the same message the gatewayapi controller uses.

pkg/render/applicationlayer/gateway_waf.go
electricjesus: yes, wafpolicies and wafplugins are both namespaced CRDs, the cluster-scoped ones are separate resources (globalwafpolicies/globalwafplugins). we deliberately don't intercept the global kinds yet, the webhook handler in calico-private#12141 only validates WAFPlugin today and the global kinds fall back to the reconciler status backstop. once the handler learns GlobalWAFPlugin we'll add a ClusterScope rule next to this one.

pkg/render/kubecontrollers/kube-controllers.go
electricjesus: follow-up: got rid of the [0] entirely in 32171cdb3. since the EnvoyExtensionPolicy image source only takes a single pullSecretRef, the operator now merges the registry auths of all Installation pull secrets into `tigera-waf-pull-secret` instead of copying just the first one. covers the realistic multi-secret case (tigera pull secret + private mirror creds, disjoint registries). if the same registry shows up in two secrets the first one wins, unparseable secrets are skipped and logged rather than failing the reconcile.

hack/gen-versions/enterprise.go.tpl
rene-dekker: It adds extra maintenance on our side and the customer when we add extra images. We could add the WASM image to an existing image like envoyproxy and eliminate needing to pull anything altogether, which would result in better performance. File loading is also supported, we could even point to a file path, rather than an image path.

rene-dekker (approved):

PR #4826, opened 2026-05-19
fix(gatewayapi): inject CALICO_API_GROUP into waf-http-filter and l7-log-collector

electricjesus:
## Description

Bug fix for **EV-6620** (no-api-server mode: empty CIG/WAF dashboards because the l7-log-collector's license check fails).

**Why:** the `waf-http-filter` and `l7-log-collector` containers are templated into an `EnvoyProxy` CR (rendered at `pkg/render/gatewayapi/gateway_api.go:889`) rather than created as a Deployment/DaemonSet directly. The operator's `componentHandler.injectAPIGroupEnv` walks `Deployment/DaemonSet/StatefulSet/Job/CronJob` — it never recurses into `EnvoyProxy.Spec.Provider.Kubernetes.EnvoyDeployment.InitContainers`. So these two sidecars never get `CALICO_API_GROUP` stamped on them and fall through to libcalico-go's auto-discovery in `clientv3.NewFromEnv()`.

In **no-api-server mode** the operator installs both `crd.projectcalico.org/v1` and `projectcalico.org/v3` CRD groups. `UsingV3CRDs()` sees both groups present and (per its old `v3present && !v1present` heuristic) picks v1. But `kubectl apply` writes `LicenseKey` to the newer group → v3. The license lookup then 404s in v1 and `IsLicensed()` returns `false`. l7-log-collector silently drops every log line; waf-http-filter silently routes through the unlicensed request processor.

**What changed:** append `apigroup.EnvVars()` (which yields `CALICO_API_GROUP=projectcalico.org/v3` when the operator already determined v3 mode at startup, otherwise nil) to the `Env` slice of each init container at render time. Discovery is bypassed entirely on both sidecars — no extra RBAC, no extra API roundtrips, deterministic.

**Components affected:** `pkg/render/gatewayapi` only.

**Testing:**
- New unit specs in `gateway_api_test.go`: `should inject CALICO_API_GROUP …` and `should not inject CALICO_API_GROUP when v3 CRDs are not in use`. Both pass alongside the existing 20 specs.
- Verified on cluster `seth-ez-82a4` (no-api-server, both v1+v3 groups present, APIServer CR absent) that the existing pods lack the env var and exhibit the bug. After this fix, fresh pods will receive `CALICO_API_GROUP=projectcalico.org/v3` and skip the buggy discovery path.

### Related but out of scope

- `pkg/apis/version.go:78` has the **same** `v3present && !v1present` heuristic. In EV-6620 clusters the operator only avoids returning `false` because `checkDatastoreMigration` short-circuits when a `DatastoreMigration` CR is in `PhaseConverged`/`PhaseComplete`. Fresh `USE_API_SERVER=false` installs without a migration CR would hit the same bug; worth a follow-up.
- calico-private libcalico-go discovery fix (#11986) is defense-in-depth for any libcalico-go client that doesn't get an injected env var (calicoctl, third-party tooling). Not needed for the gateway path once this lands.

## Release Note

```release-note
Fixed empty CIG and WAF dashboards in no-api-server clusters by injecting CALICO_API_GROUP into the waf-http-filter and l7-log-collector init containers, so libcalico-go skips the auto-discovery path that picks the wrong Calico CRD group when both v1 and v3 groups are installed.
```

## For PR author

- [x] Tests for change.
- [ ] If changing pkg/apis/, run `make gen-files` — N/A
- [ ] If changing versions, run `make gen-versions` — N/A

### DIFF
--- a/pkg/render/gatewayapi/gateway_api.go
+++ b/pkg/render/gatewayapi/gateway_api.go
@@ -23,6 +23,7 @@ import (
 
 	envoyapi "github.com/envoyproxy/gateway/api/v1alpha1"
 	operatorv1 "github.com/tigera/operator/api/v1"
+	"github.com/tigera/operator/pkg/apigroup"
 	"github.com/tigera/operator/pkg/common"
 	"github.com/tigera/operator/pkg/components"
 	"github.com/tigera/operator/pkg/render"
@@ -843,17 +844,20 @@ func (pr *gatewayAPIImplementationComponent) envoyProxyConfig(className string, 
 						MountPath: "/var/log/calico",
 					},
 				},
-				Env: []corev1.EnvVar{
+				// apigroup.EnvVars() injects CALICO_API_GROUP so libcalico-go skips
+				// auto-discovery. Required because these sidecars are templated into
+				// an EnvoyProxy CR and bypass the operator's componentHandler path.
+				Env: append([]corev1.EnvVar{
 					GatewayNameEnvVar,
 					GatewayNamespaceEnvVar,
-				},
+				}, apigroup.EnvVars()...),
 				SecurityContext: securitycontext.NewRootContext(true),
 			}
 			// need to make changes to the envoy container to mount the socket
 			l7LogCollector := corev1.Container{
 				Name:  "l7-log-collector",
 				Image: pr.L7LogCollectorImage,
-				Env: []corev1.EnvVar{
+				Env: append([]corev1.EnvVar{
 					{
 						Name:  "LOG_LEVEL",
 						Value: "info",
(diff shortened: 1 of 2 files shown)


electricjesus:
Closing in favor of moving the license gate (and the Gateway/HTTPRoute enrichment that l7-collector does today) into felix, so the sidecars hold no v3 / gateway-api client at all and the v1/v3 discovery ambiguity stops being a sidecar problem. Sketch in seth/code/EV-6620-felix-migration-sketch.md; will pull together a design doc once Rene confirms the delay. calico-private#11986 stays open as defense-in-depth for any other libcalico-go consumer.

PR #4837, opened 2026-05-20
Discover served MutatingAdmissionPolicy API version at runtime

caseydavenport:
## Description

The operator hardcoded the use of `admissionregistration.k8s.io/v1beta1` for MutatingAdmissionPolicy. K8s 1.36 promoted MutatingAdmissionPolicy to `v1` and v1beta1 is scheduled for removal in 1.37, so clusters that only serve v1 produced reconcile errors like `no matches for kind "MutatingAdmissionPolicy" in version "admissionregistration.k8s.io/v1beta1"` (see https://github.com/projectcalico/calico/issues/6412#issuecomment-4465188064).

This PR discovers the served version via the RESTMapper and uses it for parsing, listing, creating, and stale-cleanup. Preference order is v1, then v1beta1. If neither is served the controller reports a degraded condition and skips policy defaulting, same as before.

Split into two commits:

1. Bump `k8s.io/*` to `v0.36.1` and `sigs.k8s.io/controller-runtime` to `v0.24.1` so the v1 typed import is available.
2. Add discovery + dual typed paths in `pkg/imports/admission` and `pkg/controller/installation/core_controller.go`. The previous static `ProvidesMutatingAdmissionPolicyV1Beta1` minor-version check is replaced by RESTMapper-based discovery, which is more accurate across forks and feature gates.

```release-note
Fix operator reconcile failure on Kubernetes clusters that only serve the v1 (not v1beta1) MutatingAdmissionPolicy API.
```


cmd/main.go
```diff
@@ -507,8 +512,10 @@ If a value other than 'all' is specified, the first CRD with a prefix of the spe
 		ShutdownContext:     ctx,
 		K8sClientset:        clientset,
 		MultiTenant:         multiTenant,
-		ElasticExternal:     utils.UseExternalElastic(bootConfig),
+		ElasticExternal:     discovery.UseExternalElastic(bootConfig),
 		UseV3CRDs:           v3CRDs,
+
```
caseydavenport:
```suggestion
```

mohag:
A somewhat tangential note: The v3 CRD chart also have v1beta1 MutatingAdmissionPolicy manifests pulled in from [here](https://github.com/projectcalico/calico/tree/master/api/admission).

caseydavenport:
@mohag yep, will definitely need to handle those as well (I believe helm can detect API versions and use the correct one, with some tweaks)

rene-dekker (approved):

PR #4865, opened 2026-05-26
Add opt-in RBAC management UI via rbacManagement flag

dimitri-nicolo:
## Description

**Type:** New feature (Calico Enterprise only). Addresses PMREQ-824.

Adds an opt-in **RBAC management UI** feature, toggled by a new `Manager.spec.rbac.ui` field (`Enabled` | `Disabled`, **defaults to `Disabled`**). When enabled, the operator renders the additional RBAC and network access the UI needs to let an administrator manage role/group assignments from the Manager: maintaining a catalog of managed `ClusterRoles`, binding IdP groups to roles via the `rbacsync` controller, and resolving groups from an external LDAP/AD directory.

This should be merged because it is the operator-side enablement for the RBAC management UI; without it the UI's backend has no RBAC or egress to function. Gating on `spec.rbac.ui` keeps the feature — and its escalation-capable permissions — entirely dormant on clusters that don't opt in.

### What the flag turns on, by layer

- **API** (`api/v1/manager_types.go`) — new `RBAC` spec struct with a `UI RBACUI` enum field and a nil-safe `Manager.RBACManagementEnabled()` helper (returns `false` for a nil Manager or any value other than `Enabled`). Regenerated deepcopy and the Manager CRD.
- **Render**
  - `calico-manager` role (`pkg/render/manager.go`): adds the cluster-scoped RBAC management UI rules (full CRUD on managed RBAC objects + escalation-prevention coverage for tier/resource roles), plus a **namespaced** `Role`+`RoleBinding` in `calico-system` that scopes the `create`/named-resource access to the IdP `ConfigMap`/`Secret`s (`tigera-idp-groups`, `tigera-idp-ldap-config`) instead of granting it cluster-wide. Sets `RBAC_UI_ENABLED` on the `ui-apis` container and opens LDAP/AD egress (389/636) when OIDC authentication is configured.
  - `calico-kube-controllers` (`pkg/render/kubecontrollers/kube-controllers.go`): enables the `rbacsync` controller and its RBAC only when the flag is set (dormant otherwise), and adds a namespaced `Role`+`RoleBinding` granting `rbacsync` read on the `tigera-idp-groups` `ConfigMap` in `calico-system`.
  - `tigera-network-admin` (`pkg/render/apiserver.go`): grants the `bind`/`escalate`-capable rule a network-admin needs to assign managed roles via the UI, also gated on the flag.
  - Shared escalation-prevention rules live in a new `pkg/render/rbac_management.go` so the manager and kube-controllers roles stay aligned (the SA writing a managed role must already hold every permission it grants, or K8s rejects the write as privilege escalation).

### Why the installation and apiserver controllers read the Manager CR

The flag conceptually belongs to the Manager, but two of the resources it gates are **not** rendered by the manager controller:

- The `rbacsync` controller runs **inside `calico-kube-controllers`**, rendered by the **installation** controller.
- The `tigera-network-admin` ClusterRole is rendered by the **apiserver** controller.

So each of those controllers reads the Manager CR itself and passes a `RBACManagementEnabled` bool into its render config. To support this, `GetManager` was moved out of the manager controller into `pkg/controller/utils` so the apiserver and installation controllers can reuse it without importing the manager controller package. Both callers treat a missing Manager as not-found (feature off) rather than an error, and the installation controller only reads it for the enterprise variant.

Both controllers also add a **watch** on the `Manager` CR. Without it, toggling `spec.rbac.ui` would not re-trigger an installation/apiserver reconcile, so the `rbacsync` controller and the network-admin rule wouldn't appear or disappear until some unrelated event re-ran those controllers.

### Per-cluster behavior

The feature is keyed off **each cluster's own** `Manager.spec.rbac.ui`, so a management cluster and a managed cluster opt in independently. Calico Enterprise evaluates a user's managed-cluster actions against that managed cluster's own RBAC (Voltron proxies the request to the managed cluster's apiserver), so the network-admin `bind`/`escalate` grant and the manager-SA rules are intentionally rendered on managed clusters too — that is where a user managing per-cluster RBAC needs them. In multi-tenant management clusters the manager-side rules and namespaced Role are not rendered (gated on `!MultiTenant()`).

**Testing:** unit tests added for all three render surfaces (manager, apiserver, kube-controllers) covering the enabled/disabled gate and the multi-tenant exclusion; `make gen-files` produces no drift; `go build ./...` and the affected `pkg/render*` / `pkg/controller/apiserver` suites pass.

**Components affected:** apiserver (`tigera-network-admin`), manager (`calico-manager` role + namespaced Role + network policy), kube-controllers (`rbacsync`), Manager CRD, installation & apiserver controllers, `pkg/controller/utils`.

## Release Note

```release-note
Added an opt-in `Manager.spec.rbacManagement.enabled` flag that enables the RBAC management UI. When enabled, the operator grants the additional RBAC and LDAP egress the UI requires; it defaults to disabled and is supported in zero-tenant management clusters only. Disabling the flag does not remove RBAC objects already created while it was enabled.
```

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


api/v1/manager_types.go
```diff
@@ -26,6 +26,31 @@ type ManagerSpec struct {
 	// ManagerDeployment configures the Manager Deployment.
 	// +optional
 	ManagerDeployment *ManagerDeployment `json:"managerDeployment,omitempty"`
+
+	// RBACManagement configures the RBAC management UI feature. Only honored
+	// in zero-tenant (non-multi-tenant) management clusters. Disabling this
+	// after enabling does not garbage-collect previously rendered RBAC
+	// objects; they remain on the cluster until removed manually.
+	// +optional
+	RBACManagement *RBACManagement `json:"rbacManagement,omitempty"`
```
rene-dekker: Nit: Manager.spec.RBACManagement is sort of double, redundant, what do you think of simply RBAC/rbac?

api/v1/manager_types.go
```diff
@@ -26,6 +26,31 @@ type ManagerSpec struct {
...
+}
+
+// RBACManagement controls the RBAC management UI feature surface.
+type RBACManagement struct {
+	// Enabled is the master switch for the RBAC management UI. Defaults to
+	// false.
+	// +optional
+	Enabled *bool `json:"enabled,omitempty"`
```
rene-dekker: Kubernetes recommends against boolean fields. We should do something like `mode`/`display`: Enabled|Disabled

pkg/controller/utils/utils.go
```diff
@@ -463,6 +463,25 @@ func GetApplicationLayer(ctx context.Context, c client.Client) (*operatorv1.Appl
...
 }
 
+// GetZeroTenantManagerOrNil returns the zero-tenant Manager CR (named
+// tigera-secure) if present, or nil with no error if absent. The Manager CR
+// is optional, so downstream renderers must not block when it isn't
+// installed. This helper deliberately ignores tenant namespaces: callers
+// that need a tenant-scoped Manager (the manager controller itself) must use
+// GetManager from pkg/controller/manager. Non-manager controllers reading
+// zero-tenant-only flags (e.g. spec.rbacManagement) belong here.
+func GetZeroTenantManagerOrNil(ctx context.Context, c client.Client) (*operatorv1.Manager, error) {
```
rene-dekker: Can't we move the GetManager func from manager_controller to the utils package and re-use it? There is no reason that apiserver should throw an error if manager is not present. We should however only fetch it when enterprise CRDs exist and we are not running in cloud.

api/v1/manager_types.go
dimitri-nicolo: Going to got with `RBAC`

api/v1/manager_types.go
```diff
@@ -26,6 +26,40 @@ type ManagerSpec struct {
 	// ManagerDeployment configures the Manager Deployment.
 	// +optional
 	ManagerDeployment *ManagerDeployment `json:"managerDeployment,omitempty"`
+
+	// RBAC configures the RBAC management UI feature. Only honored in
```
rene-dekker: I don't think we should leak cloud (internal) implications into the API comments. Our users don't need to be aware/think about those.

api/v1/manager_types.go
```diff
@@ -26,6 +26,40 @@ type ManagerSpec struct {
...
+	RBAC *RBAC `json:"rbac,omitempty"`
+}
+
+// RBAC controls the RBAC management UI feature surface.
+type RBAC struct {
+	// Mode controls whether the RBAC management UI is enabled. Defaults to
+	// Disabled.
+	// +optional
+	// +kubebuilder:validation:Enum=Enabled;Disabled
+	Mode RBACMode `json:"mode,omitempty"`
```
rene-dekker: What do you think of the following field names?
- Management: Enabled|Disabled
- UI: Enabled|Disabled

or values: Visible | Hidden

pkg/controller/utils/utils.go
rene-dekker: I see that you moved the func to here, but I also meant to delete this func entirely as well.

rene-dekker: I think the code is more clear if we delete this method, remove any notion of "zero tenant" from the code including comments (it's not yargon that is used today) and use the multitenant booleans explicitly at the caller. I think that will be more understandable to the average reader.

pkg/render/kubecontrollers/kube-controllers.go
```diff
@@ -155,6 +161,18 @@ func NewCalicoKubeControllers(cfg *KubeControllersConfiguration) *kubeController
 			},
 		)
 		enabledControllers = append(enabledControllers, "service", "federatedservices", "usage")
+
+		// RBAC management UI: when enabled on the Manager CR, the rbacsync
+		// controller runs inside calico-kube-controllers to reconcile the
+		// catalog of managed ClusterRoles (per-tier + 32 fine-grained + 6
```
rene-dekker: nit: This number is not likely to hold up over time, best to keep the comment succinct.

rene-dekker:
General comment: the AI generated comments in this code are a bit excessive. An extreme example may be the permissions. The func explains in comments twice which permissions are added + the code to add them. They are making the code less readable. And now we need to also keep the comments up to date when we add more. The operator code is not super complicated, I'd much rather have comments be only used if it adds critical information you couldn't get from simply reading the code (or have your ai read it for you).

pkg/render/manager.go
```diff
@@ -1174,6 +1187,68 @@ func managerClusterRole(managedCluster bool, kubernetesProvider operatorv1.Provi
...
+		// Cluster-wide create on ConfigMaps — K8s RBAC does not allow pairing
+		// the create verb with resourceNames, so this needs a separate rule.
+		rbacv1.PolicyRule{
+			APIGroups: []string{""},
+			Resources: []string{"configmaps"},
+			Verbs:     []string{"create"},
+		},
+		// Escalation coverage for webhook-mod resource roles (create/patch on
+		// Secrets to wire up webhook auth).
+		rbacv1.PolicyRule{
```
rene-dekker: Could we be more specific with secrets and configmaps and bind them to the namespaces we actually need?

pkg/render/manager.go
```diff
@@ -1244,6 +1319,21 @@ func (c *managerComponent) managerCalicoSystemNetworkPolicy() *v3.NetworkPolicy
 		})
 	}
 
+	// RBAC management UI: the feature walks an external LDAP directory to
+	// discover groups. The directory lives outside the cluster (or in an
+	// arbitrary namespace), so we allow egress to the standard LDAP ports
+	// 389 (ldap/StartTLS) and 636 (ldaps). Only opened when the feature is
+	// enabled so non-RBAC-UI clusters keep the tighter default egress.
+	if c.cfg.Manager.RBACManagementEnabled() {
```
rene-dekker: We could do this only when ldap is enabled.

dimitri-nicolo: I can follow up on this, but will leave this out for now

dimitri-nicolo: https://tigera.atlassian.net/browse/EV-6665

pkg/render/manager.go
dimitri-nicolo: The named-resource access moved to a new Role + RoleBinding in `calico-system`, where `tigera-idp-groups` and `tigera-idp-ldap-config` actually live. The broad `create` is bound to that one namespace now.

Dropped the `tigera-known-oidc-users` rule (it lives in `tigera-elasticsearch` and is already granted by `logstorage.go`), and tightened the activation gate to `!Tenant.MultiTenant()` since the IdP resources are pinned to `calico-system` and the RBAC UI is zero-tenant-only anyway.

Is that what you meant by "bind them to the namespaces we actually need" — or were you pointing at something narrower?

pkg/render/rbac_management.go
```diff
@@ -0,0 +1,103 @@
...
+		{
+			APIGroups: []string{"lma.tigera.io"},
+			Resources: []string{"cluster"},
+			Verbs:     []string{"get"},
+		},
+		// Read on the customer-owned IdP directory ConfigMap (drives the RBAC
+		// management UI's directory cache and the cascading cleanup of managed
+		// bindings when a group is removed).
+		{
+			APIGroups:     []string{""},
```
rene-dekker: This one should be in a role.

pkg/controller/installation/core_controller.go
```diff
@@ -1071,6 +1080,17 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 			return reconcile.Result{}, err
 		}
 
+		if instance.Spec.Variant.IsEnterprise() {
```
pasanw: Wondering if you found the variant check to be necessary here?

pkg/render/kubecontrollers/kube-controllers.go
```diff
@@ -658,6 +672,76 @@ func kubeControllersRoleEnterpriseCommonRules(cfg *KubeControllersConfiguration)
...
+
+// rbacSyncControllerRules returns the extra rules the rbacsync controller
+// needs on top of render.RBACManagementEscalationRules.
+func rbacSyncControllerRules() []rbacv1.PolicyRule {
+	rules := render.RBACManagementEscalationRules()
+	return append(rules,
+		rbacv1.PolicyRule{
+			APIGroups: []string{"rbac.authorization.k8s.io"},
+			Resources: []string{"clusterroles", "clusterrolebindings"},
+			Verbs:     []string{"escalate", "bind"},
```
pasanw: Is it possible to avoid giving escalate and bind privileges? It essentially means that this controller can become a superuser if it chooses to be (i.e. is compromised)

pkg/render/apiserver.go
```diff
@@ -2030,6 +2034,17 @@ func (c *apiServerComponent) tigeraNetworkAdminClusterRole() *rbacv1.ClusterRole
...
 	}
 
+	// Grant network-admin the role/binding access the RBAC management UI needs.
+	// bind+escalate let it assign managed roles without holding their rules, so
+	// gate on the feature to keep these verbs off non-RBAC-UI clusters.
+	if c.cfg.RBACManagementEnabled {
+		rules = append(rules, rbacv1.PolicyRule{
+			APIGroups: []string{"rbac.authorization.k8s.io"},
+			Resources: []string{"clusterroles", "roles", "clusterrolebindings", "rolebindings"},
+			Verbs:     []string{"get", "list", "watch", "create", "update", "patch", "delete", "bind", "escalate"},
```
pasanw: Similar to previous comment - can we avoid bind and escalate? This one seems a bit more serious since it is given to users

pkg/render/kubecontrollers/kube-controllers.go
```diff
@@ -658,6 +672,76 @@ func kubeControllersRoleEnterpriseCommonRules(cfg *KubeControllersConfiguration)
...
+		},
+		rbacv1.PolicyRule{
+			APIGroups:     []string{""},
+			Resources:     []string{"configmaps"},
+			ResourceNames: []string{"tigera-known-oidc-users"},
```
pasanw: Does this controller access this config map? I thought only the ui-apis endpoint would consult this config map when listing users

pkg/render/manager.go
```diff
@@ -1195,6 +1207,72 @@ func managerClusterRole(managedCluster bool, kubernetesProvider operatorv1.Provi
 	return cr
 }
 
+// rbacManagementUIRules returns the cluster-scoped rules the RBAC management
+// UI adds to calico-manager-role. Named-resource access is scoped separately
+// on rbacManagementUINamespacedRole.
+func rbacManagementUIRules() []rbacv1.PolicyRule {
+	rules := RBACManagementEscalationRules()
```
pasanw: Does the manager use its own SA when creating RBAC via the UI? Or is it impersonating the user?

pkg/render/manager.go
```diff
@@ -1265,6 +1343,19 @@ func (c *managerComponent) managerCalicoSystemNetworkPolicy() *v3.NetworkPolicy
 		})
 	}
 
+	if c.cfg.Manager.RBACManagementEnabled() && c.cfg.Authentication != nil && c.cfg.Authentication.Spec.OIDC != nil {
```
pasanw: Why do we gate on OIDC being enabled? I'm wondering what that has do to with the user deciding they want to do an LDAP sync for RBAC UI.

To me I would have expected that this block conditionally renders depending on whether the LDAP secret is placed by the user, and that we could parse the LDAP URL to scope the network policy (we do a similar thing for guardian policy)

api/v1/manager_types.go
```diff
@@ -26,6 +26,37 @@ type ManagerSpec struct {
 	// ManagerDeployment configures the Manager Deployment.
 	// +optional
 	ManagerDeployment *ManagerDeployment `json:"managerDeployment,omitempty"`
+
+	// RBAC configures the RBAC management UI feature.
+	// +optional
+	RBAC *RBAC `json:"rbac,omitempty"`
```
pasanw: I think a field named RBAC is bound to be first perceived as a field to configure K8S RBAC for Manager. Perhaps RBACUI?

pkg/controller/installation/core_controller.go
```diff
@@ -1071,6 +1080,16 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 			return reconcile.Result{}, err
 		}
 
+		// On Calico/OSS the Manager CRD is absent, so the read returns NoMatchError.
+		managerCR, err = utils.GetManager(ctx, r.client, false, "")
+		if err != nil && !apierrors.IsNotFound(err) && !meta.IsNoMatchError(err) {
+			r.status.SetDegraded(operatorv1.ResourceReadError, "Error reading Manager", err, reqLogger)
+			return reconcile.Result{}, err
+		}
+		if apierrors.IsNotFound(err) || meta.IsNoMatchError(err) {
```
pasanw: I think if we hit NoMatchError here it should arguably be treated as an error condition - even though we are in enterprise the expected CRDs have not been loaded yet. I don't think we should view that as equivalent to the user not specifying a Manager - we don't know their intent yet

pkg/controller/installation/core_controller.go
```diff
@@ -1071,6 +1080,16 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 			return reconcile.Result{}, err
 		}
 
+		// On Calico/OSS the Manager CRD is absent, so the read returns NoMatchError.
+		managerCR, err = utils.GetManager(ctx, r.client, false, "")
```
pasanw, line 1083: nit: should we update GetManager to return nil, nil on NotFound?

pkg/render/kubecontrollers/kube-controllers.go
```diff
@@ -658,6 +672,65 @@ func kubeControllersRoleEnterpriseCommonRules(cfg *KubeControllersConfiguration)
...
+	return append(rules,
+		rbacv1.PolicyRule{
+			APIGroups: []string{"operator.tigera.io"},
+			Resources: []string{"compliances"},
+			Verbs:     []string{"get", "list"},
+		},
+		rbacv1.PolicyRule{
+			APIGroups: []string{""},
+			Resources: []string{"secrets"},
+			Verbs:     []string{"get", "create", "patch"},
```
pasanw: Can this be scoped down to the specific resource names that need it?

pkg/render/kubecontrollers/kube-controllers.go
```diff
@@ -658,6 +672,65 @@ func kubeControllersRoleEnterpriseCommonRules(cfg *KubeControllersConfiguration)
...
+			},
+		},
+	}
+}
+
+// rbacSyncControllerRules returns the extra rules the rbacsync controller
+// needs on top of render.RBACManagementEscalationRules.
+func rbacSyncControllerRules() []rbacv1.PolicyRule {
+	rules := render.RBACManagementEscalationRules()
+	return append(rules,
```
pasanw: Would be nice to have comments explaining the need for these rules

pkg/render/manager.go
```diff
@@ -1161,6 +1176,12 @@ func managerClusterRole(managedCluster bool, kubernetesProvider operatorv1.Provi
 		},
 	}
 
+	// Zero-tenant-only. Must match the rbacManagementUINamespacedRole gate so
```
pasanw: Nit: PR seems to be using the terminology "zero-tenant" - should be single tenant?

pkg/render/manager.go
```diff
@@ -1195,6 +1216,72 @@ func managerClusterRole(managedCluster bool, kubernetesProvider operatorv1.Provi
...
+			Rules: []rbacv1.PolicyRule{
+				{
+					APIGroups: []string{""},
+					Resources: []string{"configmaps", "secrets"},
+					Verbs:     []string{"create"},
+				},
+				{
+					APIGroups: []string{""},
+					Resources: []string{"secrets"},
+					Verbs:     []string{"patch"},
```
pasanw: I believe patch can have a resource name specified

pkg/render/manager.go
```diff
@@ -1195,6 +1216,72 @@ func managerClusterRole(managedCluster bool, kubernetesProvider operatorv1.Provi
 	return cr
 }
 
+// rbacManagementUIRules returns the cluster-scoped rules the RBAC management
+// UI adds to calico-manager-role. Named-resource access is scoped separately
+// on rbacManagementUINamespacedRole.
+func rbacManagementUIRules() []rbacv1.PolicyRule {
```
pasanw, line 1220: Just to elaborate on what I meant in my previous review: don't these rules not really matter? They are to ensure escalation checks are passed when managing RBAC, but don't we use the user's auth via impersonation? Therefore rules added to the manager SA are irrelevant?

pkg/render/manager_test.go
```diff
@@ -1730,6 +1731,154 @@ var _ = Describe("Tigera Secure Manager rendering tests", func() {
...
+			It("omits LDAP egress when the LDAP config secret is absent, even with RBAC UI enabled", func() {
+				resources, _ := renderObjects(renderConfig{
+					installation: installation, ns: render.ManagerNamespace,
+					manager: rbacUIManager, rbacManagementLDAP: false,
+				})
+				policy := testutils.GetCalicoSystemPolicyFromResources(policyName, resources)
+				Expect(policy.Spec.Egress).NotTo(ContainElement(ldapEgress))
+			})
+
+			It("omits LDAP egress when OIDC is configured but the LDAP config secret is absent", func() {
```
pasanw: Can be deleted? Not really a regression case

pkg/render/manager.go
```diff
@@ -746,6 +759,7 @@ func (c *managerComponent) managerUIAPIsContainer() corev1.Container {
 		{Name: "LINSEED_CLIENT_KEY", Value: keyPath},
 		{Name: "ELASTIC_KIBANA_DISABLED", Value: strconv.FormatBool(c.cfg.Tenant.MultiTenant())},
 		{Name: "VOLTRON_URL", Value: ManagerService(c.cfg.Tenant)},
+		{Name: "RBAC_UI_ENABLED", Value: strconv.FormatBool(c.cfg.Manager.RBACManagementEnabled())},
```
pasanw: Should this be gated on single tenancy as well?

pkg/render/kubecontrollers/kube-controllers.go
dimitri-nicolo: `Get`, `Patch` can be name scoped. `Create` cannot.

pkg/render/manager.go
dimitri-nicolo: Keeping "non multi-tenant". Enterprise is zero-tenant (`Tenant == nil`),  single-tenant is a separate mode (Tenant CR, no namespace). The gate is `!MultiTenant()`, so "non multi-tenant" matches it exactly without naming the wrong mode.

pkg/render/manager.go
dimitri-nicolo: Gated on `!MultiTenant()`. Here `Tenant` is only ever nil (zero-tenant) or namespaced (multi-tenant), so this enables it for enterprise and excludes multi-tenant.

api/v1/manager_types.go
```diff
@@ -27,36 +27,36 @@ type ManagerSpec struct {
...
 	// +optional
 	// +kubebuilder:validation:Enum=Enabled;Disabled
-	UI RBACUI `json:"ui,omitempty"`
+	Enabled RBACUIStatus `json:"enabled,omitempty"`
 }
 
-// RBACUI toggles the RBAC management UI feature.
-type RBACUI string
+// RBACUIStatus toggles the RBAC management UI feature.
+type RBACUIStatus string
```
pasanw: nit: conventions are typically to call this a Type, Option, or Mode rather than a Status

api/v1/manager_types.go
```diff
@@ -26,6 +26,37 @@ type ManagerSpec struct {
...
+	RBACUI *RBACUI `json:"rbacUI,omitempty"`
+}
+
+// RBACUI controls the RBAC management UI feature surface.
+type RBACUI struct {
+	// Enabled controls whether the RBAC management UI is enabled. Defaults to
+	// Disabled.
+	// +optional
+	// +kubebuilder:validation:Enum=Enabled;Disabled
+	Enabled RBACUIStatus `json:"enabled,omitempty"`
```
pasanw:
```suggestion
	Type *RBACUIType `json:"rbacUI,omitempty"`
```

pkg/controller/apiserver/apiserver_controller.go
```diff
@@ -378,14 +378,15 @@ func (r *ReconcileAPIServer) Reconcile(ctx context.Context, request reconcile.Re
 			return reconcile.Result{}, err
 		}
 
+		// A nil managerCR (no Manager created) means callers fall back to their
```
pasanw: nit: this comment serves as storage of rationale/history that is useful for the author, rather than context that helps the reader - would just remove (personally)

pkg/controller/installation/core_controller.go
```diff
@@ -1080,15 +1080,16 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 			return reconcile.Result{}, err
 		}
 
-		// On Calico/OSS the Manager CRD is absent, so the read returns NoMatchError.
+		// We are in enterprise, so the Manager CRD is expected. A nil managerCR
+		// (no Manager created) means callers fall back to their defaults. A
+		// NoMatchError (Manager CRD not registered yet) is treated as an error:
```
pasanw: nit: this comment serves as storage of rationale/history that is useful for the author, rather than context that helps the reader - would just remove (personally)

pkg/render/kubecontrollers/kube-controllers.go
pasanw: What I meant was that it would be nice to have comments that store the context behind why we need these specific rules - my understanding is that they map to a specific permission needed from a specific managed role - can we reference the name of that managed role here?

pkg/render/rbac_management.go
```diff
@@ -16,10 +16,13 @@ package render
 
 import rbacv1 "k8s.io/api/rbac/v1"
 
-// RBACManagementEscalationRules returns the rules shared by the calico-manager
-// and calico-kube-controllers roles for the RBAC management UI. The SA writing
-// a managed role must already hold every permission it grants, else K8s rejects
-// the write as privilege escalation; keep this aligned with the managed roles.
+// RBACManagementEscalationRules returns the permission set the rbacsync
```
pasanw: Sorry for catching this just now and not as part of my previous feedback, but if we have realized now that escalation rules only apply to the rbacsync controller and not ui-apis, I guess this no longer needs to be shared code? Everything should just be defined within rbacSyncControllerRules and everything there is an escalation rule? (and as per my other comment, each rule can have a comment explaining which managed role it covers)

pkg/render/manager.go
```diff
@@ -759,7 +759,11 @@ func (c *managerComponent) managerUIAPIsContainer() corev1.Container {
 		{Name: "LINSEED_CLIENT_KEY", Value: keyPath},
 		{Name: "ELASTIC_KIBANA_DISABLED", Value: strconv.FormatBool(c.cfg.Tenant.MultiTenant())},
 		{Name: "VOLTRON_URL", Value: ManagerService(c.cfg.Tenant)},
-		{Name: "RBAC_UI_ENABLED", Value: strconv.FormatBool(c.cfg.Manager.RBACManagementEnabled())},
+		// The RBAC management UI is not supported on multi-tenant management
```
pasanw: nit: I would remove this comment, feels like a comment written for the person steering Claude rather than the eventual reader of the code

pkg/render/manager.go
```diff
@@ -1265,6 +1357,20 @@ func (c *managerComponent) managerCalicoSystemNetworkPolicy() *v3.NetworkPolicy
 		})
 	}
 
+	if c.cfg.Manager.RBACManagementEnabled() && c.cfg.RBACManagementLDAPConfigured {
```
pasanw: Should this also gate on multi-tenancy?

api/v1/manager_types.go
dimitri-nicolo: What about the following options:
```
// RBACUI configures the RBAC management UI.
type RBACUI struct {
	// Type selects the RBAC management UI mode. Unmanaged leaves Calico
	// Enterprise RBAC to be configured manually; ManagedCalicoEnterpriseRBAC
	// enables the RBAC management UI for Calico Enterprise RBAC. Defaults to
	// Unmanaged.
	// +optional
	// +kubebuilder:validation:Enum=Unmanaged;ManagedCalicoEnterpriseRBAC
	Type *RBACUIType `json:"type,omitempty"`
}

// RBACUIType selects the RBAC management UI mode.
type RBACUIType string

const (
	// RBACUIUnmanaged leaves RBAC to be configured manually.
	RBACUIUnmanaged RBACUIType = "Unmanaged"
	// RBACUIManagedCalicoEnterpriseRBAC enables the RBAC management UI for
	// Calico Enterprise RBAC.
	RBACUIManagedCalicoEnterpriseRBAC RBACUIType = "ManagedCalicoEnterpriseRBAC"
)
```

pasanw (approved):
LGTM 🎉 As discussed, let's see what the other maintainers think about the API changes before we merge

api/v1/manager_types.go
```diff
@@ -26,6 +26,29 @@ type ManagerSpec struct {
...
+	RBACUI *RBACUI `json:"rbacUI,omitempty"`
+}
+
+// RBACUI configures the RBAC management UI. This is a separate control plane
+// for Calico Enterprise RBAC that lives alongside, and does not replace, the
+// user's ability to configure RBAC themselves.
+type RBACUI struct {
+	// Enabled turns the RBAC management UI on or off. Defaults to false.
+	// +optional
+	Enabled *bool `json:"enabled,omitempty"`
```
Brian-McM: We have an updated api spec in the docs that specifies these toggles,

```
Use a named string type with Enabled/Disabled constants and an Enum marker — not a *bool. This is the established repo idiom:

type NotificationMode string

const (
    Disabled NotificationMode = "Disabled"
    Enabled  NotificationMode = "Enabled"
)

```

pkg/controller/installation/core_controller.go
```diff
@@ -1705,7 +1720,8 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 		// the kube-controllers component (and deleted when the WAF extension is
 		// disabled); the caBundle is the operator CA that issued the serving
 		// cert above.
-		WAFWebhookCABundle: certificateManager.KeyPair().GetCertificatePEM(),
+		WAFWebhookCABundle:    certificateManager.KeyPair().GetCertificatePEM(),
+		RBACManagementEnabled: managerCR.RBACManagementEnabled(),
```
Brian-McM, line 1724: Since utils.GetManager returns nil, nil on not found won't this now panic?

Brian-McM (changes requested):

api/v1/manager_types.go
dimitri-nicolo: @Brian-McM Thanks, I had considered raising the boolean vs. enum question with the maintainers before merging anything, but I'm happy to follow the idiom. Done in the latest commit: replaced the *bool with a named `RBACUIStatusType enum` (Enabled/Disabled + Enum marker), exposed as a State field.

cc: @pasanw

pkg/controller/installation/core_controller.go
dimitri-nicolo: @Brian-McM It shouldn't panic, `RBACManagementEnabled()` guards `m == nil` before touching the pointer. The latest commit adds a table test (api/v1/manager_types_test.go) covering the nil-managerCR, nil-RBACUI, and nil-State cases.

rene-dekker (changes requested):
The host should be taken from Authentication.spec.ldap.host.

rene-dekker (approved):

rene-dekker:
I reviewed only the ldap changes, which look good now.

pkg/controller/apiserver/apiserver_controller.go
```diff
@@ -497,6 +511,7 @@ func (r *ReconcileAPIServer) Reconcile(ctx context.Context, request reconcile.Re
 		KubernetesVersion:            r.opts.KubernetesVersion,
 		ClusterDomain:                r.opts.ClusterDomain,
 		RequiresAggregationServer:    !r.opts.UseV3CRDs,
+		RBACManagementEnabled:        managerCR.RBACManagementEnabled(),
```
dimitri-nicolo, line 514:
```suggestion
		RBACManagementEnabled:        managerCR != nil && managerCR.RBACManagementEnabled(),
```

pkg/controller/installation/core_controller.go
dimitri-nicolo:
```suggestion
		RBACManagementEnabled: managerCR != nil && managerCR.RBACManagementEnabled(),
```

Brian-McM (approved):

PR #4879, opened 2026-06-02
EV-6666: Surface Alertmanager alerts on the manager Alerts page

rene-dekker:
Wires Prometheus/Alertmanager alerts through to the manager Alerts page.

- Run Alertmanager as a dedicated `calico-alertmanager` service account; issue it a Linseed bearer token (a service-account-token Secret, validated via TokenReview) plus RBAC to create Linseed events.
- Deliver the Alertmanager config as a raw `alertmanager.yaml` secret whose webhook forwards **all** alerts to Linseed's `/api/v1/events/alertmanager` over mTLS. Users suppress individual alerts via AlertExceptions, not the Monitor API.
- Managed clusters: reach Linseed through Guardian via a namespace-local `tigera-linseed` ExternalName service; the token is provisioned by Linseed's token controller.
- Add IP-pool-exhaustion alerts to the Calico PrometheusRule.
- Hash the Alertmanager config into the pod so config changes roll the pod and reload it.

Companion PRs: calico-private (Linseed ingest + dedup), ui-modules (`prometheus_alert` rendering).

```release-note
Prometheus/Alertmanager alerts (including IP pool exhaustion) are now surfaced on the Calico Enterprise manager Alerts page.
```


pkg/controller/monitor/monitor_controller.go
```diff
@@ -577,70 +584,29 @@ func PrometheusTLSServerDNSNames(clusterDomain string) []string {
...
-	// If it is modified by the user, Monitor controller will not set the owner reference.
-	secret, err = utils.GetSecret(ctx, r.client, monitor.AlertmanagerConfigSecret, common.TigeraPrometheusNamespace)
+// readAlertmanagerConfig returns the AlertmanagerConfig to render in the tigera-prometheus
+// namespace, where it is referenced by Alertmanager.spec.alertmanagerConfiguration.
+//
+// If the user supplies their own AlertmanagerConfig in the tigera-operator namespace (named
+// monitor.AlertmanagerConfigName), the operator renders a copy of it in tigera-prometheus.
+// Otherwise it renders the operator's default config (the Linseed webhook receiver when the UI
+// alerts integration is enabled, or a null receiver when disabled).
+func (r *ReconcileMonitor) readAlertmanagerConfig(ctx context.Context, uiAlertsEnabled bool) (*monitoringv1alpha1.AlertmanagerConfig, error) {
```
electricjesus: This drops existing customer Alertmanager config on upgrade, and I don't think we can ship it that way.

Today the customization path is the raw `alertmanager-calico-node-alertmanager` secret. It's documented, and the old `readAlertmanagerConfigSecret` carried all that owner-ref logic precisely to leave a user-modified secret alone. This PR deletes that secret and only reads config from an `AlertmanagerConfig` CR the customer has never created. So on upgrade, anyone who set their own receivers (PagerDuty, Slack, email) loses them and falls back to the default Linseed webhook. Their external paging stops and the alerts quietly reroute to the manager UI instead.

Options, in order of how much I'd trust them:
- Migrate: if the legacy secret exists and differs from the old default, parse it and seed the `AlertmanagerConfig` before deleting the secret.
- Failing that, detect a non-default legacy secret and `SetDegraded` with a clear message instead of silently replacing it, so the upgrade isn't invisible.

Either way the release note has to call this out as a breaking change. Right now it only describes the new feature.

pkg/render/monitor/monitor.go
```diff
@@ -259,23 +295,41 @@ func (mc *monitorComponent) Objects() ([]client.Object, []client.Object) {
...
-			}},
+			mc.alertmanagerLinseedClusterRole(),
+			mc.alertmanagerLinseedClusterRoleBinding(),
+			mc.cfg.AlertmanagerConfig,
 		)
 	}
 
+	// The Linseed bearer-token secret is only needed when Alertmanager is running and forwarding
+	// alerts to Linseed (the UI alerts integration is enabled); otherwise remove it.
+	if mc.alertmanagerReplicas() > 0 && mc.cfg.Monitor.UIAlertsEnabled() {
```
electricjesus: Two things about the disable toggle when a user brings their own AlertmanagerConfig.

The toggle only swaps the default. If a user has their own `AlertmanagerConfig` in the operator namespace, `uiAlertsIntegration: Disabled` does nothing, since we copy their spec verbatim. The field doc says it "controls whether alerts are forwarded to Linseed," which won't hold for those users. Worth documenting the precedence, or deciding whether disable should win regardless.

Separately, the Linseed token secret and the `tigera-alertmanager-linseed` ClusterRole/Binding get created whenever Alertmanager runs with the integration enabled, even if the user's own config never talks to Linseed. That leaves a token secret and an event-create grant nothing uses. Not harmful, but it's a dangling credential. Could gate those on the default-config path rather than on `UIAlertsEnabled` alone.

api/v1/monitor_types.go
```diff
@@ -36,6 +36,25 @@ type MonitorSpec struct {
...
 	Alertmanager *Alertmanager `json:"alertmanager,omitempty"`
+
+	// UIAlertsIntegration controls whether Prometheus/Alertmanager alerts are forwarded to Linseed and surfaced
+	// on the manager Alerts page. When omitted, the integration is Enabled.
+	// +optional
+	UIAlertsIntegration *UIAlertsIntegrationStatusType `json:"uiAlertsIntegration,omitempty"`
+}
+
+// +kubebuilder:validation:Enum=Enabled;Disabled
+type UIAlertsIntegrationStatusType string
```
electricjesus: `UIAlertsIntegrationStatusType` reads like a status field, but this is a spec enum. It's public API and awkward to rename after release, so I'd fix it now. `UIAlertsIntegrationType` or `UIAlertsIntegrationMode` matches what it actually is.

electricjesus (commented):
Drive-by review, courtesy of a quest from Tigera Town 🤣. Mostly looks good. One thing I think blocks merge, plus a few mediums, left inline.

**Cross-PR ordering:** these two have to ship together, and this PR can't stand alone. The operator's own RBAC for `alertmanagerconfigs` lives in the calico-private charts in tigera/calico-private#12184, so if this vendors into the operator ahead of that, the monitor controller can't create the AlertmanagerConfig and goes degraded. Same on the receiving end: without #12184 the `/api/v1/events/alertmanager` endpoint 404s and Linseed rejects the `prometheus_alert` type. Worth pinning both to the same release and noting the dependency on each PR while they're still draft.

One nice-to-have I noticed but won't block on: the config-hash annotation that rolls the pod doesn't include the token secret data, so the pod won't roll when Kubernetes first populates the token. It relies on the config-reloader watching the mounted secret. Probably fine, worth a sanity check.

api/v1/monitor_types.go
```diff
@@ -186,7 +186,7 @@ type Alertmanager struct {
 }
 type AlertmanagerSpec struct {
 	// Replicas defines the number of Alertmanager replicas. When set to 0, Alertmanager is not rendered.
-	// Default: 0
+	// Default: 1
```
rene-dekker, line 189: note: the default of 0 was never released, so this is not a change.

caseydavenport:

```

```

PR #4894, opened 2026-06-04
fix(gatewayapi): split envoy-gateway NetworkPolicy ingress by IP family

electricjesus:
## Description

Bug fix. The `calico-system.envoy-gateway` NetworkPolicy ingress allow put both `0.0.0.0/0` and `::/0` in a single rule's `Source.Nets`, which Calico rejects ("rule contains both IPv4 and IPv6 CIDRs"), so the policy fails to apply on dual-stack and IPv6 clusters. Split the allow into one rule per address family.

Split out of #4821 — standalone fix, no WAF dependency.

Tested: `go test ./pkg/render/gatewayapi/...` — asserts the ingress is rendered as two per-family rules.

## Release Note

```release-note
Fixed the envoy-gateway NetworkPolicy failing to apply on dual-stack and IPv6 clusters: the ingress allow mixed IPv4 and IPv6 CIDRs in a single rule, which Calico rejects. The rule is now split per IP family.
```

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:
  - `release-note-required`
  - `docs-not-required`


electricjesus:
Superseded by #4887 (identical NP ingress split, merged). Closing as dupe — test coverage for the per-family split to follow separately.

PR #4895, opened 2026-06-04
feat(applicationlayer): WAF observability — gateway audit-log capture + Felix/fluentd event log (EV-6650)

electricjesus:
## Description

Operator render for the Gateway API WAF observability path (EV-6650), split out of #4821:

- **Audit-log capture (gateway proxy render)** — `pkg/render/gatewayapi/gateway_api.go`:
  - `EnvoyProxy.Spec.Logging.Level = {default: warn, wasm: info}` — surfaces the Coraza WASM filter's `AuditLog:` lines in Envoy's application log.
  - `EnvoyProxy.Spec.ExtraArgs += --log-path /access_logs/envoy.log` — redirects the application log onto the existing `access-logs` emptyDir (ExtraArgs rather than a container-args Patch, which would replace EG's generated args; the path sits directly under `/access_logs` because Envoy doesn't create `--log-path` parent dirs).
  - `WAF_AUDIT_LOG_PATH=/access_logs/envoy.log` on the l7-log-collector, which tails the file and forwards WAF decision records via `PolicySync.ReportWAF`.
- **Felix + fluentd event-log legs** — `WAFEventLogsFileEnabled` (FelixConfiguration) and `WAF_LOG_FILE` (fluentd) so WAF block / would-block decisions land in the `tigera_secure_ee_waf` index.

## Release Note

```release-note
Add operator render for Gateway API WAF observability (EV-6650): capture the Coraza audit log from the gateway proxy, and enable the Felix + fluentd legs (WAFEventLogsFileEnabled, WAF_LOG_FILE) so WAF block / would-block decisions land in the tigera_secure_ee_waf index.
```

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:
  - `release-note-required`
  - `docs-pr-required`

## Linked

- Design: `tigera/designs#25` (PMREQ-384)
- Producer pair: `tigera/calico-private#12142` (EV-6650) — must merge first
- Stacked on: #4821

### DIFF
--- a/pkg/controller/applicationlayer/applicationlayer_controller.go
+++ b/pkg/controller/applicationlayer/applicationlayer_controller.go
@@ -181,7 +181,10 @@ func (r *ReconcileApplicationLayer) Reconcile(ctx context.Context, request recon
 			// Owned objects are automatically garbage collected. For additional cleanup logic use finalizers.
 			reqLogger.Info("ApplicationLayer object not found")
 			// Patch tproxyMode if it's  needed after crd deletion.
-			if err = r.patchFelixConfiguration(ctx, nil, r.isGatewayWAFEnabled(ctx)); err != nil {
+			gatewayWAFEnabled, gwErr := r.isGatewayWAFEnabled(ctx)
+			if gwErr != nil {
+				reqLogger.Error(gwErr, "Error checking GatewayAPI WAF state; skipping felix configuration patch")
+			} else if err = r.patchFelixConfiguration(ctx, nil, gatewayWAFEnabled); err != nil {
 				reqLogger.Error(err, "Error patching felix configuration")
 			}
 			r.status.OnCRNotFound()
@@ -245,7 +248,12 @@ func (r *ReconcileApplicationLayer) Reconcile(ctx context.Context, request recon
 	}
 
 	// Patch felix configuration if necessary.
-	if err = r.patchFelixConfiguration(ctx, instance, r.isGatewayWAFEnabled(ctx)); err != nil {
+	gatewayWAFEnabled, err := r.isGatewayWAFEnabled(ctx)
+	if err != nil {
+		r.status.SetDegraded(operatorv1.ResourceReadError, "Error checking GatewayAPI WAF state", err, reqLogger)
+		return reconcile.Result{}, err
+	}
+	if err = r.patchFelixConfiguration(ctx, instance, gatewayWAFEnabled); err != nil {
 		r.status.SetDegraded(operatorv1.ResourcePatchError, "Error patching felix configuration", err, reqLogger)
 		return reconcile.Result{}, err
 	}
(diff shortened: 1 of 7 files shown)


pkg/render/gatewayapi/gateway_api.go
```diff
@@ -927,6 +967,12 @@ func (pr *gatewayAPIImplementationComponent) envoyProxyConfig(className, ns stri
 						Name:  "ENVOY_ACCESS_LOG_PATH",
 						Value: "/access_logs/access.log",
 					},
+					// WAF audit capture: file the collector tails for the wasm filter's
+					// Coraza "AuditLog:" lines (Envoy's app log, redirected via --log-path).
+					{
+						Name:  "WAF_AUDIT_LOG_PATH",
+						Value: wafAuditLogPath,
```
jaderhs, line 1006: Is it expected behaviour for `WAF_AUDIT_LOG_PATH` to be set to `wafAuditLogPath` even if envoy proxy has a user-supplied `--log-path` ?

pkg/controller/applicationlayer/applicationlayer_controller.go
```diff
@@ -563,3 +581,21 @@ func (r *ReconcileApplicationLayer) patchFelixConfiguration(ctx context.Context,
...
+		(al != nil && ((al.Spec.SidecarInjection != nil && *al.Spec.SidecarInjection == operatorv1.SidecarEnabled) ||
+			(al.Spec.WebApplicationFirewall != nil && *al.Spec.WebApplicationFirewall == operatorv1.WAFEnabled)))
+}
+
+// isGatewayWAFEnabled reports whether the GatewayAPI WAF data-plane extension is enabled. A missing or
+// unreadable GatewayAPI CR is treated as disabled.
+func (r *ReconcileApplicationLayer) isGatewayWAFEnabled(ctx context.Context) bool {
+	gw, _, err := gatewayapi.GetGatewayAPI(ctx, r.client)
+	if err != nil {
+		return false
```
jaderhs: I don't think `err` should get masked here and this function should return `(bool, error)` instead

pkg/render/gatewayapi/gateway_api.go
electricjesus: No, not intended. Good catch.

We don't expose a `--log-path` knob. It can only come from a custom base `EnvoyProxy`. And it can't work anyway: the collector only reads the shared `access-logs` volume.

So the operator now owns the flag. `ensureExtraArg` sets it to our path and overwrites any other value. So it always matches `WAF_AUDIT_LOG_PATH`.

Fixed in `d2c09877e`, with tests.

pkg/controller/applicationlayer/applicationlayer_controller.go
electricjesus: Agreed. `isGatewayWAFEnabled` now returns `(bool, error)`.

A missing `GatewayAPI` CR still reads as off, with no error. Any other read error is passed up. So the loop retries instead of flipping `WAFEventLogsFileEnabled` off by mistake.

Main path retries. The delete path logs it and skips the patch.

Fixed in `d2c09877e`.

pkg/render/gatewayapi/gateway_api.go
```diff
@@ -809,6 +825,24 @@ func (pr *gatewayAPIImplementationComponent) controllerObjects() []client.Object
 	return objs
 }
 
+// ensureExtraArg sets "flag value" in an Envoy Gateway ExtraArgs slice (func-e parses each token as
+// a separate element), replacing the value if flag is already present or appending the flag/value
+// pair if not. It copies the slice so it never mutates a slice backing a cached EnvoyProxy object.
+func ensureExtraArg(args []string, flag, value string) []string {
```
jaderhs, line 833: Nitpick: If the envoy binary supports `--` to separate non-option args then this function should not replace args after `--`

jaderhs (approved):

pkg/render/gatewayapi/gateway_api.go
electricjesus: Good point. `ensureExtraArg` now stops at a bare `--`. It only matches the flag among the options before it, and a newly inserted pair goes before the `--`, so anything after stays put. Specs added for both cases. Done in `74261471b`.

PR #4904, opened 2026-06-08
gatewayapi: mirror trust bundle into Gateway namespaces on OSS too

radixo:
- the data-plane proxy mounts tigera-ca-bundle from its own namespace on both variants, but the per-namespace copy was Enterprise-gated, so OSS proxies hung at ContainerCreating (configmap not found)
- mirror the bundle for both variants; controller tracks OSS namespaces via a labeled ConfigMap for cleanup

## Release Note

```release-note
None
```

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/gatewayapi/gatewayapi_controller.go
+++ b/pkg/controller/gatewayapi/gatewayapi_controller.go
@@ -17,11 +17,11 @@ package gatewayapi
 import (
 	"context"
 	"fmt"
+	"time"
 
 	v3 "github.com/tigera/api/pkg/apis/projectcalico/v3"
 	v1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
-	rbacv1 "k8s.io/api/rbac/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -52,7 +52,9 @@ import (
 	"github.com/tigera/operator/pkg/ctrlruntime"
 	"github.com/tigera/operator/pkg/render"
 	"github.com/tigera/operator/pkg/render/common/networkpolicy"
+	"github.com/tigera/operator/pkg/render/common/secret"
 	"github.com/tigera/operator/pkg/render/gatewayapi"
+	"github.com/tigera/operator/pkg/tls/certificatemanagement"
 )
 
 const (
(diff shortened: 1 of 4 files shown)


pkg/controller/gatewayapi/gatewayapi_controller.go
```diff
@@ -505,10 +505,10 @@ func (r *ReconcileGatewayAPI) Reconcile(ctx context.Context, request reconcile.R
 	}
 	gatewayConfig.GatewayNamespaces = nsSet.SortedList()
 
-	// Enterprise: read previously-provisioned namespaces from the shared CRB's
-	// Subjects so we can clean them up when their Gateway is gone.
+	// Previously-provisioned namespaces, for cleanup. Enterprise reads the shared CRB's
+	// Subjects; OSS has no CRB, so it enumerates the mirrored trust-bundle ConfigMaps.
+	gatewayConfig.CurrentGatewayNamespaces = set.New[string]()
```
rene-dekker: I think the correct approach is for the trusted bundle to be named after the gateway and have an ownerref set to the gateway (see `CreateNamedTrustedBundle(...)`). This ensures that deletion of the gw will result in a cleaned bundle.

We wouldn't need all this logic to ensure cleanup. In fact, under the current conditions, deleting the GatewayAPI CR will delete the bundles, but not the gateways, leading to broken deployments.

radixo: Good call, thanks — done. The per-namespace resources (trust bundle, and on EE the WAF SA/RoleBindings/pull-secret) are now written by the controller with an ownerRef per Gateway in the namespace, so the GC cleans them up when the last Gateway there is deleted, and they no longer ride the GatewayAPI CR's lifecycle.

I kept a single shared bundle per namespace with one ownerRef per Gateway (rather than naming it per-gateway) since the proxy mounts a fixed name and multiple Gateways can share a namespace — GC still removes it once all owners are gone. This let me drop the CurrentGatewayNamespaces cleanup tracking entirely. Verified on a cluster: add/remove a 2nd Gateway updates the owner set, and deleting the last one GCs everything.

pkg/controller/gatewayapi/gatewayapi_controller.go
```diff
@@ -505,21 +507,51 @@ func (r *ReconcileGatewayAPI) Reconcile(ctx context.Context, request reconcile.R
...
-				if s.Kind == "ServiceAccount" {
-					gatewayConfig.CurrentGatewayNamespaces.Insert(s.Namespace)
+	// Legacy tigera-gateway teardown (TODO: remove once upgrades from 3.x are unsupported).
+	// Foreground-delete the old controller so its Deployment lingers until its pods are gone — only
+	// then does the Get below return NotFound, guaranteeing nothing re-creates the proxies we then
+	// sweep. Owner-scoped, requeueing until clean. Skipped if a Gateway lives in tigera-gateway,
+	// since then the proxy there belongs to the new controller.
+	const legacyGatewayNamespace = "tigera-gateway"
+	legacyNamespaceHostsGateway := false
+	for _, ns := range gatewayConfig.GatewayNamespaces {
```
rene-dekker: legacyNamespaceHostsGateway := slices.Contains(gatewayConfig.GatewayNamespaces, legacyGatewayNamespace)

radixo: Good catch, fixed this loops and the one on tierDown, thanks!

rene-dekker (approved):

PR #4923, opened 2026-06-15
Add egress rules to OSS guardian and goldmane network policies

### LINKED ISSUE #4804: guardian and goldmane network policies missing egress DNS rules
## Description

On an AKS cluster running the OSS Calico variant with tiered network policies, the `guardian` and `goldmane` pods in `calico-system` are unable to resolve DNS. Their egress UDP/53 traffic to CoreDNS in `kube-system` is being denied.

## Observed behavior

Whisker flow logs show denied flows:

Image

The blocking policy is implicit `calico-system.default-deny`.

## Analysis

Looking at the operator render code:

- `pkg/render/goldmane/component.go` — the `networkPolicy()` method only creates a policy with `Types: Ingress`. There are no egress rules at all for goldmane.

- `pkg/render/guardian.go` — the `ossNetworkPolicy()` function also only creates a policy with `Types: Ingress`. It allows inbound TCP from goldmane and inbound UDP/53, but does not define any egress.

Meanwhile, the enterprise path in `guardianCalicoSystemPolicy()` does call `networkpolicy.AppendDNSEgressRules(egressRules, cfg.OpenShift)`, so this issue seems specific to the OSS code path.

The `calico-system.default-deny` policy applies to all pods where `k8s-app != 'calico-apiserver'` and covers both Ingress and Egress types with no rules, so any pod without an explicit egress allow gets blocked.

For comparison, `calico-system.kube-controller-access` and `calico-system.apiserver-access` both include explicit egress rules allowing UDP/53 to CoreDNS, and those components work fine.

## Expected behavior

Both `guardian` and `goldmane` should be able to reach CoreDNS for DNS resolution, similar to how `calico-kube-controllers` and `calico-apiserver` have egress DNS rules in their respective policies.

Not adding those missing network policies just pollutes whisker denied flow logs.

## Environment

- Calico OSS variant (non-enterprise)
- AKS (Azure Kubernetes Service)
- Tigera operator helm chart v3.32.0

tmjd:
## Description

**Type:** bug fix.

In the OSS (Calico) variant, the `guardian` and `goldmane` network policies in the `calico-system` tier only declared **ingress** rules. All egress therefore fell through to the implicit `calico-system.default-deny` and was dropped — blocking DNS resolution and inter-component traffic for clusters connected to a management cluster, and polluting Whisker denied-flow logs. The enterprise Guardian path already builds egress via `AppendDNSEgressRules` + kube-API + tunnel rules; the OSS path short-circuited before any of it.

This PR adds the egress (and Whisker ingress) these components need to function under `default-deny`:

- **goldmane** — allow egress to DNS, the Kubernetes API server (its `Role` grants it ConfigMap access in `calico-system`), and guardian (flow-emitter POST). All destinations are known, so they are allowed explicitly with no trailing `Pass`.
- **guardian (OSS)** — allow egress to DNS, the Kubernetes API server, and goldmane (`:7443`), then a trailing `Pass`. Egress to the management-cluster tunnel endpoint is intentionally **not** rendered here: a domain-based management address would require the enterprise-only `EgressAccessControl` license, so the OSS policy cannot express it. The `Pass` defers that (and any other) egress to subsequent tiers for an administrator to allow. The enterprise path is unchanged.
- **whisker** — pass ingress to subsequent tiers. Whisker's ingress source (an ingress controller / load balancer) is not knowable by the operator, so ingress is passed rather than hard-denied by `default-deny`.

Egress rules follow the existing convention in `compliance.go`/`intrusion_detection.go` (pod-selector + backend `targetPort`). The Guardian policy continues to render only when the `calico-system` tier exists — precisely when the `default-deny` it must coexist with is present — so there is no window where Guardian is blocked but the allow policy is absent.

**Components affected:** `pkg/render/guardian.go`, `pkg/render/goldmane`, `pkg/render/whisker`.

**Testing:** Unit tests added/updated for the new guardian, goldmane, and whisker policy rules; the existing enterprise guardian golden-file test confirms enterprise rendering is unchanged. All `./pkg/render/...` tests pass.

**Issues:** Fixes #4804

## Release Note

```release-note
Add egress rules (DNS, Kubernetes API, and inter-component traffic) to the OSS guardian and goldmane network policies so they are not blocked by the calico-system default-deny policy, and allow ingress to whisker to be permitted by lower-tier policies.
```

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/whisker/component.go
+++ b/pkg/render/whisker/component.go
@@ -278,6 +278,12 @@ func (c *Component) networkPolicy() *v3.NetworkPolicy {
 	}
 	egressRules = networkpolicy.AppendDNSEgressRules(egressRules, c.cfg.OpenShift)
 
+	// Whisker is a UI that is reached from outside calico-system (e.g. via an ingress controller or
+	// load balancer), so the source of its ingress traffic is not knowable by the operator. Pass
+	// ingress to subsequent tiers so an administrator can allow it rather than having it hard-denied
+	// by calico-system.default-deny. See https://github.com/tigera/operator/issues/4804.
+	ingressRules := []v3.Rule{{Action: v3.Pass}}
+
 	return &v3.NetworkPolicy{
 		TypeMeta:   metav1.TypeMeta{Kind: "NetworkPolicy", APIVersion: "projectcalico.org/v3"},
 		ObjectMeta: metav1.ObjectMeta{Name: WhiskerPolicyName, Namespace: WhiskerNamespace},
@@ -285,6 +291,7 @@ func (c *Component) networkPolicy() *v3.NetworkPolicy {
 			Tier:     networkpolicy.CalicoTierName,
 			Types:    []v3.PolicyType{v3.PolicyTypeIngress, v3.PolicyTypeEgress},
 			Selector: networkpolicy.KubernetesAppSelector(WhiskerDeploymentName),
+			Ingress:  ingressRules,
 			Egress:   egressRules,
 		},
 	}
--- a/pkg/render/goldmane/component.go
+++ b/pkg/render/goldmane/component.go
@@ -384,14 +384,33 @@ func (c *Component) networkPolicy() *v3.NetworkPolicy {
 			},
 		})
 	}
+	// Goldmane runs in calico-system, where the calico-system.default-deny policy blocks any
+	// egress that lacks an explicit allow. It needs DNS, the Kubernetes API server (its Role grants
+	// it ConfigMap access in calico-system) and, when connected to a management cluster, Guardian
+	// (to POST flows). See https://github.com/tigera/operator/issues/4804.
+	egressRules := networkpolicy.AppendDNSEgressRules([]v3.Rule{}, c.cfg.OpenShift)
+	egressRules = append(egressRules, []v3.Rule{
+		{
+			Action:      v3.Allow,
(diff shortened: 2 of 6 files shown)


pkg/render/whisker/component.go
```diff
@@ -278,13 +278,20 @@ func (c *Component) networkPolicy() *v3.NetworkPolicy {
 	}
 	egressRules = networkpolicy.AppendDNSEgressRules(egressRules, c.cfg.OpenShift)
 
+	// Whisker is a UI that is reached from outside calico-system (e.g. via an ingress controller or
+	// load balancer), so the source of its ingress traffic is not knowable by the operator. Pass
+	// ingress to subsequent tiers so an administrator can allow it rather than having it hard-denied
+	// by calico-system.default-deny. See https://github.com/tigera/operator/issues/4804.
```
caseydavenport, line 284: I believe we did intentionally hard-deny this here, asking the user to create their own policies to allow this. But I agree that it should be hard denied in a lower tier rather than in calico-system.

To be honest, I thought the design of calico-system was that it shouldn't hard deny traffic - it should allow the necessary flows but pass through everything else. If that's not working, then there might be a larger problem here.

tmjd:
Closing as the right fix is perhaps making the default-deny be a pass rather than deny.

PR #4946, opened 2026-06-18
Render Calico OSS from individual images and fix VAP tests for v3.32

caseydavenport:
## Description

Bug fix. `release-v1.43` ships Calico Enterprise v3.24 EP1, which builds the combined image, but pins Calico OSS to v3.32, which was cut before the combined image existed for OSS. The branch was cut from master after the operator had already moved every component onto the combined image, so the Calico OSS variant references an image tag that was never published. This shows up as persistent FV failures looking for the v3.32 combined image.

This makes image selection variant-aware. The Calico OSS variant now renders the standalone per-component images using their own entrypoints and health probes, and Calico Enterprise keeps the combined image. Components touched: calico-node (plus the CNI and flexvol installers), Typha, kube-controllers, apiserver, CSI (driver and node-driver-registrar), Goldmane, the Whisker backend, the webhooks controller, Guardian, Windows, and the certificate-management CSR init container.

Same root cause on the admission side: v3.32 ships no Calico ValidatingAdmissionPolicy (the branch-prep commit dropped it along with the v3.32 CRDs), but the VAP tests still expected one. They now expect none for the Calico variant and exercise VAP creation through Enterprise, which still ships it.

FIPS images are intentionally not restored - FIPS support was removed on this branch and no render path references it.

Tested with the operator unit tests.

## Release Note

```release-note
NONE
```

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


Makefile
```diff
@@ -388,8 +382,7 @@ calico-whisker.tar:
 	docker pull $(FV_IMAGE_REGISTRY)/$(WHISKER_IMAGE):$(VERSION_TAG)
 	docker save --output $@ $(WHISKER_IMAGE):$(VERSION_TAG)
 
-IMAGE_TARS := calico-calico.tar \
-	calico-node.tar \
+IMAGE_TARS := calico-node.tar \
```
pasanw, line 439: Since OSS is back to rendering the per-component images, we probably need to restore the per-component tars here (apiserver, cni, pod2daemon-flexvol, kube-controllers, typha, csi, node-driver-registrar, goldmane, whisker-backend) the way they were before the combined image, otherwise FV won't have them preloaded.

pkg/components/calico.go
```diff
@@ -116,18 +212,31 @@ var (
 		variant:   calicoVariant,
 	}
 
+	// CalicoImages omits ComponentCalico: Calico OSS ships the individual
```
pasanw: Might be safer to remove ComponentCalico and have CombinedCalicoImage panic if it's ever called on OSS.

Makefile
caseydavenport: Yep, done. Restored the per-component tars (apiserver, cni, pod2daemon-flexvol, kube-controllers, typha, csi, node-driver-registrar, goldmane, whisker-backend) so FV preloads them now that OSS is back on the individual images.

pkg/components/calico.go
caseydavenport: Done. Dropped `ComponentCalico` and made `CombinedCalicoImage` panic on a non-Enterprise variant. That flushed out a few Enterprise-only render tests (manager, compliance, IDS, policy rec, monitor, app layer, linseed) that weren't setting the Enterprise variant and were silently resolving the OSS combined image, so I set the variant explicitly in those.

pasanw (approved):

PR #4954, opened 2026-06-22
feat: add a 'no-dataplane' option which allows for valid products to be installed with 'linuxDataplane: None'

doucol:
## Description

A suggestion from Arvind for the hackathon: add support to the operator for installing our 2 cni/dataplane agnostic products (Calico Ingress Gateway and Istio Service mesh) without our CNI nor our dataplane enabled.

This results in cni.type = nil and linuxDataplane: None 

"None" is a new enum value for our linuxDataplane.

## Release Note

```release-note
A new value 'None' is supported for the linuxDataplane which allows our Ingress Gateway and Istio Service Mesh to be installed without our dataplane components. 
```

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


pkg/controller/installation/core_controller.go
```diff
@@ -1019,6 +1049,17 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 		}
 	}
 
+	// An install with the dataplane disabled (spec.calicoNetwork.linuxDataplane: None) runs a different set of
+	// controllers than a normal install (see internal/controller.AddToManager), and that set is
+	// fixed when the manager starts. If the dataplane mode has changed since the operator
+	// started, reboot so the correct controllers are (de)registered. This mirrors the enterprise
+	// switch below; controller-runtime cannot add or remove controllers from a running manager.
+	if instance.DeletionTimestamp.IsZero() && r.dataplaneDisabled != instance.Spec.DataplaneDisabled() {
```
electricjesus: This reboot might never fire on a `None`→`Iptables` switch. We boot with the dataplane off, so the IPPool controller isn't registered. Flip back to `Iptables` and `needsIPPools` goes true, but there are no pools and nothing running to create them, so I think we'd hit the "waiting for enabled IP pools" return above and bail before reaching this check. Looks like the operator would get stuck and couldn't restart itself out of it..

Could we move this block up to right after `dataplaneDisabled :=`, above the cni and pool gates? That direction isn't covered by the test either, so might be worth a case.

Also might be worth a `return` after `osExitOverride(0)`. `os.Exit` kills us in prod so it doesn't bite there, but `checkActive` returns right after its exit for the test no-op case and this one just falls through.

pkg/controller/installation/validation.go
```diff
@@ -40,107 +40,185 @@ import (
...
+		// (hostPorts, multiInterfaceMode, containerIPForwarding, linuxPodInterfaceType) are also
+		// validated against spec.cni further below, which is nil when the dataplane is disabled
+		// and would otherwise cause a nil-pointer panic.
+		if instance.Spec.DataplaneDisabled() {
+			cn := instance.Spec.CalicoNetwork
+
+			// There is no Calico dataplane, so spec.cni (which describes the CNI the
+			// operator integrates Felix with) must be omitted entirely.
+			if instance.Spec.CNI != nil {
+				return fmt.Errorf("spec.cni must not be set when spec.calicoNetwork.linuxDataplane is %s",
```
electricjesus, line 162: I think this might reject a config the user never actually wrote when `None` comes in via the overlay. The base install defaults `spec.cni` and we persist it, then the overlay sets `linuxDataplane: None`. `OverrideInstallationSpec` doesn't seem to clear CNI when only the base has it set (no `AOnlySet` case in merge.go), so the merged spec would carry both `cni` and `None` and land here degraded. main.go says the mode can come from the overlay, so I think this path is supposed to work..?

pkg/controller/installation/core_controller.go
```diff
@@ -1657,62 +1677,67 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
...
-		// The webhook Service + ValidatingWebhookConfiguration are rendered by
-		// the kube-controllers component (and deleted when the WAF extension is
-		// disabled); the caBundle is the operator CA that issued the serving
-		// cert above.
-		WAFWebhookCABundle: certificateManager.KeyPair().GetCertificatePEM(),
-	}
-	components = append(components, kubecontrollers.NewCalicoKubeControllers(&kubeControllersCfg))
+	// Build a configuration for rendering calico/kube-controllers, which is not rendered
+	// when the dataplane is disabled.
+	var kubeControllersCfg kubecontrollers.KubeControllersConfiguration
```
electricjesus, line 1690: Looks like kube-controllers gets gated out here, but the WAF webhook cert above (`wafWebhookTLS`) and the cert render don't. So I think a no-dataplane cluster that turns on WAF in a GatewayAPI CR would end up with an orphan `tigera-waf-webhook` secret and a webhook that never runs, with no status to say why.. Might be worth gating the cert the same way, or surfacing that WAF-on-gateway isn't supported without the dataplane.

pkg/controller/installation/validation.go
```diff
@@ -40,107 +40,185 @@ import (
...
+					operatorv1.LinuxDataplaneNone)
+			}
+			if cn.BGP != nil && *cn.BGP == operatorv1.BGPEnabled {
+				return fmt.Errorf("spec.calicoNetwork.bgp cannot be Enabled when spec.calicoNetwork.linuxDataplane is %s",
+					operatorv1.LinuxDataplaneNone)
+			}
+
+			// Reject any remaining dataplane-only field that is explicitly set, reporting them
+			// all at once so the user can fix the spec in a single pass.
+			var disallowed []string
```
electricjesus, line 176: nit, not blocking: this is a deny-list, so I think a new `calicoNetwork` field added later might slip through in `None` mode by default. Some of these (`hostPorts` and friends) look like the ones that'd nil-panic on `spec.cni` downstream, which is why the block exists. An allow-list would fail closed. Worth a comment at least so the next person adding a field knows to touch this.

pkg/controller/installation/core_controller.go
```diff
@@ -608,8 +622,12 @@ func fillDefaults(instance *operatorv1.Installation, currentPools *v3.IPPoolList
 	}
 
 	// BPF dataplane requires IP autodetection even if we're not using Calico IPAM.
+	// With the Linux dataplane disabled there is no calico-node to perform autodetection,
+	// so skip defaulting these fields (validation rejects them when the dataplane is disabled).
 	needIPv4Autodetection := *instance.Spec.CalicoNetwork.LinuxDataplane == operatorv1.LinuxDataplaneBPF
```
electricjesus, line 627: tiny: you deref `LinuxDataplane` here, then zero it out right below for the disabled case. Could maybe fold it: `needIPv4Autodetection := !instance.Spec.DataplaneDisabled() && *...LinuxDataplane == LinuxDataplaneBPF`.

electricjesus (commented):
looks pretty solid to me! just a few comments..

caseydavenport:

```

```

PR #4970, opened 2026-06-26
feat(gatewayapi): calico-system policy for namespaced data-plane proxies

electricjesus:
## Description

Bug fix. Since #4690 the data-plane envoy proxies run in each Gateway's own namespace, not `calico-system`. The one `calico-system`-tier gateway policy selects the controller and certgen pods, so the proxies match nothing there and a Gateway under a default-deny tier gets no traffic.

This renders `calico-system.envoy-gateway-proxy`, one `NetworkPolicy` per Gateway namespace, selecting proxy pods on `k8s-app == 'calico-gateway-api-proxy'`. Ingress allows that namespace's own listener ports plus 19001 for metrics, then `Pass`. Egress allows DNS and 18000/18002 to the controller, then `Pass`, so the user still owns the backend hop and the proxy answers 503 until they allow it.

Two things the diff does not show. The listener ports come off the `GatewayList` the controller already reads to build `GatewayNamespaces`, and `watchGateways` is registered, so editing a listener re-renders. And the policy is owned by the GatewayAPI CR rather than the namespace's Gateways, because `mergeState` skips a `*v3.NetworkPolicy` whose Spec has not changed, so a second owner reference never merges in; `sweepOrphanedProxyPolicies` deletes leftovers instead, through an uncached reader so it starts no informer over every Calico policy on the cluster.

## Reading guide

- `c0b1d14e2` the policy
- `a7b817e71` select on a Calico-owned label
- `1d372cdee` all three components labelled in one place
- `3b16b346e` one policy per namespace, scoped to its listener ports
- `a4e01ceb5` sweep a namespace that loses its Gateways

## Test plan

`go build ./...`, `gofmt -l` clean, `GOOS=linux go vet ./pkg/...` clean, `make static-checks` 0 issues, `go test` green on `pkg/render/gatewayapi`, `pkg/controller/gatewayapi`, `pkg/enterprise/gatewayapi` and `pkg/controller/utils`.

Unit tests cover the port scoping including a Gateway with no listeners, the `Pass` after the allow, that no `GlobalNetworkPolicy` is rendered, and the sweep deleting an orphan while leaving a live policy and a user's own policy alone.

Cluster run on OSS master (eBPF, namespaced): baseline 200; under a default-deny a stock operator times out; with this operator 503, so the proxy is reachable and routing and only the backend hop is denied; 200 once the user allows the backend. That run predates the per-namespace port-scoped ingress rule, so the current rule is not cluster-tested yet. A newly added listener port only passes traffic after the next reconcile, and the 503 lands after Envoy's ~10s upstream timeout, so a curl timeout under 10s reads as a connect failure.

## Related

Docs: https://github.com/tigera/docs/pull/2802. The Job pod-template-labels change that used to ride here now sits on `seth/job-pod-template-labels`, because a non-nil Job pod template also gets `operator.tigera.io/host-networked`, which `podiprecovery` acts on by deleting pods.

## Release Note

```release-note
Add a `calico-system`-tier NetworkPolicy in each Gateway namespace so Envoy Gateway data-plane proxies work under a default-deny tier, allowing the ports the namespace's Gateways listen on (deploy.type=GatewayNamespace).
```

## For PR author

- [x] Tests for change.
- [ ] If changing pkg/apis/, run `make gen-files` (n/a, no API change).
- [ ] If changing versions, run `make gen-versions` (n/a).


pkg/render/gatewayapi/gateway_api.go
```diff
@@ -75,6 +75,12 @@ const (
 
 	ControllerPolicyName       = networkpolicy.CalicoComponentPolicyPrefix + "envoy-gateway"
 	EnvoyGatewayPolicySelector = "k8s-app == '" + GatewayControllerLabel + "' || k8s-app == '" + GatewayCertgenLabel + "'"
+
+	// Data-plane proxies run in each Gateway's own namespace (deploy.type=GatewayNamespace),
+	// not calico-system, so they need their own policy in the calico-system tier. EnvoyProxy
+	// stamps gateway.envoyproxy.io/owning-gateway-name on every proxy pod, so use it as the selector.
+	ProxyPolicyName          = networkpolicy.CalicoComponentPolicyPrefix + "envoy-gateway-proxy"
+	EnvoyProxyPolicySelector = "has(gateway.envoyproxy.io/owning-gateway-name)"
```
rene-dekker: Could we set some calico labels on these objects and use those for policy?

Related: do the envoy related components have standard labels that we set in utils/component.go? If not,that is a gap.

electricjesus: did both, thanks.

The selector keys off a label we own now. Proxy pods get `k8s-app: calico-gateway-api-proxy`, stamped through the EnvoyProxy pod spec (`ensureGatewayProxyLabel`), and `calico-system.envoy-gateway-proxy` selects on that instead of `gateway.envoyproxy.io/owning-gateway-name`. Controller and certgen keep the labels they had, just set from one place now (`setGatewayComponentLabel`).

On your `component.go` question: you were right, there was a gap, and it's fixed here too.

`setStandardSelectorAndLabels` does set the `app.kubernetes.io/*` labels on any object with a CR, so the controller Deployment, the certgen Job and the EnvoyProxy all get those. But the pod-template pass below it only handles Deployment/DaemonSet/Prometheus/Alertmanager, so Jobs hit the default case and return early and their pods miss the lot. Not an envoy thing, it's all four Jobs we render. Added `*batchv1.Job` to that switch.

Two bits worth flagging on that one. It labels the pod template only, since a Job's selector belongs to the job controller and is immutable, so we can't do the `Spec.Selector` thing the Deployment case does. And existing Jobs don't get rewritten, because `mergeState` compares Jobs on container images and pod-template annotations rather than labels. No churn on upgrade, but it does mean a Job already on a cluster keeps its old template until something recreates it. Seemed like the better trade than recreating four Jobs on every upgrade, though shout if you'd rather it were immediate..

The proxy is the one case the labeller genuinely can't reach. It derives the label from the object name, so we'd get `envoy-gateway` rather than a stable `calico-gateway-api-*`, and the proxy pods are created by the EG controller at runtime so we never render them to label. The EnvoyProxy pod spec is the only hook we own. That's why the `k8s-app` labelling still sits in the gateway render, with a comment saying why.

One bit of luck: the pod-template pass only sets `k8s-app` when it's empty, so our explicit labels survive it.

pkg/render/gatewayapi/gateway_api.go
```diff
@@ -486,6 +496,9 @@ func (pr *gatewayAPIImplementationComponent) Objects() ([]client.Object, []clien
 	// Installation's default-deny.
 	if pr.cfg.IncludeV3NetworkPolicy {
 		objs = append(objs, gatewayAPIControllerPolicy(common.CalicoNamespace, openShift))
+		// Data-plane proxies live in each Gateway's own namespace; a GlobalNetworkPolicy
+		// covers them all and auto-extends to new Gateway namespaces without a re-render.
```
caseydavenport: Think this comment would make more sense on top of the function rather than here.

caseydavenport: In fact it is already there, so we can just delete this and refactor the comment on line 485

pkg/render/gatewayapi/gateway_api.go
```diff
@@ -856,6 +866,37 @@ func ensureExtraArg(args []string, flag, value string) []string {
...
+	if template.Labels == nil {
+		template.Labels = map[string]string{}
+	}
+	template.Labels["k8s-app"] = label
+}
+
+// ensureGatewayProxyLabel stamps the same Calico-owned k8s-app label on the data-plane
+// proxy pods, via the EnvoyProxy pod spec, so gatewayAPIProxyPolicy can select them.
+// Any user-supplied pod labels carried over from a custom EnvoyProxy are preserved.
+func ensureGatewayProxyLabel(pod *envoyapi.KubernetesPodSpec) {
```
caseydavenport, line 772: Alternatively, we could update the component handler so that it catches envoy proxy resources and labels them automatically too? WDYT?

pkg/render/gatewayapi/gateway_api.go
```diff
@@ -1397,3 +1440,59 @@ func gatewayAPIControllerPolicy(namespace string, openShift bool) *v3.NetworkPol
 		},
 	}
 }
+
+// gatewayAPIProxyPolicy lets the data-plane envoy proxies — which run in each
+// Gateway's own namespace (deploy.type=GatewayNamespace), not calico-system —
+// punch through any default-deny in those namespaces. It is a GlobalNetworkPolicy
+// rather than a per-namespace NetworkPolicy fanned out over GatewayNamespaces so
+// that it automatically covers new Gateway namespaces with no re-render.
```
caseydavenport: Is "automatically with no re-render" actually important / desirable? 

On its own I don't think this is a reason one way or another - but could there be reasons we want global vs. not? 

i.e., differentiation in flow logging? or RBAC reasons?

pkg/render/gatewayapi/gateway_api.go
```diff
@@ -1397,3 +1440,59 @@ func gatewayAPIControllerPolicy(namespace string, openShift bool) *v3.NetworkPol
...
+			Types:    []v3.PolicyType{v3.PolicyTypeIngress, v3.PolicyTypeEgress},
+			// Allow all inbound TCP from any source (this also covers the 19001 metrics
+			// scrape). Gateway listener ports are user-defined and dynamic, so a managed
+			// Gateway has to accept arbitrary ports to serve traffic out of the box,
+			// including under a default-deny tier. Verified on a cluster: the narrower
+			// alternative (allow only 19001, then Pass) lets listener ingress fall
+			// through to the user's default-deny and silently breaks every Gateway in a
+			// default-deny namespace. The cost of allowing all TCP is that an Allow is
+			// terminal in this tier, so a user cannot narrow ingress to the proxy with
+			// their own policy.
```
caseydavenport: This comment is massive AI overcommenting - we especially don't need any "Verified on a cluster" type comments.

This should just explain why we want such an open ingress rule.

Do we actually want this? Or do we want to allow users to define where ingress can be allowed from?

pkg/render/gatewayapi/gateway_api_test.go
```diff
@@ -362,6 +365,9 @@ var _ = Describe("Gateway API rendering tests", func() {
 		proxy, err := rtest.GetResourceOfType[*envoyapi.EnvoyProxy](objsToCreate, GatewayClassName, common.CalicoNamespace)
 		Expect(err).NotTo(HaveOccurred())
 		Expect(proxy.Spec.Provider.Kubernetes.EnvoyDeployment.Pod.Labels).To(HaveKeyWithValue("g-rural", "urban"))
+		// Our Calico-owned label is stamped alongside any user-supplied pod labels, so the
+		// calico-system-tier proxy policy can select the proxy pods by a label we control.
```
caseydavenport: More claude overcommenting
```suggestion
```

caseydavenport (changes requested):

pkg/render/gatewayapi/gateway_api.go
electricjesus: done. now in the comment above the `if` and dropped the rest, since `gatewayAPIProxyPolicy`'s own comment says it already.

pkg/render/gatewayapi/gateway_api.go
caseydavenport: @electricjesus I think my last question still stands here - it's a bit sketchy to blanket allow all traffic in a way that can't be overridden by users. Are we sure that's the right model here?

pkg/render/gatewayapi/gateway_api.go
electricjesus: i looked at that first. i don't think it reaches, though.

the switch in `setStandardSelectorAndLabels` builds `k8s-app` from the object's own name (`component.go:1107`, and `addNameLabel` just above it). an EnvoyProxy is named after its GatewayClass (`gateway_api.go:803`), and the user picks that name in `GatewayAPI.Spec.GatewayClasses[].Name`. so the handler would stamp whatever they called the class, and a different value per class if they declare a few. `EnvoyProxyPolicySelector` is a constant, so it can't match that.

we could hardcode the value inside the case. but then the shared handler carries a gateway-specific constant, and it's this helper again with a longer path to it.

there's a smaller snag too. the EnvoyProxy pod spec is `envoyapi.KubernetesPodSpec`, not `corev1.PodTemplateSpec`, so the case can't set `podTemplate` and reuse the tail. it'd label inline like the Prometheus and Alertmanager cases do, and skip the `app.kubernetes.io/*` pod-template stamping. so it wouldn't really be uniform with the Deployment and DaemonSet cases anyway.

happy to move it there if you still want it central.

pkg/render/gatewayapi/gateway_api.go
electricjesus: you were right to push on it. that line wasn't a real reason and it's gone.

it's namespaced now, one `NetworkPolicy` per Gateway namespace instead of the GNP. the RBAC angle you raised is what decided it. the GNP had no namespaceSelector. so any pod anywhere that set `k8s-app: calico-gateway-api-proxy` picked up a terminal all-TCP allow at `calico-system` order 100. a tenant could label their own pod and skip every tier under us. per-namespace that's gone, and they can see the policy in their own namespace.

cost is a little on timing. the policy lands on the reconcile after the Gateway shows up, so a new namespace has a short gap where the proxy is up and the policy isn't.

details on the ingress side are over on the other thread..

pkg/render/gatewayapi/gateway_api.go
electricjesus: yeah, this doesn't hold up. i'd assumed we couldn't know the listener ports. we can.

the controller already lists every Gateway to build `GatewayNamespaces` (`gatewayapi_controller.go:490`), and each item carries `Spec.Listeners[].Port`. we just never read it. `watchGateways` is already registered, so a listener edit re-runs reconcile.

so it's namespaced now, one `NetworkPolicy` per Gateway namespace, ingress allowing that namespace's own listener ports plus 19001 for metrics, then `Pass`. egress is unchanged. the controller policy right above it in the same file is already a namespaced `NetworkPolicy`, so the proxy one being global was the odd one out anyway.

two catches. 19001 isn't a listener, so it goes in by hand. and a new listener port won't serve until the next reconcile. small window, but real.

one thing you'll probably ask: it's owned by the GatewayAPI CR, not by the namespace's Gateways like the other per-namespace stuff. `mergeState` returns nil for a `*v3.NetworkPolicy` whose Spec hasn't changed (`component.go:851`), so the second Gateway's owner ref never merges in. i had a test showing owners `["gw1"]` instead of `["gw1","gw2"]`. rather than touch shared code i left it CR-owned and added a sweep for namespaces that drop out..

this also kills what you poked at on the other thread. the old GNP had no namespaceSelector, so any pod anywhere that set `k8s-app: calico-gateway-api-proxy` inherited a terminal all-TCP allow at order 100. per-namespace, that's gone.

pushed at `a4e01ceb5`, body updated.

pkg/render/gatewayapi/gateway_api_test.go
electricjesus: gone.

caseydavenport:

```

```

PR #4972, opened 2026-06-26
Grant kube-controllers patch permission on tiers

caseydavenport:
## Description

calico-kube-controllers is switching to a merge patch, rather than a full update, when it adds and removes the tier finalizer (see projectcalico/calico#13082). The ClusterRole grants `update` on tiers but not `patch`, and those are distinct RBAC verbs, so the patch would be rejected in operator-managed clusters. This adds `patch` to the tiers rule.

Related: projectcalico/calico#13082

## Release Note

```release-note
Grant calico-kube-controllers permission to patch tiers.
```


electricjesus (approved):
LGTM

pasanw (approved):

PR #4973, opened 2026-06-26
Default new installs to v3 CRD mode when the cluster supports it

caseydavenport:
Makes a brand-new operator-managed install default to v3 CRD mode (projectcalico.org/v3 served directly by CRDs, no aggregation API server) when the cluster can support it. Existing and upgraded clusters are unaffected - if Calico v1 CRDs are already present, the install stays in API-server mode.

The default only flips on a greenfield cluster (no Calico CRDs of either kind yet) that serves MutatingAdmissionPolicy, which v3 mode relies on to default policy types (Kubernetes 1.32+). On older clusters a greenfield install stays on the API server. An admin can still opt a new install back to API-server mode by applying the v1 CRDs before creating the Installation.

This is the operator-only piece of CORE-12574. The Helm and manifest defaults, and the docs updates, are tracked separately. libcalico-go's detection is deliberately left unchanged - components only run after the operator has installed the CRDs, so the existing rule resolves correctly once one set is present.

```release-note
New installs on Kubernetes clusters with MutatingAdmissionPolicy support enabled now default to serving the projectcalico.org/v3 API directly via CRDs instead of through the aggregated API server. Existing and upgraded clusters are unaffected.
```

### DIFF
--- a/pkg/apis/version.go
+++ b/pkg/apis/version.go
@@ -21,6 +21,7 @@ import (
 	v3 "github.com/tigera/api/pkg/apis/projectcalico/v3"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
@@ -34,6 +35,11 @@ var datastoreMigrationGVR = schema.GroupVersionResource{
 	Version:  "v1beta1",
 	Resource: "datastoremigrations",
 }
+
+const (
+	mutatingAdmissionPolicyGroup = "admissionregistration.k8s.io"
+	mutatingAdmissionPolicyKind  = "MutatingAdmissionPolicy"
+)
 
 var log = ctrl.Log.WithName("apis")
 
@@ -122,4 +128,30 @@ func decideV3CRDs(v1present, v3present, mapServed bool) bool {
 		return true
 	}
 	return mapServed
+}
+
+// mutatingAdmissionPolicyServed reports whether the cluster serves the MutatingAdmissionPolicy
+// API (any version). v3 CRD mode relies on a MutatingAdmissionPolicy to default policy types, so
+// a greenfield install only defaults to v3 when this is available (k8s 1.32+).
+//
+// This is best-effort: ServerGroupsAndResources can return a partial result with an error when an
+// aggregated API is unhealthy, so we log and continue with whatever was returned rather than
+// failing. It is only called in the greenfield branch, where no aggregated APIs exist yet.
+func mutatingAdmissionPolicyServed(disco discovery.DiscoveryInterface) bool {
+	_, resourceLists, err := disco.ServerGroupsAndResources()
+	if err != nil {
+		log.Info("Partial error discovering server resources while checking for MutatingAdmissionPolicy; continuing with partial results", "error", err)
+	}
+	for _, rl := range resourceLists {
+		gv, parseErr := schema.ParseGroupVersion(rl.GroupVersion)
+		if parseErr != nil || gv.Group != mutatingAdmissionPolicyGroup {
+			continue
+		}
+		for _, r := range rl.APIResources {
+			if r.Kind == mutatingAdmissionPolicyKind {
+				return true
+			}
+		}
+	}
+	return false
 }
--- a/pkg/apis/version.go
+++ b/pkg/apis/version.go
@@ -56,8 +56,23 @@ func UseV3CRDS(cfg *rest.Config) (bool, error) {
 	if err != nil {
 		return false, err
 	}
-	disco := cs.Discovery()
+	dyn, err := dynamic.NewForConfig(cfg)
+	if err != nil {
+		return false, err
+	}
+	return useV3CRDs(cs.Discovery(), dyn)
+}
 
+// useV3CRDs holds the actual decision logic, taking the discovery and dynamic clients as
+// interfaces so tests exercise it end-to-end with fakes rather than poking at internal helpers.
+//
+//   - If the v1 CRDs are present, the cluster is an existing/upgraded install (or has opted out
+//     by pre-installing v1 CRDs), so use v1.
+//   - If only the v3 CRDs are present, the cluster is already on v3; never downgrade it, but v3
+//     mode needs MutatingAdmissionPolicy, so error out if the cluster can't serve it.
+//   - If neither is present, this is a brand-new install. Default to v3, but only when the cluster
+//     can serve MutatingAdmissionPolicy (needed to default policy types in v3 mode); otherwise v1.
+func useV3CRDs(disco discovery.DiscoveryInterface, dyn dynamic.Interface) (bool, error) {
 	if apiGroup := os.Getenv("CALICO_API_GROUP"); apiGroup != "" {
 		log.Info("CALICO_API_GROUP environment variable is set, using its value to determine API group", "CALICO_API_GROUP", apiGroup)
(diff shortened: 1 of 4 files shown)


pkg/imports/admission/enterprise/protect-builtin-tiers.yaml
```diff
@@ -15,7 +20,7 @@ spec:
         resources: ["tiers"]
         operations: ["DELETE"]
   validations:
-    - expression: "!(oldObject.metadata.name in ['default', 'kube-admin', 'kube-baseline'])"
+    - expression: "!(oldObject.metadata.name in ['default', 'adminnetworkpolicy', 'baselineadminnetworkpolicy'])"
```
Brian-McM, line 23: Is this a result of your changes here or just something you noticed that needed to be fixed?

pkg/apis/version.go
```diff
@@ -103,3 +142,55 @@ func checkDatastoreMigration(cfg *rest.Config) (bool, error) {
...
+
+// mutatingAdmissionPolicyServed reports whether the cluster serves the MutatingAdmissionPolicy
+// API (any version). v3 CRD mode relies on a MutatingAdmissionPolicy to default policy types, so
+// a greenfield install only defaults to v3 when this is available (k8s 1.32+).
+//
+// This is best-effort: ServerGroupsAndResources can return a partial result with an error when an
+// aggregated API is unhealthy, so we log and continue with whatever was returned rather than
+// failing. MutatingAdmissionPolicy is served by the core kube-apiserver, so an unhealthy
+// aggregated API server doesn't affect whether we find it.
+func mutatingAdmissionPolicyServed(disco discovery.DiscoveryInterface) bool {
```
Brian-McM: Nit, personally I prefer having this prefixed with "is" for this framing, "isMutatingAdmissionPolicyServed" or changing to "mutatingAdmissionPolicyExists".

pkg/apis/version.go
```diff
@@ -35,31 +37,60 @@ var datastoreMigrationGVR = schema.GroupVersionResource{
...
+// cluster CRD evidence - the CALICO_API_GROUP override and a converged DatastoreMigration.
+func requireMAPForV3(useV3 bool, disco discovery.DiscoveryInterface) (bool, error) {
+	if useV3 && !mutatingAdmissionPolicyServed(disco) {
+		return false, errV3RequiresMAP
 	}
-	apiGroups, err := cs.Discovery().ServerGroups()
+	return useV3, nil
+}
+
+// useV3CRDsFromDiscovery is the discovery-success path of UseV3CRDS, split out so it can be tested
```
Brian-McM: Personally I really don't like when we have to explicitly test internal bits with direct calls, and would change this so that we pass the k8s interface into UseV3CRDS, or just pass the discovery objects in (I see there's another one used).

Dependency injection can usually get around having to explicitly test internals as opposed to more valuable implicit tests that go through the actual calls the UseV3CRDS calls.

pkg/apis/version.go
```diff
@@ -103,3 +142,55 @@ func checkDatastoreMigration(cfg *rest.Config) (bool, error) {
...
+// override and the DatastoreMigration check are handled by the caller.
+//
+//   - If the v1 CRDs are present, the cluster is an existing/upgraded install (or has opted out
+//     by pre-installing v1 CRDs), so use v1.
+//   - If only the v3 CRDs are present, the cluster is already on v3; never downgrade it, but v3
+//     mode needs MutatingAdmissionPolicy, so error out if the cluster can't serve it.
+//   - If neither is present, this is a brand-new install. Default to v3, but only if the cluster
+//     can serve MutatingAdmissionPolicy (needed to default policy types in v3 mode); otherwise
+//     fall back to v1.
+func decideV3CRDs(v1present, v3present, mapServed bool) (bool, error) {
```
Brian-McM: Nit, but I'm not sure why this function needs to exist. It's pretty a pretty small amount of logic used once that's all testable through fake discovery clients. Personally I find the indirection for this a little less readable than more.

pkg/apis/version.go
```diff
@@ -74,8 +105,16 @@ func UseV3CRDS(cfg *rest.Config) (bool, error) {
...
+	// v1 present always wins (existing/upgraded install), so only pay for the MutatingAdmissionPolicy
+	// discovery call when v3 is still in play - an existing v3 install we must gate, or a greenfield
+	// default we can only take when MAP is servable.
+	mapServed := false
+	if !v1present {
+		mapServed = mutatingAdmissionPolicyServed(disco)
+	}
+
+	log.Info("Detected API groups from API server", "v3present", v3present, "v1present", v1present, "mapServed", mapServed)
+	return decideV3CRDs(v1present, v3present, mapServed)
```
Brian-McM: If v1 is present (which is decided in the loop, then there's no point in any of the logic after the loop. An early retrun gets rid of some of this logic.

pkg/apis/version.go
```diff
@@ -103,3 +142,55 @@ func checkDatastoreMigration(cfg *rest.Config) (bool, error) {
...
+// a greenfield install only defaults to v3 when this is available (k8s 1.32+).
+//
+// This is best-effort: ServerGroupsAndResources can return a partial result with an error when an
+// aggregated API is unhealthy, so we log and continue with whatever was returned rather than
+// failing. MutatingAdmissionPolicy is served by the core kube-apiserver, so an unhealthy
+// aggregated API server doesn't affect whether we find it.
+func mutatingAdmissionPolicyServed(disco discovery.DiscoveryInterface) bool {
+	_, resourceLists, err := disco.ServerGroupsAndResources()
+	if err != nil {
+		log.Info("Partial error discovering server resources while checking for MutatingAdmissionPolicy; continuing with partial results", "error", err)
```
Brian-McM: How can you tell if this is a partial error or a full error, and is it worth it to log at the error level or is this actually expected that the aggregated API will be unhealthy.

I'd think the aggregated API shouldn't be expected to be unhealthy (not sure if there's caveats to this here).

pkg/imports/admission/enterprise/protect-builtin-tiers.yaml
caseydavenport: Just something I had to pull in, not related to the v3-mode logic here. The protected tier list changed in calico-private and `make gen-versions` regenerates this file, so I needed the sync to get the validate-gen-versions check green.

pkg/apis/version.go
caseydavenport: Good call, tightened it to only swallow `ErrGroupDiscoveryFailed` (the partial case where some aggregated APIService is unhealthy but the healthy groups, including the core `admissionregistration.k8s.io` that serves MAP, still come back). Any other discovery error now logs at error level and treats MAP as unavailable. The partial tolerance is deliberate: we don't want an unrelated aggregated API being down (Calico's own apiserver mid-rollout, metrics-server, whatever) to block the CRD-mode decision.

pkg/imports/admission/enterprise/protect-builtin-tiers.yaml
```diff
@@ -1,7 +1,12 @@
...
 # ValidatingAdmissionPolicy that prevents deletion of the built-in Calico tiers
-# (default, kube-admin, kube-baseline). These tiers are required for correct
-# operation and should never be deleted.
+# (default, adminnetworkpolicy, baselineadminnetworkpolicy). These tiers are
+# required for correct operation and should never be deleted.
+#
+# NOTE: Calico Enterprise does not yet support ClusterNetworkPolicy, so the OSS
+# kube-admin / kube-baseline tiers are not created here and are not protected.
+# When ClusterNetworkPolicy lands, add kube-admin / kube-baseline to the list
+# below (see EnsureInitialized in libcalico-go/lib/clientv3/client.go).
```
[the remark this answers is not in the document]
caseydavenport: This file is generated by make gen-versions (synced from calico-private), so the comment wording has to change in the calico-private source rather than here.

Brian-McM (approved):

PR #4980, opened 2026-06-29
Un-fork operator-cloud: bring cloud differences behind a runtime Cloud gate

Brian-McM:
## Description

**Type of change:** new feature (Calico Cloud support / un-fork of `tigera/operator-cloud`).

This brings the differences carried by the long-lived `tigera/operator-cloud` fork back into
`tigera/operator`, so the fork can be retired. All cloud behavior is gated behind the **`CALICO_CLOUD`
environment variable** (`cloud.EnableCloudEnvVar`), set on the operator Deployment by the cloud
installer. When the gate is off (every Calico / Calico Enterprise install), behavior is unchanged —
the cloud code paths are inert.

**Why merge this:** it removes the maintenance burden of a separate fork that had to be continually
re-synced with this repo, and makes Calico Cloud a first-class, gated mode of the upstream operator.

**How it works:**
- `pkg/cloud` parses `CALICO_CLOUD` at startup and returns `Cloud=false` (no-op) unless it is set,
  threading `options.ControllerOptions.Cloud` to controllers.
- Controllers populate typed per-component cloud extension structs (e.g. `render.ManagerCloudResources`)
  only when `opts.Cloud` is true; render decorators are no-ops otherwise.
- Cloud version pins are generated from `config/cloud_versions.yml` into `pkg/components/cloud.go` via a
  new standalone `make gen-versions-cloud` target (not part of the default `gen-versions` aggregate, so
  enterprise generated output is byte-identical).

**Components affected (all gated):** manager, logstorage (linseed / esgateway / kibana / elastic /
external-elastic / dashboards / kube-controllers + ES ILM), api-server RBAC, compliance, fluentd, tiers,
intrusion-detection, policy-recommendation, monitor, packet-capture. Intrusion-detection additionally
cleans up orphaned Image Assurance resources.

**Testing:**
- `make build` / `go build ./pkg/... ./cmd/...` succeed.
- Unit tests pass for every touched package, including new cloud-path tests (manager / fluentd / kibana /
  esgateway / linseed render and elastic / compliance / manager controllers), each with cloud-on and
  cloud-off (negative) cases confirming enterprise output is unchanged.
- Branch rebased cleanly on `master`; cloud version generation is idempotent.

A full migration map and rationale is in `docs/cloud-unfork-migration-plan.md`.

Prerequisite context: `tigera/operator-cloud#1059` (removal of dead Image Assurance / runtime security)
landed in the fork first.

## Release Note

```release-note
NONE
```

## For PR author

- [x] Tests for change.
- [ ] If changing pkg/apis/, run `make gen-files` (no `api/v1` / CRD changes in this PR)
- [x] If changing versions, run `make gen-versions` (`make gen-versions-cloud` → `pkg/components/cloud.go`)

## For PR reviewers

- [ ] Appropriate labels:


cmd/main.go
```diff
@@ -331,7 +341,17 @@ If a value other than 'all' is specified, the first CRD with a prefix of the spe
 		// not being this mapper (which has since been rectified). It was a tough issue to figure out when the default
 		// had changed out from under us, so better to continue to explicitly set it as we know this is the mapper we want.
 		MapperProvider: apiutil.NewDynamicRESTMapper,
-	})
+	}
+
+	if cloudOpts.Cloud {
```
caseydavenport: Is this real? I'm a bit surprised this would be needed for CC and not for everyone else who uses the operator 🤷

cmd/main.go
```diff
@@ -518,7 +538,9 @@ If a value other than 'all' is specified, the first CRD with a prefix of the spe
 		ShutdownContext:     ctx,
 		K8sClientset:        clientset,
 		MultiTenant:         multiTenant,
-		ElasticExternal:     discovery.UseExternalElastic(bootConfig),
+		ElasticExternal:     discovery.UseExternalElastic(bootConfig) || cloudOpts.ElasticExternal,
```
caseydavenport: UseExternalElastic is _almost_ the same thing as ElasticExternal, I think - the main different possibly being the name of the config map that the EXTERNAL_ELASTIC var is set in? 

Is there a way we can avoid having this multi-headed config knob?

cmd/main.go
```diff
@@ -518,7 +538,9 @@ If a value other than 'all' is specified, the first CRD with a prefix of the spe
...
+		Cloud:               cloudOpts.Cloud,
```
caseydavenport: Fine for now, I think longer term this should be folded into the ProductVariant

docs/cloud-unfork-migration-plan.md
```diff
@@ -0,0 +1,304 @@
+# Un-forking `operator-cloud` into `tigera/operator`
```
caseydavenport: Don't think we want to commit this plan here?

hack/gen-versions/main.go
```diff
@@ -22,47 +22,58 @@ import (
...
-	} else {
-		log.Println("must specify either -os-versions or -ee-versions")
-		flag.PrintDefaults()
-		os.Exit(1)
+	case cloudVersionsPath != "":
+		if err := run(cloudVersionsPath, filepath.Join(templateDir, cloudVersionsTpl)); err != nil {
+			log.Fatalln(err)
+		}
 	}
 }
```
caseydavenport, line 79: Can't comment on them, but we shouldn't be checking in binary files (.pyc) below.

hack/hashrelease/management_cluster_updater.py
```diff
@@ -0,0 +1,146 @@
...
+import argparse
+import logging
+import subprocess
+import os
+import sys
+
+import requests
+import ruamel.yaml
+
+class ManagementClusterUpdater:
```
caseydavenport: Not sure this is the right home for this? Seems like this belongs in the cloud tooling repos somewhere instead of the operator?

pkg/cloud/watch.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+
+	go informer.Run(make(chan struct{}))
+	for !informer.HasSynced() {
+		time.Sleep(1 * time.Second)
+	}
+	return nil
+}
+
+func compareMap(m1, m2 map[string]string) bool {
```
caseydavenport: there's a go std lib function for this now.

pkg/cloud/watch.go
```diff
@@ -0,0 +1,82 @@
...
+	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/cache"
+)
+
+var configWatchLog = ctrl.Log.WithName("cloud_config_watch")
+
+// watch spawns a goroutine which should exit if a configmap's data is changed.
+// it is stubbed for testing.
+var watch = func(cs kubernetes.Interface, cmData map[string]string) error {
```
caseydavenport: No need to change for this PR, but longer term seems like this should be handled by a controller rather than goroutine spawned as a side-effect of Load() (if we need this configmap long term at all)

pkg/controller/apiserver/apiserver_controller.go
```diff
@@ -496,6 +496,7 @@ func (r *ReconcileAPIServer) Reconcile(ctx context.Context, request reconcile.Re
 		KeyValidatorConfig:           keyValidatorConfig,
 		KubernetesVersion:            r.opts.KubernetesVersion,
 		ClusterDomain:                r.opts.ClusterDomain,
+		Cloud:                        r.opts.Cloud,
```
caseydavenport, line 539: I've already started doing this in other places - I'll handle it in my own refactoring but worth calling out that I'm moving to just having the ControllerOptions as a field on the controller instead of copying each var individually.

pkg/controller/compliance/compliance_controller.go
```diff
@@ -158,6 +166,28 @@ func newReconciler(mgr manager.Manager, opts options.ControllerOptions, licenseA
 	return r
 }
 
+// NewReconcilerWithShims constructs a ReconcileCompliance with the provided dependencies. It is
+// intended for tests that need to inject a fake client, scheme, and status manager.
+func NewReconcilerWithShims(
```
caseydavenport: The type `ReconcileCompliance` is public, so not sure why we need this?

pkg/controller/compliance/compliance_controller.go
```diff
@@ -411,6 +441,20 @@ func (r *ReconcileCompliance) Reconcile(ctx context.Context, request reconcile.R
 		return reconcile.Result{}, nil
 	}
 
+	// Calico Cloud: update the trusted bundle with system root certificates on management clusters,
+	// as they may be needed by compliance-server when the OIDC provider is external.
+	// TODO: we should do this closer to instantiation of the trustedBundle but can't because that code
```
caseydavenport: Is this TODO fixable now that it's merged upstream?

pkg/controller/logstorage/dashboards/dashboards_controller.go
```diff
@@ -272,11 +274,22 @@ func (d DashboardsSubController) Reconcile(ctx context.Context, request reconcil
 			return reconcile.Result{RequeueAfter: utils.StandardRetry}, nil
 		}
 	} else {
-		// If we're using an external ES and Kibana, the Tenant resource must specify the Kibana endpoint.
-		if tenant == nil || tenant.Spec.Elastic == nil || tenant.Spec.Elastic.KibanaURL == "" {
-			reqLogger.Error(nil, "Kibana URL must be specified for this tenant")
-			d.status.SetDegraded(operatorv1.ResourceValidationError, "Kibana URL must be specified for this tenant", nil, reqLogger)
-			return reconcile.Result{}, nil
+		if d.multiTenant || !d.cloud {
```
caseydavenport: Hm. This feels a little bit wrong to me - might just come out in the wash if EE doesn't use external ES at all, but an EE cluster that is not multi-tenant will unconditionally error here? 

If EE doesn't support this branch, then why do we need the `!d.cloud` clause? 

Might just need an explainer comment.

pkg/controller/logstorage/elastic/elastic_controller.go
```diff
@@ -480,6 +483,24 @@ func (r *ElasticSubController) Reconcile(ctx context.Context, request reconcile.
 			r.status.SetDegraded(operatorv1.ResourceReadError, "Failed to retrieve the Kibana service", err, reqLogger)
 			return reconcile.Result{}, err
 		}
+
+		if r.cloud {
+			// Read the cloud-kibana-config ConfigMap and parse it into a global map in the render package. The render code will read this map
+			// and use it to override the default Kibana configuration.
+			// TODO: We should instead be passing this via arguments to the render code.
```
caseydavenport: Can we fix this TODO now that we're merged upstream?

pkg/controller/logstorage/elastic/external_elastic_controller.go
```diff
@@ -101,6 +102,13 @@ func AddExternalES(mgr manager.Manager, opts options.ControllerOptions) error {
 		return fmt.Errorf("log-storage-external-es-controller failed to watch tigera-elasticsearch namespace: %w", err)
 	}
 
+	if opts.Cloud {
+		// Calico Cloud addition.
```
caseydavenport: nit: comment is redundant with the code here.

pkg/controller/logstorage/linseed/linseed_controller.go
```diff
@@ -310,11 +320,22 @@ func (r *LinseedSubController) Reconcile(ctx context.Context, request reconcile.
 			return reconcile.Result{RequeueAfter: utils.StandardRetry}, nil
 		}
 	} else {
-		// If we're using an external ES, the Tenant resource must specify the ES endpoint.
-		if tenant == nil || tenant.Spec.Elastic == nil || tenant.Spec.Elastic.URL == "" {
-			reqLogger.Error(nil, "Elasticsearch URL must be specified for this tenant")
-			r.status.SetDegraded(operatorv1.ResourceValidationError, "Elasticsearch URL must be specified for this tenant", nil, reqLogger)
-			return reconcile.Result{}, nil
+		if r.multiTenant || !r.cloud {
```
caseydavenport: Hm. common code that maybe could use a helper func to avoid desychronizing?

pkg/controller/manager/manager_controller.go
```diff
@@ -484,6 +490,31 @@ func (r *ReconcileManager) Reconcile(ctx context.Context, request reconcile.Requ
 	if err != nil {
 		r.status.SetDegraded(operatorv1.ResourceCreateError, "Error creating trusted bundle for manager", err, logc)
 	}
+
+	// Handle all the resources that are specific to Calico Cloud. For non-cloud installs this is
+	// skipped entirely, leaving mcr at its zero value and enterprise behavior unchanged.
+	var mcr render.ManagerCloudResources
+	if r.opts.Cloud {
+		var reconcileResult *reconcile.Result
+		bundleMaker, mcr, tenant, reconcileResult, err = r.handleCloudReconcile(
```
caseydavenport, line 482: This is probably fine, but I think we probably only structure the code this way because it used to be a fork and we didn't want to interleave merge conflicts - we should probably unpack this as a follow on to clean it up instead of overwriting / recalculating what was done above.

pkg/controller/options/options.go
```diff
@@ -48,6 +48,15 @@ type ControllerOptions struct {
 	// and instead will configure the cluster to use an external Elasticsearch.
 	ElasticExternal bool
 
+	// Cloud indicates the operator is running as a Calico Cloud install. When set, controllers
```
caseydavenport: "Calico Cloud Install" -> I assume this actually means "Calico Cloud Management Cluster" ?

pkg/controller/tiers/tiers_controller_cloud.go
```diff
@@ -0,0 +1,47 @@
...
+import (
+	"context"
+
+	v3 "github.com/tigera/api/pkg/apis/projectcalico/v3"
+	"github.com/tigera/operator/pkg/render/common/networkpolicy"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// cloudPatchTier removes the allow-tigera tier's app.kubernetes.io/instance label to fix Calico
```
caseydavenport: allow-tigera doesn't exist any more. 

But also, can't we do this inline now instead of a separate patch step? i.e., an if statement in the render code?

caseydavenport: I think a lot of the code is structured the way it is due to the history of being a fork and wanting to minimize conflicts - but now that it's upstreamed that structuring adds complexity rather than removing it.

Makefile
```diff
@@ -169,6 +169,19 @@ EXCLUDE_MANIFEST_REGISTRIES?=""
 PUSH_MANIFEST_IMAGE_PREFIXES=$(PUSH_IMAGE_PREFIXES:$(EXCLUDE_MANIFEST_REGISTRIES)%=)
 PUSH_NONMANIFEST_IMAGE_PREFIXES=$(filter-out $(PUSH_MANIFEST_IMAGE_PREFIXES),$(PUSH_IMAGE_PREFIXES))
 
+# Calico Cloud build variant. `make <target> VARIANT=cloud` builds and pushes the operator-cloud
+# image to GCR (gcr.io/tigera-tesla/operator-cloud), amd64 only, instead of the enterprise quay
+# image. Enterprise/OSS builds (VARIANT unset) are completely unaffected. PUSH_MANIFEST_IMAGE_PREFIXES
+# and PUSH_NONMANIFEST_IMAGE_PREFIXES above use recursive `=`, so they pick up these overrides.
+ifeq ($(VARIANT),cloud)
```
radTuti, line 190: if we are setting `VARIANT=cloud`, do we need all the separate `?-cloud` targets? (maybe `gen-versions-cloud`)

pkg/controller/utils/cloudconfig.go
```diff
@@ -0,0 +1,69 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// CloudAuthConfig is the name of the ConfigMap holding cloud authentication configuration
+// (e.g. the tenantID) for a single-tenant cloud management cluster backed by internal Elasticsearch.
+const CloudAuthConfig = "cloud-auth-config"
+
+// GetCloudConfig retrieves the config map containing the configuration values needed to set up communications with
+// external Elasticsearch and Kibana, such as the externalESDomain and externalKibanaDomain.
+func GetCloudConfig(ctx context.Context, cli client.Client) (*cloudconfig.CloudConfig, error) {
```
caseydavenport, line 37: Does it make sense for this to be separate from the other utility functions in `pkg/cloud/` ?

pkg/render/common/test/testing.go
```diff
@@ -260,7 +260,7 @@ func ExpectVolumeMount(vms []corev1.VolumeMount, name, path string) {
 			return
 		}
 	}
-	Expect(false).To(BeTrue(), fmt.Sprintf("Missing expected volume mount %s", name))
+	ExpectWithOffset(1, false).To(BeTrue(), fmt.Sprintf("Missing expected volume mount %s", name))
```
caseydavenport, line 263: Glad we finally fixed this one upstream 👍

pkg/render/kubecontrollers/kube-controllers.go
```diff
@@ -116,6 +116,14 @@ type KubeControllersConfiguration struct {
 	WASMCACert                   *corev1.ConfigMap
 	TrustedBundle                certificatemanagement.TrustedBundleRO
 
+	// Calico Cloud additions. TenantId is only set by the cloud-gated controller path; when empty
+	// (regular Calico/Calico Enterprise) no cloud env is emitted.
+	TenantId string
```
caseydavenport:
```suggestion
	TenantID string
```

pkg/render/logstorage/esgateway/cloud.go
```diff
@@ -0,0 +1,198 @@
...
+func removeEnv(evs []corev1.EnvVar, name string) []corev1.EnvVar {
+	for i, ev := range evs {
+		if ev.Name == name {
+			evs = append(evs[:i], evs[i+1:]...)
+		}
+	}
+	return evs
+}
+
+func (e *esGateway) modifyDeploymentForCloud(d *appsv1.Deployment) {
```
caseydavenport: Does anywhere in cloud still use es-gateway?

pkg/render/logstorage/esgateway/cloud.go
```diff
@@ -0,0 +1,198 @@
...
+		// Enable this so that fluentd cannot modify ILM but the POSTs fluentd performs
+		// still succeed.
+		envs = append(envs, corev1.EnvVar{
+			Name:  "ES_GATEWAY_ILM_DUMMY_ROUTE_ENABLED",
+			Value: "true",
+		})
+	}
+
+	if e.cfg.Cloud.EnableMTLS {
+		// todo: delete these from the envVars
```
caseydavenport: When can we address this TODO? Doesn't really say.

pkg/render/logstorage/esgateway/esgateway.go
```diff
@@ -304,8 +310,8 @@ func (e *esGateway) esGatewayDeployment() *appsv1.Deployment {
 		}
 	}
 
+	e.modifyDeploymentForCloud(&d)
```
caseydavenport: I'd say this pattern is wrong, but tbh it does look a fair bit like what I'm doing in https://github.com/tigera/operator/pull/4871

So maybe it's actually best to keep it like this and eventually I rebase my PR and refactor lots of these to use the new extension mechanism instead.

pkg/render/logstorage/kibana/kibana.go
```diff
@@ -66,6 +66,11 @@ const (
 
 var EntityRule = networkpolicy.CreateEntityRule(Namespace, CRName, Port)
 
+// CloudKibanaConfigOverrides holds Calico Cloud Kibana config overrides. It is populated only by the
+// cloud-gated controller path; for regular Calico/Calico Enterprise it stays empty and is a no-op.
+// TODO: This shouldn't be done with a global variable set by the controller; thread it through Configuration instead.
```
caseydavenport: hear, hear

Makefile
```diff
@@ -169,6 +169,19 @@ EXCLUDE_MANIFEST_REGISTRIES?=""
...
+REPO:=tigera/operator-cloud
```
caseydavenport: Think this might need to be tweaked?

Makefile
```diff
@@ -551,6 +570,23 @@ hack/bin/release: $(shell find ./hack/release -type f)
 	sh -c '$(GIT_CONFIG_SSH) \
 	go build -buildvcs=false -o hack/bin/release ./hack/release'
 
+# Calico Cloud release tooling. The cloud release tool is the same tool compiled with `-tags cloud`,
```
caseydavenport: Do we need the build tags now? Would rather avoid those if possible.

CC @tigera/release-team / @radTuti to consider how we integrate the release tooling here.

cmd/main.go
Brian-McM: It probably wouldn't hurt to use it for everybody.

cmd/main.go
Brian-McM: Maybe lets just do that now, I didn't actually think about that.

Brian-McM: I looked into folding this into `ProductVariant` and I don't think it's the right move:

- `Variant` is user-facing config on the `Installation` CR (enum-validated), but "this is a Calico Cloud management cluster" is an operator-deployment property that the cloud installer sets — not a product a user picks. They're different layers.
- Cloud installs essentially all the enterprise components, so a `CalicoCloud` variant would have to be treated as enterprise everywhere (all ~76 `IsEnterprise()` call sites), and we'd *still* need a separate cloud flag for the cloud-specific modifications on top. So it doesn't actually remove the knob — it adds a variant on top of it.
- Exact `== CalicoEnterprise` comparisons (e.g. the CRD watches in `core_controller.go`) would silently not match a new variant value, which is an easy source of subtle breakage.

So I'd keep it as the deployment-level `CALICO_CLOUD` flag: cloud runs as `CalicoEnterprise` + `Cloud=true`, which leaves the EE behavior intact and keeps the cloud tweaks orthogonal. Happy to revisit if cloud ever diverges from EE enough to justify a real variant.

cmd/main.go
caseydavenport: I think it would potentially slow down rolling update?

cmd/main.go
caseydavenport: @Brian-McM yeah, definitely not for this PR - that's what I meant to say by "longer term"

But we could have a `IsEnterprise` and an `IsCloud` (and IsEnterprise would return true for both).

> Happy to revisit if cloud ever diverges from EE enough to justify a real variant.

I don't think it's about divergence, it's about keeping our configuration model clean.

Makefile
```diff
@@ -551,6 +570,23 @@ hack/bin/release: $(shell find ./hack/release -type f)
...
+# which activates hack/release/cloud.go (GCR/tesla image defaults, cloud-vX.Y.Z version format,
+# hashrelease support). The regular `release`/`release-publish` targets above are unaffected.
+hack/bin/release-cloud: $(shell find ./hack/release -type f)
```
radTuti: I don't we should need a new binary build to be able to get a cloud release. The release binary should handle be able to handle both.

Brian-McM: My preference would be that we use a new binary so that you can't accidentally modify an env variable and have the operator uninstall cloud components.

This is managing part of the cloud infrastructure, so even if it costs a little more to build and release I would much prefer the guarantee that the cloud flags are always set in the binary when running in cloud.

cmd/main.go
Brian-McM: Ohh right, ya it would. It seems like there's a better solution to this issue, how long is the apiserver taking that we need this sort of duration?

hack/gen-versions/main.go
Brian-McM: Ya that was certain a mistake.

Also for future reference you can comment on files, there's a little text box next to the "Viewed" check box on the right.

pkg/render/logstorage/esgateway/cloud.go
Brian-McM: We still have single tenant clusters which I think still use this.

Makefile
Brian-McM: Landed here: one release binary, no build tags. `cloud.go`'s cloud behavior now activates at runtime when `VARIANT=cloud`, so `make release VARIANT=cloud` / `make release-tag VARIANT=cloud …` drive the cloud release from the same `hack/bin/release` — your "the release binary should handle both" point. Removed the `-tags cloud` build and the `release-cloud` / `release-publish-cloud` / `release-tag-cloud` targets.

Brian's "flags must be guaranteed" concern is handled a layer down: `VARIANT=cloud` bakes cloud mode into the cloud *operator image* at build time (`-ldflags -X …/pkg/cloud.buildVariant=cloud` → `pkg/cloud.IsCloudBuild()`), so a mutated Deployment env can't flip cloud off. Two images (enterprise + cloud), one release binary.

Makefile
Brian-McM: Collapsed the separate `-cloud` targets into `VARIANT=cloud` on the existing `release` / `release-publish` / `release-tag` targets, so there's no `release-cloud` twin anymore. Kept `gen-versions-cloud` (as you suggested) since it generates a distinct file (`pkg/components/cloud.go`) rather than a variant of an existing target's output.

cmd/main.go
Brian-McM: > But we could have a IsEnterprise and an IsCloud (and IsEnterprise would return true for both).

Ya that would be much nicer, I shouldn't have let Claude respond on behalf of me there lol.

I might just end up doing the refactor here, so I'm going to keep this open as a maybe todo.

docs/cloud-unfork-migration-plan.md
Brian-McM: removed

pkg/controller/manager/manager_controller.go
Brian-McM: Oh I fully intend on doing that right now, this is the initial pass.

cmd/main.go
Brian-McM: I'm going to use the discovery config map and ensure that cloud creates that cm. I think we might be able to get rid of the cloud specific CR, but for now I'll just update both with the same values.

.semaphore/push_images.yml
```diff
@@ -34,14 +34,20 @@ global_job_config:
...
     task:
       secrets:
         - name: quay-robot-semaphore_v2
         - name: operator-redhat-connect
+        - name: google-service-account-for-gcr
       prologue:
         commands:
           - docker login -u="$QUAY_USERNAME" -p="$QUAY_TOKEN" quay.io;
+          - gcloud auth activate-service-account --key-file=/home/semaphore/secrets/secret.google-service-account-key.json
+          - docker login -u _json_key -p "$(cat ~/secrets/secret.google-service-account-key.json)" https://gcr.io
```
radTuti: i think it is better to have registry cred isolation: move the quay auth to "Build" job and gcloud to "Build Cloud" job

.semaphore/release.yml
```diff
@@ -40,16 +40,23 @@ global_job_config:
...
     task:
       secrets:
         - name: quay-robot-semaphore_v2
         - name: operator-redhat-connect
+        - name: google-service-account-for-gcr
       prologue:
         commands:
           - docker login -u="$QUAY_USERNAME" -p="$QUAY_TOKEN" quay.io;
+          - gcloud auth activate-service-account --key-file=/home/semaphore/secrets/secret.google-service-account-key.json
+          - docker login -u _json_key -p "$(cat ~/secrets/secret.google-service-account-key.json)" https://gcr.io
```
radTuti: same comment as above

config/cloud_versions.yml
tmjd, line 1: This doesn't seem to be needed since the cc-rbac-api isn't deployed by operator for cloud.

pkg/components/combined.go
```diff
@@ -30,3 +30,13 @@ func CombinedCalicoImage(installation *operatorv1.InstallationSpec) Component {
 	}
 	return ComponentCalico
 }
+
+// CalicoCloudImage returns the tesla-compiled variant of the combined calico image. It is the same
+// image repository as ComponentTigeraCalico (so ImageSet digests still resolve by that name),
+// published under a tesla- tag. It carries the Calico Cloud behavior for the components that need it
+// — currently only kube-controllers, the sole component with tesla-gated code. See TSLA-11580.
```
tmjd: Maybe we should also point to https://tigera.atlassian.net/browse/TSLA-11650 for possible removal of this code.

.semaphore/release.yml
```diff
@@ -58,6 +65,13 @@ blocks:
           env_vars:
             - name: CONFIRM
               value: "true"
+        - name: Publish Cloud Release
+          # Same commit as the enterprise job; -cloud suffix marks the cloud image/version.
+          commands:
+            - make release-tag VARIANT=cloud RELEASE_TAG=${SEMAPHORE_GIT_TAG_NAME}-cloud
```
radTuti, line 71: what doesn't passing `VARIANT=cloud` handle adding the `-cloud`?

hack/release/utils.go
```diff
@@ -79,10 +83,6 @@ func addRepoInfoToCtx(ctx context.Context, repo string) (context.Context, error)
 	return ctx, nil
 }
 
-// isValidReleaseVersion validates the operator release version format.
-// It defaults to standard release format (vX.Y.Z) but can be overridden if a different format is needed.
-var isValidReleaseVersion = isReleaseVersionFormat
-
 // isReleaseVersionFormat checks if the version in the format vX.Y.Z.
 func isReleaseVersionFormat(version string) (bool, error) {
```
radTuti: is this still used/needed?

Makefile
```diff
@@ -628,6 +662,14 @@ gen-versions-calico: $(BINDIR)/gen-versions update-calico-crds
 gen-versions-enterprise: $(BINDIR)/gen-versions update-enterprise-crds
 	$(BINDIR)/gen-versions -ee-versions=$(EE_VERSIONS) > pkg/components/enterprise.go
 
+# gen-versions-cloud regenerates pkg/components/cloud.go from config/cloud_versions.yml. It is a
+# Calico Cloud concern and is intentionally NOT part of the default `gen-versions` aggregate, so the
+# enterprise build's generated output is unchanged. The cloud release pipeline invokes this target.
+# Unlike the calico/enterprise targets, it does not fetch CRDs (cloud relies on the upstream operator
+# repo for CRD updates).
+gen-versions-cloud: $(BINDIR)/gen-versions
```
radTuti: the release tool should probably call this when building the operator-cloud. I don't think it currently does now

cmd/main.go
caseydavenport: > how long is the apiserver taking that we need this sort of duration?

Indeed, that's basically my question. This smells a lot like code that was added to fix a symptom rather than a root cause and that might not be needed any more.

hack/hashrelease/management_cluster_updater.py
Brian-McM: Removed completely.

pkg/render/logstorage/esgateway/esgateway.go
Brian-McM: Ya it's definitely wrong and I'm removing it. Just inline it without the decorator pattern.

You're pattern is good for what you're doing, the code will eventually be split across two repos and will be required. Those no need to decorate here anymore for cloud.

Makefile
Brian-McM: Removed.

Makefile
Brian-McM: Update on this, it's not one release binary, it's one release tool. Two separate binaries, to separate images.

config/cloud_versions.yml
Brian-McM: Removed.

.semaphore/push_images.yml
Brian-McM: Done

.semaphore/release.yml
Brian-McM: Done.

Makefile
Brian-McM: This actually isn't needed at all anymore since @tmjd pointed out the RBAC api isn't actually installed here in cloud.

hack/hashrelease/management_cluster_updater.py
Glen-Tigera: FYI this is used for bleeding-edge cluster promotion, would this be migrated to a different repo?

Brian-McM: Not sure exactly, we'll find the right spot for auto promotion after we figure out how to get this working and released.

It might just end up back in here, but I don't want to deal with that as part of this PR.

Glen-Tigera: Sounds good

pkg/controller/logstorage/elastic/elastic_controller.go
```diff
@@ -480,6 +483,24 @@ func (r *ElasticSubController) Reconcile(ctx context.Context, request reconcile.
...
+			kbCm := &corev1.ConfigMap{}
+			if err = r.client.Get(ctx, types.NamespacedName{Name: "cloud-kibana-config", Namespace: common.OperatorNamespace()}, kbCm); err != nil {
+				if !errors.IsNotFound(err) {
```
caseydavenport, line 491: When `cloud-kibana-config` gets deleted we land in this NotFound branch and leave `CloudKibanaConfigOverrides` set to whatever the last reconcile wrote, so stale overrides stick around until the operator restarts. We should reset it to empty on NotFound too.

pkg/render/logstorage/kibana/kibana.go
```diff
@@ -66,6 +66,11 @@ const (
...
+var CloudKibanaConfigOverrides = map[string]interface{}{}
```
caseydavenport: Can we thread this through `Configuration` instead of a package global? A global set as a side effect of the controller is easy to leave stale (see my comment on the elastic controller), and it's the one spot in here that won't fold cleanly into the variant work later. Same shape as what you did for the compliance bundle.

pkg/cloud/cloud.go
```diff
@@ -0,0 +1,214 @@
...
+	if !cloudEnabled {
+		setupLog.Info("cloud mode not enabled; running in non-cloud mode")
+		return Options{Cloud: false}, nil
+	}
+
+	var cmData map[string]string
+	cloudConfig, err := cs.CoreV1().ConfigMaps(common.OperatorNamespace()).Get(ctx, configMapName, metav1.GetOptions{})
+	if err != nil {
+		if !kerrors.IsNotFound(err) {
+			return Options{}, fmt.Errorf("failed to read configmap '%s': %v", configMapName, err)
```
caseydavenport: We should probably use `%w` here (and throughout this file) so the wrapped errors stay unwrappable with `errors.Is`/`errors.As`.

Not super important but a nice consistency cleanup.

cmd/main.go
caseydavenport: @Brian-McM bumping this one - did you get a chance to dig into the apiserver latency numbers? Still want to make sure these longer durations aren't just papering over a slow apiserver rather than fixing the actual cause.

pkg/controller/logstorage/dashboards/dashboards_controller.go
caseydavenport: @Brian-McM bump on this - does an EE non-multitenant cluster actually hit this error path? Want to be sure we're not breaking EE here. If the `!d.cloud` clause is load-bearing, can we get an explainer comment?

pkg/controller/logstorage/linseed/linseed_controller.go
caseydavenport: @Brian-McM did you get a chance to look at this one? Wondering if a shared helper would keep these two paths from drifting apart over time.

pkg/controller/utils/cloudconfig.go
caseydavenport: @Brian-McM any thoughts here? Feels like it'd sit more naturally alongside the other helpers in `pkg/cloud/`.

cmd/main.go
Brian-McM: Just looking into this now, this was something added a long time ago so might not be relevant anymore, cc @tmjd.

pkg/controller/logstorage/dashboards/dashboards_controller.go
Brian-McM: An EE non multitenant cluster (which is all EE clusters) hit's this path but then does the check below it which is the same check as before.

The else branch is only ever hit if this is a single tenant cloud, in which case we have external ES.

I'm wondering if I should just if this as:
```
if d.cloud && !d.multiTenant {
  
} else if tenant == nil || tenant.Spec.Elastic == nil || tenant.Spec.Elastic.KibanaURL == "" {

}
```

That's doesn't functionally changing, but maybe reads better?

pkg/controller/utils/cloudconfig.go
Brian-McM: Might be splitting hairs here but this logic is pretty controller centric, while the pkg/cloud is used for initial setup.

cmd/main.go
Brian-McM: Chatted with @tmjd and I'm going to remove the hard coded override but put in flags to update these values  in case we need to quickly override these again in cloud.

Sound good @caseydavenport ?

pkg/cloud/watch.go
Brian-McM: We actually already do this with utils.MonitorConfigMap, and it's in our main.go file.

I'm not sure if this is an appropriate case for a controller, that might be overkill, given that this isn't for reconciliation. We just want to kill the process if change is detected.

This should also be unneeded since we're picky backing on the bootstrap config now and we can hopefully get rid of that duplicate cloud config map.

Brian-McM: This is removed now.

pkg/cloud/cloud.go
Brian-McM: File is gone.

cmd/main.go
```diff
@@ -86,6 +86,18 @@ var (
 // configuration for the operator loaded at startup.
 const bootstrapConfigMapName = "operator-bootstrap-config"
 
+// buildVariant is injected at build time via -ldflags "-X main.buildVariant=cloud" when building the
+// Calico Cloud operator image (see CLOUD_LDFLAGS in the Makefile). It is empty for the regular
+// Calico/Calico Enterprise image. Baking it into the binary means cloud mode is immutable: it cannot
+// be disabled by editing the operator Deployment's environment.
```
caseydavenport: > Baking it into the binary means cloud mode is immutable: it cannot be disabled by editing the operator Deployment's environment.

Just a bit of a pressure test... why is this important?

cmd/main.go
```diff
@@ -121,6 +140,22 @@ func main() {
 		&enableLeaderElection, "enable-leader-election", true,
 		"Enable leader election for controller manager. Enabling this will ensure there is only one active controller manager.",
 	)
+	// Leader-election timings. Defaults match the controller-runtime defaults (i.e. the values a
+	// regular Calico Enterprise install runs with). They are exposed as flags so a deployment on a
+	// higher-latency API server (e.g. a Calico Cloud management cluster) can loosen them without a
+	// separate code path.
```
caseydavenport:
```suggestion
```

self-explanatory and inconsistent with the other flags.

cmd/main.go
```diff
@@ -518,9 +578,14 @@ If a value other than 'all' is specified, the first CRD with a prefix of the spe
 		ShutdownContext:     ctx,
 		K8sClientset:        clientset,
 		MultiTenant:         multiTenant,
-		ElasticExternal:     discovery.UseExternalElastic(bootConfig),
-		UseV3CRDs:           v3CRDs,
-		APIDiscovery:        apiDiscovery,
+		// External-ES is a single knob, sourced only from the operator bootstrap configmap
+		// (operator-bootstrap-config) via discovery.UseExternalElastic, shared by both cloud and
+		// enterprise so there is one downstream knob rather than a per-variant source.
```
caseydavenport:
```suggestion
```

Claude likes to document historical behavior ... but it just leads to bloat. The current behavior is self-explanatory.

pkg/controller/logstorage/dashboards/dashboards_controller.go
caseydavenport: I think `d.cloud && !d.multiTenant` as you framed it above is more clear.

Alternatively, this?

```go
if d.multiTenant {
  // Any multi-tenant cluster.
} else if d.cloud {
  // CC single-tenant.
}
```

With either of those I think we could get rid of the big "load bearing" comment that was added above.

pkg/controller/logstorage/linseed/linseed_controller.go
```diff
@@ -320,6 +320,15 @@ func (r *LinseedSubController) Reconcile(ctx context.Context, request reconcile.
 			return reconcile.Result{RequeueAfter: utils.StandardRetry}, nil
 		}
 	} else {
+		// External ES. Two ways to obtain the tenant's ES endpoint:
```
caseydavenport: Should make sure the structure here mirrors whatever we end up on with the equivalent Kibana change above.

caseydavenport (commented):
Just a few more thoughts / questions. Probably the last round though!

cmd/main.go
Brian-McM: I don't want us to somehow have some weird backdoor of setting a operator cloud image into EE mode.

cmd/main.go
Brian-McM: I really just need to tell claude to stop commenting altogether, that bot has opposite commenting preferences then I do where it describes every bit of decision and line of code in a comment where I only explain when it's not pretty clear from the code what's going on.

I'll clean all the comments up.

caseydavenport: Yeah, I have the same problem

cmd/main.go
caseydavenport: I'm generally against having two builds of this code if we can avoid it. 

I don't see there being a huge risk of someone having a backdoor, and if they do it's not actually a problem? It just won't work, right?

Brian-McM: The risk is that a cloud environment gets turned into an ee environment and cloud production cluster is hosed, even worse if it's a multi tenant one. 

I don't like uneccessarily having multiple builds, but I'll stand by it being worth it in this case.

caseydavenport: Yeah, I just don't see why that's a risk worth propagating into our build + code, rather than just having CI that catches misconfigurations before merging manifest changes. A CLI argument would be pretty hard to mess up, especially if we had some pre-flight checks that caught a misconfiguration and prevented the operator from taking action.

Brian-McM: I sort of think the opposite, why push it to testing when you can just make it not possible. I'm not sure this is huge burden having two builds here tbh, but I'd just like to roll with this for my peace of mind and if it is for some reason actual issue then maybe we can just revisit then.

pkg/controller/logstorage/dashboards/dashboards_controller.go
Brian-McM: Done.

caseydavenport (approved):

hack/release/README.md
```diff
@@ -87,6 +87,19 @@ By default, it creates a draft release.
      make release-publish VERSION=vA.B.C DRAFT_GITHUB_RELEASE=false
      ```
 
+### Calico Cloud variant
+
+Run any release target with `VARIANT=cloud` to build the Calico Cloud operator instead of the
+enterprise one, e.g. `make release-tag VARIANT=cloud RELEASE_TAG=vA.B.C-cloud`. It is the same
```
tmjd, line 93: Is it correct here that `RELEASE_TAG=vA.B.C-cloud` includes the suffix `-cloud`?
In the next paragraph it says
> There are no separate cloud release tags or pipelines

pkg/render/apiserver.go
```diff
@@ -1891,19 +1915,22 @@ func (c *apiServerComponent) tigeraUserClusterRole() *rbacv1.ClusterRole {
...
+	}...)
 
 	// Privileges for lma.tigera.io have no effect on managed clusters.
 	if c.cfg.ManagementClusterConnection == nil {
 		// Access to flow logs, audit logs, and statistics.
 		// Access to log into Kibana for oidc users.
+		// Calico Cloud additionally grants access to runtime logs.
+		resourceNames := []string{"flows", "audit*", "l7", "events", "dns", "waf", "kibana_login", "recommendations"}
+		if c.cfg.Cloud {
+			resourceNames = append([]string{"runtime"}, resourceNames...)
```
tmjd, line 1927: Do we even need to include this anymore?

Brian-McM: ooof, I'm not sure, I would think not....

PR #4983, opened 2026-06-30
Don't autoscale typha after the autoscaler is shut down

caseydavenport:
The typha autoscaler unit tests flake intermittently with a panic from an unexpected mock call (a degraded-status update fired after the test had already finished).

The "initialize an autoscaler" test starts an autoscaler with no nodes. Its goroutine waits for the informer to sync, then runs a startup autoscale that reports degraded (one typha wanted, zero linux nodes). The test never cancels and waits for that goroutine, so the degraded call lands on a later spec's mock and crashes the suite. Ginkgo's randomized ordering decides which spec gets hit, which is why it only shows up sometimes.

The autoscaler now bails out of its startup path as soon as its context is cancelled, and exposes a way to wait for the goroutine to exit. The tests cancel and wait in AfterEach so a leaked goroutine can't contaminate a later spec.

```release-note
None
```

### DIFF
--- a/pkg/controller/installation/typha_autoscaler.go
+++ b/pkg/controller/installation/typha_autoscaler.go
@@ -147,8 +147,7 @@ func (t *typhaAutoscaler) start(ctx context.Context) {
 			}
 		}
 
-		// If we were cancelled while waiting for the informers to sync, don't autoscale or
-		// report degraded - we're shutting down.
+		// Don't autoscale or report degraded if the context has been cancelled - we're shutting down.
 		if ctx.Err() != nil {
 			typhaLog.Info("typha autoscaler shutting down")
 			return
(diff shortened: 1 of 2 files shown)


pkg/controller/installation/typha_autoscaler.go
```diff
@@ -124,17 +127,31 @@ func newTyphaAutoscaler(cs kubernetes.Interface, indexInformer cache.SharedIndex
...
-			time.Sleep(100 * time.Millisecond)
+			select {
+			case <-ctx.Done():
+				typhaLog.Info("typha autoscaler shutting down")
+				return
+			case <-time.After(100 * time.Millisecond):
+			}
+		}
+
+		// If we were cancelled while waiting for the informers to sync, don't autoscale or
```
pasanw: nit: comment seems a bit misleading? It seems more appropriate for the check within the for loop. The check here is not catching cancellation while we wait for sync (it's catching cancellation before or after we wait?)

pasanw (approved):

PR #4987, opened 2026-07-03
[CI-1951] fix(podiprecovery): add Pod watch so recovery is level-triggered

coutinhop:
Follow-up to #4784. Two fixes to the podiprecovery controller.

## 1. Recovery was edge-triggered (add a Pod watch)

The Node watch alone fired only when a node's host IPs changed. On a KubeVirt
reboot, the node's new IP is reported while its host-networked pods are still
restarting (empty `status.podIPs`), so the reconcile skips them; when a
surviving pod later comes back with its old, now-stale IP, the node IP has
already settled and no further Node event fires — so the stale pod is never
re-checked.

Add a second watch on operator-managed host-networked Pods that re-enqueues a
pod's node when it settles (IPs appear/change, or it becomes Ready). Both
watches feed the same idempotent, node-keyed Reconcile, making recovery
level-triggered on both inputs (node addresses + pod IPs) while staying
event-driven. The predicate is gated on the host-networked label +
`spec.hostNetwork`.

## 2. Don't force a cluster-wide Pod cache (use a scoped cache)

As written, #4784 registered a `Pod.spec.nodeName` field index in
`cmd/main.go` and read pods via the shared cache — either forces a
cluster-wide Pod informer holding every pod in memory, always-on from startup.
That's a sizable memory cost at scale and a regression (the operator's other
pod readers are conditional, so the shared Pod informer was previously
lazy/absent).

Give the controller its own `cache.New` scoped **server-side** to the
`operator.tigera.io/host-networked` label; the watch, the per-node List, and
the `spec.nodeName` index all live on it, so only the handful of
host-networked pods are cached and the shared cache is untouched. The
`cmd/main.go` field index is removed (supersedes that part of #4784). The
scoped reader is wrapped in a `hostNetworkedPodLister` exposing only
"list host-networked pods on a node", so it can't be misused as a full-pod
reader (which would silently drop non-host-networked pods).

## Changes
- `podiprecovery`: Pod watch + `podToNode` + `hostNetPodSettledPredicate`;
  dedicated label-scoped `cache.New` (with `spec.nodeName` index), wrapped in
  `hostNetworkedPodLister`.
- `ctrlruntime`: add `WatchObjectInCache` (watch via a supplied cache).
- `cmd/main.go`: remove the shared-cache `Pod.spec.nodeName` index from #4784.
- update `mockController` in utils/egressgateway tests for the new method.

## Testing
Unit tests for the pod-settle predicate and `podToNode` mapping;
`go build`/`go vet ./...` clean; podiprecovery/utils/egressgateway tests pass.

## Release Note

```release-note
None
```

For PR author

- [x] Tests for change.
- [ ] If changing pkg/apis/, run make gen-files
- [ ] If changing versions, run make gen-versions

For PR reviewers

- [ ] Appropriate labels:
  - [x] kind/bug
  - [ ] kind/enhancement
  - [ ] enterprise

### DIFF
--- a/pkg/controller/podiprecovery/podiprecovery_controller.go
+++ b/pkg/controller/podiprecovery/podiprecovery_controller.go
@@ -33,6 +33,15 @@
 // spec.hostNetwork == true as a side effect of the normal apply path.
 // The controller additionally verifies spec.hostNetwork == true on each
 // candidate as a safety net before deleting.
+//
+// Recovery is triggered by two watches so it stays level-triggered rather
+// than reacting to a single node event: a Node watch (fires when a node's
+// host IPs change) and a Pod watch (fires when a host-networked pod finishes
+// (re)starting). The Pod watch is what catches a pod that was still
+// restarting at the moment its node's IP changed and only later comes back
+// reporting its old, stale IP — after which no further Node event would
+// fire. Both watches enqueue the same Node key and share one idempotent
+// Reconcile.
 package podiprecovery
 
 import (
@@ -41,6 +50,7 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/sets"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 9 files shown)


pkg/controller/podiprecovery/podiprecovery_controller.go
```diff
@@ -83,9 +131,37 @@ func Add(mgr manager.Manager, _ options.ControllerOptions) error {
 		return fmt.Errorf("podiprecovery-controller failed to watch Nodes: %w", err)
 	}
 
+	// Watch host-networked pods too, and re-enqueue the pod's node when one
```
caseydavenport: This comment (and a few others) are really way too long. We're duplicating explanations of the same behavior / fix in a few places (as Claude likes to do) and it makes the code pretty tough to read. 

I think this probably doesn't need a comment at all.

pkg/controller/podiprecovery/podiprecovery_controller.go
```diff
@@ -112,6 +188,73 @@ func hostIPsChangedPredicate() predicate.Predicate {
...
+// Add). It fires when such a pod finishes (re)starting — either its reported
+// IPs change, or it transitions to Ready — which is the moment a pod that
+// survived a node-IP change comes back reporting its old, now-stale IP.
+func hostNetPodSettledPredicate() predicate.Predicate {
+	return predicate.Funcs{
+		CreateFunc: func(e event.CreateEvent) bool {
+			pod, ok := e.Object.(*corev1.Pod)
+			// An already-running host-net pod observed on (re)start of the
+			// controller: evaluate it if it already reports IPs.
+			return ok && isManagedHostNetPod(pod) && len(pod.Status.PodIPs) > 0
```
caseydavenport, line 194: Do we need `isManagedHostNetPod` now that we're using a label selector on our watch? Seems redundant?

caseydavenport (approved):

PR #4991, opened 2026-07-06
feat(apiserver): provision per-tenant calico-apiserver Linseed RBAC

Brian-McM:
## Description

**Type of change:** new feature (Calico Enterprise, multi-tenant only).

On a multi-tenant management cluster the aggregated API server stays a single, cluster-scoped component in `calico-system` — there is no per-tenant API server deployment. But each tenant's managed clusters run a `calico-apiserver` (queryserver) that reads Linseed `policyactivity`, authenticating with a Linseed-signed token whose identity is namespace-overridden to the tenant namespace (`system:serviceaccount:<tenant-namespace>:calico-apiserver`). Linseed authorizes every request with a **cluster-scoped `SubjectAccessReview`** against the management cluster (the token only authenticates), so the management cluster must grant that tenant-namespaced identity `linseed.tigera.io/policyactivity` access via a `ClusterRoleBinding`. This PR provisions that RBAC.

This mirrors how compliance, intrusion-detection, and the manager already grant their managed-cluster components Linseed access — all of which carry the same `TenantNamespaceOverride` in Linseed's token controller.

**What changed:**
- Added a dedicated least-privilege `calico-apiserver-linseed-access` ClusterRole (only `linseed.tigera.io/policyactivity: get`).
- Bound it via a **single** ClusterRoleBinding whose subjects are the `calico-apiserver` `ServiceAccount` in every tenant namespace (one `ServiceAccount` subject per tenant), rendered from the API server component over `BindingNamespaces` using the shared `rcomponents.ClusterRoleBinding` helper. Multi-tenant only; cleaned up on zero/single-tenant clusters.
- The apiserver controller computes `BindingNamespaces` by listing the current tenant namespaces during reconcile, and watches `Tenant` resources so the binding's subject list is re-rendered when tenants are added or removed. Because it is one shared object rendered over the current namespace set, it **self-heals** on tenant deletion (the deleted tenant's subject is simply dropped) — no per-tenant cluster-scoped objects to garbage-collect.
- Restored the full `calico-apiserver` ClusterRole to its ungated rule set and scoped its binding to the `calico-system` API server only. (An in-progress change had gated those enterprise rules behind multi-tenant, which stripped them from single-tenant Enterprise; this restores correct behavior.)

The API server reconcile is unchanged in shape: a single idempotent path that renders the full component (the tenant binding included) on every trigger. No per-tenant dispatch, and no `ServiceAccount` is created in tenant namespaces — the SA object is not required, since Linseed's SAR matches the synthesized username string whether or not the object exists.

**Affected components:** `pkg/controller/apiserver`, `pkg/render/apiserver.go`.

**Testing:**
- `go test ./pkg/render/...` and `go test ./pkg/controller/apiserver/...` pass.
- Render specs assert: the `calico-apiserver` ClusterRoleBinding binds only the `calico-system` ServiceAccount; the Linseed-access ClusterRole is least-privilege; and its single binding has one `calico-apiserver` ServiceAccount subject per tenant namespace.
- A controller spec asserts that, with multiple tenants present, one reconcile produces the `calico-apiserver-linseed-access` ClusterRoleBinding with a subject per tenant namespace.

## Release Note

```release-note
NONE
```

## For PR author

- [x] Tests for change.

### DIFF
--- a/pkg/controller/apiserver/apiserver_controller.go
+++ b/pkg/controller/apiserver/apiserver_controller.go
@@ -255,14 +255,25 @@ func (r *ReconcileAPIServer) Reconcile(ctx context.Context, request reconcile.Re
 	reqLogger := log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
 	reqLogger.V(2).Info("Reconciling APIServer")
 
-	// In multi-tenant mode a request scoped to a tenant namespace is handled by the per-tenant path: the
-	// aggregated API server itself is a cluster-scoped, calico-system component (reconciled by the
-	// empty-namespace request), so all a tenant namespace needs is the calico-apiserver ServiceAccount.
-	// That ServiceAccount is the subject the shared calico-apiserver ClusterRoleBinding grants Linseed
-	// access to, allowing this tenant's managed clusters to reach Linseed. We short-circuit here to avoid
-	// running any of the cluster-scoped API server setup (certificates, deployment, etc.) for tenants.
+	// In multi-tenant mode, a request scoped to a tenant namespace is handled by the per-tenant path, which
+	// provisions only that tenant's Linseed-access RBAC. The aggregated API server itself is a single
+	// cluster-scoped component in calico-system, reconciled by the cluster-scoped path below.
+	//
+	// We must only divert to the per-tenant path for genuine tenant namespaces. Many of this controller's
+	// watches (secrets and config maps in the operator namespace, the API server deployment in calico-system,
+	// etc.) enqueue requests carrying a non-empty, non-tenant namespace; those must fall through to the
+	// cluster-scoped reconcile, or the API server would stop reconciling on those events. A namespace is a
+	// tenant namespace iff it contains a Tenant, so we key off that rather than "namespace is non-empty".
 	if r.opts.MultiTenant && request.Namespace != "" {
-		return r.reconcileTenant(ctx, request.Namespace, reqLogger)
+		tenant, _, err := utils.GetTenant(ctx, true, r.client, request.Namespace)
+		if err != nil && !errors.IsNotFound(err) {
+			r.status.SetDegraded(operatorv1.ResourceReadError, "An error occurred while querying Tenant", err, reqLogger)
+			return reconcile.Result{}, err
+		}
+		if tenant != nil {
+			return r.reconcileTenant(ctx, tenant, reqLogger)
+		}
+		// No Tenant in this namespace - fall through to the cluster-scoped reconcile.
 	}
 
 	// Check if a datastore migration is in progress. If so, the migration controller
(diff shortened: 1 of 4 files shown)


tmjd (commented):
> 🤖 **Note: this review was drafted by Claude (Claude Code), at Erik's request.** Erik reviewed the investigation but the analysis, code tracing, and wording below are Claude's. Treat the findings as a starting point for discussion, not authoritative verdicts — please sanity-check the claims against your own knowledge of the system.

---

I traced the end-to-end policyactivity auth path to understand the requirement this PR addresses, and confirmed the premise is sound: the managed cluster's queryserver presents a Linseed-signed JWT whose subject is overridden to the tenant namespace (`linseed/cmd/main.go:309` registers `calico-apiserver` with `TenantNamespaceOverride`, `GenerateSubjectLinseed` applies it), and Linseed's middleware runs a `SubjectAccessReview` against the management cluster for `system:serviceaccount:<tenant-ns>:calico-apiserver` (`linseed/pkg/middleware/token.go`). So the management cluster genuinely needs RBAC granting that tenant-namespaced identity `policyactivity get`, which is what this PR provides. 👍

That said, I have four points — the first could meaningfully simplify the change.

**1. Does the `ServiceAccount` object itself actually need to exist?**

The authorization is a `SubjectAccessReview` on a *synthesized* username, and the token is a Linseed **self-signed JWT** (`token_controller.go` `createToken`), not a Kubernetes `TokenRequest`. As far as I can tell, neither the SAR nor the JWT requires the `ServiceAccount` object to exist on the management cluster — only the `ClusterRole` + `ClusterRoleBinding` (with the tenant-namespaced SA as subject) are load-bearing.

If that's correct, then creating the per-tenant `ServiceAccount` — and much of the machinery to make the apiserver controller run per tenant just to create it — may be unnecessary; you'd only need the CR/CRB carrying the tenant-namespaced subject. Could you confirm whether the SA object is required by anything? If not, dropping it removes a lot of the per-tenant surface here.

**2. The `MultiTenant && request.Namespace != ""` dispatch swallows the cluster-scoped API server's own reconcile events.**

```go
if r.opts.MultiTenant && request.Namespace != "" {
    return r.reconcileTenant(ctx, request.Namespace, reqLogger)
}
```

This treats *any* namespaced request as a tenant request, but several existing watches (inside `if opts.EnterpriseCRDExists`, which MT always is) enqueue the object's **own** namespace via `EnqueueRequestForObject`: `calico-apiserver-certs`, the CA secret, Dex/Prometheus/Voltron-Linseed cert secrets, the `k8s-service-endpoint` ConfigMap, and the `tigera-apiserver`/`calico-apiserver`/webhooks Deployments (namespaces `tigera-operator`, `calico-system`, `calico-apiserver`). Those now route to `reconcileTenant`, hit `GetTenant` → `NotFound`, and are silently dropped.

Net effect: in multi-tenant mode the cluster-scoped API server no longer reconciles in response to cert rotations, CA-bundle changes, or deployment drift — it only self-heals on the periodic reconcile. I'd suggest dispatching on a positive tenant-namespace check rather than "namespace is non-empty" (or routing non-tenant namespaced requests to the cluster path).

**3. The per-tenant `ClusterRole` and `ClusterRoleBinding` are not garbage-collected on tenant teardown.**

The code and comments state all three objects are owned by the `Tenant` and GC'd when the tenant is removed. That holds for the namespaced `ServiceAccount`, but not for the two cluster-scoped objects: `componentHandler`'s `skipAddingOwnerReference` (`pkg/controller/utils/component.go:607`) deliberately skips the owner reference when a namespaced owner (Tenant) would own a cluster-scoped object. So `calico-apiserver-linseed-access-<ns>` (ClusterRole + ClusterRoleBinding) get **no** owner ref, and nothing deletes them on teardown (`reconcileTenant` early-returns when the Tenant is gone). They'll accumulate per deleted tenant. These are exactly the objects that matter for #1, so they need an explicit cleanup path (finalizer, or delete-when-tenant-absent), and the comment/description should be corrected.

**4. (Minor) The per-tenant `ClusterRole` is identical for every tenant.**

`tenantLinseedAccessClusterRole(ns)` produces byte-identical rules for all tenants, differing only in name. A single shared least-privilege `calico-apiserver-linseed-access` ClusterRole + per-tenant `ClusterRoleBinding` referencing it gives the same isolation with half the cluster-scoped objects (the per-tenant *binding* is what you need for the cluster-scoped SAR and per-tenant subject; the per-tenant *role* isn't).

---

Since this is up for discussion: the Linseed token controller already enumerates tenants and registers `calico-apiserver` per tenant (`linseed/cmd/main.go:309`), so the tenant/namespace knowledge exists elsewhere already — worth weighing whether newly making the apiserver controller tenant-aware is the right home for this RBAC, versus provisioning it where per-tenant state is already managed.

> 🤖 *Drafted by Claude (Claude Code). Flagging AI authorship for transparency — please verify before acting on any of it.*

tmjd (approved):

PR #4992, opened 2026-07-06
node: mount securityfs into calico-node for lockdown detection

tomastigera:
### Description

Felix detects kernel `lockdown=confidentiality` by reading `/sys/kernel/security/lockdown`, and under that level loads trace-printk-free variants of its BPF preamble programs. This avoids the kernel logging `could not enable bpf_trace_printk events` on every BPF program load (ftrace is disabled at boot under confidentiality lockdown, e.g. Talos).

calico-node only mounts `/sys/fs` (sysfs), **not** securityfs, so from inside the container the lockdown file is absent and detection always returns `false` — leaving the workaround inert on exactly the nodes it targets. This mounts securityfs read-only at `/sys/kernel/security`.

No `hostPath` type is set (matching the existing `/proc`/`nodeproc` mount) so nodes without securityfs still start; Felix treats an unreadable lockdown file as "not locked down".

Companion to projectcalico/calico#13142 (CORE-13107), which adds the same mount to the manual manifests / Helm charts. Verified end-to-end on a `lockdown=confidentiality` node: with the mount, Felix loads the `_notrace` preambles and the per-attach dmesg spew drops to zero.

### For PR author

- [x] Tests for change.
- [x] If changing pkg/apis/, run `make gen-files` to regenerate CRDs/docs. (n/a — no API change)
- [x] If changing versions, run `make gen-versions`. (n/a)

### Release Note

```release-note
Mount securityfs into calico-node so Felix can detect kernel lockdown=confidentiality and avoid bpf_trace_printk-related kernel log spam.
```


tmjd (approved):
Looks good to me as an operator-maintainer.

PR #4998, opened 2026-07-08
fix(waf): sync WAFPolicy CRD with cross-scope plugin CEL rule

electricjesus:
## Description

Bundled-CRD sync for a bug fix in calico-private. A namespaced `WAFPolicy` could name a cluster-scoped `GlobalWAFPlugin` in `spec.plugins[].kind`. Admission accepted it, but the reconciler always resolves a plugin ref as a namespace-scoped `WAFPlugin` and ignores `kind`, so the ref never resolved and failed later with a confusing `PluginNotFound`.

The calico-private PR adds a CRD CEL rule (`x-kubernetes-validations`) on `WAFPolicy.spec.plugins` that rejects the cross-scope ref at admission. kube-apiserver enforces it no matter what, so it backstops the `/validate-waf` webhook when the webhook isn't there (cert not provisioned yet, gateway license off, or an older operator that hasn't wired the webhook config).

This PR is the operator side. It syncs the bundled enterprise WAFPolicy CRD (`pkg/imports/crds/enterprise/applicationlayer.projectcalico.org/applicationlayer.projectcalico.org_wafpolicies.yaml`) so the operator ships the rule. The diff is exactly the CEL block plus the field description, and it matches what a full `make fetch-crds` re-sync from calico-private master will produce once the paired PR lands.

- Type: bug fix (generated-CRD sync)
- Affects: enterprise WAFPolicy CRD only. No controller or render changes.
- Paired with https://github.com/tigera/calico-private/pull/12606
- Jira: EV-6753

## Release Note

```release-note
Fixed a namespaced WAFPolicy that referenced a GlobalWAFPlugin being silently accepted and then failing with PluginNotFound. Such cross-scope plugin references are now rejected at admission with a clear message.
```

## For PR author

- [x] Tests for change. The webhook and CEL behavior are tested in the paired calico-private PR; this PR is a generated-CRD sync, so no operator-side test applies.
- [x] If changing pkg/apis/, run `make gen-files`. N/A, no `api/v1` change; the bundled CRD under `pkg/imports/crds` is synced from calico-private, not produced by `gen-files`.
- [x] If changing versions, run `make gen-versions`. N/A.

### DIFF
--- a/pkg/imports/crds/enterprise/applicationlayer.projectcalico.org/applicationlayer.projectcalico.org_wafpolicies.yaml
+++ b/pkg/imports/crds/enterprise/applicationlayer.projectcalico.org/applicationlayer.projectcalico.org_wafpolicies.yaml
@@ -95,10 +95,17 @@ spec:
                   type: object
                 plugins:
                   description: |-
-                    Plugins references WAFPlugins in this namespace. Cross-scope references
-                    (kind: GlobalWAFPlugin) are not yet resolved by the reconciler, so the
-                    CEL rule rejects them at admission as a backstop to the validating
-                    webhook — see EV-6753 and the PluginRef doc above.
+                    Plugins references WAFPlugins in this namespace. A cross-scope reference
+                    (kind: GlobalWAFPlugin) is rejected at admission by the CEL rule below and
+                    the validating webhook, because the reconciler resolves plugin scope from
+                    the parent policy and does not yet honor Kind. See EV-6753.
+
+                    Only this namespaced direction is guarded. The mirror case, a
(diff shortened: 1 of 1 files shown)


electricjesus:
Closing — superseded by the auto-sync bot. The "Sync versions" workflow regenerates the bundled enterprise CRDs from calico-private (`make gen-versions` → `update-enterprise-crds`), and https://github.com/tigera/operator/pull/5005 already carries this exact cross-scope CEL rule, pulled straight from cp#12606 on calico-private master. Hand edits under `pkg/imports/crds/enterprise/` fail gen/dirty-check by design. release-v1.43 picks it up via the same bot once cp#12681 propagates. Tracking in EV-6800.

PR #2545, opened 2023-03-11
Remove LogCollector Splunk public certificate

hjiawei:
## Description

This feature was never officially released to our customer and was mostly tested internally before. When the public certificate isn't available in the cluster, it continuously outputs "Assuming http protocol" in operator logs. This causes confusions to our customers.

## For PR author

- [ ] Tests for change.
- [ ] If changing pkg/apis/, run `make gen-files`
- [ ] If changing versions, run `make gen-versions`

## For PR reviewers

A note for code reviewers - all pull requests must have the following:

- [ ] Milestone set according to targeted release.
- [ ] Appropriate labels:
  - `kind/bug` if this is a bugfix.
  - `kind/enhancement` if this is a a new feature.
  - `enterprise` if this PR applies to Calico Enterprise only.

PR #5055, opened 2026-07-16
logcollector: give fluent-bit its own trusted bundle; user CA for Splunk HEC

hjiawei:
## Description

Bug fix plus a small follow-on feature for the fluent-bit log collector, in two commits:

**1. Render fluent-bit's trusted bundle under its own name** (bug fix, cherry-pick candidate)

The fluentd→fluent-bit migration moved the log collector into `calico-system` but kept rendering its trusted bundle as `tigera-ca-bundle` — the same ConfigMap the core Installation controller renders there with a different certificate set. The component handler replaces ConfigMap data wholesale and flips the controllerRef on every write, so the two controllers overwrite each other and the additions only the logcollector makes (the `syslog-ca` user CA, BYO prometheus/linseed certs) were lost to whichever reconcile ran last. Net effect: syslog forwarding over TLS with a user-supplied CA never worked — fluent-bit failed `certificate verify failed` because the CA never persisted in the bundle it mounts. Pre-migration there was no collision because fluentd's bundle lived in `tigera-fluentd`.

The fix follows the pattern of the other components sharing `calico-system` (`calico-manager-ca-bundle`, `whisker-ca-bundle`, `policy-recommendation-ca-bundle`): a named bundle, `calico-fluent-bit-ca-bundle-system-certs`. Mount directory and data keys are unchanged, so the rendered fluent-bit configuration is byte-identical; only the DaemonSet volume reference changes (one DaemonSet roll on upgrade).

**2. User CA for Splunk HEC + watch the CA ConfigMaps** (feature)

The Splunk output already verifies `https` HEC endpoints against the trusted bundle, but nothing could put a private CA there — a self-hosted Splunk with a self-signed certificate was unusable since the splunk-public-certificate secret was removed (#2545). Accept one the same way syslog does: a `splunk-ca` ConfigMap in `tigera-operator` (key `tls.crt`), folded into fluent-bit's bundle. The controller also now watches the `syslog-ca`/`splunk-ca` ConfigMaps, so creating or rotating a CA takes effect immediately instead of waiting for an unrelated reconcile.

**Testing**

- New regression tests: the bundle CM name and syslog CA content, the shared `tigera-ca-bundle` no longer rendered by this controller, and the Splunk CA path; render suites updated to the named bundle. Rendered-config goldens are unchanged by design.
- Validated live on an Enterprise dev cluster (Linux + Windows nodes): with this change the syslog user CA persists in the mounted bundle and the calico-private TLS syslog e2e passes end to end (53s), and the Splunk e2e forwards flow logs to an https HEC endpoint serving Splunk's self-signed certificate via `splunk-ca` (5m10s). Companion calico-private e2e changes (Splunk TLS mode) and a docs update for `splunk-ca` (tigera/docs#2815) follow.

Part of the fluent-bit migration epic (EV-6164/EV-6684).

## Release Note

```release-note
- Splunk HTTP Event Collector endpoints with a custom CA are supported via the splunk-ca ConfigMap.
```

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:
  - `kind/bug` if this PR fixes a bug

### DIFF
--- a/pkg/controller/logcollector/logcollector_controller.go
+++ b/pkg/controller/logcollector/logcollector_controller.go
@@ -142,6 +142,14 @@ func add(mgr manager.Manager, c ctrlruntime.Controller) error {
 		return fmt.Errorf("logcollector-controller failed to watch ConfigMap %s: %v", rlogcollector.FluentBitFilterConfigMapName, err)
 	}
 
+	// Watch the user-supplied CA ConfigMaps so creating or rotating a syslog or
+	// Splunk CA takes effect without waiting for an unrelated reconcile.
+	for _, caConfigMap := range []string{rlogcollector.SyslogCAConfigMapName, rlogcollector.SplunkCAConfigMapName} {
+		if err = utils.AddConfigMapWatch(c, caConfigMap, common.OperatorNamespace(), &handler.EnqueueRequestForObject{}); err != nil {
+			return fmt.Errorf("logcollector-controller failed to watch ConfigMap %s: %v", caConfigMap, err)
+		}
+	}
+
 	// Watch the rendered configuration ConfigMaps so tampering with them
 	// triggers a reconcile that restores the rendered content.
 	for _, configMapName := range []string{
@@ -524,7 +532,7 @@ func (r *ReconcileLogCollector) Reconcile(ctx context.Context, request reconcile
 	var useSyslogCertificate bool
 	if instance.Spec.AdditionalStores != nil {
 		if instance.Spec.AdditionalStores.Syslog != nil && instance.Spec.AdditionalStores.Syslog.Encryption == operatorv1.EncryptionTLS {
-			syslogCert, err := getSysLogCertificate(r.client)
+			syslogCert, err := getUserCACertificate(r.client, rlogcollector.SyslogCAConfigMapName)
 			if err != nil {
 				r.status.SetDegraded(operatorv1.ResourceReadError, "Error loading Syslog certificate", err, reqLogger)
 				return reconcile.Result{}, err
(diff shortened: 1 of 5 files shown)


pkg/controller/logcollector/logcollector_controller.go
```diff
@@ -534,6 +542,19 @@ func (r *ReconcileLogCollector) Reconcile(ctx context.Context, request reconcile
 				trustedBundle.AddCertificates(syslogCert)
 			}
 		}
+		// The Splunk output verifies https HEC endpoints against the trusted
+		// bundle, so a user CA for a self-hosted Splunk rides the same way as
+		// the syslog one.
+		if instance.Spec.AdditionalStores.Splunk != nil {
```
pasanw: We should consider gating on whether this is an HTTPS connection (like we do above for Syslog)

pkg/render/logcollector/logcollector.go
```diff
@@ -73,6 +73,7 @@ const (
 	SysLogPublicCertKey                      = "ca-bundle.crt"
 	SysLogPublicCAPath                       = SysLogPublicCADir + SysLogPublicCertKey
 	SyslogCAConfigMapName                    = "syslog-ca"
+	SplunkCAConfigMapName                    = "splunk-ca"
```
pasanw, line 76: Two thoughts (unrelated to line)
* The release note on the PR description should not have the line about the bundle rename
* Let's make sure we track the docs changes needed for this CA feature

pasanw (approved):

pkg/controller/logcollector/logcollector_controller.go
hjiawei: Done — the CA is now loaded only for https endpoints, parsed with `url.ParseEndpoint` the same way the render decides TLS for the splunk output, plus a test that an http endpoint leaves the CA out of the bundle. (Syslog keeps its `encryption: TLS` gate since TLS-ness there is a field, not the scheme.)

pkg/render/logcollector/logcollector.go
hjiawei: Both done:
- Release note trimmed to the splunk-ca line only.
- Docs for the CA feature are already written on the tigera/docs migration branch (`EV-6164-fluent-bit-docs`, PR tigera/docs#2815): commit 9618b8600 adds a `splunk-ca` step to the Splunk archiving instructions (mirroring the syslog-ca step) and a release-notes line. That docs PR should land with this one.

PR #4910, opened 2026-06-10
Replace fluentd with fluent-bit in the operator

hjiawei:
## Description

New feature (Calico Enterprise): the LogCollector controller now deploys **fluent-bit** (`calico-fluent-bit` in `calico-system`) in place of fluentd, completing the log-collector migration on the operator side.

- Resource identity migration: namespace `tigera-fluentd` → `calico-system`; DaemonSet/ServiceAccount `fluentd-node` → `calico-fluent-bit`; TLS secret → `calico-fluent-bit-tls`; metrics port 9081 → 2020 (fluent-bit's built-in HTTP server).
- Configuration is rendered in fluent-bit's YAML schema into per-OS ConfigMaps (`calico-fluent-bit-conf`/`-windows`), subPath-mounted on Linux and directory-mounted on Windows, started with `-c`. The render loads the Go plugins via `plugins_file`, defines parsers inline, applies the record-transform Lua filter, and inlines user-provided fluent-bit YAML filter lists. A rendered-config hash annotation rolls the pods on config-only changes.
- Tail inputs use the producing components' real log paths, with SQLite offset DBs and filesystem buffering under `/var/log/calico/calico-fluent-bit`; a `pos-migrator` init container (Linux and Windows) seeds offsets from the legacy fluentd `.pos` files and pre-creates the tailed directories. Windows tails the same log types the fluentd Windows variant shipped.
- The linseed output matches only Linseed-bound tags, authenticates with mTLS + the pod's ServiceAccount token, and retries without limit against bounded filesystem storage. S3/Splunk/Syslog outputs mirror fluentd's per-type fan-out (standard AWS credential env vars, endpoint scheme honored, syslog ships the whole record as JSON via a per-output Lua processor with TLS properly enabled).
- NonClusterHost renders the `:9880` HTTP input with client-certificate verification; the input Service is cleaned up when the resource is removed.
- `eks-log-forwarder` runs the fluent-bit image with a rendered `in_eks` → `linseed` pipeline and health probes (no startup init container; the input plugin resolves its resume point from Linseed).
- Probes hit `:2020/api/v1/health`; the ServiceMonitor scrapes plain HTTP (fluent-bit's monitoring server has no TLS) with access restricted by the component NetworkPolicy, and legacy fluentd monitors are deleted.
[...]

PR #4021, opened 2025-06-26
Fix missing kube-controller secret access

vara2504:
## Release Note

```release-note
TBD
```

## For PR author

- [ ] Tests for change.
- [ ] If changing pkg/apis/, run `make gen-files`
- [ ] If changing versions, run `make gen-versions`

## For PR reviewers

A note for code reviewers - all pull requests must have the following:

- [ ] Milestone set according to targeted release.
- [ ] Appropriate labels:
  - `kind/bug` if this is a bugfix.
  - `kind/enhancement` if this is a a new feature.
  - `enterprise` if this PR applies to Calico Enterprise only.

PR #5056, opened 2026-07-16
logstorage: stop deleting the tigera-linseed RoleBinding it no longer owns

hjiawei:
## Description

Bug fix for managed Enterprise clusters (EV-6837).

The managed-cluster logstorage cleanup still lists the `tigera-linseed` RoleBinding in `calico-system` in `deprecatedObjects()`, but that object is live again: the apiserver controller renders it for the MCM queryserver wiring (#4786) and the logcollector controller renders it since fluent-bit moved into `calico-system` (#4910), both so guardian's identity can manage Linseed token secrets there. The delete fought their create on every reconcile, so the RoleBinding flapped and guardian intermittently lost permission to provision Linseed tokens for calico-system components.

The entry was correct when it was added: the guardian RBAC consolidation (#3969, #4021) obsoleted logstorage's namespace-scoped bindings, and the cleanup removed the stragglers on upgrade. It just was never pruned when new controllers started rendering the same object again.

Found by a static audit of objects rendered by more than one controller (follow-up to #5055's `tigera-ca-bundle` fix — same root pattern: the component handler replaces object state wholesale, so two controllers with different desired state fight). The render test now pins the RoleBinding's absence from the delete list.

## Release Note

```release-note
None
```

No release note: both renders of this RoleBinding (#4786 and the fluent-bit migration #4910) are unreleased, so the create/delete fight never shipped.

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:
  - `kind/bug` if this PR fixes a bug

### DIFF
--- a/pkg/render/logstorage.go
+++ b/pkg/render/logstorage.go
@@ -1372,10 +1372,6 @@ func (m *managedClusterLogStorage) deprecatedObjects() []client.Object {
 			TypeMeta:   metav1.TypeMeta{Kind: "ClusterRole", APIVersion: "rbac.authorization.k8s.io/v1"},
 			ObjectMeta: metav1.ObjectMeta{Name: "tigera-linseed-configmap"},
 		},
-		&rbacv1.RoleBinding{
-			TypeMeta:   metav1.TypeMeta{Kind: "RoleBinding", APIVersion: "rbac.authorization.k8s.io/v1"},
-			ObjectMeta: metav1.ObjectMeta{Name: "tigera-linseed", Namespace: "calico-system"},
-		},
 
 		// Remove legacy ExternalName service pointing to Guardian for linseed
 		&corev1.Service{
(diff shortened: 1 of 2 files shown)


tianfeng92 (approved):

pkg/render/logstorage.go
```diff
@@ -1372,10 +1372,6 @@ func (m *managedClusterLogStorage) deprecatedObjects() []client.Object {
 			TypeMeta:   metav1.TypeMeta{Kind: "ClusterRole", APIVersion: "rbac.authorization.k8s.io/v1"},
 			ObjectMeta: metav1.ObjectMeta{Name: "tigera-linseed-configmap"},
 		},
-		&rbacv1.RoleBinding{
```
hjiawei, line 1375: This is added in apiserver render: https://github.com/tigera/operator/pull/4786/changes#diff-88957f5c859486b2cbf260254b87b36b75fadbc99bfd38c01ea29dc0ac22a260R1345.

Brian-McM (approved):

PR #5065, opened 2026-07-17
istio: Add cleanup for stale istio resources to waypoint controller

alexh-tigera:
If a Gateway class name is changed, it will result in an abandoned resource. Add clean up for these stale istio references.

## Release Note

```release-note
TBD
```

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/istio/waypoint/waypoint_cleanup.go
+++ b/pkg/controller/istio/waypoint/waypoint_cleanup.go
@@ -75,9 +75,8 @@ func (r *ReconcileWaypoint) staleGatewaySets(ctx context.Context, reqLogger logr
 		return nil, fmt.Errorf("failed to list Gateways: %w", err)
 	}
 	gateways := map[types.NamespacedName]*gapi.Gateway{}
-	for i := range gatewayList.Items {
-		gw := &gatewayList.Items[i]
-		gateways[types.NamespacedName{Namespace: gw.Namespace, Name: gw.Name}] = gw
+	for _, gw := range gatewayList.Items {
+		gateways[types.NamespacedName{Namespace: gw.Namespace, Name: gw.Name}] = &gw
 	}
 
 	// The kinds istiod renders for every managed Gateway. It can also render a
--- a/pkg/controller/istio/waypoint/waypoint_cleanup.go
+++ b/pkg/controller/istio/waypoint/waypoint_cleanup.go
@@ -1,0 +1,193 @@
+// Copyright (c) 2026 Tigera, Inc. All rights reserved.
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


pkg/controller/istio/waypoint/waypoint_cleanup.go
```diff
@@ -0,0 +1,292 @@
...
+	lists := []client.ObjectList{
+		&appsv1.DeploymentList{},
+		&corev1.ServiceList{},
+		&corev1.ServiceAccountList{},
+		&policyv1.PodDisruptionBudgetList{},
+	}
+
+	var toDelete []client.Object
+	classManaged := map[string]bool{}
+	for _, list := range lists {
```
Brian-McM, line 96: Correct me if I'm wrong, but you're basically looking for old gateway class names. If you have the name, then you can target the specific set of resources.

I think we might be able to have a more efficient approach of finding the orphaned gateway names, by picking one of the resources and searching for all of that specific resource that has a gateway class name not in the list of current class names.

Like look for all PodDisruptionBUdgetList items that don't have a label in the list of active class names. If non come up, then you're good. If some do come up, then you have your list and you can just target the resources with that exact label to delete them.

alexh-tigera: It gets a bit complicated because some of the resources are optional. `appsv1.DeploymentList` might work, but my understanding is that all of these reads are coming from the operator's cache anyway, so it's not a huge performance hit to get the other items.

However, WDYT about also storing the namespaces of the gateways and only checking _those_ namespaces? We know they must be in the same ns as the Gateway.

pkg/controller/istio/waypoint/waypoint_cleanup.go
```diff
@@ -0,0 +1,193 @@
...
+		return managed, nil
+	}
+	var managed bool
+	gc := &gapi.GatewayClass{}
+	err := r.Get(ctx, types.NamespacedName{Name: className}, gc)
+	switch {
+	case err == nil:
+		managed = strings.HasPrefix(string(gc.Spec.ControllerName), istioGatewayControllerPrefix)
+	case errors.IsNotFound(err):
+		managed = builtinIstioGatewayClasses[className]
```
Brian-McM, line 175: This might not matter (since the gateway class is likely cached), but you could save a k8s call by checking if it's a default class first, then if it's not do the k8s call.

pkg/controller/istio/waypoint/waypoint_cleanup.go
```diff
@@ -0,0 +1,193 @@
...
+// strands the previous class's set — for waypoints, a running Deployment whose
+// l7-collector sidecar stays connected to Felix.
+func (r *ReconcileWaypoint) staleGatewaySets(ctx context.Context, reqLogger logr.Logger) ([]client.Object, error) {
+	gatewayList := &gapi.GatewayList{}
+	if err := r.List(ctx, gatewayList); err != nil {
+		return nil, fmt.Errorf("failed to list Gateways: %w", err)
+	}
+	gateways := map[types.NamespacedName]*gapi.Gateway{}
+	for i := range gatewayList.Items {
+		gw := &gatewayList.Items[i]
```
Brian-McM: Nit, but I don't think you need the index, just use the second parameter in the for loop.

Brian-McM (approved):
Couple nits but looks good.

PR #5068, opened 2026-07-20
EV-6825: istio: Create RoleBindings so the waypoint controller can copy pull secrets

alexh-tigera:
## Description

Bug fix / follow-up for the istio waypoint feature (Calico Enterprise only).

The waypoint controller copies Installation pull secrets into user namespaces that contain
istio-waypoint Gateways, but the operator has no cluster-wide permission to write Secrets —
writes are only authorized in namespaces where a RoleBinding binds the
`tigera-operator-secrets` ClusterRole to the operator's ServiceAccount.

This only happened to work because of the combined calico images, and calico-node is running as a DaemonSet. However, the operator was constantly trying to create the pull secret and failing, writing error logs regularly.

I've used the Gateway API code as an example of how to fix this.

- Create the `tigera-operator-secrets` RoleBinding in every namespace it writes pull secret
  copies into.
- Preserve copies and RoleBindings co-owned by other features (e.g. egress gateway): shared
  objects keep their owner references via the multiple-owners label, and stale objects are left for Kubernetes GC. This follows the same behaviour as the Gateway, and a general solution to clean up stale objects will be implemented later.
- Skip reserved namespaces (`tigera-operator`, `calico-system`, `tigera-gateway`) whose pull
  secrets and RoleBindings are managed by other controllers.

Testing: unit tests extended in `pkg/controller/istio/waypoint` covering RoleBinding
creation ordering, co-owned secret/binding preservation, reserved namespaces, and
watch-not-ready behavior; `go build`, `go vet`, and the istio/waypoint test suites pass.

## Release Note

```release-note
TBD
```

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/istio/waypoint/waypoint_controller.go
+++ b/pkg/controller/istio/waypoint/waypoint_controller.go
@@ -45,15 +45,6 @@ const (
 	// IstioWaypointClassName is the GatewayClass name used by Istio waypoints.
 	IstioWaypointClassName = "istio-waypoint"
 
-	// WaypointPullSecretLabel identifies the pull secret copies and tigera-operator-secrets
-	// RoleBindings created by this controller. Cleanup follows the egress gateway pattern:
-	// the objects carry an owner reference to the Istio CR and are removed by Kubernetes
-	// garbage collection when the CR is deleted. Objects stranded while the CR still exists
-	// (a namespace losing its last waypoint Gateway, or a pull secret being renamed in
-	// Installation) are not yet cleaned up — that gap is shared with the egress gateway and
-	// will be addressed for both together; the label marks the objects for that work.
-	WaypointPullSecretLabel = "operator.tigera.io/istio-waypoint-pull-secret"
-
 	// legacyGatewayNamespace is the pre-namespaced gateway-API install namespace, kept in
 	// sync with the const of the same name in the gatewayapi controller.
 	legacyGatewayNamespace = "tigera-gateway"
@@ -269,7 +260,6 @@ func (r *ReconcileWaypoint) pullSecretResources(ctx context.Context, reqLogger l
 	for ns := range targetNamespaces {
 		rb := render.CreateOperatorSecretsRoleBinding(ns)
 		rb.Labels = common.MapExistsOrInitialize(rb.Labels)
-		rb.Labels[WaypointPullSecretLabel] = "true"
 		rb.Labels[common.MultipleOwnersLabel] = "true"
 		objs = append(objs, rb)
 
@@ -278,7 +268,6 @@ func (r *ReconcileWaypoint) pullSecretResources(ctx context.Context, reqLogger l
 			if s.Labels == nil {
 				s.Labels = map[string]string{}
 			}
-			s.Labels[WaypointPullSecretLabel] = "true"
 			s.Labels[common.MultipleOwnersLabel] = "true"
 			objs = append(objs, s)
 		}
--- a/pkg/controller/istio/waypoint/waypoint_controller.go
+++ b/pkg/controller/istio/waypoint/waypoint_controller.go
@@ -19,11 +19,9 @@ import (
 	"fmt"
 
 	"github.com/go-logr/logr"
-	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/event"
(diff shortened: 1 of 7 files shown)


pkg/controller/istio/waypoint/waypoint_controller.go
```diff
@@ -205,58 +233,217 @@ func (r *ReconcileWaypoint) pullSecretChanges(ctx context.Context, active bool,
...
-			toDelete = append(toDelete, s)
+		if desiredSecrets[key] || hasForeignOwnerRef(s.OwnerReferences) {
+			continue
+		}
+		if terminating, err := nsTerminating(s.Namespace); err != nil {
+			return nil, nil, err
+		} else if terminating {
+			continue
 		}
+		toDelete = append(toDelete, &s)
```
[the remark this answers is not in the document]
alexh-tigera: This isn't true in Go 1.23+

pkg/controller/istio/waypoint/waypoint_controller.go
```diff
@@ -189,74 +221,68 @@ func (r *ReconcileWaypoint) Reconcile(ctx context.Context, request reconcile.Req
...
+	// Build the objects for each target namespace. The RoleBinding comes first: it grants
+	// the operator permission to write secrets in the namespace, so it must exist before
+	// the secret copies are created. Both carry MultipleOwnersLabel so the component
+	// handler merges owner references with any other owners (e.g. an egress gateway CR
+	// sharing the namespace) rather than replacing them.
+	var objs []client.Object
+	for ns := range targetNamespaces {
+		rb := render.CreateOperatorSecretsRoleBinding(ns)
+		rb.Labels = common.MapExistsOrInitialize(rb.Labels)
+		rb.Labels[WaypointPullSecretLabel] = "true"
```
pasanw: I think we should be able to remove this label and its assignments, give that we have removed cleanup?

pkg/controller/istio/waypoint/waypoint_controller.go
```diff
@@ -189,74 +221,68 @@ func (r *ReconcileWaypoint) Reconcile(ctx context.Context, request reconcile.Req
...
 	}
 
-	// List all existing secrets managed by this controller and mark stale ones for deletion.
-	existingSecrets := &corev1.SecretList{}
-	if err := r.List(ctx, existingSecrets, client.MatchingLabels{WaypointPullSecretLabel: "true"}); err != nil {
-		return nil, nil, fmt.Errorf("failed to list waypoint pull secrets: %w", err)
+	targetNamespaces := map[string]bool{}
+	for i := range gatewayList.Items {
+		gw := &gatewayList.Items[i]
+		if string(gw.Spec.GatewayClassName) == IstioWaypointClassName && !reservedNamespace(gw.Namespace) {
```
pasanw, line 258: Could we do without the reserved namespace check? Seems to me like the purpose of this logic is to ensure that we have the pull secrets in place that the proxy gateways need, therefore this logic should do its best to ensure that happens? Having no check seems like the conceptually right approach for me: this controller needs that secret, so it ensures its created (in case no one else does) and marks its need of this secret via ownership

pkg/controller/istio/waypoint/waypoint_controller.go
pasanw, line 1: General comment: lets make sure both the commit and PR desc have a clear call-out explaining why we removed cleanup of secrets

pkg/controller/istio/waypoint/waypoint_controller.go
```diff
@@ -156,31 +185,34 @@ func (r *ReconcileWaypoint) Reconcile(ctx context.Context, request reconcile.Req
...
-		toDelete = append(toDelete, stale...)
+	toDelete, err := r.staleGatewaySets(ctx, reqLogger)
+	if err != nil {
+		return reconcile.Result{}, err
 	}
 
-	// Use a single passthrough component to handle both creation and deletion.
-	hdlr := utils.NewComponentHandler(log, r, r.scheme, nil)
+	// Pass the Istio CR as the owner: created objects carry an owner reference to it
+	// (merged with any other owners, such as an egress gateway CR sharing the namespace,
```
pasanw: >merged with any other owners

I think this is only true if the label is set on the resource being created/updated, otherwise it will force sole ownership to Istio CR. Does this have any unintended consequences?

pkg/controller/istio/waypoint/waypoint_controller.go
```diff
@@ -156,31 +185,34 @@ func (r *ReconcileWaypoint) Reconcile(ctx context.Context, request reconcile.Req
...
+	// via the multiple-owners label) and are garbage collected when the CR is deleted —
```
pasanw:
```suggestion
	// via the multiple-owners label) and are garbage collected by owner reference —
```

pkg/controller/istio/waypoint/waypoint_controller.go
```diff
@@ -189,74 +221,68 @@ func (r *ReconcileWaypoint) Reconcile(ctx context.Context, request reconcile.Req
...
-			if string(gw.Spec.GatewayClassName) == IstioWaypointClassName &&
-				gw.Namespace != common.OperatorNamespace() {
-				targetNamespaces[gw.Namespace] = true
-			}
+// pullSecretResources returns the objects each waypoint namespace needs so its pods can
+// pull images from private registries: a tigera-operator-secrets RoleBinding granting the
+// operator permission to write secrets in the namespace, followed by copies of the
+// Installation pull secrets. Cleanup mirrors the egress gateway pattern: the objects are
+// owned by the Istio CR and garbage collected when it is deleted. Objects stranded while
+// the CR still exists (a removed Gateway, a renamed pull secret) are not yet cleaned up.
```
pasanw:
```suggestion
// Installation pull secrets. Cleanup mirrors the egress gateway pattern: the objects are
// garbage collected by owner reference. Objects stranded while the CR still exists and
// is the only owner (a removed Gateway, a renamed pull secret) are not yet cleaned up.
```

pkg/controller/istio/waypoint/waypoint_controller.go
pasanw: Only scenario that comes to mind: we might take ownership of something that has no owners, and then once we are gone, then it gets deleted. i.e. if others depend on it but don't mark that via ownerref, then we might remove it from under them

pkg/controller/istio/waypoint/waypoint_controller_test.go
pasanw, line 1: Would be nice to have tests that cover the "upgrade" or transition to the new version of the controller that doesn't cleanup resources (if the previous implementation made it to production)

pkg/controller/istio/waypoint/waypoint_controller.go
```diff
@@ -189,74 +221,68 @@ func (r *ReconcileWaypoint) Reconcile(ctx context.Context, request reconcile.Req
...
-		if !desiredSecrets[key] {
-			toDelete = append(toDelete, s)
+
+	// Build the objects for each target namespace. The RoleBinding comes first: it grants
+	// the operator permission to write secrets in the namespace, so it must exist before
+	// the secret copies are created. Both carry MultipleOwnersLabel so the component
+	// handler merges owner references with any other owners (e.g. an egress gateway CR
+	// sharing the namespace) rather than replacing them.
+	var objs []client.Object
+	for ns := range targetNamespaces {
```
Brian-McM, line 264: I'm wondering if this is starting to unravel the split between the render package and the controllers. We use the passthrough for simple things, and it's the exception to the rule but this is now getting to be more complicate resource logic and I wouldn't normally expect to find this here.

I'm OK to leave this here for the moment if you think it's best (also I don't want to turn this into a much larger refactor PR) but the exception to rule **_may_** be getting pushed in this controller.

Brian-McM (commented):
@alexh-tigera is there a compelling reason not to use a cluster role binding here? What this is doing is essentially a cluster role binding, the operator needs permission to write rolebindings in every namespace in the cluster to achieve this, so it essentially has the ability to create this secret in every namespace.

Anybody posing as the operator service account would have the ability to create the role binding in any namespace and then be able to create the secret (same as with a cluster role binding), but since these are user namespaces we're now littering them with extra role bindings.

I'm also not entirely sure about the cluster role that's being used, but this is a more fundamental issue. With that cluster role we can update or delete any secret in any customer namespace. I'd personally rather see dedicated cluster roles with resource names in them so we can only update the secrets we control. I don't think we have a case for secrets with dynamic names, certainly not in this case anyway.

pkg/controller/istio/waypoint/waypoint_controller.go
alexh-tigera: Yes, fixed in a735cdb2e

pkg/controller/istio/waypoint/waypoint_controller.go
alexh-tigera: > Only scenario that comes to mind: we might take ownership of something that has no owners, and then once we are gone, then it gets deleted. i.e. if others depend on it but don't mark that via ownerref, then we might remove it from under them

Right, this is the scenario that I was thinking of. Although that can only happen in `tigera-operator` and `tigera-gateway`, so I've removed `calico-system` from the list.

Deleting the secret in `tigera-operator` would be quite bad since that's the source of the pullsecret.

alexh-tigera:
> is there a compelling reason not to use a cluster role binding here? What this is doing is essentially a cluster role binding, the operator needs permission to write rolebindings in every namespace in the cluster to achieve this, so it essentially has the ability to create this secret in every namespace.

Yes, 100% agree. I was following the patterns that the GatewayAPI code used, but I think a follow up PR to manage user controlled namespaces specifically for `tigera-pull-secret` for both components is the way to go. I tried to find discussion of this in the design docs for Gateway but didn't see anything specific there.

pkg/controller/istio/waypoint/waypoint_controller.go
alexh-tigera: I've updated the PR - the 5897d0646 commit already notes that will be done in a follow up.

pkg/controller/istio/waypoint/waypoint_controller.go
alexh-tigera: Right. I've added some fixes in c057192bb to work around conflicts with the GatewayAPI, plus the reservedNamespaces should handle any cases that don't track the ownership.

pasanw (approved):

PR #5083, opened 2026-07-21
Add pre-push checklist to CLAUDE.md

vara2504:
## Summary
  - Add a Pre-push Checklist section to .claude/CLAUDE.md with the minimum commands to run before pushing: gen-files, fix, dirty-check, static-checks

## Release Note
None

### DIFF
--- a/.claude/CLAUDE.md
+++ b/.claude/CLAUDE.md
@@ -42,14 +42,15 @@ make cluster-destroy                   # Tear down
 ```
 
 ### Pre-push Checklist
-Run the relevant commands before pushing to avoid CI failures:
+Only run the commands matching what you changed — none are needed for a
+docs/YAML/comment-only push:
 ```bash
-# After API type changes (api/v1/, config/*_versions.yml):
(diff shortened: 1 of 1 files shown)


.claude/CLAUDE.md
```diff
@@ -41,6 +41,16 @@ go run ./ --enable-leader-election=false
 make cluster-destroy                   # Tear down
 ```
 
+### Pre-push Checklist
+After making changes, run these before pushing to avoid CI failures:
```
caseydavenport: I don't think we should be having Claude run these command unconditionally - many changes don't need these, and it's going to add extra time to each push. 

At a minimum we should be a bit smarter about explaining what sort of changes should trigger running which commands.

vara2504: @caseydavenport  updated so each command is gated by change type rather than run unconditionally:
  - static-checks (the slow golangci-lint container pass) now only runs after Go code changes, not on every push.
- dirty-check moved next to gen-files, where it just confirms regenerated files are committed.
- gen-files stays scoped to API/CRD/config changes.

PR #5087, opened 2026-07-22
fix(apiserver): prevent calico-apiserver deadlock on direct 3.21-to-3.23 upgrade

xiumozhan:
## Description

Bug fix.

On a direct Calico Enterprise 3.21.x → 3.23.x upgrade, the operator moves the `calico-apiserver` from the `tigera-system` namespace into `calico-system` and retargets the `v3.projectcalico.org` APIService before the apiserver's allow policy is effective. The leftover 3.21-era `allow-tigera.default-deny` (selector `all()`, `calico-system` namespace, `allow-tigera` tier) still selects the moved pod and drops its egress to the Kubernetes API, so the aggregated `projectcalico.org/v3` API goes down and the operator's write of `calico-system.apiserver-access` returns 503 indefinitely. The result is a deadlock that does not self-heal.

This PR makes the apiserver controller write a transitional policy "bridge" through the still-serving old apiserver, before the workload cutover, reproducing the intermediate state that the safe two-hop upgrade (via 3.22.x) passes through:

- rewrite `allow-tigera.default-deny` to exclude `calico-apiserver`,
- create an order-1 `allow-tigera.apiserver-access` in the `allow-tigera` tier,
- create `calico-system.apiserver-access`.

The bridge component is ordered before the workload component, so the controller's fail-fast apply loop requeues before any cutover if a bridge write fails. Detection uses a shared predicate — the deprecated `allow-tigera` tier still exists and `calico-apiserver` is not yet Ready in `calico-system` — gated on the tier watch and a no-op on the native-CRD (`UseV3CRDs`) path. While migrating, the existing `APIServerPolicy` component and the kube-controllers and tiers deletions of the transitional `allow-tigera.*` objects are suppressed; they resume once the apiserver is stable in `calico-system`, and a periodic reconcile on the tiers controller removes the emptied `allow-tigera` tier promptly.

**Components affected:** apiserver controller, installation/kube-controllers policy rendering, tiers controller, apiserver policy renderer.

**Testing:** unit tests (Ginkgo v2 + Gomega + controller-runtime fake client) covering the migration predicate, the bridge rendering, the controller wiring (bridge-before-cutover, failed-write requeue, and the `APIServerPolicy` gate), and the two cross-controller deletion gates. Verified locally: `pkg/render`, `pkg/controller/apiserver`, `pkg/controller/installation` (308/308 with envtest), `pkg/controller/tiers`, and `pkg/controller/utils` all pass; `go vet` clean.

No API/CRD changes. Intended to cherry-pick to `release-v1.42`.

Fixes EV-6821.

## Release Note

```release-note
Fixed a deadlock in which the Calico Enterprise API server (calico-apiserver) could become permanently unavailable during a direct upgrade from Calico Enterprise 3.21 to 3.23.
```

## For PR author

- [x] Tests for change.
- [ ] If changing pkg/apis/, run `make gen-files` — n/a, no API changes.
- [ ] If changing versions, run `make gen-versions` — n/a, no version changes.

## For PR reviewers

- [ ] Appropriate labels:
  - `kind/bug` — this is a bug fix.
  - `enterprise` — the affected path (aggregated API server, `allow-tigera` → `calico-system` migration) is Calico Enterprise; the change is a no-op on the OSS native-CRD path.
- [ ] Consider `docs-not-required` (restores expected upgrade behaviour; no user-facing configuration or API change).

### DIFF
--- a/pkg/controller/apiserver/apiserver_controller.go
+++ b/pkg/controller/apiserver/apiserver_controller.go
@@ -487,6 +487,19 @@ func (r *ReconcileAPIServer) Reconcile(ctx context.Context, request reconcile.Re
 		}
 	}
 
+	// Detect the allow-tigera -> calico-system apiserver migration window. Gate on the Tier
+	// watch being ready, matching the includeV3NetworkPolicy read above: before the watch is
+	// ready the aggregated v3 API may be unavailable, and we must not degrade/return on that.
+	// During a real migration the old apiserver is serving, so the watch is ready.
+	migrationInProgress := false
+	if !r.opts.UseV3CRDs && r.tierWatchReady.IsReady() {
+		migrationInProgress, err = utils.APIServerMigrationInProgress(ctx, r.client)
+		if err != nil {
+			r.status.SetDegraded(operatorv1.ResourceReadError, "Error checking apiserver migration state", err, reqLogger)
+			return reconcile.Result{}, err
+		}
+	}
+
 	err = utils.PopulateK8sServiceEndPoint(r.client)
 	if err != nil {
 		r.status.SetDegraded(operatorv1.ResourceReadError, "Error reading services endpoint configmap", err, reqLogger)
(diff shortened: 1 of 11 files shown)


xiumozhan:
Closing in favour of #5143, which takes the readiness-gate approach from #5139: it holds the cutover until the new API server reports ready, rather than writing a bridge policy before the move. That needs no knowledge of the deprecated policy and leaves the previous API server serving throughout, so there is no window where the aggregated API is unavailable.

PR #5090, opened 2026-07-22
Default generated CNI config to cniVersion 1.0.0; add spec.cni.specVersion

### LINKED ISSUE #4969: Calico fails on OpenShift 4.22 due to CNI version mismatch
OpenShift uses Multus CNI which has been bumped to use CNI 1.1.0 in [cluster-network-operator PR](https://github.com/openshift/cluster-network-operator/pull/2941). Multus has a strict version compatibility check in [generator.go](https://github.com/openshift/multus-cni/blob/b4ec7d8239ce4bd3ed949bce9816a013377b44c7/pkg/server/config/generator.go#L93) which requires the primary CNI delegate (Calico) to use at least CNI 0.4.0. The current version is hardcoded as 0.3.1 in [node.go](https://github.com/tigera/operator/blob/master/pkg/render/node.go#L833).

This results in the following error in kube-multus Pod:
```
[error] failed to create the configuration manager for the primary CNI plugin: 
failed to load the primary CNI configuration as a multus delegate with error 
'delegate cni version is 0.3.1 while top level cni version is 1.1.0'
```

sridhartigera:
## Description

Fixes #4969 (Calico fails on OpenShift 4.22+/4.23 due to CNI version mismatch); part of the fix for projectcalico/calico#12965.

OpenShift 4.23 bumps multus to a top-level `cniVersion: 1.1.0`, and multus rejects any primary-CNI delegate whose config declares a version below 0.4.0. The operator has hardcoded `0.3.1` in the generated CNI config since the early days, so multus crash-loops and nodes never go Ready.

Two commits:

1. **Render the generated CNI config with `cniVersion: 1.0.0` (was `0.3.1`), all platforms.**
   - Every supported container runtime accepts 1.0.0 configs (containerd ≥ 1.6 / CRI-O ≥ 1.24 — already below the supported minimum).
   - Calico and the bundled chained plugins (portmap, bandwidth, tuning) have supported spec 1.0.0 for years; the calico KDD test lane runs 1.0.0 permanently.
   - Existing attachments are unaffected on upgrade: libcni caches each attachment's config and replays the original `cniVersion` at DEL.

2. **Add `spec.cni.specVersion` to the Installation CR** (per review discussion with @fasaxc on Slack): enum `Auto` (default) | `"0.3.1"` | `"0.4.0"` | `"1.0.0"`. `Auto` lets the operator choose — currently 1.0.0 — and may be raised in future operator versions. Explicit values pin the version for environments where a chained CNI plugin or container runtime requires one (escape hatch back to `0.3.1` included). `1.1.0` is deliberately not offered yet: the bundled third-party CNI plugins are built against libcni v1.0.1 and top out at spec 1.0.0. Only valid with `spec.cni.type: Calico`.

### Testing

- Unit: render tests (default→1.0.0, pinned→0.3.1, Auto→1.0.0, Windows), validation (field rejected for non-Calico CNI), defaulting round-trip; CEL CRD-schema suite run against the regenerated CRD via envtest.
- **OpenShift 5.0.0-ec.3** (multus at 1.1.0, the breaking config): with `cniVersion: 0.3.1` multus crash-loops with `delegate cni version is 0.3.1 while top level cni version is 1.1.0`; patching the generated config to `1.0.0` (stock Calico images) immediately recovers multus, nodes go Ready, and the cluster install completes. Reproduced on two clusters.
- **GA OpenShift 4.18** (multus at 0.4.0): patching to `1.0.0` with stock images — no regression; pods and hostPort pods (portmap chain) network correctly, verified down to the programmed DNAT rules; clean DEL.

### Notes for reviewers

- After upgrade, `fillDefaults` materializes `specVersion: Auto` on existing Installations using Calico CNI (consistent with the operator's defaulting of other fields).
- Backport intent: this PR is master-only. For release branches the proposal is a minimal, no-API-change variant (default 1.0.0 only when `kubernetesProvider: OpenShift`), since z-streams shouldn't add CRD fields or change defaults for healthy non-OpenShift clusters.

## Release Note

```release-note
The operator-generated CNI config now declares cniVersion 1.0.0 (previously 0.3.1), required for multus compatibility on OpenShift 4.23+. Requires containerd >= 1.6 or CRI-O >= 1.24. A new Installation field, spec.cni.specVersion, allows pinning the version (including back to 0.3.1) or leaving it operator-managed (Auto).
```

## For PR author

- [x] Tests for change.


pkg/controller/installation/core_controller.go
```diff
@@ -558,6 +558,13 @@ func fillDefaults(instance *operatorv1.Installation, currentPools *v3.IPPoolList
 		}
 	}
 
+	// Default the CNI spec version of the generated CNI config. Only
+	// meaningful when the operator generates the config, i.e. Calico CNI.
```
tmjd: > Only meaningful when the operator generates the config

I don't understand what that means.

pkg/controller/installation/validation_test.go
tmjd, line 1: A validation test when the Plugin is not Calico would be good.

pkg/controller/installation/defaults_test.go
tmjd, line 1: It'd be good to have a test that validates the new field is defaulted to Auto with Calico CNI and another that doesn't have it set when the Plugin is not Calico.

tmjd (commented):
Do we need to have any kinds of guards around changing the CNI version?
Is it safe for folks upgrading, that the CNI version will switch from 0.3.1 to 1.0.0 will be seamless?

pkg/controller/installation/core_controller.go
sridhartigera: Fair - reworded to "Default the CNI spec version for Calico CNI." in ec80d750a. (What it was trying to say: the operator only writes a CNI config file when installing Calico CNI; for GKE/AmazonVPC/AzureVNET the platform installs the config, so the field doesn't apply.)

pkg/controller/installation/validation_test.go
sridhartigera: There was one folded into the combined test; split into two explicit cases in ec80d750a: "should allow specVersion with Calico CNI" and "should reject specVersion when the CNI plugin is not Calico".

pkg/controller/installation/defaults_test.go
sridhartigera: Added both in ec80d750a: "should default CNI spec version to Auto for Calico CNI" and "should not set CNI spec version for other CNI plugins" (mirroring the adjacent BGP defaulting tests).

sridhartigera:
> Do we need to have any kinds of guards around changing the CNI version? Is it safe for folks upgrading, that the CNI version will switch from 0.3.1 to 1.0.0 will be seamless?

I did test upgrade and the cluster upgrade was seamless.

caseydavenport (approved):

PR #5110, opened 2026-07-27
Apply calico-node before Typha during upgrades to prevent cluster-wide NotReady

AdheipSingh:
## Description

**Type of change:** bug fix

**Problem:** During upgrades, the operator applies the Typha Deployment and the calico-node DaemonSet in the same reconcile, so both roll simultaneously. Typha (typically 3 replicas) finishes its rollout in about a minute, while calico-node on a large cluster takes much longer. Felix is compatible with an older Typha, but an older Felix cannot sync with a newer Typha — so once all Typha replicas are on the new version, every still-old calico-node pod loses sync, fails its readiness probe after ~90s (3 × 30s), and goes NotReady. The DaemonSet controller does not count already-unavailable pods against `maxUnavailable`/`maxSurge`, so it then replaces pods across the whole cluster at once instead of honoring the configured rolling-update limits. On a ~200 node production cluster this presented as a cluster-wide wave of NotReady calico-node pods during a 3.31 → 3.32 upgrade.

**Fix:**
- Apply the calico-node component before the remaining components.
- After applying, read the calico-node DaemonSet directly from the API server rather than the cached client — immediately after the write, the informer cache can still hold the pre-update object whose status reports the previous rollout as complete, which would open the gate prematurely (this race was observed in testing: identical code passed or failed depending on whether the watch event beat the read).
- Gate the Typha component on the calico-node rollout being complete (`ObservedGeneration >= Generation`, all pods updated and ready) using the existing component `Ready()` mechanism.

With this ordering, new calico-node pods connect to old Typha (the compatible direction) while rolling under the configured limits, and Typha updates last, when every Felix is already on the new version. First installs are unaffected (no DaemonSet exists, so the gate is open), and in steady state the check passes immediately because the DaemonSet generation is unchanged.

**Testing:** Validated on a 30-node cluster upgrading Calico v3.31.3 → v3.32.1:
- With `nodeUpdateStrategy: {maxSurge: 4, maxUnavailable: 0}` (the affected user's configuration): calico-node Ready stayed 30/30 for the entire upgrade across every DaemonSet status transition; peak 34 concurrent pods (30 + exactly 4 surge), no limit bypass; node phase ~4 minutes; Typha updated only after all 30 nodes completed.
- With the default strategy (`maxUnavailable: 1`): Ready never dropped below 29/30 (the single pod being replaced), Typha gated for the full ~18 minute rollout and updated within seconds of the last node becoming ready.
- Without this change, the same upgrade dropped Ready to 11/30, with the DaemonSet controller replacing ~12 pods within 2 seconds.

**Affected components:** installation controller (core_controller.go), Typha render component.

## Release Note

```release-note
During upgrades, the operator now waits for the calico-node DaemonSet rollout to complete before updating Typha, preventing cluster-wide calico-node readiness disruption caused by older Felix being unable to sync with newer Typha.
```

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/installation/core_controller.go
+++ b/pkg/controller/installation/core_controller.go
@@ -1503,7 +1503,7 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 		NonClusterHost:         nonclusterhost,
 		FelixHealthPort:        *felixConfiguration.Spec.HealthPort,
 	}
-	components = append(components, render.Typha(&typhaCfg))
+	typhaComponent := render.Typha(&typhaCfg)
 
 	// See the section 'Use of Finalizers for graceful termination' at the top of this file for terminating details.
 	canRemoveCNI := false
@@ -1613,6 +1613,8 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 		}
 	}
 
+	components = append(components, typhaComponent)
+
 	// Build a configuration for rendering calico/node.
 	nodeCfg := render.NodeConfiguration{
 		GoldmaneRunning:               goldmaneRunning,
@@ -1671,7 +1673,7 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 		warnOnce.Reset()
 	}
 
-	components = append(components, render.Node(&nodeCfg))
+	nodeComponent := render.Node(&nodeCfg)
 
 	csiCfg := render.CSIConfiguration{
 		Installation: &instance.Spec,
@@ -1771,6 +1773,42 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 		r.status.SetDegraded(operatorv1.ResourceValidationError, "Error validating ImageSet", err, reqLogger)
 		return reconcile.Result{}, err
 	}
+
+	// Apply calico-node before the remaining components so that its rollout
+	// status can be checked before Typha is applied. Per the Calico version
+	// skew policy, Felix may be at most one minor version ahead of Typha, but
+	// not behind it. Updating Typha while older calico-node pods are still
+	// running would leave those pods unable to sync, marking them NotReady
+	// and allowing the DaemonSet controller to exceed the configured surge
+	// limits when replacing them.
+	if err = imageset.ResolveImages(imageSet, nodeComponent); err != nil {
+		r.status.SetDegraded(operatorv1.ResourceValidationError, "Error resolving ImageSet for calico-node", err, reqLogger)
+		return reconcile.Result{}, err
+	}
+	if err := handler.CreateOrUpdateOrDelete(ctx, nodeComponent, nil); err != nil {
+		r.status.SetDegraded(operatorv1.ResourceUpdateError, "Error creating / updating calico-node", err, reqLogger)
+		return reconcile.Result{}, err
+	}
+
+	// Check whether the calico-node rollout is complete; Typha is only
+	// applied once it is. Read the DaemonSet directly from the API server
+	// rather than the informer cache: immediately after the update above,
+	// the cache may still hold the pre-update object whose status reports
+	// fully rolled out, which would open the gate before the rollout has
+	// started.
+	nodeRolledOut := true
+	nodeDS, err := r.clientset.AppsV1().DaemonSets(common.CalicoNamespace).Get(ctx, common.NodeDaemonSetName, metav1.GetOptions{})
+	if err != nil {
+		if !apierrors.IsNotFound(err) {
+			r.status.SetDegraded(operatorv1.ResourceReadError, "Unable to read calico-node DaemonSet", err, reqLogger)
+			return reconcile.Result{}, err
+		}
+	} else {
+		nodeRolledOut = nodeDS.Status.ObservedGeneration >= nodeDS.Generation &&
+			nodeDS.Status.UpdatedNumberScheduled == nodeDS.Status.DesiredNumberScheduled &&
+			nodeDS.Status.NumberReady == nodeDS.Status.DesiredNumberScheduled
+	}
+	typhaCfg.NodeRolledOut = nodeRolledOut
 
 	if err = imageset.ResolveImages(imageSet, components...); err != nil {
 		r.status.SetDegraded(operatorv1.ResourceValidationError, "Error resolving ImageSet for components", err, reqLogger)
--- a/pkg/controller/installation/core_controller.go
+++ b/pkg/controller/installation/core_controller.go
@@ -25,6 +25,7 @@ import (
 	"reflect"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/elastic/cloud-on-k8s/v2/pkg/utils/stringsutil"
 	"github.com/sirupsen/logrus"
@@ -1492,6 +1493,73 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 		}
 	}
 
+	imageSet, err := imageset.GetImageSet(ctx, r.client, instance.Spec.Variant)
+	if err != nil {
+		r.status.SetDegraded(operatorv1.ResourceReadError, "Error getting ImageSet", err, reqLogger)
+		return reconcile.Result{}, err
+	}
+
+	if imageSet == nil {
+		// There is no imageSet for the configured variant, but check to see if there are any
+		// ImageSets with a different variant so we can give the user some kind of indication
+		// to why an existing ImageSet is being ignored.
+		nvis, err := imageset.DoesNonVariantImageSetExist(ctx, r.client, instance.Spec.Variant)
+		if err != nil {
+			r.status.SetDegraded(operatorv1.ResourceReadError, "Error checking for non-variant ImageSet", err, reqLogger)
+			return reconcile.Result{}, err
+		} else {
+			if nvis {
+				reqLogger.Info("An ImageSet exists for a different variant")
+			}
+		}
+	}
+
+	if err = imageset.ValidateImageSet(imageSet); err != nil {
+		r.status.SetDegraded(operatorv1.ResourceValidationError, "Error validating ImageSet", err, reqLogger)
+		return reconcile.Result{}, err
+	}
+
+	// Determine whether the Typha Deployment update must be deferred until the
+	// calico-node DaemonSet finishes rolling out. Per the Calico version skew
+	// policy, Felix may be newer than Typha, but not older: updating Typha
+	// while older calico-node pods are still running leaves those pods unable
+	// to sync, marking them NotReady, at which point the DaemonSet controller
+	// no longer honors the configured surge limits and replaces pods
+	// cluster-wide.
+	deferTyphaUpdate := false
+	desiredNodeImage, err := render.NodeImage(&instance.Spec, imageSet)
+	if err != nil {
+		r.status.SetDegraded(operatorv1.ResourceValidationError, "Error resolving calico-node image", err, reqLogger)
+		return reconcile.Result{}, err
+	}
+	nodeDS := &appsv1.DaemonSet{}
+	if err := r.client.Get(ctx, types.NamespacedName{Name: common.NodeDaemonSetName, Namespace: common.CalicoNamespace}, nodeDS); err != nil {
+		if !apierrors.IsNotFound(err) {
+			r.status.SetDegraded(operatorv1.ResourceReadError, "Unable to read calico-node DaemonSet", err, reqLogger)
+			return reconcile.Result{}, err
+		}
+		// DaemonSet doesn't exist (fresh install) - nothing to defer for.
+	} else {
+		deferTyphaUpdate = deferTyphaDeploymentUpdate(nodeDS, desiredNodeImage)
+	}
+	if deferTyphaUpdate {
+		// Only defer when a Typha Deployment already exists; if it is missing
+		// it must be created regardless, since calico-node depends on it.
+		typhaDeployment := &appsv1.Deployment{}
+		if err := r.client.Get(ctx, types.NamespacedName{Name: common.TyphaDeploymentName, Namespace: common.CalicoNamespace}, typhaDeployment); err != nil {
+			if !apierrors.IsNotFound(err) {
+				r.status.SetDegraded(operatorv1.ResourceReadError, "Unable to read Typha Deployment", err, reqLogger)
+				return reconcile.Result{}, err
+			}
+			deferTyphaUpdate = false
+		}
+	}
+	if deferTyphaUpdate {
+		reqLogger.Info("Deferring Typha Deployment update until the calico-node DaemonSet rollout completes",
+			"updatedNumberScheduled", nodeDS.Status.UpdatedNumberScheduled,
+			"desiredNumberScheduled", nodeDS.Status.DesiredNumberScheduled)
+	}
+
 	// Build a configuration for rendering calico/typha.
 	typhaCfg := render.TyphaConfiguration{
 		K8sServiceEp:           k8sapi.Endpoint,
@@ -1502,8 +1570,9 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 		ClusterDomain:          r.clusterDomain,
 		NonClusterHost:         nonclusterhost,
 		FelixHealthPort:        *felixConfiguration.Spec.HealthPort,
+		DeferDeploymentUpdate:  deferTyphaUpdate,
 	}
-	typhaComponent := render.Typha(&typhaCfg)
+	components = append(components, render.Typha(&typhaCfg))
 
 	// See the section 'Use of Finalizers for graceful termination' at the top of this file for terminating details.
 	canRemoveCNI := false
@@ -1613,8 +1682,6 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 		}
 	}
 
-	components = append(components, typhaComponent)
-
 	// Build a configuration for rendering calico/node.
 	nodeCfg := render.NodeConfiguration{
 		GoldmaneRunning:               goldmaneRunning,
@@ -1673,7 +1740,7 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 		warnOnce.Reset()
 	}
 
-	nodeComponent := render.Node(&nodeCfg)
+	components = append(components, render.Node(&nodeCfg))
 
 	csiCfg := render.CSIConfiguration{
 		Installation: &instance.Spec,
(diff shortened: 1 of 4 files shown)


pkg/controller/installation/core_controller.go
```diff
@@ -1772,6 +1774,42 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
...
+	if err = imageset.ResolveImages(imageSet, nodeComponent); err != nil {
+		r.status.SetDegraded(operatorv1.ResourceValidationError, "Error resolving ImageSet for calico-node", err, reqLogger)
+		return reconcile.Result{}, err
+	}
+	if err := handler.CreateOrUpdateOrDelete(ctx, nodeComponent, nil); err != nil {
+		r.status.SetDegraded(operatorv1.ResourceUpdateError, "Error creating / updating calico-node", err, reqLogger)
+		return reconcile.Result{}, err
+	}
+
+	// Check whether the calico-node rollout is complete; Typha is only
```
caseydavenport: Is this the right behavior for a fresh rollout? 

On a brand new cluster, we don't want to wait for calico/node to roll out before deplying Typha, since typha is needed for calico/node to become ready. 

I suspect this will cause a deadlock on a fresh cluster.

pkg/controller/installation/core_controller.go
```diff
@@ -1772,6 +1774,42 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
...
+	// applied once it is. Read the DaemonSet directly from the API server
+	// rather than the informer cache: immediately after the update above,
```
caseydavenport: I don't think we should be doing a direct read against the API server here for perf reasons - we can use the cache if we get the existing node DS generation prior to applying the new version, I think.

pkg/controller/installation/core_controller.go
```diff
@@ -1772,6 +1774,42 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
...
+	nodeDS, err := r.clientset.AppsV1().DaemonSets(common.CalicoNamespace).Get(ctx, common.NodeDaemonSetName, metav1.GetOptions{})
+	if err != nil {
+		if !apierrors.IsNotFound(err) {
+			r.status.SetDegraded(operatorv1.ResourceReadError, "Unable to read calico-node DaemonSet", err, reqLogger)
+			return reconcile.Result{}, err
+		}
+	} else {
+		nodeRolledOut = nodeDS.Status.ObservedGeneration >= nodeDS.Generation &&
+			nodeDS.Status.UpdatedNumberScheduled == nodeDS.Status.DesiredNumberScheduled &&
+			nodeDS.Status.NumberReady == nodeDS.Status.DesiredNumberScheduled
```
caseydavenport: `NumberReady` is a steady state health check rather than a rollout check. One permanently unready calico-node pod (cordoned node, wedged kubelet, unrelated crashloop) leaves this false forever, so Typha stops getting config updates entirely, not just during upgrades.

What we actually care about is that no old-version Felix is left, which is the `UpdatedNumberScheduled` term on its own. We should drop the `NumberReady` check.

pkg/controller/installation/core_controller.go
```diff
@@ -1772,6 +1774,42 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
...
+	}
```
caseydavenport: Nothing reliably wakes us back up when the rollout finishes. The DaemonSet watch from `secondaryResources()` goes through `createPredicateForObject`, whose `UpdateFunc` filters on generation change, and rollout progress is status only. So the gate stays shut until the 5 minute periodic reconcile fires.

I suspect the testing here looked fast because other watches (FelixConfiguration, secrets) are noisy enough to mask it. We should return a `RequeueAfter` while gated, and log that we're deliberately holding Typha back.

pkg/controller/installation/core_controller.go
```diff
@@ -1613,6 +1613,8 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 		}
 	}
 
+	components = append(components, typhaComponent)
```
caseydavenport: Moving the append down here also reorders Typha relative to the cleanup passthrough above. Probably harmless, but it doesn't look intentional - can we leave the append where it was?

pkg/render/typha.go
```diff
@@ -165,7 +171,7 @@ func (c *typhaComponent) typhaPodDisruptionBudget() *policyv1.PodDisruptionBudge
 }
 
 func (c *typhaComponent) Ready() bool {
-	return true
+	return c.cfg.NodeRolledOut
```
caseydavenport: Worth keeping in mind that `CreateOrUpdateOrDelete` skips the whole component when `Ready()` is false, not just the Deployment. So this also stops reconciling the typha Service, ServiceAccount, RBAC, PDB, cert secret and NetworkPolicy.

That's my main concern with gating rather than just ordering the two applies: any state where calico-node is unready because something on the Typha side is missing or broken now becomes unrecoverable, since the operator won't touch Typha until node is healthy.

pkg/controller/installation/core_controller.go
```diff
@@ -1772,6 +1774,42 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 		return reconcile.Result{}, err
 	}
 
+	// Apply calico-node before the remaining components so that its rollout
```
caseydavenport: This ordering is inverted for downgrades, where node rolls to the older Felix first while Typha is still new, which is exactly the direction the comment says doesn't work.

Wonder if we need more awareness on if this is an upgrade / downgrade / or simply a configuration change.

pkg/controller/installation/core_controller.go
```diff
@@ -1772,6 +1774,42 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
...
+		if !apierrors.IsNotFound(err) {
+			r.status.SetDegraded(operatorv1.ResourceReadError, "Unable to read calico-node DaemonSet", err, reqLogger)
+			return reconcile.Result{}, err
+		}
+	} else {
+		nodeRolledOut = nodeDS.Status.ObservedGeneration >= nodeDS.Generation &&
+			nodeDS.Status.UpdatedNumberScheduled == nodeDS.Status.DesiredNumberScheduled &&
+			nodeDS.Status.NumberReady == nodeDS.Status.DesiredNumberScheduled
+	}
+	typhaCfg.NodeRolledOut = nodeRolledOut
```
caseydavenport: This mutates `typhaCfg` a few hundred lines after we handed it to `render.Typha()` - I think we need to adjust this so that we are setting this field before passing the config to the renderer.

caseydavenport:
@AdheipSingh thanks for the PR - was there a specific issue you noticed that prompted this? 

We try to make the upgrade not depend on ordering regardless, so curious if we introduced an unexpected dependency in v3.32.

caseydavenport (changes requested):

AdheipSingh:
> **Problem:** During upgrades, the operator applies the Typha Deployment and the calico-node DaemonSet in the same reconcile, so both roll simultaneously. Typha (typically 3 replicas) finishes its rollout in about a minute, while calico-node on a large cluster takes much longer. Felix is compatible with an older Typha, but an older Felix cannot sync with a newer Typha — so once all Typha replicas are on the new version, every still-old calico-node pod loses sync, fails its readiness probe after ~90s (3 × 30s), and goes NotReady. The DaemonSet controller does not count already-unavailable pods against `maxUnavailable`/`maxSurge`, so it then replaces pods across the whole cluster at once instead of honoring the configured rolling-update limits. On a ~200 node production cluster this presented as a cluster-wide wave of NotReady calico-node pods during a 3.31 → 3.32 upgrade.

Hi @caseydavenport , thanks for the review this is main reason ˆˆ for this PR. We have large nodes ( 1000 + ), and when we increase maxSurge to speed up rollouts, keeping maxUnavailable as 0, we see a large disruption. This was the reason for fixing this. We saw no issue when maxSurge is set to 1, as we increase surge we see this issue. Also this is visible when we tried to upgrade 30 nodes with maxSurge 4, maxSurge 1/2 works fine on 30 nodes.....so the disruption is proportional to  no of nodes and maxSurge count. 


A newer Felix can talk to an older Typha just fine. But an older Felix cannot talk to a newer Typha.... the connection just dies. So during any upgrade there's exactly one safe order: upgrade all the calico-nodes first, Typha last. Then every mixed pairing that ever exists during the transition is "newer Felix --> older Typha," the direction that works.

The operator ignored that order. When you upgrade Calico, the tigera-operator updates Typha and calico-node in the same breath. Typha is only 3 pods it finishes upgrading in about a minute. calico-node on 200 nodes, at 4 at a time, takes half an hour. So one minute into the upgrade you have: 3 brand-new Typhas, and ~196 still-old calico-nodes — every one of them now in the forbidden pairing. They all lose their instruction feed simultaneously, all fail their health checks 90 seconds later, and all flip to NotReady at once.

pkg/controller/installation/core_controller.go
AdheipSingh: You were right , the previous revision deadlocked fresh installs (the gate waited for node readiness, and node needs Typha to become ready). Fixed in the latest revision in two ways: readiness is no longer part of the condition, and only the Typha Deployment *update* is deferred on a fresh install (no existing Deployment) everything is created immediately. Verified on a cluster: Typha was created on the first reconcile, all nodes came up, and the deferral stops once the DaemonSet is fully scheduled.

pkg/controller/installation/core_controller.go
AdheipSingh: Removed the direct read. The check now runs entirely from the informer cache.

pkg/controller/installation/core_controller.go
AdheipSingh: Agreed, dropped. The condition is now only `ObservedGeneration >= Generation && UpdatedNumberScheduled == DesiredNumberScheduled`. Added a unit test pinning the "all pods updated but some unready must not block Typha" case so it can't creep back in.

pkg/controller/installation/core_controller.go
AdheipSingh: The latest revision returns `utils.StandardRetry` while the update is deferred and logs "Deferring Typha Deployment update until the calico-node DaemonSet rollout completes" with the updated/desired counts on each reconcile.

pkg/controller/installation/core_controller.go
AdheipSingh: Restored.

pkg/render/typha.go
AdheipSingh: agreed. The latest revision removes the `Ready()`gating entirely.

pkg/controller/installation/core_controller.go
AdheipSingh: Correct — as written, the deferral applies to any image change, so a downgrade would also roll node first, which is the direction the skew policy doesn't support. Might consider detecting direction by comparing the resolved desired version against the deployed one and only deferring for upgrades (falling back to current behavior for downgrades and unparseable references like digests), but didn't want to add version-parsing machinery without your input. Would you prefer that in this PR, as a follow-up, or is there an existing pattern in the operator for knowing upgrade vs downgrade that we should use?

pkg/controller/installation/core_controller.go
AdheipSingh: Fixed, the rollout check now runs before the Typha configuration is constructed, and the field is set at construction time. No mutation after `render.Typha()`

AdheipSingh:
Unit test is generated via Claude.

caseydavenport:
>  But an older Felix cannot talk to a newer Typha.... the connection just dies. 

This is the part that I'm worried about - do you have log output by chance from when this happened? We try to avoid this!

AdheipSingh:
> > But an older Felix cannot talk to a newer Typha.... the connection just dies.
> 
> This is the part that I'm worried about - do you have log output by chance from when this happened? We try to avoid this!

Let me share with you RCA document.

AdheipSingh:
## RCA: cluster-wide calico-node NotReady during operator upgrade to Calico v3.32

All evidence below was captured on a 30-node kubeadm cluster (Kubernetes v1.33, AWS EC2), upgrading Calico v3.31.3 → v3.32.0 with the stock, unpatched operator v1.42.0 and `nodeUpdateStrategy: {maxSurge: 4, maxUnavailable: 0}`.

---

## Summary

Typha v3.32.0 **actively disconnects** any sync client that does not advertise `SupportsModernPolicyKeys` in its hello message.  This looks like an explicit server-side disconnect, not a transient error.

Because the operator applies the Typha Deployment and the calico-node DaemonSet in the same reconcile, Typha (3 replicas) finishes rolling ~30× faster than the DaemonSet. Every not-yet-upgraded calico-node then loses Typha permanently, fails its readiness probe after 90s (3 × 30s), and the DaemonSet controller which does not count already-unavailable pods against the rolling-update limits, replaces the entire fleet at once, ignoring `maxSurge`/`maxUnavailable`.

Observed impact: **Ready collapsed from 29/30 to 7/30 in 33 seconds** on a cluster configured for `maxUnavailable: 0`.

---

## Root cause: the exact point where the connection breaks

Reproduced in isolation (operator scaled to 0; calico-node container image set to `v3.31.3` against healthy, Ready Typha `v3.32.0` no other variables).

**Typha v3.32.0 server log**  receives the v3.31.3 hello and disconnects it, for every syncer type (`felix`, `bgp`, `node-status`):

```
typha/sync_server.go 940: Received Hello message from client.
    msg=syncproto.MsgClientHello{Hostname:"ip-172-31-26-176", Info:"node-status v3.31.3",
    Version:"v3.31.3", SyncerType:"node-status", SupportsDecoderRestart:true,
    SupportedCompressionAlgorithms:[]syncproto.CompressionAlgorithm{"snappy"},
    SupportsModernPolicyKeys:false, ClientConnID:0x165ddefa00000285}
typha/sync_server.go 973: Client does not support modern policy keys, disconnecting.
    client=172.31.26.176:36936 connID=0x9e233bdf000054ce type="node-status"
```

Source: `projectcalico/calico` v3.32.0, `typha/pkg/syncserver/sync_server.go:973`.

**Felix v3.31.3 client log** — the same moment from the client side; the disconnect surfaces as EOF, and Felix loops on reconnect indefinitely:

```
status-reporter/sync_client.go 435: Failed to read from server
    connection=172.31.16.97:5473/ip-172-31-16-97 error=EOF myID=0xfec4f69700000001 type="node-status"
confd/sync_client.go 435: Failed to read from server
    connection=172.31.16.97:5473/ip-172-31-16-97 error=EOF myID=0xc983b23200000001 type="bgp"
sync_client.go 220: Typha connection failed but client callback is restart-aware. Restarting connection...
```

**Health consequence** — readiness probe on the affected pod:

```
$ kubectl exec <calico-node> -- /bin/calico-node -felix-ready
calico/node is not ready: felix is not ready: readiness probe reporting 503
command terminated with exit code 1
```

State during this isolation test: calico-node image `quay.io/calico/node:v3.31.3`, Typha image `quay.io/calico/typha:v3.32.0` (3/3 Ready), affected pods `0/1 Running` indefinitely. The direction matters: v3.32 Felix against v3.31 Typha connects and becomes Ready (verified — this is what makes node-first ordering safe).

---

## The cascade: timeline of the operator-driven upgrade (unpatched v1.42.0)

DaemonSet watch excerpt (columns: DESIRED CURRENT READY UP-TO-DATE AVAILABLE):

```
01:58:54.236Z calico-node 30 30 29  4 29   ← rolling correctly, 4 at a time
01:59:21.060Z calico-node 30 30 27  4 27   ← readiness fuse expires
01:59:22.446Z calico-node 30 30 22  6 22
01:59:23.664Z calico-node 30 30 17 10 17
01:59:52.724Z calico-node 30 30  9 19  9
01:59:53.562Z calico-node 30 30  8 24  8   ← 20 pods replaced in 32s, surge limit bypassed
```

**7/30 Ready**. The `maxSurge: 4 / maxUnavailable: 0` strategy was configured and verified on the DaemonSet; the DaemonSet controller bypassed it ecause self-NotReady pods are not counted against availability limits.

---

The compatible pairing is strictly one-directional: newer Felix ↔ older Typha works; older Felix ↔ newer Typha is refused at `sync_server.go:973`. The PR therefore defers only the Typha Deployment update until the calico-node DaemonSet reports `UpdatedNumberScheduled == DesiredNumberScheduled`, so no older Felix remains when Typha restarts. Same cluster, same configuration, with the patched operator: Ready stayed 30/30 through every DaemonSet status
transition in repeated runs, Typha updated within seconds of the last node completing, and surge limits were honored throughout (peak 34 pods = 30 + 4).

caseydavenport:
@AdheipSingh so, I've been talking a bit more about this internally and have a few updates here.

The way this is designed to work:

- Typha gets a MaxSurge of 100%, allowing new replicas to start immediately.
- Old Typhas have a graceful termination window of 300s, during which they slowly shed connections.
- A felix that was shed will attempt to reconnect to a new Typha without a restart
  - If it succeeds, all is good, and it will wait for the DS controller to restart it.
  - If it encounters an error (like you are seeing) it will go NotReady and upgrade ahead of schedule in order to fix itself.

In normal operation, this is fine - the real "bug" here is that we intentionally made a breaking change between v3.31 and v3.32, which means that all downlevel calico/nodes hit that second branch and go Unready, get upgraded, and then take 30-60 seconds for our slow readiness probe to detect them as ready again. 

Given that the breaking change was done intentionally (forced by a policy improvement), I don't think we can adjust that.

I also think that attempting to force an ordering dependency here is also a bit problematic for a few reasons:
- The upgrade vs. downgrade considerations we talked about earlier. 
- The fact that, in practice, this same breakage could have happened the other direction (we support +/- 1 version, so either direction could happen).
- A hung calicoi/node could cause Typha to stall forever, even when an upgraded Typha could be the solution to the hung calico/node!

not technically impossible, but certainly complicated for what is a one-time blip from v3.31 -> v3.32.

What I think we can do:
- I've already sped up our probes in master in this PR: https://github.com/tigera/operator/pull/4609
- I've also raised a new PR to add a startup probe which should detect it even faster: https://github.com/tigera/operator/pull/5163

This should both detect that calico/node is `NotReady` faster, and also move new pods into `Ready` state faster as well, closing the gap between the Typha shedding and the DS controller's rates.

AdheipSingh:
Hi @caseydavenport ,

Thanks for the internal follow-up and the detailed design context the shed, reconnect, self-heal flow makes sense, and understood that the v3.31→v3.32 protocol change was intentional and can't be adjusted.

One data point from our reproductions that I think is worth weighing against the probe-speed mitigations is that the impact at scale isn't primarily detection latency, it's simultaneity. When the downlevel calico/nodes hit the error branch, they go NotReady together, and the DaemonSet controller then no longer counts them against the rolling-update limits  in our captures (30 nodes, maxSurge: 4, maxUnavailable: 0), UP-TO-DATE jumped 4→24 within ~2 seconds and Ready collapsed 29→7/30 in ~30s. Faster readiness/startup probes shrink the window per pod, but the entire fleet's dataplane still restarts at effectively the same moment. On the ~200-node production cluster where we first hit this, that remains a meaningful event even at reduced duration.

Given the objections to a default ordering dependency (all of which I think are fair particularly the hung-node-stalls-Typha case), would you be open to an opt-in variant instead? E.g. a field on the Installation (default off) that defers only the Typha Deployment update until the node DaemonSet reports all pods updated. Opt-in addresses the three concerns directly: no default behavior change, direction is explicit rather than inferred, and operators who worry about the hung-node interaction simply don't enable it. If that's still more machinery than the one-time transition justifies, happy to close this PR in favor of the probe improvements no hard feelings, the review feedback was valuable either way.

Two follow-ups regardless of the outcome:

1. Will the probe changes ) be backported to release-v1.42? That's the operator line v3.32.x users are on, so the mitigation only helps this transition if it ships there.
2. For large clusters making the v3.31....v3.32 hop today, we've validated a workaround using only supported mechanisms: annotate the Typha Deployment with unsupported.operator.tigera.io/ignore before applying the new operator manifest, wait for the node DaemonSet to reach UpdatedNumberScheduled == Desired, then remove the annotation (the operator immediately reconciles Typha forward). Tested end-to-end on the versions in question node rollout completes against old Typha with zero Ready dips, Typha rolls in seconds afterwards. It seems better for users hitting this to find a documented procedure than to discover the failure mode in production, as we did.

caseydavenport:
@AdheipSingh I suppose the question I'd ask is if you noticed any actual disruption to your cluster as part of this, or was it primarily the optics of many pods getting updated at once? Is there a tangible impact that this had on your apps?

I am reluctant to go ahead with the ordered rollout, even behind a flag, as I think it could be a pretty large "foot-gun" and the complexity isn't quite justified in my mind. 

I'd be happy to backport the timing fixes to older branches.

PR #5114, opened 2026-07-27
Gate the RBAC management UI on a per-cluster ConfigMap

dimitri-nicolo:
## Description

Replaces the `Manager.spec.rbacUI.state` gate for the RBAC management UI with a per-cluster `rbac-ui-config` ConfigMap in `calico-system`, keyed `rbac-ui-enabled`. The admin owns the ConfigMap outright; the operator, ui-apis and rbacsync only read it, so a toggle needs no restart or rollout.

The consuming side is on calico-private branch https://github.com/tigera/calico-private/pull/12954 — that PR will be merged alongside this one.

### Changes

- **API** — `Manager.spec.rbacUI` removed, CRD regenerated, along with the Manager CR reads and watches in the installation and apiserver controllers.
- **Read** — the installation, apiserver and manager controllers each read and watch the ConfigMap. A missing ConfigMap, missing key or unparsable value reads as disabled. The operator never writes it.
- **Render** — the access the feature depends on is gated on the value: the `tigera-network-admin` rolebinding rules, the `calico-manager` cluster rules, the namespaced Role scoping ui-apis to the IdP resources, the LDAP/AD egress, and the rbacsync controller plus its Role. Multi-tenant renders none of it.
- **Enablement** — `tigera-network-admin` gets write access to `rbac-ui-config`.

### Notes

- **Fail-closed** — a missing ConfigMap reads as disabled, and the operator does not re-create it, so deleting it switches the feature off and keeps it off. An unreadable ConfigMap degrades the reconcile rather than rendering as disabled.
- **`create` on ConfigMaps is namespace-wide** — RBAC cannot restrict `create` by resource name, so the write grant admits creating any ConfigMap in the namespace. A namespaced Role would narrow it but `tigera-network-admin` is a ClusterRole the customer binds themselves.

## Release Note

```release-note
NONE
```

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


pkg/render/kubecontrollers/kube-controllers.go
```diff
@@ -707,6 +704,14 @@ func (c *kubeControllersComponent) rbacSyncIDPGroupsRole() []client.Object {
 					ResourceNames: []string{"tigera-idp-groups"},
```
pasanw: Should this permission be gated on whether or not the config map is present and has the feature enabled?

pkg/controller/installation/core_controller.go
```diff
@@ -2378,6 +2376,32 @@ func (r *ReconcileInstallation) checkActive(log logr.Logger) (*corev1.ConfigMap,
...
 }
 
+// seedRBACManagementConfigMap creates the rbac-ui-config ConfigMap, disabled, if it is
+// not already present. This controller owns it because every cluster needs one in order
+// to control its own access to the RBAC management UI.
+//
+// Create-only and ownerless, so an admin's toggle survives both reconciles and
+// Installation deletion. Deleting the ConfigMap is fail-closed: consumers read a missing
+// one as disabled and this re-seeds it disabled, since the previous value is not tracked.
+func (r *ReconcileInstallation) seedRBACManagementConfigMap(ctx context.Context, log logr.Logger) error {
```
pasanw: I would lean towards the operator having a read-only relationship with this config map, leaving the user fully in control of whether this config map exists. Any components that need to read from the config map can assume the feature is disabled entirely if the configmap doesn't exist

pkg/render/kubecontrollers/kube-controllers.go
pasanw, line 1: Claude says that the RBACManagementEnabled check provided protection against the controller and role being enabled in multi-tenant environments. If so we'd need a tenancy check instead?

pkg/render/kubecontrollers/kube-controllers.go
```diff
@@ -220,12 +216,11 @@ func NewCalicoKubeControllers(cfg *KubeControllersConfiguration) *kubeController
...
-		// bindings against the tigera-idp-groups ConfigMap.
-		if cfg.RBACManagementEnabled {
-			enabledControllers = append(enabledControllers, "rbacsync")
-			kubeControllerRolePolicyRules = append(kubeControllerRolePolicyRules, rbacSyncControllerRules()...)
-		}
+		// Runs the rbacsync controller to reconcile managed ClusterRoles and bindings
+		// against the tigera-idp-groups ConfigMap. It reads this cluster's
+		// rbac-ui-config and reconciles nothing while the feature is off there.
+		enabledControllers = append(enabledControllers, "rbacsync")
+		kubeControllerRolePolicyRules = append(kubeControllerRolePolicyRules, rbacSyncControllerRules()...)
```
pasanw: Should this permission be gated on whether or not the config map is present and has the feature enabled?

pkg/render/manager.go
```diff
@@ -284,11 +292,11 @@ func (c *managerComponent) Objects() ([]client.Object, []client.Object) {
 
 	objsToCreate = append(objsToCreate,
 		managerClusterRoleBinding(c.cfg.Tenant, c.cfg.BindingNamespaces, c.cfg.OSSTenantNamespaces),
-		managerClusterRole(false, c.cfg.Installation.KubernetesProvider, c.cfg.Tenant, c.cfg.Manager.RBACManagementEnabled()),
+		managerClusterRole(false, c.cfg.Installation.KubernetesProvider, c.cfg.Tenant),
 		c.managedClustersWatchRoleBinding(),
 	)
 	objsToCreate = append(objsToCreate, c.managedClustersUpdateRBAC()...)
-	if c.cfg.Manager.RBACManagementEnabled() && !c.cfg.Tenant.MultiTenant() {
+	if !c.cfg.Tenant.MultiTenant() {
```
pasanw: Should this permission be gated on whether or not the config map is present and has the feature enabled?

pkg/render/apiserver.go
```diff
@@ -2145,20 +2141,18 @@ func (c *apiServerComponent) tigeraNetworkAdminClusterRole() *rbacv1.ClusterRole
...
-				Verbs:     []string{"get", "list", "watch"},
-			},
-			rbacv1.PolicyRule{
-				APIGroups: []string{"rbac.authorization.k8s.io"},
-				Resources: []string{"clusterrolebindings", "rolebindings"},
-				Verbs:     []string{"get", "list", "watch", "create", "update", "delete"},
-			},
-		)
-	}
+	rules = append(rules,
```
pasanw, line 2200: Should this permission be gated on whether or not the config map is present and has the feature enabled?

pasanw (commented):
The TL:DR of most of my comments: we've moved the enable knob outside of the operator API, but the operator should still consume it

pkg/render/kubecontrollers/kube-controllers.go
dimitri-nicolo: Added in "Gate the rbacsync controller on the feature gate, and skip multi-tenant".

`rbacSyncEnabled(cfg)` requires Enterprise, `RBACManagementEnabled`, and `!Tenant.MultiTenant()`. It gates this Role and the controller's entry in `ENABLED_CONTROLLERS`.

pkg/controller/installation/core_controller.go
dimitri-nicolo: Added in "Leave the RBAC management UI gate entirely to the admin".

The seeding is dropped: the installation controller no longer creates the ConfigMap, and `render.RBACManagementConfigMap` is removed. The operator's only contact with the gate is the three controllers reading it. A missing ConfigMap already read as disabled, so the default is unchanged; deleting it now stays deleted instead of reappearing on the next reconcile.

pkg/render/kubecontrollers/kube-controllers.go
dimitri-nicolo: Added in "Gate the rbacsync controller on the feature gate, and skip multi-tenant".

The old protection was incidental: the installation controller looked the Manager CR up by a cluster-scoped key, which never resolves on a multi-tenant management cluster, so the feature always read as disabled there. Reading a ConfigMap removed that side effect, so `rbacSyncEnabled` now checks `!cfg.Tenant.MultiTenant()` explicitly.

pkg/render/kubecontrollers/kube-controllers.go
dimitri-nicolo: Done, same `rbacSyncEnabled` gate as the Role above.

pkg/render/manager.go
dimitri-nicolo: Done in "Gate the manager RBAC management UI access on the feature gate".

`rbacManagementUIActive()` (`RBACManagementEnabled && !Tenant.MultiTenant()`) gates the `calico-manager` cluster rules, the namespaced Role scoping ui-apis to the IdP resources, and the LDAP/AD egress policy. ui-apis keeps read access to the gate ConfigMap whenever the Role is rendered, so it can observe the feature being switched off.

pkg/render/apiserver.go
dimitri-nicolo: Done in "Gate the tigera-network-admin RBAC management rules on the feature gate".

The `clusterrolebindings`/`rolebindings` create/update/delete rule and the `clusterroles`/`roles` read are now behind the gate.

pkg/controller/installation/core_controller.go
```diff
@@ -1355,6 +1345,20 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 		return reconcile.Result{}, err
 	}
 
+	// The admin owns this ConfigMap; the operator only reads it, and an absent one reads
+	// as disabled.
+	var rbacManagementEnabled bool
+	if instance.Spec.Variant.IsEnterprise() {
```
pasanw: Should there be a multi-tenancy check here? Wondering if the setting of rbacManagementEnabled should be gated on the three conditions (enterprise, multi-tenancy, config map), and then the component itself just uses this boolean rather than performing the same checks again

pkg/render/manager.go
```diff
@@ -1235,6 +1240,22 @@ func managerClusterRole(managedCluster bool, kubernetesProvider operatorv1.Provi
 	return cr
 }
 
+// RBACManagementEnabled reports whether the RBAC management UI is switched on for this
+// cluster. A missing ConfigMap, missing key or unparsable value reads as disabled.
+func RBACManagementEnabled(cm *corev1.ConfigMap) bool {
```
pasanw: Consider migrating this to render/common

pkg/controller/manager/manager_controller.go
```diff
@@ -735,6 +750,7 @@ func (r *ReconcileManager) Reconcile(ctx context.Context, request reconcile.Requ
 		Manager:                    instance,
 		Authentication:             authenticationCR,
 		KibanaEnabled:              kibanaEnabled,
+		RBACManagementEnabled:      render.RBACManagementEnabled(rbacGate),
```
pasanw: Similar comment as the one I left on the core controller: consider setting this based on the full check with tenancy so we don't need to repeat the same checks inside the render component

pkg/controller/apiserver/apiserver_controller.go
```diff
@@ -538,7 +541,7 @@ func (r *ReconcileAPIServer) Reconcile(ctx context.Context, request reconcile.Re
 		ClusterDomain:                r.opts.ClusterDomain,
 		Cloud:                        r.opts.Cloud,
 		RequiresAggregationServer:    !r.opts.UseV3CRDs,
-		RBACManagementEnabled:        managerCR.RBACManagementEnabled(),
+		RBACManagementEnabled:        rbacManagementEnabled,
```
pasanw, line 540: Same thought here as the one I left in the core controller

pkg/render/manager.go
```diff
@@ -83,6 +83,11 @@ const (
 	// Keep in sync with ui-apis rbacmanagement/idp LDAPConfigSecretName.
 	RBACManagementLDAPConfigSecretName = "tigera-idp-ldap-config"
 
+	// The admin-owned switch for the RBAC management UI, read by the operator, ui-apis
+	// and rbacsync. Keep in sync with ui-apis rbacmanagement/gate.
+	RBACManagementConfigMapName = "rbac-ui-config"
```
pasanw: Consider moving these to render/common

pkg/controller/installation/core_controller.go
```diff
@@ -1324,18 +1325,10 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
...
-	if instance.Spec.Variant.IsEnterprise() {
-		gate, err := utils.GetIfExists[corev1.ConfigMap](ctx, client.ObjectKey{
-			Name: render.RBACManagementConfigMapName, Namespace: common.CalicoNamespace,
-		}, r.client)
-		if err != nil {
-			r.status.SetDegraded(operatorv1.ResourceReadError, "Error reading the RBAC management UI ConfigMap", err, reqLogger)
-			return reconcile.Result{}, err
-		}
-		rbacManagementEnabled = render.RBACManagementEnabled(gate)
+	rbacManagementEnabled, err := utils.RBACManagementEnabled(ctx, r.client, instance.Spec.Variant, false)
```
pasanw: I can't say I'm very knowledgeable on multi-tenancy, but Claude is saying that just because kube-controllers is always a cluster-scoped deployment, it doesn't mean that it is never run on a multi-tenant management cluster. It's suggesting that we wire through the multi-tenancy signal from opts.MultiTenant.

Happy to leave this up to you to evaluate and determine the correct approach

pasanw (approved):

PR #5127, opened 2026-07-30
Read the DatastoreMigration CRD at v1

caseydavenport:
## Description

The DatastoreMigration CRD moves to `migration.projectcalico.org/v1` in Calico v3.33, and the v3.33 CRD serves v1beta1 (deprecated) alongside v1 (storage). There's no single version the operator can hardcode and be right on every cluster, so it now resolves the served version from discovery at startup, preferring v1 and falling back to v1beta1. That version is used for:

- scheme registration
- the startup check for an already-migrated cluster
- both controller watches

Get it wrong and the watch never becomes ready, the migration phase reads as empty, and the apiserver controller re-creates the aggregated APIService while the migration is deleting it.

An earlier version of this PR assumed an upgraded v3.32 cluster had no CR left to watch, since applying the new CRD meant deleting the old one. Deleting the CRD runs the migration finalizer for real, which is why the new CRD serves both versions instead.

Related: [CORE-12573](https://tigera.atlassian.net/browse/CORE-12573)

```release-note
None
```

[CORE-12573]: https://tigera.atlassian.net/browse/CORE-12573?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/controller/migration/datastoremigration/version.go
+++ b/pkg/controller/migration/datastoremigration/version.go
@@ -15,52 +15,89 @@
 package datastoremigration
 
 import (
+	"context"
+	"fmt"
+	"sync"
+	"time"
+
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/discovery"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	initialRetry = 1 * time.Second
+	maxRetry     = 30 * time.Second
 )
 
 var log = ctrl.Log.WithName("datastoremigration")
 
-// ResolveServedVersion points the package at the version the cluster serves. Call it before
-// AddToScheme: scheme, cache and watches must agree.
-func ResolveServedVersion(disco discovery.DiscoveryInterface) {
-	gv, ok := ServedGroupVersion(disco)
-	if !ok {
-		// The CRD is installed by the user when they migrate, so it's usually absent at startup.
-		log.V(1).Info("No DatastoreMigration CRD served, defaulting", "groupVersion", SchemeGroupVersion)
-		return
-	}
-	log.Info("Resolved served DatastoreMigration version", "groupVersion", gv)
-	SchemeGroupVersion = gv
+var (
+	servedMutex   sync.RWMutex
+	servedVersion = GroupVersionV1
+)
+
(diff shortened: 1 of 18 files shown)


pkg/controller/migration/datastoremigration/version.go
```diff
@@ -0,0 +1,66 @@
...
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/discovery"
+	ctrl "sigs.k8s.io/controller-runtime"
+)
+
+var log = ctrl.Log.WithName("datastoremigration")
+
+// ResolveServedVersion points the package at the version the cluster serves. Call it before
+// AddToScheme: scheme, cache and watches must agree.
+func ResolveServedVersion(disco discovery.DiscoveryInterface) {
```
MichalFupso: Is there a way for a user to install v1beta crd in 3.33 cluster? If there is no crd at the start of the day and we default to v1 and the user installs v1beta we default to the wrong version

pkg/controller/migration/datastoremigration/version.go
```diff
@@ -0,0 +1,66 @@
...
+	SchemeGroupVersion = gv
+}
+
+// ServedGroupVersion returns the DatastoreMigration group/version the cluster serves, preferring v1.
+// The second return is false when the CRD isn't installed.
+func ServedGroupVersion(disco discovery.DiscoveryInterface) (schema.GroupVersion, bool) {
+	for _, gv := range []schema.GroupVersion{GroupVersionV1, GroupVersionV1beta1} {
+		resources, err := disco.ServerResourcesForGroupVersion(gv.String())
+		if err != nil {
+			if !apierrors.IsNotFound(err) {
```
MichalFupso: Should there be a mechanism to retry if it's a transient error instead of not-found

pkg/controller/migration/datastoremigration/version.go
caseydavenport: Fixed!

pkg/controller/migration/datastoremigration/version.go
caseydavenport: Fixed!

pkg/controller/migration/datastoremigration/version.go
```diff
@@ -0,0 +1,103 @@
...
+			}
+		}
+	}
+	return schema.GroupVersion{}, false, nil
+}
+
+// WatchObject returns an empty DatastoreMigration at the served version, for controllers
+// registering a watch.
+func WatchObject() client.Object {
+	u := &unstructured.Unstructured{}
```
MichalFupso, line 100: *unstructured.Unstructured implements metav1.Object but not metav1.ObjectMetaAccessor — it has no GetObjectMeta()

MichalFupso (approved):

PR #5131, opened 2026-07-31
Resolve the product variant once at startup

caseydavenport:
The operator decides which variant it is running in four places: the `--variant` flag, Enterprise CRD discovery, the Installation spec, and the Installation status. The one gating most controllers is CRD discovery, which is not what the user set and stays true after a downgrade because the CRDs outlive the variant change.

This resolves the variant once from the Installation before any controller starts, and restarts the operator from a single watch when it changes. Design: https://github.com/tigera/designs/pull/89 (CORE-13240).

## Description

- the variant is resolved in `main` from the Installation, merging the overlay, and passed to controllers as one option
- restart-on-variant-change moves to a watch on the manager cache; the reboot in the installation controller goes away, along with its ability to fire mid-uninstall
- the bootstrap ConfigMap watch moves onto the manager cache too, which drops a goroutine that could never be stopped
- registration gates key on the variant rather than on whether the Enterprise CRDs exist
- `GetInstallationSpec` no longer returns the status variant, so controllers cannot re-derive it
- the `--variant` flag is now bootstrap-only and validated; an unrecognised value used to be treated as Calico, which installs the wrong CRDs for the bootstrap-crds path where nothing runs afterwards to correct them
- `updateCRDs` installs CRDs for the overlay-merged variant. It read the spec variant before the overlay was merged in, so an overlay that set the variant got the wrong CRD set.
- `fillDefaults` defaults to the process variant instead of hardcoding Calico, so it and `main` cannot disagree

Behaviour changes worth calling out:

- an Enterprise install whose Installation omits the variant now comes up as Enterprise. Previously `fillDefaults` overrode it to Calico.
- asking for Enterprise without the Enterprise CRDs now fails at startup with a message rather than degrading on every reconcile.

Also fixes a latent ImageSet bug. During an Enterprise install the status variant lags the spec, so controllers asked for a `calico-` ImageSet. `GetImageSet` only errors when an ImageSet with the matching prefix exists, so on a cluster carrying only `enterprise-<version>` it found nothing, returned no error, and the components deployed with built-in image tags. Air-gapped clusters fail to pull; everyone else silently gets floating tags instead of the digests they pinned.

```release-note
Fixes an issue where components could be deployed with default image tags instead of the images from a configured ImageSet during a Calico Enterprise installation.
```

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -435,6 +435,23 @@ If a value other than 'all' is specified, the first CRD with a prefix of the spe
 		}
 	}
 
+	// The enterprise controllers can't register without their APIs, so stop here with something
+	// actionable rather than failing later. A restart picks it up once the CRDs are installed.
+	if bootVariant.IsEnterprise() {
+		enterpriseAPIs, err := discovery.RequiresTigeraSecure(cs)
+		if err != nil {
+			setupLog.Error(err, "Failed to determine whether the Enterprise APIs are available")
+			os.Exit(1)
+		}
+		if !enterpriseAPIs {
+			setupLog.Error(
+				fmt.Errorf("missing Tigera Secure custom resource definitions"),
+				"Cannot run as Calico Enterprise, install the Enterprise CRDs first",
+			)
+			os.Exit(1)
+		}
+	}
+
 	// Start a goroutine to handle termination.
 	go func() {
 		// Cancel the main context when we are done.
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -26,6 +26,7 @@ import (
 	"time"
 
 	"github.com/cloudflare/cfssl/log"
+	"github.com/go-logr/logr"
 	"github.com/tigera/operator/pkg/render/common/cloudconfig"
 
 	v3 "github.com/tigera/api/pkg/apis/projectcalico/v3"
@@ -57,7 +58,6 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apiextensions-apiserver/pkg/apis/apiextensions"
 	"k8s.io/apimachinery/pkg/api/errors"
-	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
@@ -430,11 +430,7 @@ admission policy installation; once an Installation exists it is the authority o
 	}
 
 	// Resolve the variant now that the operator CRDs exist.
-	bootVariant, err := resolveBootVariant(ctx, c, operatortigeraiov1.ProductVariant(bootstrapVariant))
-	if err != nil {
-		setupLog.Error(err, "Failed to resolve the product variant")
-		os.Exit(1)
-	}
+	bootVariant := waitForVariant(ctx, c, setupLog)
 	setupLog.WithValues("variant", bootVariant).Info("Resolved product variant")
 
 	// The bootstrap pass above used the flag default, which doesn't cover the enterprise APIs.
@@ -456,10 +452,7 @@ admission policy installation; once an Installation exists it is the authority o
 			os.Exit(1)
 		}
 		if !enterpriseAPIs {
-			setupLog.Error(
-				fmt.Errorf("the Calico Enterprise CRDs are not installed"),
-				"Cannot run as Calico Enterprise",
-			)
+			setupLog.Error(fmt.Errorf("the Calico Enterprise CRDs are not installed"), "Cannot run as Calico Enterprise")
 			os.Exit(1)
 		}
 	}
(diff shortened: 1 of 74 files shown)


cmd/main.go
```diff
@@ -413,6 +429,41 @@ If a value other than 'all' is specified, the first CRD with a prefix of the spe
...
+	// The enterprise controllers can't register without their APIs. Exiting lets the kubelet
+	// retry us once the CRDs are installed.
+	if bootVariant.IsEnterprise() {
+		enterpriseAPIs, err := discovery.EnterpriseAPIsExist(cs)
+		if err != nil {
+			setupLog.Error(err, "Failed to determine whether the Enterprise APIs are available")
+			os.Exit(1)
+		}
+		if !enterpriseAPIs {
+			setupLog.Error(
```
caseydavenport: no need to truncate lines this short.

cmd/main.go
```diff
@@ -413,6 +429,41 @@ If a value other than 'all' is specified, the first CRD with a prefix of the spe
 		}
 	}
 
+	// Resolve the variant now that the operator CRDs exist.
+	bootVariant, err := resolveBootVariant(ctx, c, operatortigeraiov1.ProductVariant(bootstrapVariant))
```
caseydavenport: This is wrong - we need to wait until an Installation exists before we continue here. We don't want to ever fall back to using the botstrap variant for anything more than bootstrapping the CRDs above. Eventually that flag will just be gone.

pkg/controller/csr/csr_controller.go
```diff
@@ -128,14 +128,14 @@ func newReconciler(mgr manager.Manager, opts options.ControllerOptions) (reconci
...
-		allowedTLSAssets:    allowedAssets(opts.ClusterDomain),
-		enterpriseCRDExists: opts.EnterpriseCRDExists,
+		client:           mgr.GetClient(),
+		clientset:        opts.K8sClientset,
+		calicoClient:     calicoClient,
+		scheme:           mgr.GetScheme(),
+		provider:         opts.DetectedProvider,
+		clusterDomain:    opts.ClusterDomain,
+		allowedTLSAssets: allowedAssets(opts.ClusterDomain),
+		enterprise:       opts.Variant.IsEnterprise(),
```
Brian-McM: Nit, could you rename the variable to isEnterprise just so it's apparent this is a bool when reading it / answering the question if it's enterprise or not.

pkg/controller/gatewayapi/gatewayapi_controller.go
```diff
@@ -73,10 +73,10 @@ func Add(mgr manager.Manager, opts options.ControllerOptions) error {
 	r := &ReconcileGatewayAPI{
 		client:              mgr.GetClient(),
 		scheme:              mgr.GetScheme(),
-		enterpriseCRDsExist: opts.EnterpriseCRDExists,
 		tierWatchReady:      &utils.ReadyFlag{},
 		status:              status.New(mgr.GetClient(), "gatewayapi", opts.KubernetesVersion),
 		clusterDomain:       opts.ClusterDomain,
+		variant:             opts.Variant,
```
Brian-McM, line 79: In some places you've change this to `enterprise` as a bool, and some you've just passed on the variant. Is there a reason to do this differently, or can we pass the variant everywhere or just pass the `enterprise`?

cmd/main.go
```diff
@@ -413,6 +429,34 @@ If a value other than 'all' is specified, the first CRD with a prefix of the spe
 		}
 	}
 
+	// Resolve the variant now that the operator CRDs exist.
+	bootVariant := waitForVariant(ctx, c, setupLog)
```
Brian-McM: This isn't the boot variant now, is it? It's the real variant, or am I missing something?

cmd/main.go
```diff
@@ -664,6 +706,100 @@ func setKubernetesServiceEnv(kubeconfigFile string) error {
...
+func waitForVariant(ctx context.Context, c client.Client, log logr.Logger) operatortigeraiov1.ProductVariant {
+	waiting := false
+	for {
+		variant, err := effectiveVariant(ctx, c)
+		switch {
+		case err != nil:
+			log.Error(err, "Failed to read the Installation, will retry")
+		case variant != "":
+			return variant
+		case !waiting:
```
Brian-McM: Is the point of the waiting variable just so that we log that info once?

cmd/main.go
```diff
@@ -664,6 +706,100 @@ func setKubernetesServiceEnv(kubeconfigFile string) error {
...
+			log.Info("Waiting for an Installation")
+			waiting = true
+		}
+
+		select {
```
Brian-McM, line 723: I wonder if it would be more effective to add a watch instead of poll, since we require a change to happen if the variant isn't set.

caseydavenport: Yeah, I looked at that. The problem is ordering: this runs before `mgr.Start()`, so there's no informer to hang a watch off yet, and building a standalone one means a second client plus its own cache for what is a one-shot wait. It's also two objects, since `effectiveVariant` merges the overlay in. A 2s poll that ends the moment an Installation lands felt like the better trade.

Brian-McM (approved):

PR #5132, opened 2026-07-31
Reconfigure Linseed and linseed user when migrating from multi-index to single-index

asincu:
## Description

- Linseed will reconfigure its environment variables to set the correct backend for the
indices it is using and also set the base index name used for Cloud.
- ES Kubecontrollers will stop running elasticsearch configuration
- LogStorage Controllers Users will configure Linseed user instead of es-kubecontrollers.

## Release Note

```release-note
TBD
```

## For PR author

- [ ] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/common/discovery/discovery.go
+++ b/pkg/common/discovery/discovery.go
@@ -306,3 +306,19 @@ func ElasticIsMigrating(config *corev1.ConfigMap) bool {
 	}
 	return false
 }
+
+// UseSingleIndex returns true if this single tenant management cluster is in the last phase of a migration to single-index
+// storage, during which the operator must reconfigure Linseed to use the single-index names.
+func UseSingleIndex(config *corev1.ConfigMap) bool {
+	if config == nil {
+		return false
+	}
+
+	// Load the operator bootstrap configuration from its configmap.
+	if val, ok := config.Data["USE_SINGLE_INDEX"]; ok && val != "" {
+		if strings.ToLower(val) == "true" {
+			return true
+		}
+	}
+	return false
+}
--- a/pkg/common/discovery/discovery.go
+++ b/pkg/common/discovery/discovery.go
@@ -308,3 +308,19 @@ func ElasticIsMigrating(config *corev1.ConfigMap) bool {
 	}
 	return false
 }
+
+// IndexIsMigrating returns true if this cluster is in the last phase of a migration to single-index
+// storage, during which the operator must reconfigure Linseed to use the single-index names.
+func IndexIsMigrating(config *corev1.ConfigMap) bool {
+	if config == nil {
+		return false
+	}
+
+	// Load the operator bootstrap configuration from its configmap.
+	if val, ok := config.Data["INDEX_MIGRATION"]; ok && val != "" {
+		if strings.ToLower(val) == "true" {
+			return true
+		}
+	}
+	return false
+}
--- a/pkg/controller/intrusiondetection/intrusiondetection_controller.go
+++ b/pkg/controller/intrusiondetection/intrusiondetection_controller.go
@@ -437,7 +437,7 @@ func (r *ReconcileIntrusionDetection) Reconcile(ctx context.Context, request rec
 			r.status.SetDegraded(operatorv1.ResourceReadError, "Failed to read cloud config", err, reqLogger)
 			return reconcile.Result{}, err
 		}
-		tenant = cloudConfig.ToTenant()
+		tenant = cloudConfig.ToTenant(false)
 	}
 
 	// Create a component handler to manage the rendered component.
--- a/pkg/controller/logstorage/dashboards/dashboards_controller.go
+++ b/pkg/controller/logstorage/dashboards/dashboards_controller.go
@@ -324,8 +324,7 @@ func (d DashboardsSubController) Reconcile(ctx context.Context, request reconcil
 	}
 
 	// Query the username and password this Dashboards Installer instance should use to authenticate with Elasticsearch.
-	// For multi-tenant systems, credentials are created by the elasticsearch users controller.
-	// For single-tenant system, these are created by es-kube-controllers.
+	// For cloud, credentials are created by the elasticsearch users controller.
 	key = types.NamespacedName{Name: dashboards.ElasticCredentialsSecret, Namespace: helper.InstallNamespace()}
 	credentials := corev1.Secret{}
 	if err = d.client.Get(ctx, key, &credentials); err != nil && !errors.IsNotFound(err) {
--- a/pkg/controller/logstorage/dashboards/dashboards_controller.go
+++ b/pkg/controller/logstorage/dashboards/dashboards_controller.go
@@ -288,7 +288,7 @@ func (d DashboardsSubController) Reconcile(ctx context.Context, request reconcil
 				d.status.SetDegraded(operatorv1.ResourceReadError, "Failed to read cloud config", err, reqLogger)
 				return reconcile.Result{}, err
 			}
-			tenant = cloudConfig.ToTenant()
+			tenant = cloudConfig.ToTenant(false)
 		}
 
 		// Determine the host and port from the URL.
@@ -321,7 +321,8 @@ func (d DashboardsSubController) Reconcile(ctx context.Context, request reconcil
 
 	// Query the username and password this Dashboards Installer instance should use to authenticate with Elasticsearch.
 	// For multi-tenant systems, credentials are created by the elasticsearch users controller.
-	// For single-tenant system, these are created by es-kube-controllers.
+	// For single-tenant systems, these are created by es-kube-controllers, unless the cluster is
+	// migrating to single-index storage - then the users controller creates them as well.
 	key = types.NamespacedName{Name: dashboards.ElasticCredentialsSecret, Namespace: helper.InstallNamespace()}
 	credentials := corev1.Secret{}
 	if err = d.client.Get(ctx, key, &credentials); err != nil && !errors.IsNotFound(err) {
--- a/pkg/controller/logstorage/initializer/conditions_controller.go
+++ b/pkg/controller/logstorage/initializer/conditions_controller.go
@@ -17,6 +17,7 @@ package initializer
 import (
 	"context"
 	"fmt"
+	"sort"
 	"time"
 
 	"k8s.io/apimachinery/pkg/api/errors"
@@ -44,6 +45,7 @@ func AddConditionsController(mgr manager.Manager, opts options.ControllerOptions
 		client:      mgr.GetClient(),
 		scheme:      mgr.GetScheme(),
 		multiTenant: opts.MultiTenant,
+		cloud:       opts.Cloud,
 	}
 
 	return ctrl.NewControllerManagedBy(mgr).
@@ -62,6 +64,10 @@ type LogStorageConditions struct {
 	client      client.Client
 	scheme      *runtime.Scheme
 	multiTenant bool
+
+	// cloud indicates that this is a Calico Cloud install, in which case the log-storage users
+	// controller runs in single-tenant mode too and reports status.
+	cloud bool
 }
 
 func (r *LogStorageConditions) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
@@ -91,6 +97,12 @@ func (r *LogStorageConditions) Reconcile(ctx context.Context, request reconcile.
 	ls.Status.Conditions = updateConditions(currentConditions, desiredConditions)
 
 	if err := r.client.Status().Update(ctx, ls); err != nil {
+		if errors.IsConflict(err) {
+			// The LogStorage was modified after we read it - our cached copy is stale. Requeue and
+			// recompute the conditions from the updated object instead of reporting an error.
+			reqLogger.V(3).Info("Conflict updating LogStorage status conditions, retrying")
+			return reconcile.Result{Requeue: true}, nil
+		}
 		log.WithValues("reason", err).Info("Failed to update LogStorage status conditions")
 		return reconcile.Result{}, err
 	}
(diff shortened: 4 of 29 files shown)


pkg/render/kubecontrollers/kube-controllers.go
```diff
@@ -282,7 +287,12 @@ func NewElasticsearchKubeControllers(cfg *KubeControllersConfiguration) *kubeCon
 	var enabledControllers []string
 	if !cfg.Tenant.MultiTenant() {
 		// Zero and single tenant cluster needs elasticsearch configuration
-		enabledControllers = append(enabledControllers, "authorization", "elasticsearchconfiguration")
+		enabledControllers = append(enabledControllers, "authorization")
```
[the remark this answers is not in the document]
asincu: MultiTenant() is a nil-safe method on *Tenant (api/v1/tenant_types.go:140):

func (t *Tenant) MultiTenant() bool {
      return t != nil && t.GetNamespace() != ""
}

pkg/common/discovery/discovery.go
```diff
@@ -306,3 +306,19 @@ func ElasticIsMigrating(config *corev1.ConfigMap) bool {
 	}
 	return false
 }
+
+// UseSingleIndex returns true if this cluster is in the last phase of a migration to single-index
+// storage, during which the operator must reconfigure Linseed to use the single-index names.
+func UseSingleIndex(config *corev1.ConfigMap) bool {
```
caseydavenport: The comment suggests that this function is only applicable for migration, but the name of the function sounds more general. Perhaps this would be better named "SingleIndexMigrationDone" or similar? Or is the comment just overly specific?

pkg/controller/intrusiondetection/intrusiondetection_controller.go
```diff
@@ -437,7 +437,7 @@ func (r *ReconcileIntrusionDetection) Reconcile(ctx context.Context, request rec
 			r.status.SetDegraded(operatorv1.ResourceReadError, "Failed to read cloud config", err, reqLogger)
 			return reconcile.Result{}, err
 		}
-		tenant = cloudConfig.ToTenant()
+		tenant = cloudConfig.ToTenant(false)
```
caseydavenport: A nit, but I might suggest using a typed constant for this instead of true/false bools. Readers of ToTenant don't really know what true/false means without reading the function itself.

pkg/controller/logstorage/dashboards/dashboards_controller.go
```diff
@@ -323,7 +323,8 @@ func (d DashboardsSubController) Reconcile(ctx context.Context, request reconcil
 
 	// Query the username and password this Dashboards Installer instance should use to authenticate with Elasticsearch.
 	// For multi-tenant systems, credentials are created by the elasticsearch users controller.
-	// For single-tenant system, these are created by es-kube-controllers.
+	// For single-tenant systems, these are created by es-kube-controllers, unless the cluster is
+	// migrating to single-index storage - then the users controller creates them as well.
 	key = types.NamespacedName{Name: dashboards.ElasticCredentialsSecret, Namespace: helper.InstallNamespace()}
```
caseydavenport: // For single-tenant systems using multi-index mode, these are created by es-kube-controllers.

pkg/controller/logstorage/initializer/conditions_controller.go
```diff
@@ -88,9 +95,22 @@ func (r *LogStorageConditions) Reconcile(ctx context.Context, request reconcile.
...
 
 	// Compare and update the current StatusCondition if there are any new changes
-	ls.Status.Conditions = updateConditions(currentConditions, desiredConditions)
+	conditions := updateConditions(currentConditions, desiredConditions)
+
+	// Skip the write if nothing changed. This controller watches LogStorage, so a no-op write would
+	// re-trigger it and spin: each reconcile would bump the resourceVersion and enqueue another one.
+	if equality.Semantic.DeepEqual(ls.Status.Conditions, conditions) {
+		return reconcile.Result{}, nil
+	}
```
caseydavenport: This seems unnecessary if we're doing the sorting below - the k8s API server will notice that nothing has changed, and won't update the generation nor send an update. So I thinkwe can remove this check (not that it's a problem really, but it's another place something could go wrong if the equality check is not correct).

pkg/controller/logstorage/initializer/conditions_controller.go
```diff
@@ -88,9 +95,22 @@ func (r *LogStorageConditions) Reconcile(ctx context.Context, request reconcile.
...
+	ls.Status.Conditions = conditions
 
 	if err := r.client.Status().Update(ctx, ls); err != nil {
+		if errors.IsConflict(err) {
+			// The LogStorage was modified after we read it - our cached copy is stale. Requeue and
+			// recompute the conditions from the updated object instead of reporting an error.
+			reqLogger.V(1).Info("Conflict updating LogStorage status conditions, retrying")
```
caseydavenport: I think this should be V(2) for debug. V(1) is info IIRC?

pkg/controller/logstorage/linseed/linseed_controller.go
```diff
@@ -89,6 +90,7 @@ func Add(mgr manager.Manager, opts options.ControllerOptions) error {
 		status:          status.New(mgr.GetClient(), "log-storage-access", opts.KubernetesVersion),
 		elasticExternal: opts.ElasticExternal,
 		cloud:           opts.Cloud,
+		useSingleIndex:  opts.UseSingleIndex,
```
caseydavenport: a nit, but would be a bit cleaner to just do `opts: opts` instead of copying every field from opts.

pkg/controller/logstorage/linseed/linseed_controller.go
```diff
@@ -363,7 +365,8 @@ func (r *LinseedSubController) Reconcile(ctx context.Context, request reconcile.
 
 	// Query the username and password this Linseed instance should use to authenticate with Elasticsearch.
 	// For multi-tenant systems, credentials are created by the elasticsearch users controller.
-	// For single-tenant system, these are created by es-kube-controllers.
+	// For single-tenant systems, these are created by es-kube-controllers, unless the cluster is
+	// migrating to single-index storage - then the users controller creates them as well.
```
caseydavenport: same nit on comment phrasing here.

pkg/controller/logstorage/linseed/linseed_controller.go
```diff
@@ -420,7 +423,8 @@ func (r *LinseedSubController) Reconcile(ctx context.Context, request reconcile.
...
 	// Delay installing Linseed until available.
 	// TODO: Switch single-tenant to using operator-provisioned users.
```
caseydavenport: I think this TODO can probably just be deleted now?

pkg/controller/logstorage/users/users_controller.go
```diff
@@ -72,9 +79,10 @@ func Add(mgr manager.Manager, opts options.ControllerOptions) error {
 	if !opts.Variant.IsEnterprise() {
 		return nil
 	}
-	if !opts.MultiTenant {
-		// For now, the operator only creates users in multi-tenant mode. In single-tenant mode,
-		// user creation is handled by es-kube-controllers instead.
+	if !opts.MultiTenant && !opts.Cloud {
+		// The operator creates users for multi-tenant clusters, and for the single-tenant clusters of
+		// Calico Cloud - which build their tenant configuration from the cloud config ConfigMap. Anywhere
+		// else, user creation is handled by es-kube-controllers instead.
```
caseydavenport: I think this conditional is a bit confusing - EE doesn't support multi-tenant at all, so if it's not multi-tenant and it's not cloud, it's EE, but that's the same thing as !opts.Cloud (you'll never have !opts.Cloud && opts.MultiTenant).

pkg/controller/logstorage/users/users_controller.go
```diff
@@ -119,6 +128,12 @@ func Add(mgr manager.Manager, opts options.ControllerOptions) error {
 		if err = c.WatchObject(&operatorv1.Tenant{}, &handler.EnqueueRequestForObject{}); err != nil {
 			return fmt.Errorf("log-storage-user-controller failed to watch Tenant resource: %w", err)
 		}
+	} else {
+		// In single-tenant mode, the tenant configuration - including the Elasticsearch endpoint -
+		// comes from the cloud config ConfigMap.
```
caseydavenport: Seems like longer term, it might be nice to move off of this config map and instead using a proper Tenant CR even for single-tenant deployments? That way we standardize on the same configuration struct for both modes and don't need as many branches. WDYT?

pkg/controller/logstorage/users/users_controller.go
```diff
@@ -325,6 +363,35 @@ func (r *UserController) Reconcile(ctx context.Context, request reconcile.Reques
...
+
+	if clusterID == "" {
+		err := fmt.Errorf("%s/%s ConfigMap value for key 'cluster-id' must be non-empty",
+			clusterIDConfigMap.Namespace, clusterIDConfigMap.Name)
+		r.status.SetDegraded(operatorv1.ResourceReadError, fmt.Sprintf("%v", err), err, reqLogger)
+		return "", err
+	}
+
+	return clusterID, nil
+}
```
caseydavenport: Huh, was there no common utility for this already? Seems odd.

pkg/controller/options/options.go
```diff
@@ -61,6 +61,10 @@ type ControllerOptions struct {
 	// LSS configuration and internal elasticsearch running. Only meaningful when Cloud is set.
 	ESMigration bool
 
+	// UseSingleIndex is enabled in the last phase of an index migration for a single tenant cluster,
+	// during which the operator reconfigures log storage to use the single-index names.
+	UseSingleIndex bool
```
caseydavenport, line 67: Is this only used in migration?

pkg/controller/utils/elasticsearch.go
```diff
@@ -220,8 +220,71 @@ var (
 	ElasticsearchUserNameDashboardInstaller = "tigera-ee-dashboards-installer"
 )
 
-func LinseedUser(clusterID, tenant string) *User {
-	username := formatName(ElasticsearchUserNameLinseed, clusterID, tenant)
+// ElasticsearchSecureUserSuffix is appended to the user names provisioned for single-tenant clusters.
+// It maintains the 1:1 mapping between the public user propagated to components and the private user
+// swapped in at ES gateway, which strips this suffix.
+const ElasticsearchSecureUserSuffix = "secure"
```
caseydavenport: Is this for backwards compat with how es-kube-controllers programmed the single-tenant users? If so we should say it in the comment and probably name this to something like LegacySingleTenantUserSuffix or something similar.

pkg/render/common/cloudconfig/cloudconfig.go
```diff
@@ -82,8 +102,12 @@ type CloudConfig struct {
 // ToTenant converts the given CloudConfig structure to a Tenant object.
 // This allows controllers that have been converted to support multi-tenancy to still leverage
 // the single-tenant CloudConfig structure using the same code path as in multi-tenancy.
-func (c CloudConfig) ToTenant() *v1.Tenant {
-	return &v1.Tenant{
+//
+// useSingleIndex declares the standard single-index base names on the returned Tenant. Only clusters
+// migrating to single-index storage should set it: index base names are otherwise not carried on the
+// artificial single-tenant Tenant, and components fall back to their default index names.
+func (c CloudConfig) ToTenant(useSingleIndex bool) *v1.Tenant {
```
caseydavenport: if we did something like this:

    ```go
    type IndexType string

    const (
        IndexTypeSingle   IndexType = "primary"
        IndexTypeMulti IndexType = "secondary"
    )
    ```

caseydavenport: if we did something like this:

```go
type IndexType string

const (
    IndexTypeSingle   IndexType = "single"
    IndexTypeMulti    IndexType = "multi"
)
```

It would make the code that calls this more readable and less error prone, perhaps.

pkg/render/kubecontrollers/kube-controllers.go
```diff
@@ -122,7 +122,9 @@ type KubeControllersConfiguration struct {
 	TenantID string
 
 	// Cloud indicates kube-controllers is being rendered for a Calico Cloud install. When false the
-	// cloud-specific RBAC below is not granted and enterprise RBAC is unchanged.
+	// cloud-specific RBAC below is not granted and enterprise RBAC is unchanged. In Calico Cloud,
+	// Elasticsearch configuration - including provisioning of the Linseed user - is handled by the
+	// operator's log-storage users controller rather than by es-kube-controllers.
```
caseydavenport: This looks like overcommenting - don't think we need to change this comment at all.

caseydavenport (commented):
Just a few thoughts, mostly nits!

pkg/controller/logstorage/users/users_controller.go
asincu: I agree, but we are not planning on investing time or changes on single tenant management clusters.

pkg/controller/logstorage/initializer/conditions_controller.go
asincu: V1 should be debug. V2 and V3 is shown when we pass the log level needed, not just plain debug. I think chatter when reconcile is set at V3. I will change it to V3.

pkg/controller/logstorage/users/users_controller.go
asincu: What about Cloud with internal elasticsearch ? In all fairness, they will go away soon. We still have a couple of environments yet.

asincu: Nvm , will make the change

pkg/controller/options/options.go
asincu: We will keep this on until we migrate traffic to the MT environment.

pkg/render/common/cloudconfig/cloudconfig.go
asincu: I ended up using options because we only need to backfill the indicies in Linseed Controller.

pkg/common/discovery/discovery.go
asincu: We will need to leave this on even after the migration is done. Will be turned off once we migrate them to a tenant.

pkg/controller/logstorage/users/users_controller.go
asincu: There was not one, but I added now.

caseydavenport (approved):
LGTM!

PR #5134, opened 2026-08-03
Create the node metrics service for Calico

caseydavenport:
## Description

The operator already creates metrics Services for Typha and kube-controllers, but not for Felix, so Calico users following the [component metrics guide](https://docs.tigera.io/calico/latest/operations/monitor/monitor-component-metrics) have to hand-roll one. This renders the existing `calico-node-metrics` Service for Calico too, carrying only Felix's port, and removes it again if Felix metrics are turned back off. Enterprise keeps its reporter and BGP ports, unchanged.

Also fixes the Felix metrics port being read out of FelixConfiguration only for Enterprise. Nothing consumed it in a Calico install before this change, but with the Service rendered it would have pointed at the default port no matter what the user configured.

Windows is not covered here. The Windows metrics Service has never carried Felix's port, even in Enterprise, and getting one to work needs a firewall rule, so it wants its own change.

Related: https://github.com/projectcalico/calico/issues/10755, [CORE-13310](https://tigera.atlassian.net/browse/CORE-13310)

## Release Note

```release-note
Calico installations now get a Service for scraping Felix Prometheus metrics when Felix metrics are enabled, so one no longer has to be created by hand.
```

## For PR author

- [x] Tests for change.

[CORE-13310]: https://tigera.atlassian.net/browse/CORE-13310?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/controller/installation/core_controller.go
+++ b/pkg/controller/installation/core_controller.go
@@ -1256,6 +1256,9 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 	calicoVersion := components.CalicoRelease
 
 	felixPrometheusMetricsPort := defaultFelixMetricsDefaultPort
+	if felixConfiguration.Spec.PrometheusMetricsPort != nil {
+		felixPrometheusMetricsPort = *felixConfiguration.Spec.PrometheusMetricsPort
+	}
 
 	if instance.Spec.Variant.IsEnterprise() {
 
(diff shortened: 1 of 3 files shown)


caseydavenport:
Marking this a draft. The gating is wrong: `nodeMetricsPort` enables Felix metrics by setting env on calico-node rather than writing the port to FelixConfiguration, so this misses anyone who enabled metrics that way, and targets the wrong port when both paths are set with different values.

`nodeMetricsPort` should be writing the port to FelixConfiguration like we already do for eBPF and nftables, which reduces this to a couple of lines. That has the operator owning fields a user may have set themselves, so it's waiting on #4523. Background in https://github.com/projectcalico/calico/issues/10755.

PR #5138, opened 2026-08-04
Add a readiness probe to the calico-webhooks deployment

caseydavenport:
Rolling the calico-webhooks deployment could briefly reject policy writes. The webhook pod joins the Service endpoints as soon as its container starts, before it is actually serving, and the tiered RBAC webhook fails closed, so a policy write that lands in that window gets rejected.

This adds a readiness probe to the webhook container so the pod only takes traffic once it is serving, matching the probe already present in the Calico helm chart.

Related: [CORE-13250](https://tigera.atlassian.net/browse/CORE-13250)

```release-note
Fixes a brief window during a calico-webhooks rollout where policy writes could be rejected.
```

[CORE-13250]: https://tigera.atlassian.net/browse/CORE-13250?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


electricjesus (commented):
looks pretty solid to me! just a couple of comments..

pkg/render/webhooks/render.go
```diff
@@ -222,6 +238,7 @@ func (c *component) Objects() ([]client.Object, []client.Object) {
 
 	// Read the final container port from the deployment (after overrides) for use in the Service.
 	containerPort := dep.Spec.Template.Spec.Containers[0].Ports[0].ContainerPort
+	dep.Spec.Template.Spec.Containers[0].ReadinessProbe.HTTPGet.Port = intstr.FromInt32(containerPort)
```
electricjesus, line 241: looks like the container port override never reaches the binary. We don't pass `--port` in the Args above, so the webhook keeps listening on 6443 whatever `containerPort` says. The Service targetPort has the same gap today, but the probe makes it louder: override to anything but 6443 and the pod never goes Ready, so with `maxUnavailable: 0` the rollout just sits there.

Both binaries take the flag (`webhooks/pkg/webhook/command.go`, OSS and enterprise), and passing the default is a no-op for anyone not using the override. Could we set it here, where `containerPort` is final?

```suggestion
	dep.Spec.Template.Spec.Containers[0].ReadinessProbe.HTTPGet.Port = intstr.FromInt32(containerPort)

	// The server binds its own port, so tell it which one. Without this a port override would
	// leave the Service, the probe and the listener disagreeing.
	dep.Spec.Template.Spec.Containers[0].Args = append(
		dep.Spec.Template.Spec.Containers[0].Args,
		fmt.Sprintf("--port=%d", containerPort),
	)
```

ran the package tests with this locally and they're green..

pkg/render/webhooks/render_test.go
```diff
@@ -468,6 +486,7 @@ var _ = Describe("Webhooks rendering tests", func() {
 		// Verify the container port is set to the custom value.
 		dep := rtest.GetResource(resources, webhooks.WebhooksName, common.CalicoNamespace, "apps", "v1", "Deployment").(*appsv1.Deployment)
 		Expect(dep.Spec.Template.Spec.Containers[0].Ports[0].ContainerPort).To(Equal(customPort))
+		Expect(dep.Spec.Template.Spec.Containers[0].ReadinessProbe.HTTPGet.Port.IntValue()).To(Equal(int(customPort)))
```
electricjesus, line 489: tiny: worth pinning the flag here too, so all three stay wired together?

```suggestion
		Expect(dep.Spec.Template.Spec.Containers[0].ReadinessProbe.HTTPGet.Port.IntValue()).To(Equal(int(customPort)))
		Expect(dep.Spec.Template.Spec.Containers[0].Args).To(ContainElement(fmt.Sprintf("--port=%d", customPort)))
```

electricjesus (approved):

pkg/render/webhooks/render.go
caseydavenport: Yep, confirmed - both binaries take `--port` with a 6443 default and we never passed it, so the probe would have been polling a port nothing was listening on. Added it where `containerPort` is final.

pkg/render/webhooks/render_test.go
caseydavenport: Done.

PR #5139, opened 2026-08-04
Hold the API server cutover until the new API server is ready

caseydavenport:
## Description

Alternative fix for EV-6821, opened alongside #5103 so the two can be compared.

On a direct upgrade the operator creates the new API server and repoints the projectcalico.org/v3 APIService at it in the same pass, without checking that the new pod can serve. A leftover default-deny policy denies that pod, and that policy can only be deleted through the API server it denies.

This PR holds the repoint - along with the removal of the old API server's namespace and RBAC - until the new API server reports ready. The old one keeps serving until then, so the installation controller's existing delete of the policy lands and the upgrade continues.

Nothing here knows the name of the deprecated policy, so any other reason the new pod fails to come up also leaves a working API server in place. Already-deadlocked clusters aren't repaired. Not validated on a live cluster yet.

## Release Note

```release-note
Fixed a deadlock on upgrade where the Calico API server was moved before a deprecated policy blocking it was removed, leaving the projectcalico.org/v3 API permanently unavailable.
```

## For PR author

- [x] Tests for change.
- [x] If changing pkg/apis/, run `make gen-files` - no API changes.
- [x] If changing versions, run `make gen-versions` - no version changes.

### DIFF
--- a/pkg/controller/apiserver/cutover.go
+++ b/pkg/controller/apiserver/cutover.go
@@ -27,8 +27,7 @@ import (
 
 // holdAPIServiceCutover reports whether the projectcalico.org/v3 APIService is still served from a
 // previous location whose API server has to keep serving, because the one that would replace it is
-// not ready. Repointing early retires a working API server for one that may never come up, and the
-// fix for that then needs the v3 API that was just taken away.
+// not ready. Errors hold too, since failing to read the state is not evidence the cutover is safe.
 func holdAPIServiceCutover(ctx context.Context, reader client.Reader) (bool, error) {
 	apiService := &apiregv1.APIService{}
 	if err := reader.Get(ctx, client.ObjectKey{Name: render.APIServiceName}, apiService); err != nil {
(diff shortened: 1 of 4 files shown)


pkg/controller/apiserver/cutover.go
```diff
@@ -0,0 +1,59 @@
...
+// not ready. Repointing early retires a working API server for one that may never come up, and the
+// fix for that then needs the v3 API that was just taken away.
+func holdAPIServiceCutover(ctx context.Context, reader client.Reader) (bool, error) {
+	apiService := &apiregv1.APIService{}
+	if err := reader.Get(ctx, client.ObjectKey{Name: render.APIServiceName}, apiService); err != nil {
+		if errors.IsNotFound(err) {
+			// A fresh install. There is no API server in service to protect.
+			return false, nil
+		}
+		return false, err
```
caseydavenport: Shouldn't an unexpected error from the apiserver cause a hold on cutting over?

pkg/controller/apiserver/cutover.go
```diff
@@ -0,0 +1,59 @@
...
+		return false, nil
+	}
+
+	deployment := &appsv1.Deployment{}
+	key := client.ObjectKey{Name: render.APIServerName, Namespace: render.APIServerNamespace}
+	if err := reader.Get(ctx, key, deployment); err != nil {
+		if errors.IsNotFound(err) {
+			return true, nil
+		}
+		return false, err
```
caseydavenport: same here - unexpected read getting the dpeloyment doens't mean its ready.

caseydavenport:
closing in favor of https://github.com/tigera/operator/pull/5143

PR #5141, opened 2026-08-05
Only create the dex TLS key pair when dex is enabled

tmjd:
## Description

**Bug fix.** `utils.DexEnabled()` returns false when `Authentication.spec.oidc.type` is `Tigera`, and dex is not deployed in that case. The authentication controller nonetheless calls `GetOrCreateKeyPair()` for `tigera-dex-tls` unconditionally (`authentication_controller.go:282`), which generates a new key pair whenever the existing secret is missing, expired or otherwise unusable.

The component that persists that key pair — `rcertificatemanagement.CertificateManagement(...)` — is only appended to `components` when `enableDex` is true. So on a cluster that does not run dex, nothing ever writes the newly generated secret. The result is that **every reconcile generates and immediately discards an RSA key pair**, and an expired `tigera-dex-tls` secret is re-detected forever:

```
{"level":"info","logger":"tls","msg":"KeyPair is an expired legacy operator cert, make a new one","name":"tigera-dex-tls"}
```

This never converges: while dex is disabled nothing can replace the expired secret, so the same log line repeats on every reconcile indefinitely. The cost is wasted key generation and signing on a hot controller, plus permanent log noise.

### What changed

- `pkg/controller/authentication/authentication_controller.go` — hoist `enableDex` (it only depends on the already-fetched `Authentication`) above the point where the key pair is built, and only call `GetOrCreateKeyPair()` when dex is enabled. `CheckKeyPairWarnings()` is guarded the same way, since it dereferences the key pair.
- `pkg/render/dex.go` — `Objects()` built `c.deployment()` and `c.configMap()` before checking `DeleteDex`, and both dereference `c.cfg.TLSKeyPair`, so a nil key pair would panic on the teardown path. Those two objects are only ever *deleted* on that path, so they are now identified by name and GVK instead of being fully rendered.

Not changed: when dex *is* enabled the behaviour is identical, and this does not alter the contents of any rendered object.

### Testing

- New unit test in `pkg/render/dex_test.go` covering `DeleteDex: true` with a nil `TLSKeyPair`, asserting nothing is rendered for creation and that the Deployment and ConfigMap are in the delete list. This test panics without the `dex.go` change.
- `go test ./pkg/render` — 516 specs pass.
- `go test ./pkg/controller/authentication/...` — passes.
- `gofmt` clean; `go vet` clean on both changed packages.

### Components affected

The authentication controller and the dex renderer only. Clusters using `oidc.type` other than `Tigera` (i.e. those that actually deploy dex) are unaffected.

## Release Note

```release-note
Stop generating a new tigera-dex-tls key pair on every reconcile when dex is not deployed.
```

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/render/dex.go
+++ b/pkg/render/dex.go
@@ -130,17 +130,34 @@ func (*dexComponent) SupportedOSType() rmeta.OSType {
 
 func (c *dexComponent) Objects() ([]client.Object, []client.Object) {
 
+	// c.deployment() and c.configMap() panic on a nil TLSKeyPair, which is what dex being disabled
+	// implies. They are only deleted on that path, so name them instead of rendering them.
+	var dexDeployment, dexConfigMap client.Object
+	if c.cfg.DeleteDex {
+		dexDeployment = &appsv1.Deployment{
+			TypeMeta:   metav1.TypeMeta{Kind: "Deployment", APIVersion: "apps/v1"},
+			ObjectMeta: metav1.ObjectMeta{Name: DexObjectName, Namespace: DexNamespace},
+		}
+		dexConfigMap = &corev1.ConfigMap{
+			TypeMeta:   metav1.TypeMeta{Kind: "ConfigMap", APIVersion: "v1"},
+			ObjectMeta: metav1.ObjectMeta{Name: DexObjectName, Namespace: DexNamespace},
+		}
+	} else {
+		dexDeployment = c.deployment()
+		dexConfigMap = c.configMap()
+	}
+
 	objs := []client.Object{
 		CreateNamespace(DexObjectName, c.cfg.Installation.KubernetesProvider, PSSRestricted, c.cfg.Installation.Azure),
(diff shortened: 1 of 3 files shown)


pkg/render/dex.go
```diff
@@ -130,17 +130,35 @@ func (*dexComponent) SupportedOSType() rmeta.OSType {
 
 func (c *dexComponent) Objects() ([]client.Object, []client.Object) {
 
+	// Rendering the Deployment and the ConfigMap requires a TLS key pair, and no key pair is created
+	// when dex is disabled. Both are only ever deleted on that path, so referring to them by name is
+	// sufficient.
+	var dexDeployment, dexConfigMap client.Object
+	if c.cfg.DeleteDex {
```
tmjd, line 136: There is no need for this check. If we're going to be deleting these resources it doesn't matter if they reference a non-existent key pair.

caseydavenport:

```

```

PR #5146, opened 2026-08-06
PMREQ-821: Whisker access via Calico Ingress Gateway

vara2504:
## Description

Exposes the **Whisker UI (Calico OSS)** through Calico Ingress Gateway via `spec.ingressGateway` on the Whisker CR, following the flow #5032 added for Manager. When set, the controller renders a Gateway, HTTPRoute, Backend, ReferenceGrant and TLS secret, and reports Degraded until the Gateway is programmed (only on the Calico variant; on others it tears the gateway down).

Main changes:
- Extracts the gateway logic shared with Manager into **`pkg/uigateway`** (one `Helper.Components` call per controller); Enterprise-only objects (WAF filter SA/RoleBinding) layer on via **`pkg/enterprise/uigateway`**, so the common code carries no variant knowledge.
- **Narrows the operator's Gateway API write permissions**: the ClusterRole keeps only cluster-wide `get/list/watch`; writes are self-granted per namespace at runtime. Needs the companion chart changes: projectcalico/calico#13521 (OSS) and tigera/calico-private#13247 (Enterprise).
- The operator **owns and cleans up a `gatewayNamespace` it creates** (a pre-existing user namespace is never touched).
- Renders the **ReferenceGrant at `v1beta1`** — the only version pre-installed Gateway API bundles (OpenShift 4.19+) serve.

## Release Note

```release-note
The Whisker UI can be exposed through Calico Ingress Gateway by setting spec.ingressGateway on the Whisker resource.
```


pkg/controller/manager/manager_controller.go
```diff
@@ -789,6 +760,13 @@ func (r *ReconcileManager) Reconcile(ctx context.Context, request reconcile.Requ
 	// observed on the cluster.
 	var gatewayComponents []render.Component
 	var gatewayTLSKeyPair certificatemanagement.KeyPairInterface
+	gwHelper := &uigateway.Config{
+		Client:           r.client,
+		ResourcePrefix:   ManagerGatewayResourcePrefix,
+		TLSSecretName:    ManagerGatewayTLSSecretName,
+		BackendNamespace: helper.InstallNamespace(),
+		Enterprise:       true,
```
rene-dekker:
```suggestion
		Enterprise:       installationSpec.Variant.IsEnterprise(),
```

pkg/controller/whisker/controller.go
```diff
@@ -267,18 +279,103 @@ func (r *Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
...
+	// Resolve gateway components. Cleanup is label-driven: every Gateway
+	// carrying this component's label outside the desired namespace (or in
+	// any namespace, when spec.ingressGateway is nil) marks leftover
+	// resources to tear down.
+	gwHelper := &uigateway.Config{
+		Client:           r.cli,
+		ResourcePrefix:   whisker.GatewayResourcePrefix,
+		TLSSecretName:    whisker.GatewayTLSSecretName,
+		BackendNamespace: whisker.WhiskerNamespace,
+		Enterprise:       false,
```
rene-dekker:
```suggestion
		Enterprise:       !installationSpec.Variant.IsEnterprise(),
```

pkg/render/gateway/component.go
```diff
@@ -105,29 +129,86 @@ func (c *gatewayComponent) Objects() (objsToCreate, objsToDelete []client.Object
...
-		// RoleBinding, and NetworkPolicy are rendered here. In a custom
-		// namespace the GatewayAPI controller creates the SA and RoleBinding,
-		// and no NetworkPolicy is rendered — the same treatment user-brought
-		// Gateways get.
-		objs = append(objs,
-			rgatewayapi.GatewayNamespaceServiceAccount(c.cfg.GatewayNamespace),
-			rgatewayapi.GatewayNamespaceRoleBinding(c.cfg.GatewayNamespace),
-			c.proxyNetworkPolicy(),
-		)
+	if c.cfg.GatewayNamespace == c.cfg.BackendNamespace {
```
rene-dekker, line 135: This gate and its rationale don't line up for Manager.

`GatewayNamespace == BackendNamespace` only holds for Whisker. For Manager, `BackendNamespace` is `helper.InstallNamespace()` (`tigera-manager`) while `IngressGatewaySpec.NamespaceOrDefault()` returns `calico-system` — so with the *default* `spec.ingressGateway` (no `gatewayNamespace` set), the Envoy proxy pods land in `calico-system`, this branch is skipped, and no proxy NetworkPolicy is rendered.

`calico-system` does carry the operator-managed default-deny (`calicoSystemDefaultDenyForCalicoSystem()`, `pkg/controller/installation/core_controller.go:2253`), and `reconcileGatewayNamespaceResources` explicitly skips `gw.Namespace == common.CalicoNamespace` — so neither path covers it.

Should the condition be `gwNS == bkNS || gwNS == common.CalicoNamespace`?

(Pre-existing from #5032, but this PR rewrites the block and the comment justifying it.)

pkg/render/gateway/component.go
```diff
@@ -430,7 +511,43 @@ func (c *gatewayDeletionComponent) Objects() (objsToCreate, objsToDelete []clien
...
+	// leftovers by its label.
+	objs = append(objs, &gapi.Gateway{
+		TypeMeta:   metav1.TypeMeta{Kind: "Gateway", APIVersion: "gateway.networking.k8s.io/v1"},
+		ObjectMeta: metav1.ObjectMeta{Name: prefix + "-gateway", Namespace: gwNS},
+	})
+
+	// The Role and RoleBinding go last, after the resources they permit deleting,
+	// and in the reverse of the render order. On a move the backend namespace
+	// keeps its resources, so it keeps its grant.
+	objs = append(objs, c.roleBinding(gwNS), c.role(gwNS))
```
rene-dekker: The comment says "on a move the backend namespace keeps its resources, so it keeps its grant", but the `!move` guard only covers line 534. Line 533 unconditionally deletes `roleBinding(gwNS)`/`role(gwNS)`.

When `gwNS == bkNS` and the gateway moves *out* of the backend namespace — e.g. Whisker moving from `calico-system` to a custom namespace — this deletes the backend namespace's write grant even though the Backend and ReferenceGrant stay there.

The render component normally re-creates it later in the same reconcile, so it's self-healing in the common case. But if that Role create fails, the operator is left with no write access to a Backend it still owns, and nothing in the teardown path restores it.

pkg/uigateway/uigateway.go
```diff
@@ -0,0 +1,276 @@
...
+		return nil, err
+	}
+	if len(namespaces) == 0 {
+		return nil, nil
+	}
+	if !slices.Contains(namespaces, c.BackendNamespace) {
+		namespaces = append(namespaces, c.BackendNamespace)
+	}
+	var components []render.Component
+	for _, ns := range namespaces {
```
rene-dekker, line 251: Ordering hazard across the emitted components.

Each non-backend component's tail drops the backend-namespace Role/RoleBinding (`component.go:534`), and namespaces here are sorted with `BackendNamespace` appended last. If a labeled Gateway exists in both `bkNS` and another namespace — a partially completed move, or a hand-labelled Gateway — the first component revokes the operator's write grant in `bkNS`, and then a later component still has to delete the ReferenceGrant there. That Delete now 403s, and `Teardown` never re-creates the grant, so the reconcile is stuck degraded with no retry that can make progress.

Deleting the `bkNS` grant only from the component whose `gwNS == bkNS` (or ordering `bkNS` first) avoids it.

More generally: the self-authorization ordering constraint is currently maintained by slice position *across* a variable-length component list. Since `CreateOrUpdateOrDelete` runs all creates then all deletes for a single component (`pkg/controller/utils/component.go:487` → `:598`), collapsing this into one component with both lists populated would make the invariant structural instead.

pkg/uigateway/uigateway.go
```diff
@@ -0,0 +1,276 @@
...
+// a corresponding Gateway. This also avoids touching kinds the cluster may not
+// serve: Backend is an Envoy Gateway resource, which may be unavailable when
+// the Gateway API CRDs were installed independently. Attempting to delete an
+// unserved kind would fail the reconcile.
+func (c *Config) Teardown(ctx context.Context) ([]render.Component, error) {
+	namespaces, err := c.Namespaces(ctx)
+	if err != nil {
+		return nil, err
+	}
+	if len(namespaces) == 0 {
```
rene-dekker, line 244: Returning nothing when no labeled Gateway exists drops the old manager backstop, which was `gatewayCRDsPresent && !slices.Contains(namespaces, installNS)` → always include the install namespace.

The previous render order emitted the ReferenceGrant *before* the Gateway, so a partial render from the pre-upgrade operator can leave a ReferenceGrant/Backend in the install namespace with no labeled Gateway pointing at them. After upgrade, clearing `spec.ingressGateway` will now never clean those up.

Related to the known limitation in the description, but this is a second, separate path into it — worth a line in that section at minimum.

pkg/controller/whisker/controller.go
```diff
@@ -267,18 +279,103 @@ func (r *Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
...
+		if err != nil {
+			if kerrors.IsNotFound(err) {
+				r.status.SetDegraded(operatorv1.ResourceNotFound, "GatewayAPI CR not found; gateway resources will not be rendered", err, reqLogger)
+				return reconcile.Result{}, err
+			}
+			r.status.SetDegraded(operatorv1.ResourceReadError, msg, err, reqLogger)
+			return reconcile.Result{}, err
+		}
+
+		gatewayClassName, err := uigateway.ResolveClassName(gw, gatewayAPI)
```
rene-dekker: A gateway *configuration* error returns before `ch.CreateOrUpdateOrDelete` runs for `certComponent` and `whisker.Whisker(cfg)`, so Whisker itself never gets deployed.

On a fresh cluster where the user sets `spec.ingressGateway` up front with, say, two GatewayClasses and no `gatewayClassName`, the result is no Whisker at all rather than Whisker without a gateway. Same for a missing GatewayAPI CR or a namespace-create failure.

This mirrors Manager, but the blast radius differs: for Manager the gateway is an add-on to an already-rendered component; here it takes down the whole thing. Worth degrading and rendering Whisker without the gateway instead.

rene-dekker (commented):
Review notes, mostly around the gateway/backend namespace split. The two NetworkPolicy ones look like they'd stop a Manager gateway from working at all; the rest are teardown edge cases.

**One finding that wouldn't attach inline** (the line isn't in this PR's diff) — `pkg/render/gateway/component.go:380`, the proxy NetworkPolicy's xDS egress rule:

It targets `calico-gateway-api-controller` in `c.cfg.BackendNamespace`, but that controller only ever runs in `calico-system` (`pkg/render/gatewayapi/gateway_api.go:68`). For Whisker this is correct by coincidence — the backend namespace *is* `calico-system`. For Manager it's wrong: with `gatewayNamespace: tigera-manager` (the only configuration in which this policy renders for Manager today) the proxy is denied egress to the xDS controller on 18000/18001, so the Gateway never reaches Programmed. Should be `common.CalicoNamespace`. Pre-existing rather than introduced here, but it becomes reachable once Whisker shares this render path.

Separately (not inline, since it isn't in the diff): the delete list in `gatewayDeletionComponent.Objects()` is a hand-maintained mirror of the create path — it re-derives every name and every `move` / `gwNS != bkNS` / `Enterprise` conditional independently. The `move` finding below is a drift between the two. Discovering the delete set by Listing on `operator.tigera.io/gateway` (labelling all the rendered objects, not just Gateway/Role/RoleBinding) and feeding the found objects into `objsToDelete` would remove the class of bug rather than this instance of it.

pkg/render/gateway/component.go
vara2504: `render.ManagerNamespace` is now `common.CalicoNamespace`, with `tigera-manager` retained only as `LegacyManagerNamespace`. In single-tenant mode, `helper.InstallNamespace()` returns `calico-system`, which is also what `NamespaceOrDefault()` returns. Therefore, `gwNS == bkNS` holds with the default spec, and the proxy NetworkPolicy and WAF ServiceAccount are rendered. Multi-tenant mode cannot reach this path either, since `spec.ingressGateway` is rejected a few lines earlier.

pkg/render/gateway/component.go
vara2504: Fixed by splitting the grant by purpose into two separately named Roles: the Gateway namespace gets `Gateways`/`HTTPRoutes`, while the backend namespace gets `ReferenceGrants`/`Backends`. This avoids overlap when namespaces coincide, scopes each Role to the resources it needs, and removes the backend Role only during teardown. Upgrades also clean up the legacy combined Role.

pkg/uigateway/uigateway.go
vara2504: Fixed. The cleanup component now deletes only objects in its own namespace. `Backend` and `ReferenceGrant` are removed by the component for the backend namespace, rather than by every component. Teardown always includes that namespace, ensuring both are cleaned up.

pkg/controller/whisker/controller.go
vara2504: Per the design doc discussion, we decided to keep the halt but make the behavior explicit. The previous message incorrectly implied that only Gateway resources would not be rendered, when the reconcile actually stops and nothing is rendered. The message now reads, **“GatewayAPI CR not found; GatewayAPI is a prerequisite for spec.ingressGateway,”**

pkg/uigateway/uigateway.go
vara2504: This version creates the access Role and RoleBinding first. The behavior is essentially the same: RBAC needs to exist before the Gateway can be created.

I'm keeping the current behavior for this PR. Hitting this requires a failed reconcile and the user clearing `spec.ingressGateway` before the retry. Otherwise, the Gateway is created, cleanup finds everything by label, and re-setting the field re-adopts the resources.

Bringing back the backstop would require another cluster-wide Role/RoleBinding discovery path and additional permissions, which doesn't seem worth the complexity right now.

pkg/uigateway/uigateway.go
```diff
@@ -0,0 +1,278 @@
...
+// the Gateway API CRDs were installed independently. Attempting to delete an
+// unserved kind would fail the reconcile.
+func (c *Config) Teardown(ctx context.Context) ([]render.Component, error) {
+	namespaces, gatewayCRDsPresent, err := c.Namespaces(ctx)
+	if err != nil {
+		return nil, err
+	}
+	if !gatewayCRDsPresent || len(namespaces) == 0 {
+		return nil, nil
+	}
```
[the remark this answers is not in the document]
vara2504: This version creates the access Role and RoleBinding first. The behavior is essentially the same: RBAC needs to exist before the Gateway can be created.

I'm keeping the current behavior for this PR. Hitting this requires a failed reconcile and the user clearing spec.ingressGateway before the retry. Otherwise, the Gateway is created, cleanup finds everything by label, and re-setting the field re-adopts the resources.

Bringing back the backstop would require another cluster-wide Role/RoleBinding discovery path and additional permissions, which doesn't seem worth the complexity right now.

pkg/controller/manager/manager_controller.go
```diff
@@ -789,6 +760,13 @@ func (r *ReconcileManager) Reconcile(ctx context.Context, request reconcile.Requ
 	// observed on the cluster.
 	var gatewayComponents []render.Component
 	var gatewayTLSKeyPair certificatemanagement.KeyPairInterface
+	gwHelper := &uigateway.Config{
+		Client:           r.client,
+		ResourcePrefix:   ManagerGatewayResourcePrefix,
+		TLSSecretName:    ManagerGatewayTLSSecretName,
+		BackendNamespace: helper.InstallNamespace(),
+		Enterprise:       installationSpec.Variant.IsEnterprise(),
+	}
```
caseydavenport: This is called a "helper" but the struct is just config, which is a bit unusual.

If we want this to fit the normal pattern, we should do something like this:

```go 
gwHelper := uigateway.NewHelper(...)
```

```go
func NewHelper(...) Helper { }
```

```go
type Helper struct // Or interface{}
```

caseydavenport: This is called a "helper" but the struct is just config, which is a bit unusual.

If we want this to fit the normal pattern, we should do something like this:

```go 
gwHelper := uigateway.NewHelper(...)
```

```go
func NewHelper(...) Helper { }
```

```go
type Helper struct // Or interface{}, for testing purposes.
```

pkg/controller/manager/manager_controller.go
```diff
@@ -800,7 +778,7 @@ func (r *ReconcileManager) Reconcile(ctx context.Context, request reconcile.Requ
 			return reconcile.Result{}, nil
 		}
 	} else if instance.Spec.IngressGateway != nil {
-		gwComp, gwKeyPair, result, err := r.resolveGateway(ctx, instance, authenticationCR, certificateManager, helper, logc)
+		gwComp, gwKeyPair, result, err := r.resolveGateway(ctx, instance, installationSpec, authenticationCR, certificateManager, helper, logc)
```
caseydavenport: Many of the other funcs moved to the helper - should this one as well?

pkg/controller/uigateway/uigateway.go
```diff
@@ -0,0 +1,287 @@
...
+	"github.com/tigera/operator/pkg/controller/utils"
+	"github.com/tigera/operator/pkg/ctrlruntime"
+	"github.com/tigera/operator/pkg/render"
+	rgateway "github.com/tigera/operator/pkg/render/gateway"
+	"github.com/tigera/operator/pkg/tls/certificatemanagement"
+)
+
+// Config identifies one UI component's gateway resources.
+type Config struct {
+	Client client.Client
```
caseydavenport: This is a bit of a smell - a Config struct shouldn't contain a client, at least how we tend to structure the rest of the operator code.

pkg/controller/uigateway/uigateway.go
```diff
@@ -0,0 +1,287 @@
...
+	// TLSSecretName is the gateway listener certificate secret.
+	TLSSecretName string
+
+	// BackendNamespace holds the Backend and ReferenceGrant.
+	BackendNamespace string
+
+	// Enterprise controls whether the proxy SA and RoleBinding are part of
+	// the component's rendered set; the proxy NetworkPolicy is rendered on
+	// both variants.
+	Enterprise bool
```
caseydavenport: Probably better to pass the Variant here.

caseydavenport: Probably better to pass the Variant here instead of a bool.

pkg/controller/uigateway/uigateway.go
```diff
@@ -0,0 +1,287 @@
...
+	}
+	slices.Sort(namespaces)
+	return namespaces, nil
+}
+
+// MoveCleanup returns deletion components for every labeled Gateway outside
+// the desired namespace — leftovers of a gatewayNamespace change. The
+// Backend is kept and the ReferenceGrant handled per the move target; see
+// DeletionConfiguration.MoveTargetNamespace.
+func (c *Config) MoveCleanup(ctx context.Context, desiredNS string) ([]render.Component, error) {
```
caseydavenport: I'm a bit confused by the name of this function. It implies it's performing an action (i.e., Move) but it seems to just be generating a list of components to clean up?

Maybe `func StaleComponents() ([]render.Component, error)` instead?

pkg/uigateway/uigateway.go
```diff
@@ -0,0 +1,287 @@
...
+		if cond.Type == condType && cond.Status != metav1.ConditionTrue {
+			return fmt.Sprintf("%s: %s", msgPrefix, cond.Message)
+		}
+	}
+	return ""
+}
+
+// EnsureNamespace creates the gateway namespace if it does not exist.
+// The namespace is created without an owner reference and is never deleted by
+// the operator: a user-provided namespace may hold other workloads.
```
caseydavenport: I know this is pre-existing, but is there a reason we should never cleanup our own namespaces?

I get it if this is installing into a pre-existing Namespace, but if we're creating the Namespace then we should clean it up when we're done with it?

pkg/controller/uigateway/uigateway.go
```diff
@@ -0,0 +1,287 @@
...
+	}
+}
+
+// AddWatches registers the watches a UI gateway reconciler depends on: the
+// gateway TLS secret across all namespaces (the truth copy lives in the
+// operator namespace, the rendered copy in the user-configurable gateway
+// namespace), and the Gateway/HTTPRoute status. Gateway health lives in
+// status, which does not bump the generation, so the default generation-based
+// predicate would drop these events — match by name and accept every event
+// instead. The Gateway/HTTPRoute watch arms once the Gateway API CRDs exist.
```
caseydavenport: I think this (and several others) comment reads as AI over-commenting to me. Most of this is self-explanatory from reading the code and will just get out of date / bloat the file.

pkg/controller/whisker/controller.go
```diff
@@ -107,6 +111,14 @@ func Add(mgr manager.Manager, opts options.ControllerOptions) error {
 		return fmt.Errorf("whisker-controller failed to watch ClusterInformation")
 	}
 
+	if err = c.WatchObject(&operatorv1.GatewayAPI{}, &handler.EnqueueRequestForObject{}); err != nil {
+		return fmt.Errorf("%s failed to watch GatewayAPI resource: %w", controllerName, err)
+	}
+
+	if err = uigateway.AddWatches(c, opts.K8sClientset, log, whisker.GatewayResourcePrefix, whisker.GatewayTLSSecretName); err != nil {
+		return fmt.Errorf("%s failed to add gateway watches: %w", controllerName, err)
+	}
```
caseydavenport, line 118: We're in the process of making the code have a clear split between OSS and enterprise library functions - looks like uigateway is a new package that contains both.

We should think about how we can make a clear split, which will be relevant when moving this code to the monorepo in a few weeks. 

I think it's mostly common? With the key bits being the variant switches - can we reasonably have a `pkg/uigateway` and `pkg/enterprise/uigateway` that is built as a layer on top?

pkg/controller/whisker/controller.go
```diff
@@ -267,18 +279,107 @@ func (r *Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
 		cfg.ClusterID = clusterInfo.Spec.ClusterGUID
 	}
 
+	gwHelper := &uigateway.Config{
+		Client:           r.cli,
+		ResourcePrefix:   whisker.GatewayResourcePrefix,
+		TLSSecretName:    whisker.GatewayTLSSecretName,
+		BackendNamespace: whisker.WhiskerNamespace,
+		// false on Calico, where the gateway is rendered; the WAF ServiceAccount
+		// is Enterprise-only and Whisker's gateway never creates it.
```
caseydavenport: AI overcommenting.

pkg/controller/whisker/controller.go
```diff
@@ -267,18 +279,107 @@ func (r *Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
...
+		Enterprise: installationSpec.Variant.IsEnterprise(),
```
caseydavenport: This defualts to false.

pkg/controller/whisker/controller.go
```diff
@@ -267,18 +279,107 @@ func (r *Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
...
-			rcertificatemanagement.NewKeyPairOption(whiskerKeyPair, true, true),
-			rcertificatemanagement.NewKeyPairOption(backendKeyPair, true, true),
-		},
-		TrustedBundle: trustedBundle,
+		KeyPairOptions:  keyPairOptions,
+		TrustedBundle:   trustedBundle,
 	})
 
 	components := []render.Component{certComponent, whisker.Whisker(cfg)}
+	components = append(components, gatewayComponents...)
```
caseydavenport, line 371: Do we need the intermediate gatewayComponents? Can't we just define `components` and add the gateway components directly to it as needed? Or is there some ordering dependency this is solving?

pkg/controller/whisker/controller.go
```diff
@@ -267,18 +279,107 @@ func (r *Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
...
+			r.status.SetDegraded(operatorv1.InvalidConfigurationError, "Failed to resolve gateway class", err, reqLogger)
+			return reconcile.Result{}, err
+		}
+
+		gwNS := gw.NamespaceOrDefault()
+		if gwNS != whisker.WhiskerNamespace {
+			if err := uigateway.EnsureNamespace(ctx, r.cli, gwNS); err != nil {
+				r.status.SetDegraded(operatorv1.ResourceCreateError, fmt.Sprintf("Failed to create gateway namespace %q", gwNS), err, reqLogger)
+				return reconcile.Result{}, err
+			}
```
caseydavenport: Feels like we have a halfway split between the new gwHelper and package level functions - what defines when one belongs as a package level function vs. a helper method?

It feels like it would be nice if we could move a lot of this boiler plate (gateway class name determination, namespace creation, TLS configuration, cleanup) within the helper so that any controller that wants to integrate with a gateway gets a simple, single entrypoint instead of needing to duplicate a lot of this.

WDYT? Could we extend the helper? What else would it need?

```go
// TODO: Extend args with the necessary bits to fully encapsulate 
// within the helper.
helper := gw.NewHelper(...)

components = append(components, helper.Components()...)
```

pkg/render/gateway/component.go
```diff
@@ -105,27 +130,109 @@ func (c *gatewayComponent) Objects() (objsToCreate, objsToDelete []client.Object
...
+		if c.cfg.Enterprise {
+			// The WAF HTTP filter's ServiceAccount and the RoleBinding giving
+			// it Gateway API reads. The proxy pod runs as that account on
+			// Enterprise. In a custom namespace the GatewayAPI controller
+			// creates them instead.
+			objs = append(objs,
+				rgatewayapi.GatewayNamespaceServiceAccount(c.cfg.GatewayNamespace),
+				rgatewayapi.GatewayNamespaceRoleBinding(c.cfg.GatewayNamespace),
+			)
+		}
```
caseydavenport: Making a note for myself that we'll need to port this to the enterprise `Extensions` mechanism.

pkg/render/gateway/component.go
```diff
@@ -105,27 +130,109 @@ func (c *gatewayComponent) Objects() (objsToCreate, objsToDelete []client.Object
...
+	}
+
+	// Before the grants were split by purpose, one Role named for the gateway
+	// covered both namespaces. Where the namespaces differ that object is no
+	// longer rendered in the backend namespace, so remove the leftover; it
+	// carried write verbs on kinds that namespace never holds. Can go once no
+	// supported upgrade path starts before the split.
```
caseydavenport: Please call out explicitly what version this will be - we should be able to predict it, and it makes it easy to remove in the future.

Ideally we also raise a ticket to track.

pkg/render/gateway/component.go
```diff
@@ -105,27 +130,109 @@ func (c *gatewayComponent) Objects() (objsToCreate, objsToDelete []client.Object
...
+		legacyRole, legacyBinding := c.access(c.cfg.ResourcePrefix+gatewayAccessSuffix, c.cfg.BackendNamespace, nil)
+		objsToRemove = append(objsToRemove, legacyBinding, legacyRole)
 	}
 
-	return objs, nil
+	return objs, objsToRemove
+}
+
+// The Gateway and HTTPRoute live in the gateway namespace, the Backend and
+// ReferenceGrant beside the backing Service.
```
caseydavenport: I don't understand why this comment is above these lines.

pkg/render/gateway/component.go
```diff
@@ -105,27 +130,109 @@ func (c *gatewayComponent) Objects() (objsToCreate, objsToDelete []client.Object
...
+const (
+	gatewayAccessSuffix = "-ingressgateway-access"
+	backendAccessSuffix = "-ingressgateway-backend-access"
+)
+
+// gatewayAccess grants the writes needed in the gateway namespace; the
```
caseydavenport:
```suggestion
// gatewayAccess grants the operator the write permissions needed in the gateway namespace; the
```

pkg/render/gateway/component.go
```diff
@@ -343,18 +456,17 @@ func (c *gatewayComponent) proxyNetworkPolicy() *v3.NetworkPolicy {
 // DeletionConfiguration is the minimal config needed to identify gateway
 // resources for cleanup. No TLS keypair, hostname, or class is required.
 type DeletionConfiguration struct {
-	ResourcePrefix   string
-	GatewayNamespace string
+	ResourcePrefix string
+	// StaleNamespace is the namespace being cleaned up.
```
caseydavenport:
```suggestion
```
Claude loves to throw obvious comments in above new struct fields even when no other field on the struct has them...

pkg/render/gateway/component.go
```diff
@@ -383,19 +495,20 @@ func (c *gatewayDeletionComponent) Objects() (objsToCreate, objsToDelete []clien
 	objs := []client.Object{
 		&corev1.Secret{
```
caseydavenport, line 490: Mostly just curious... do you know why we have a gateway deletion component instead of just using the gateway component and inverting create / delete?

caseydavenport, line 490: Mostly just curious... do you know why we have a gateway deletion component instead of just using the gateway component and inverting create / delete? Seems like we need to be careful to keep these in sync?

pkg/render/gateway/component.go
```diff
@@ -413,24 +528,61 @@ func (c *gatewayDeletionComponent) Objects() (objsToCreate, objsToDelete []clien
...
+			objs = append(objs,
+				rgatewayapi.GatewayNamespaceServiceAccount(staleNS),
+				rgatewayapi.GatewayNamespaceRoleBinding(staleNS),
+			)
+		}
+	}
+
+	// The Gateway goes after the resources found through it, mirroring the render.
+	// If an earlier delete fails, it stays and the next reconcile still finds the
+	// leftovers by its label.
```
caseydavenport, line 533: I think a finalizer is the correct way to do this? Relying on ordering can be finicky / error-prone, and finalizers are the k8s native way to say "keep this around until I am done with it".

Let's not do it in this PR, but perhaps as a follow-on to keep things tidy?

pkg/render/whisker/component.go
```diff
@@ -296,6 +306,22 @@ func (c *Component) networkPolicy() *v3.NetworkPolicy {
...
 	egressRules = networkpolicy.AppendDNSEgressRules(egressRules, c.cfg.OpenShift)
 
+	var ingressRules []v3.Rule
+	if c.cfg.IngressGatewayNamespace != "" {
+		// Envoy Gateway labels its proxy pods with the owning Gateway name.
+		ingressRules = append(ingressRules, v3.Rule{
+			Action:   v3.Allow,
+			Protocol: &networkpolicy.TCPProtocol,
+			Source: v3.EntityRule{
+				NamespaceSelector: fmt.Sprintf("%s == '%s'", selector.CalicoNameLabel, c.cfg.IngressGatewayNamespace),
```
caseydavenport, line 310: note to self: we should rename the CalicoNameLabel

caseydavenport (changes requested):

pkg/controller/manager/manager_controller.go
vara2504: Done

pkg/controller/whisker/controller.go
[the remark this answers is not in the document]
vara2504: Resolved by removing the flag rather than fixing its derivation: the render no longer carries any variant knowledge. Callers supply the variant's objects through `Config.ExtraProxyObjects` — Manager passes the WAF ServiceAccount/RoleBinding built by `pkg/enterprise/uigateway`, Whisker passes nothing — and the deletion path deletes only what its own config carries. The scenario described (a Whisker teardown on an Enterprise cluster deleting shared WAF resources it never created) is now structurally impossible: Whisker's config cannot name those objects.

pkg/controller/manager/manager_controller.go
vara2504: Done — `Config` is now data only and `Helper` owns the client via `NewHelper(client, Config)`, the same split the Reconciler structs use.

pkg/controller/manager/manager_controller.go
vara2504: resolveGateway is Manager-specific: the OIDC managerDomain check needs the Authentication CR, which Whisker doesn't have, and the listener keypair has to flow back to the controller for the certificate component.

pkg/controller/uigateway/uigateway.go
vara2504: Fixed — the client moved to `NewHelper`; `Config` holds only data, matching how the reconcilers hold `cli` themselves.

pkg/controller/uigateway/uigateway.go
vara2504: Went one further: neither is passed now. The common code carries no variant knowledge — callers supply `Config.ExtraProxyObjects` (Manager passes what `pkg/enterprise/uigateway` builds, Whisker passes nil), so there's no flag left to derive or default.

pkg/controller/uigateway/uigateway.go
vara2504: Renamed to `StaleComponents` — it builds and returns the list, as suggested.

pkg/uigateway/uigateway.go
vara2504: Agreed and done: a namespace the operator creates is stamped with the gateway label, and teardown deletes it; a namespace the user already had is never labeled and never touched. Verified live — an operator-created namespace is removed on teardown, a pre-existing one survives with its labels intact.

pkg/controller/uigateway/uigateway.go
vara2504: Culled across the PR — comments now state only what the code can't (a why, a precondition), and duplicated rules were reduced to one statement at the site that enforces them.

pkg/controller/whisker/controller.go
vara2504: Done, in exactly that layout: `pkg/uigateway` is the variant-free common code; `pkg/enterprise/uigateway` layers the Enterprise proxy objects on top via `Config.ExtraProxyObjects`.

pkg/controller/whisker/controller.go
vara2504: Removed.

pkg/controller/whisker/controller.go
vara2504: The flag is gone entirely — the common code carries no variant knowledge; callers supply `Config.ExtraProxyObjects` instead, so there's nothing left to default.

pkg/controller/whisker/controller.go
vara2504: Yes — ordering: the gateway components are applied after the certificate and Whisker components, so the slice is held apart and appended at the end. Defining `components` directly would interleave them.

pkg/controller/whisker/controller.go
vara2504: Resolved by the fold-in: everything reconcile-shaped is a `Helper` method behind `Components()`; the only package-level function left is `AddWatches`, which runs at controller setup before a helper exists.

pkg/render/gateway/component.go
vara2504: The cleanup was dead code and is gone, along with its test: no shipped branch (release-calient-v3.24-1/-2, release-v1.43-2, release-v1.44) ever rendered the combined Role, so no upgrade path can carry it.

pkg/render/gateway/component.go
vara2504: Fixed — moved onto the declarations it describes.

pkg/render/gateway/component.go
vara2504: Applied verbatim.

pkg/render/gateway/component.go
vara2504: Deleted.

pkg/render/gateway/component.go
vara2504: Inverting would work only by faking the config: the create path needs a hostname, class, and TLS keypair to build its objects, and all of those come from `spec.ingressGateway`, which is nil during teardown. The delete set is also not a mirror of the create set — teardown covers every labelled namespace, keeps the Backend on a move, and must not touch a custom namespace's SA/RoleBinding, so those rules would become delete-only branches inside a shared component anyway. To keep the two paths honest, a test now asserts every rendered object has a matching delete, so a missed mirror fails CI rather than leaking.

pkg/render/gateway/component.go
vara2504: Agreed as a follow-up.

vara2504:
@caseydavenport Addressed the review comments. Should be good for another look when you have a chance.

pkg/uigateway/uigateway.go
```diff
@@ -0,0 +1,376 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package uigateway carries the controller-side logic shared by the UI
+// components (Manager, Whisker) that expose themselves through Calico
+// Ingress Gateway: label-driven cleanup, gateway health read-back, namespace
+// provisioning, class resolution, and watch setup. The rendering lives in
+// pkg/render/gateway; this package holds what a reconciler needs around it.
+package uigateway
```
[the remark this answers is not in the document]
vara2504: Aligned — the description now says `pkg/uigateway` (with `pkg/enterprise/uigateway` layering the Enterprise objects on top).

pkg/controller/whisker/controller.go
```diff
@@ -275,18 +284,62 @@ func (r *Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
...
+		if err != nil {
+			r.status.SetDegraded(operatorv1.CertificateError, "Error getting or creating gateway TLS certificate", err, reqLogger)
+			return reconcile.Result{}, err
+		}
+
+		gatewayComponents, err = gwHelper.Components(ctx, gw, gatewayTLSKeyPair)
+		if err != nil {
+			r.status.SetDegraded(operatorv1.ResourceCreateError, "Failed to render gateway resources", err, reqLogger)
+			return reconcile.Result{}, err
+		}
```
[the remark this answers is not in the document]
vara2504: ResourceCreateError is valid: whatever the cause, the failure blocks creating the gateway resources, and the error chain carries the specific cause into the status message ("Failed to
  render gateway resources: GatewayAPI CR not found; …")

pkg/enterprise/controller/manager/manager_controller.go
```diff
@@ -1077,122 +969,27 @@ func (r *ReconcileManager) resolveGateway(
...
-		if msg := unhealthyCondition(ps.Conditions, string(gapi.RouteConditionAccepted), "HTTPRoute not accepted"); msg != "" {
-			return msg
-		}
-		if msg := unhealthyCondition(ps.Conditions, string(gapi.RouteConditionResolvedRefs), "HTTPRoute refs not resolved"); msg != "" {
-			return msg
-		}
+	comps, err := gwHelper.Components(ctx, gw, gwTLSKeyPair)
+	if err != nil {
+		r.status.SetDegraded(operatorv1.ResourceCreateError, "Failed to render gateway resources", err, logc)
+		return nil, nil, err
```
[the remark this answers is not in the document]
vara2504: ResourceCreateError is accurate: whatever the cause, the failure blocks creating the gateway resources, and the error chain carries the specific cause into the status message ("Failed to
  render gateway resources: GatewayAPI CR not found; …")

pkg/uigateway/uigateway.go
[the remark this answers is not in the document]
vara2504: Updated

pkg/uigateway/uigateway.go
```diff
@@ -0,0 +1,369 @@
...
+	if err == nil || !errors.IsNotFound(err) {
+		return err
+	}
+	ns := render.CreateNamespace(name, h.cfg.Provider, render.PSSPrivileged, nil)
+	// The gateway label marks the namespace as ours, so teardown can tell a
+	// namespace the operator created from one the user already had.
+	ns.Labels[rgateway.GatewayLabel] = h.cfg.ResourcePrefix
+	if err := h.cli.Create(ctx, ns); err != nil && !errors.IsAlreadyExists(err) {
+		return err
+	}
```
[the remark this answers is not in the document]
vara2504: Good catch — fixed. Installation.Azure is now threaded through the helper Config and passed to CreateNamespace, so on AKS the gateway namespace gets the same Azure-policy labels as other operator-created namespaces (the nil only diverged when Azure PolicyMode was non-default). Added a UT for the AKS label path alongside the OpenShift one.

pkg/render/gateway/component.go
```diff
@@ -374,28 +457,27 @@ func (c *gatewayDeletionComponent) SupportedOSType() rmeta.OSType              {
...
 		&gapi.HTTPRoute{
 			TypeMeta:   metav1.TypeMeta{Kind: "HTTPRoute", APIVersion: "gateway.networking.k8s.io/v1"},
-			ObjectMeta: metav1.ObjectMeta{Name: prefix + "-route", Namespace: gwNS},
+			ObjectMeta: metav1.ObjectMeta{Name: prefix + "-route", Namespace: staleNS},
 		},
 	}
 
-	if !move {
+	// The Backend and ReferenceGrant live in the backend namespace, so only the
+	// component cleaning that namespace deletes them. A component for another
```
caseydavenport: I am not sure about this comment - we don't do per-component RBAC here - both components are going to run with the operator RBAC, right?

pkg/render/gateway/component.go
```diff
@@ -60,14 +62,14 @@ type Configuration struct {
...
-	// user-brought Gateways.
-	Enterprise bool
+	// RouteRequestTimeout, when set, becomes the HTTPRoute rule's request
+	// timeout; nil keeps the Envoy Gateway default.
+	RouteRequestTimeout *string
+
+	// ExtraProxyObjects are variant-specific objects rendered beside the
+	// proxy, only when the Gateway shares the backend namespace — elsewhere
+	// the GatewayAPI controller provisions per-namespace resources itself.
+	ExtraProxyObjects []client.Object
```
caseydavenport, line 79: Hm, is this the right approach? I think we should be registering an Extension for this and have the extension add the extra objects?

pkg/render/whisker/component.go
```diff
@@ -292,6 +302,22 @@ func (c *Component) networkPolicy() *v3.NetworkPolicy {
...
+				Selector:          fmt.Sprintf("gateway.envoyproxy.io/owning-gateway-name == '%s'", GatewayResourcePrefix+"-gateway"),
```
caseydavenport: This `Prefix + "-gateway"` string seems brittle - I think a utils function would help ensure we don't accidentally let these drift / remove typo errors.

pkg/render/gateway/component.go
```diff
@@ -250,24 +331,26 @@ func (c *gatewayComponent) backend() *envoyapi.Backend {
 	}
 }
 
-func (c *gatewayComponent) referenceGrant() *gapi.ReferenceGrant {
+// The ReferenceGrant is v1beta1: the standard has not promoted it to v1, so a
+// cluster serving pre-installed Gateway API CRDs (OpenShift 4.19+) has no v1.
```
caseydavenport:
```suggestion
// ReferenceGrant is written as v1beta1, the only version served by every
// Gateway API bundle we may find pre-installed (OpenShift 4.19 ships v1.2.1,
// and CRDManagementPreferExisting leaves it alone). v1beta1 is still the
// storage version as of Gateway API v1.6.
```

caseydavenport (approved):
A few minor questions / suggestions, but otherwise LGTM.

pkg/render/gateway/component.go
vara2504: Reworded the comment so it no longer implies per-component RBAC.

pkg/render/gateway/component.go
vara2504: Doing it as a follow-up.

pkg/render/whisker/component.go
vara2504: Done.

pkg/render/gateway/component.go
vara2504: Done.

pkg/uigateway/uigateway.go
```diff
@@ -0,0 +1,380 @@
...
+	go utils.WaitToAddResourceWatch(c, k8sClientset, log, nil, []client.Object{
+		&gapi.Gateway{
+			TypeMeta:   metav1.TypeMeta{Kind: "Gateway", APIVersion: "gateway.networking.k8s.io/v1"},
+			ObjectMeta: metav1.ObjectMeta{Name: resourcePrefix + "-gateway"},
+		},
+		&gapi.HTTPRoute{
+			TypeMeta:   metav1.TypeMeta{Kind: "HTTPRoute", APIVersion: "gateway.networking.k8s.io/v1"},
+			ObjectMeta: metav1.ObjectMeta{Name: resourcePrefix + "-route"},
+		},
+	}, gatewayWatchPredicate)
```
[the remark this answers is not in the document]
vara2504: Fixed.

PR #5163, opened 2026-08-12
Add a startup probe to the calico-node DaemonSet

caseydavenport:
## Description

Type of change: enhancement

calico-node typically becomes ready within a couple of seconds of container start, but nothing notices until the next readiness poll. During a rolling update that latency is paid once per node, so on a large cluster it dominates how long the DaemonSet takes to roll.

This adds a startup probe to calico-node running the same health check readiness already runs, polling every 2s with a 300s total budget. Two effects:

- Pods are marked ready within ~2s of actually being ready instead of on the next readiness poll.
- Kubernetes suspends the liveness probe until the startup probe passes, so a slow start on a large cluster can no longer be restarted part way through. Previously the liveness probe was live from container start and would restart the container after ~3 minutes.

The tradeoff is that a container hung at startup now takes up to 300s to be restarted rather than ~3 minutes. That felt like the right direction, since the restart is unlikely to help and the 300s budget is only ever paid when startup is actually failing.

Readiness and liveness are unchanged. Windows nodes are not changed.

Note that the operator's probe defaulting only fills in liveness and readiness probes, so every startup probe field is set explicitly here - the Kubernetes defaults would give a 30s budget, which would restart calico-node on any large cluster.

The startup probe's timing is overridable through the calico-node DaemonSet overrides on the Installation resource, alongside the existing readiness and liveness overrides. The probe handler stays operator-owned.

TigeraStatus also accounts for the new probe: the window it waits before reporting a pod as running but not ready now includes the startup budget, since Kubernetes holds off readiness until the startup probe passes. Without that, every node would be reported as failing part way through a slow start.

Affected components: calico-node DaemonSet, TigeraStatus reporting.

## Release Note

```release-note
Reduces the time calico-node takes to be marked ready after starting, which shortens calico-node rolling updates on large clusters.
```

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:


pkg/render/node.go
```diff
@@ -1695,8 +1696,8 @@ func (c *nodeComponent) nodeLifecycle() *corev1.Lifecycle {
 	}
 }
 
-// nodeLivenessReadinessProbes creates the node's liveness and readiness probes.
-func (c *nodeComponent) nodeLivenessReadinessProbes() (*corev1.Probe, *corev1.Probe) {
+// nodeProbes creates the node's probes. The startup probe stops a slow Felix start from tripping liveness.
```
tmjd: The "The startup probe stops a slow Felix start from tripping liveness." feels like an AI added comment that doesn't feel appropriate as a comment on the function. Maybe it should be moved to the definition of the SP.
I would like to see a comment or named return variables to make it clear which probe is which returned by the function.

caseydavenport:
```suggestion
// nodeProbes creates calico/node health probes. It returns in order: startupProbe, livenessProbe, readinessProbe
```

tmjd (approved):

PR #5170, opened 2026-08-13
Gate the remaining Enterprise-only code behind the variant extensions

caseydavenport:
## Description

Prep for splitting the operator across the OSS and Enterprise monorepos. Core code should never name an Enterprise-only resource, and a handful of places still did.

- The certificate signing controller now takes its signable certificates, its watches, and its signing-role decision from the variant, rather than hard-coding the Prometheus server certificate and reading the monitoring resource itself.
- Dex and Prometheus client certificate watches move into the Enterprise API server hook; the Prometheus and packet capture certificates come out of the cluster connection controller's watches and trusted bundle the same way.
- Elasticsearch client helpers move out of the shared controller utilities into the log storage tree, where the only callers live.
- The startup check that refuses a contradictory internal/external Elasticsearch configuration becomes an Enterprise startup hook, replacing two near-identical copies.
- The Enterprise custom resource getters move out of the shared controller utilities into their own package, along with the conversion from cloud configuration to a tenant.
- The kube-controllers render configuration drops the Enterprise resources only the Elasticsearch assembler reads, and the tiers controller stops deciding for itself which namespaces need DNS access when the variant is Enterprise.
- The tenant secrets controller moves into a package of its own. It only runs in multi-tenant mode, so the OSS side deletes it wholesale.
- A new test fails when anything outside the Enterprise packages imports an Enterprise-only renderer, or names one of the fifteen operator API kinds that only Enterprise installs. Only the operator's main entry point is still on an exception list.

One behavior change worth flagging: with no monitoring resource present, the signing-role decision now falls through to the non-cluster host check instead of returning early. The non-cluster host check was being skipped in that case.

```release-note
None
```


cmd/main.go
```diff
@@ -615,6 +607,9 @@ admission policy installation; once an Installation exists it is the authority o
 		Cloud:       isCloudBuild(),
 		ManageCRDs:  manageCRDs,
 		UseV3CRDs:   v3CRDs,
+
```
caseydavenport, line 610:
```suggestion
```

cmd/main.go
```diff
@@ -589,12 +587,6 @@ admission policy installation; once an Installation exists it is the authority o
 
 	if isCloudBuild() {
```
caseydavenport, line 588: This entire branch, plus the elasticIsMigrating / useExternalElastic are eventually going to need to be inputs to something like `operator.New(...)` instead of living here. The code in `New()` will need to be variant agnostic as well.

pkg/controller/apiserver/apiserver_controller.go
```diff
@@ -345,15 +330,12 @@ func (r *ReconcileAPIServer) Reconcile(ctx context.Context, request reconcile.Re
...
-	var managementCluster *operatorv1.ManagementCluster
-	if r.opts.Variant.IsEnterprise() {
-		managementCluster, err = utils.GetManagementCluster(ctx, r.client)
-		if err != nil {
-			r.status.SetDegraded(operatorv1.ResourceReadError, "Error reading ManagementCluster", err, reqLogger)
-			return reconcile.Result{}, err
-		}
+	// The webhooks component (v3-CRD mode) needs the management-cluster setup to
+	// register the managed-cluster webhook.
+	managementCluster, err := r.ext.ManagementCluster(ctx, r.client)
```
caseydavenport: This is wrong - the extension interface shouldn't have a ManagementCluster() call on it, nor should the OSS controller know anything about ManagementClusters - this is a violation of the break between OSS and enterprise that we are trying to build.

This needs to be plumbed through the controller extension mechanism.

pkg/controller/certificatemanager/certificatemanager.go
```diff
@@ -138,9 +138,10 @@ func WithLogger(log logr.Logger) Option {
 	}
 }
 
-func WithTenant(t *operatorv1.Tenant) Option {
+// WithMultiTenant tells the manager to use the per-tenant CA secret.
+func WithMultiTenant(multiTenant bool) Option {
```
caseydavenport, line 142: Is this change necessary?

pkg/controller/clusterconnection/clusterconnection_controller.go
```diff
@@ -278,9 +275,10 @@ func (r *ReconcileConnection) Reconcile(ctx context.Context, request reconcile.R
 
 	includeSystem := managementClusterConnection.Spec.TLS.CA == operatorv1.CATypePublic
 
+	bundleSecrets := append([]string{render.CalicoAPIServerTLSSecretName, goldmane.GoldmaneKeyPairSecret},
```
caseydavenport: No need for this to be split across lines, and if it is we should do each argument on its own line, and closing paren on its own line.

pkg/controller/csr/csr_controller.go
```diff
@@ -194,25 +184,12 @@ func (r *reconcileCSR) Reconcile(ctx context.Context, request reconcile.Request)
...
 	needsCSRRole := instance.Spec.CertificateManagement != nil
-	if !needsCSRRole && r.variant.IsEnterprise() {
-		monitorCR := &operatorv1.Monitor{}
-		if err := r.client.Get(ctx, utils.DefaultEnterpriseInstanceKey, monitorCR); err != nil {
-			if apierrors.IsNotFound(err) {
-				return reconcile.Result{}, nil
-			}
+	if !needsCSRRole {
+		var err error
+		needsCSRRole, err = r.extensions.CSR().NeedsCSRRole(ctx, r.client)
```
caseydavenport: This PR is doing too much of this - adding single-use purpose built methods to the extensions interface, which is really defeating the purpose.

We should be aiming to do this all via the existing Extensions interface with minimal changes. Why can't we use ExtendInputs for this logic?

pkg/controller/installation/core_controller.go
```diff
@@ -1213,44 +1213,42 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 		}))
 
 	// Check if non-cluster host feature is enabled.
-	var nonclusterhost *operatorv1.NonClusterHost
-	if instance.Spec.Variant.IsEnterprise() {
-		nonclusterhost, err = utils.GetNonClusterHost(ctx, r.client)
+	nonclusterhost, err := r.ext.NonClusterHostEnabled(ctx, r.client)
```
caseydavenport, line 1216: same here - the base controller shouldn't know anything about non cluster hosts.

pkg/controller/intrusiondetection/intrusiondetection_controller.go
```diff
@@ -437,14 +439,14 @@ func (r *ReconcileIntrusionDetection) Reconcile(ctx context.Context, request rec
...
 		}
-		tenant = cloudConfig.ToTenant()
+		tenant = eutils.TenantFromCloudConfig(cloudConfig)
 	}
 
 	// Create a component handler to manage the rendered component.
 	handler := utils.NewComponentHandler(log, r.client, r.scheme, instance)
 
 	// Determine the namespaces to which we must bind the cluster role.
-	namespaces, err := helper.TenantNamespaces(r.client)
```
caseydavenport, line 447: Why change the signature on this? Feels unrelated.

pkg/controller/manager/manager_controller_cloud.go
```diff
@@ -122,10 +123,10 @@ func (r *ReconcileManager) handleCloudReconcile(
 				r.status.SetDegraded(operatorv1.ResourceReadError, "Failed to read cloud config", err, reqLogger)
 				return nil, render.ManagerCloudResources{}, nil, nil, err
 			}
-			tenant = cloudConfig.ToTenant()
+			tenant = eutils.TenantFromCloudConfig(cloudConfig)
```
caseydavenport, line 126: I think this is actually a nicer signature, unlikely some of the other signature changes - shows that we're talking about Elasticsearch config specifically.

pkg/controller/utils/cloudconfig.go
```diff
@@ -16,22 +16,13 @@ package utils
...
 	"github.com/tigera/operator/pkg/render/common/cloudconfig"
 	corev1 "k8s.io/api/core/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-// CloudAuthConfig is the name of the ConfigMap holding cloud authentication configuration
-// (e.g. the tenantID) for a single-tenant cloud management cluster backed by internal Elasticsearch.
-const CloudAuthConfig = "cloud-auth-config"
-
 // GetCloudConfig retrieves the config map containing the configuration values needed to set up communications with
```
caseydavenport: I think this whole file belongs in pkg/enterprise/utils.

pkg/enterprise/utils/namespace_helper.go
```diff
@@ -52,14 +44,6 @@ type NamespaceHelper interface {
...
-	// TenantNamespaces returns all namespaces in the cluster for this component, across all tenants. This is useful when
-	// binding global resources to potentially several different Tenant namespaces.
-	// For single-tenant clusters, this simply returns the InstallNamespace.
-	TenantNamespaces(client.Client) ([]string, error)
-
-	// FilteredTenantNamespaces returns all namespaces for all Tenants that match the given filter.
-	FilteredTenantNamespaces(client.Client, TenantFilter) ([]string, error)
-
 	// Returns whether or not this is a multi-tenant helper.
 	MultiTenant() bool
```
caseydavenport, line 48: MultiTenant is an enterprise only concept.

pkg/enterprise/options/options.go
```diff
@@ -28,4 +28,9 @@ type Options struct {
 	// installation hook watches the CRDs the variant adds.
 	ManageCRDs bool
 	UseV3CRDs  bool
+
+	// ExternalElastic and ESMigration mirror the operator's Elasticsearch mode. The
+	// startup hook rejects a mode the cluster's certificates contradict.
```
caseydavenport, line 33: Bad comment. Explain what these mean, not "they mirror" something....

pkg/enterprise/policysync/policysync.go
```diff
@@ -0,0 +1,45 @@
...
+
+package policysync
+
+import (
+	operatorv1 "github.com/tigera/operator/api/v1"
+)
+
+// ApplicationLayerRequires reports whether any enabled ApplicationLayer feature needs
+// policySyncPathPrefix set. A nil CR returns false.
+func ApplicationLayerRequires(al *operatorv1.ApplicationLayer) bool {
```
caseydavenport: An entire package for this one function? Seems excessive.

pkg/enterprise/boundary_test.go
```diff
@@ -0,0 +1,269 @@
...
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+const (
+	modulePath = "github.com/tigera/operator/"
+	apiPath    = "github.com/tigera/operator/api/v1"
+)
+
+// enterpriseRender are the render packages that only Calico Enterprise installs. Core
```
caseydavenport, line 35: Needs a top-level comment explaining this is a temporary file to prevent re-addition of enterprise code into the core operator code during the window between now and monorepo.

pkg/enterprise/register.go
```diff
@@ -36,6 +40,10 @@ func New(variant operatorv1.ProductVariant, o eoptions.Options) extensions.Exten
 			Windows:           windows.New(variant),
 			APIServer:         apiserver.New(variant, o),
 			ClusterConnection: clusterconnection.New(variant),
+			CSR:               csr.New(),
```
caseydavenport, line 43: We don't need to tell these about the variant? What about when we're running as OSS Calico with an enterprise operator?

pkg/extensions/apiserver.go
```diff
@@ -28,6 +30,10 @@ type APIServerExtension interface {
 	ExtendInputs(ctx context.Context, ci controller.Inputs) (controller.Inputs, []certificatemanagement.KeyPairInterface, error)
 	Watches(c ctrlruntime.Controller) error
 
+	// ManagementCluster returns the management-cluster setup the webhooks register
+	// against, or nil when the cluster does not manage others.
+	ManagementCluster(ctx context.Context, c client.Client) (*render.ManagementCluster, error)
```
caseydavenport: Yeah - again this is quite wrong. The extensions interfaces shouldn't know anything about the enterprise resources / implementation. They are meant to be agnostic. 

The intention is that ExtendInputs handles additional logic.

pkg/extensions/clusterconnection.go
```diff
@@ -38,6 +38,9 @@ type ClusterConnectionExtension interface {
 	// wait on a CRD.
 	Watches(c ctrlruntime.Controller, cs kubernetes.Interface) error
 
+	// TrustedBundleSecrets are the certificates Guardian must trust for the variant.
+	TrustedBundleSecrets() []string
```
caseydavenport: Can ExtendInputs not handle this?

pkg/extensions/clusterconnection.go
caseydavenport, line 36:
```suggestion
	ValidateAndDefault(cr *operatorv1.ManagementClusterConnection) error
	
```

pkg/extensions/installation.go
```diff
@@ -42,6 +43,10 @@ type InstallationExtension interface {
 	// it changed fc. It runs before Felix defaulting persists.
 	DefaultFelixConfiguration(install *operatorv1.InstallationSpec, fc *v3.FelixConfiguration) (bool, error)
 
+	// NonClusterHostEnabled reports whether the variant serves hosts outside the
+	// cluster, which Typha renders a second deployment for.
+	NonClusterHostEnabled(ctx context.Context, c client.Client) (bool, error)
```
caseydavenport, line 48: Yep - this is an enterprise concept and shouldn't appear on the generic extensions interface.

pkg/enterprise/cloudconfig/cloudconfig.go
```diff
@@ -79,27 +78,6 @@ type CloudConfig struct {
 	enableMTLS           bool
```
caseydavenport, line 78: This whole file also probably belongs in an enterprise/ package of some sort.

pkg/render/kubecontrollers/kube-controllers.go
```diff
@@ -69,12 +69,10 @@ type KubeControllersConfiguration struct {
...
 
-	Installation   *operatorv1.InstallationSpec
-	Authentication *operatorv1.Authentication
+	Installation *operatorv1.InstallationSpec
 
-	// ManagementCluster and ManagementClusterConnection are inputs for the enterprise
-	// es-kube-controllers assembler. No base rendering reads them.
-	ManagementCluster           *operatorv1.ManagementCluster
+	// ManagementClusterConnection is an input for the enterprise es-kube-controllers
+	// assembler. No base rendering reads it.
```
caseydavenport: This is a breach. No base rendering needs it -> should be handled in extension.

pkg/render/tenant.go
```diff
@@ -0,0 +1,42 @@
...
+import (
+	"fmt"
+
+	operatorv1 "github.com/tigera/operator/api/v1"
+)
+
+// LinseedNamespace determine the namespace in which Linseed is running.
+// For management and standalone clusters, this is always the tigera-elasticsearch
+// namespace. For multi-tenant management clusters, this is the tenant namespace
+func LinseedNamespace(tenant *operatorv1.Tenant) string {
```
caseydavenport, line 26: This file should also go into an `enterprise/` package of some sort - let's make it clear what files are strictly for enterprise.

pkg/render/typha.go
```diff
@@ -82,7 +82,7 @@ type TyphaConfiguration struct {
 	TLS               *TyphaNodeTLS
 	MigrateNamespaces bool
 	ClusterDomain     string
-	NonClusterHost    *operatorv1.NonClusterHost
```
caseydavenport, line 85: This is lazy - it's not just the type that matters, its' the concept of an NCH altogether. This isn't actually meeting the requirement.

PR #5100, opened 2026-07-23
chore(deps): bump helm to v3.21.3 and oras-go to v2.6.2 to drop vulnerable containerd

electricjesus:
## Description

This bumps Helm to v3.21.3 and floors oras-go to v2.6.2 so the vulnerable containerd code stops shipping in the operator binary.

The operator never imports `github.com/containerd/containerd` itself. It only pulled it in transitively through `helm.sh/helm/v3`. Helm v3.21.3 dropped that dependency, so once we bump Helm the whole containerd module family (containerd, errdefs, log, platforms) falls out of the build graph. That clears three containerd CVEs from the image:

- GO-2026-5338 / CVE-2026-50195 (crit 9.9, checkpoint tag poisoning)
- GO-2026-5064 / CVE-2026-53492 (crit 9.6, CDI device smuggling)
- CVE-2026-53489 / GO-2026-5622 (high, symlink log read)

The oras-go bump to v2.6.2 clears CVE-2026-50163 (high 7.1).

Only `go.mod` and `go.sum` change. Helm v3.21.3 also pulls `cyphar/filepath-securejoin` up to v0.7.0, which is why that line moves too. No operator code changed, so this is a pure dependency graph update.

Part of the CVE / third-party updates epic, EV-6408.

## Release Note

```release-note
Removed the vulnerable containerd dependency from the operator image by bumping Helm to v3.21.3, and cleared an oras-go CVE by flooring it to v2.6.2.
```

## For PR author

- [x] Tests for change. (existing render unit tests exercise the changed graph)
- [ ] If changing pkg/apis/, run `make gen-files` (n/a, no API change)
- [ ] If changing versions, run `make gen-versions` (n/a, this is a go.mod dep, not a component version)

### Test plan

Built the operator binary before and after the bump and read the embedded build info. containerd is present before and gone after.

Before (`go version -m` on the pre-bump binary):

```
dep  github.com/containerd/containerd  v1.7.33
dep  helm.sh/helm/v3                   v3.21.2
dep  oras.land/oras-go/v2              v2.6.1
```

After (`go version -m` on the post-bump binary):

```
dep  helm.sh/helm/v3      v3.21.3
dep  oras.land/oras-go/v2  v2.6.2
```

No `github.com/containerd/containerd` line remains in the binary. Cross-checks:

```
$ go version -m operator | grep 'containerd/containerd'   # (no output)
$ go list -deps ./... | grep 'github.com/containerd/containerd$'
NONE
```

[...]

PR #5174, opened 2026-08-13
Drop containerd by moving to helm v3.21.3

dimitri-nicolo:
## Description

Cherry-pick of master's `3fce8aca0` ("chore(deps): bump helm to v3.21.3 and oras-go to v2.6.2 to drop vulnerable containerd", #5100), to remove containerd from the operator on this branch instead of bumping it again.

containerd isn't used here. It arrives via `pkg/render/istio` → helm's `pkg/action` → helm's OCI `pkg/registry` → `containerd/remotes`. That render code only uses helm as a local template renderer — it never contacts a registry — but `pkg/action` imports `pkg/registry` unconditionally, so containerd is linked into the binary on a path nothing can reach. helm v3.21.3 swapped containerd for `oras-go/v2`, so moving to it drops containerd entirely.

**This branch is the furthest behind of the three, so this is the heaviest of the set.** Reviewers should weigh it on that basis rather than by analogy to the v1.42 pick.

### 1. It needs a Go toolchain lift

helm v3.21.3 declares `go 1.26.0`; this branch builds with Go 1.25.7, so `go mod tidy` refuses outright:

```
helm.sh/helm/v3@v3.21.3 requires go >= 1.26.0 (running go 1.25.7; GOTOOLCHAIN=local)
```

There is no way around this by picking a lower helm — v3.21.3 is the *only* containerd-free release, and everything from v3.21.1 up requires Go 1.26:

| helm | go directive | containerd in go.mod |
|---|---|---|
| v3.20.0 – v3.21.0 | 1.25.0 | yes |
| v3.21.1, v3.21.2 | 1.26.0 | yes |
| **v3.21.3** | 1.26.0 | **no** |

`GO_BUILD_VER` and the build image do **not** change. Instead `CONTAINERIZED` gains one line — `-e GOTOOLCHAIN=go1.26.4+auto` — letting the container fetch the newer toolchain, and `go.mod`'s directive moves to `go 1.26.0`. This is the same arrangement release-v1.40 already runs (it builds with the Go 1.25.12 image and a `go 1.26.5` directive), so it's an established pattern on a release branch, not a new one.

### 2. It's a two-minor k8s jump

helm v3.21.3 requires client-go v0.36.2, and controller-runtime has to move with it:

| | before | after |
|---|---|---|
| helm | v3.19.5 | v3.21.3 |
| controller-runtime | v0.22.5 | v0.24.1 |
| k8s.io/{api,client-go,apimachinery,apiserver,apiextensions-apiserver,kube-aggregator} | v0.34.3 | v0.36.2 |
| oras-go/v2 | v2.6.0 | v2.6.2 |
| containerd | v1.7.29 + 3 more | **gone** |

`oras-go` goes to v2.6.2 rather than the v2.6.1 tidy selects, to match the CVE-2026-50151 / CVE-2026-50163 pin the other branches carry.

These can't be split into separate buildable commits: k8s.io/api v0.36 removed `scheduling/v1alpha1`, which helm v3.19.5's `k8s.io/kubectl` still imports, and controller-runtime v0.22.5 doesn't support client-go v0.36.

### Verified

- `go build ./...` passes under Go 1.26.4. **No code changes were needed**, despite the two-minor controller-runtime jump.
- containerd drops to 0 linked packages and leaves `go.mod` entirely.
- `go test ./pkg/render/istio/...` (the only helm consumer) passes.
- `go test ./pkg/...` behaves identically to pristine release-v1.41: every package passes on both trees except `pkg/controller/installation`, which hits `go test`'s default 600s timeout on both alike. CI runs ginkgo rather than `go test`, so that timeout is an artifact of the local invocation, not a regression. Diffed against a clean checkout of the branch.

### Companion PRs

- release-v1.42: #5172 — the cheap one. That branch is already on controller-runtime v0.24.1 / client-go v0.36.3, so only helm and oras-go move (+43/-55).
- release-v1.40: #5168 — mid-weight. Carries controller-runtime v0.23.3 → v0.24.1 and k8s.io v0.35.7 → v0.36.2, no Go change (+83/-98).

If the Go toolchain lift isn't wanted on this branch, the alternative is to keep bumping containerd here as advisories land — it's unreachable code, so exposure is limited either way. That's a reasonable call to make for v1.41 specifically, and it doesn't affect the other two PRs.

## Release Note

```release-note
None
```


dimitri-nicolo:
Closing in favour of #5176.

This PR needed Go 1.26 on a branch that builds with Go 1.25.7, and we do not want to move release-v1.41 to Go 1.26. The CI failure made the cost concrete: `golangci-lint` in the branch's go-build image is built with Go 1.25 and refuses to lint a module targeting 1.26 (`the Go language version (go1.25) used to build golangci-lint is lower than the targeted Go version (1.26.0)`). That is fixable — release-v1.40 works around it with a separate `CALICO_BUILD_LINT` image — but it only papers over the fact that the operator binary would then be compiled by Go 1.26.4 via `GOTOOLCHAIN`.

There is no smaller helm that achieves the containerd removal: v3.21.3 is the only containerd-free release and every helm from v3.21.1 on declares `go 1.26.0`.

#5176 instead takes everything that is reachable under Go 1.25 — oras-go v2.6.0 → v2.6.2 (10 advisories → 0), helm v3.19.5 → v3.20.2 (chart extraction path traversal), containerd v1.7.29 → v1.7.34, and the k8s/controller-runtime floor those require. containerd stays, with three CRI-checkpoint advisories that only the helm v3.21.3 route can clear.

PR #5176, opened 2026-08-13
Bump helm, oras-go, containerd and k8s libraries to clear advisories

dimitri-nicolo:
## Description

Advisory cleanup for release-v1.41, staying on the branch's own Go 1.25 line.

### What moves

Every version here is the newest that still builds under Go 1.25, and the resulting set is the same one release-v1.40 already ships:

| | before | after | why |
|---|---|---|---|
| `oras.land/oras-go/v2` | v2.6.0 | v2.6.2 | **10 advisories → 0** |
| `helm.sh/helm/v3` | v3.19.5 | v3.20.2 | GO-2026-5435 — chart extraction path traversal |
| `containerd` | v1.7.29 | v1.7.34 | 6 of 9 advisories cleared; newest v1.7.x |
| `k8s.io/*` | v0.34.3 | v0.35.7 | required by helm v3.20.2 |
| `controller-runtime` | v0.22.5 | v0.23.3 | required by client-go v0.35 |
| `GO_BUILD_VER` | `1.25.7-llvm18.1.8-k8s1.34.3` | `1.25.13-llvm18.1.8-k8s1.35.7` | keep the build image's k8s tooling level with the k8s libraries |

The k8s and controller-runtime moves aren't optional extras — helm v3.20.2 requires client-go v0.35.1, and controller-runtime v0.22.5 doesn't support client-go v0.35.

All eight versioned `k8s.io/*` modules are held at v0.35.7 together, including the indirect `cli-runtime` and `kubectl`. MVS would otherwise leave those two at v0.35.1 — the minimum helm v3.20.2 asks for — which builds fine but leaves a skewed k8s module set (thanks @copilot for catching it). `klog` v2.130.1 and `kube-openapi` 20250910 are deliberately left alone: those are what the v0.35.7 line itself requires, not stragglers. Neither v0.35.1 nor v0.35.7 of `cli-runtime`/`kubectl` carries a known advisory, so this is a consistency fix rather than a security one.

**Still on Go 1.25, and no `GOTOOLCHAIN` override is added** — avoiding Go 1.26 is the point of this branch and that is unchanged. The `go` directives in `go.mod` (1.25.7) and `api/go.mod` (1.25.5) are also unchanged; they are minimums that the newer toolchain satisfies as they stand.

### Follow-up commits

Two commits sit on top of the original bump, each reviewable on its own.

**`7086ea6d6` — disable the client-go `WatchListClient` gate in the installation UT suite.** client-go turns this feature on by default from v0.35 (`v0.34.3`: `Default: false`; `v0.35.7`: `{Version: 1.35, Default: true}`). With it on, an informer's reflector issues a streaming list and then blocks until the server sends the bookmark that marks the end of the initial event stream. These tests run against `kfake.NewSimpleClientset`, which never sends that bookmark, so the typha autoscaler's cache sync never returns.

The suite stopped dead at *"should create a separate Typha deployment for non-cluster hosts"*, the reflector logged `awaiting required bookmark event for initial events stream` every 10s, and Semaphore killed the job at ~30 minutes. Nothing failed — the job never finished, and UT had not completed once since the k8s bump. release-v1.42 and master both carry this exact fix, added incidentally inside their own k8s-1.35 moves (5d5f1aa4 / #4694 and 8d9201e0 / #4613). Neither cherry-picks here — both are large toolchain bumps that conflict, and the master one *is* the Go 1.26 bump this branch avoids — so it lands standalone, which also gives the next branch crossing client-go v0.35 something pickable.

**`d652bc301` — align `GO_BUILD_VER` with the bumped k8s libraries.** The `k8s` component of the tag selects the k8s tooling bundled in the build image, and it was still on 1.34.3 while the libraries moved to v0.35.7. Every branch keeps these level (master `k8s v1.37.0-beta.0` ↔ `client-go v0.37.0-beta.0`; release-v1.42 `1.36.3` ↔ `v0.36.3`), and #4694 — this same k8s-1.35 move on release-v1.42 — bumped the tag in the same commit.

`k8s1.35.7` is the patch-exact match for client-go v0.35.7 and is only published against Go 1.25.12/1.25.13, so the Go patch moves with it; 1.25.13 is newest on the 1.25 line and picks up the stdlib fixes released since 1.25.7, which suits a PR whose purpose is clearing advisories. `HELM3_VERSION` is deliberately left at v3.11.3 — it pins the helm *binary* used to render charts at build time, unrelated to the linked `helm.sh/helm/v3` library; release-v1.42 and release-v1.40 both ship helm v3.21.3 against that same v3.11.3 binary. `KINDEST_NODE_VERSION` (v1.31.12) is likewise independent.

### Why not just drop containerd, like master did

master removed containerd outright by moving to helm v3.21.3, which replaced containerd's `remotes` with `oras-go/v2`. That isn't reachable from this branch:

| helm | go directive | containerd in go.mod |
|---|---|---|
| v3.20.0 – v3.21.0 | 1.25.0 | yes |
| v3.21.1, v3.21.2 | 1.26.0 | yes |
| **v3.21.3** | 1.26.0 | **no** |

v3.21.3 is the only containerd-free release, and everything from v3.21.1 up needs Go 1.26. So dropping containerd here means moving the branch to Go 1.26, which we don't want on v1.41.

### What this does not fix

containerd stays in the binary, and three advisories remain open against v1.7.34 — CVE-2026-53492, CVE-2026-50195, CVE-2026-53489. All three are in containerd's CRI checkpoint paths. The operator links containerd only through helm's OCI registry client:

```
pkg/render/istio -> helm.sh/helm/v3/pkg/action
                 -> helm.sh/helm/v3/pkg/registry
                 -> github.com/containerd/containerd/remotes
```

and that renderer loads four embedded chart tarballs with `DryRun`/`ClientOnly` set, so it never contacts a registry — the CRI paths aren't reachable. There is no containerd v1.7.35+; those fixes are in containerd v2, which is a different module path (`github.com/containerd/containerd/v2`) that helm's v1 import can't resolve. So v1.7.34 is the ceiling for as long as this branch stays on Go 1.25.

### Verified

- `go build ./...` passes under the branch's own Go 1.25.7, and `go vet ./...` is clean for the whole module under the new Go 1.25.13.
- `go test ./pkg/render/istio/...` (the only helm consumer) passes.
- The UT hang reproduces and clears exactly as described: with the fix removed, the named spec hangs and `go test` panics at its timeout with the goroutine parked in `client-go@v0.35.7/tools/cache/reflector.go:368`; with it, that spec passes in 1.5s and the full installation suite in ~37s.
- The other packages using a fake clientset — `pkg/common`, `pkg/controller/csr`, `pkg/controller/utils`, `pkg/controller/utils/imageset` — all pass, so the fix isn't just moving the hang one suite along.
- All four arch tags CI builds exist for the new build image: `amd64`, `arm64`, `ppc64le`, `s390x`.
- The only production-code change in this PR is the dependency bump itself; the WatchList change is test-only.

### Companion PRs

- release-v1.42: #5172 — takes the full master fix; that branch is already on Go 1.26.5, so containerd goes to zero there.
- release-v1.40: #5168 — same, via a `GOTOOLCHAIN` override that branch already carries.

## Release Note

```release-note
None
```


dimitri-nicolo:
Good catch — fixed in f57e9f9.

`k8s.io/cli-runtime` and `k8s.io/kubectl` are now at v0.35.7 with the rest of the set, so all eight versioned `k8s.io/*` modules move together. They were lagging because MVS picks the minimum that satisfies the graph, and helm v3.20.2 only asks for v0.35.1 — `component-base` got pulled to v0.35.7 by `apiserver`, but nothing pulled those two.

On the advisory half of the question: no, nothing was left unresolved. Both v0.35.1 and v0.35.7 of `cli-runtime` and `kubectl` are clean in OSV, so this is a consistency fix, not a security one.

I left `klog/v2` v2.130.1 and `kube-openapi` 20250910 alone deliberately — those are not stragglers, they are exactly what the v0.35.7 line pins (`k8s.io/apiserver@v0.35.7` requires both), so raising them would create skew rather than remove it.

Verified after the change: `go mod tidy` keeps v0.35.7 (an explicit require is not downgraded), `go build ./...` passes under the branch Go 1.25.7, and `go test ./pkg/render/istio/...` passes.

go.mod
pasanw, line 1: I believe we need to update GO_BUILD_VER and potentially the helm version in `Makefile`?

pasanw (approved):
One comment to consider - it's a potential gap that exists in #5100 too

PR #5177, opened 2026-08-13
Align third-party versions with calico-private release-calient-v3.23

dimitri-nicolo:
## Description

Moves this branch's third-party pins to match calico-private `release-calient-v3.23`, which is bumping them in [tigera/calico-private#13239](https://github.com/tigera/calico-private/pull/13239).

This branch tracks `release-calient-v3.23` (`config/enterprise_versions.yml` → `libcalico-go`), and the operator deploys those images, so its pins have to move with calico's or it will deploy versions that no longer get built.

| component | before | after |
|---|---|---|
| `eck-kibana` | 8.19.16 | **8.19.20** |
| `eck-elasticsearch` | 8.19.16 | **8.19.20** |
| `coreos-prometheus` | v3.9.1 | **v3.13.2** |
| `coreos-alertmanager` | v0.30.1 | **v0.33.1** |
| `eck-elasticsearch-operator` | 3.4.0 | **3.4.1** |
| `ISTIO_VERSION` (Makefile) | 1.29.4 | **1.29.6** |

`pkg/components/enterprise.go` is regenerated from the yaml by `gen-versions` — not hand-edited — so `validate-gen-versions` stays clean.

### Also fixes an inconsistency

The `eck-elasticsearch-operator` bump repairs a mismatch already on this branch: `pkg/imports/crds/enterprise/01-crd-eck-bundle.yaml` was synced to ECK **3.4.1** in #5173, while this pin still claimed **3.4.0**. They now agree.

### Deliberately not changed

**Go and k8s.** [calico-private#13254](https://github.com/tigera/calico-private/pull/13254) moves v3.23 to Go 1.25.13, but this branch is on the Go 1.26 line (`GO_VERSION=1.26.5`, `K8S_VERSION=v1.36.3`). Those pins are not meant to track calico-private, and #5172 (helm v3.21.3) **requires** Go 1.26 here — aligning them downward would break it. The operator/calico Go versions have been independent on this branch for a while.

**`hack/release/prep_test.go`.** It contains the literals `8.19.16` and `v3.9.1`, so a grep for the old versions finds them. They're inputs to a self-contained round-trip test asserting that `updateConfigVersions` leaves third-party pins alone while rewriting product versions; the values are arbitrary and match nothing real. Changing them would be diff noise.

### Verified

- `go build ./...` passes.
- Istio charts re-fetched at 1.29.6 — `pkg/render/istio` passes (no golden fixtures are pinned to the chart version).
- `pkg/render/logstorage/...` (elasticsearch, kibana, eck, esgateway, esmetrics, linseed, dashboards), `pkg/components` and `hack/release` all pass.

### Companion PR

- release-v1.40: the equivalent alignment against `release-calient-v3.22`, which additionally carries Go 1.25.12 → 1.25.13.

## Release Note

```release-note
None
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -239,7 +239,7 @@ endif
 
 # To update the Istio version, see "Updating the bundled version of Istio" in docs/common_tasks.md.
 ISTIO_HELM_REPO ?= https://istio-release.storage.googleapis.com/charts
-ISTIO_VERSION ?= 1.29.4
+ISTIO_VERSION ?= 1.29.6
 ISTIO_RESOURCES_DIR = pkg/render/istio
 ISTIO_CHARTS = base istiod cni ztunnel
 ISTIO_CHART_FILES = $(addprefix $(ISTIO_RESOURCES_DIR)/,$(addsuffix .tgz,$(ISTIO_CHARTS)))
--- a/config/enterprise_versions.yml
+++ b/config/enterprise_versions.yml
@@ -47,13 +47,13 @@ components:
     version: v3.23.1
   # eck-kibana holds the version of Kibana built for tigera/kibana
   eck-kibana:
-    version: 8.19.16
+    version: 8.19.20
   kibana:
     image: kibana
     version: v3.23.1
   # eck-elasticsearch holds the version of Elasticsearch built for tigera/elasticsearch
   eck-elasticsearch:
-    version: 8.19.16
+    version: 8.19.20
   elasticsearch:
     image: elasticsearch
     version: v3.23.1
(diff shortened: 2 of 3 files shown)


dimitri-nicolo:
Valid finding, thanks — the pointer really is dead. `docs/common_tasks.md` has "Updating the bundled version of Envoy Gateway" and nothing for Istio.

It is not specific to this PR though: the comment has been wrong on master and on every release branch back to release-v1.40, and this PR only changes the version number two lines below it. So rather than patch the comment on two release branches (which would leave master still wrong, and have docs flowing backwards from release to master), I have written the missing section on master in #5179.

I went with adding the section rather than rewording the comment, since the repo already documents the equivalent Envoy Gateway procedure there and the Istio one has two non-obvious traps worth recording:

- The chart tarballs are gitignored and never committed, so a stale copy in `pkg/render/istio` keeps being used after a bump — `make istio_charts` will not re-fetch over an existing file. You get a green build and green tests rendering the *old* version. (This is why this PR deletes them before re-fetching.)
- The version has to stay in step with the Istio version the calico repo builds `istio-pilot` / `istio-install-cni` / `istio-ztunnel` / `istio-proxyv2` from, which is exactly what this PR is doing.

Leaving this PR to version pins only; #5179 can be picked back to the release branches if the docs are wanted there.

rene-dekker (approved):

PR #5179, opened 2026-08-14
docs: add the missing "Updating the bundled version of Istio" section

dimitri-nicolo:
## Description

The `Makefile` has pointed at a doc section that was never written:

```make
# To update the Istio version, see "Updating the bundled version of Istio" in docs/common_tasks.md.
```

`docs/common_tasks.md` has the equivalent **Envoy Gateway** section but nothing for Istio, so that pointer is dead — on master and on every release branch back to release-v1.40. Anyone following it to bump Istio finds nothing.

This writes the section, following the Envoy Gateway one, from the procedure as it actually works today: bump `ISTIO_VERSION`, delete the downloaded charts, re-run `istio_charts`, rebuild, re-run the render tests.

### Two things worth writing down

Both are easy to get wrong and neither is discoverable from the Makefile:

- **The chart tarballs are gitignored and never committed**, so a stale copy left in `pkg/render/istio` silently keeps being used after a version bump — `make istio_charts` won't re-fetch over an existing file. They have to be deleted, not just re-made. (The failure mode is worse than it sounds: you get a green build and green tests rendering the *old* Istio version.)
- **The version has to stay in step with the calico repo.** The operator renders the upstream Helm charts; `projectcalico/calico` builds the images those manifests point at (`istio-pilot`, `istio-install-cni`, `istio-ztunnel`, `istio-proxyv2`) under `istio/` and `third_party/istio-ztunnel/`. If the two drift you get manifests from one Istio release referencing images built from another. The Makefile comment now says this too.

### Origin

Found by the review bot on #5177, a version-alignment PR against release-v1.42. The finding was valid but pre-existing and repo-wide, so it's fixed here at the source rather than in the release-branch PRs — those keep to version pins only, and this can be picked back if it's wanted on the release branches.

Docs-only; no code or generated files change.

## Release Note

```release-note
None
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -260,7 +260,10 @@ else
   GIT_VERSION?=$(shell git describe --tags --dirty --always --abbrev=12)
 endif
 
-# To update the Istio version, see "Updating the bundled version of Istio" in docs/common_tasks.md.
+# To update the Istio version, see "Updating the bundled version of Istio" in
+# docs/common_tasks.md.  Keep this in step with the Istio version the calico
+# repo builds the Istio images from, or the rendered manifests will point at
+# images built from a different release.
 ISTIO_HELM_REPO ?= https://istio-release.storage.googleapis.com/charts
 ISTIO_VERSION ?= 1.29.2
 ISTIO_RESOURCES_DIR = pkg/render/istio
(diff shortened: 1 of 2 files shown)


caseydavenport:

```

```

PR #5187, opened 2026-08-14
Grant calico-kube-controllers access to Networks

caseydavenport:
Automatic host endpoints and LoadBalancer IP assignment stop being programmed on Enterprise clusters. kube-controllers has no permission to list networks, so its datastore syncer retries the list forever and never reaches in-sync, which leaves the node and load balancer controllers idle.

This grants get, list and watch on networks to the calico-kube-controllers cluster role.

```release-note
Fixes automatic host endpoints and LoadBalancer IP addresses not being programmed.
```


fasaxc (approved):
Was just about to raise this!  Looks like #4871 reverted #5148.

PR #5196, opened 2026-08-14
Move FelixConfiguration writes onto server-side apply on v3 clusters

caseydavenport:
## Description

Moves the FelixConfiguration write path onto server-side apply on v3-native clusters, and migrates every controller that writes FelixConfiguration onto it.

- The write path sits behind an interface with one implementation per API group, chosen by whether the operator is running against v3 CRDs.
- On `crd.projectcalico.org/v1` it stays the existing merge patch, with an annotation recording what the operator last wrote so a user's edit is still detectable. That's the aggregated-apiserver case, where the operator writes v1 and users write v3 and managed fields aren't comparable across the boundary.
- On `projectcalico.org/v3` writes go through server-side apply. Each write site owns a named set of fields, and per field it either defers to a user's edit, overrides it, or degrades the Installation status.
- Every controller declares the fields it drives on every reconcile, instead of only when the field is unset, because an owner that stops declaring a field deletes it.
- The policy sync socket path gets one owner that computes it from the ApplicationLayer, GatewayAPI, EgressGateway, and Istio resources together, so no controller can clear another's claim.
- Fields the operator wrote before this change are recorded as updates rather than applies. The apply path recognizes that older manager name, so the first apply that changes a value takes ownership back from it.

Behavior changes worth a look during review:

- eBPF mode gets written as explicitly disabled in cases where the operator previously wrote nothing at all.
- The eBPF kube-proxy health port is cleared when a cluster leaves eBPF mode, restoring Felix's own default instead of leaving it pinned at zero.
- The policy sync socket path is cleared once no feature needs it, where before it stuck around forever after the first write.
- On aggregated-apiserver clusters the operator only removes a field it has a record of having written, so a value inherited from before the upgrade is left alone.
- The application layer's TPROXY mode stays on the old merge path on purpose. Declaring it unconditionally would set it on clusters upgrading from before the field existed, which restarts Felix.

Remaining work is out of scope here: generated apply configuration types in place of hand-built payloads.

### DIFF
--- a/pkg/controller/applicationlayer/applicationlayer_controller.go
+++ b/pkg/controller/applicationlayer/applicationlayer_controller.go
@@ -24,6 +24,7 @@ import (
 	"github.com/tigera/operator/pkg/common"
 	"github.com/tigera/operator/pkg/controller/gatewayapi"
 	"github.com/tigera/operator/pkg/controller/options"
+	"github.com/tigera/operator/pkg/controller/sharedconfig"
 	"github.com/tigera/operator/pkg/controller/status"
 	"github.com/tigera/operator/pkg/controller/utils"
 	"github.com/tigera/operator/pkg/controller/utils/imageset"
@@ -521,7 +522,7 @@ func (r *ReconcileApplicationLayer) patchFelixConfiguration(ctx context.Context,
 	}
 	istioNeeds := utils.IstioRequiresPolicySync(istioCR, r.variant)
 
-	_, err = utils.PatchFelixConfiguration(ctx, r.client, func(fc *v3.FelixConfiguration) (bool, error) {
+	_, err = sharedconfig.NewWriter(r.client).UpdateFelixConfiguration(ctx, func(fc *v3.FelixConfiguration) (bool, error) {
 		wafEventLogsFileEnabled := wafEventLogsFileRequired(al, gatewayWAFEnabled)
 
 		var tproxyMode string
--- a/pkg/controller/applicationlayer/applicationlayer_controller.go
+++ b/pkg/controller/applicationlayer/applicationlayer_controller.go
@@ -81,6 +81,7 @@ func newReconciler(mgr manager.Manager, opts options.ControllerOptions, licenseA
 		provider:        opts.DetectedProvider,
 		status:          status.New(mgr.GetClient(), "applicationlayer", opts.KubernetesVersion),
 		clusterDomain:   opts.ClusterDomain,
+		useV3CRDs:       opts.UseV3CRDs,
 		variant:         opts.Variant,
 		licenseAPIReady: licenseAPIReady,
 	}
@@ -167,6 +168,7 @@ type ReconcileApplicationLayer struct {
 	provider        operatorv1.Provider
 	status          status.StatusManager
 	clusterDomain   string
+	useV3CRDs       bool
 	variant         operatorv1.ProductVariant
 	licenseAPIReady *utils.ReadyFlag
 }
@@ -522,7 +524,7 @@ func (r *ReconcileApplicationLayer) patchFelixConfiguration(ctx context.Context,
 	}
 	istioNeeds := utils.IstioRequiresPolicySync(istioCR, r.variant)
 
-	_, err = sharedconfig.NewWriter(r.client).UpdateFelixConfiguration(ctx, func(fc *v3.FelixConfiguration) (bool, error) {
+	_, err = sharedconfig.NewWriter(r.client, r.useV3CRDs).UpdateFelixConfiguration(ctx, func(fc *v3.FelixConfiguration) (bool, error) {
 		wafEventLogsFileEnabled := wafEventLogsFileRequired(al, gatewayWAFEnabled)
 
 		var tproxyMode string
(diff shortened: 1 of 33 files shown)


pkg/controller/egressgateway/egressgateway_controller.go
```diff
@@ -84,6 +85,7 @@ func newReconciler(mgr manager.Manager, opts options.ControllerOptions, licenseA
 		provider:        opts.DetectedProvider,
 		status:          status.New(mgr.GetClient(), "egressgateway", opts.KubernetesVersion),
 		clusterDomain:   opts.ClusterDomain,
+		useV3CRDs:       opts.UseV3CRDs,
```
caseydavenport: We should just do `Opts: opts` and remove all the copies.

pkg/controller/applicationlayer/applicationlayer_controller_test.go
```diff
@@ -150,17 +150,15 @@ var _ = Describe("Application layer controller tests", func() {
 			_, err = r.Reconcile(ctx, reconcile.Request{})
 			Expect(err).ShouldNot(HaveOccurred())
 
-			By("ensuring that felix configuration PolicySyncPathPrefix is left as is, even after ALP deletion")
+			By("ensuring that felix configuration PolicySyncPathPrefix is cleared after ALP deletion")
```
caseydavenport: Hm, this test makes me thinkg it might have been intentional?

pkg/controller/applicationlayer/applicationlayer_controller.go
```diff
@@ -507,75 +505,56 @@ func (r *ReconcileApplicationLayer) getTProxyMode(al *operatorv1.ApplicationLaye
...
-				return false, nil
-			}
 
-			// If the mode is already set, fall through to the normal logic, it's safe to force-set the field now.
-			// This also avoids churning the config if a previous version of the operator set it to Disabled already,
-			// we avoid setting it back to nil.
+	_, err := writer.UpdateFelixConfiguration(ctx, func(fc *v3.FelixConfiguration) (bool, error) {
+		ok, tproxyMode := r.getTProxyMode(al)
+		if !ok && fc.Spec.TPROXYMode == "" {
+			// Setting this during an upgrade from before the field existed makes Felix restart,
```
caseydavenport: According to the original comment, this is long since not needed based on version.

pkg/controller/applicationlayer/applicationlayer_controller.go
```diff
@@ -507,75 +505,56 @@ func (r *ReconcileApplicationLayer) getTProxyMode(al *operatorv1.ApplicationLaye
...
+// update path for the upgrade workaround below.
 func (r *ReconcileApplicationLayer) patchFelixConfiguration(ctx context.Context, al *operatorv1.ApplicationLayer, gatewayWAFEnabled bool) error {
-	// Fetch the Istio CR and Installation variant so DesiredPolicySyncPathPrefix
-	// can see whether the istio side still needs the field. Both reads tolerate
-	// NotFound — the istio side has no claim if either is absent.
-	istioCR, err := utils.GetIstio(ctx, r.client)
-	if err != nil {
+	writer := sharedconfig.NewWriter(r.client, r.useV3CRDs)
+
+	if _, err := writer.ApplyFelixConfiguration(ctx, declareWAFEventLogsFile(al, gatewayWAFEnabled)); err != nil {
```
caseydavenport: Back-to-back, maybe this should support a variadic set of applications to make?

pkg/controller/istio/istio_controller.go
```diff
@@ -116,10 +115,11 @@ func Add(mgr manager.Manager, opts options.ControllerOptions) error {
...
 func newReconciler(mgr manager.Manager, opts options.ControllerOptions) *ReconcileIstio {
 	r := &ReconcileIstio{
-		Client:   mgr.GetClient(),
-		scheme:   mgr.GetScheme(),
-		status:   status.New(mgr.GetClient(), "istio", opts.KubernetesVersion),
-		provider: opts.DetectedProvider,
+		Client:    mgr.GetClient(),
+		scheme:    mgr.GetScheme(),
+		status:    status.New(mgr.GetClient(), "istio", opts.KubernetesVersion),
+		provider:  opts.DetectedProvider,
```
caseydavenport: same here - should just be `opts: opts,`

pkg/controller/istio/istio_controller.go
```diff
@@ -257,13 +258,15 @@ func (r *ReconcileIstio) Reconcile(ctx context.Context, request reconcile.Reques
...
-	_, err = utils.PatchFelixConfiguration(ctx, r.Client, func(fc *v3.FelixConfiguration) (bool, error) {
-		return r.setIstioFelixConfiguration(ctx, instance, fc, false)
-	})
-	if err != nil {
+	writer := sharedconfig.NewWriter(r.Client, r.useV3CRDs)
+	if _, err = writer.ApplyFelixConfiguration(ctx, r.declareIstioFelixConfiguration(instance, false)); err != nil {
 		r.status.SetDegraded(operatorv1.ResourceCreateError, "Error patching felix configuration with Istio settings", err, log)
 		return reconcile.Result{}, err
 	}
+	if _, err = writer.ApplyFelixConfiguration(ctx, sharedconfig.DeclarePolicySyncPathPrefix(ctx, r.Client)); err != nil {
```
caseydavenport: Any way to make this not require a Client?

PR #5197, opened 2026-08-14
Move guardian's Enterprise trusted certificates behind the extension

caseydavenport:
## Description

Split out of https://github.com/tigera/operator/pull/5170 to keep that review smaller, and the second of the per-controller ones. The cluster connection controller built Guardian's trusted bundle from a fixed list that included the packet capture and Prometheus server certificates, neither of which exists outside Enterprise.

- the controller builds the bundle from the core secrets, and the extension adds what the variant needs Guardian to trust
- the watches on those two secrets move to the extension along with them
- the bundle gets created earlier in the reconcile so the extension has something to add to

No behavior change on either variant.

## Release Note

```release-note
None
```

### DIFF
--- a/pkg/controller/clusterconnection/clusterconnection_controller.go
+++ b/pkg/controller/clusterconnection/clusterconnection_controller.go
@@ -53,7 +53,6 @@ import (
 	"github.com/tigera/operator/pkg/render"
 	"github.com/tigera/operator/pkg/render/common/networkpolicy"
 	"github.com/tigera/operator/pkg/render/goldmane"
-	"github.com/tigera/operator/pkg/render/monitor"
 	"github.com/tigera/operator/pkg/render/whisker"
 	"github.com/tigera/operator/pkg/tls/certificatemanagement"
 )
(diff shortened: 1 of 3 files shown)


pkg/controller/clusterconnection/clusterconnection_controller.go
```diff
@@ -252,13 +249,25 @@ func (r *ReconcileConnection) Reconcile(ctx context.Context, request reconcile.R
...
+	// Guardian to trust, and produces the Enterprise-specific Guardian inputs the
+	// controller reads back below (the managed cluster version and the license-gated
+	// egress policy flag). For the core operator this is a no-op and the render inputs
+	// carries no extension data, so the OSS defaults apply.
 	ci := controller.Inputs{
-		RenderInputs:       render.Inputs{Installation: installationSpec, ClusterDomain: r.opts.ClusterDomain},
+		RenderInputs: render.Inputs{
+			Installation:  installationSpec,
+			ClusterDomain: r.opts.ClusterDomain,
+			TrustedBundle: trustedBundle,
```
Brian-McM, line 269: Just to confirm, the trusted bundle is modified through reference in the extension, so we don't need to get the TrustedBundle from the returned ci object before passing it to render, right?

I don't see anything wrong with this (sort of unavoidable anyway), just want to confirm.

Brian-McM (approved):

pkg/controller/clusterconnection/clusterconnection_controller.go
caseydavenport: Yep, confirmed. `TrustedBundle` is an interface field and `AddCertificates` is declared on the pointer receiver (`certificatebundle.go:115`), so the extension mutates the same bundle the local variable points at. Copying `ci` by value copies the pointer, not the bundle.

Issue #5102, opened 2026-07-23
Issue when using helm v4


manuelbuil:
We moved to Helm v4 and we are starting to see issues related to the --server-side flag, which is enabled by default. In that case, kube-api verifies who is the manager of each field in the Installation object. 

When installing the Tigera Operator, helm creates the deployment and the Installation CR object, which at this point, has all the fields managed by helm. However, after a few seconds, the operator injects some values in a few fields of the same installation object and hence those fields get a new manager ==> manager: operator.  Unfortunately, helm does not know what values the operator wrote after the initial install, so there is a drift between the object in the cluster and what helm knows.  Moreover, the helm chart is defaulting those fields to empty or different values. And that is a problem if we try to change any field with helm. Any change will trigger helm to send the whole installation object it knows and kube-api will complain with the ownership error because helm is trying to change fields which are not "managed" by helm.

I was wondering if you are thinking about changing the architecture to avoid this issue (e.g. leaving Install as a user-only CR) or how would you recommend users with helm v4 to operate it.

Thanks

caseydavenport:
Just seeing this now, sorry for the delay - I'll do some thinking on this and come up with some options. Seems like we certainly need to do something here.

caseydavenport:
The conflict is the operator writing its computed defaults back into `Installation.spec` under its own field manager, so it ends up owning `kubernetesProvider`, `nodeUpdateStrategy`, `logging`, and parts of `calicoNetwork` and `cni`.

Here's my current thinking: stop writing the spec entirely.

- constants become CRD schema defaults, which the apiserver fills in without taking field ownership (a server-side apply that omits `spec.calicoLibHostPath` gets the default stored with nobody owning it, so Helm neither owns it nor conflicts on it)
- the ones needing cluster state (detected provider, pod CIDRs, aws-node) move into the status, and get dropped once a user sets the field in the spec

`helm upgrade --force-conflicts` unblocks you today, though it's only a workaround: the operator's next reconcile takes those fields straight back, since an update-type write doesn't do conflict detection at all. Once the operator stops writing the spec, that same forced apply becomes the permanent reset for existing clusters, for whichever fields your chart renders.

caseydavenport:
The operator now lives in the projectcalico/calico repository, so this moved to https://github.com/projectcalico/calico/issues/13697. Please follow it there.

PR #5203, opened 2026-08-17
Record installation defaults in the status instead of the spec

caseydavenport:
## Description

The operator writes its computed defaults back into the Installation spec, which makes the operator the field manager for anything it defaulted. Helm and Argo then conflict on those fields, and a server-side apply from the chart gets rejected. See #5102.

This changes the defaulting to leave the spec alone:

- computed defaults are recorded on the status instead of the spec, as a sparse object holding only the fields the user left unset
- defaulting seeds from the recorded values first, so a default that changes between operator versions doesn't change an existing cluster
- a field drops out of the recorded set as soon as the user sets it, and the effective config is still published in the status as before
- every other controller reads that effective config and waits until the core controller has published it, so there is one place that merges the spec, the defaults and the overlay
- the write that persists finalizers and migrated Calico config stays, so an upgrade from a manifest install is unaffected

Also fixes a bug this uncovered: an empty ipPools list marshalled the same as an unset one, so asking for no IP pools got the default pool instead. The documented behavior now holds.

IP pool semantics, spelled out in the CRD docs as part of this:

- `ipPools: []` means no pools. The operator deletes the pools it owns and creates none.
- Omitting `ipPools` leaves the choice to the operator. On a cluster where it has already recorded pool defaults, that keeps the pools in use, where before it deleted them.

First-install pool defaulting takes two status round-trips: the IP pool controller records the pool defaults, then the core controller republishes the effective config with them included. A few FV timeouts went up to cover that window.

Existing clusters where the operator already owns spec fields need one `helm upgrade --force-conflicts` to hand ownership back to the chart.

Constants that could be plain CRD schema defaults are left alone here - that's a follow-up.

Related: [CORE-13376](https://tigera.atlassian.net/browse/CORE-13376)

```release-note
Fixes server-side apply conflicts on Installation fields between the operator and the tool that manages the resource, such as Helm or Argo CD. The operator records the values it defaults in the Installation status rather than writing them into the spec.
```

```release-note
Fixes an issue where an empty list of IP pools in the Installation was treated as unset, causing a default IP pool to be created.
```

[CORE-13376]: https://tigera.atlassian.net/browse/CORE-13376?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pkg/controller/installation/core_controller.go
```diff
@@ -791,11 +791,31 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
...
 	}
 
+	// Snapshot the user's spec before defaulting, so defaults can be told apart from it later.
+	declaredSpec := *instance.Spec.DeepCopy()
+
+	// Seed from recorded defaults, so changing a default doesn't change existing clusters.
+	if instance.Status.Defaults != nil {
+		instance.Spec = utils.OverrideInstallationSpec(*instance.Status.Defaults, instance.Spec)
+
+		// Pool lists merge whole, so layer the per-pool defaults back under the declared pools.
```
caseydavenport, line 790: Is this the right behavior?

pkg/controller/installation/core_controller.go
```diff
@@ -853,21 +873,19 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
...
 		return reconcile.Result{}, err
 	}
 
-	// Write the discovered configuration back to the API. This is essentially a poor-man's defaulting, and
-	// ensures that we don't surprise anyone by changing defaults in a future version of the operator.
-	// Note that we only write the 'base' installation back. We don't want to write the changes from 'overlay', as those should only
-	// be stored in the 'overlay' resource.
-	if err := r.client.Patch(ctx, instance, preDefaultPatchFrom); err != nil {
-		r.status.SetDegraded(operatorv1.ResourceUpdateError, "Failed to write defaults", err, reqLogger)
+	// Persist metadata and migrated config, but not defaults. Writing defaults
```
caseydavenport: What is "metadata and migrated config" ?

pkg/controller/ippool/pool_controller.go
```diff
@@ -144,6 +144,44 @@ func hasOwnerLabel(pool *v3.IPPool) bool {
 	return false
 }
 
+// recordDefaults adds the pool defaults to the Installation status, leaving the spec untouched.
+func (r *Reconciler) recordDefaults(ctx context.Context, installation *operatorv1.Installation, declared operatorv1.InstallationSpec) error {
```
caseydavenport: Seems like we want a utility library for working with CRs and defaulting - especially if multiple controllers are going to be acting on the same object. Routing that through a well defined interface will help reduce bugs in the future.

Additionally, we're going to need this same status defaulting logic for all of the other CRDs used by the operator - not just Installation.

pkg/controller/ippool/pool_controller.go
```diff
@@ -178,10 +216,16 @@ func (r *Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
 	// The core installation controller adds a specific finalizer as part of performing defaulting,
 	// so wait for that before we continue.
 	readyToGo := slices.Contains(installation.GetFinalizers(), render.OperatorCompleteFinalizer)
-	if !readyToGo {
+	if !readyToGo || installation.Status.Computed == nil {
```
caseydavenport: Shouldn't the finalizer be gated on the Computed being set?

pkg/controller/ippool/pool_controller.go
```diff
@@ -178,10 +216,16 @@ func (r *Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
...
 		r.status.SetDegraded(operatorv1.ResourceNotReady, "Waiting for Installation defaulting to occur", nil, reqLogger)
 		return reconcile.Result{}, nil
 	}
+
+	// Recorded defaults are diffed against the raw spec; everything else works from the
+	// effective config.
+	declaredSpec := *installation.Spec.DeepCopy()
+	installation.Spec = *installation.Status.Computed.DeepCopy()
```
caseydavenport: I don't like this pattern of overwriting the `installation.Spec` with the computed - it makes it a lot less clear what config we're working with.

I think we should have a very clear pattern here, with consistent terminology that makes is very explicit whether you're acting on the declared vs. computed data.

```go
declared := installation.Spec.DeepCopy()
computed := installation.Status.Computed.DeepCopy()
```

and then future references pick either delcared / computed based on what they need (almost all should be using computed!)

pkg/controller/utils/utils.go
```diff
@@ -505,28 +505,24 @@ func GetInstallationStatus(ctx context.Context, client client.Client) (*operator
 	return &instance.Status, nil
 }
 
-// GetInstallationSpec returns the current installation, accounting for overlays. Controllers take
-// the variant from their ControllerOptions instead, so that the whole process agrees on one value.
+// GetInstallationSpec returns the effective config the core controller publishes
+// on the status, which already has defaults and the overlay applied.
 func GetInstallationSpec(ctx context.Context, client client.Client) (*operatorv1.InstallationSpec, error) {
```
caseydavenport: Might be worth a rename in order to make the code read more clearly - `GetComputedInstallation` perhaps?

pkg/controller/utils/utils.go
```diff
@@ -505,28 +505,24 @@ func GetInstallationStatus(ctx context.Context, client client.Client) (*operator
...
-	overlay := operatorv1.Installation{}
-	if err := client.Get(ctx, OverlayInstanceKey, &overlay); err != nil {
-		if !errors.IsNotFound(err) {
-			return nil, err
-		}
-	} else {
-		spec = OverrideInstallationSpec(spec, overlay.Spec)
+	if instance.Status.Computed == nil {
+		// Callers treat this like a missing Installation and wait.
+		return nil, errors.NewNotFound(schema.GroupResource{
```
caseydavenport: I think a distinct error string (or maybe even a distinct type?) would help here - i.e., "installations.status.computed" - so we can differentiate between "Installation not created" and "Installation status is not ready yet"

test/gatewayapi_test.go
```diff
@@ -385,6 +391,8 @@ var _ = Describe("GatewayAPI tests", func() {
 		Expect(c.Create(shutdownContext, instance)).NotTo(HaveOccurred())
 		Expect(c.Get(shutdownContext, utils.DefaultInstanceKey, instance)).NotTo(HaveOccurred())
 		instance.Status.Variant = operator.CalicoEnterprise
+		// The core controller, which normally publishes this, does not run in this suite.
```
caseydavenport:
```suggestion
```

These aren't needed.

pkg/controller/installation/core_controller.go
caseydavenport: Yeah. `OverrideInstallationSpec` merges `IPPools` all or nothing, so a user who declares a pool with only a CIDR would wipe out the name and block size we assigned it, and we'd rename the pool out from under an existing cluster. Layering by CIDR keeps the recorded per-pool defaults underneath whatever the user set.

pkg/controller/ippool/pool_controller.go
caseydavenport: Agreed on the interface. All defaulting goes through `MergeRecordedDefaults`, and each controller passes the dotted paths it owns, so the core controller and the pool controller can't clobber each other's recorded defaults.

The other CRDs each need their own `status.defaults` and `status.computed` before that library is worth writing, so I'd rather do it as a follow-on than grow this PR.

pkg/controller/installation/core_controller.go
```diff
@@ -791,11 +791,31 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 		}
 	}
 
+	// Snapshot the user's spec before defaulting, so defaults can be told apart from it later.
+	declaredSpec := *instance.Spec.DeepCopy()
+
+	// Seed from recorded defaults, so changing a default doesn't change existing clusters.
+	if instance.Status.Defaults != nil {
+		instance.Spec = utils.OverrideInstallationSpec(*instance.Status.Defaults, instance.Spec)
```
caseydavenport: I don't like that we're updating instance.Spec in place here without the intention to write it back to the API server (which is the intent of this PR). 

I think we should be defining a new Spec struct, filling it from the defaults, and then applying instance.Spec to that - to avoid polluting the object given to use by the user and to make it more clear when we're working with the declared spec vs. the defaulted one.

instance.Spec == declaredSpec.
defaulted := instance.Spec + defaults

pkg/controller/installation/core_controller.go
```diff
@@ -791,11 +791,31 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
...
+		}
+	}
+
 	// update Installation with defaults
 	if err := updateInstallationWithDefaults(ctx, r.client, instance, r.opts.DetectedProvider, r.opts.Variant); err != nil {
 		r.status.SetDegraded(operatorv1.ResourceReadError, "Error querying installation", err, reqLogger)
 		return reconcile.Result{}, err
 	}
+
+	recordedDefaults, err := utils.MergeRecordedDefaults(nil, declaredSpec, instance.Spec)
```
caseydavenport:
```suggestion
	recordedDefaults, err := utils.MergeRecordedDefaults(nil, declaredSpec, defaultedSpec)
```

pkg/controller/installation/core_controller.go
```diff
@@ -853,21 +873,19 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
...
 	}
 
-	// Write the discovered configuration back to the API. This is essentially a poor-man's defaulting, and
-	// ensures that we don't surprise anyone by changing defaults in a future version of the operator.
-	// Note that we only write the 'base' installation back. We don't want to write the changes from 'overlay', as those should only
-	// be stored in the 'overlay' resource.
-	if err := r.client.Patch(ctx, instance, preDefaultPatchFrom); err != nil {
-		r.status.SetDegraded(operatorv1.ResourceUpdateError, "Failed to write defaults", err, reqLogger)
+	// Write finalizers and migrated manifest config, never defaults, which would steal
+	// spec ownership from whoever manages the CR.
```
caseydavenport: We do write defaults - we just don't write them to the spec. This comment is confusing.

pkg/controller/installation/core_controller.go
```diff
@@ -853,21 +873,19 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
...
+	persisted := instance.DeepCopy()
+	persisted.Spec = declaredSpec
```
caseydavenport: Yeah - this confirms the awkwwardness. We're storing instance.Spec as declared, and then writing declared back to instance.Spec here. Roundabout.

pkg/controller/installation/core_controller.go
```diff
@@ -905,12 +923,24 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
...
 		instance.Status = operatorv1.InstallationStatus{}
-		if err := r.client.Status().Update(ctx, instance); err != nil {
+		if err := r.writeStatus(ctx, instance); err != nil {
 			r.status.SetDegraded(operatorv1.ResourceUpdateError, "Failed to write default status", err, reqLogger)
 			return reconcile.Result{}, err
 		}
 	}
 
+	// Publish the effective config before the IP pool gate below, since other
+	// controllers read it instead of the spec.
```
caseydavenport:
```suggestion
	// controllers read it instead of the spec. Once status.Computed is present, other controllers will kick in.
```

pkg/controller/installation/core_controller.go
```diff
@@ -1614,15 +1644,25 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
...
+// writeStatus updates the status without letting the response overwrite the defaulted spec.
+func (r *ReconcileInstallation) writeStatus(ctx context.Context, instance *operatorv1.Installation) error {
+	written := instance.DeepCopy()
+	if err := r.client.Status().Update(ctx, written); err != nil {
+		return err
+	}
+	instance.ObjectMeta = written.ObjectMeta
+	instance.Status = written.Status
+	return nil
+}
```
caseydavenport: Don't think we'd need this if we didn't do the instance / declared spec copying shenanigans above?

pkg/controller/utils/defaults.go
```diff
@@ -0,0 +1,237 @@
...
+	"fmt"
+	"reflect"
+	"strings"
+
+	operatorv1 "github.com/tigera/operator/api/v1"
+)
+
+// MergeRecordedDefaults merges the defaults absent from declared over the recorded ones. Owned
+// holds the dotted JSON paths the caller defaults.
+func MergeRecordedDefaults(recorded *operatorv1.InstallationSpec, declared, defaulted operatorv1.InstallationSpec, owned ...string) (*operatorv1.InstallationSpec, error) {
```
caseydavenport: I find recorded / delcared / defaulted to be a bit confusing here. What do each of these mean? What is meant to go into each of these?

pkg/controller/installation/windows_controller_test.go
```diff
@@ -667,7 +667,7 @@ var _ = Describe("windows-controller installation tests", func() {
 						Status: operator.InstallationStatus{
 							Variant: operator.Calico,
 							Computed: &operator.InstallationSpec{
-								Registry: "my-reg",
+								Registry: "some.registry.org/",
```
caseydavenport: `windows_controller.go` never moved to `Status.Computed`. It reads `instance.Spec` throughout and merges the overlay itself at line 264, so now that defaulting only lands on the core's private copy, a user who sets `windowsDataplane: HNS` and omits `windowsNodes` hits the degraded at line 274 forever and never gets calico-node-windows. We should migrate it along with the rest.

pkg/controller/ippool/pool_controller.go
```diff
@@ -174,15 +193,18 @@ func (r *Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
...
+	if installation.Status.Computed == nil {
 		r.status.SetDegraded(operatorv1.ResourceNotReady, "Waiting for Installation defaulting to occur", nil, reqLogger)
 		return reconcile.Result{}, nil
 	}
-	if installation.Spec.CNI == nil || installation.Spec.CNI.Type == "" {
+
+	// Everything below works from the computed config, except the recorded defaults, which are
+	// diffed against what the user declared.
+	declared := *installation.Spec.DeepCopy()
+	computed := installation.Status.Computed.DeepCopy()
```
caseydavenport: `computed` already has the overlay merged in, so `recordDefaults` diffs post-overlay against the pre-overlay declared spec and writes the overlay's pools into `status.defaults` on the base Installation, where they survive deleting the overlay. The core deliberately derives `recordedDefaults` before its overlay merge; we should do the same here.

pkg/controller/utils/defaults.go
```diff
@@ -0,0 +1,243 @@
...
+	}
+
+	recorded := map[string]operatorv1.IPPool{}
+	for _, pool := range defaults.CalicoNetwork.IPPools {
+		recorded[pool.CIDR] = pool
+	}
+
+	for i := range spec.CalicoNetwork.IPPools {
+		declared := spec.CalicoNetwork.IPPools[i]
+		base, ok := recorded[declared.CIDR]
```
caseydavenport: Keyed on the raw CIDR string, but the ippool controller normalizes CIDRs precisely because the Installation can hold a non-canonical spelling. A user writing `192.168.0.1/16` misses the lookup, we `continue`, and `status.computed` publishes a half-defaulted pool. We should normalize on both sides of this map.

pkg/controller/utils/defaults.go
```diff
@@ -0,0 +1,243 @@
...
+	return base, nil
+}
+
+// addedKeys walks two decoded specs and keeps only what defaulted added on top of declared.
+func addedKeys(declared, defaulted map[string]any) map[string]any {
+	added := map[string]any{}
+	for key, defaultedValue := range defaulted {
+		declaredValue, present := declared[key]
+		if !present {
+			if isVacuous(defaultedValue) {
```
caseydavenport: `isVacuous` drops empty lists, which kills the one default the `omitempty` removal is there to preserve: `fillDefaults` sets `IPPools = []IPPool{}` to mean pools are managed out-of-band. That never reaches `status.defaults`, so `status.computed.calicoNetwork.ipPools` comes back unset rather than explicitly empty, and the nil-vs-empty distinction is invisible to every consumer.

pkg/controller/utils/defaults.go
```diff
@@ -0,0 +1,243 @@
...
+		defaultedObject, defaultedIsObject := defaultedValue.(map[string]any)
+		if declaredIsObject && defaultedIsObject {
+			if nested := addedKeys(declaredObject, defaultedObject); len(nested) > 0 {
+				added[key] = nested
+			}
+			continue
+		}
+
+		// Lists are all-or-nothing; per-element merging would need a merge key we don't have.
+		if !reflect.DeepEqual(declaredValue, defaultedValue) {
```
caseydavenport: All-or-nothing on lists means a user's own declared pool CIDRs get recorded as operator defaults. Remove `ipPools` from the manifest afterwards and `OverrideInstallationSpec` restores the recorded list, so the pools come back and we recreate them. Same ownership problem this PR fixes for the spec, moved into the status, and it applies to every list field.

pkg/controller/utils/defaults.go
```diff
@@ -0,0 +1,243 @@
...
+	content := map[string]any{}
+	if err := json.Unmarshal(raw, &content); err != nil {
+		return nil, fmt.Errorf("unmarshal installation spec: %w", err)
+	}
+	pruneNulls(content)
+	return content, nil
+}
+
+// pruneNulls drops null fields, which the API server prunes too, so they can't read as defaults.
+func pruneNulls(content map[string]any) {
```
caseydavenport: `pruneNulls` doesn't recurse into array elements, so the pool objects inside `calicoNetwork.ipPools` keep their explicit nulls. `Defaults` is Schemaless with preserve-unknown-fields so the API server doesn't prune them either, which makes the parity claim in the comment wrong and leaves `addedKeys` comparing null-bearing lists with `DeepEqual`.

pkg/controller/installation/core_controller.go
```diff
@@ -853,19 +874,14 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 		utils.SetInstallationFinalizer(instance, render.OperatorCompleteFinalizer)
 	}
 
-	// Update CRDs before persisting defaults. Defaulting can set a value only this operator version's
-	// CRD accepts (e.g. an autodetected kubernetesProvider=Kind); on upgrade the old served CRD would
-	// otherwise reject the write and the reconcile would loop before ever reaching the CRD update.
 	if err = r.updateCRDs(ctx, r.opts.Variant, reqLogger); err != nil {
```
caseydavenport: We dropped the comment recording why the CRD update has to come before the write: defaulting can produce a value only this operator version's CRD accepts, with autodetected `kubernetesProvider: Kind` as the case. That constraint moved to the status write below, which is later still and fully schema-validated, so we should keep the note here.

pkg/controller/installation/core_controller.go
```diff
@@ -911,6 +927,19 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 		}
 	}
 
+	// Publish the effective config before the IP pool gate below, since other
+	// controllers read it instead of the spec. Once status.Computed is present, other
+	// controllers will kick in.
+	computed := defaulted.Spec.DeepCopy()
+	if !reflect.DeepEqual(instance.Status.Defaults, recordedDefaults) || !reflect.DeepEqual(instance.Status.Computed, computed) {
```
caseydavenport: We pass `recorded=nil` with no owned paths, so every reconcile rewrites all of `status.defaults` from what it just derived, while the ippool controller writes the same subtree independently. Line 923 also blanks the whole status on a first reconcile, which can drop what ippool just recorded. Convergence relies on every default being re-derivable from `defaulted.Spec`. Do we want to scope this write to the paths the core owns instead?

api/v1/installation_types.go
```diff
@@ -1135,6 +1135,13 @@ type InstallationStatus struct {
 	// +optional
 	Computed *InstallationSpec `json:"computed,omitempty"`
 
+	// Defaults holds fields the operator supplied because the spec omitted them
+	// (schemaless, so schema defaults can't inflate).
+	// +optional
+	// +kubebuilder:validation:Schemaless
+	// +kubebuilder:pruning:PreserveUnknownFields
+	Defaults *InstallationSpec `json:"defaults,omitempty"`
```
caseydavenport: `Defaults` is Schemaless and `Computed` isn't, so the API server injects kubebuilder defaults into one and not the other. `IPPool.DisableBGPExport` has `default:=false` and nothing in Go sets it, so the `DeepEqual` on `Computed` fails every reconcile and we rewrite the status, waking every Installation watcher, in a loop.

pkg/controller/utils/utils.go
```diff
@@ -505,28 +505,28 @@ func GetInstallationStatus(ctx context.Context, client client.Client) (*operator
...
+// ComputedInstallationResource names what a caller is waiting for when the core controller has
+// not published the effective config yet.
+var ComputedInstallationResource = schema.GroupResource{
+	Group:    operatorv1.GroupVersion.Group,
+	Resource: "installations.status.computed",
+}
+
+// GetComputedInstallationSpec returns the effective config the core controller publishes on the
+// status, which already has defaults and the overlay applied.
+func GetComputedInstallationSpec(ctx context.Context, client client.Client) (*operatorv1.InstallationSpec, error) {
```
caseydavenport: Every caller now depends on the core reaching its status write, and there are several early returns above it (`updateCRDs`, the spec patch, overlay validation, the admission policy updates). Any of those leaves `status.computed` stale or absent and stalls every other controller at once, where before each one read spec plus overlay itself and could make progress independently.

pkg/controller/podiprecovery/podiprecovery_controller.go
```diff
@@ -330,7 +330,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 
 	// Gate on Installation: if Calico hasn't been installed yet, the
 	// operator-managed pods we'd act on don't exist. Bail out silently.
-	if _, err := utils.GetInstallationSpec(ctx, r.client); err != nil {
+	if _, err := utils.GetComputedInstallationSpec(ctx, r.client); err != nil {
```
caseydavenport: This controller has no Installation watch, unlike applicationlayer, kubeproxy and goldmane, so an event dropped on the NotFound below is never retried. A host IP change in the window before `status.computed` is published leaves the stale host-networked pod around until some unrelated Node event. We should add `utils.AddInstallationWatch` here.

pkg/controller/secrets/cluster_ca_controller_test.go
```diff
@@ -82,13 +82,12 @@ var _ = Describe("ClusterCA controller", func() {
...
 			},
-			Status: operatorv1.InstallationStatus{
-				Computed: &operatorv1.InstallationSpec{},
-			},
+			Status: operatorv1.InstallationStatus{},
 			Spec: operatorv1.InstallationSpec{
 				Variant: operatorv1.Calico,
 			},
 		}
+		install.Status.Computed = &install.Spec
```
caseydavenport: `&install.Spec` aliases the two, so this test can't distinguish a controller reading `Status.Computed` from one reading `Spec`. Same in `whisker/controller_test.go` and `tenantsecrets/tenant_controller_test.go`. We should diverge them the way the test below does, otherwise a missed migration still passes.

pkg/controller/secrets/cluster_ca_controller_test.go
```diff
@@ -134,21 +133,14 @@ var _ = Describe("ClusterCA controller", func() {
...
-				Name: "overlay",
-			},
-			Spec: operatorv1.InstallationSpec{
-				Variant: operatorv1.CalicoEnterprise,
-			},
-		}
-		Expect(cli.Create(ctx, &install)).ShouldNot(HaveOccurred())
+	// The broken Calico ImageSet fails to load if the reconciler reads the spec
+	// instead of the status.
+	It("should Reconcile from the computed spec", func() {
```
caseydavenport: This was the only test exercising an overlay Installation through the shared getter, and the shared getter no longer merges overlays, so nothing covers that path. We should keep an overlay case somewhere, either here or against the core's status write.

pkg/controller/istio/istio_controller.go
```diff
@@ -387,7 +387,7 @@ func (r *ReconcileIstio) configurePolicySyncPathPrefix(ctx context.Context, inst
 		// installationSpec.Variant (i.e. Installation.Spec.Variant), so the
 		// policy-sync field tracks the renderer's decision to ship the L7
 		// waypoint sidecar even before Status.Variant catches up.
-		installationSpec, err := utils.GetInstallationSpec(ctx, r.Client)
+		installationSpec, err := utils.GetComputedInstallationSpec(ctx, r.Client)
```
caseydavenport: Comment above is stale once this reads `Status.Computed`. The race it describes, the spec variant leading `Status.Variant`, is gone, since the core writes `Computed` and `Variant` in the same status update.

pkg/controller/ippool/validation.go
```diff
@@ -23,16 +23,16 @@ import (
...
 	names := map[string]bool{}
-	for _, pool := range instance.Spec.CalicoNetwork.IPPools {
+	for _, pool := range spec.CalicoNetwork.IPPools {
 		_, cidr, err := net.ParseCIDR(pool.CIDR)
 		if err != nil {
 			return fmt.Errorf("IP pool CIDR (%s) is invalid: %s", pool.CIDR, err)
 		}
 
-		// Validate that there is only a single instance of each CIDR and Name.
+		// Validate that there is only a single spec of each CIDR and Name.
```
caseydavenport: The rename caught the English here: this was "a single instance of each CIDR and Name".

test/pool_test.go
```diff
@@ -242,10 +242,21 @@ var _ = Describe("IPPool FV tests", func() {
...
+			}
+			if instance.Status.Computed == nil || instance.Status.Computed.CalicoNetwork == nil {
+				return fmt.Errorf("Installation defaults have not been recorded yet")
+			}
+			pools := instance.Status.Computed.CalicoNetwork.IPPools
+			if len(pools) != 1 || pools[0].Name == "" {
+				return fmt.Errorf("IP pool defaults have not been recorded yet: %+v", pools)
+			}
+			return nil
+		}, 30*time.Second, 1*time.Second).ShouldNot(HaveOccurred())
```
caseydavenport: 6x on the timeout in both cases is measuring something real: first-install pool defaulting now needs two controllers and at least two status round-trips instead of one spec write. Everything gated on `status.computed.calicoNetwork.ipPools` sees that longer window on first install.

pkg/controller/installation/windows_controller.go
```diff
@@ -239,45 +245,19 @@ func (r *ReconcileWindows) Reconcile(ctx context.Context, request reconcile.Requ
 		return reconcile.Result{}, err
 	}
 
-	// We rely on the core controller for defaulting, so wait until it has done so before continuing
-	if reflect.DeepEqual(instanceStatus, operatorv1.InstallationStatus{}) {
-		err := fmt.Errorf("InstallationStatus is empty")
-		r.status.SetDegraded(operatorv1.ResourceNotReady, "InstallationStatus is empty", err, reqLogger)
```
tmjd, line 245: We use to degrade in this case, should we do the same now when status.Computed is not set?

pkg/controller/ippool/pool_controller.go
```diff
@@ -144,6 +143,27 @@ func hasOwnerLabel(pool *v3.IPPool) bool {
...
+	recorded, err := utils.MergeRecordedDefaults(installation.Status.Defaults, declared, computed,
+		utils.DefaultsScope{Owned: []string{utils.PoolDefaultsPath}})
+	if err != nil {
+		return err
+	}
+	if reflect.DeepEqual(installation.Status.Defaults, recorded) {
+		return nil
+	}
+
+	// Write a copy, so the response can't replace the effective spec we're working from.
```
tmjd, line 157: That's interesting that this is needed. I wouldn't have thought that Status().Update would not have updated the main spec in what was passed to it.

pkg/controller/ippool/pool_controller.go
```diff
@@ -201,21 +234,19 @@ func (r *Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
 		}
 	}
 
-	// Write default IP pool configuration back to the Installation object using patch.
-	preDefaultPatchFrom := client.MergeFrom(installation.DeepCopy())
```
tmjd, line 205: Any idea why we patched before? I'm wondering if we should still be patching for updating the defaults

tmjd (commented):
All the code for having the IPPools managed by the ippool controller is quite complicated. I'm not a big fan of that.

tmjd (approved):
LGTM

pkg/controller/installation/windows_controller.go
caseydavenport: I don't think we need to degrade here, it (1) should be a transient condition and (2) isn't really meaningfully actionable by the user.

We do skip the reconcile and requeue above though, so we're still gating here.

pkg/controller/ippool/pool_controller.go
caseydavenport: Yeah, it just unpacks the whole returned object and that includes the spec.

pkg/controller/ippool/pool_controller.go
caseydavenport: I switched it back to Patch - I think that's the correct operation.

PR #5206, opened 2026-08-17
Move Enterprise-only webhook rendering into the extension

caseydavenport:
## Description

Follow-up to #5164, moving the rest of the Enterprise-only webhook rendering behind the extensions boundary. The base render now produces the same objects for both variants, and the Enterprise extension layers on:

- audit logging, both the webhook registration and the host directory the container writes it to
- the mutating webhook configuration and the UISettings webhook in it
- permissions on managed clusters and UI settings groups
- a root (and on OpenShift, privileged) security context on the container

When the operator runs as Calico, the mutating webhook configuration is queued for deletion so an install downgraded from Enterprise cleans up after itself.

One behavior change worth calling out: Calico installs no longer mount the audit log host directory into the webhooks container. Only the Enterprise audit handler ever wrote there, so it was a mounted-but-unused hostPath, but dropping it does roll the webhooks deployment on upgrade.

## Release Note

```release-note
None
```

### DIFF
--- a/pkg/enterprise/apiserver/webhooks.go
+++ b/pkg/enterprise/apiserver/webhooks.go
@@ -44,15 +44,14 @@ const (
 // modifyWebhooks layers audit logging, the mutating webhooks, the extra RBAC, and
 // management-cluster support onto the rendered objects.
 func modifyWebhooks(cfg *webhooks.Configuration, managementCluster *operatorv1.ManagementCluster, multiTenant bool, create, del []client.Object) ([]client.Object, []client.Object) {
-	if dep, ok := extensions.FindObject[*appsv1.Deployment](create, webhooks.WebhooksName); ok {
-		enableAuditLogging(dep, cfg.Installation)
-	}
-	if vwc, ok := extensions.FindObject[*admregv1.ValidatingWebhookConfiguration](create, webhookConfigName); ok {
-		vwc.Webhooks = append(vwc.Webhooks, auditLoggingWebhook(cfg))
-	}
-	if cr, ok := extensions.FindObject[*rbacv1.ClusterRole](create, webhooks.WebhooksName); ok {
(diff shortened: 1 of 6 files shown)


pkg/enterprise/apiserver/webhooks.go
```diff
@@ -28,35 +29,183 @@ import (
...
+	webhookConfigName = "api.projectcalico.org"
+
+	webhookAuditLogVolumeName = "audit-logs"
+	webhookAuditLogPath       = "/var/log/calico/audit"
+)
+
+// modifyWebhooks layers audit logging, the mutating webhooks, the extra RBAC, and
+// management-cluster support onto the rendered objects.
 func modifyWebhooks(cfg *webhooks.Configuration, managementCluster *operatorv1.ManagementCluster, multiTenant bool, create, del []client.Object) ([]client.Object, []client.Object) {
+	if dep, ok := extensions.FindObject[*appsv1.Deployment](create, webhooks.WebhooksName); ok {
```
Brian-McM: We probably want to panic or be very very load if we couldn't find an object through the extensions.

Brian-McM (changes requested):

pkg/enterprise/apiserver/webhooks.go
caseydavenport: Added `extensions.MustFindObject`, which panics with `BUG: no object named %q to modify` when the base render did not produce it, mirroring `render.MustContainer` in `pkg/render/containers.go`. Converted the four lookups here that are always guaranteed to hit.

Your point applies to the whole mechanism though - there are about 20 other `if ok` sites across `pkg/enterprise/` that silently no-op the same way. I'll do those in a separate PR, minus `guardian.go:46`, `kubecontrollers.go:62`, and `apiserver/extension.go:733`, which return early on a miss deliberately.

Brian-McM (approved):

PR #5212, opened 2026-08-17
Allow konnectivity agents to reach the Calico webhooks

caseydavenport:
## Description

On AKS and GKE the API server reaches in-cluster webhooks through konnectivity agent pods, so the admission call arrives at `calico-webhooks` with a pod source IP. The policy we render denies every pod source on the webhook port, and the allow that precedes it matches the `default/kubernetes` service, whose endpoint is the public control plane address. Result: every `projectcalico.org/v3` write times out on the tiered RBAC webhook.

This adds an allow for the konnectivity agents ahead of the deny. AKS labels them `app: konnectivity-agent` and GKE uses `k8s-app`, so the selector matches either.

Found while running the e2e suite against AKS 1.36.2 with the v3 CRDs served directly. Adding this rule by hand took the run from 74 failures to zero.

## Release Note

```release-note
Fix Calico webhook admission requests being denied on clusters where the API server connects through konnectivity, including AKS and GKE.
```

### DIFF
--- a/pkg/render/webhooks/render.go
+++ b/pkg/render/webhooks/render.go
@@ -239,11 +239,11 @@ func (c *component) Objects() ([]client.Object, []client.Object) {
 			},
 		)
 
-		// TODO: Calico policy can't match "any node" as a source. Such a selector would
-		// let us drop the fallback allow.
+		// TODO: Calico policy can't match "any node" or "any tunnel IP" as a source. Such
+		// a selector would let us allow nodes explicitly and drop the fallback allow.
 		ingressRules := []v3.Rule{
-			// The apiserver source IP usually survives to the webhook pod and matches the
-			// kubernetes Service endpoints.
+			// The apiserver source IP usually survives unmodified to the webhook pod, where it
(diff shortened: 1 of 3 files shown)


pkg/render/webhooks/render.go
```diff
@@ -243,21 +243,25 @@ func (c *component) Objects() ([]client.Object, []client.Object) {
 		// This covers the common case where the apiserver source IP survives unmodified
 		// to the webhook pod and matches the kubernetes Service's endpoints.
 		//
-		// Rule 2: Deny traffic from any Kubernetes workload endpoint. The selector
+		// Rule 2: Allow traffic from the konnectivity agents. On AKS and GKE the apiserver
```
tmjd: I know you're matching what was here before but is there any reason not to put these comments inline with each of the rules that they are about?

tmjd (approved):

pkg/render/webhooks/render.go
caseydavenport: Yeah, that is nicer - I'll do that!

PR #5223, opened 2026-08-19
refactor(networkpolicy): consolidate egress destination parsing

tianfeng92:
## Description

`ParseExternalDestination` and `ExternalDestinationEntityRule` already turn an endpoint into the tightest egress rule available. Only the OTel Collector used them. Dex and Guardian carried their own copies; both now go through the shared pair.

Dex's `parseHostPortFromURL` was a duplicate of `pkg/url.ParseHostPortFromHTTPProxyURL`, which Guardian already used, so it is deleted rather than reimplemented.

**Preserved:** Dex's `0.0.0.0/0` and `::/0` any-destination rules, and its unconditional domain rule — it has no license input.

**Changed:** a `<svc>.<ns>.svc` destination now renders a Services match instead of a Domains rule, so it follows the Service's own ports. A Domains rule never worked for a ClusterIP. Only the domain branch needs `egress-access-control`; a Services match does not.

## Testing

- lint clean; render, enterprise, clusterconnection and authentication suites pass
- new test covers an unlicensed in-cluster destination
- draft until verified on a cluster: Dex reaching its IdP, Guardian's tunnel, with and without a proxy

## Release Note

```release-note
None
```

EV-6963


pkg/enterprise/clusterconnection/guardian.go
pasanw, line 1: A large portion of my feedback is about the `ParseExternalDestination` implementation, which I should have given to you in the previous PR - apologies that I missed that.

I'll give that feedback here:
* `clusterService` logic doesn't look quite right - if it's passed something like proxy.corp.svc.example.com it will treat it as a kubernetes service. I think this function should be strict and only parse service name and namespace if the host string ends with `.svc`, `.svc.<cluster-domain>`, or `.svc.<cluster-domain>.`. This scheme doesn't support if a customer specifies their service without a .svc based suffix, but I couldn't think of a clean way to support that - we can brainstorm on that if you'd like
* `ExternalDestinationEntityRule` results in a fail-open style of policy, where if we aren't able to restrict the host in the egress rule (with a Domains or Nets field), we return an egress rule that allows all hosts. I don't think a general policy helper should be making a decision to open broad access by default - that kind of decision should be made by the component. (And I don't think components should be choosing broad access in such a scenario, including otel collector)
* Guardian, Dex, and other components all utilized the flow of `SplitHostPort` > `PortFromString` > `ParseIP`. Looks like we replaced `PortFromString` with `Atoi` - could we revert that? `Atoi` doesn't support all of the different scenarios that `PortFromString` does and in general I think it's better to not change longstanding logic
* I think it would be nicer for components to provide their destination string to one function and get back an `EntityRule`. This way it's only one function call and no new concept (`ExternalDestination`) needs to be introduced. I'd recommend structuring it as such:
    * `func X(destination string) v3.EntityRule`, which itself calls `func Y(destination string) (host string, port string)` and then `func Z(host string, port string) v3.EntityRule` to do its work
    * OTel, Guardian, Dex would use `X` directly to get their entity rule
    * Manager and ServiceEndpoint would use `Z` to get their entity rule

pasanw, line 1: General comment on this PR: looks like we are missing the opportunity for manager.go (`managerComponent`) and k8s-endpoint.go (`ServiceEndpoint`) to utilize our shared logic. They could use the function `Z`, as mentioned above

pkg/enterprise/clusterconnection/guardian.go
```diff
@@ -131,44 +130,27 @@ func enterpriseGuardianPolicySpec(gpc *render.GuardianConfiguration) (v3.Network
...
-			})
-			allowedDestinations[tunnelDestinationHostPort] = true
+
+		// Only a domain destination needs the EgressAccessControl feature; an IP or
+		// an in-cluster Service does not. Ask for the rule with the feature we
+		// actually have, and skip it only when that left nothing to match on --
+		// the Pass rule below already governs the tunnel, and a port-only allow
+		// here would be wider than what ships today.
+		dest := networkpolicy.ExternalDestinationEntityRule(parsed, gpc.IncludeEgressNetworkPolicy)
+		if dest.Services == nil && len(dest.Nets) == 0 && len(dest.Domains) == 0 {
```
pasanw: Seems like this is a large guard trying to detect if the fail-open I mentioned in the larger comment occurred? If so, I would hope that we could remove it as a result of removing the fail-open behaviour

pkg/enterprise/clusterconnection/guardian.go
```diff
@@ -131,44 +130,27 @@ func enterpriseGuardianPolicySpec(gpc *render.GuardianConfiguration) (v3.Network
 			continue
 		}
 
-		host, port, err := net.SplitHostPort(tunnelDestinationHostPort)
-		if err != nil {
-			return v3.NetworkPolicySpec{}, err
+		parsed, ok := networkpolicy.ParseExternalDestination(tunnelDestinationHostPort)
```
pasanw: Claude: This replaced `net.SplitHostPort`, which rejected anything that was not `host:port`. `ParseExternalDestination`'s `url.Parse` fallback now silently accepts a URL-shaped `gpc.URL` — e.g. `https://mgmt.example.com` parses as host `mgmt.example.com` port **443** (the scheme default), not the tunnel port. So a misconfig that used to fail the render loudly now ships a rule for the wrong port. Note the `// gpc.URL has host:port form` comment just above is no longer enforced. Since — unlike OTel's endpoint — this field is not meant to be a URL, Guardian should keep rejecting non-`host:port` here (a strict entry point, or validate before parsing).

pkg/render/dex.go
```diff
@@ -563,7 +562,7 @@ func resolveEgressDestinationsForPod(podProxy *httpproxy.Config) ([]string, erro
 			return nil, err
 		}
 
-		httpProxyDestination, err := parseHostPortFromURL(httpProxyURL)
+		httpProxyDestination, err := operatorurl.ParseHostPortFromHTTPProxyURL(httpProxyURL)
```
pasanw: Claude: Robustness note — pre-existing, not introduced here, low priority. Swapping to `ParseHostPortFromHTTPProxyURL` (here and at line 579) widens the set of proxy values that error out: it rejects any non-http/https scheme up front, whereas the old `parseHostPortFromURL` accepted anything that carried a port. The problem is what an error here costs: `resolveEgressDestinationsForPod` returns the error *before* the `0.0.0.0/0`/`::/0` catch-alls are appended, and the caller (`resolveEgressRulesByDestination`) then `continue`s the **whole pod** — so a single unparseable proxy value silently drops that pod's *entire* egress, IdP catch-alls included (log line only, status stays healthy). The ordering and the whole-pod skip both predate this PR, and the newly-rejected schemes (e.g. `socks5://`, scheme-relative `//host:port`) aren't a supported proxy type anyway (Guardian's tunnel dialer is HTTP-CONNECT-only), so this is hardening, not a regression. If we want to harden it: append the catch-alls unconditionally, and handle a proxy-parse failure locally (skip just that one destination) so a bad proxy value can't wipe the broad allows.

pasanw (changes requested):

pkg/enterprise/clusterconnection/guardian.go
```diff
@@ -131,44 +129,21 @@ func enterpriseGuardianPolicySpec(gpc *render.GuardianConfiguration) (v3.Network
...
-				Action:   v3.Allow,
-				Protocol: &networkpolicy.TCPProtocol,
-				Destination: v3.EntityRule{
-					Nets:  []string{parsedIP.String() + netSuffix},
-					Ports: []numorstring.Port{parsedPort},
-				},
-			})
-			allowedDestinations[tunnelDestinationHostPort] = true
+
+		if len(dest.Domains) > 0 && !gpc.IncludeEgressNetworkPolicy {
```
pasanw: confusing name

pkg/render/common/networkpolicy/networkpolicy.go
```diff
@@ -339,87 +338,73 @@ var PrometheusSourceEntityRule = v3.EntityRule{
...
-		if p, err := strconv.Atoi(portStr); err == nil && p > 0 && p <= 65535 {
-			return ExternalDestination{Host: host, Port: uint16(p)}, true
+// EntityRuleForHostPort builds the tightest destination rule for a host: an
+// exact net for a literal IP, otherwise the domain. It always constrains the
+// destination, never returning a ports-only rule.
+//
+// Service matching is opt-in via clusterDomain, since a Services match cannot
+// carry ports.
+func EntityRuleForHostPort(host, clusterDomain string, ports ...numorstring.Port) v3.EntityRule {
+	if clusterDomain != "" {
```
pasanw: we should always be passing a cluster domain and detecting if the host is a svc. where do we conditionally pass?

pkg/render/common/networkpolicy/networkpolicy.go
```diff
@@ -339,87 +338,73 @@ var PrometheusSourceEntityRule = v3.EntityRule{
...
-	if err != nil {
-		return ExternalDestination{}, false
-	}
-	if portStr := u.Port(); portStr != "" {
-		if p, err := strconv.Atoi(portStr); err == nil && p > 0 && p <= 65535 {
-			return ExternalDestination{Host: u.Hostname(), Port: uint16(p)}, true
+
+	rule := v3.EntityRule{Ports: ports}
+	if ip := net.ParseIP(host); ip != nil {
+		suffix := "/128"
```
pasanw: inverted, ideally /32 default

pkg/render/common/networkpolicy/networkpolicy.go
```diff
@@ -339,87 +338,73 @@ var PrometheusSourceEntityRule = v3.EntityRule{
...
 		}
+		rule.Nets = []string{ip.String() + suffix}
+		return rule
 	}
-	switch u.Scheme {
-	case "https":
-		return ExternalDestination{Host: u.Hostname(), Port: 443}, true
-	case "http":
-		return ExternalDestination{Host: u.Hostname(), Port: 80}, true
+	rule.Domains = []string{host}
```
pasanw: Is this how the existing logic handles it? if parse ip fails just raw domain?

pkg/render/common/networkpolicy/networkpolicy.go
```diff
@@ -339,87 +338,73 @@ var PrometheusSourceEntityRule = v3.EntityRule{
...
-// clusterService splits an in-cluster Service DNS name --
-// <service>.<namespace>.svc[.cluster.local] -- into its namespace and name.
-// Each element is checked with the upstream DNS-label validator rather than a
-// pattern of our own.
-func clusterService(host string) (namespace, name string, ok bool) {
+// ClusterServiceWithDomain reports the namespace and name of an in-cluster
+// Service DNS name: <service>.<namespace>.svc, optionally followed by the
+// cluster domain and a trailing dot. Anything longer is an external host that
+// merely carries an "svc" label.
+func ClusterServiceWithDomain(host, clusterDomain string) (namespace, name string, ok bool) {
```
pasanw: review closer

pkg/render/otelcollector/component.go
```diff
@@ -832,25 +831,27 @@ func (c *component) statefulSet() *appsv1.StatefulSet {
...
+		return v3.EntityRule{}, err
+	}
+	host := u.Hostname()
+	if host == "" {
+		return v3.EntityRule{}, fmt.Errorf("exporter endpoint %q has no host", exp.Endpoint)
 	}
-	port := uint16(OTLPGRPCPort)
-	if exp.Protocol == operatorv1.OpenTelemetryProtocolHTTP {
-		port = OTLPHTTPPort
+	portStr := u.Port()
```
pasanw: default to https, on what basis?

pkg/render/dex.go
```diff
@@ -546,7 +539,7 @@ func (c *dexComponent) resolveEgressRulesByDestination() map[string]v3.Rule {
 // resolveEgressDestinationsForPod collects all possible http proxy destinations and all possible IdP destinations.
 // In the future, this function may return only the specific destinations it expects Dex pods to connect to given the
 // current issuer configuration (in the Authentication CR) and the HTTP proxy configuration.
-func resolveEgressDestinationsForPod(podProxy *httpproxy.Config) ([]string, error) {
+func resolveEgressDestinationsForPod(podProxy *httpproxy.Config) []string {
```
pasanw: lots of change to a high danger function. understand this but probably back out?

pkg/render/otelcollector/component.go
```diff
@@ -832,25 +831,27 @@ func (c *component) statefulSet() *appsv1.StatefulSet {
...
+	if portStr == "" {
+		portStr = "443"
 	}
```
pasanw: Not sure I understand what's happening here, why do we default to 443?

pkg/render/dex.go
```diff
@@ -557,38 +550,27 @@ func resolveEgressDestinationsForPod(podProxy *httpproxy.Config) ([]string, erro
...
-		httpsProxyURL, err := url.Parse(podProxy.HTTPSProxy)
+		proxyURL, err := url.Parse(proxy)
 		if err != nil {
-			return nil, err
+			log.Error(err, fmt.Sprintf("ignoring unparseable proxy %q when rendering Dex egress", proxy))
+			continue
 		}
-
-		httpsProxyDestination, err := parseHostPortFromURL(httpsProxyURL)
+		destination, err := operatorurl.ParseHostPortFromHTTPProxyURL(proxyURL)
```
pasanw: To be safe I think we'll need to keep the local `parseHostPortFromURL`. The function that we've switched to on this line only tolerates HTTP(S) proxies. That works for Guardian since it only supports HTTP(S) proxies, but it looks like Dex could support socks proxies. Technically `parseHostPortFromURL` could support socks proxies if the port is specified in the URL so to be safe I think it's best we don't change behaviour

pkg/render/common/networkpolicy/networkpolicy.go
pasanw: The right behaviour here would be that we always pass clusterDomain in - could we ensure that all callers of `EntityRuleForHostPort` and `EntityRuleForDestination` do that?

pkg/render/otelcollector/component.go
```diff
@@ -861,10 +862,14 @@ func (c *component) networkPolicy() *v3.NetworkPolicy {
 	// ports: the exporters below are the only egress destinations we need, and a
 	// catch-all would let the collector reach any host on 4317/4318.
 	for _, exp := range c.cfg.OpenTelemetry.Exporters {
```
pasanw: Background: Calico Enterprise is licensed. One thing the license can turn on is *domain-based egress* — allow-rules that name a DNS domain instead of an IP.

Before this PR, the operator only emitted domain rules for OTel when the license included the `EgressAccessControl` feature. This PR deletes that check. The loop here now emits domain rules for every exporter, regardless of the license.

The catch: Guardian, changed in this same PR, still keeps the check. So one PR both drops and keeps the same gate.

Two ways this bites a cluster without the feature:
- On a current apiserver it silently gains domain egress it isn't entitled to.
- On an older apiserver that still enforces the feature, the **whole** policy is rejected and the controller degrades. The old code fell back to a ports-only rule and kept working.

Please gate OTel the same way Guardian does, or drop Guardian's gate too — and say which, and why, in the PR description.

pkg/render/dex.go
```diff
@@ -557,38 +550,27 @@ func resolveEgressDestinationsForPod(podProxy *httpproxy.Config) ([]string, erro
...
-
-		egressDestinations = append(egressDestinations, httpProxyDestination)
-	}
-
-	if podProxy.HTTPSProxy != "" {
-		httpsProxyURL, err := url.Parse(podProxy.HTTPSProxy)
+		proxyURL, err := url.Parse(proxy)
 		if err != nil {
-			return nil, err
+			log.Error(err, fmt.Sprintf("ignoring unparseable proxy %q when rendering Dex egress", proxy))
```
pasanw: Claude: `HTTP_PROXY` / `HTTPS_PROXY` are URLs, and a proxy URL may carry credentials — e.g. `http://user:pass@proxy:3128`.

This new line logs the raw proxy value with `%q`. So when a proxy fails to parse, its username and password go into the operator log. The sibling line just below (skipping a non-HTTP scheme) does the same.

Log a redacted form instead — `proxyURL.Redacted()`, or just the `host:port` — on both lines.

pkg/render/otelcollector/component.go
```diff
@@ -861,10 +862,14 @@ func (c *component) networkPolicy() *v3.NetworkPolicy {
...
+		dest, err := exporterDestination(exp, c.cfg.ClusterDomain)
+		if err != nil {
+			continue
```
pasanw, line 859: We should log at error level when we fail to create a rule for an exporter destination

pkg/render/otelcollector/component.go
```diff
@@ -861,10 +862,14 @@ func (c *component) networkPolicy() *v3.NetworkPolicy {
...
+		}
 		egressRules = append(egressRules, v3.Rule{
 			Action:      v3.Allow,
 			Protocol:    &networkpolicy.TCPProtocol,
-			Destination: networkpolicy.ExternalDestinationEntityRule(exporterDestination(exp), c.cfg.DomainEgressAllowed),
+			Destination: dest,
 		})
 	}
```
pasanw: Background: Calico Enterprise is licensed. One thing the license can turn on is *domain-based egress* — allow-rules that name a DNS domain instead of an IP.

Before this PR, the operator only named a domain when the license carried the `EgressAccessControl` feature; without it, the old helper fell back to a ports-only rule. This PR deletes that gate. The loop now emits a `Domains` rule for every exporter, regardless of the license.

Guardian — changed in this same PR — still gates on the identical feature (`clusterconnection/extension.go:138`). So one PR both drops and keeps the same gate.

Two ways this bites a cluster without the feature:
- On a current apiserver it silently gains domain egress it isn't entitled to.
- On an older apiserver that still enforces the feature, the **whole** policy is rejected and the controller degrades.

Suggested fix — restore the gate, skipping the domain rule when unlicensed (IP and in-cluster-service exporters are unaffected, and we don't reintroduce a blanket port allow):

```suggestion
	for _, exp := range c.cfg.OpenTelemetry.Exporters {
		dest, err := exporterDestination(exp, c.cfg.ClusterDomain)
		if err != nil {
			continue
		}
		// Domain-based egress requires the EgressAccessControl license feature.
		// Without it, skip domain-named exporters instead of emitting a Domains
		// rule the apiserver may reject (or that grants egress the license does
		// not cover) — mirroring how Guardian gates the same feature.
		if len(dest.Domains) > 0 && !c.cfg.DomainEgressAllowed {
			continue
		}
		egressRules = append(egressRules, v3.Rule{
			Action:      v3.Allow,
			Protocol:    &networkpolicy.TCPProtocol,
			Destination: dest,
		})
	}
```

This also needs the deleted `DomainEgressAllowed` field restored on the config struct and recomputed in the controller via `utils.IsFeatureActive(license, common.EgressAccessControlFeature)` — the same call Guardian uses. Alternatively, if domain egress no longer needs the license, drop Guardian's gate instead and note it in the PR description.

pkg/render/otelcollector/component.go
```diff
@@ -861,10 +862,14 @@ func (c *component) networkPolicy() *v3.NetworkPolicy {
...
 	for _, exp := range c.cfg.OpenTelemetry.Exporters {
+		dest, err := exporterDestination(exp, c.cfg.ClusterDomain)
+		if err != nil {
+			continue
+		}
 		egressRules = append(egressRules, v3.Rule{
 			Action:      v3.Allow,
 			Protocol:    &networkpolicy.TCPProtocol,
-			Destination: networkpolicy.ExternalDestinationEntityRule(exporterDestination(exp), c.cfg.DomainEgressAllowed),
+			Destination: dest,
```
pasanw, line 864: Looks like we dropped the domain egress check, could we add it back into this function similar to how it is done in guardian (`if len(dest.Domains) > 0 && !gpc.IncludeEgressNetworkPolicy...`)?

tianfeng92: https://github.com/tigera/operator/pull/5223/changes#diff-4b08b3b19dde288685e682b4b7ab31da5775d3906bf8ead59f9f3a10812275a9R857 put it back here.

pasanw (approved):
Thanks for taking this on!

PR #5224, opened 2026-08-19
Add ingestionCompression to LogCollector and NonClusterHost, and compress fluent-bit log ingestion to Linseed

hjiawei:
## Description

The fluent-bit outputs that ship logs to Linseed now compress each batch — zstd by default — cutting log ingestion bandwidth by roughly **12x** measured on a kubeadm cluster, and both LogCollector and NonClusterHost gain an `ingestionCompression` field to control it.

- `ingestionCompression: None | Gzip | Zstd`, default `Zstd` through the CRD (`+kubebuilder:default`), which the API server fills on every read — so no reconciler write-back is needed, matching the other CRD-defaulted fields in this API.
- **LogCollector** drives the in-cluster outputs: the rendered `compress` follows the field, and `None` omits it so `out_http` posts plain NDJSON. Only the Linseed `http` outputs carry it — S3, Splunk, Syslog and OpenTelemetry are unaffected, asserted per output in the render test.
- **NonClusterHost** drives non-cluster hosts, which read the resource directly and hot-reload the value. On a cluster whose CRD predates the field, hosts fall back to gzip, which every release's ingestion listener inflates.
- The field is named for the ingestion path and "in transit" deliberately: felix already zstd-compresses rotated log files on disk.

Both zstd and gzip are compiled into the fluent-bit image, so nothing about the build changes. zstd is the default because it beat gzip on both axes on real Calico log data (43x vs 28x compression, roughly 6x less CPU); gzip is kept for symmetry with Linseed's accepted encodings and for hosts attached to older clusters.

## Requires the Linseed-side change

Depends on tigera/calico-private#13367, which teaches Linseed to inflate zstd and gzip request bodies (and adds the host plugin and Go client sides). Ship together, or Linseed first: fluent-bit's `out_http` drops a chunk on a 4xx rather than retrying it.

## Upgrade order

Clusters are upgraded before non-cluster host packages, and a management plane before its managed clusters — the same rule for MCM and Calico Cloud. A host at version N only meets a cluster at N or newer. `ingestionCompression: None` is the escape hatch.

## CPU cost

Decompression costs Linseed roughly 5–20% more CPU per request (less for bigger batches), measured on a dev laptop and a cluster node (`e2-standard-2`). It runs at 488 MB/s of logs per node vCPU against about 30 MB/s for the NDJSON parsing that follows, so it is ~15x cheaper per byte than work Linseed already does. Sender-side cost is distributed across nodes: a 10KB batch encodes in 254us.

## Release Note

```release-note
Log ingestion to Linseed is compressed with zstd by default, reducing bandwidth by roughly 12x. The new `ingestionCompression` field on LogCollector and NonClusterHost (None, Gzip, Zstd) controls it; upgrade the cluster before non-cluster host packages.
```

## For PR author

- [x] Tests: render test asserts every Linseed output carries the configured value (Gzip / None / default) and no other output does; controller tests cover that an explicit value is left untouched; golden files regenerated with `UPDATE_RENDERED_CONFIGS=1`.
- [x] `make gen-files` run — CRDs and deepcopy regenerated for the new field.
- [x] `make gen-versions` — n/a, no version change.

Validated end to end on a real cluster: the new operator asserted the CRDs and both resources read `Zstd`; in-cluster outputs rendered `compress: zstd`; setting `None` re-rendered the ConfigMap without it and ingestion stayed clean (38 bulk responses, `"failed":0`); non-cluster hosts hot-reloaded on every field change.

**AI assistance:** Written with Claude Code (Claude Fable 5); design decisions, review and cluster validation by the author.


hjiawei:
Closing: this repo is no longer where operator changes land. tigera/operator has been imported into the Calico monorepos (tigera/calico-private#13606, projectcalico/calico#13674), both at `operator/`.

Both commits here have moved to **tigera/calico-private#13367**, which already carried the Linseed, host-plugin and Go-client sides of EV-6973 — so the change that had to ship together is now one PR in one repo, rather than two PRs with an ordering constraint between them.

The port is unchanged in substance; paths mapped as `api/...` → `operator/api/...`, `pkg/...` → `operator/pkg/...`, and `pkg/imports/crds/operator/*.yaml` → `operator/pkg/crds/operator/*.yaml` (the one directory the import relocated). Two things follow from the monorepo layout that did not apply here: the CRDs now reach `charts/` through symlinks into `operator/pkg/crds/operator`, so no chart copy is needed, and `manifests/{operator-crds,v1_crd_projectcalico_org,v3_projectcalico_org,v3_projectcalico_org-v1beta1}.yaml` are generated copies that the CRD change regenerates in the same PR — replacing the old post-merge `make get-operator-crds` flow.

Nothing goes to projectcalico/calico: LogCollector and NonClusterHost are Enterprise resources, and OSS ships neither CRD outside `operator/pkg/crds/operator`.

Review continues on tigera/calico-private#13367.

PR #5228, opened 2026-08-20
Move the whisker non-Calico cleanup behind an Enterprise modifier

caseydavenport:
## Description

The whisker render was deleting everything it produced when the variant wasn't Calico, which put variant knowledge in a base render package. The base render now produces the same objects either way, and the Enterprise extension moves them to the delete list, so an install upgraded from Calico still cleans up after itself.

## Release Note

```release-note
None
```

### DIFF
--- a/pkg/enterprise/whisker/extension.go
+++ b/pkg/enterprise/whisker/extension.go
@@ -1,0 +1,52 @@
+// Copyright (c) 2026 Tigera, Inc. All rights reserved.
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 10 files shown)


pkg/enterprise/whisker/extension.go
```diff
@@ -0,0 +1,52 @@
...
+	case *rwhisker.Component:
+		return extensions.Decorate(c, ri, e.variant, deleteWhisker)
+	default:
+		return c
+	}
+}
+
+// deleteWhisker moves everything whisker rendered to the delete list. Enterprise has
+// no whisker, so an install upgraded from Calico cleans up after itself.
+func deleteWhisker(create, del []client.Object) ([]client.Object, []client.Object) {
```
Brian-McM, line 50: Sure is easy to remove a calico only component lol.

Brian-McM (approved):

PR #5230, opened 2026-08-20
Move Enterprise gateway API rendering behind the extensions boundary

caseydavenport:
## Description

Another variant-gating split. The gateway API render no longer branches on variant, and the Enterprise extension layers on what only Enterprise installs get:

- the tigera envoy gateway, proxy and ratelimit images
- WAF filtering on the data-plane proxy, plus the l7-log-collector that ships the access and audit logs
- the WAF filter's cluster-scoped roles, and its service account and role binding in each Gateway namespace

Cleanup of the pre-namespaced install moves with it: the old service account in "tigera-gateway" and the two cluster role bindings that bound it are queued for deletion by the extension rather than the base render.

No behavior change for either variant. Test coverage for the Enterprise pieces moved into the new package alongside the code.

Related: [CORE-13396](https://tigera.atlassian.net/browse/CORE-13396)

## Release Note

```release-note
None
```

[CORE-13396]: https://tigera.atlassian.net/browse/CORE-13396?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pkg/enterprise/gatewayapi/extension.go
```diff
@@ -0,0 +1,162 @@
...
+	// once no Gateway namespaces remain.
+	if len(cfg.GatewayNamespaces) > 0 {
+		create = append(create, gatewayapi.GatewayNamespacesCRB(cfg.GatewayNamespaces))
+	} else {
+		del = append(del, gatewayapi.GatewayNamespacesCRB(nil))
+	}
+
+	image := gatewayAPIData(ri).l7LogCollectorImage
+	for _, o := range create {
+		if proxy, ok := o.(*envoyapi.EnvoyProxy); ok {
```
Brian-McM, line 125: If we don't find this component in the create list, that's a problem and we should probably panic or log an error.

Brian-McM (approved):

PR #5232, opened 2026-08-20
Move Enterprise istio rendering behind the extensions boundary

caseydavenport:
## Description

Another variant-gating split. The istio render no longer branches on variant, and the Enterprise extension layers on what only Enterprise installs get:

- the tigera istio pilot, install-cni, ztunnel and proxyv2 images
- the waypoint L7 logging resources: the defaults ConfigMap, the two EnvoyFilters, and the EnvoyFilter-writer Role and binding

When waypoint logging is turned off, those five objects are queued for deletion in the order that keeps the RBAC grant alive until the EnvoyFilters are gone. That was already the behavior; it moves into the extension with the rest.

The EnvoyFilter shim types stay in the render package, since the operator registers them in the scheme for the controller-runtime cache.

No behavior change for either variant. Test coverage for the Enterprise pieces moved into the new package alongside the code.

Related: [CORE-13396](https://tigera.atlassian.net/browse/CORE-13396)

## Release Note

```release-note
None
```

[CORE-13396]: https://tigera.atlassian.net/browse/CORE-13396?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pkg/render/istio/istio.go
```diff
@@ -80,6 +82,21 @@ const (
 	istioFakeImageProxyv2 = "fake.io/fakeimg/proxyv2:faketag"
 )
 
+// Component names, which key the image overrides a variant resolves through.
```
Brian-McM, line 85: In the future (certainly not now) we might want to see if we can actually separate out these "components" into proper components, instead of a component "facade" used just for image name mapping.

Not sure how feasible that is since the "components" may be very intertwined.

Brian-McM (approved):

PR #5233, opened 2026-08-20
Pass FelixConfiguration into the node and typha render

caseydavenport:
## Description

The installation controller was reading Felix's health port and the cgroup v2 path override off FelixConfiguration and copying them into the node and typha render config as scalars. The render takes the FelixConfiguration itself instead, so a render modifier can read anything else it needs from the same object without the controller having to copy another field in.

The health port keeps its 9099 default, which the render falls back to when FelixConfiguration does not set one.

No behavior change. Adds test coverage for the cgroup path, which had none.

Related: [CORE-13402](https://tigera.atlassian.net/browse/CORE-13402)

## Release Note

```release-note
None
```

[CORE-13402]: https://tigera.atlassian.net/browse/CORE-13402?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/controller/installation/core_controller.go
+++ b/pkg/controller/installation/core_controller.go
@@ -1225,12 +1225,12 @@ func (r *ReconcileInstallation) Reconcile(ctx context.Context, request reconcile
 
 	// Build a configuration for rendering calico/typha.
 	typhaCfg := render.TyphaConfiguration{
-		K8sServiceEp:      k8sapi.Endpoint,
-		Installation:      &defaulted.Spec,
-		TLS:               typhaNodeTLS,
-		MigrateNamespaces: needsNamespaceMigration,
-		ClusterDomain:     r.opts.ClusterDomain,
-		FelixHealthPort:   *felixConfiguration.Spec.HealthPort,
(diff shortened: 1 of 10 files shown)


pkg/controller/installation/core_controller.go
```diff
@@ -1720,7 +1719,7 @@ func (r *ReconcileInstallation) setDefaultsOnFelixConfiguration(ctx context.Cont
 
 	// Determine the felix health port to use. Prefer the configuration from FelixConfiguration,
 	// but default to 9099 (or 9199 on OpenShift). We will also write back whatever we select to FelixConfiguration.
-	felixHealthPort := 9099
+	felixHealthPort := render.DefaultFelixHealthPort
 	if install.Spec.KubernetesProvider.IsOpenShift() {
 		felixHealthPort = 9199
```
caseydavenport, line 1749: Constant for this too?

Brian-McM (approved):

PR #5234, opened 2026-08-20
Mirror the new RBAC UI role grants onto rbacsync's ClusterRole (EV-6977)

dimitri-nicolo:
## Description

Companion to **[calico-private#13388](https://github.com/tigera/calico-private/pull/13388)** (EV-6977). **That PR cannot merge until this one does.**

calico-private adds resources to the managed `calico-ui-*` role catalogue to close gaps against what `tigera-ui-user` and `tigera-network-admin` already ship. Kubernetes' privilege-escalation guard only lets `calico-kube-controllers` create a ClusterRole granting permissions it already holds, so each of those grants has to be mirrored in `rbacSyncControllerRules()` — which is exactly what that function exists for, per its own doc comment.

The failure without this change is not partial. The rbacsync controller ensures `cluster-context` first and returns on error, so the `gatewayapis` grant alone would abort the whole sync cycle — tier roles and LMA roles included — with a forbidden error on every tick.

### What's added

| Grant | Verbs | Managed role that needs it |
| --- | --- | --- |
| `operator.tigera.io/gatewayapis` | `get` | `calico-ui-cluster-context` |
| `applicationlayer.projectcalico.org` (6 WAF resources) | full CRUD | `calico-ui-waf-{view,mod}` |
| `gateway.networking.k8s.io/gateways,httproutes` | `get,list,watch` | `calico-ui-waf-{view,mod}` |
| `apps/deployments` | `get,list,watch,patch` | `calico-ui-waf-{view,mod}` |
| `policy.networking.k8s.io` (3 resources) | full CRUD | `calico-ui-admin-network-policies-{view,mod}` |
| `projectcalico.org/networks` | full CRUD | `calico-ui-egress-gateways-mod` |
| `projectcalico.org/securityeventwebhooks` | full CRUD | `calico-ui-webhooks-{view,mod}` |
| `crd.projectcalico.org/securityeventwebhooks` | `get,list,watch` | `calico-ui-webhooks-{view,mod}` |

Three of these deserve a second look:

- **`networks`** is already granted `watch,list,get` by the common rules, for the node controller's IPAM syncer. The egress-gateways *modify* role writes it, so the read-only grant does not clear the guard. It joins the existing `projectcalico.org` CRUD rule next to `externalnetworks`.
- **`apps/deployments` carries `patch`.** `calico-ui-waf-mod` toggles sidecar WAF by patching a deployment's pod template, which the console does against the apiserver directly rather than through ui-apis, so a read-only grant would not clear the guard for that role. Nothing else grants it: the base rules cover `apps/daemonsets` scoped to `calico-node`, and the migration role is read-only. This mirrors what `tigera-network-admin` already ships — see the [thread on calico-private#13388](https://github.com/tigera/calico-private/pull/13388#discussion_r3865685001) for the discussion of the cluster-wide scope and the follow-up it needs.
- **The WAF and Gateway API rules already exist** in `KubeControllersEnterpriseCommonRules` via `wafRules()` — but gated on `gatewayAPIPresent`. rbacsync needs them whenever RBAC management is enabled, independent of whether the GatewayAPI CR exists, so they are repeated in `rbacSyncControllerRules()` rather than leaning on that gate. Without that, the managed WAF roles would be creatable on Gateway API clusters and rejected by the escalation guard everywhere else. Open to instead widening the `wafRules()` gate to `gatewayAPIPresent || rbacManagementEnabled` if that reads better — it trades the duplication for a wider blast radius.

`securityeventwebhooks` moves out of the `projectcalico.org`-only list into its own rules now that it spans two API groups. The two groups are granted asymmetrically, mirroring the catalogue: the managed role writes through the aggregated `projectcalico.org` API, which validates the object, and only reads the CRD group behind it, so the escalation guard needs write on the former and reads on the latter. A test fails if the CRD group regains the write verbs.

Not needed, already held: `configmaps` (common rules — this is what covers the `coreruleset-default` read on the WAF roles), `services` including `watch` (base rules grant `get,list,update,watch`), and the two new `lma.tigera.io` log types (the existing `lma.tigera.io/cluster get` rule is unscoped by resource name).

I checked coverage in the other direction too: every API group and resource the catalogue grants — across `resourceroles.go`, `tierroles.go` and the LMA log roles — is held by the rendered role once this lands.

### Testing

- A `DescribeTable` asserting each new grant is present on the rendered `calico-kube-controllers` ClusterRole when the gate is on, one entry per resource.
- A test that `networks` carries the modify verbs specifically, not just the IPAM syncer's reads.
- A test that `apps/deployments` carries `patch`.
- A test that the `crd.projectcalico.org` group stays read-only for `securityeventwebhooks` while the aggregated group carries the write verbs.
- A negative test that `policy.networking.k8s.io` — granted by nothing else in the role — is absent when the feature is off.
- I confirmed the tests are not vacuous by reverting the `gatewayapis` and `networks` additions; both fail as intended.
- `go test ./pkg/enterprise/installation/...` passes.

Affects Calico Enterprise only, and only when RBAC management is enabled.

## Release Note

```release-note
None
```

## For PR author

- [x] Tests for change.

### DIFF
--- a/pkg/enterprise/apiserver/extension.go
+++ b/pkg/enterprise/apiserver/extension.go
@@ -87,10 +87,6 @@ type apiServerRenderData struct {
 	// run. The base render resolves it too, but a modifier runs with no ImageSet.
 	calicoImage string
 
-	// rbacManagementEnabled mirrors the rbac-ui-config gate. ui-apis writes role bindings
-	// impersonating the caller, so tigera-network-admin needs the verbs itself.
-	rbacManagementEnabled bool
-
 	// bindingNamespaces is the set of tenant namespaces whose calico-apiserver ServiceAccount
 	// should be granted Linseed access. Empty for zero/single-tenant clusters; every tenant
 	// namespace for multi-tenant management clusters.
@@ -225,8 +221,10 @@ func (e *Extension) ExtendInputs(ctx context.Context, ci controller.Inputs) (con
 		return ci, nil, extensions.InvalidConfigf("having both a ManagementCluster and a ManagementClusterConnection is not supported")
 	}
 
-	rbacManagementEnabled, err := utils.RBACManagementEnabled(ctx, ci.Client, e.variant, e.opts.MultiTenant)
-	if err != nil {
+	// Nothing the apiserver renders branches on the switch any more, but an unreadable
+	// switch still means unknown cluster state, so degrade rather than reconcile on a
+	// guess.
+	if _, err := utils.RBACManagementEnabled(ctx, ci.Client, e.variant, e.opts.MultiTenant); err != nil {
 		return ci, nil, extensions.Degradedf(operatorv1.ResourceReadError, "error reading the RBAC management UI ConfigMap: %w", err)
 	}
 
@@ -345,7 +343,6 @@ func (e *Extension) ExtendInputs(ctx context.Context, ci controller.Inputs) (con
 		calicoImage:                 calicoImage,
 		bindingNamespaces:           bindingNamespaces,
 		cloud:                       e.opts.Cloud,
-		rbacManagementEnabled:       rbacManagementEnabled,
 	}
 	return ci, nil, nil
 }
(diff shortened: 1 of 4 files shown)


dimitri-nicolo:
@electricjesus there's one decision here I'd like you to sign off on. The WAF grants from wafRules() are repeated in rbacSyncControllerRules() without the GatewayAPI gate, because rbacsync needs them whenever RBAC management is enabled. As a result, kube-controllers can patch any Deployment in the cluster while the feature is on. That is what lets it create the calico-ui-waf-mod role, which grants the same verb for the sidecar WAF toggle (see [tigera/calico-private#13388](https://github.com/tigera/calico-private/pull/13388)). Let me know if that's too broad for this service account.

PR #5237, opened 2026-08-21
Add a startup extension so main stops calling into pkg/enterprise

caseydavenport:
## Description

The operator's startup path calls into the Enterprise package directly for several things. Those go through the extensions boundary here instead, via a new startup slot that no-ops when nothing registers.

- checking that the variant's CRDs are served before any controller starts
- verifying the cluster's Elasticsearch certificates match the configured internal or external mode
- listing the namespaces the operator manages and so must not run in itself
- discovering the tenancy mode, which only ever comes back true for Enterprise or Cloud
- interpreting the cloud build flag, which the Makefile bakes in through linker flags

`cmd/main.go` builds the extensions once it has a client and has read the variant off the Installation, which leaves it naming a variant in exactly one place.

Behavior is unchanged with two exceptions worth calling out. The protected namespaces were already applied whatever the variant, so the startup extension registers for every variant and skips the checks that only apply to Enterprise. The version banner prints the build variant it was given rather than a hardcoded product name, so a cloud build reports "cloud"; the line the release tooling parses for the operator version is untouched.

Related: [CORE-13420](https://tigera.atlassian.net/browse/CORE-13420)

```release-note
None
```

[CORE-13420]: https://tigera.atlassian.net/browse/CORE-13420?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


cmd/main.go
```diff
@@ -87,13 +87,9 @@ var (
...
-// buildVariant is set to "cloud" via -ldflags "-X main.buildVariant=cloud" when building the Calico
-// Cloud operator image (see CLOUD_LDFLAGS in the Makefile), and is empty otherwise.
-var buildVariant string
-
-func isCloudBuild() bool {
-	return buildVariant == "cloud"
-}
+// buildExtensions is the variant's extensions factory, and the only place this binary
+// names a variant. A variant's own main supplies its own once main becomes a shim.
+var buildExtensions extensions.Factory = enterprise.Build
```
Brian-McM: As this PR stands the factory methodology here seems uneccessary (might be missing something though).

I could understand this though if the follow up work / goal is that this factory is defaulted in OS and then overwritten in enterprise, although I might still have some questions if that's the case.

caseydavenport: Yeah, it might be overkill - the idea is that we're preparing for a future where we have an "operator as a library" structure, and enterprise imports it.

Something like this:

- calico:
  - `lib/operator/`
  - `cmd/main.go` - imports `lib/operator/` and calls `New()` without an extension.
- calico-private
  - `cmd/main.go` - imports `calico/lib/operator/` and calls `New()` with this extension factory.

The reason for a factory is that the context needed to build the extension is queried from the cluster in `main`

It mostly just removes the need to duplicate that context gathering to build the extension - perhaps overkill, perhaps not. But happy to hear either way.

Otherwise, we'll just need to copy lines 424-449 below in both places.

Brian-McM: I think in this mono repo merged world we'd might still have separate main files for calico versus calico-private, so the buildExtensions variable that's overwritten in main.go doesn't seem like it would even be something that's viable to use in that future.

The common lines you referenced will have to go into some common invocable library if you want to share it between the main OS and EE main files.

The main functions just call whatever they need to get the extension registry and pass it to the common tooling so the "factories" don't even have to have the same function definition between the two variants.

I might just advocate for taking out that factory variable / type because whether or not that's needed is very much dependent on the final step we take to merge this into the mono repo.

caseydavenport: Yeah, I took out the factory.

Brian-McM (approved):

PR #5238, opened 2026-08-21
Move the goldmane non-Calico cleanup behind an Enterprise modifier

caseydavenport:
Goldmane only ships on Calico, and the render enforced that itself: for any other variant it moved everything it had just built onto the delete list. That branch goes into an Enterprise modifier, matching the whisker split in https://github.com/tigera/operator/pull/5228.

Core renders goldmane unconditionally and keeps the one deletion that is not about the variant, the metrics service when no metrics port is configured. The Enterprise-only expectation moved out of the render test and into the extension's own test.

Related: [CORE-13421](https://tigera.atlassian.net/browse/CORE-13421)

```release-note
None
```

[CORE-13421]: https://tigera.atlassian.net/browse/CORE-13421?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/controller/goldmane/controller.go
+++ b/pkg/controller/goldmane/controller.go
@@ -37,6 +37,7 @@ import (
 	"github.com/tigera/operator/pkg/controller/utils/imageset"
 	"github.com/tigera/operator/pkg/ctrlruntime"
 	"github.com/tigera/operator/pkg/dns"
+	"github.com/tigera/operator/pkg/extensions"
 	"github.com/tigera/operator/pkg/render"
 	rcertificatemanagement "github.com/tigera/operator/pkg/render/certificatemanagement"
 	"github.com/tigera/operator/pkg/render/goldmane"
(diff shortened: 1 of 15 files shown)


pkg/controller/goldmane/controller.go
```diff
@@ -256,7 +259,16 @@ func (r *Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
...
 	}
 
-	ch := utils.NewComponentHandler(log, r.cli, r.scheme, goldmaneCR)
+	ri := render.Inputs{
+		Installation:  installationSpec,
+		ClusterDomain: r.clusterDomain,
+		TrustedBundle: trustedBundle,
+	}
+	ch := utils.NewComponentHandler(log, r.cli, r.scheme, goldmaneCR,
+		utils.WithModifier(func(c render.Component) render.Component {
```
caseydavenport: Let's not do this inline...

```go
mod := utils.WithModifier(...)
```

Brian-McM (approved):

PR #5255, opened 2026-08-26
Write the chosen IP pools to the Installation spec

caseydavenport:
## Description

A fresh install that leaves the IP pool list unset could end up with no IP pool at all, leaving Calico degraded on "waiting for enabled IP pools to be created" and every pod unable to get a sandbox. The IP pool controller creates its default pool, reconciles again on that create before the effective config has been republished, and reads the pool it just made as a sign that pools are managed out-of-band, so it records an empty list over its own default and deletes the pool.

- Pool membership moves to the Installation spec. When nobody declared pools, the operator writes the ones it chose there and reads them back from there, so nothing re-derives the list from cluster state, and an edit that appends a pool can see the pools already in use.
- A declared list is left alone, and per-pool field defaults still go to the status and layer by CIDR.
- Second commit: that write is a server-side apply under its own field manager, and the pool list is a map keyed on CIDR, so each manager owns the pools it declared rather than fighting over the whole list.

Diagnosis by Jiawei Huang, who hit it on gcp-kubeadm clusters running the master hashrelease.

```release-note
Fixes a fresh install creating no IP pool when the Installation leaves the pool list unset; the operator now fills the IP pool list in on the Installation when it is not specified.
The Installation IP pool list is a map keyed on CIDR, so server-side apply clients keep ownership of the pools they declare.
```

### DIFF
--- a/api/v1/installation_types.go
+++ b/api/v1/installation_types.go
@@ -695,10 +695,9 @@ type CalicoNetworkSpec struct {
 	// +optional
 	ClusterRoutingMode *ClusterRoutingMode `json:"clusterRoutingMode,omitempty"`
 
-	// IPPools contains a list of IP pools to manage. If nil, the operator chooses: on a cluster with no
-	// operator-managed pools it creates a single IPv4 pool, and on a cluster that already has them it keeps
-	// the pools in use. Provide an empty list to manage IP pools out-of-band; the operator then deletes the
-	// pools it owns and creates none.
+	// IPPools contains a list of IP pools to manage. If nil, the operator fills this field in: a single IPv4
+	// pool on a cluster with no IP pools, or an empty list on a cluster that already has them, which leaves
+	// those pools to whoever created them.
 	// IP pools in this list will be reconciled by the operator and should not be modified out-of-band.
 	// +optional
 	// +kubebuilder:validation:MaxItems=25
--- a/api/v1/installation_types.go
+++ b/api/v1/installation_types.go
@@ -695,10 +695,9 @@ type CalicoNetworkSpec struct {
 	// +optional
 	ClusterRoutingMode *ClusterRoutingMode `json:"clusterRoutingMode,omitempty"`
 
-	// IPPools contains a list of IP pools to manage. If left unset, the operator fills it in with the pools
-	// it already owns, or with the platform's pod CIDRs or a built-in default on a cluster that has none. A
-	// cluster with pools the operator does not own gets an empty list. Provide an empty list to manage IP
-	// pools out-of-band; the operator then deletes the pools it owns and creates none.
+	// IPPools contains a list of IP pools to manage. If unset, the operator fills it in with defaults based
+	// on cluster state. Provide an empty list to manage IP pools out-of-band; the operator then deletes the
+	// pools it owns and creates none.
 	// IP pools in this list will be reconciled by the operator and should not be modified out-of-band.
 	// +optional
 	// +kubebuilder:validation:MaxItems=25
@@ -888,7 +887,6 @@ type IPPool struct {
 	Name string `json:"name,omitempty"`
 
 	// CIDR contains the address range for the IP Pool in classless inter-domain routing format.
-	// MaxLength keeps the rule below inside the CEL cost budget.
 	// +kubebuilder:validation:MaxLength=43
 	// +kubebuilder:validation:XValidation:rule="isCIDR(self) && string(cidr(self).masked()) == self",message="cidr must be in canonical form, for example 192.168.0.0/16 or fd00:10:244::/64"
 	CIDR string `json:"cidr"`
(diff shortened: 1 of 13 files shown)


Brian-McM (approved):

pkg/controller/ippool/pool_controller.go
```diff
@@ -216,6 +269,15 @@ func (r *Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
 		declared = utils.OverrideInstallationSpec(declared, *overlay)
 	}
 
+	poolsDeclared := declared.CalicoNetwork != nil && declared.CalicoNetwork.IPPools != nil
+	if poolsDeclared {
+		// Pool membership comes from the spec, since the computed config lags a write-back by a reconcile.
```
caseydavenport:
```suggestion
		// The pools section of the computed configuration may lag, since it is written by this controller.
		// Set it explicitly from the spec.
```

Also, I assume this means pools can't be set in the `overlay` resource effectively with this logic?

pkg/controller/ippool/pool_controller.go
```diff
@@ -242,6 +304,19 @@ func (r *Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
 		r.status.SetDegraded(operatorv1.InvalidConfigurationError, "error validating IP pool configuration", err, reqLogger)
 		return reconcile.Result{}, err
 	}
+
+	if !poolsDeclared && computed.CalicoNetwork != nil && computed.CalicoNetwork.IPPools != nil {
+		// Nobody declared pools, so the chosen list goes on the spec, where whoever edits it next can see it.
```
caseydavenport:
```suggestion
		// The spec didn't declare any pools, so the chosen list goes back on the spec, where whoever edits it next can see it.
```

pkg/controller/ippool/pool_controller_test.go
```diff
@@ -157,6 +158,104 @@ var _ = Describe("IP Pool controller tests", func() {
...
+				applied = &installation.ManagedFields[i]
+			}
+		}
+		Expect(applied).NotTo(BeNil())
+		Expect(applied.Operation).To(Equal(metav1.ManagedFieldsOperationApply))
+		Expect(applied.FieldsV1).NotTo(BeNil())
+		Expect(applied.FieldsV1.GetRawString()).To(ContainSubstring("ipPools"))
+	})
+
+	It("should keep its own default pool when it reconciles again before the computed config is republished", func() {
```
caseydavenport, line 205: In addition to these, I think we should have an FV test that explicitly exercises the field management logic with multiple writers against a real cluster.

api/v1/installation_types.go
```diff
@@ -886,6 +888,9 @@ type IPPool struct {
 	Name string `json:"name,omitempty"`
 
 	// CIDR contains the address range for the IP Pool in classless inter-domain routing format.
+	// MaxLength keeps the rule below inside the CEL cost budget.
```
caseydavenport:
```suggestion
```

api/v1/installation_types.go
```diff
@@ -695,13 +695,15 @@ type CalicoNetworkSpec struct {
 	// +optional
 	ClusterRoutingMode *ClusterRoutingMode `json:"clusterRoutingMode,omitempty"`
 
-	// IPPools contains a list of IP pools to manage. If nil, the operator chooses: on a cluster with no
-	// operator-managed pools it creates a single IPv4 pool, and on a cluster that already has them it keeps
-	// the pools in use. Provide an empty list to manage IP pools out-of-band; the operator then deletes the
-	// pools it owns and creates none.
+	// IPPools contains a list of IP pools to manage. If left unset, the operator fills it in with the pools
+	// it already owns, or with the platform's pod CIDRs or a built-in default on a cluster that has none. A
```
caseydavenport: This is confusing wording - just "if unset, the operator fills it in with defaults based on cluster state"

pkg/controller/ippool/defaults.go
```diff
@@ -75,8 +75,8 @@ func fillDefaults(ctx context.Context, client client.Client, spec *operator.Inst
 		return fmt.Errorf("cannot perform IP pool defaulting until CNI configuration is available")
 	}
 
-	// Only add default CIDRs if there are no existing pools in the cluster. If there are existing pools in the cluster,
-	// then we assume that the user has configured them correctly out-of-band and we should not install any others.
+	// Platform CIDRs and the built-in default only apply to a cluster with no IP pools at all. On any other
```
caseydavenport: this comment change feels like a regression in readability.

pkg/controller/ippool/defaults.go
```diff
@@ -136,14 +136,6 @@ func fillDefaults(ctx context.Context, client client.Client, spec *operator.Inst
 				}
 			}
 		}
-	} else if spec.CalicoNetwork == nil || spec.CalicoNetwork.IPPools == nil {
```
caseydavenport, line 139: Why do we need to remove this? How does the default in the out-of-band case get set to `[]` now?

PR #5270, opened 2026-08-27
Resolve variant images from the component lists

caseydavenport:
## Description

Calico and Enterprise builds of the same image differed by exactly one thing: the field that already picks the default registry and image path, and that also keys an ImageSet lookup.

- so the key constants, the override table, its registrations and its twenty call sites all expressed "use the Enterprise registry"
- the components package answers that directly now, out of the same generated lists that ImageSet validation reads
- core resolves the images this build ships; a variant registers its own as it builds its extensions, so no core package branches on the variant
- a variant with no image under a key is an error, not a silent fall back to the Calico one
- a unit test holds the keys and the lists in sync
- net 214 lines deleted

Calico Cloud is the one image outside the lists, so it stays explicit: supplied by the installation extension for kube-controllers, and by the enterprise assembler for es-kube-controllers.

Two gateway API specs asserted Calico envoy images under an Enterprise variant, which only passed because the fixture left the override table nil. A real Enterprise install renders the Enterprise images, so those expectations moved.

Related: [CORE-13465](https://tigera.atlassian.net/browse/CORE-13465)

```release-note
None
```

[CORE-13465]: https://tigera.atlassian.net/browse/CORE-13465?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pkg/enterprise/register.go
```diff
@@ -40,6 +38,10 @@ func New(variant operatorv1.ProductVariant, o eoptions.Options) extensions.Exten
 	// an Installation asking for the deprecated TigeraSecureEnterprise still gets the
 	// Enterprise extensions.
 	if variant.IsEnterprise() {
+		// Registered here rather than in an init, so the images arrive with the
+		// extensions and cannot be missed by a build that has one without the other.
+		components.RegisterVariantImages(variant, components.EnterpriseImages)
```
caseydavenport: Hm, shouldn't `EnterpriseImages` live in `pkg/enterprise/images` ?

pkg/components/imagefor.go
```diff
@@ -0,0 +1,85 @@
...
+var ImageKeys = []string{
+	ImageKeyCalico, ImageKeyNode, ImageKeyCNIPlugins,
+	ImageKeyNodeWindows, ImageKeyCNIWindows,
+	ImageKeyEnvoyGateway, ImageKeyEnvoyProxy, ImageKeyEnvoyRatelimit,
+	ImageKeyIstioPilot, ImageKeyIstioInstallCNI, ImageKeyIstioZTunnel, ImageKeyIstioProxyv2,
+}
+
+// variantImages is what a variant registered over the images this build ships. It is
+// written once while the operator resolves its variant, before any render runs.
+var variantImages = map[operator.ProductVariant][]Component{}
```
Brian-McM: Do we need to have a map here, or should it just be a flat list that can be overwritten?

My "naive" thinking to seeing the map is that we restart if we switch variants (like on downgrade), but I'm also not sure if there's something else I'm missing (maybe some bootstrapping things?).

pkg/components/imagefor.go
```diff
@@ -0,0 +1,85 @@
...
+
+// ImageFor returns the image the given variant runs for key. A miss is an error rather
+// than a fallback to another image, which would ship the wrong one silently.
+func ImageFor(v operator.ProductVariant, key string) (Component, error) {
+	imgs := CalicoImages
+	if registered, ok := variantImages[v]; ok {
+		imgs = registered
+	}
+	for _, c := range imgs {
+		if c.Image == key {
```
Brian-McM: If the lookup is always done with this key then maybe a map is a slightly better solution here?

Brian-McM (approved):

pkg/components/imagefor.go
caseydavenport: Hm, good point. Let me think about that a bit.

PR #5272, opened 2026-08-28
feat(manager): show the WAF management UI when the WAF extension is on

electricjesus:
Jira: https://tigera.atlassian.net/browse/EV-6793

Turning on `GatewayAPI.spec.extensions.waf` now gives you the WAF management UI. There is no second switch to find.

The manager controller reads the WAF extension state and sets `WAF_UI_ENABLED` on the ui-apis container. That is the variable ui-apis already reads at startup, so this is the whole change. No new API field and no ConfigMap.

Nothing new gets watched. The manager controller has watched `GatewayAPI` since it started resolving the ingress gateway class (`manager_controller.go:172`), so flipping the extension already re-reconciles the manager.

The read mirrors `isGatewayWAFEnabled` in the applicationlayer controller. A missing `GatewayAPI` CR means no gateways, so no WAF and no UI, and that is not an error. Any other read error degrades so the controller requeues instead of reporting the feature off. Enterprise only, and off on multi-tenant, matching `utils.RBACManagementEnabled`.

Flipping `waf.state` rolls the manager Deployment. That happens on enable or disable, not during normal use.

Affected component: operator (manager controller, manager render).

## Testing

`go test ./pkg/controller/manager/ ./pkg/render/` pass. New controller specs cover no CR, extensions unset, `Disabled`, `Enabled`, the legacy `tigera-secure` CR, the Calico variant, and multi-tenant. New render spec asserts `WAF_UI_ENABLED` tracks the flag both ways.

Also run on a cluster, swapping only the operator image. One master hashrelease cluster, one `GatewayAPI` CR, nothing else changed:

| operator | `waf.state` | `WAF_UI_ENABLED` |
|---|---|---|
| stock | no CR | absent |
| this PR | no CR | `false` |
| this PR | `Enabled` | `true` |
| stock | `Enabled` | **absent** |
| this PR | `Enabled` | `true` |

Row 4 is the bug: stock reconciles, bumps the Deployment generation, and strips the variable back out while WAF is on.

## Release Note

```release-note
The WAF management UI now appears when the GatewayAPI WAF extension is enabled.
```

## For PR author

- [x] Tests for change.

## For PR reviewers

- [ ] Appropriate labels:

### DIFF
--- a/pkg/controller/manager/manager_controller.go
+++ b/pkg/controller/manager/manager_controller.go
@@ -737,6 +737,12 @@ func (r *ReconcileManager) Reconcile(ctx context.Context, request reconcile.Requ
 		return reconcile.Result{}, err
 	}
 
+	wafUIEnabled, err := wafManagementEnabled(ctx, r.client, installationSpec.Variant, tenant.MultiTenant())
+	if err != nil {
+		r.status.SetDegraded(operatorv1.ResourceReadError, "Error reading the GatewayAPI WAF extension", err, logc)
+		return reconcile.Result{}, err
+	}
+
 	managerCfg := &render.ManagerConfiguration{
 		VoltronRouteConfig:         routeConfig,
 		KeyValidatorConfig:         keyValidatorConfig,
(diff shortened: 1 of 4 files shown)


electricjesus:
going back to #5001

PR #5290, opened 2026-09-01
Set require_mtu_file when CNI MTU is auto-detected

### LINKED ISSUE #5289: Operator should set require_mtu_file when CNI MTU is auto-detected
## Expected Behavior

With `spec.calicoNetwork.mtu` unset (auto-detect), a pod created before Felix has written `/var/lib/calico/mtu` shouldn't end up with the wrong MTU.

## Current Behavior

The pod gets a veth at the kernel default of 1500 and starts normally. The veth MTU is fixed at creation, so the pod keeps that value until it's recreated.

projectcalico/calico#10270, the fix for projectcalico/calico#9718, added a `require_mtu_file` key to the CNI config and shipped in Calico v3.31.0. The operator's rendered CNI config doesn't set that key — `pkg/render/node.go` on master renders `"mtu": mtu` with no `require_mtu_file` — so it defaults to `false` and operator-managed clusters keep the older behavior. I also don't see a corresponding field on the `Installation` API, so as far as I can tell operator users can't currently enable it.

The path through the code:

1. `createCalicoPluginConfig()` renders `"mtu": 0` when `spec.calicoNetwork.mtu` is unset, meaning auto-detect.
2. Felix writes the detected value to `/var/lib/calico/mtu` in `writeMTUFile()`, called from `NewIntDataplaneDriver()` at startup.
3. On CNI ADD the plugin reads that file:

```go
// cni-plugin/pkg/plugin/plugin.go
if mtu, err := utils.MTUFromFile(utils.MTUFilePath, conf); err != nil {
    return fmt.Errorf("failed to read MTU file: %s", err)
} else if conf.MTU == 0 && mtu != 0 {
    conf.MTU = mtu
}
```

4. If the file isn't there yet, the error is skipped because `RequireMTUFile` is false:

```go
// cni-plugin/internal/pkg/utils/utils.go
if os.IsNotExist(err) {
    if conf.RequireMTUFile {
        return 0, err
    }
    logrus.WithField("filename", filename).WithError(err).Errorf("File does not exist, skipping the error since RequireMTUFile is false")
    return 0, nil
}
```

5. `conf.MTU` stays `0`, and that's passed to netlink as-is (`la.MTU = d.mtu` in `cni-plugin/pkg/dataplane/linux/dataplane_linux.go`), so the kernel applies its default veth MTU of 1500.

There's a window on node join where kubelet is admitting pods but Felix hasn't written the file yet. `hostNetwork` pods have no veth, so in my capture the pods that hit this window were the first non-hostNetwork DaemonSets scheduled onto the node.

## Possible Solution

A conditional key in `createCalicoPluginConfig()`:

```go
if mtu == 0 {
    calicoPluginConfig["require_mtu_file"] = true
}
```

When the MTU is auto-detected, a missing MTU file becomes a failed CNI ADD that kubelet retries rather than a silently wrong veth. When the MTU is pinned the key is omitted, leaving the rendered config identical to today.

Gating on `mtu == 0` keeps the change narrow: when the MTU is pinned the plugin never reads the file, so requiring it would add a failure mode for no benefit. Omitting rather than rendering `false` also matters because the CNI config is hashed into the calico-node pod template annotation, so an unused key would roll calico-node for no behavior change. It does not change the unreadable-file case: `MTUFromFile` only gates `os.IsNotExist` behind this flag, and other read errors (e.g. the SELinux case in projectcalico/calico#7851) already return an error today.

I have this working locally with updated `pkg/render/node_test.go` fixtures and `go test ./pkg/render/` passing; PR to follow.

Two things I'd want a maintainer's view on:

- `pkg/render/windows.go` has the same `"mtu": mtu` line and I've left it alone — projectcalico/calico#10270 did touch `install_windows.go`, but the veth race is Linux-specific and I'd rather ask than guess.
- Calico's default is also off via `REQUIRE_MTU_FILE=false`; should the operator enable it for auto-MTU installs, or should Calico change the default more broadly?

1. Operator-managed Calico with `spec.calicoNetwork.mtu` unset and VXLAN encapsulation. Confirm `cm/cni-config` in `calico-system` contains `"mtu":0` and no `require_mtu_file`.
2. Join a new node, so kubelet starts admitting pods before Felix has written `/var/lib/calico/mtu`.
3. Let the first non-hostNetwork DaemonSet pods be scheduled onto it.
4. On the node, compare veth MTUs against `vxlan.calico`: the pods that landed in the window are at 1500, `vxlan.calico` and every later pod are at 1450.

This is timing-dependent and didn't reproduce on every node — two of the nodes in the same cluster came up entirely at 1450, matching the intermittent behavior described in projectcalico/calico#9718.

## Context

I found this while investigating pods with 1500-byte veths on a VXLAN cluster where everything else was correct: `vxlan.calico` was 1450, the routes to remote pod blocks carried an explicit `mtu 1450`, and `/var/lib/calico/mtu` contained `1450`. Sorting the node's interfaces by index and container start time showed a clean boundary — lowest indices and earliest start times all at 1500, everything later at 1450. The affected pods were promtail, loki-canary, velero node-agent and virt-handler; everything starting earlier was `hostNetwork` and had no veth.

Affected pods advertise an MSS 50 bytes too large for the tunnel. Where PMTUD works this costs a drop and retransmit; where it does not, large payloads can be lost. Either way it's silent — the pods run normally and nothing reports the mismatch, so the only way I found it was by checking interface MTUs directly.

Pinning `spec.calicoNetwork.mtu` avoids the whole thing, since a non-zero MTU means the plugin never consults the file. But that's a workaround for a default that silently misconfigures pods, and it needs the affected pods recreated afterwards either way.

## Your Environment

* Operating System and version: Ubuntu 24.04.4 LTS, kernel 6.8.0
* Kubernetes: v1.34.5, containerd 2.2.1
* tigera/operator: v1.40.7
* Calico: v3.31.4 (so the projectcalico/calico#10270 code is present)
* Encapsulation: VXLAN, 1500-byte underlay
* `spec.calicoNetwork.mtu`: unset

Vemula-Rohith:
## Description

Bug fix, affecting Linux CNI config rendering in `pkg/render/node.go`. Fixes #5289.

projectcalico/calico#10270 (the fix for projectcalico/calico#9718) added a `require_mtu_file` key to the CNI config and shipped in Calico v3.31.0. The operator doesn't set it, so it defaults to `false` and operator-managed clusters keep the older behavior even on Calico builds that contain the fix.

When `spec.calicoNetwork.mtu` is unset we render `"mtu": 0`, and the plugin resolves the real value by reading `/var/lib/calico/mtu`, which Felix writes at startup. If a pod is created before that file exists, `MTUFromFile` skips the `os.IsNotExist` error because `RequireMTUFile` is false, `conf.MTU` stays `0`, and that reaches netlink as-is — so the kernel applies its default veth MTU of 1500.

The veth MTU is fixed at creation, so the pod keeps that value for its lifetime. Since `hostNetwork` pods have no veth, the pods most likely to hit this are early non-hostNetwork DaemonSets scheduled onto a joining node.

This sets the key only when the MTU is auto-detected:

```go
if mtu == 0 {
    calicoPluginConfig["require_mtu_file"] = true
}
```

A missing MTU file then becomes a failed CNI ADD that kubelet retries, rather than a silently wrong veth.

Gating on `mtu == 0` keeps the change narrow. When the MTU is pinned the plugin never reads the file, so requiring it would only add a failure mode. The key is omitted rather than rendered as `false` so the emitted config stays identical to today: `nodeCNIConfigMap` data is hashed into the calico-node pod template annotation, so adding an unused key would roll calico-node on upgrade for no behavior change.

This also does not change behavior for clusters where the MTU file exists but is unreadable — `MTUFromFile` only gates `os.IsNotExist` behind the flag, so other read errors (e.g. projectcalico/calico#7851) already return an error today. The one new failure mode is a retried CNI ADD during the startup window, which is what projectcalico/calico#10270 intends.

`pkg/render/windows.go` has the same `"mtu": mtu` line, and projectcalico/calico#10270 did modify `install_windows.go`, but the veth race is Linux-specific so I'd rather have a maintainer's call. I left Windows unchanged and can extend the PR if maintainers want the same rendering there.

### Testing

- Updated the 15 CNI config fixtures in `pkg/render/node_test.go` — 14 auto-detect cases expecting `true`, one pinned-MTU case expecting `false`.
- `go build ./pkg/render/` and `go test ./pkg/render/` pass.
- I also verified the fixtures fail if only the expected output is reverted, confirming the key actually reaches the rendered ConfigMap.
- Checked against a live operator v1.40.7 / Calico v3.31.4 cluster: `cm/cni-config` contains `"mtu":0` and no `require_mtu_file`, while `/var/lib/calico/mtu` on the nodes later contained the expected value.
- No dataplane functional test: the underlying behavior is covered by projectcalico/calico#10270, and this change only controls whether that path is enabled.

## Release Note

```release-note
Set require_mtu_file in the rendered CNI configuration when the MTU is auto-detected, so that pods created before Felix writes /var/lib/calico/mtu no longer receive the kernel default MTU of 1500.
```

## For PR author

- [x] Tests for change.
- [ ] If changing pkg/apis/, run `make gen-files` — n/a, no API change.
- [ ] If changing versions, run `make gen-versions` — n/a.

## For PR reviewers

- [ ] Appropriate labels: `kind/bug`.


pkg/render/node.go
```diff
@@ -675,9 +675,14 @@ func (c *nodeComponent) createCalicoPluginConfig() map[string]any {
...
-		"mtu":                    mtu,
+		"type":           "calico",
+		"datastore_type": "kubernetes",
+		"mtu":            mtu,
+		// When MTU is auto-detected (0), the CNI plugin reads the value from
+		// /var/lib/calico/mtu, which Felix writes on startup. If that file is not
+		// yet present the plugin would silently fall back to the kernel default
+		// veth MTU (1500), so require it to be there and let kubelet retry instead.
+		"require_mtu_file":       mtu == 0,
 		"nodename_file_optional": nodenameFileOptional,
```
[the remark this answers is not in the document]
Vemula-Rohith: The key is now added only when the MTU is auto-detected, rather than always rendered:

if mtu == 0 {
calicoPluginConfig["require_mtu_file"] = true
}

There is a concrete reason to omit the key rather than render false: nodeCNIConfigMap data is hashed into the calico-node pod template via hash.operator.tigera.io/cni-config. Adding an unused key would change that hash and trigger a calico-node rollout for pinned-MTU clusters on upgrade, with no behavior change.

Vemula-Rohith:
Closing this for now. The operator is being imported into the monorepo (projectcalico/calico#13674), where this file moves to operator/pkg/render/node.go, and issue intake has already moved there (#5283). Rather than have this sit against a path that is about to change, I'll re-open it against projectcalico/calico once the import lands. The branch is kept and the change is unaffected by the move.

pkg/render/node.go
```diff
@@ -692,6 +692,16 @@ func (c *nodeComponent) createCalicoPluginConfig() map[string]any {
 		"calico_api_group":             apiGroup,
 	}
 
+	// When MTU is auto-detected (0), the CNI plugin reads the value from
+	// /var/lib/calico/mtu, which Felix writes on startup. If that file is not yet
+	// present the plugin falls back to the kernel default veth MTU (1500), so
+	// require it to be there and let kubelet retry instead. The key is omitted
+	// when the MTU is pinned, since the plugin never reads the file in that case
+	// and rendering it would change the CNI config hash for no reason.
```
caseydavenport:
```suggestion
	// When MTU is auto-detected, the CNI plugin reads the value from
	// /var/lib/calico/mtu, which Felix writes on startup. Wait for it to avoid 
	// setting the wrong MTU.
```

caseydavenport:
Haha, I was just about to try to rush this one in before the cutoff. but happy to pick it up on the other side of the move!

Vemula-Rohith:
Reopening. This repository still cuts release-v1.41 through release-v1.44 for the rest of their supported lives, so the fix is worth landing here rather than waiting on the monorepo import (projectcalico/calico#13674). If maintainers would prefer it go straight to the monorepo instead, I'm happy to close this and re-file there — just say which you'd rather have.

pkg/render/node.go
caseydavenport: Looks like AI over-commenting, we can do with a tidier one.

caseydavenport:
One nitty comment, but otherwise LGTM.

pkg/render/node.go
Vemula-Rohith: Fair. I was trying to make the commit easier to review and definitely overdid it. Trimmed the comment down

caseydavenport (approved):

PR #5292, opened 2026-09-01
Recover a GatewayClass wedged in Terminating

electricjesus:
## Description

Deleting the `GatewayAPI` CR marks the `GatewayClass` for deletion, since we set a controller ownerRef on it. Envoy Gateway's `gateway-exists-finalizer` blocks that delete while any Gateway still uses the class, so it never finishes and nothing repairs it.

Envoy Gateway keeps managing a terminating class but skips its `parametersRef`, because it only resolves the ref when `DeletionTimestamp` is nil (`internal/provider/kubernetes/controller.go:366`, EG v1.8.3). Gateway deployments fall back to the upstream `docker.io/envoyproxy/envoy` image with no `imagePullSecrets`, and the class still reports `Accepted=True`. Re-creating the CR restores the `EnvoyProxy` but can't clear a `DeletionTimestamp`, so the cluster stays wrong until someone clears the finalizer by hand.

The gateway pod carries on serving and tigerastatus stays green, so this shows up later as a pod that won't start after a restart or drain.

This watches `GatewayClass`, clears the finalizers on a terminating one, and reports the repair on tigerastatus. The next reconcile re-creates the class and Envoy Gateway resolves `parametersRef` normally.

Clearing another controller's finalizer is deliberate. The wedge only exists because our ownerRef marked a class that Envoy Gateway won't release, and there's no other way out of it. Gateways lose their managed class for a few seconds while the class is re-created, so the proxy deployment is torn down and re-provisioned.

### Test plan

`go build`, `go vet`, `gofmt` and `go test ./pkg/controller/gatewayapi/...` all pass.

New unit test covers a class with `DeletionTimestamp` set, and asserts both the release and the tigerastatus message. It fails when the repair is short-circuited.

Checked by hand on a GCP kubeadm cluster running Enterprise master. Waiting, restarting envoy-gateway, annotating the Gateway and deleting the deployment all left it wedged. Clearing the finalizer and reconciling fixed it in 6 seconds.

Not covered: no e2e, and the upstream skip in EG is unchanged since v0.5.0 and still on EG main, so this only handles our side of it.

## Release Note

```release-note
Fixed a bug where deleting and re-creating the GatewayAPI resource could leave gateway deployments running an incorrect image without image pull secrets.
```

## For PR author

- [x] Tests for change.

### DIFF
--- a/pkg/controller/gatewayapi/gatewayapi_controller.go
+++ b/pkg/controller/gatewayapi/gatewayapi_controller.go
@@ -169,6 +169,23 @@ func Add(mgr manager.Manager, opts options.ControllerOptions) error {
 		return nil
 	}
 
+	// Watch GatewayClass for the same reason, and lazily for the same reason. Without this
+	// we cannot see a class wedged in Terminating: we own it by ownerRef, so deleting the
+	// GatewayAPI CR marks it, and Envoy Gateway's gateway-exists finalizer then blocks the
+	// delete for as long as any Gateway uses the class.
+	gatewayClassesWatched := false
+	r.watchGatewayClasses = func() error {
+		if gatewayClassesWatched {
+			return nil
+		}
+		log.V(1).Info("Adding watch for GatewayClass resources")
+		if err = c.WatchObject(&gapi.GatewayClass{}, &handler.EnqueueRequestForObject{}); err != nil {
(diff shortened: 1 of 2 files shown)


caseydavenport:

```

```

PR #5293, opened 2026-09-01
PMREQ-821: Gateway access-grant finalizer and UIGateway extension

vara2504:
## Description

Follow-ups from the #5146 review (Whisker CIG), both requested by @caseydavenport:

**1. A finalizer replaces delete-ordering for the gateway access grants** ([r3816269564](https://github.com/tigera/operator/pull/5146#discussion_r3816269564)). The access Role/RoleBinding were protected only by their position in the deletion list — emitted last so the operator kept its own write grant while deleting the Gateway/HTTPRoute/Backend. A failed or reordered delete could strip the grant first and orphan those resources. Both grants now carry `operator.tigera.io/gateway-rbac-finalizer`; teardown deletes only mark them, and a sweep in the uigateway helper removes the finalizer once the resources they cover are gone. The sweep discovers grants by the gateway label, so a grant that outlives its Gateway is still found and can never wedge. Uses `Update` rather than `Patch`: the operator's ClusterRole grants `update` on roles/rolebindings but not `patch`. No RBAC changes needed.

**2. The Enterprise-only gateway objects move behind a UIGateway extension** ([r3883122289](https://github.com/tigera/operator/pull/5146#discussion_r3883122289)). The WAF filter ServiceAccount/RoleBinding reached the shared render as a raw `ExtraProxyObjects` pass-through the manager controller built by hand, mirrored across the create and delete paths. A `UIGatewayExtension` in the variant extension set replaces that; the Enterprise implementation returns the WAF objects for the Manager prefix and nothing for other components, so behavior is unchanged on every variant.

Also adds `BackendName`, `ReferenceGrantName` and `ListenerName` helpers so each generated name has one definition (follow-up to the earlier "brittle string" comment).

**Testing:** unit tests for the finalizer stamping, the sweep's gating (clears only marked grants whose resources are gone), and the extension flow. Verified end to end on a GKE Enterprise cluster: enable → Gateway Programmed, grants carry the finalizer, WAF objects present, UI serves HTTP 200 through the gateway; teardown → all gateway objects and grants removed in one reconcile cycle, nothing left Terminating; the sweep also recovered grants deliberately wedged by an earlier build.

This PR was written in part with the assistance of generative AI.

## Release Note

```release-note
Ingress gateway teardown protects its RBAC grants with a finalizer instead of delete ordering, so a partially failed teardown can no longer orphan gateway resources.
```

### DIFF
--- a/pkg/enterprise/controller/manager/manager_controller.go
+++ b/pkg/enterprise/controller/manager/manager_controller.go
@@ -753,14 +753,13 @@ func (r *ReconcileManager) Reconcile(ctx context.Context, request reconcile.Requ
 	// observed on the cluster.
 	var gatewayComponents []render.Component
 	var gatewayTLSKeyPair certificatemanagement.KeyPairInterface
-	gwHelper := uigateway.NewHelper(r.client, uigateway.Config{
+	gwHelper := uigateway.NewHelper(r.client, r.opts.Extensions.UIGateway(), uigateway.Config{
 		ResourcePrefix:               ManagerGatewayResourcePrefix,
 		TLSSecretName:                ManagerGatewayTLSSecretName,
 		BackendNamespace:             helper.InstallNamespace(),
 		BackendServiceName:           render.ManagerServiceName,
 		BackendPort:                  render.ManagerPort,
 		BackendCABundleConfigMapName: certificatemanagement.TrustedCertConfigMapName,
-		ExtraProxyObjects:            euigateway.ProxyObjects(helper.InstallNamespace()),
 		Provider:                     r.opts.DetectedProvider,
 		Azure:                        installationSpec.Azure,
 	})
@@ -941,7 +940,7 @@ func (r *ReconcileManager) resolveAdditionalTunnelCert(
 
 const (
 	ManagerGatewayTLSSecretName  = "calico-manager-gateway-tls"
-	ManagerGatewayResourcePrefix = "calico-manager"
+	ManagerGatewayResourcePrefix = euigateway.ManagerGatewayResourcePrefix
 )
 
 // resolveGateway validates the Manager spec.ingressGateway configuration, resolves the
--- a/pkg/enterprise/uigateway/uigateway.go
+++ b/pkg/enterprise/uigateway/uigateway.go
@@ -13,16 +13,37 @@
 // limitations under the License.
 
 // Package uigateway layers the Calico Enterprise pieces onto a UI component's
-// ingress gateway. The common gateway code carries no variant knowledge; an
-// Enterprise controller passes what this package builds, and an OSS controller
-// passes nothing.
+// ingress gateway. The common gateway code carries no variant knowledge; it
+// reaches this package only through the registered UIGateway extension.
 package uigateway
 
 import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
+	"github.com/tigera/operator/pkg/extensions"
 	rgatewayapi "github.com/tigera/operator/pkg/render/gatewayapi"
 )
+
+// ManagerGatewayResourcePrefix names the Manager's CIG resources; the OSS
+// counterpart is whisker.GatewayResourcePrefix.
+const ManagerGatewayResourcePrefix = "calico-manager"
+
+// Extension is the Calico Enterprise behavior for UI ingress gateways.
+type Extension struct{}
+
+var _ extensions.UIGatewayExtension = Extension{}
+
+// New returns the UI gateway extension for Calico Enterprise.
+func New() Extension { return Extension{} }
+
+// ProxyObjects returns the WAF filter objects for the Manager gateway; other
+// components get nothing.
+func (Extension) ProxyObjects(resourcePrefix, namespace string) []client.Object {
+	if resourcePrefix != ManagerGatewayResourcePrefix {
+		return nil
+	}
+	return ProxyObjects(namespace)
+}
 
 // ProxyObjects returns the Enterprise-only objects that run beside a UI
 // gateway's Envoy proxy in the backend namespace: the WAF HTTP filter's
--- a/pkg/render/gateway/component.go
+++ b/pkg/render/gateway/component.go
@@ -49,6 +49,12 @@ const (
 	// once no labeled Gateway from any component remains, so components that
 	// share a namespace never delete it out from under each other.
 	GatewayNamespaceLabel = "operator.tigera.io/gateway-namespace"
+
+	// RBACFinalizer holds a gateway access Role and RoleBinding until the
+	// gateway resources they authorize deleting are gone, so teardown cannot
+	// strip the operator's own write grant first and orphan them. The
+	// uigateway helper removes it once those resources no longer exist.
+	RBACFinalizer = "operator.tigera.io/gateway-rbac-finalizer"
 )
 
 // Configuration holds everything the shared gateway component needs to render
@@ -162,6 +168,16 @@ func GatewayName(prefix string) string { return prefix + "-gateway" }
 // RouteName is the HTTPRoute object name for a component's resource prefix.
 func RouteName(prefix string) string { return prefix + "-route" }
 
+// BackendName is the Envoy Gateway Backend object name for a component's resource prefix.
+func BackendName(prefix string) string { return prefix + "-backend" }
+
+// ReferenceGrantName is the ReferenceGrant object name for a component's resource prefix.
+func ReferenceGrantName(prefix string) string { return prefix + "-allow-gateway" }
+
+// ListenerName is the Gateway's HTTPS listener name for a component's resource
+// prefix. The HTTPRoute's parentRef sectionName must match it to attach.
+func ListenerName(prefix string) string { return prefix + "-https" }
+
 // gatewayAccess grants the operator the write permissions needed in the gateway namespace; the
 // cluster-wide ClusterRole keeps the reads.
 func (c *gatewayComponent) gatewayAccess() (*rbacv1.Role, *rbacv1.RoleBinding) {
@@ -191,22 +207,25 @@ func (c *gatewayComponent) backendAccess() (*rbacv1.Role, *rbacv1.RoleBinding) {
 }
 
 // access builds a Role with rules and a RoleBinding tying it to the operator's
-// own ServiceAccount, the identity that renders the gateway resources.
+// own ServiceAccount, the identity that renders the gateway resources. Both
+// carry RBACFinalizer so the grant outlives the resources it covers.
 func (c *gatewayComponent) access(name, namespace string, rules []rbacv1.PolicyRule) (*rbacv1.Role, *rbacv1.RoleBinding) {
 	return &rbacv1.Role{
 			TypeMeta: metav1.TypeMeta{Kind: "Role", APIVersion: "rbac.authorization.k8s.io/v1"},
 			ObjectMeta: metav1.ObjectMeta{
-				Name:      name,
-				Namespace: namespace,
-				Labels:    map[string]string{GatewayLabel: c.cfg.ResourcePrefix},
+				Name:       name,
+				Namespace:  namespace,
+				Labels:     map[string]string{GatewayLabel: c.cfg.ResourcePrefix},
+				Finalizers: []string{RBACFinalizer},
 			},
 			Rules: rules,
 		}, &rbacv1.RoleBinding{
 			TypeMeta: metav1.TypeMeta{Kind: "RoleBinding", APIVersion: "rbac.authorization.k8s.io/v1"},
 			ObjectMeta: metav1.ObjectMeta{
-				Name:      name,
-				Namespace: namespace,
-				Labels:    map[string]string{GatewayLabel: c.cfg.ResourcePrefix},
+				Name:       name,
+				Namespace:  namespace,
+				Labels:     map[string]string{GatewayLabel: c.cfg.ResourcePrefix},
+				Finalizers: []string{RBACFinalizer},
 			},
 			RoleRef: rbacv1.RoleRef{
 				APIGroup: "rbac.authorization.k8s.io",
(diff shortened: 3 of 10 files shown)


pkg/uigateway/uigateway.go
```diff
@@ -258,13 +279,87 @@ func (h *Helper) Teardown(ctx context.Context) ([]render.Component, error) {
...
 		}))
 	}
 	return components, nil
 }
 
+// clearRBACFinalizers removes RBACFinalizer from every labeled access
+// Role and RoleBinding that is marked for deletion, once the gateway
+// resources it covers are gone. The finalizer keeps the operator's write
+// grant in place until then, so teardown does not depend on delete order.
+func (h *Helper) clearRBACFinalizers(ctx context.Context) error {
```
caseydavenport: This seems like it goes against the grain in terms of how we handle finalizers elsewhere - i.e., it's normally just a toggle include/exclude in the renderer.

Is this extra logic needed? Can't we calculate "Needs Finalizer" in the controller and pass it to the renderer?

pkg/uigateway/uigateway.go
```diff
@@ -258,13 +279,87 @@ func (h *Helper) Teardown(ctx context.Context) ([]render.Component, error) {
...
+		}
+	}
+	return nil
+}
+
+// accessResourcesGone reports whether none of the component's gateway
+// resources remain in the namespace — the point at which an access grant
+// there has nothing left to cover. A kind the cluster does not serve counts
+// as gone.
+func (h *Helper) accessResourcesGone(ctx context.Context, namespace string) (bool, error) {
```
caseydavenport: This checks all four resources in one namespace, but the Backend and ReferenceGrant legitimately stay in the backend namespace after a `gatewayNamespace` change. So on a move out of `calico-system`, the stale gateway-access grant there never clears and sits Terminating forever, which also blocks deleting that namespace later.

We should scope the check to what each grant actually authorizes: Gateway and HTTPRoute for the gateway grant, Backend and ReferenceGrant for the backend grant.

pkg/enterprise/controller/manager/manager_controller.go
```diff
@@ -753,14 +753,13 @@ func (r *ReconcileManager) Reconcile(ctx context.Context, request reconcile.Requ
 	// observed on the cluster.
 	var gatewayComponents []render.Component
 	var gatewayTLSKeyPair certificatemanagement.KeyPairInterface
-	gwHelper := uigateway.NewHelper(r.client, uigateway.Config{
+	gwHelper := uigateway.NewHelper(r.client, r.opts.Extensions.UIGateway(), uigateway.Config{
```
caseydavenport: Deleting the Manager CR never reaches this code. Reconcile returns at the `instance == nil` check above, so nothing runs the sweep, and the grants have an owner ref so GC marks them right away. The finalizer holds them and deleting the gateway namespace hangs.

Whisker covers this with an explicit teardown branch in its CR-not-found path (`pkg/controller/whisker/controller.go`); we should do the same here.

pkg/render/gateway/component.go
```diff
@@ -49,6 +49,12 @@ const (
 	// once no labeled Gateway from any component remains, so components that
 	// share a namespace never delete it out from under each other.
 	GatewayNamespaceLabel = "operator.tigera.io/gateway-namespace"
+
+	// RBACFinalizer holds a gateway access Role and RoleBinding until the
+	// gateway resources they authorize deleting are gone, so teardown cannot
+	// strip the operator's own write grant first and orphan them. The
+	// uigateway helper removes it once those resources no longer exist.
+	RBACFinalizer = "operator.tigera.io/gateway-rbac-finalizer"
```
caseydavenport: There's no escape hatch when a grant gets marked for deletion in a state the sweep won't clear: delete one by hand while the gateway is live, or remove the operator mid-teardown, and it's stuck Terminating with nothing left to reap it.

`core_controller.go` handles this for `CNIFinalizer` by stripping the finalizer whenever an object is marked for deletion but the CR isn't terminating. We should do the same for this one.

pkg/uigateway/uigateway.go
```diff
@@ -258,13 +279,87 @@ func (h *Helper) Teardown(ctx context.Context) ([]render.Component, error) {
...
 }
 
+// clearRBACFinalizers removes RBACFinalizer from every labeled access
+// Role and RoleBinding that is marked for deletion, once the gateway
+// resources it covers are gone. The finalizer keeps the operator's write
+// grant in place until then, so teardown does not depend on delete order.
+func (h *Helper) clearRBACFinalizers(ctx context.Context) error {
+	byLabel := client.MatchingLabels{rgateway.GatewayLabel: h.cfg.ResourcePrefix}
+	roles := &rbacv1.RoleList{}
+	if err := h.cli.List(ctx, roles, byLabel); err != nil {
```
caseydavenport: These two Lists are cluster-wide with no namespace or field selector, and the manager cache isn't restricted, so this pins every Role and RoleBinding in the cluster in memory. Both controllers call `Teardown` on every reconcile when the gateway is off, which is always on Enterprise and by default on OSS.

`Namespaces()` gates on `gatewayKindServed()` for this reason. We should gate the sweep the same way and scope the Lists to the namespaces we care about.

pkg/uigateway/uigateway.go
```diff
@@ -258,13 +279,87 @@ func (h *Helper) Teardown(ctx context.Context) ([]render.Component, error) {
...
+		rgateway.GatewayName(prefix):        &gapi.Gateway{},
+		rgateway.RouteName(prefix):          &gapi.HTTPRoute{},
+		rgateway.BackendName(prefix):        &envoyapi.Backend{},
+		rgateway.ReferenceGrantName(prefix): &gapiv1b1.ReferenceGrant{},
+	} {
+		err := h.cli.Get(ctx, types.NamespacedName{Name: name, Namespace: namespace}, obj)
+		if err == nil {
+			return false, nil
+		}
+		if !errors.IsNotFound(err) && !apimeta.IsNoMatchError(err) {
```
caseydavenport: `apimeta.IsNoMatchError` type-asserts without unwrapping, so a wrapped no-match gets treated as a real error and takes down the whole sweep. We use `stderrors.As(err, &noMatch)` everywhere else in this file for exactly that reason.

pkg/uigateway/uigateway.go
```diff
@@ -258,13 +279,87 @@ func (h *Helper) Teardown(ctx context.Context) ([]render.Component, error) {
...
+		}
+		if !gone {
+			continue
+		}
+		grant.SetFinalizers(slices.DeleteFunc(grant.GetFinalizers(), func(f string) bool {
+			return f == rgateway.RBACFinalizer
+		}))
+		// Update, not Patch: the operator's ClusterRole grants update on
+		// roles and rolebindings but not patch.
+		if err := h.cli.Update(ctx, grant); err != nil && !errors.IsNotFound(err) {
```
caseydavenport: A bare Update skips the two things `componentHandler` gives us: the conflict retry, and honoring `unsupported.operator.tigera.io/ignore` on a Role someone has taken over by hand. The cached read can lag the apiserver, so the Conflict is real, and we only tolerate NotFound here.

pkg/uigateway/uigateway.go
```diff
@@ -234,9 +250,14 @@ func (h *Helper) StaleComponents(ctx context.Context, desiredNS string) ([]rende
...
 // Teardown returns deletion components for every labeled Gateway namespace,
 // plus the backend namespace, which holds the Backend and ReferenceGrant.
-// No labeled Gateway means nothing to do: the Gateway is rendered first, so
-// nothing else can exist without one.
+// It first clears access finalizers whose resources are gone — a grant can
+// outlive its Gateway, so this runs even when no labeled Gateway remains.
+// After that, no labeled Gateway means nothing to do: the Gateway is
+// rendered first, so nothing else can exist without one.
 func (h *Helper) Teardown(ctx context.Context) ([]render.Component, error) {
+	if err := h.clearRBACFinalizers(ctx); err != nil {
```
caseydavenport: The sweep runs before the deletes it's meant to finish, so the pass that marks the grants can never clear them and we always need a second reconcile. Nothing guarantees one: the Gateway watch is registered asynchronously by `WaitToAddResourceWatch` and only once the CRD is served, so if it isn't attached when the Gateway goes away we wait on the periodic resync with the namespace stuck Terminating.

Same reachability problem for the early returns in `Components()` above the `StaleComponents` call. Set a bad `gatewayClassName` while grants are marked and they can't be reaped until the CR is fixed.

pkg/enterprise/uigateway/uigateway.go
```diff
@@ -13,17 +13,38 @@
...
+type Extension struct{}
+
+var _ extensions.UIGatewayExtension = Extension{}
+
+// New returns the UI gateway extension for Calico Enterprise.
+func New() Extension { return Extension{} }
+
+// ProxyObjects returns the WAF filter objects for the Manager gateway; other
+// components get nothing.
+func (Extension) ProxyObjects(resourcePrefix, namespace string) []client.Object {
```
caseydavenport: `go test ./pkg/enterprise/uigateway/...` reports no test files. If this prefix compare were inverted, Enterprise Whisker teardown would start deleting the GatewayAPI controller's per-namespace ServiceAccount and RoleBinding out of `calico-system`, which is the thing the guard exists to prevent, and nothing in CI would catch it. The fake extension in `uigateway_test.go` returns its objects unconditionally, so it doesn't cover this.

pkg/enterprise/controller/manager/manager_controller.go
```diff
@@ -941,7 +940,7 @@ func (r *ReconcileManager) resolveAdditionalTunnelCert(
 
 const (
 	ManagerGatewayTLSSecretName  = "calico-manager-gateway-tls"
-	ManagerGatewayResourcePrefix = "calico-manager"
+	ManagerGatewayResourcePrefix = euigateway.ManagerGatewayResourcePrefix
```
caseydavenport: This makes the Manager controller's own name for its resources an alias of a constant owned by a package it calls into, and the round trip exists only so the extension can compare the prefix back to `calico-manager`. The caller already knows which component it's serving.

We should put the component identity on the Config, or hand the helper the resolved objects the way `ExtraProxyObjects` did.

pkg/uigateway/uigateway_test.go
```diff
@@ -256,6 +258,90 @@ var _ = Describe("Cleanup helpers", func() {
...
+			for _, c := range components {
+				_, toDelete := c.Objects()
+				for _, obj := range toDelete {
+					if sa, ok := obj.(*corev1.ServiceAccount); ok {
+						deleted = append(deleted, sa.Namespace+"/"+sa.Name)
+					}
+				}
+			}
+			Expect(deleted).To(ConsistOf(backendNS+"/waf-http-filter"),
+				"only the backend namespace's component deletes the extension objects")
```
caseydavenport: The fixture has a single Gateway in the backend namespace, so there's only ever one deletion component and this can't fail. Dropping the `staleNS == bkNS` guard in the renderer leaves it green. A second labeled Gateway in another namespace would make it a real test.

pkg/render/gateway/component_test.go
```diff
@@ -110,6 +110,17 @@ var _ = Describe("Gateway component render", func() {
 			Expect(gw.Labels).To(HaveKeyWithValue(gateway.GatewayLabel, prefix))
 		})
 
+		It("stamps the access grants with the cleanup finalizer", func() {
```
caseydavenport: This only runs in the same-namespace context, where both grants land in `bkNS`. In the split layout the gateway grant goes to `cfg.GatewayNamespace` and nothing asserts the finalizer lands there, which is the configuration that breaks in `accessResourcesGone`.

pkg/render/gateway/component.go
```diff
@@ -536,8 +555,10 @@ func (c *gatewayDeletionComponent) Objects() (objsToCreate, objsToDelete []clien
 		ObjectMeta: metav1.ObjectMeta{Name: GatewayName(prefix), Namespace: staleNS},
 	})
 
-	// The grants go after the resources they permit deleting. The backend grant
-	// is dropped only by the backend namespace's own component.
+	// The grants carry RBACFinalizer, so these deletes only mark them; they
+	// hold the operator's write access until the uigateway helper clears the
+	// finalizer once the resources above are gone. The backend grant is
+	// dropped only by the backend namespace's own component.
```
caseydavenport: These are over our comment limit, and so are the `RBACFinalizer` block above and the doc comments on `Teardown`, `clearRBACFinalizers` and `accessResourcesGone`. 1-3 sentences and 35 words, cut to the claim that matters.

Line 56 also describes the change rather than the code ("once those resources no longer exist").

pkg/uigateway/uigateway.go
```diff
@@ -84,12 +85,22 @@ type Config struct {
...
-	return &Helper{cli: cli, cfg: cfg}
+// NewHelper returns the helper for one UI component's gateway. ext supplies
+// the variant's additions beside the proxy; nil adds none.
+func NewHelper(cli client.Client, ext extensions.UIGatewayExtension, cfg Config) *Helper {
+	return &Helper{cli: cli, ext: ext, cfg: cfg}
+}
+
+// proxyObjects returns the variant's additions beside the proxy.
+func (h *Helper) proxyObjects() []client.Object {
+	if h.ext == nil {
```
caseydavenport: The nil branch is dead in production. `extensions.UIGateway()` returns `noopUIGateway{}` when nothing is registered and is documented never to return nil, and both call sites go through it. It's here for the tests, which could pass `extensions.New(extensions.Set{}).UIGateway()` instead.

pkg/uigateway/uigateway.go
caseydavenport: To be clear on the above: the finalizer itself is what I asked for, it's the clearing I want to look like the rest of the operator.

The read half has to stay in the helper, since we need to find the grants and check whether the resources they cover are gone. But that can produce a bool on the deletion config, and the renderer emits the grant with or without the finalizer, the way `CanRemoveCNIFinalizer` does in `pkg/render/node.go`. The handler's update does the clearing, which also gets us the conflict retry and the ignore-annotation check for free.

caseydavenport:

```

```

PR #5295, opened 2026-09-01
build: bump Go to 1.27.0 and align the go-build image with k8s 1.37.0

electricjesus:
## Description

Moves the operator to `GO_VERSION=1.27.0` and `K8S_VERSION=v1.37.0`, where calico-private and calico OSS already are.

This unblocks the Envoy Gateway v1.9.1 bump, which stacks on top of this PR. EG v1.9.1 declares `go 1.26.7`. There is no `calico/go-build` image for 1.26.6 or 1.26.7, and the container runs `GOTOOLCHAIN=local`, so it will not fetch one. 1.27.0 is the only tag paired with a k8s1.37 image, so `K8S_VERSION` moves with it. `K8S_VERSION` only feeds `GO_BUILD_VER`; `ENVTEST_K8S_VERSION` stays at `1.34.x`.

The 1.27.0 image carries golangci-lint 2.13.1 instead of 2.12.2, which reports seven pre-existing staticcheck issues. They are not from this diff: linting clean `master` with the same image gives the identical seven at the same lines.

Three are fixed. The two namespace-migration requeues in `core_controller.go` move from rate-limited backoff to a flat 30s `RequeueAfter`, matching the other controllers; the old path already exceeded 30s per attempt after about 13 requeues. `PacketCaptureAPIPolicy` returns `NewPassthrough`, which is never nil, so its dead nil check is gone.

Four keep their behaviour and take a targeted `nolint` with a reason, matching ones already in the tree. Three are deprecated fields we still want: the prometheus-operator `BearerTokenSecret` pair, where moving to `authorization` would change the rendered ServiceMonitor, and the `FluentdDaemonSet` alias, which that test exists to cover. The fourth is the LogStorage status conflict retry, which wants to be prompt; a flat 30s would leave the conditions stale after a conflicting write.

`pkg/render/gateway/component.go` is gofmt-only. The 1.27.0 gofmt reindents multi-value returns of composite literals.

## Test plan

- `make static-checks`: **0 issues**. Reports 7 without this change.
- `make ut`: 109 suites, all passed, 10m28s.
- `go build ./...` and `go vet ./...` clean.

Not covered: nothing here runs against a live cluster. The migration requeue path needs an upgrade that actually triggers namespace migration, which CI does not do.

## Release Note

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -102,9 +102,9 @@ REPO?=tigera/operator
 PACKAGE_NAME?=github.com/tigera/operator
 LOCAL_USER_ID?=$(shell id -u $$USER)
 # The project Go version.
-GO_VERSION?=1.26.5
+GO_VERSION?=1.27.0
 # Version of Kubernetes to use for dependencies, tests, and kubectl.
-K8S_VERSION?=v1.37.0-beta.0
+K8S_VERSION?=v1.37.0
 # The version of LLVM to use for the go-build image.
 LLVM_VERSION?=21.1.8
 # Calico toolchain versions and the calico/go-build image to use.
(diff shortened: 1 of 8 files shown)


caseydavenport:

```

```

PR #5296, opened 2026-09-01
build(gatewayapi): bump bundled Envoy Gateway to v1.9.1

electricjesus:
## Description

Bumps the bundled Envoy Gateway from v1.8.3 to v1.9.1. That moves the embedded helm chart's Gateway API bundle from v1.5.1 to v1.6.1, and the vendored `sigs.k8s.io/gateway-api` with it.

Stacked on the Go 1.27.0 PR, which it needs: EG v1.9.1 declares `go 1.26.7` and no go-build image exists below 1.27.0 that satisfies it.

v1.9.1 splits `EnvoyGatewayKubernetesProvider` into two embedded structs, `EnvoyGatewayKubernetesInfrastructureConfiguration` and `EnvoyGatewayKubernetesConfiguration`. Field access still promotes, so `gateway_api.go` is unchanged. Only a composite literal has to name the embedded struct, which is the one test change here. The JSON tags did not move, so parsing an `EnvoyGateway` config is unaffected.

The v1.6.1 bundle adds one CRD, `XBackend` in `gateway.networking.x-k8s.io`. The name-suffix split in `gateway_api.go` already routes `x-k8s.io` into `k8sCRDs`, so it lands in the right bucket with no code change. Its operator RBAC lives in the calico repos, not here.

The v1.5.1 to v1.6.1 CRD change is additive: `TCPRoute` and `UDPRoute` gain `v1` and keep `v1alpha2`, nothing loses a served version. So the `safe-upgrades` ValidatingAdmissionPolicy does not block an in-place upgrade, unlike v1.7 to v1.8.

## Test plan

- `make ut`: 109 suites, all passed, 9m36s. That run includes the parent commit, so it covers the Go 1.27.0 move and the EG bump together.
- `go build ./...` and `go vet ./...` clean.
- Confirmed the regenerated `gateway-helm.tgz` is the published v1.9.1 chart. Its sha256 matches the registry manifest digest, and its `gatewayapi-crds.yaml` reads `bundle-version: v1.6.1`, `channel: experimental`.

Not covered: no live cluster. Two things want a real one before release. EG v1.9.1 moves EnvoyExtensionPolicy Lua from per-route to listener-level filters, which changes xDS filter names next to the coraza WASM filter. And `EndpointSliceIndex` now defaults on, which may raise controller memory.

## Release Note

```release-note
Update the bundled Envoy Gateway to v1.9.1, which requires the Gateway API v1.6 CRDs and raises the supported Kubernetes floor to v1.33. Review the Envoy Gateway v1.9.0 and v1.9.1 release notes for upstream changes: https://github.com/envoyproxy/gateway/blob/main/release-notes/v1.9.0.yaml
```


electricjesus:
Moved to https://github.com/projectcalico/calico/pull/14025 now that the operator lives in the monorepo.

PR #5322, opened 2026-09-18
Derive CNI address families from workload-capable IP pools

caseydavenport:
## Description

Bug fix. The address families in the rendered CNI config came from every enabled IP pool in the cluster, whatever that pool was allowed to be used for:

- One IPv6 pool reserved for something other than workloads turns on IPv6 assignment for every pod on a single-stack IPv4 cluster.
- Every sandbox then fails with "cannot find a qualified ippool, no pools match the required use (Workload)".
- It stays latent until the next calico-node restart rewrites the CNI config, so the cluster can run for days after the pool lands and die at the next reboot or upgrade.
- Recovery needs the IP pool's finalizer patched off by hand, since kube-controllers cannot get an address either.

A pool that serves only tunnels or load balancers says nothing about what an ordinary pod can get, so the CNI config now takes its address families from pools that allow `Workload`. A pool with no allowed uses set still counts, matching the documented back-compatibility default. The same filter applies to host-local, which picks single or dual stack ranges off the same list.

Unit tests in `pkg/render` cover both IPAM plugins, including the case where the only pool of a family is reserved for another use.

AI assistance: Claude Code wrote the fix and the tests. I reviewed it and am responsible for it.

Related: [CORE-13664](https://tigera.atlassian.net/browse/CORE-13664)

## Release Note

```release-note
Fixed pod networking breaking on single-stack clusters when an IP pool of the other address family exists that doesn't allow workloads.
```

[CORE-13664]: https://tigera.atlassian.net/browse/CORE-13664?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


caseydavenport:
Wrong repo - operator master lives in projectcalico/calico since the monorepo import. Reopening against operator/pkg/render there.

